PR #333, opened 2026-08-13
fix: the auto-nudge status-checks a blocked-on-the-user goal mid-oscillation

ksarma:
## The incident class

A goal blocked on the **user's** decision gets flipped to `working` by judge unblocks that don't actually answer the block — rulings like *"new work was filed on this branch"* or *"answered in passing"* — while the decision is still outstanding. The auto-nudge reads each flip as a stall and fires a status ask at a session that is waiting on a human; the closer then re-files essentially the same block minutes later. Five false status checks in one weekend fired in exactly that blocked↔working ping-pong window, including while the user was actively messaging the session about *other* threads (so "has the user replied since the block" doesn't discriminate).

## The fix

The fire list gains a third hold, keyed on the goal's own diary: **a goal whose newest judge row is an unblock, and whose diary already holds an earlier judge unblock, stands down until the closer's next word.** The discriminator was checked against all five live fires and the doctrine cases: every flap fire had a prior judge unblock on record; the legitimate cases had none.

Plumbing rides the hold system's own rules (the 2026-08-13 rework): `held` rows become `(goal, why, ev_t)` triples so each hold defers under its own why; `WHY_UNBLOCK_UNSETTLED` joins `WHY_IN_FLIGHT` (it presents as the Analyzing… swirl — the closer's pending pass is romp working the card, not a stall); and `_deferral_sweep_tick` gains its retirement case — the closer's next filed word — placed **above** the in-flight class branch, because no-judge-running is not this hold's event.

## Consequences to weigh (the part worth your review)

This hold deliberately trades a little nudge latency for not interrupting a human who is the actual blocker. The edges we checked, and one real tradeoff:

1. **The first-unblock doctrine is preserved.** A goal's FIRST block→unblock leaving it working still nudges — `test_awaiting_same_trigger_wedge` (your pinned case, whose own incident was a *wrongly-gagged* nudge) passes unchanged. The hold requires a *prior* judge unblock, so it can only engage once a goal has already ping-ponged.
2. **User gestures can't trigger or extend the hold.** Only judge rows count on both ends (`NUDGE_HOLD_EXEMPT_SRC`): a user's own unblock is the event that re-ARMS the nudge, and holding on it would invert it. Pinned in tests on both ends, plus the case where exempt bookkeeping rows land after the judge's unblock.
3. **No deadlock if the closer never speaks again.** The hold defers loudly (a deferral record with its own why), rides the standard 6h backstop, and the sweep also pops it when the goal leaves `working` for any reason. Worst case is a status ask delayed by 6h on a goal the judges abandoned mid-flap.
4. **The known tradeoff: the prior-unblock memory is unbounded.** A goal legitimately unblocked once long ago, then blocked and unblocked again much later, engages the hold on the second unblock even if no flap is happening *now*. In practice the cost is small — the hold releases on the closer's very next filed word on that goal (usually the next judge pass), and the backstop caps it — but a time- or count-bounded discriminator would be a coherent alternative if you'd rather not hold on ancient history. We chose the diary-only form because it's exactly event-based (no tunable window) and because every real fire we audited had its prior unblock minutes-to-hours earlier, not weeks.
5. **Presentation:** cards under this hold show the Analyzing… swirl rather than the yellow stalled chip. If you'd rather these read as their own state, the why is already distinct — routing is one branch.

## Tests

`tests/test_nudge_unblock_hold.py` (10 tests) rebuilds the incident shapes synthetically: the flap holds; the first unblock fires; the closer's next word releases; a re-block is dropped not held; user unblocks count on neither end; exempt rows don't release; the ev_t hold still wins first with its own why; the class membership and sweep-case ordering are pinned. The existing held-shape consumers (`test_nudge_hold_judge_rows`, `test_nudge_injected_turn_arm`, `test_interrupt_lift_evidence_time`, `test_kernel_nudge_last_resort`) are updated for the triple and all pass, as do `test_stall_note` and `test_state_isolation_order`.

Running on a fork since 2026-08-11 (pre-rework form) and since 2026-08-13 in this exact form; the false status checks stopped with it.

### DIFF
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -269,6 +269,16 @@ _WHY_JUDGING_LEGACY = "a judge pass is mid-flight"
 # INSPECTABLE: this drop used to leave NO record anywhere, which is how a card sat in Working for half an
 # hour with nothing to read (see the kernel's _nudge_fire_list).
 WHY_TURN_IN_FLIGHT = "a judge has ruled on a turn that hasn't finished yet"
+# The fire list's THIRD hold (2026-08-11): a goal whose diary ENDS on a judge's UNBLOCK, with an
+# EARLIER judge unblock already on record, is mid-OSCILLATION — the column has ping-ponged
+# blocked↔working at least once without settling, and the closer's next word is pending. Five false
+# status checks in one live weekend fired in that window: a blocked-on-the-user goal was repeatedly
+# flipped to 'working' by "new work filed" / "answered in passing" rulings while the user's decision was
+# still outstanding, and the nudge read each flip as a stall. A goal's FIRST unblock never holds — that
+# is the 2026-07-30 considered-verdict case, whose own incident was a wrongly-gagged nudge. Same
+# never-paint rule as the holds above; clears on the closer's next word (any newer judge row on the
+# goal) or the deferral backstop.
+WHY_UNBLOCK_UNSETTLED = "an unblock is awaiting the closer's next word on this goal"
 # The CONSOLIDATOR (the user 2026-06-19): the grouper's twin for the COMPLETED column. The working grouper
 # only ever sees OPEN tops, so related goals that finish before they get
 # grouped land as separate cards. The consolidator groups related ALL-COMPLETED sibling tops under a
@@ -8101,7 +8111,11 @@ def stall_llm(goal_text, work_text, holding):
 # so one frozen between retries showed nothing at all — six live records were dark up to 20 hours.
 # Now the kernel's deferral sweep retires each record on its reason's own event, and whatever stands
 # presents somewhere by definition.
-WHY_IN_FLIGHT = (WHY_JUDGING, _WHY_JUDGING_LEGACY, WHY_TURN_IN_FLIGHT)
+# The unblock-unsettled hold is in-flight for the same reason the turn hold is: the closer's next
+# pass is romp's own review mid-flight, not a state the user acts on. Its sweep case retires it on
+# the closer's next filed word (kernel _deferral_sweep_tick, ABOVE the class branch — the class
+# branch's no-judge-running event would retire it early).
+WHY_IN_FLIGHT = (WHY_JUDGING, _WHY_JUDGING_LEGACY, WHY_TURN_IN_FLIGHT, WHY_UNBLOCK_UNSETTLED)
 
 
 def stalled_facts(fsid):
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -3371,6 +3371,8 @@ def _deferral_sweep_tick(now):
       * the goal left plain 'working' (resolved/cleared/gone) — pop, whatever the why;
       * WHY_TURN_IN_FLIGHT — the held-on evidence's turn has ENDED: newest ended-turn watermark
         >= the record's evT (stamped at mint from the offending diary row);
+      * WHY_UNBLOCK_UNSETTLED — the closer's next word: the goal's newest judge row is no
+        longer an unblock;
       * WHY_JUDGING (+ legacy) — no judge call in flight for the session (jd.active_runs, the
         registry that deregisters in a finally);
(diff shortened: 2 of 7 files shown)


henrypinkard:
Closing this in favor of a smaller upstream fix that removes the need for a new hold entirely.

The flap you're targeting is real: during a blocked↔working oscillation the auto-nudge can status-check a session that is actually waiting on the user. Your guard handles it by deferring the nudge under a new `WHY_UNBLOCK_UNSETTLED` reason. The concern is that the discriminator ("an earlier unblock exists") is unbounded, so a goal that flapped once long ago can hold its nudge up to 6h later.

The cleaner fix is at the writer, not the nudge. The flap is manufactured when the closer re-blocks a just-lifted node by anchoring its block to the unblock's own evidence time (equal timestamp, later filing wins the fold tiebreak). romp already refuses a judge block that isn't strictly newer than the user-reply floor (equality voids); extending that same floor to also cover the node's newest judge unblock makes the re-block stand down at the source, so the card never flaps and no nudge-side hold is needed. That deletes a mechanism rather than adding one.

I'm writing that now and it will land as its own PR. Thanks for the thorough repro and tests here — they made the root cause easy to confirm.

henrypinkard:
Reopening this — I owe you a correction.

I closed this saying a smaller upstream (writer-side) fix would replace it. I implemented that fix (floor a judge re-block on the node's newest lift's evidence time, so a re-block that isn't strictly newer than the lift stands down) and the **test suite refuted it**. It breaks the interrupt/resume path, which by design stamps the lift AND the legitimate follow-up block at the *same* evidence time: in `test_interrupt_lift_evidence_time.py`, a resumed cut turn is unblocked, then its new question re-blocks the card at that same `RESUME_T`, and that block must land. The flap and a legitimate re-block are the same mechanism at identical timestamps, so there's no evidence-time signal to separate them at the writer. Four tests broke; I reverted it.

So your nudge-side approach here is the right one, and I was wrong to supersede it. The one thing worth changing is the discriminator we flagged as unbounded ("an earlier unblock exists anywhere in the goal's life"), which can hold a nudge up to 6h late on a goal that flapped once long ago. Bound it to the **adjacent** cycle instead: hold only when the newest judge row is an `unblock` AND the row immediately before it is a `block`. That removes the edge with no time window and no unbounded memory — it reads only the last two judge rows, so a lone historical lift no longer arms the hold — and it keeps your `WHY_UNBLOCK_UNSETTLED` mechanism and its sweep-release intact; it just tightens when it arms.

Happy to pair on that, or if you'd rather I push the one-line discriminator tweak to this branch, say the word. Thanks for the original repro and tests — they're exactly what let me confirm the writer-side dead end quickly.

henrypinkard:
Correction to my last comment: **disregard** the "bound it to the adjacent block→unblock cycle" suggestion. I was wrong, and your discriminator is right as written.

I checked it empirically. Bounding to the adjacent cycle would fire the hold on a goal's FIRST unblock — block→unblock is the normal adjacency, since an unblock lifts a block — which is exactly the case your `test_a_goals_first_unblock_still_fires_the_pinned_doctrine` asserts must NOT hold (the 2026-07-30 considered-verdict doctrine). Your "an earlier judge unblock already on record" test is precisely what separates a first unblock (nudgeable) from a repeat (the flap), and only judge unblocks counting on both ends keeps an interrupt re-engage from arming it. I ran the suite at your head: all green, and it leaves `test_interrupt_lift_evidence_time` untouched (my writer-side replacement broke that one — it can't tell a lift from a legitimate re-block, since the interrupt path stamps both at the same evidence time).

The only residual is the one you already handle loudly: a goal with ≥2 judge unblocks whose newest unblock sticks and whose session then goes fully quiet has its nudge deferred to the 6h backstop. That's a bounded responsiveness cost, it errs in the safe direction (better than a false status-check on a blocked-on-the-user session), it presents as in-flight rather than a stall, and the closer's next word releases it in the common case. I'd merge this as-is.

Sorry for the runaround.

PR #336, opened 2026-08-13
fix: an unreadable session state renders as nothing, exactly like a healthy idle one

ksarma:
## The gap

The status pips speak two words: working (gold dot) and awaiting-bg (straw dot). Every other session state renders as **nothing** — including states the kernel has read and knows (a session whose latest state record is `waiting` draws no pip at all). A knowable state encoded as a blank is indistinguishable from a rendering hole, which is exactly how a real rendering bug hid on a fork for a while. The sessions pane is worse: its dot is working-only, so the same session shows a straw dot on the feed and nothing in the pane beside it.

## The fix

`build_feed` emits the other two quarters of a **total partition**: `ready` (alive and quiet — live state read, nothing running) and `stateUnknown` (listed while the live state could not be read). Every session the feed lists lands in exactly one of the four lists. All three surfaces — feed, sessions pane, chat tab strip — render the four states explicitly:

- working — gold dot (unchanged)
- awaiting-bg — straw dot (unchanged)
- ready/idle — a **hollow steel ring** (hollow = at rest, filled = active)
- state unknown — an **explicit gray ring** (fail loudly: no-information renders as itself, never as an empty spot)

Each pip explains itself on hover. The tab strip's ladder keeps the OPENING pulse as its own branch, and states with their own designed treatments (blocked's dashed ring, compacting's bar, closed's strikethrough) deliberately keep them — the pip encodes turn state, not an attention taxonomy.

**Degradation is deliberate:** a bare name is reserved for payloads that predate the lists. An old kernel — including an old *remote* kernel in a federated merge, whose sessions ride in none of the lists (federation concatenates `ready`/`stateUnknown` per host like `working`/`awaiting`) — keeps today's look instead of reading falsely as "unknown".

## Tests

`tests/test_feed_status_lists.py` (partition totality, `hideFromFeed` parity with the cards, old-payload degradation), `tests/test_tab_strip_pips.py` (the ladder, and the mobile scrape contract that reads `.tab-dot.await`), `ui/webview/feed-status-pips.test.ts` (dotFor ranking, retint-in-place on re-render, the federation ride-along). The two existing pins that quoted the old dot code (`awaiting-state.test.ts`, `provisional.test.ts`) follow it. Extension suite: 1881/1881, typecheck clean.

Running on a fork since 2026-08-09 (the tab strip since 2026-08-10); the empty-pip reports stopped with it.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -15215,6 +15215,27 @@ def _boundary_clear_notices(alive):
     return out
 
 
+def _state_unknown_names(alive, tmux, working, awaiting):
+    """Sessions the feed LISTS but whose live state it could not read — no row in the merged live
+    map (e.g. the headless file-derived fallback). The client draws these a gray ring, so a blank
+    pip goes back to meaning exactly one thing: alive and quiet.
+
+    Without this list the two cases are the same nothing, which is how a rendering hole hid: a
+    session whose state was unreadable looked identical to one that was simply idle. Working and
+    awaiting sessions are excluded by construction — their state was read. Mirrors build_feed's own
+    filters, so a hideFromFeed session is in no list, matching its absence from the cards."""
+    out = []
+    for s in alive:
+        if _session_flag(s["sid"], "hideFromFeed"):
+            continue
+        nm = s["name"]
+        if nm in working or nm in awaiting:
+            continue                                  # state read: it is working or awaiting
+        if tmux.get(s["sid"]) is None:
+            out.append(nm)
+    return out
+
+
 def build_feed(now, tmux=None):
     """The {type:"feed"} message the tuned feed.js bundle consumes (ui-parity.md: feed = ADAPT).
     Goals map onto the AskItem/AskTreeNode shape the render already speaks: the goal tree IS the
@@ -16016,6 +16037,9 @@ def build_feed(now, tmux=None):
         _a["notify"] = True if _notify_card_effective(_ncards, _a["itemId"], str(_a.get("sid") or "")) else None
     return {"type": "feed", "asks": asks, "now": now,
             "working": working, "awaiting": awaiting,   # awaiting = idle-but-waiting-on-bg-work names → straw dot (the user 2026-07-13)
+            # listed-but-unreadable names → an explicit gray ring, so a BLANK pip means "alive and
+            # quiet" and nothing else (see _state_unknown_names)
+            "stateUnknown": _state_unknown_names(alive, tmux, working, awaiting),
             # session name -> live judge-classified SERVICE descs (a dev server the session keeps around;
             # _bg_split) → the grouped-mode session header's neutral chip, never a waiting state (2026-07-24)
             "bgServices": bg_services,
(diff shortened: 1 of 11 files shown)


henrypinkard:
Design call after looking at this: a pip should mark something happening or something wrong, and no pip should keep meaning "alive and quiet". Concretely:

- Drop the ready ring on all three surfaces. A healthy idle session stays unmarked, and a blank goes back to meaning ready.
- Keep the existing working and background-work dots as they are.
- Keep the gray unreadable-state ring. That case is the real win of this PR: it is exactly where a blank used to hide breakage, and once it renders loudly, an unmarked session is trustworthy again.

Colors check out as they stand: #2b7fb8 is exactly --st-ready-bg and the timeline ready chip (BADGE in ui/romp-timeline-view.js), and #8a8a8a matches the network strip down-host gray (strip.ts:533). After the ready ring goes, the gray is the only new pip color; in styles.css reference the palette (a var, or a comment pointing at the strip gray) rather than a bare hex. The hardcoded values in feed.css and fleet-pane.css are fine since those sheets load standalone.

Wording nit for the same pass: a couple of the added comment lines use "the fleet" as a noun (ui/webview/fleet.ts and the new test). Say "the sessions" there; mechanical reuse of existing identifiers like FLEET_CSS is fine.

Tests will want the flip too: ready should assert no pip. Everything else read clean, thanks for the thorough coverage.

ksarma:
Took the design call — rebuilt as a single commit off current `main` with the ready ring gone.

**What changed from the first version**

- **Ready ring dropped** on all three surfaces, and the `ready` list is gone from the payload entirely. A quiet session needs no entry, and its absence is what a blank already says. A nice side effect: the old-kernel degradation worry disappears with it — a host too old to send `stateUnknown` contributes nothing, so its sessions stay blank rather than reading falsely as unknown. No special-casing needed.
- **Gray unreadable-state ring kept** — the case that was genuinely invisible.
- **`styles.css` names its source** for the gray (a comment pointing at the strip's down-host pick) instead of a bare hex; `feed.css` / `fleet-pane.css` keep literals since those sheets load standalone, per your note.
- **Wording fixed** — the added comments say "the sessions"; `FLEET_CSS`-style identifiers left alone. There's now a test asserting that, so it can't drift back.
- **Tests flipped**: a readable idle session asserts *no* pip and no `.ready` rule survives in any sheet.

**One thing I added, and the argument for it — say the word and I'll drop it**

The sessions pane now also draws the straw **awaiting** dot. That reads like scope creep against "keep the existing dots as they are", but it's the same principle you set: the feed has shown that dot since 2026-07-13 and this pane never did, so an awaiting session is blank *there* and reads as idle. A blank can't be trustworthy while a real state is still hiding behind it — the gray ring fixes one leak of that kind and this fixes the other. Happy to pull it into its own PR if you'd rather keep this one to exactly the unreadable case.

Verified: extension suite **1940 pass / 0 fail**, `tsc --noEmit` clean, `npm run build` clean, and the Python side green (`test_feed_status_lists.py`, `test_state_isolation_order.py`).

PR #337, opened 2026-08-13
fix(security): a same-site loopback page rides the cookie into an authenticated WebSocket

ksarma:
**DRAFT — needs live client verification before merge (see bottom).** Chain 1 of 5 from a security re-verification against v0.8.0. Each chain is a separate PR.

## The hole

Cookies are host- not port-scoped (RFC 6265 §8.5), so every `http://127.0.0.1:<port>` page on the machine — e.g. a dev server in a repo an agent cloned — is **same-site** with the dashboard, and the browser attaches the `romp_token` cookie to its requests, **including the WebSocket handshake** (which CORS does not cover). `_authorize` accepted a valid cookie *before* it ever consulted the Origin gate (`kernel.py`), so that page could open an authenticated `/ws`, stream every session, and drive `sendMessage` / `answerAsk` (approve a pending permission prompt) / `forkSession` — full control of the romp instance and, through injected prompts + approvals, the host.

SECURITY.md already states *"An Origin check additionally protects the browser surfaces against cross-site requests, including the WebSocket upgrade"* — this makes the code match that promise.

## The fixes

1. **Cookie no longer bypasses Origin.** The cookie branch of `_authorize` now also requires `_origin_ok()`. An **explicitly presented** `?token` / `X-Romp-Token` still bypasses Origin (federation needs it; a drive-by page can't obtain one, it rides only the cookie). Absent-Origin (same-origin nav, curl) and `vscode-webview://` still authorize, so no shipped client regresses.
2. **nosniff** on every response (`_send`) — a response whose bytes look like markup can otherwise be sniffed and run as HTML on the dashboard's origin.
3. **X-Frame-Options: SAMEORIGIN + CSP frame-ancestors 'self'** on every response — clickjacking: the authenticated dashboard must not be framable by the same loopback attacker. (This finding was noted-but-deferred in the original audit; authored fresh here.)
4. **The `/remote/<host>/file` relay derives its own Content-Type** from the requested extension (`_PREVIEW_MIME`, same 404 as the local route) and discards the remote's — a compromised remote answering `text/html` for a `.pdf` the lightbox opens in a same-origin unsandboxed iframe was script on the dashboard's origin.

## Tests

`tests/test_kernel_auth_hardening.py` gains the cookie-vs-origin cases (denied from a foreign loopback port and from offsite; still authorized with absent-Origin / same-origin / `vscode-webview://` / explicit token) plus source-pins for the nosniff + frame headers and the relay's local-mime derivation. `test_kernel_remote_file_relay`, `test_kernel_ws_auth`, and `test_state_isolation_order` pass unchanged.

## Why this is a draft

The framing header and the cookie/Origin tightening **can't be verified offline against the real clients**: the VS Code webview loads the dashboard from a `vscode-webview://` origin, and phone/tailnet access must keep working. The `frame-ancestors 'self'` policy in particular may need a webview allowlist entry. These need confirming on every shipped client before merge — happy to adjust once maintainers weigh in on which client surfaces must keep framing / cross-origin cookie behavior.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -21614,6 +21614,19 @@ class Handler(BaseHTTPRequestHandler):
         self.send_response(code)
         self.send_header("Content-Type", ctype)
         self.send_header("Content-Length", str(len(body)))
+        # Take the declared type at its word. Without this a browser may sniff a response whose
+        # bytes look like markup and run it as HTML on the dashboard's own origin, which is the
+        # same escalation the /remote/…/file relay's own type check closes from the other side.
+        self.send_header("X-Content-Type-Options", "nosniff")
+        # Framing gate (clickjacking): the authenticated dashboard must not be loadable in a
+        # cross-origin frame, or a hostile same-site loopback page (the same attacker the cookie
+        # Origin gate below defends against) could frame it and drive it invisibly. SAMEORIGIN keeps
+        # the dashboard's own same-origin sub-frames (/chat, /feed) working. NOTE: the VS Code
+        # webview loads the dashboard from a vscode-webview:// origin — if that surface is still
+        # needed it must be added to a frame-ancestors allowlist; confirm against every shipped
+        # client (webview / phone / tailnet) before relying on this.
+        self.send_header("X-Frame-Options", "SAMEORIGIN")
+        self.send_header("Content-Security-Policy", "frame-ancestors 'self'")
         for k, v in (headers or {}).items():
             self.send_header(k, v)
         if cache:                                     # e.g. "no-cache" — keeps a tab from running a stale bundle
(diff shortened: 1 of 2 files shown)


henrypinkard:
Traced this from first principles: it closes a real, previously-open surface, and the VS Code fear in the draft note rests on a false premise, so it can land as-is.

**The hole is real.** The `romp_token` cookie is host-scoped, not port-scoped (RFC 6265), so a page on any other `127.0.0.1` port is same-site and the browser auto-attaches the cookie to a `ws://127.0.0.1:<dashport>/ws` handshake. On main the cookie branch of `_authorize` returns authorized *before* `_origin_ok()` runs, so that foreign-origin socket authorizes — full read plus the whole drive vocabulary (`sendMessage`, `answerAsk`, `endSession`, `setAuth`, ...). Token enforcement didn't cover it because the cookie *is* the token, auto-presented by the browser. The one-line fix (AND `_origin_ok()` onto the cookie branch) closes it; `?token`/`X-Romp-Token`, absent-Origin, same-origin, and `vscode-webview://` all still pass, so no shipped client regresses. The `nosniff` and relay-MIME extras are both sound — the relay-MIME one is a genuinely distinct federation surface (a remote answering `text/html` for a `.pdf` runs as script in the same-origin preview iframe).

**VS Code is not broken by `frame-ancestors 'self'`.** The draft caveat assumes the panes load the dashboard as `<iframe src="http://127.0.0.1:…">`. They don't: each pane is an extension-authored webview that loads its bundle via `asWebviewUri` and reaches the kernel only over `connect-src` (fetch + WebSocket). `X-Frame-Options`/`frame-ancestors` gate document framing only — they have zero effect on fetch/XHR/WebSocket, and WS handshakes are never subject to XFO. Phone/tailnet frame the same kernel origin, which `'self'` permits. So keep `'self'` tight (no `vscode-webview:` allowlist needed) and the headers ship safely.

Adversarial bypass checks (Origin-less/native clients; the `/remote` federation relay) both come back closed: every authorizing path still requires the secret token, which a browser drive-by can't read (HttpOnly) or forge, and the relay sits behind the same `_authorize`. Only nit: the new `_send` comment warns of a VS Code break that can't happen — worth correcting. Recommend un-drafting and landing.

PR #340, opened 2026-08-13
fix(security): the VS Code extension execs an install path taken off the auth-exempt /version

ksarma:
**DRAFT — needs live VS Code verification before merge (see bottom).** Chain 2 of 5 from a security re-verification against v0.8.0.

## The hole

`updateExtension()` reads `rompDir` from the kernel's `/version` — an **auth-exempt** route — and runs `execFile("bash", [<rompDir>/vscode-extension/install.sh], { cwd })`. So whatever answers the kernel port (a local process that grabbed it before the real kernel, say) gets to choose the directory a shell command runs in, and the same `/version` keepalive `dv` raises the "newer build" prompt that invites the click. A path that arrives over a socket is not a path to run.

## The fixes

1. **The install target is resolved locally** (new `vscode-extension/src/update-target.ts`): this VSIX's own `context.extensionPath`, else `ROMP_DIR` from the extension host's own environment — both set by whoever launched VS Code. A candidate must look like a real **checkout** (both `install.sh` *and* `esbuild.js` present; a packaged `.vsix` ships the first but not the second via `.vscodeignore`), so a packaged install resolves to nothing and the extension says so and points at the terminal, rather than shelling out to a guess.
2. **`rompDir` is dropped from `/version`.** With #1 it has no reader left, and a repo path ($HOME-collapse hides only the username, never the location or the project name) has no business on an auth-exempt route.

`defaultDir` / `nativeDialogs` still ride `/version` because the gear reads them there — they *also* travel the authenticated `sessionList` message, and repointing the gear onto that is the clean follow-up that lets `defaultDir` leave this payload too. I left that out here deliberately: it's a gear refactor, not a security fix, and doesn't belong bundled into a security PR. Happy to send it separately if you'd like.

## Tests

- `update-target.test.ts` — the resolver: requires both checkout markers, prefers `extensionPath` over `ROMP_DIR`, returns null when this copy isn't a checkout.
- `build-drift.test.ts` — `updateExtension` resolves locally and no longer reads `info.rompDir` / `fetchJson("/version")`.
- `test_kernel_cachebust.py` — `/version` carries no repo path.

Extension suite 1883/1883, typecheck clean, bundles build.

## Why this is a draft

The resolver change alters how the extension finds its own install dir. It needs confirming on a **real VS Code host** that a genuine checkout still resolves and one-click self-update still works (and that a packaged-VSIX install shows the terminal hint) — none of that is verifiable offline.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -502,8 +502,17 @@ def _kernel_ver():
 def _version_info():
     """What this kernel is running — code sha + per-bundle build mtimes + pid/uptime. Lets the feed's
     settings gear / `romp version` / a curl tell at a glance whether the browser is on a stale bundle
-    (compare the served ?v= against bundles[].mtime here). Any path here is $HOME-collapsed for privacy
-    (like defaultDir/rompDir) — never a raw /Users/<name> path."""
+    (compare the served ?v= against bundles[].mtime here).
+
+    /version is AUTH-EXEMPT — any local process reads it with no token, and it rides the remote poll
+    between machines — so it must reveal nothing that grants leverage. `rompDir` used to ride here,
+    $HOME-collapsed: the VS Code extension read it and turned it into an `execFile("bash", …)` target,
+    so whatever answered this port chose the directory a shell ran in (the extension resolves its own
+    install path now — see vscode-extension/src/update-target.ts — and never asks). A $HOME-collapse
+    hid the username, never the checkout's location or name, so the path had no business on an
+    auth-exempt route at all. (`defaultDir`/`nativeDialogs` still ride here because the gear reads
+    them here; they also travel the authenticated sessionList, and moving the gear onto that is the
+    clean follow-up that lets `defaultDir` leave this payload too.)"""
     bundles = {}
     try:
         for p in sorted(DIST.glob("*.js")) + sorted(DIST.glob("*.css")):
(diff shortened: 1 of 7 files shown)


henrypinkard:
The core security fix here is right and carries over unchanged, but two things block this branch as-is, one of them a privacy issue that needs a history rewrite rather than a follow-up commit:

1. It commits a `vscode-extension/node_modules` symlink whose target is an absolute local home path. That leaks a username/home path into history permanently once merged — deleting it later is not enough, the branch has to be rebuilt so the path never enters history at all.
2. `driftNotice()` / `COPY_ACTION` are added and thoroughly tested here but never wired into `maybeBuildNotice`, which still offers "Update extension" unconditionally — a button that errors on any packaged VSIX (the ordinary install). The tested-but-unwired path is effectively dead code.

I built a cleaned-up version on branch `henrypinkard:sec-vscode-fix`: a single commit off current `main` (so no leaked path is in history), `driftNotice` wired into `maybeBuildNotice` so a non-checkout install is offered "Copy install command" instead of a failing button, and your local-resolution fix (drop `rompDir` from the auth-exempt `/version`; resolve the install script behind both checkout markers) intact. Green locally on all four CI gates (pytest, `tsc --noEmit`, `node --test` 1891 pass, `npm run build`). Suggest landing that in place of this branch.

henrypinkard:
Opened that cleaned-up version as #352 (single commit off current main, no leaked path in history, `driftNotice` wired in, all CI gates green). Suggest landing #352 in place of this.

henrypinkard:
Closing in favor of #352, which is the same core fix rebuilt clean (no leaked home path in history, driftNotice wired in, all CI gates green). Merged as of now.

PR #341, opened 2026-08-13
Hive: a game-feel 3D command view of your sessions

jfonsecarivera:
A fifth dashboard pane — a Tron-styled 3D honeycomb, one hex pad per session, a
cute bean character on each pad acting out that session's live state. Click a
hex: the camera flies in and a pane-local card shows the gist (state in plain
words, current goal, needs-you decision brief) with a composer that delivers
into that session and an Open session jump. A quiet ghost hex on the spiral's
first free slot opens the new-session picker.

## How it rides the existing machinery
- `app=hive` is a feed-payload rider (three one-line additions to the push
  loop); the pane consumes `ledgers` + `asks` — no new payload builders.
- The scene mutates ONLY on diff events from a pure model (`hive-model.ts`):
  identical pushes produce zero motion (tested), goal-done celebrations fire
  once per observed transition, hex slots are stable for a session's life
  (persisted) so spatial memory works.
- The talk composer posts the same `sendMessage` drive op the chat uses;
  foreign-sid refusals surface on the card via the err channel.
- Status colors keep their standing meanings (pinned against styles.css by
  test); the accent is reserved for selection/hover/ghost chrome.
- Click-safety: the card is built once and updated in place, actions delegated
  (`actions.ts`); canvas clicks acknowledge on pointer-down with a pad dip.
- Rendering pauses fully when the pane is hidden (IntersectionObserver +
  visibilitychange); DPR capped at 2.

## Found along the way
A backward rAF clock step (laptop sleep, tab restore, virtual time) flipped
every exponential camera ease into a runaway AWAY from its target. `frameDt()`
clamps dt into [0, 50ms]; tested.

## Tests
`hive-layout` (spiral/slot invariants, frameDt), `hive-model` (every chip,
no-flap invariant, goal-done semantics, state lines), `hive-wiring` (pins the
kernel route/landing strings, esbuild entries, palette parity, talk protocol).
Full suite: 1901 passing. `npm run typecheck` clean. three.js (MIT) joins the
webview bundle via the existing esbuild pipeline.

Design doc: `plans/hive.md`. Screenshotted end-to-end via a headless-Chrome
harness (synthetic sessions only).

### DIFF
--- a/bin/romp
+++ b/bin/romp
@@ -948,6 +948,29 @@ if [[ "${1:-}" == "default-dir" ]]; then
     exit 0
 fi
 
+# `romp default-host [NAME]` — the machine NEW sessions are created on by default. Persisted to
+# ~/.config/romp/default-host and read by the kernel's _default_create_host(), which ships it to the
+# client: the + picker preselects that machine in its Host row and a hive tray drop lands there. No arg
+# prints it; an empty arg ("") clears it (back to this machine). Read live — no restart.
+if [[ "${1:-}" == "default-host" ]]; then
+    _dh_file="$HOME/.config/romp/default-host"
+    if [[ $# -lt 2 ]]; then
+        if [[ -s "$_dh_file" ]]; then cat "$_dh_file"; else echo "(unset — new sessions land on this machine)"; fi
+        exit 0
+    fi
+    if [[ -z "$2" ]]; then
+        rm -f "$_dh_file"; echo "default new-session host cleared (new sessions land on this machine)."
+        exit 0
+    fi
+    if [[ ! "$2" =~ ^[A-Za-z0-9._@:-]+$ ]]; then
+        echo "romp default-host: not a host name: $2" >&2; exit 1
+    fi
+    mkdir -p "$(dirname "$_dh_file")"
+    printf '%s\n' "$2" > "$_dh_file"
+    echo "default new-session host set to: $2"
+    exit 0
+fi
+
 # NB: `romp --serve` (a persisted 0.0.0.0 opt-in for direct tailnet binding) was
 # REMOVED 2026-07-19: phone/tailnet reach goes through `tailscale serve` proxying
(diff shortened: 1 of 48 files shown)


henrypinkard:
Really like this — it's a genuinely fun surface and the engineering under it is solid (it reads the authoritative feed over the WS rather than scraping, and it's diff-driven). Two things before it can move forward, plus a note on where it's headed.

**1. It should ship as romp's first *plugin*, not merge into core.** A plugin-system design landed on `main` (`plans/plugins.md`): a plugin is a directory + `manifest.json` served into a sandboxed pane that reads the same feed you already consume and issues manifest-scoped drive ops. Hive maps onto it almost 1:1. We're holding off on building the plugin host itself until you settle what you want Hive to actually do and show — once that's firm, we build the host and you rebase the 3D view onto it (no more hardcoding a pane into the landing shell).

**2. Please split the non-Hive changes into their own PRs.** Bundled in here are several core changes unrelated to the 3D view, and they can't really be evaluated while buried in a viewer PR — a couple of them look genuinely useful on their own:
- session-create defaults (a per-host default model/effort/perm applied to every new session)
- the `default-host` picker
- the "hookless host card" in `build_feed`
- the interrupt-chip fix
- `bin/romp` auto-start (bare `romp` boots the stack)

Each as its own PR would let them land on their merits.

**3. Housekeeping.** It currently conflicts with `main`, and CI never ran (fork Actions didn't fire), so nothing here is verified yet. Minor: `three` (~600KB) is bundled, and `@types/three` is listed as a runtime `dependency` (should be a `devDependency`).

On emojis: that's a romp-*core* rule only. Once Hive is its own plugin, use whatever glyphs you like in its own surface — no need to strip the ✓/❗ from the plugin itself, just from anything that would land in core.

Net: park the 3D view here, spin the core bits out into their own PRs, and let's line up the plugin host once the Hive scope is settled.

jfonsecarivera:
Superseded — Hive will land as the first plugin per plans/plugins.md.

PR #379, opened 2026-08-14
Run the judge and the announcer from a private dir, not /tmp

ksarma:
`kernel/judge.py` already runs its model calls with `--safe-mode`, and its own docstring says why: it *"drops auto-discovered CLAUDE.md/memory + skills + hooks"*. `hooks/romp-summarize.sh` runs the same CLI without it, from `cd /tmp`. This PR makes the announcer consistent with the judge, and moves both helpers' working directories out of `/tmp`.

The two halves ship together on purpose, because separating them is self-contradicting: arguing "a helper's cwd must not be world-writable" while `JUDGE_SCRATCH = "/tmp/romp-judge"` sits in the tree is not an argument anyone should accept.

### 1. `--safe-mode` on the announcer

The summarizer gates tools (`--tools ""`) and MCP (`--strict-mcp-config` + empty config), but nothing stops a *discovered* `settings.json` from running hook commands. From a world-writable cwd, another local account can plant `/tmp/.claude/settings.json` and have its commands run as you.

Being straight about the preconditions, because they matter and a reviewer will find them:

- the hook is **off by default** — it needs `~/.claude/romp-summarize-on`, and the feature is already marked deprecated in its own header;
- it is tmux-only (`exit 0` with no `$TMUX`);
- there is a recursion guard.

So this is not "every prompt on every install". For anyone who *has* opted in, it is once per prompt and once per stop.

I also cannot demonstrate the last mile from this repo: whether headless `claude -p` really executes a cwd-discovered `.claude/settings.json` without a trust prompt is not something I can prove here. The argument rests on your own docstring and your own test (`tests/test_judge.py`), which I think is enough for a defense-in-depth change — but I am not presenting it as a working exploit.

### 2. Both cwds move under the state root

`$STATE/summarize` and `$STATE/judge-scratch`, created `0700`, with the same `${ROMP_STATE_DIR:-${XDG_STATE_HOME:-$HOME/.local/state}/romp}` idiom already used verbatim in `hooks/romp-postal-revive.sh` and `hooks/romp-wake.sh`. `_ensure_judge_scratch()` adds an `lstat` + `S_ISDIR` + owner check and repairs loose modes, refusing rather than proceeding if the directory is not ours.

**The strongest reason for this half is one I did not expect.** Nothing is ever written into the judge scratch — it is a bare cwd, and judge transcripts land under `~/.claude/projects/`. But `prune_judge_scratch` **realpaths** `JUDGE_SCRATCH` at every kernel boot and unlinks day-old `*.jsonl` in the project directory it derives from that path. A symlink planted at the guessable `/tmp/romp-judge`, pointed at a directory the user actually works in, turns romp's own housekeeping into a deleter of their real Claude Code transcripts. That is the concrete impact, and it is what the ownership/symlink guard closes.

When the scratch refuses, the judge call is skipped, logged once per reason as an `err == "scratch"` row, and `_judge_ctx.paused` is set — so a permissions problem is not counted by the give-up counters. Without that flag three refusals blank a card summary to the empty sentinel, which is unrepairable loss rather than a visible failure.

### A test that was asserting the implementation against itself

`tests/romp-uninstall.bats` *"removes romp's judge scratch and its transcripts"* changes here, and it is worth saying why it is not me weakening coverage. The test built its expected path with the same shell `${p//\//-}` substitution the script under test used — so it ratified the implementation against itself and stayed green while `--purge` swept a directory that never existed.

That shell slug agreed with Claude Code's real encoding only while the path was `/tmp/romp-judge`. The dot in `.local` breaks the agreement the moment the scratch moves, silently stranding the transcripts. The `_proj_slug` helper encodes the path the way Claude Code and `kernel/judge.py`'s `_proj_dir` do.

### Tests

13 cases pin this, **all 13 fail on `main` today**:

- `tests/test_judge_scratch_private.py` (new, 9 cases): under the state root not `/tmp`; rebind moves it; `0700` whatever the umask; a loose existing directory is tightened; a directory owned by someone else is refused; a symlink in its place is refused; an unsafe scratch skips the call and logs it *and* sets `paused`; prune sweeps the new project dir.
- `tests/romp-summarize-gate.bats`: `--safe-mode` is passed; it runs from a private dir it owns.
- `tests/romp-uninstall.bats`: the new default follows the state root, plus the slug correction above.

Full suite: **4560 passed vs 4551 on unmodified `main` — exactly +9, my new Python cases, zero of your tests flipped.** Full bats: 313 ok, 0 not ok.

### Adjacent, out of scope

`bin/romp-uninstall`'s scratch guards compare raw strings, so several spellings of the same directory (a trailing slash or newline, `//`, `/./`) walk straight past them and reach `rm -rf`. That is pre-existing and independent of this change, so I have deliberately left it alone rather than widen the diff — happy to send it as a follow-up if you want it.

This has run in production on our fork since early August, though that is evidence you cannot re-run yourself.

### DIFF
--- a/bin/romp-uninstall
+++ b/bin/romp-uninstall
@@ -154,19 +154,26 @@ rm -f  "$ROMP_DIR/vscode-extension"/*.vsix "$ROMP_DIR/vscode-extension/package.j
 echo "==> SDK venv"
 rm -rf "$STATE/sdkvenv"
 
-# ── 4b. the judge scratch, which lives OUTSIDE the state dir ───────────
-# romp runs its judges as `claude` processes rooted at /tmp/romp-judge (kernel/judge.py's
-# JUDGE_SCRATCH), so Claude Code writes their transcripts into a project dir keyed on that
-# path — outside $STATE, and therefore untouched by --purge. A purged, reinstalled romp kept
-# showing judge records from the PREVIOUS install because of it (the user 2026-07-27, who
-# rightly read that as an incomplete teardown).
+# ── 4b. the judge scratch's Claude Code project dir, which lives OUTSIDE the state dir ───────────
+# romp runs its judges as `claude` processes rooted at JUDGE_SCRATCH (kernel/judge.py), so Claude
+# Code writes their transcripts into a project dir keyed on that path — outside $STATE, and
+# therefore untouched by --purge. A purged, reinstalled romp kept showing judge records from the
+# PREVIOUS install because of it (the user 2026-07-27, who rightly read that as an incomplete
+# teardown).
+#
+# The scratch moved from /tmp/romp-judge to "$STATE/judge-scratch": a world-writable /tmp with a
+# guessable name let another local account create the path first and hand romp's judges — and this
+# teardown — a directory they control. The scratch itself is inside $STATE now, so --purge takes it
+# either way; but the DERIVED project dir is still keyed on the path, so this default has to move
+# with it or the 07-27 symptom comes back under a new name.
 #
 # These are romp's own droppings: every transcript here is a judge call romp made, never a
 # session anyone started, so removing them completes the teardown rather than deleting user
 # data. Scoped to exactly that one derived directory — YOUR project dirs are never touched.
-# Claude Code names a project dir after its cwd with '/' → '-', so /tmp/romp-judge maps to
-# '-tmp-romp-judge'; derived, not hardcoded, so it follows JUDGE_SCRATCH if that ever moves.
-JUDGE_SCRATCH="${ROMP_JUDGE_SCRATCH:-/tmp/romp-judge}"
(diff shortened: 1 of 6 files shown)


ksarma:
One ordering note, since these two touch the same block: I have opened #387, which canonicalises `ROMP_JUDGE_SCRATCH` before the guards read it.

Worth merging #387 first. This PR moves the project-dir slug to `os.path.realpath` but adds no symlink guard, so after it lands the validated string and the deleted target are no longer the same path — a symlinked scratch would resolve the slug through the link and `rm -rf` the user's own `~/.claude/projects/<their-project>` on a plain `--yes`. #387's `islink` refusal closes that before this PR can open it.

They conflict in two places in `bin/romp-uninstall` (the guard block and one comment); `tests/romp-uninstall.bats` auto-merges. #387's description has the resolution, including one line in the merged test file that the auto-merge will not flag.

PR #380, opened 2026-08-14
Keep the serve token out of curl's argv

ksarma:
The serve token is passed to `curl` as `-H "X-Romp-Token: ..."`, which puts a full-control credential into process argv. On Linux `/proc/<pid>/cmdline` is world-readable, so any other local account can read it out of `ps` while the request is in flight.

### Scope, stated honestly

**The shell clients stop publishing it.** This does *not* mean the token never appears in any argv — `bin/romp` still hands it to the browser when it opens the dashboard, which is a separate problem I am flagging at the bottom rather than fixing here.

Severity is Medium **on Linux only**: on macOS another user cannot read a process's argv (`KERN_PROCARGS2` is same-uid-or-root). And these curls are short-lived, so this is a race, not a durable exposure — but a race is not a boundary, and `romp-wake` fires at every turn boundary, re-offering the attempt.

### The change

A `_romp_token_cfg()` helper emits a curl config line (`header = "X-Romp-Token: %s"`, with `\` and `"` escaped for curl's quoted-config syntax) and the five shell call sites pipe it into `curl --config -`. That is four sites in `bin/romp` plus the `/tick` poke in `hooks/romp-wake.sh`. `-d` keeps the request body in argv, so nothing else contends for stdin.

`romp new` keeps `_tok` — it still gates the "kernel isn't running (no serve token)" error — only the header moves.

One thing worth pre-empting: `bin/romp` runs under `set -euo pipefail`, so the new pipeline is pipefail-sensitive. `_romp_token_cfg` can only fail on a printf failure, the config line is far under the pipe buffer so there is no realistic SIGPIPE, and any non-zero status lands on the existing `|| { echo "kernel not reachable" ...}` branch, which is the message the old code already produced.

### The token file's mode — consistency, not an exploit

`kernel/kernel.py` wrote the token with `write_text` and chmod'd it afterwards, leaving a brief window at the umask's mercy. It now opens `0600` from the start, keeping the trailing `chmod` for a pre-existing file.

I want to be precise that **this is not an exploitable window today**: `kernel/judge.py` chmods the state root to `0700` at import, well before `_load_token` runs, so the gap sat behind a closed door. The honest argument is narrower — this file's own mode is what `_load_token`'s docstring names as the same-user gate, and the function should not have to assume the directory did its job. (I used `os.open` rather than `_atomic_write`, which is defined after `_load_token` and whose temp file has the same window on a predictably-named path.)

### Tests this flips — three of yours, and why

Three existing assertions currently pin the leak by grepping the token out of the mock's **argv** record:

- `tests/romp-wake-hook.bats:61` and `:71`
- `tests/romp-sessions.bats:75`

All three now assert against the mock's **stdin** record instead. `tests/romp-wake-hook.bats` also gains a case asserting the token is absent from argv *while the header really is sent*, so this cannot pass by quietly not authenticating.

New `tests/test_kernel_serve_token_mode.py` (3 cases). The pinning one monkeypatches `os.chmod` to a no-op and asserts the file is already `0600` — only that interposed form actually fails on unfixed code. The other two are guards (the retained chmod still tightens a pre-existing loose file; the env-override path writes nothing) and pass either way; I kept them because they are what a future simplifier would delete, but say the word and I will drop them.

Your existing end-to-end coverage against a **real** curl is unchanged and still passes: `tests/romp.bats` and all seven `tests/romp-headless.bats` cases stand up a real HTTP server and assert the header arrives.

Full suite: **4554 passed vs 4551 on unmodified `main` — exactly +3, the new file, zero regressions.** 250 bats assertions across 20 suites, 0 failures.

### One exposure I am flagging rather than fixing

The fix here is a design call that belongs to you, so I have not made it.

The URL romp opens carries the serve token, so it becomes the argv of `open`/`xdg-open` and then of the browser itself (`bin/romp:795` + `:828`, the path a normal install takes since `bin/romp-serve:107` sets `ROMP_KERNEL_NO_OPEN=1`; `kernel/kernel.py:24760` for a hand-run kernel). On Linux that argv is world-readable, so another local account can lift a full-control credential out of `ps` for as long as that browser process lives. The token is read back from the state file rather than re-minted per boot, so one read is good until the file is deleted. A *cold* browser start leaves the tokened URL in the long-lived main-process argv — a certainty, not a race; a warm start is the short race.

Shapes that would close it: a single-use short-TTL code spent on arrival, a POST-once bootstrap, or simply opening the bare URL and letting the login page do the work. Each has real costs (a new credential type needs its own mint cap and a no-kernel fallback). Happy to implement whichever you prefer.

Note the *printed* link should keep the token deliberately — that is the one you copy to a phone.

### DIFF
--- a/bin/romp
+++ b/bin/romp
@@ -713,6 +713,19 @@ _romp_token() {
     cat "${ROMP_STATE_DIR:-${XDG_STATE_HOME:-$HOME/.local/state}/romp}/serve-token" 2>/dev/null || true
 }
 
+# The auth header as a curl CONFIG on stdin, for `curl --config -`. Never `-H "X-Romp-Token: $tok"`:
+# argv is world-readable through /proc/<pid>/cmdline, so every other account on the machine can read
+# the token off a running curl — and the token is full control of every session. These curls are
+# short-lived, which makes reading one a race rather than a certainty; that is not a boundary.
+# curl's config syntax is quoted, so a token carrying " or \ (only possible via ROMP_SERVE_TOKEN —
+# the minted one is base64url) is escaped rather than silently truncating the header.
+_romp_token_cfg() {
+    local _t _e
+    _t="$(_romp_token)"
+    _e="${_t//\\/\\\\}"; _e="${_e//\"/\\\"}"
+    printf 'header = "X-Romp-Token: %s"\n' "$_e"
+}
+
 # `romp url` — print the tokened dashboard URL, nothing else (scripting/piping; bare `romp` is the
 # human entry point). The first open seeds a year-long cookie; after that the bare URL works.
 # `--url` stays as a silent alias (agent-facing text names it).
@@ -762,7 +775,7 @@ if [[ "${1:-}" == "sessions" ]]; then
         *)      echo "usage: romp sessions [--json]" >&2; exit 2 ;;
     esac
     _kport="${ROMP_KERNEL_PORT:-29855}"
-    _resp="$(curl -sf -m 10 "http://127.0.0.1:$_kport/sessions" -H "X-Romp-Token: $(_romp_token)")" \
+    _resp="$(_romp_token_cfg | curl -sf -m 10 --config - "http://127.0.0.1:$_kport/sessions")" \
         || { echo "romp sessions: kernel not reachable on :$_kport (is romp running?)" >&2; exit 1; }
     if $_json; then
         printf '%s\n' "$_resp"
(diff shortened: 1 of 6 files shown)


ksarma:
Rebased onto `c3f9ecc5`. Two notes, one of which changes what this PR covers.

**The rebase found a fifth leak site, added after this branch was cut.** `romp new -m` (the first-prompt send) posts to `/send` with `-H "X-Romp-Token: $_tok"`. This PR originally converted four call sites; a mechanical rebase would have merged it while its own claim — that the shell clients stop publishing the token — had quietly stopped being true. All five now go through `curl --config -`, and the only `-H "X-Romp-Token"` strings left anywhere in `bin/` or `hooks/` are the comments explaining why not to use it.

The `/new` conflict resolved by keeping your `--model`/`--effort` payload and moving only the auth header to stdin. `_tok` stays where it is — it still gates the "kernel isn't running (no serve token)" error.

**Heads-up on auth-merge:** you have auto-merge enabled here, and force-pushing the rebase cancels it (same thing happened on #382). It needs re-enabling or a manual merge — it looks queued from the outside and isn't.

Re-verified against `main` as it stands now, not the old base. With `bin/romp` and `hooks/romp-wake.sh` reverted and the tests kept, 5 fail:

```
not ok 4  romp sessions: reads the KERNEL, authorizing with the serve token
not ok 11 romp-wake sends the serve token header (env override form)
not ok 12 romp-wake reads the serve token from the state file
not ok 14 romp-wake never puts the serve token in curl's argv
not ok 15 romp-wake escapes a token that would otherwise break curl's config syntax
```

Full suites on the rebased branch: **4648 passed / 40 skipped** against a 4645 baseline measured on pristine `c3f9ecc5` — exactly +3, the new test file, no regressions. **324/324 bats.** The one shellcheck SC2064 in `bin/romp` is pre-existing on `main` and untouched here.

PR #387, opened 2026-08-14
Canonicalise the judge scratch path before the guards read it

ksarma:
## What

`ROMP_JUDGE_SCRATCH` is validated by string comparison, but the `rm -rf` it guards resolves paths the way the kernel does. One directory has many spellings, so several values name a directory the guards mean to refuse while matching none of them.

Reachable on a plain `--yes` with **no `--purge`** — this teardown is not gated on `--purge`:

| value | result on `main` today |
|---|---|
| `ROMP_JUDGE_SCRATCH="$STATE"` | **`$STATE` deleted.** Every recorded session, summary and message a no-`--purge` uninstall promises to keep. |
| `"$STATE/"`, `"$DIR//romp"`, `"$DIR/./romp"` | same |
| `"$HOME/"`, `"$HOME//"`, `"$DIR//home"`, `"$DIR/./home"` | **`$HOME` deleted**, when the home path contains `romp` |
| `"$HOME/."`, `"$HOME/./"` | uninstall aborts mid-run (GNU rm refuses a `.` path) |
| a symlinked scratch spelt with a trailing slash | **the link's target is emptied** — `rm -rf "$link/"` can't unlink the link, so it recurses into the directory behind it |

The state-dir case needs no mangling at all: `$STATE` contains `romp` on every default install, so it passes the `*romp*` name check, and nothing else refuses the state root. The `$HOME` cases need a home path containing `romp` to clear the same check — real ones exist (a user named after the project, a home under `/srv/romp`) but it is a precondition, not the common case. `"$HOME"` on its own is already refused; `"$HOME/"` is one character away and passes both the equality check and the ancestor pattern.

**Preconditions, stated plainly.** All of these require `ROMP_JUDGE_SCRATCH` to be set in the environment running the uninstaller. Nothing in romp sets it — `kernel/judge.py` hardcodes `JUDGE_SCRATCH = "/tmp/romp-judge"` and never reads the variable, so it is an uninstaller-only override. This is not remotely triggerable and there is no attacker in the story: it is a guard that is supposed to refuse dangerous values and doesn't. The default `/tmp/romp-judge` path is not affected.

Two things I checked and am **not** claiming:

- A trailing newline (`"$STATE/"$'\n'`) walks past every guard but is inert on `main` — the path doesn't exist, so `rm -rf` no-ops. It is in the tests because it is part of the class and goes live under any normalisation that trims suffixes in the shell (a shell trimmer never sees the newline at all: `$(...)` strips it only *after* the trimmer has run).
- A symlinked scratch spelt *without* a trailing slash only removes the link on `main` today. It matters more once the slug is derived from `realpath` — see the note on #379 below.

## Fix

Canonicalise once, and run every check against the canonical path, so the path that is validated is the path that gets deleted. Done in `python3` — already required twice in this script — rather than a shell suffix-trimmer, which only closes the spellings at the end of the string.

- `os.path.normpath` after stripping surrounding whitespace and collapsing a leading `//` (which normpath is required to preserve) handles `//`, `/./` and trailing slashes.
- The checks the shell already made move across unchanged.
- Two are new: the state root itself, which only `--purge` removes and only behind its own confirmation; and a symlinked final component, which romp never created and which a `rm -rf "$link/"` would empty. A symlinked *ancestor* — a relocated state dir — still works.
- `os.path.samefile` is a second layer, for the spellings no string comparison can see: a case variant on a case-insensitive filesystem, a symlinked `$HOME` or `$STATE`. It applies whenever the path actually exists, which is the only case where anything can be destroyed.
- `..` is still refused on the raw value rather than normalised away: `..` through a symlink is not lexical, so collapsing it would point the `rm` at a different directory than the one checked.

The project-dir slug keeps the existing `${p//\//-}` rule and only changes the input it is derived from, so it now comes from the same canonical path the `rm` deleted.

## Tests

Four cases in `tests/romp-uninstall.bats`, enumerating the spelling class as *inputs* rather than the two or three spellings any one normalisation happens to handle — trailing slash, `/.`, `//`, `/./`, trailing newline, internal `//` and `/./`, aimed at both `$HOME` and the state dir — plus the symlink case with and without the slash, and a positive case: romp's own scratch is still torn down, transcripts and all, when spelt with `//` or a trailing slash.

All four were run against the unfixed script and fail there, each on the data assertion (the protected file is gone), not on a message assertion. 17/17 pass with the fix; the Python suite is unchanged at 4551 passed / 40 skipped.

## Note for whoever merges this alongside #379

#379 also edits this block. They conflict in two places in `bin/romp-uninstall` (the guard block, and one comment); the bats file auto-merges. Either order works, but merging **this one first** is worth considering: #379 moves the slug to `os.path.realpath` without a symlink guard, which means the validated string and the deleted target stop being the same path — a symlinked scratch would resolve the slug through the link and `rm -rf` the user's own `~/.claude/projects/<their-project>` on a plain `--yes`. The `islink` refusal here closes that before #379 can open it.

Resolving the conflict: keep this PR's python block, and fold in #379's two additions — its widened name rule becomes `("romp" in cp or cp.startswith(cstate + "/"))`, and its slug (`re.sub(r"[^A-Za-z0-9]", "-", os.path.realpath(...))`) moves into the same python call, emitted from the canonical path. One line in the merged test file also needs updating: the positive case builds its expected project dir with `${scratch//\//-}`, which must become `$(_proj_slug "$scratch")` under #379's slug rule. The auto-merge will not flag it.

## Known remaining edge

Only equality with `$STATE` is refused, not ancestry. On a default install the `romp` name check covers it, since `$STATE` is `~/.local/state/romp` and no ancestor contains `romp`. With a custom `ROMP_STATE_DIR` like `/data/romp-state/x`, the ancestor `/data/romp-state` would still be accepted. Happy to widen it if you want that closed too.

macOS is unverified here — I avoided `realpath` on the path components so the slug is unchanged there, but the `samefile` layer is only exercised on Linux in this run.

### DIFF
--- a/bin/romp-uninstall
+++ b/bin/romp-uninstall
@@ -182,39 +182,82 @@ echo "==> judge scratch"
 # guard test caught it, and only GNU rm's --preserve-root stopped it.) So require a path
 # that could only be romp's own scratch: absolute, at least two components deep, not $HOME
 # or any ancestor of it, and named for romp. Anything else is skipped loudly, never guessed at.
-_judge_ok=1
-# Claude Code names a project dir after its cwd by replacing EVERY non-alphanumeric character with
-# '-', over the REALPATH — kernel/judge.py's _proj_dir implements exactly that, and this has to
-# agree with it or --purge sweeps a directory that does not exist. A shell `${p//\//-}` replaces
-# only the slashes, which agreed by luck while the scratch was /tmp/romp-judge (no other
-# non-alnum char in it) and stopped agreeing the moment it moved under ~/.local/state: the dot in
-# `.local` survives, so the two spell the same path differently and the transcripts are left
-# behind. Derived here the same way judge.py derives it, so they cannot drift apart again.
-_scratch_slug="$(python3 - "$JUDGE_SCRATCH" <<'PYEOF'
+#
+# Every one of those checks compares STRINGS, and one directory has many spellings: "$HOME/",
+# "$HOME/.", "$STATE//x", a value with a trailing newline, a case variant on a case-insensitive
+# filesystem. Each is a different string for the same directory, so the string checks miss it
+# and the `rm -rf` below does not — a trailing slash alone is enough to walk $HOME past both
+# the equality check and the ancestor pattern. A trailing slash also switches OFF a symlink
+# refusal, because `[[ -L "$p/" ]]` resolves the link instead of lstat'ing it.
+#
+# So canonicalise ONCE, and run every check against the canonical path, so the path that is
+# validated is the path that gets deleted. It happens in python3 (already required above)
+# rather than a shell suffix-trimmer: a trimmer only closes the spellings that sit at the END
+# of the string, and it never sees a trailing newline at all, because $(...) strips that after
+# the trimmer has already run.
+#
+# The same call emits the Claude Code project slug, because that has to be derived from the very
+# same path: Claude Code names a project dir after its cwd by replacing EVERY non-alphanumeric
+# character with '-', over the REALPATH — kernel/judge.py's _proj_dir implements exactly that. A
+# shell `${p//\//-}` replaces only the slashes, which agreed by luck while the scratch was
+# /tmp/romp-judge (no other non-alnum char in it) and stopped agreeing the moment it moved under
+# ~/.local/state: the dot in `.local` survives, so the two spell the same path differently and the
+# transcripts are left behind.
+_judge_lines="$(python3 - "$JUDGE_SCRATCH" "$STATE" "$HOME" <<'PYEOF'
 import os, re, sys
-print(re.sub(r"[^A-Za-z0-9]", "-", os.path.realpath(sys.argv[1])))
+
+raw, state, home = sys.argv[1], sys.argv[2], sys.argv[3]
+
+
+def canon(p):
+    """One spelling per directory. Surrounding whitespace goes first (a trailing newline is a
+    different string for the same dir), then a doubled leading slash — which normpath is
+    required to preserve — then normpath collapses '//', '/./' and any trailing slash."""
+    p = re.sub(r"^//+", "/", p.strip())
+    return os.path.normpath(p) if p else ""
+
+
+cp, chome, cstate = canon(raw), canon(home), canon(state)
+ok = (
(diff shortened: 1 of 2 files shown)


ksarma:
Rebased onto current `main`. Two things changed the picture, so re-stating the situation rather than leaving the original description to speak for it.

**The merge-order note is overtaken.** I'd suggested landing this before #379. #379 landed first, which is fine — but it means the symlink case is live on `main` right now rather than hypothetical. #379 moved the project slug to `os.path.realpath` without a symlink guard, so the validated string and the deleted target are no longer the same path: a symlinked scratch resolves the slug through the link, and `--purge` removes the `~/.claude/projects/<...>` directory belonging to whatever the link points at.

**Re-verified against `main` as it stands today, not against the old base.** All four new cases still fail on the current post-#379 script:

```
not ok 15 no spelling of $HOME reaches rm -rf (trailing slash, /., //, newline)
not ok 16 no spelling of the state dir reaches rm -rf — a plain --yes keeps every record
not ok 17 a symlinked scratch is refused, trailing slash or not
not ok 18 romp's own scratch is still cleaned when spelt with // or a trailing slash
```

So `ROMP_JUDGE_SCRATCH="$STATE"` with a plain `--yes` and no `--purge` still deletes `$STATE` outright — `$STATE` contains `romp`, so it clears the name check, and nothing refuses the state root.

**How the conflict resolved.** Both intents are kept, in one place:

- the canonicalisation runs first, and every check now reads the canonical path, so what is validated is what is deleted;
- #379's widened name rule survives as `("romp" in cp or cp.startswith(cstate + "/"))`, so a custom `ROMP_STATE_DIR` with no "romp" in it still tears down;
- #379's `realpath` non-alnum slug survives too, and is now emitted from that **same** python call off the same canonical path — it has to stay `realpath`-based to agree with `kernel/judge.py`'s `_proj_dir`, and a symlinked *ancestor* (a relocated state dir) still works because only the final component is refused.

One thing the auto-merge did not flag: the positive test built its expected project dir with the old shell `${p//\//-}`, which silently stops agreeing with #379's rule the moment a path contains any other non-alphanumeric character. It now uses the `_proj_slug` helper #379 added, which is exactly what that helper is for.

18/18 in `tests/romp-uninstall.bats`, 326/326 across the full bats suite, no Python changes.

PR #462, opened 2026-08-17
Creating, opening, forking or reviving a session moves only the asker's chat

jfonsecarivera:
## The report

Chats kept opening at random: while working in one session's chat, the dashboard would switch to some other session unbidden.

## The cause

Every chat pane switches session on a kernel `focus` message (`render.ts` → `revealSelfPane()` + `setActive`), and four kernel paths still **broadcast** that focus to every connected dashboard — every browser window, phone PWA, VS Code pane, federated viewer:

| path | trigger | effect |
|---|---|---|
| `_create_sdk_session` | any create, incl. the CLI's POST `/new` (`romp new` in a terminal, a script) | every window's chat jumped to the new session |
| `_open_or_revive` | a feed/fleet/picker tap in **any** window | every window followed the click (a dead session's tap popped confirmRevive everywhere) |
| `_fork_session` | fork | same |
| `_revive_session` | revive | same, plus a broadcast `reviveFailed` |

`openSession` is the click-op the 2026-07-29 per-viewer fix (b36aa679, "which tab you are looking at is yours") missed: its guard test counted `_reveal_chat_for(client, …)` call sites, and the branches it did cover satisfied the count.

## The fix

Every reveal now names its asker: the WS ops pass their `client` through, the revive thread carries it, and `reviveFailed` goes to the window whose revive loader is up. An op with **no** asking dashboard (POST `/new`) moves nobody — the new tab still reaches every window via the push, just unselected. The broadcast helper `_reveal_chat` is retired so the yank cannot quietly return; `_reveal_chat_for` absorbs its shell-nudge and federation notes.

## Tests

`tests/test_per_viewer_focus.py` grows behavioral pins (open moves the asking window alone; a client-less create moves nobody) plus wiring pins for all four call sites and the helper's removal. The create-ack, revive, fork, mobile, color-pick, session-auth and new-session-dir suites follow the new signatures. Full local run: 4401 passed, 32 skipped.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -4608,7 +4608,7 @@ def _apply_new_session_prefs(sid, body):
     return out
 
 
-def _create_sdk_session(nm, cwd, auth="", prefs=None):
+def _create_sdk_session(nm, cwd, auth="", prefs=None, client=None):
     """Create + open a new SDK-backed session, ACK-FAST (the user 2026-07-14, who asked why it took so long
     to open a new SDK session). spawn() is file writes and connect() is threaded (~0.4s to a booting
     CLI) — the 7-10s the user waited was the handler's inline _push_all(): a new session invalidates the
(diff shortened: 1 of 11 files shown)


henrypinkard:
Looks good, but some tests failing

PR #501, opened 2026-08-18
The judging band reads the shared judge-usage cache — bars frames ship again

henrypinkard:
Live incident, user-reported 2026-08-18 with a screenshot: every lane served by one kernel showed skeleton furniture (state chips, comment marks, awaiting stripes) but **no work bars**, while another kernel's lanes rendered fully. That kernel's pusher thread sat at ~100% CPU.

**Root cause** (py-spy on the live kernel): `_run_judging` — the judging band — still slurps and JSON-parses the **entire** append-only `judge-usage.jsonl` on every bars build, on the GIL. At 60 MB / 225k rows, the parse eats the pusher whole: the cheap lanes skeleton (`with_bars=False`, which skips the band) keeps flowing, but the heavy `{type:"bars"}` frame effectively never ships. Hence all-furniture-no-bars, and only on the kernel with the big log.

**Fix**: the 2026-08-13 change that introduced the shared incremental reader (`_judge_usage_rows`: parse once, consume appends from the saved byte offset, 31-day retention) converted the analytics reader and missed this one. `_run_judging` now consumes the cache. The timeline window (48h) sits far inside the retention, so band content is unchanged — only the per-push cost drops from O(whole file) to O(new lines).

**Tests**: exercises the band through the cache including the incremental-append path, filters (window, dead sids), and pins that no full-file reader of `judge-usage.jsonl` ever returns — the band body itself and the file-wide pattern.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -18248,16 +18248,14 @@ def _run_judging(t0, alive_sids, semantic):
     for v in by.values():
         v.sort(key=lambda m: m["t"])
     out = []
-    try:
-        lines = (jd.STATE / "judge-usage.jsonl").read_text(errors="replace").splitlines()
-    except OSError:
-        return out
     done = set()                                      # (sid, judge, sent) of completed runs — to dedup live ones
-    for ln in lines:
-        try:
-            o = json.loads(ln)
-        except Exception:
-            continue
+    # Rows come from the SHARED incremental cache (_judge_usage_rows), never a full re-read: this used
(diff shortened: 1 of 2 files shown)


henrypinkard:
Superseded: the same root cause landed on main via the barsjudge branch with a more complete treatment (shared incremental reader, a bounded wire with loud trims, and per-stage degradation), plus its own regression tests. Closing rather than rebasing a duplicate.

PR #522, opened 2026-08-19
CI: the bats install survives a hung apt mirror — 90s per call, two retries

henrypinkard:
A hung apt mirror eats the bats job's entire 35-minute time-box: the job sits on "Install bats + tmux (Linux)" until the box cancels it, and the PR needs a manual rerun. It happened to three PRs on 2026-08-19 alone.

Every apt call now runs under `timeout 90` with two retries (a fresh attempt usually lands on a healthy mirror) and a 6-minute step-level box as the backstop. A bad mirror now costs seconds and self-heals; a genuinely broken runner fails fast instead of silently burning 35 minutes.

This PR's own bats job exercises the changed step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,16 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install bats + tmux (Linux)
         if: runner.os == 'Linux'
-        run: sudo apt-get update && sudo apt-get install -y bats tmux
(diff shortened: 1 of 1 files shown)


henrypinkard:
Superseded: bats now installs from the pinned release tarball with apt off the critical path entirely (merged via the batsboot branch), which beats hardening the apt call. Closing as redundant.

PR #527, opened 2026-08-19
The shell's viewport fit is pinch-aware: --app-h holds the layout height

henrypinkard:
User report 2026-08-19: a Mac trackpad reverse-pinch (zoom in) makes the dashboard's bottom bar jump up over the content being zoomed — with or without the timeline pane open.

**Root cause**: pinch zoom shrinks `visualViewport.height` by the zoom factor, and the shell blindly re-fit `--app-h` (its own layout var) to it on every vv resize, re-laying the whole column into the zoomed-in window. The keyboard detector shared the blind spot: a deep pinch read as "keyboard open" and collapsed the mobile bar's reservation.

**Fix**: `height * scale` recovers the layout viewport height, which is constant through a pinch, so the layout holds still while you zoom. The mobile keyboard and browser toolbars keep `scale` at 1, so address-bar and keyboard tracking behave exactly as before (the original mobile behaviors this code exists for).

The mobile-shell test pins both scale-aware expressions.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -23048,7 +23048,18 @@ _LANDING_MOBILE_JS = """
 // Chat/Feed/Timeline bar on real Android Chrome — dvh didn't track the painted area (the user 2026-06-19).
 // visualViewport.height IS the live visible height (address-bar- AND keyboard-aware), so drive --app-h
 // off it and re-fit on every resize/orientation change. Runs even if #mtabs is missing.
-function fit(){try{var h=(window.visualViewport&&window.visualViewport.height)||window.innerHeight;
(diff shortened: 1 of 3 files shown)


henrypinkard:
Strengthened after a Firefox report: desktop Firefox does not reliably update visualViewport.scale during a pinch, so the scale arithmetic was a Chrome-ism. Desktop (fine pointer) now reads innerHeight outright — pinch-immune in every browser by definition — and the visual-viewport path applies only on coarse-pointer devices, where the keyboards and collapsing toolbars it exists for live. Re-verified headless: the bar holds still through a 2x pinch on the new path.

PR #717, opened 2026-08-26
T101: the card unit is the dictated ask — dispatches link into it, containers retire

henrypinkard:
## The ruling

A dictated round contains multiple asks; each ask fans to possibly-multiple workers. The round is never the relevant unit for anything, and a dispatch-card multiplies one ask into near-duplicates on fan-out. Target structure: **round → asks → dispatches**, where the CARD is the ask — the planner's decomposition of the dictated prompt, in the user's own phrasing, with a definite done — and fan-out lives INSIDE the card (per-dispatch progress one click down, never sibling cards).

## What changed (all on existing rails, in a `notes-api` manager→workers world)

**Courier — link, never mint, when the ask is carded.** A dispatch whose chain roots to an ask the courier linked (oldest-wins, unchanged) plants its tracking node UNDER that ask; the recipient files quietly through the existing machinery (fyi placement + the `handoff.quiet` reply-sweep ending — no recipient goal will ever carry the msgId). One ask to two workers = one card with two handoff children; two asks to one worker = two cards. Only a rooted dispatch with NO resolvable ask still mints the recipient top (there the recipient card IS the ask's card; the frame enrichment rides). Linking alone never moves the ask card's column — planting a tracking child writes no verdict.

**Grouper/consolidator — containers RETIRE (not demote).** A store-level container is unavoidably a tracked unit: it owns rollup, and the provenance audit measured that it swallowed the chain evidence of every stranded ask. The visual-grouping job already has a display-side owner with no store footprint (the feed's group fold). `merge`/`split`/`retitle` stay; the retired `mint`/`group` ops parse away and are ignored if hand-built; the follow-up pivot's structural umbrella tie retires with them (pivotFrom provenance stays). **Legacy containers dissolve idempotently in every writer's rollup**: children re-parent to the first non-container ancestor with their provenance intact, placements at the container retire processed, nested containers resolve transitively.

**Trace — the container-sibling rescue.** Archives keep their containers (dissolution is live-store only), so a chain dead-ending at an evidence-free archived container now looks once — bounded, one level, cap 40 — at the container's children for the round's human record: the audit's own measured predicate, now in-walk, still a confident rule (a sibling's human record IS the round's evidence).

**Rescue, measured** over the dashboard team's real 36-hour chains: **40 of 41 origin-minted tops now trace user-rooted** (was 7 of 30 at audit time). The one miss is a parentless tracking top with no evidence anywhere — the known worst shape, genuinely untraceable.

## Tests

`tests/test_ask_unit_cards.py`: the courier fan-out matrix both directions (one-ask-two-workers = one card; two-asks-one-worker = two cards), the linkless fallback minting with frame, link-never-moves-the-column, retired-op parse/apply pins, the prompt steer, dissolution idempotency on legacy stores (nested containers, placement retirement, provenance survival), and the archived-container sibling rescue both ways. The grouper/consolidator/pivot suites reworked to the ask-unit semantics with retired behaviors pinned as no-ops. Private synthetic sids, journal cleanup in tearDown throughout.

Suites: Python 5712 passed / 0 failed; UI 2432/2432.

### DIFF
--- a/docs/judges.md
+++ b/docs/judges.md
@@ -229,23 +229,30 @@ the distiller.
 
 ## Board shape: the grouper and the consolidator
 
-As cards accumulate, related tops nest under umbrellas so the board stays
-glanceable. Both judges move whole subtrees and append no diary events:
+Since 2026-08-26 (T101, the user's ruling) the board's unit is the
+INDIVIDUAL ASK: every top-level goal is its own card, tops never nest under
+other tops, and container ("umbrella") goals are retired — a store-level
+container is unavoidably a tracked unit (it owns rollup and, measured in the
+provenance audit, swallowed the chain evidence of every stranded ask), while
+the visual-grouping job belongs to the feed's display-side group fold, which
+has no store footprint. Both judges keep only the housekeeping that serves
+the ask-unit rule, move whole subtrees, and append no diary events:
 structure, never status.
 
-**grouper.** Given the open top-level cards: nest one under another, or
-mint an umbrella when several serve one outcome, and "doing nothing is a
-valid, common outcome". Called only when the open-top set actually changed.
-Hard rules in `apply_group`: never touch a view-cleared card, no cycles,
-depth clamp 4, same-session only. A to-do-mirror top (planted flat by
-plan-sync) is explicitly the grouper's to nest.
+**grouper.** Given the open top-level cards: merge true twins into one line,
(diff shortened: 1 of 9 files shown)


henrypinkard:
Retriggering CI: the original run was stranded in queue by the GitHub Actions incident (zombie state — uncancellable, unrerunnable). Reopening immediately.

PR #731, opened 2026-08-26
The queued header counts only what is actually queued — popAll resolves, an overtaken echo says it was lost

Chris-mian:
Three separate ways a delivered or lost message kept reading as "queued", all visible at once on one session's header: four items, of which one had already run and three were sends from days earlier.

## What was wrong

- **`_pending_queued` ignored `popAll`.** Its enqueues stayed pending, so every later `dequeue` resolved an entry one slot too old and the newest *delivered* message sat in the chat as a queued bubble for good — holding the queue-aware nudge gate (`_backend_queued`) shut with it. Live corpus: 11 `popAll` records, 5 sessions each stuck with one phantom; handling the op clears all 5 and hides nothing new.
- **`_undelivered_wake_tail` ignored it too**, so the idle-queue drive read a recalled backlog as signals still owed a turn (one session was holding twelve).
- **The busy-session echo fold enlisted any unlanded tmux echo, with no bound.** An echo the transcript has overtaken with a later genuine-human turn cannot still be waiting in a FIFO pane — it is a loss. It now settles as `dropped`, which draws the already-shipped "never delivered" treatment instead of an ordinary sent bubble.

## Changes

- `_pending_queued`: `popAll` clears the queue; a content-bearing `remove` discards that entry (the rule `_undelivered_wake_tail` measured live 2026-08-18) instead of shifting the FIFO. The one deliberate divergence between the two readers — an unmatched content-remove still resolves the oldest here, because for a *display* a stranded bubble is the worse error — is documented at both ends.
- `_undelivered_wake_tail`: `popAll` withdraws the tail.
- `_human_turn_floor` extracted from `_merge_live_atoms`, so the SDK floor, the tmux echo's settle and the queued fold read one rule. `_echo_overtaken` is strictly-later, never at-or-later, so a send sharing a second with a landed turn keeps the no-flicker queued treatment.
- `TmuxBackend.dismiss_echo`: the ✕ on a never-delivered bubble was posting into a `hasattr(be, "dismiss_echo")` gate only the SDK satisfied, so the click came undone on the next push. Matching mirrors `sdk_backend.dismiss_echo`.
- `/diag/sendvis` reports each tmux echo's `dropped` — the question it could not answer while this was being diagnosed.

## Tests

New coverage in `tests/test_kernel.py` (`TestPendingQueued`, `TmuxEchoSettledByALaterTurn`, one `ViewBuilder` case), `tests/test_idle_queue_drive.py` and `tests/test_kernel_sendvis_diag.py`: `popAll` clears / does not shift later resolutions / spares what follows it; content-addressed remove; overtaken echo marked but never pruned; in-flight and same-second echoes untouched; an interrupt record does not settle an echo; dismiss clears a settled echo and refuses an in-flight one. All synthetic.

Full Python suite green locally except a known ordering flake in the nudge tests that reproduces identically on the base commit (3 failures on base, 1 on this branch, same family) — it appears only under a bare `unittest` run, which skips `conftest`'s autouse placer stub. CI runs pytest and is the arbiter. The node/webview suite was not run (no TS/CSS change); the four `kernel.py` strings its tests pin are intact.


Chris-mian:
Superseded — this was branched off a fork branch, so it carried 13 unrelated commits. Reopened clean off main: #740 (queued header) and #741 (pane clear).

PR #739, opened 2026-08-27
The clear before a pane paste actually clears — Ctrl+A never selected, so every send onto a held input was concatenated

Chris-mian:
A paste APPENDS, so the CLI's input box has to be empty before romp injects a message. `_clear_pane_input` was written for that in June as `Ctrl+A` + `Backspace`, on the understanding that Ctrl+A selects the whole input and Backspace deletes the selection.

**Measured against Claude Code 2.1.223** (a scratch CLI in tmux, 2026-08-26): Ctrl+A does not select, and the Backspace deletes exactly ONE character. Seeded box `https://example.com/pull/123␤` → after the clear: `https://example.com/pull/123`. The clear has been a silent no-op on any non-empty box, and every composer send into a pane holding text went in joined onto it.

The damage is not cosmetic: the CLI receives two texts as one prompt and answers something nobody wrote, and the delivered text no longer matches the echo romp is showing — so the send becomes indistinguishable from a lost one.

## Changes

- `Ctrl+U` empties the box, one line per press (two per line: the text, then the emptied line). `_clear_pane_input` presses until the box READS empty and returns whether it got there — event-based per press, not a sleep. An empty box costs one capture and zero keystrokes, so the ordinary send is cheaper than before.
- It gives up after two presses that move nothing (the shape the Ctrl+A regression itself had), with the press cap only as an outer guard.
- A clear that cannot finish **refuses the send**, loudly, rather than pasting anyway. Keeping the leftover and the message apart is the only safe outcome, and a refused send surfaces as never-delivered instead of as a message that vanished. `_interrupt` reports the same failure without pasting anything.

## Tests

`tests/test_kernel_inject.py` and `test_kernel.py::TmuxInject` pinned the old keystroke pair and faked every capture as `""`, so they were asserting a sequence the CLI no longer honors; both now drive a stateful fake pane modelling real Ctrl+U semantics. New `tests/test_kernel_pane_clear.py` covers the multi-line clear, the no-keystroke empty case, the refusal on a box that will not move, and the refusal on a box that cannot be read.

Verified end-to-end against the live CLI as well as in unit tests: with a leftover URL in the box, a send now leaves only the new message. Full Python suite green (5,800 tests).


Chris-mian:
Superseded — this was branched off a fork branch, so it carried 13 unrelated commits. Reopened clean off main: #740 (queued header) and #741 (pane clear).

PR #763, opened 2026-08-28
The Yatharth theme: a Claude-desktop-warm light mode, principled color semantics, one interaction vocabulary

yatharth:
The requested personal theme plus the unconditional semantics/consistency work around it. Everything opt-in stays opt-in: classic and yatharth-dark render byte-identically (test-pinned); the light theme is a third value of one setting.

**Unconditional (both themes)**
- Model / effort / context colors leave the recency colormap for single-tone ramps — model = oranges by capability (haiku → fable), effort = violets, context = calm teal with the shared amber/red taking over past ONE unified threshold pair (warn 70 / danger 88; was 60/85 × 3 copies and 70/90 × 2, some bypassing the colormap). Kills the fable == ultracode == 100%-context purple collision. Kernel-canonical, client fallbacks + Obsidian mirror grep-pinned.
- ONE button system: the documented accent-triple hover on every action button, a padding/type token scale, one transition + :active press cue; destructive stays red, selected stays reverse-highlight.
- Permission modes wear house-style glyphs (shield / pencil / bolt / route / slashed shield) beside their text in the badge + picker; the stop button moves to the statusline's far right.
- Small sheets fully tokenized (gear/strip/sessions-pane/timeline-pane, 116→46 raw hexes, census-pinned, byte-identical renders).

**The theme**
- `theme: classic | yatharth | yatharth-light` supersedes chatTabTheme (migration + derived legacy alias); gear section renamed **Appearance** with the Theme dropdown leading it.
- One applier (theme.ts) on every pane + a no-flash inline shell reader; VS Code panes honor the light choice over the editor theme, deliberately.
- The light look: ivory #FAF9F5, cream raised surfaces, warm hairlines, CLAY #C2410C accent (light only — dark keeps romp blue), re-skinned shell/banners/loader, timeline via a PAL() indirection (Obsidian host stays dark by construction).
- Typography: Space Grotesk UI + Source Serif 4 assistant prose in light (OFL, bundled + licensed; the claude.ai reading split — one token to revert); dark keeps Inter untouched.

**Guardrails (the anti-rot design)**
- Key-parity test: a token added to :root without a light value fails in the same commit.
- WCAG contrast smoke test over designated (fg, ground) pairs in BOTH themes.
- shot.sh `--body-class`: every fixture renders in both themes; raw-hex census may only go down.

Verified per commit: typecheck + 2533 webview tests + 5871 pytest + build; fixture screenshots in both themes throughout.


yatharth:
Two more batches since the description: ONE tooltip system (shared .romp-tip dress + instant-in/160ms-grace helper; statusline badges, stop button, composer buttons and feed badges upgraded from native title=) and the px type set closed (11.5/12.5 retired; the rung set pinned like the em ladder). Final state: 2541 webview tests + 5871 pytest + build green; every fixture renders in both themes via shot.sh --body-class.

henrypinkard:
The owner has looked at this and welcomes the theme: he is happy for it to exist as an opt-in. The mechanics check out on our side as well. All three font binaries are byte-identical to the official upstream releases, and dark users fetch nothing new at runtime. So the revision ahead is one scoping decision plus the list below.

**1. The default classic view must keep all of its current colors.** This is the owner's call and it gates the merge. The new model-orange and effort-violet ramps and the teal-amber-red context thresholds are welcome inside the yatharth themes, but as written they land on everyone: the colors are computed kernel-side and shipped in the payload, so every client gets them whatever theme it is running. Scoping them is awkward, since the theme choice lives client-side in localStorage. Two shapes work: ship raw values in the payload and derive the colors client-side, or have the kernel ship both palettes and let the client pick. Either works from our side; your call.

One correction in fairness, since the PR body leans on it: the claim that the ctx gauges "said 60/85" before is not accurate for the primary path on main. That path samples the global recency colormap server-side, with no thresholds at all. The 60/85 thresholds exist only in the stale-kernel fallback and on secondary surfaces.

**Confirmed defects.** Each of these was reproduced against the branch.

2. The browser `/timeline` page never sets `body.theme-light`. There is no applier and no inline reader, so in the browser dashboard every pane goes light and the timeline stays dark, permanently. `timeline-pane.css` carries a comment claiming the opposite.

3. The VS Code outline webview never receives theme changes. `fleet.ts` reads localStorage at boot and listens for storage events, but VS Code webviews have per-origin storage, and gear saves arrive as `settingsSync` host messages. There is no `installSettingsSync` call.

4. Light gauge readability. The kernel-shipped gauge, model and effort colors are dark-tuned and land at roughly 2.2:1 on the ivory ground. The `colormap.py` comment saying light clients re-encode lightness does not hold for any client here, and the timeline's four `TODO(toneReadableLight)` markers say so outright.

5. The timeline jump-to-now chevron keeps `stroke #ffffff` over the light stub background, about 1.08:1, so it is invisible there.

6. The theme-parity/WCAG test parser is comment-blind. Comments inside `feed.css`'s `:root` swallow `--accent` and `--card-border` and corrupt `--err`, so three contrast pairs skip in silence and the suite still passes green. Strip comments before parsing, and add count assertions on the parsed tokens and pairs so a future skip fails loudly. This repo has been burned by silent no-op tests before.

7. The css-census ratchet pins `gear.css` at 19 raw hexes; the sheet holds 16. That leaves slack for three new raw hexes to arrive unnoticed. Pin the actual counts.

8. The chat statusline adopted the new `tip.ts` tooltips without the prune half, so tips orphan on screen after every kernel push. Hover the Stop button while a session is working and you will see it. `feed.ts` calls `pruneTip` after render; `render.ts` needs the same.

9. The "px type set closes" commit leaves six `11.5px` rules in kernel-served CSS, in the rnet and rerr panels, with comments still documenting 11.5px as canonical.

10. Smaller things, worth sweeping in while you are in here: the Theme picker's own dropdown and the settings text input are inline-styled dark in `gear.js`, so they render dark inside the light modal; `.fl-back` comes out clay-on-navy at about 2.3:1 in light; the styled tips are mouse-only, with no keyboard focus path; and `feed.css` declares `--font-prose` only inside the `theme-light` block, never in `:root`.

Items 4, 5 and 10 are internal to the light theme, so the bar there is yours to set. Item 1 and the shared-path items gate the merge: the tooltip prune, the census pins, the test parser and the leftover 11.5px rules all sit on code classic users run. Push the revision and CI plus a re-review will follow.

yatharth:
Revision pushed — every item addressed, plus the merge conflicts with current main resolved (main's T141/T151 button rest and the 2026-08-30 gear order win; this branch's hover/press additions ride on top).

**Item 1 (gating):** took the ship-both shape. The kernel's `modelColor`/`effortColor`/`ctxColor`/usage `seg.color` revert to the recency-colormap sample byte-identical to main — classic keeps every current color — and the single-hue tones ride beside them (`modelTone`/`effortTone`/`ctxTone`/`seg.tone`, `tone=` on `/models` choices). Clients pick via `ctx-color.ts pickTone()`, probing the one body class both yatharth themes wear; a class-less document (including the Obsidian host) is classic by construction, and an older kernel's payload degrades to classic colors. Fallbacks split the same way: classic = main's 60/85 ctx + 70/90 usage verbatim (executable-tested with a faked body), yatharth = the unified 70/88 pair. Thanks for the 60/85 correction — the commit message and tests now describe the fallback paths as what they are.

2. `/timeline` gets the theme via a shared `_THEME_READER` (the landing's inline script, extracted — one mirror, grep-pinned).
3. `fleet.ts` calls `installSettingsSync()`.
4. `readableRgb()` re-encodes kernel RGB on light (L down, S up, hue + vividness order preserved; executable-tested) at every application site; the four TODO markers are paid.
5. Chevron strokes PAL ink.
6. The parity/contrast parser strips comments first and pins parsed-token + evaluated-pair counts — the previously-skipped pairs now run (and pass). Point taken on silent no-op tests.
7. Census pins are exact equality per sheet (gear at its true 16).
8. `updateStatusline` prunes orphaned tips (the Stop-hover repro).
9. The kernel-served 11.5px rules join the closed set, comments updated.
10. Gear's inline-styled controls tokenized (light modal renders light), `.fl-back` gets a light chip, styled tips gain focus/blur, and `--font-prose` was in feed's `:root` — item 6's comment-blind parser was hiding it from both of us.

State: 2609 webview tests + 6153 pytest + build green.

henrypinkard:
Re-review of the revision: all ten items verified addressed — nice work, especially the ship-both shape and the executable two-theme fallback tests. Merging on the owner's word. One commit folded directly (you'd already fixed the two stale battery-fill pins yourself; our fix for those got dropped in favor of yours):

**Folded on the branch** — eight findings from an adversarial re-review pass, all verified against the code before acting:
- The landing shell's usage rail was the one surface still painting `seg.color` unconditionally; it now picks `seg.tone` on the yatharth themes, with a theme-aware stale-kernel fallback (classic keeps the flat green verbatim).
- `.romp-tl-tip` rules are adopted into the topmost document, which defines none of the pane sheet's tokens — the fallback-less `var(--text-faint)` lost the tip's faint ink in the classic browser dashboard. Every tip-rule `var()` now carries its literal fallback, with a theme.test guard.
- The comment-popover badge's session-status fallback path skipped `readableRgb` (a text tint; the /models branch re-encoded).
- `PAL_LIGHT.menuBg` stayed `#FFFFFF` when the final commit swept the sheets' menu surface to `#FBF6EF`.
- Two warm-resweep stragglers (`#rerr-clear:hover`, `.mtabs-div`) sat at 1.001:1 on their swept grounds.
- The kernel tests had swapped the classic `cm.ramp` assertions for tone ones, leaving item 1's classic-verbatim guarantee unenforced for ctx/usage — both are pinned side by side now, and the picker-row tint got a webview pin.
- Source Serif 4 retired fully (the mono decision left 755KB of dead binaries + @font-face pairs + a stale comment), and `--font-prose` joined feed.css's `:root` (it was light-only, so dark's `.fileview-md` rule silently no-opped).
- Stale `TODO(toneReadableLight)` markers and the pre-warm contrast figures caught up with the shipped state.

**Yours to take or leave, no gate** (the owner's word: the theme's internals are your call):
- The VS Code strip's usage bars always use the stepped fallback pair on the yatharth themes, where the timeline ramps through `seg.tone` — same data, different curve.
- The comment dialog's own model/effort dropdown rows are untinted, while the main picker rows now carry rank colors.
- The shell/gear inline theme readers treat an unrecognized stored `theme` value as classic; `settings.ts` migration normalizes it — same outcome today, but the readers would drift if a third theme ever lands.

Suites on the head merged with current main: 6172 pytest + 2627 webview/extension tests + typecheck + build, all green. Auto-merge armed.

PR #847, opened 2026-09-01
After T209: the eclipse keeps one property-selected chain, so machine-spur siblings never ride the salvage

ksarma:
Rebased onto main after T209 (1fa891a8) and reworked to build on the eclipsed verdict rather than duplicate it.

T209 now covers the healing this PR originally shipped: its fork probe (at least one `system/api_error` record strictly between the fork and the spine's next conversational record) detects the machine bypass, and the kept branch renders again. The probe is also stronger than the first-record signature this PR proposed: it reads through leading non-error system records, refuses to double a spine that re-replied, and eclipses the tail-spur and corrupt-cycle windows. This rework keeps all of that unchanged.

What this PR adds on top is the selection half: which uuids the eclipse keeps. T209 keeps the whole fork component, and three sibling shapes ride along with the reply:

- A parallel tool-stub pair inside the branch renders beside its kept originals. The same stub pair drops as a rewind at every other fork.
- Two persisted attempts at one fork both render: a doubled reply.
- A branch headed by a user record behind a completed flush is kept. A rollback typed mid-storm forks in that same place (the buffered flush lands the user's replacement prompt past the error spur), so keeping the branch can re-show a prompt its user deleted, the one direction the eclipsed verdict is designed never to take.

`em._select_eclipsed_chains` narrows the keep to one chain per fork, and all three shapes are pinned red-first: of the 14 golden scenarios, six fail on the unnarrowed T209 keep and eight already pass on it (the healing and contrast cases).

## Selection is by property, not CLI write order

Nothing contracts the byte order in which the CLI flushes its buffers, so the kept chain is chosen by what it carries. Among the eclipsed fork's component chains (one per leaf, walked leaf to fork), the winner is the max-seq chain whose head is an assistant record and which carries reply text (witnessed by `landed_text_uuids`, the one existing definition); everything else in the component demotes to `rewind`, exactly as its on-spine twins classify. A late-written stub pair therefore cannot steal the keep, and a late-written second error burst cannot veto it. The (max-seq, leaf-seq) tiebreak for the two-qualifying-chains case (unobserved in the corpus) is deliberate and pinned by a test: the latest write is the CLI's final word on the turn, and it matches the single-chain pick whenever exactly one chain qualifies.

## When no chain qualifies, the two eclipse terminals part ways

At a completed flush (the probe ended on the user's landed next prompt) the fork stands down whole: a machine-orphaned reply would qualify if one existed, and what remains is indistinguishable from a rollback the storm raced. At an exhausted spur (the transcript's tail mid-flush or mid-storm, or a corrupt cycle) everything stays eclipsed: no next prompt exists, so no user gesture can have abandoned anything, and keep-on-unprovable holds. T209's tests pass unchanged under the composition, including the cycle case's kept real ask.

## Corpus evidence (counts)

Scanning 4,678 transcripts: 49 machine-orphaned bypasses, every one an assistant-headed branch carrying reply text, against 2,809 rewind forks that are not (2,035 parallel tool stubs, 700 superseded retries, 42 genuine user-gesture rollbacks, 32 other), zero overlap. 5 of the 49 storms wrote no stop-hook record, so the stop hook is characterization, not predicate; stop-hook-only bypasses (15 observed, unstudied) deliberately keep today's behavior. Two incidents were verified at the transcript level (8 episodes); the shape reaches back to at least mid-Aug 2026.

## PLACEMENTS_V

Bumps 10 to 11 (T209 took 9 to 10). The narrowed keep shrinks the parsed atom set of transcripts whose eclipsed fork carried sibling chains: v8's shape, same seal. The canary is re-pinned in the same commit; the canary fixture carries no api_error spur, so every pinned id is unchanged.

## Companion CLI report

The buffered flush hijacking the leaf instead of chaining from the current one is a CLI bug, filed as anthropics/claude-code#91113. That report is the real fix; this parse-side selection defends regardless.

## Tests (all written red-first)

- `tests/test_event_model_golden.py`, `EclipsedChainSelection` (14 tests): single, consecutive, and mid-turn-storm episodes; user-gesture contrast with the ghost gate intact; marker stand-down; stub isolation; stop-hook-only bypass stays dropped; the completed-flush user-headed belt and its tail-spur counterpart; pending-cut precedence; and the write-order defense (late-written stub pair, sibling error burst, order-independent membership, two-chain tiebreak). T209's `EclipsedBranch` scenarios pass unchanged beside it.
- `tests/test_event_model_incremental.py`: warm/cold equivalence through the flush append, including the late-stub append order.
- `tests/test_judge_rewind_cleanup.py`: membership sets, sweep predicates spare the kept chain, kept-parity, beside T209's eclipsed classification, tail-spur, and cycle cases, all unchanged.
- `tests/test_placements_canary.py`: v11 seal.

### DIFF
--- a/kernel/event_model.py
+++ b/kernel/event_model.py
@@ -520,6 +520,13 @@ def _read_jsonl_incremental(path):
     return records
 
 
+def _is_tool_result(r):
+    """True when a user-typed record is a machine-made tool_result (its first content block),
+    not a person's prompt — the discriminator the eclipse stand-down needs."""
+    c = _content((r or {}).get("message"))
+    return bool(c) and isinstance(c[0], dict) and c[0].get("type") == "tool_result"
+
+
 def _content(message):
     """The content[] of a message, normalized: a bare string becomes one text block,
     so every atom carries a list of blocks (the 'one atom, many blocks' shape)."""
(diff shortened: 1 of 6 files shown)


ksarma:
Rebased onto main after T209 and reworked to build on the eclipsed verdict — the description above matches the current head (2d191838, re-rebased once more across the event-model append-fold work): your fork probe decides whether a fork is machine-abandoned, unchanged; this PR narrows which chain the eclipse keeps.

henrypinkard:
Reviewed the rework in depth — the pivot to a delta on the eclipsed verdict is exactly the right shape, and the terminal split (completed flush stands down, mid-storm tail keeps) closes a genuine hazard in T209 as merged: a rollback typed mid-storm rejoins at an api_error fork, where the whole-component keep could re-show a deleted prompt. Owner's call was to fold follow-ups directly and merge, so one commit rides on your head (`030953cd`); all its findings came out of adversarial construction against the selection logic, each pinned red-first:

- **Selection is now per BRANCH (the sub-tree under each fork-side head), kept whole, ranked by latest reply text.** Two per-leaf-chain designs fell to constructions: your full-chain gates let a junk tail below a text-carrying record (a grafted second burst, a stub pair one level down) inherit the prefix's qualifications and out-seq the real reply on write order — and our first repair (unique-suffix gates) fell the same way, when a branch's own twin sub-branches (storm-cut tool result + stub twin) stripped their turn's shared text record from both suffixes, disenfranchising the real turn so an older sibling stole the keep. The branch unit is immune to both: junk carries no text, so nothing textless can move the pick whatever its file position. The accepted cost is documented and deliberate: a stub twin inside the winning branch renders one output-less duplicate tool row, and a grafted burst inside it costs nothing (system records never emit atoms). Never trade a possible reply for a duplicate row.
- **The completed-flush stand-down demotes only user-PROMPT-headed branches.** A turn of pure tool activity (assistant-headed, no reply text) is provably not rollback residue and now survives; so does a stranded tool_result head (user-typed but machine-made — its output even reunites with the spine's own tool call).
- **The reply witness excludes `isApiErrorMessage` records** (error-as-text failure echoes, which `atoms()` refuses to treat as replies) — a late-written failed-attempt chain can no longer qualify or steal.
- Docstrings updated to the branch semantics, both no-qualifier terminals, and the corpus figures attributed to your scan; `kept_uuids`/`chain_membership` stop overclaiming "the one reply chain".

Five construction tests pin the holes (grafted burst below text, stub pair below text, twin-sub-branch disenfranchise, failure echo, stranded tool_result), your in-branch stub test now pins the over-keep trade, and the battery + full suites are green (6337 pytest, 2628 node, typecheck, build). The corpus work and the terminal-split design made this a pleasure to build on. Auto-merge armed.

PR #868, opened 2026-09-02
The /busy drain arm is a write: gate it on an explicit serve token, presented by the manager

ksarma:
T121's deploy-drain gave the auth-exempt `GET /busy` route a write side effect: `?drain=1` calls `be.refresh_drain_hold()` in the exempt block before `_authorize` runs, arming or extending a ~12s lease (`DRAIN_HOLD_TTL`) that holds new turn starts for every session, and the lease is refreshable forever. Any drive-by web page can reach it with a no-cors GET to loopback (no preflight, no token), and any tailnet client can when the dashboard is tailscale-served. Looping the request every ~10s freezes all new turn starts indefinitely, with only a stderr line as the tell. This is the drive-by-loopback adversary that `_authorize`'s own docstring names.

## The gate

- The `/busy` read stays auth-exempt: a bare count leaks nothing, and healthz-style probes rely on it. Only the drain arm is gated.
- The gate is an explicit token (`?token=` or `X-Romp-Token`), deliberately not the full `_authorize` check. `_authorize`'s cookie branch requires only the cookie plus `_origin_ok`, and `_origin_ok` accepts an absent Origin (non-browser clients), so a cookie-carrying subresource load (an `` or `<script>` pointed at this route, which sends no Origin and rides the host-scoped cookie: RFC 6265 §8.5, cookies are port-blind) would still arm it. An explicit token clears both vectors: a no-cors request cannot set a custom header, and a subresource load cannot guess `?token=`.
- A refused drain returns the plain read response without arming. That is expected traffic, not a fail-loud case.

## The manager keeps arming, and degrades loudly

The manager is the legitimate drain caller, and it held no token. It now reads the serve token (env first, then the 0600 state file) and sends `X-Romp-Token` on its parked poll, so the deploy-drain flow keeps arming. When it cannot, the failure is visible instead of silent:

- One line per park when the token read fails, naming the path and the error.
- One refused-drain line per refusal episode: `fetchBusy` now parses the `draining` field the kernel already returns. `null` means an older kernel without the field, which is no evidence and raises no alarm.
- An armed answer after a refusal logs the recovery and re-arms the notice, so a hold lost again later is reported again.
- The park counts refusals and arms, stamped only on the park that issued the poll, so a backstop cut's cause line reports what its own park saw: the hold never armed, or it was refused N times before arming (turns started before a late arm may still be cut).

## Tests

- `tests/test_kernel_auth_hardening.py::BusyDrainWriteGate` drives the real `do_GET` dispatcher over a fake socket. The token-less drain, the cookie-only subresource drive-by, and the cross-origin fetch each read the count but arm nothing (all three red at b878b1ad before the fix); `X-Romp-Token` and `?token=` arm it; the bare read never holds.
- `tests/manager-drain.test.js` covers the token attach, both loud-degradation lines, episode re-arming, null-field silence, and real parks (fast polls, near backstop) for the evidence counters and the issue-time stamping rule.
- `tests/test_deploy_drain.py`'s source pin follows the gated spelling.

Note for reviewers: `tests/manager-*.test.js` is not wired into CI (`npm test` bundles only the `.test.ts` suites, and the shell job runs bats). Run `node --test tests/manager-drain.test.js`, the same convention as the existing `manager-restart.test.js`.

### DIFF
--- a/bin/romp-manager
+++ b/bin/romp-manager
@@ -323,19 +323,86 @@ function quietGate(st, busy, now, opts) {
   return { action: 'wait', reason: `${busy} turn(s) in flight` };
 }
 
-let pendingQuiet = null;   // {since, count, mode:'all'|<kernel id>, timer}
+let pendingQuiet = null;   // {since, count, mode:'all'|<kernel id>, timer, drainRefusedCount, drainArmedCount}
+
+// The primary kernel's serve token. The drain arm of '/busy?drain=1' is a WRITE — it holds every
+// session's new turn starts — so the kernel now requires the token for it (a drive-by loopback page
+// or a tailnet client, holding no token, can no longer freeze turns by looping the poll). We are the
+// same-user local daemon: read the 0600 token file (the kernel writes it under its state root on
+// startup), or the env a profile baked in. Absent/unreadable → the poll still reads the count (the
+// read stays auth-exempt), the drain hold just does not arm; the quiet-window gate still works off
+// the count. Read fresh each poll so a kernel that reminted its token is picked up.
+//
+// FAIL LOUDLY, NOT DIFFERENTLY: left silent, both halves of that degrade disappear — a token-read
+// failure collapses to '' with zero log, and the kernel's `draining` answer (the refusal signal
+// it returns in the same /busy response) gets parsed out and dropped — so a parked deploy would
+// believe new turn starts were held while the kernel armed nothing; busy never quiets, and the
+// 15-minute backstop SIGTERMs the kernel MID-TURN (the exact cut the quiet window exists to
+// prevent) with no line anywhere naming the cause. The degrade BEHAVIOR is deliberate
+// (count-only polling keeps working; the /busy read arm needs no token) — these notices
+// make it visible, once per refusal EPISODE, not per 3s poll: queueQuietRestart re-arms them when
+// a fresh park begins, and an ARMED drain answer re-arms the refusal notice mid-park
+// (kernelToken() reads fresh per poll, so a hold can recover when the token file
+// comes back and be refused AGAIN when it goes; each transition is new information, said once).
+let tokenNoticed = false;   // the serve-token read failed → said once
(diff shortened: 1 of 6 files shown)


henrypinkard:
Reviewed this with an adversarial pass aimed at the gate, and it holds: header case and duplicate headers fail closed, the query edges fail closed, HEAD, OPTIONS and POST all sit behind auth, and no exempt route carries another pre-authorization side effect. Kernel and manager token parity checks out, both reading the same environment variable and the same 0600 file written before the server binds. The fake-socket `do_GET` tests are the right way to pin a route's side of the gate. Merging on the owner's word.

Two follow-ups worth doing, both slated in-house unless you want them:

1. The kernel stays silent on a refused drain, so the refusal, the one event the gate exists for, is only inferred by the client, and an older manager against a new kernel degrades with no line anywhere. One latched kernel-side log per refusal episode, naming the peer and whether a token was presented, would close it.
2. `tests/manager-*.test.js` is not wired into CI, so the handshake proof does not gate merges. That is repo infrastructure.

Smaller notes, take or leave:

- The manager's "but it rejected the token" wording infers a cause the kernel never stated; the kernel also returns `draining:false` when no SDK backend exists.
- The notice bits are module-level, so a stale probe from a cleared park can log a refusal during its replacement. The counters are correctly issue-stamped; the notices are not.
- The backstop cause string cannot distinguish refused-then-armed from refused-armed-refused-again, where the hold is off at the cut.
- `tokenNoticed` never re-arms within a park, while `drainNoticed` does.
- Test copy says "once per park" where the code is once per episode.
- The fake headers dict leaves the real header object's case-insensitivity untested.

One pre-existing item outside your diff, flagged for in-house follow-up: `quietTick` still reads `pendingQuiet` at response time, so a cleared park's stale busy count can drive the replacement park's apply decision.

Suites are green here as well: the full python and webview suites, plus your manager node tests. Thanks for the careful write-up and the drive-by analysis.

PR #875, opened 2026-09-02
Arm the ask future before presenting the ask, so a synchronous answer is not dropped

ksarma:
## What breaks

An answer that arrives between an ask's presentation and the arming of its future is dropped, and the ask never completes. Since 8ab85123 (T214), `SdkSession.resolve_ask` reads `_cur_ask_fut` synchronously on the caller's thread and returns `False` when it is `None` or done, so the caller can report the answer lost. But every ask coroutine presents first and arms second: the permission branch of `_can_use_tool`, `_approve_plan`, and `_ask_one`'s `emit()` each call `self.backend._emit_ask(...)` and only then `await self._next_ask_action()`, which is where the future was created. An answer landing in that gap gets `False` back, no `_set` is scheduled, and the coroutine awaits a future nothing will resolve. Before 8ab85123 the read was deferred onto the loop, which hid the ordering.

Live, the gap is microseconds, because only the kernel's websocket click handlers answer asks, from another thread. When it is hit, the user sees the `askLost` toast (the question was lost with a restart) while the question is still on screen, and that question stays open.

In tests the gap is hit every time. The repo's own SDK-gated round-trip classes answer synchronously inside the `askLive` notify callback, on `_emit_ask`'s call stack: `AskRoundTrip` (`notify` calls `on_ask(msg["id"], "answer", 1)`), `CustomAnswerRoundTrip`, and `PermissionAndPlanRoundTrip`. With `claude_agent_sdk` importable at 70a36077, the three `AskRoundTrip` tests that send a prompt and wait for the answer fail at their 6 s bound ("can_use_tool never returned an answer"), and `CustomAnswerRoundTrip` hangs the process (`asyncio.run` with no timeout). CI installs only `pytest` and `cryptography`, so all three classes skip there and none of this was visible.

## Reproduction

At 70a36077 with the SDK importable:

```
timeout 120 python -m pytest tests/test_sdk_backend.py -q \
  -k "AskRoundTrip or CustomAnswerRoundTrip or PermissionAndPlanRoundTrip"
```

prints `.F.FF` (the three `AskRoundTrip` failures), then hangs on `CustomAnswerRoundTrip` until `timeout` kills it (exit 124).

Without the SDK, the new test in this PR reproduces it: `AskArmedBeforePresent` drives the real `_ask_one` with a `notify` that answers through `backend.on_ask` inside the `askLive` callback. Against 70a36077 it fails after its 5 s bound: "the answer was dropped: _ask_one presented its ask before arming the future".

## The fix

`SdkSession._arm_ask()` creates the future, or keeps a live one, and every ask site calls it immediately before `_emit_ask`. `_next_ask_action` awaits the armed future and clears it in `finally` as before. The invariant, stated in the docstring: an ask is never presented before the future its answer lands on exists. `resolve_ask` is untouched, so 8ab85123's outcomes hold (`None` or done gives `False`; a live future gives `True` and delivers), and `tests/test_ask_lost.py` and `tests/test_mode_truth.py` stay green. Two files change; the kernel part is 17 net lines.

## Tests

- New SDK-free regression test `AskArmedBeforePresent::test_an_answer_delivered_inside_the_presentation_callback_is_not_lost`: red at 70a36077, green with the fix (0.2 s). It needs no SDK, so CI exercises the invariant.
- The three existing round-trip classes, SDK importable, no test edits: from `.F.FF` plus a hang to `14 passed in 0.88s`.
- The rest of `tests/test_sdk_backend.py` under the SDK is unchanged by this PR: the same five tests fail before and after (`OptionsAssembly::test_ultracode_is_a_choice_everywhere_but_never_the_seeded_default`, both `ApiRetryState` tests, `ReconnectReconcilesInflight::test_reconnect_settles_a_turn_stranded_by_the_teardown`, `PushSessionCallback::test_without_the_callback_it_falls_back_to_the_pusher_wake`), none ask-related.
- Full suite, `python -m pytest -n 8 -q`: 6494 passed, 44 skipped.

## Companion suggestion (not in this PR)

Install `claude-agent-sdk` in the CI test job, or keep an SDK-free test next to each `_HAVE_SDK`-gated class as this PR does, so those classes stop reading as green when they skip.

### DIFF
--- a/kernel/sdk_backend.py
+++ b/kernel/sdk_backend.py
@@ -1812,6 +1812,10 @@ class SdkSession:
             self._clearing = True
         self._input_wake: asyncio.Event | None = None
         self._cur_ask_fut: asyncio.Future | None = None
+        self._ask_serial = asyncio.Lock()            # ONE live ask per session: the SDK dispatches every control
+        #   request as its own detached task, so two asks CAN overlap; unserialized, the second would either
+        #   overwrite the first's future (a hang) or share it (one click answering both — a silently wrong
+        #   permission). Each ask site holds this from present to resolve (PR #875 review, 2026-09-02).
         self._lock = threading.Lock()
         self._ready = threading.Event()
         # Boot-stagger hook (see BOOT_RESUME_CONCURRENCY): fired exactly once when this session's CLI
@@ -2303,7 +2307,10 @@ class SdkSession:
         lands between _emit_ask and the coroutine's first await must already find the future armed —
         otherwise it is reported lost and the ask waits forever. Before T214 that read was deferred
         into the loop, which hid the gap; every ask site now calls this right before _emit_ask, and
-        _next_ask_action awaits the armed future rather than minting a second one."""
+        _next_ask_action awaits the armed future rather than minting a second one. The keep branch
+        exists for _next_ask_action alone: the sites hold _ask_serial from present to resolve, so a
+        site never finds another ask's live future here (two overlapping asks sharing one future
+        would let a single click answer both)."""
         fut = self._cur_ask_fut
         if fut is None or fut.done():
             self._cur_ask_fut = asyncio.get_running_loop().create_future()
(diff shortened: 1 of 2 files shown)


henrypinkard:
Merging on the owner's word, with one commit folded onto your branch (`d765be70`): the fix is right and closes a real forever-hang, but arming the future at the ask sites let two overlapping asks (the SDK dispatches each control request as its own detached task) find and share one live future, so a single click answered both — the visible hang became a silently wrong permission. Each ask site now holds a per-session `asyncio.Lock` from present to resolve: the second ask is not presented until the first is answered, and each answer lands on the ask it was given to. `_arm_ask`'s keep branch is `_next_ask_action`'s alone now (docstring updated). Pinned by a test that races two asks and checks presentation order and per-ask answers. Full suites green on the branch (6505 pytest, 2648 node). Nice catch on the original gap.

PR #879, opened 2026-09-02
Kernel settings survive socket churn: queue per host and flush on open, and the store applies them in gesture order

ksarma:
## What breaks

Two related losses, both reproduced with a synthetic `TESTHOST` remote.

1. `FederationManager.outbound()` drops any message routed to a host whose socket is not OPEN: a warn toast, no client-diag record, no retry. During a kernel restart every remote socket churns for a few seconds, so a settings gesture made in that window vanishes. The file viewer's Edit consent, OK'd mid-churn, lost its `setFileEditing` broadcast; the save that followed was refused by a kernel that never heard the yes, with refusal copy pointing back at a popup already answered. Every `KERNEL_SETTING` broadcast loses the same way and leaves the machines silently disagreeing until the next gear touch.

2. The queue's dual: a delayed replay is as dangerous as a drop. The kernel applies any arriving setting unconditionally, and the judge tiers re-propagate what they apply. A dashboard tab frozen for hours (a backgrounded phone, a throttled tab) that re-dials and flushes a superseded pick reverts every linked kernel to the hours-old value, and because every kernel then agrees, the gear's disagreement marks show nothing. The same race decides two dashboards flushing one setting on a host's recovery by socket order rather than gesture order.

## Reproduction

Sender: attach a remote, close its socket (devtools offline), toggle any gear setting, reconnect. Before this change: a warn toast, and the remote never hears the setting. `ui/webview/federation-send-queue.test.ts` drives this sequence through the real `FederationManager` over stubbed sockets.

Ordering: two dashboards on one kernel with a linked remote. Pick a judge model on dashboard A. On dashboard B, whose remote socket you closed, pick a different value, then restore the socket. Before this change: the flush lands, applies, and fans out, and every kernel converges on B's older pick. `tests/test_setting_gesture_order.py::TheIncidentReplay` runs the same sequence through the real WS handler.

## The fix (three commits)

**Sender (a0ac0dda).** All remote sends go through one path, `sendRemote`. `KERNEL_SETTING` messages queue per connection while the socket is down, latest per type, and flush synchronously on the socket's open event before any post-reconnect traffic (ahead of the `romp:hostRelayUp` dispatch, so a re-shipped upload cannot overtake a queued setting). Non-setting drops keep the toast and gain a `senddrop` client-diag breadcrumb; replaying arbitrary actions after a long disconnect stays a deliberate non-goal. The open and detach `hostconn` breadcrumbs name what was flushed or discarded.

**Ordering at the store (356c6a81).** Every `KERNEL_SETTING` emitter (all gear change handlers including the comment trio and `cmtFastGate`'s inner post, the file viewer's consent post, the feed's judge-limit switch) stamps `gt`, epoch ms at the click, inside the message literal; the queue delivers it unchanged. Each store persists the last-applied stamp beside the value: `auto-nudge.json` gains `gt` and `compactSuggestGt`, `file-editing.json` and `update-mode.json` gain `gt`, the bare judge-tier files gain a `<name>.gt` sidecar; absent reads 0. An arriving older-or-equal stamp stands down: no apply, no propagation, one stderr line. The fan-out body carries the applied `gt`, so a stale value cannot win via the second hop either. Unstamped messages (older dashboards, direct HTTP) apply as before and record arrival time.

This supersedes the per-field mtime `stamps` added on 2026-08-30 for the distill-pick stomp (`_JUDGE_SETTING_FILES`, `_setting_stamp`, the `stamps`/`utime` block in `_apply_judge_settings`). That mechanism guards the `/judge-settings` hop only: a stale WS flush is still applied locally at a fresh mtime and then fans out as the newest pick. Two recency clocks deciding one write would also fight. One clock now decides every write, and the earlier tests' semantics (older never overwrites newer, a refused value never steals recency, a stampless body keeps the legacy apply) are re-pinned in gesture-stamp terms in `tests/test_judge_settings_sync.py`. If you would rather keep the `stamps` wire vocabulary, that is a small rename on top.

Store hardening in the same commit: every gt-gated read-check-write span holds one `_SETTINGS_LOCK` (auto-nudge keeps `_NUDGE_LOCK`), so racing flushes land in gesture order; the judge value and sidecar publish via `_atomic_write`, sidecar first, so a crash between the two errs toward one extra stand-down rather than an inversion; setter OSError paths are loud on stderr and never propagate a value the local store did not take (`_set_judge_state` previously swallowed OSError silently and wrote with `write_text`).

Scope note: `setAutoNudge` keeps its current handler shape (gt-gated only, no re-tick); `setCompactSuggest` keeps its act-now tick, on a real apply only.

**The losing dashboard is told (20dc3193).** A stand-down answers the delivering socket with a `settingStale` frame (the saveFile-ack `_reply` idiom). The gear toasts it in plain words with the kept value and re-fills if open, so it never keeps displaying a refused pick as applied. The toast uses the warn-toast family's dismissal (visible ✕, Escape, fade), copied into gear.js/gear.css because the gear is its own document loaded by panes that ship only that sheet.

## Relation to #883 (relay-socket reconnect)

#883 and this PR close different halves of the same loss. #883 bounds how long a relay socket stays dead: its watchdog abandons a socket that is OPEN but silent past the keepalive bound (or stuck CONNECTING), dials a fresh one in the same tick, and every surface shows which hosts are still coming. It leaves `outbound()` unchanged, so both remote send sites still drop any message routed to a socket that is not OPEN, and the abandon-and-dial itself puts the fresh socket through a CONNECTING window in which a gear gesture is lost. This PR is the outbound half. A `KERNEL_SETTING` sent into any not-OPEN window (restart churn, the onclose retry, the poll's redial, or #883's watchdog redial) queues on the connection and flushes on the fresh socket's open event, ahead of the `romp:hostRelayUp` dispatch and before the watchdog's `lastRecv` stamp. Two tests cover the composition directly: a send after the abandon-and-dial, and a setting queued before the lost-timer redial. The branch is rebased onto #883's merge commit; the overlap was textual only (the `Conn` fields, the `Conn` literal, the `onopen` block). The `hostconn` breadcrumb family gains `flushed` and `pendingDropped` beside #883's `watchdog-close`. With only this PR's test hunks applied to the #883 merge commit, the 26 tests this PR adds or rewrites fail and the other 2668 pass, so #883 covers none of these cases.

## Compatibility

- File format additions are additive and read-as-0 compatible. A downgrade to an older kernel leaves `.gt` sidecars orphaned (harmless).
- Mixed-version meshes: a kernel on this code treats an old kernel's `stamps` body as unstamped and applies it unconditionally, as before 2026-08-30. The ordering protection needs both ends on this code, as the mtime version did.
- Clocks: cross-device wall clocks order gestures. Contention is at human timescales (the motivating flush arrived hours late; a two-dashboard race is seconds apart), so sub-second skew is immaterial; equal stamps keep the stored value.

## Tests

New: `ui/webview/federation-send-queue.test.ts` (14: queue, latest-wins, per-host isolation, flush ordering, drop diagnostics, gt survives the flush un-restamped, both send sites route through `sendRemote`, no timer in the flush, and two watchdog-composition cases: a send after the abandon-and-dial, a setting queued across the lost-timer redial), `ui/webview/setting-stale.test.ts` (5), `tests/test_setting_gesture_order.py` (42: incident replay through the real WS handler, reverse-arrival race, no-gt compat, propagation carries the stamp end to end, pre-gt file formats read as 0, update-mode and compact-suggest gating, deterministic two-flush lock race, sidecar-crash safety, OSError loudness, the `settingStale` reply), and a `gear.test.ts` completeness pin that reads federation's actual `KERNEL_SETTING` set and requires a stamped emitter per member.

Updated pins: flicker-diag, warn-toast, gear-judge-versions, feed-limit-banner, file-view, test_kernel_update, test_judge_settings_sync (the stamps tests rewritten on `gt`).

Every new test fails on the base commit without the code hunks. Full suites on the rebased branch (on the #883 merge commit): `npm run typecheck` clean, `npm test` 2694/2694, pytest 6574 passed and 40 skipped, bats 364/364.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -2859,6 +2859,85 @@ def _nudge_text(count, stalled=False):
 _autonudge_cache = {}   # str(path) -> ((mtime_ns,size), dict)
 
 
+# ── kernel settings: gesture-time ordering (2026-08-29) ──────────────────────────────────────────
+# A setting message can reach this kernel long after the click that made it: federation queues
+# KERNEL_SETTING sends per host while a socket is down and flushes them on reconnect
+# (ui/webview/federation.ts sendRemote/flushPending), and that queue is latest-per-TAB — a frozen
+# dashboard tab (a backgrounded phone, a throttled browser tab) re-dials hours later and flushes a
+# pick the user has since superseded from another device. This kernel used to apply whatever
+# arrived, and the judge tiers RE-PROPAGATE what they apply (_propagate_judge_settings), so one
+# stale flush could walk the whole mesh back to an hours-old pick — and with every kernel then
+# agreeing, the gear's mixed-marks disagreement surface showed nothing. The writer's evidence
+# predated the store's newer state, so it stands down at the write moment (the standing card rule,
+# applied to settings). Ordering is enforced HERE, at the authoritative store, so every path is
+# covered at once: live sends, queue flushes, racing dashboards, and the propagation legs.
+#   - Every dashboard gesture carries `gt`: epoch ms captured ONCE where the message is built (the
+#     gear's change handlers, the file viewer's consent post, the feed's judge-limit switch), so a
+#     queued flush delivers the ORIGINAL gesture's time, never its send time.
+#   - Each setting's store persists the last-APPLIED gt beside the value (auto-nudge.json /
+#     file-editing.json gain a "gt" field; the bare judge-tier stores gain a `<name>.gt` sidecar);
+#     an arriving stamp older-or-equal to it stands down — no apply, no propagation, one loud
+#     stderr line naming both stamps.
+#   - A message with NO stamp (an older dashboard, a direct HTTP caller) applies as it always did
+#     and records its arrival time, so stamped and unstamped writers still order against each other.
+# Clock skew: cross-device wall clocks (NTP) are the ordering source. Gestures contend at human
+# timescales — the flush that motivated this arrived hours late, and a two-dashboard race is
+# seconds apart — so sub-second NTP skew is immaterial; equal stamps keep the STORED value (stand
+# down) for determinism.
+
+def _gesture_ms(msg):
+    """The gesture stamp riding `msg` (epoch ms), or None when the sender didn't stamp one."""
+    gt = msg.get("gt") if isinstance(msg, dict) else None
+    return int(gt) if isinstance(gt, (int, float)) and gt > 0 else None
+
+
+def _gt_int(v):
+    """A stored stamp, defensively: anything unreadable orders as 0, so it never blocks an apply."""
+    try:
+        return int(v or 0)
+    except (TypeError, ValueError):
+        return 0
+
+
+def _setting_stale(name, gt, applied_gt):
+    """True when gesture `gt` must stand down against the store's last-applied stamp — loudly.
+    A stand-down is also recorded on THIS thread (_pop_stale_notice) so the WS branch that
+    delivered the gesture can answer its own socket with a settingStale frame: the stderr line
+    alone left the refusing kernel's verdict invisible to the dashboard that made the gesture
+    (the open gear kept showing the refused pick, and with the mesh AGREEING on the kept value
+    the mixed marks showed nothing). Every check clears the previous verdict first, so a popped
+    notice is always the CURRENT call's."""
+    _stale_seen.last = None
+    if gt is None or gt > applied_gt:
+        return False
+    _stale_seen.last = {"setting": name, "storedGt": applied_gt}
+    sys.stderr.write("setting %s: stale gesture stood down (gesture %d <= applied %d) — "
+                     "no apply, no propagation\n" % (name, gt, applied_gt))
+    return True
+
+
+_stale_seen = threading.local()   # per-thread: the last _setting_stale stand-down (the WS reply seam)
+
+
+def _pop_stale_notice():
+    """The current thread's last stand-down verdict, consumed. The gt-gated setters run
+    synchronously on the thread that dispatched the message, so this is exact — no id plumbing."""
+    d = getattr(_stale_seen, "last", None)
+    _stale_seen.last = None
+    return d
+
+
+# ONE lock for every gt-gated setting's read-check-write span (2026-08-29): the stores read the
+# stamp, check _setting_stale, then write, on a ThreadingHTTPServer — with no lock, two
+# near-simultaneous flushes of one setting (two dashboards flushing on a host's recovery) could
+# both pass the check against the same stored stamp and then land in SOCKET order, the exact
+# inversion the gesture-time mechanism exists to prevent. The critical sections are tiny (a stat
+# + small read, a compare, an atomic write), so one module-level lock covers them all;
+# _set_auto_nudge is the one exception — its file doubles as the nudge ledger, whose other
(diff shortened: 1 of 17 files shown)


ksarma:
Rebased onto #883's merge commit. The two fixes compose rather than overlap — #883 bounds how long a socket stays dead, this PR queues the sends that land in the CONNECTING windows every redial (including #883's watchdog redial) still opens; the body's new "Relation to #883" section has the details and two tests pin the composition. Head is now 20dc3193.

henrypinkard:
Merging on the owner's word, with two commits on your branch: a merge of current main (resolving the collision with #882 — every judge-tier setter now validates against main's `_judge_model_values()` AND carries your gesture stamp; main's new test pins take the `gt=gt` argument), and one fold (`d7729f9b`):

**The gesture's own echo is quiet.** A judge pick reaches every remote kernel twice with one `gt` — the dashboard broadcasts `KERNEL_SETTING` to every host, and the origin kernel's WS branch fans the applied pick out over `/judge-settings` — so whichever copy landed second was classed stale: a `settingStale` toast told the user their own pick had "changed somewhere else" (the kept value being exactly their pick), and a stand-down line hit stderr on every ordinary pick in any mesh with a remote. The five gt-gated setters now consult `_gesture_echo` first: an equal stamp whose value the store already holds applies nothing, says nothing. `_setting_stale` keeps its signature (your two-flush race harness wraps it) and its loud verdict for an equal stamp carrying a *different* value. Pinned on the judge and boolean setters.

**Left as a follow-up, your call:** ordering by device wall clocks. A laptop ten minutes ahead picks X; every pick from a correctly-clocked phone is refused for ten minutes with copy asserting another device changed it, and no gesture can win. The repo's doctrine asks for the event rather than the clock; a cheap tightening that keeps your design is to stamp gestures as `max(Date.now(), lastSeenGt + 1)` (the gear already fetches `/version` on open and could learn each store's current `gt`), plus an explicit "apply anyway" action on the toast — a new user gesture is legitimate new information — and toast copy that does not claim another device acted when the kernel cannot know that. Smaller notes: a setting queued for a down host is silent at queue time (a client-diag breadcrumb would make a later disagreement attributable); `flushPending` clears the map only after the whole loop, so a send that throws mid-flush replays delivered entries on the next open; N kernels refusing one stale flush yield N identical toasts naming no host; `setCompactSuggest` is not in federation's `KERNEL_SETTING` set though the comments say it rides the queued door.

Suites on the rebased head: 6684 pytest, 2714 node, typecheck, build — all green. The composition tests with #883's watchdog were a good addition.

ksarma:
Thanks for the careful pass — the composition tests and the echo-quiet fold both landed exactly where I'd have wanted them. One note on the follow-up list: `setCompactSuggest` is absent from `KERNEL_SETTING` on purpose. It (and the thinking-summaries setting a coming PR adds) is a per-install, non-propagating setting: each kernel keeps its own answer and the gear shows the local kernel's value as authoritative, so broadcasting it to every remote host would be the bug rather than the fix — and `gear.test.ts` pins the set's membership by parsing federation.ts for that reason. The comments that say it rides the queued door are the misleading part; I'll tighten that wording in the next PR touching the file. The other four (clock-ordered stamps with an explicit "apply anyway", the queue-time breadcrumb, per-entry clear in `flushPending`, host-attributed dedup of stale toasts) are all fair and I've queued them.

PR #880, opened 2026-09-02
Index judges: pick the cost lever the model honors (effort on adaptive-thinking models, thinking-off on the rest)

ksarma:
## What breaks

The index tier (captioner + archiver) sets `MAX_THINKING_TOKENS=0` for every model. That lever was tuned on Haiku, where it turns a ~385-token thinking block into a ~15-token caption (722 -> 24 output tokens, 7.1s -> 0.9s per call).

On a model carrying the CLI's `rejects_disabled_thinking` capability - Fable 5 / 5.1 and Mythos - the CLI drops the thinking parameter from the request instead of sending `thinking: disabled` (the API refuses it on those models). Verified by reading the decision procedure in the Claude Code 2.1.257 and 2.1.258 binaries. So on those models the env var was a silent no-op, and every "thinking-off" index call ran full adaptive thinking at default effort. Anyone who pinned the indexing model to Fable paid for that with nothing in the log to show it. The gear's Indexing effort sub-copy described the same wrong behavior ("indexing runs with thinking disabled as a cost lever").

## Reproduction (before this change)

1. Set the indexing model to `fable` (gear -> Indexing model, or `STATE/index-model`).
2. Let the captioner run a few segments, then read `STATE/judge-usage.jsonl`: `out` tokens per caption are in the hundreds, not ~15-25, and `_judge_run` passed no `--effort` while `_judge_env("index")` set `MAX_THINKING_TOKENS=0` - the CLI dropped it.
3. At the unit level: `jd._judge_env("index")` returned `MAX_THINKING_TOKENS == "0"` regardless of model, and `_judge_run(model, ..., tier="index")` built argv without `--effort` unless the gear pick was set. The new tests in `tests/test_judge.py` (`JudgeEnv`, `EffortCapability`, `IndexTierLever`) fail against the previous `kernel/judge.py` for exactly these reasons.

## The fix

`_judge_run` resolves the tier's model first and `_judge_env(tier, auth, model)` takes it. The lever is keyed on whether the model runs adaptive thinking, using the CLI's own rule (catalog capability, then a hardcoded denylist of `claude-3-*`, Opus 4.0/4.1/4.5, Sonnet 4.0/4.5 and Haiku 4.5, then a provider default that is yes on first-party auth):

| Model | Env var | Effort flag |
|---|---|---|
| Fable, Mythos (every version), Opus 4.6+, Sonnet 4.6+ | none | `--effort <Indexing effort setting, default low>` |
| Haiku 4.5, Sonnet 4.5, Opus 4.5 and older, claude-3 | `MAX_THINKING_TOKENS=0` (as before) | none by default; the gear pick still flows through if set |
| Bare alias (`fable`, `opus`, `sonnet`, `haiku`) | placed by the version the CLI resolves it to (5.1 / 5 / 5 / 4.5) | so `haiku` stays on the env var |
| An id the rule cannot place | none | `--effort low`, plus one stderr line naming the id |

Why the boundary is generational rather than per family: a rule of "not Haiku" sends an index tier pinned to Sonnet 4.5 (a pick the version submenu offers) an ignored `--effort low` and no env var, i.e. full extended thinking - measured on a caption workload at roughly 3.5x the cost and 2.3x the latency of the env-var path. Sonnet 4.5 and Haiku have `--effort` deleted by the CLI anyway; Opus 4.5 accepts it but effort never turns its thinking off, so the env var stays its lever.

Why a stranger gets effort rather than the env var: the CLI's first-party default treats an unlisted model as adaptive, forwards `--effort`, and drops `thinking: disabled` for it - so the env var is the one lever guaranteed not to work there.

One stderr line per model says which lever applied (`romp-judge: index tier on <model> - cost lever ...`), re-announced only when it changes. The existing Indexing effort setting is the knob; nothing new to propagate. Its gear sub-copy now describes the rule. Triage and distill tiers are untouched.

Code: `kernel/judge.py` - `INDEX_EFFORT_DEFAULT`, `_EFFORT_FLOOR` / `_ALIAS_HEAD`, `_model_family_version`, `_adaptive_thinking`, `_judge_env(tier, auth, model)`, the index branch of `_judge_run`'s effort resolution, `_LEVER_LOGGED`. `ui/webview/gear.js` - the Indexing effort row's sub-copy only.

## A design choice to flag

On Opus 4.6+ and Sonnet 4.6+ the CLI does honor `thinking: disabled`, so the old env var worked there. This change moves them to `--effort low` anyway, so every adaptive model shares one lever and the effort setting is the single knob. If you would rather keep the env var wherever the CLI honors it and use effort only where it is dropped, that is a two-line change in `_judge_env` / `_judge_run`, and the tests name each model, so it is easy to flip.

## Scope

First-party auth only (a login token or an Anthropic API key, the two `_judge_env` manages). On Bedrock/Vertex the CLI's provider default flips to no and a bare `sonnet` resolves to Sonnet 4.5; neither is modeled. The denylist and alias heads are the 2.1.257 / 2.1.258 binaries' - a future CLI that moves them needs an `_EFFORT_FLOOR` / `_ALIAS_HEAD` edit; strangers follow the CLI default and are announced once so drift shows up in the log. `_ALIAS_HEAD` duplicates alias-head knowledge `kernel/kernel.py`'s `MODEL_VERSIONS` also holds; `judge.py` cannot import `kernel.py` (the kernel imports judge), so it keeps its own small table. They agree today.

## Tests (red-first)

- `tests/test_judge.py` `JudgeEnv`: the index pin rewritten to the model-keyed contract - models without adaptive thinking get the env var, the boundary is generational (Sonnet 4.5 / Opus 4.5 / Opus 4.1 / dated ids / claude-3 vs Sonnet 4.6 / Opus 4.6 / Opus 4.8 / Fable 5.1), and a bare `_judge_env("index")` answers about the configured pick.
- `EffortCapability`: the rule in both directions over 30 ids (aliases, `[1m]` suffix, dated, provider-prefixed, Vertex `@date`, 2024-shaped, Mythos), the family/version parser, the stranger announced once per id with the lever named, and alias heads (`haiku` -> 4.5 -> False).
- `IndexTierLever`: drives `_judge_run` with `subprocess.run` stubbed and captures argv + env per model - Fable / Opus / Sonnet / Mythos get `--effort low` and no var; Haiku, Sonnet 4.5 and Opus 4.5 get the var and no flag; a stranger follows the CLI fallback and is announced once; the Indexing effort setting flows through; a caller's explicit effort wins; one log line per model, re-logged on change; triage untouched.
- `tests/test_judge_auth_billing.py`: the env-contract pin passes `model="haiku"` explicitly and asserts Fable never sees the var.
- `tests/test_kernel_judge_model.py::test_judge_run_applies_tier_effort` stays green unchanged (a gear pick on Haiku still flows; only the default is model-keyed).

Full pytest suite, bats, and the vscode-extension typecheck / tests / build pass.


henrypinkard:
Merging on the owner's word, with one commit folded onto your branch (`6499afd9`): the lever table is right about what the CLI honors, but withholding `MAX_THINKING_TOKENS=0` from adaptive-thinking models traded a measured thinking-off path for unmeasured adaptive-low on Sonnet and Opus 4.6+, which the CLI does honor it on — a cost fix that could raise cost on exactly those pins. The variable now rides the index tier unconditionally: where the CLI drops it (Fable, Mythos, strangers) it is a harmless no-op and `--effort` is the lever that lands; where it honors it, it is the lever. Both ride together; the log line still names the one the CLI honors. Pins updated to assert the variable present on every index-tier call. Full suites green on the branch (6519 pytest, 2648 node). Two small notes, no gate: the bare `haiku` alias is pinned to 4.5 in `_ALIAS_HEAD` with no drift signal if the catalog moves it, and `test_index_tier_with_no_model_resolves_the_tier_pick` passes against main too, so it does not bind what its name says.

PR #882, opened 2026-09-02
Model pickers: a family click floats on the alias, a typed /model reaches the registry, and a refused pick unwinds only its own write

ksarma:
Two defects, both reproducible on `main`, plus the supporting changes they need. Seven commits, each separable; the last is a design decision for you. If one review is too much, I can split this into two PRs: the pickers (commits 1 to 4 and 6) first, then the backend verdict path (commit 5), which depends on commit 1's `_forget_model_pick(only=)` and the `on_model_refused` wiring.

## What breaks

1. **A family click pins an exact id.** `GET /models` computes each family's `default` as the user's remembered pick, else the head of the version list. Clicking "Fable" in any picker (chat statusline, comment popover, timeline lane, gear judge tiers) therefore sends `--model claude-fable-5-1`, an exact pin, while the CLI's own `fable` alias follows the family's newest release. Before Fable 5.1 was seeded, this left every picker-set session on `claude-fable-5` after the CLI had moved on. It recurs at every release, for every family.

   To reproduce: on a fresh state, `curl /models` returns `"default": "claude-fable-5-1"` for fable. Click Fable, then read `reg.model` and `sdk-defaults.json`.

2. **A typed `/model X` bypasses the registry.** The composer's `sendMessage` arm hands `/model X` to the CLI as literal text. The CLI executes it, but the reg, `sdk-defaults.json` and the reconnect's `--model` keep the old value, so the switch silently reverts at the next reconnect or kernel restart. Only the timeline's `sendCommand` arm routed `/model` through the setter.

   To reproduce: type `/model sonnet` in the chat composer of an SDK session. `reg.model` is unchanged. Restart the kernel; the session comes back on the old model.

## What changes, per commit

1. **Alias default and one router.** `default = pick or family alias`. `sendMessage`, `sendCommand` and `POST /send` share `_route_meta_command`: `/model`, `/effort` and `/fast on|off` take the kernel's park-aware setters, but only for a value the kernel can vouch for (`_vouched_model`: an alias, `default`, a catalog id, or a well-formed first-party id of a known family; the CLI validates the exact version). A typo, a multiline message that opens with the command, or a bare `/model` (the CLI's own picker) stays the CLI's, verbatim. The pick memory's writes take a lock. `_forget_model_pick(fam, only=)` and a `floating` flag on the set ops land here for later commits. The stale "fable resolved to Opus 5" rationale in `bin/romp` and `_apply_new_session_prefs` is corrected; it is not reproducible on any installed CLI.
2. **Learned versions.** The SDK backend persists the raw id the CLI reports (`reg.liveModelId`, from `_learn_model` and `_do_refresh_context`). The kernel completes each family's list with the ids the catalog lacks, marked `learned: true`, and every picker marks them " new". A sighting is T222's staleness event (`_note_unknown_model`), so the catalog fetch runs and the mark drops. `_note_unknown_model` now marks an id asked only when its refresh actually started; a sighting during the inflight boot fetch used to spend the id's one refresh on a no-op. The judge-tier setters validate against the live set, so the kernel never refuses a value the gear offered.
3. **The models frame.** The kernel sends `{type:"models", rev}` to chat, timeline and feed (the gear lives in the feed bundle) whenever the pick memory moves or the catalog fetch adds ids. Every picker re-reads `/models` in place on it, keeps the highest `rev` it applied, and drops an older response that lands late. This also replaces `_refresh_model_catalog`'s `_push_soon()`, whose comment claimed a re-read no client performed. The gear repaints its selects from whichever list won and hands each select its held value back through `setShow`, so a re-read that overtakes the page-load fill never blanks the pickers.
4. **Latest row and create-dialog parity.** Every version submenu opens with "Latest", the one gesture back to floating once a family is pinned; it sends the alias with `floating`. The new-comment dialog sends the family's remembered default like the other pickers, names the pin on the family row, and offers "<Family> · Latest". The gear rows use the T226 tokens.
5. **`set_model` reverts on the CLI's verdict.** The mode path's T139 rule, applied to models with three refinements. Revert only on a verdict: `_cli_refusal` tells a CLI error response from a lost answer (the installed SDK, `claude_agent_sdk` 0.2.132, raises the same bare `Exception` for both), and a lost answer leaves the optimistic state, which the next connect asserts. The session's own layers go back to the last ACCEPTED model (`_model_accepted`), never to what the refused write replaced; two refusals in a row would otherwise write the first refused id back. The shared `sdk-defaults.json` `model` is ruled per write by a `modelTok` token (`_seed_write_*`), so another session's accepted pick is never rolled back and two cross-session refusals never seed a refused id. A refused version leaves the pick memory (`on_model_refused` to `_model_pick_refused`, a compare-and-swap). The ack chip fires on a dormant session too (`_ack_cmd_chip`, the one builder `set_model`, `set_effort` and `set_auth` share). Both reflect checks strip the `[1m]` tag.
6. **One-time boot migration.** `_model_alias_boot_pass`, gated by the marker `STATE/model-alias-migration.done`, turns stored pre-fix seed heads into the family alias in `sdk-defaults.json`, `model-picks.json` and every reg, mirroring the CLI's own `migration_fable5_to_fable_alias`. The marker matters because three of the four heads are current releases a user legitimately pins after this change. Permanent garbage (an unparseable or undecodable store) is named and skipped; a transient read failure withholds the marker so the next boot retries.
7. **Design decision: the T223 thread-wake remap.** After commits 1 to 6, a full id in a reg is a deliberate pin, so `thread_wake_model = _family_newest_model` would re-point only deliberate pins at a dormant thread's next wake. This commit unwires the hook (the helper and the `_ensure` consult stay) and rewrites the thread-rows test to "pins stand". Drop this commit if you would rather keep the remap. Then a version pinned on a comment thread is re-pointed at its next wake, and "pins stand" does not hold for threads.

## Notes for review

- `/models` gains a leading `rev` and version rows gain `learned`; both are additive. A client on an older kernel still works, because a payload without `rev` always applies.
- `_learned_versions` globs `STATE/sdk/*.json` on every `/models` read. That is fine at ordinary session counts; reusing `list_regs` would be an improvement if that lands.
- `_cli_refusal` depends on the behavior of `claude_agent_sdk` 0.2.132's `_internal/query.py`. A future SDK that types these exceptions makes it conservative (never a revert), which is the safe failure.
- The credential path for the Models API fetch is untouched.
- No `docs/reference.md` change: there is no `romp new --model` row to update.

## Tests

All fixtures are synthetic (placeholder uuids, `TESTHOST`, the notes-api demo names). Each commit's tests were red against the base first: the alias default and typed-`/model` routing (`test_model_versions.py` ModelsRoute, `test_sdk_kernel.py`); learned versions and the staleness event (`LearnedVersions`, `test_model_catalog.py` StalenessEvent); the models frame (kernel fan-out and `rev`, plus executed loader tests in `ui/webview/models-rev.test.ts` and `ui/webview/gear-models-frame.test.ts`); the Latest row pins; about 20 backend verdict and token tests in `test_sdk_backend.py`; `AliasMigration`; `ThreadWakePinsStand`.

Final counts on the finished stack: `pytest -n 8 tests`, 6563 passed and 40 skipped; `npm test` in vscode-extension, 2676 passed; `npx tsc --noEmit`, clean; `bats tests/romp.bats tests/romp-headless.bats`, 100 passed.

### DIFF
--- a/bin/romp
+++ b/bin/romp
@@ -633,7 +633,7 @@ EOF
     _romp_cmd "romp new -t <name>"         -            "Start it as a terminal (tmux) session and attach (--detach: don't attach)"
     _romp_cmd "romp new -d <dir> <name>"   -            "Working directory for the new session (default: the current folder)"
     _romp_cmd "romp new -m <text> <name>"  -            "Start it AND deliver <text> as its first prompt (one command, idea to working)"
-    _romp_cmd "romp new --model <id> <name>" -          "Model for the SDK session, as a FULL id (e.g. claude-fable-5); re-asserted if <name> already runs"
+    _romp_cmd "romp new --model <id> <name>" -          "Model for the SDK session: a family alias (fable — follows the newest) or a full id (claude-fable-5 — a pin); re-asserted if <name> already runs"
     _romp_cmd "romp new --effort <level> <name>" -      "Reasoning effort for the SDK session (e.g. high, ultracode); re-asserted if <name> already runs"
     _romp_cmd "romp resume [<id>]"         -            "Resume a past conversation (full-screen picker; or an exact UUID)"
     _romp_cmd "romp status"                romp-manager "Show manager + kernel status"
@@ -1685,8 +1685,10 @@ case "${1:-}" in
                 --model)   shift
                            # per-spawn model, riding POST /new (the user 2026-08-14: the nightly
                            # optimizer must always run fable, and a headless script has no WS).
-                           # FULL id as the picker sends it (claude-fable-5) — the alias table
-                           # stops at opus/sonnet/haiku and quietly resolved "fable" to Opus 5.
+                           # VERBATIM: a family alias (fable — the CLI resolves it live, so it
+                           # follows the family's newest) or a full id (claude-fable-5, a pin). A
+                           # note here once said the CLI resolved "fable" to Opus 5 and so required
+                           # full ids; that is not reproducible on any installed CLI.
                            if [[ $# -eq 0 || -z "$1" || "$1" == -* ]]; then
                                echo "usage: romp new [-t] [-d <dir>] [-m <text>] [--tag <label>] [--model <id>] [--effort <level>] <name>" >&2; exit 2
                            fi
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -838,8 +838,9 @@ MODEL_CHOICES = [{"value": "fable", "label": "Fable"}, {"value": "opus", "label"
 # picked for it (model-picks.json below), else the family ALIAS itself. The CLI resolves an alias
 # live, so a bare family click follows the family's newest release exactly as the CLI's own default
 # does. (Before this the default fell to this table's head, which pinned every picker-set session
-# to claude-fable-5 while `fable` moved on.) Full ids ride every set path verbatim already — the
-# CLI's alias table stops at the family names, so a version pick needs no new transport.
+# to claude-fable-5 while `fable` moved on; see _model_alias_boot_pass for the state that left
+# behind.) Full ids ride every set path verbatim already — the CLI's alias table stops at the
+# family names, so a version pick needs no new transport.
 MODEL_VERSIONS = {
     "fable":  [{"value": "claude-fable-5-1", "label": "Fable 5.1"},   # verified live 2026-09-01 (Models API + CLI 2.1.257)
                {"value": "claude-fable-5", "label": "Fable 5"}],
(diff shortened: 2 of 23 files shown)


henrypinkard:
Merging on the owner's word (all three of the decisions your body raised resolved as shipped: the one-time migration accepts the lost-pin day, the thread-wake remap stays unwired as in commit 7, and the alias note goes), with one commit folded onto your branch (`2d207d76`): the seed set stopped at `claude-fable-5`, but fable's head moved to `claude-fable-5-1` on 2026-09-01 and every family click since wrote that id — the exact artefact the boot pass exists for — while the migration test enshrined the miss (a 5.1 pin was asserted to survive as never-a-family-click). The seed set carries the 5.1 head now; the counts and the test follow. Full suites green on the branch (6569 pytest, 2676 node). Two notes for later, no gate: `_learned_versions` re-reads every reg on hot paths (the /models read, each pick, the SDK loop), and the composer's `/effort` now reconnects an SDK session while tmux `/model` auto-confirms — both fine, both worth a line in the body.

PR #884, opened 2026-09-02
The live subagent count retires failed agents: the CLI's stop hook only fires for clean finishes

lczh:
## What you saw

The timeline's working badge read things like `Working · 37 subagents` on sessions with nothing running, and those sessions stayed "working" forever. The number only ever went up.

## Why

The SDK backend keeps a per-session set of live subagents, filled by the CLI's `SubagentStart` hook and emptied only by `SubagentStop`. Claude CLI 2.1.257 fires `SubagentStop` **only for agents that finish cleanly**. A workflow agent that fails leaves a start and nothing else.

Two independent measurements agree:

- On a live kernel, every stale entry was a workflow agent and every one was hours old. Counting the `failed` rows in each session's workflow journals gave the same numbers exactly: three sessions carried 192, 37 and 10 stale entries against 192, 37 and 10 failed agents.
- A standalone SDK probe reproduced it on demand: a two-agent workflow with one agent on a nonexistent model produced two start hooks, one stop hook, and a `state: "error"` slot for the failed agent in the run's own progress list.

## What changes

Entries are retired on events that do arrive, never on age or inference:

- A Workflow run's `task_progress` carries `workflow_progress`: one slot per agent with its latest state, never capped (only the run's log lines are trimmed), the full list re-sent on every state change. A `done`/`error` state retires the agent, and a slot re-minted with a new id for a retry ends the attempt it displaced.
- The run's end retires everything it ever listed.
- A Task/Agent subagent's lifecycle task carries the agent id as its `task_id`, so the task ending (failed or not) is the agent's end.
- The reconnect-loop teardown drops the abandoned CLI's agents and lifecycle tasks together, queuing the existing death notice with the cause named truthfully (a settings switch, not a crash).

An agent no run has listed is deliberately left alone rather than retired because "no run is live any more": that would infer death from the absence of something else instead of an event about the agent. It ends on its own stop hook, its task's end, or the teardown.

## Tests and verification

- `LiveSubagentsRetire` in `tests/test_sdk_backend.py` (11 tests), payload shapes taken from the probe recordings, synthetic ids only.
- Full Python suite and the node suite green on the branch.
- The recorded probe streams replayed through the patched session end at zero live agents, with the failed agent retired at the moment the CLI reported it.
- An adversarial review pass (three lenses, two refuters per finding) produced eleven findings; all were either folded into this branch or shown to describe pre-existing behaviour.

## Not in this PR

- A stale `_bg_tasks` entry can still be re-created by a `task_progress` that arrives for an already-ended task (pre-existing self-heal behaviour, untouched).
- The CLI also emits a `background_tasks_changed` level signal with replace semantics; consuming it as the task-set authority would remove the need for that self-heal. Left for a separate change.
- The launch ledger is not tombstoned on a reconnect (pre-existing; the next Stop hook reconciles it as before).

### DIFF
--- a/kernel/sdk_backend.py
+++ b/kernel/sdk_backend.py
@@ -3621,7 +3621,9 @@ class SdkSession:
                                                             len(died), "" if len(died) == 1 else "s", reason))
             self.backend._poke()
 
-    _RECONNECT_CAUSE = "a settings switch restarted it"   # the death notice's cause on a reconnect
+    _RECONNECT_CAUSE = "its process was restarted by a settings switch or a rewind"   # the death notice's cause on
+    #   a reconnect: every trigger of the reconnect loop (effort/fast/auth/mode switches, edit-message rewinds and
+    #   rollbacks) is named truthfully — the loop cannot tell them apart here, so the notice names the family
 
     _WF_AGENT_ENDED = frozenset(("done", "error"))
 
--- a/kernel/sdk_backend.py
+++ b/kernel/sdk_backend.py
@@ -1054,20 +1054,21 @@ def is_launch_limit(text: str) -> bool:
     return bool(text and _LAUNCH_LIMIT_RE.search(text))
 
 
-def task_death_notice(tasks: list) -> str:
+def task_death_notice(tasks: list, cause: str = "a restart or crash") -> str:
     """The visible romp notice for BACKGROUND TASKS that died with their claude process. Bg tasks are
     the CLI's children, so a kernel restart or CLI crash silently kills a session's timers/watchers —
     and a session idle-waiting on one would wait FOREVER for a completion notification that can never
     arrive (nimbus's dead campaign watcher, the user 2026-07-11). The notice names what was lost (the
     task descriptions from the lifecycle stream) so the session can relaunch exactly what still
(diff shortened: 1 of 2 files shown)


henrypinkard:
Merging on the owner's word, with one commit folded onto your branch (`12e526e5`): the reconnect death notice hardcoded "a settings switch restarted it", but the same reconnect loop is entered by edit-message rewinds and rollbacks (and the permission-mode switch), so a rewound session holding a live background task was told a false cause — and `test_f` pinned that wording. The notice now names the trigger family ("restarted by a settings switch or a rewind"), true for every trigger since the loop top cannot tell them apart. Suites green on the branch (6514 pytest, 2648 node).

Two follow-ups worth your time, no gate: the parser keys on probe-verified field names (`workflow_progress`, `workflow_agent`, `agentId`, `index`, `state`) with no shape diagnostic, so a future CLI rename brings the growth back silently — this file already has a once-per-process shape warning for `api_retry` (`_retry_shape_warned`) worth mirroring; and the docstrings say the CLI fires SubagentStop "only for clean finishes", while the bundle does attempt the stop hook on interrupted queries (the probe showed one failure kind, a nonexistent model), so softening that to what was observed would keep them honest. The event-keyed design and the probe-backed tests were exactly right.

PR #885, opened 2026-09-02
Codex session backend: run OpenAI Codex agents beside Claude Code sessions

lczh:
## Codex session backend

Adds **OpenAI Codex** as a third session backend beside tmux and the Claude Agent SDK, so a romp install can run Codex agents alongside — or instead of — Claude Code sessions, with the same dashboard, feed, timeline, judges and postal service.

### What's in it
- **`kernel/codex_backend.py`** — a `SessionBackend` that drives Codex threads through `codex app-server`, materializes Claude-shaped transcripts under `STATE/codex/projects/` (so every existing reader — the event model, judges, feed — works unchanged), keeps a registry of stable session ids, and supports spawn / resume / rename / model & effort pickers.
- **`kernel/codex_events.py`** — translation of app-server events into the transcript event model.
- **Kernel wiring** (`kernel/kernel.py`): backend selection (`POST /new` and the WS `create` op accept `backend: "codex"`, honoring a machine default), `Sessions.backend_for` / `live()` / revive / rename aware of the third backend, and a `codex` section on `GET /models` (the models the app-server actually offers, the four efforts Codex accepts) so a Codex session's pickers never show another vendor's vocabulary.
- **Judges on Codex** (`kernel/judge.py`): `STATE/judge-engine` = `codex` runs every judge as a one-shot `codex exec` with an invocation-local read-only permission profile; Codex sessions join `discover()` through the registry.
- **CLI**: `romp new --codex <name>`, and `romp engine claude|codex|status` — one switch for both knobs (which vendor runs the judges, what new sessions default to).
- **UI**: the create picker and the gear's default-backend select offer Codex; tab tooltips and timeline lanes label it; a Codex session's model/effort menus use Codex's lists.
- **Setup**: `bin/romp-codex-setup` builds an isolated venv with the Codex CLI; `docs/codex.md` documents the flow (`codex login`, engine switch, limitations).
- **Tests**: backend unit tests with a fake app-server client, event-translation goldens, discovery, judge-engine, revive, and a bats file for `romp engine` / `romp new --codex`.

### Scope and known limitations
- Purely additive: 22 files, ~5.1k lines added, 30 removed; no behavior change for tmux/SDK sessions.
- When both a Claude login and a Codex login exist and Claude's usage window is exhausted, judge calls on the Codex engine still respect the Claude usage pause (the gate reads `usage.json` before the engine branch). A machine with no Claude login is unaffected. Follow-up candidate.
- Codex reports no token usage for `codex exec`, so judge usage rows carry duration only.

### Verification
- Python: full suite green on this branch; UI/extension: 2677 pass, 0 fail; bats: 368 ok.

### DIFF
--- a/bin/romp
+++ b/bin/romp
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# romp — direct a fleet of Claude Code sessions; the dashboard is the front door
+# romp — direct your Claude Code and Codex sessions; the dashboard is the front door
 #
 # Usage (see `romp help`):
 #   romp                     # open the dashboard in your browser
@@ -624,12 +624,13 @@ if [[ "${1:-}" == "help" || "${1:-}" == "-h" || "${1:-}" == "--help" ]]; then
         printf '  %-28s %s\n' "$inv" "$*"
     }
     cat <<'EOF'
-romp — direct a fleet of Claude Code sessions; the dashboard is the front door
+romp — direct your Claude Code and Codex sessions; the dashboard is the front door
 
 Usage:
 EOF
     _romp_cmd "romp"                       -            "Open the dashboard in your browser (prints the link too)"
-    _romp_cmd "romp new <name>"            -            "Start a session (SDK; lives in the kernel, watch it on the dashboard)"
+    _romp_cmd "romp new <name>"            -            "Start a session (SDK by default; \`romp engine\` sets the machine default)"
+    _romp_cmd "romp new --codex <name>"    -            "Start it as an OpenAI Codex session (docs/codex.md)"
     _romp_cmd "romp new -t <name>"         -            "Start it as a terminal (tmux) session and attach (--detach: don't attach)"
     _romp_cmd "romp new -d <dir> <name>"   -            "Working directory for the new session (default: the current folder)"
     _romp_cmd "romp new -m <text> <name>"  -            "Start it AND deliver <text> as its first prompt (one command, idea to working)"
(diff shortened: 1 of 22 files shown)


henrypinkard:
The owner has accepted the Codex backend as an opt-in surface, on one condition: a machine that never sets up Codex must be untouched and default to Claude Code. Merging on that word, with one commit folded onto your branch (`0e47a15c`) to make the condition hold:

- **`GET /models` consulted the Codex backend unconditionally.** `model_catalog()` builds the client, which spawns `codex app-server`, so every dashboard load spawned (or, without a venv, repeatedly failed to spawn) it on every install. Codex is now consulted only where the machine opted in: the Codex default backend, the Codex judge engine, or a live Codex session (`_judge_engine_name` reads the engine file without importing the judge). Pinned in `tests/test_codex_discovery.py`.
- **`docs/codex.md` carried a fork README's install section**, sending readers to `curl | bash` a bootstrap from your fork and to clone that fork as the distribution; it points at `install.md` now, with the Codex steps after.
- **The codex-engine judge child inherited the kernel environment including the Anthropic API key** that `_judge_env` re-injects for key-billed sessions; another vendor's process has no use for it, so it is stripped from that child. Pinned in `tests/test_judge_engine.py`.
- The engine test now isolates the machine's standing user notes (read from outside the rebound state dir), so its byte-exact stdin pin holds on a developer's box as it does on a clean runner.

Two doc notes for a later round, no gate: the AppArmor `kernel.apparmor_restrict_unprivileged_userns=0` advice re-enables unprivileged user namespaces host-wide and should name that trade-off; and `romp-codex-setup` pins the SDK by version but the native binary arrives through an unpinned transitive wheel. Suites green on the branch (6636 pytest, 2677 node, build). The security screen found no exfiltration and a careful installer — thank you for the thorough write-up and the isolated venv design.

henrypinkard:
Sequencing note: this branch now conflicts with main (the per-session env and other changes that landed in the last hour touch the same session-creation and judge regions), and GitHub creates no workflow run for a pull request without a mergeable merge commit — which is why the checks vanished on the last two heads. One more change to those regions (the session-move PR) is minutes from landing, so rather than resolve twice, the rebase happens once after it: the watcher will do the mechanical resolution and re-run the suites, or, if you get there first, a rebase onto main from your side is equally welcome. Auto-merge stays armed; nothing else is needed from you.

PR #886, opened 2026-09-02
Theme feedback round three: links, the file viewer, the bells, mobile — plus the recovered sienna round

yatharth:
Seven fixes off the live yatharth-light board, plus a recovery: the "round two" commit (your voice goes sienna, quotes read mono, risky modes red, the gear's last dark corners) was committed sixteen minutes **after** PR #763 auto-merged and died with the branch — it survived only in the reflog. It is cherry-picked here, reconciled with main's newer T226 menu tokens, and its quote-tint rule — which a specificity tie had silently disabled — actually works now.

## Light-theme fixes
- **Links**: one `--link` token in both sheets. The light stand-in was `#B24A28` — a hair off `--err #B02A1C`, so every rendered link read as an error. Now an inked cobalt `#1D63A8` (5.2:1 on the page); dark unchanged (follows the editor's link color). Bubble links stay white — `.md a` was outranking `.user-bubble a` by source order.
- **File viewer / editor**: the hljs palette is `--hl-*` tokens (dark byte-identical, light re-inked for cream — the old raw palette was pale tan on `#F1EAE2`); the CodeMirror theme builds per mount reading the live body class (`{ dark: true }` was frozen at module load — the near-black search panel), chrome on shared tokens; viewer/browser raws (modal edge, row hover, focus ring, cite highlight, `th` washes) resolve through tokens / `color-mix` of the accent.
- **The master bell**: the light rail recolor outspecified the bare `.on` rules — on/off rendered pixel-identical (the network glyph's connected cue died with it). The light block restates `.on`, and OFF now reads as a slashed bell, the app's one bell-off idiom.
- **Mobile session chip** (`#mcur`/`#madd`): joins the token migration (`--btn-bg`/`--hairline`, old literals as fallbacks, light text overrides) — it sat as a raw `#2a2a2a` slab on the light chat page.

## Cross-theme fixes
- **Transcript opens at the top** (open/fork): a provisional/revive tab holds zero events, so its first real payload routed down the APPEND path, measured "at bottom?" against a one-line placeholder, landed the history at `scrollTop 0`, and the never-yank rule held it there. A first content-bearing build now lands like a brand-new tab (bottom).
- **Context battery reads 100% after a model switch**: the CLI's percentage is documented "0–100+" — a 1M→200k switch legitimately overflows and the kernel clamp presented a silent full battery. `ctxOver` now rides snapshot → merge → chat status, and the battery says **100%+**, "over this model's window". The in-flight refresh guard also queues one rerun instead of dropping a switch-time ask.
- **Mobile keyboard scrolls the page away**: `interactive-widget=resizes-content` on the shell viewport (layout shrinks instead of panning; Android Chrome), plus `fit()` undoing the stray page offset iOS still forces. The composer stays visible above the keyboard.

## Guardrails
Key-parity + WCAG contrast pairs now cover `--link` and all eight `--hl-*` tokens in both themes; the hljs pin moved to the tokenized form (dark hexes still pinned verbatim in `:root`); css-vocab pins the doubled bubble-override selectors; new pinned tests for the first-build scroll land, the bell's light on-state + slash, the chip tokens, the viewport meta + scroll reset, `ctxOver` end-to-end, and the queued context refresh.

Verified: typecheck, 2676 webview tests, full pytest (6542), build; both-theme fixture screenshots for links/hljs/bubbles/browser/queued/quote.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -10393,6 +10393,8 @@ class Sessions:
                                 "connected": bool(st.get("connected")),   # SDK handshake up → the opening-chip override stands down (fresh sessions have no transcript yet)
                                 "spawning": bool(st.get("spawning")),   # spawn/handshake in flight NOW — the ONLY window the opening chip covers (a dormant created session must read ready, the user 2026-08-13)
                                 "context": ctx if isinstance(ctx, (int, float)) else None, "compactPct": None,
+                                "ctxOver": bool(st.get("ctxOver")),   # context% is CLAMPED at 100 — this says the CLI
+                                #   reported 100+ (tokens exceed the CURRENT model's window, e.g. after a 1M→200k switch)
                                 "ctxTokens": st.get("ctxTokens"),   # raw totalTokens (SDK only) — the
                                 #   compaction-suggestion thresholds key on true tokens (2026-08-30)
                                 "fast": st.get("fast", ""),   # fast-mode state from the CLI's init ("on"/"off"/"cooldown"; "" = unknown → no badge)
@@ -20236,6 +20238,10 @@ def build_session(sid, now, tmux=None, path_override=None, tail_cap_t=None):
                   "modelPending": _model_pending_now(sid, tm),   # switching-dots on the model badge until the pick lands, from EITHER surface (the user 2026-07-03)
                   "effortPending": bool(tm.get("effortPending")),   # switching-dots on the effort badge while the /effort reconnect applies (SDK-only; the user 2026-07-06)
                   "ctx": str(tm["context"]) if tm["context"] is not None else "",
+                  # the % above is clamped at 100 — ctxOver says the CLI reported 100+ (tokens exceed
+                  # the CURRENT model's window, e.g. right after a 1M→200k model switch), so the
+                  # battery can say "over this model's window" instead of a silent 100% (2026-09-02)
+                  "ctxOver": bool(tm.get("ctxOver")),
                   # model name + effort tinted on the GLOBAL colormap by capability/effort rank (the user
                   # 2026-07-02): the statusline meta buttons just apply these (mirrors ctxColor). None = default.
                   "modelColor": _model_color(tm["model"], cm.stops_for(_colormap())),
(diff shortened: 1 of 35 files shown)


yatharth:
Round two pushed onto this PR (same live-board session):

- **The settings "double tooltip"**: the Account row's live login status accidentally wore `rs-sub`, so the hover CSS floated a second bordered popover on top of the description's (even empty it painted a box) — it's an inline `rs-note` now. Same sweep: picker buttons speak through `aria-label`, the row description stands down while a picker dropdown is open or the tri-state `mixed` mark is hovered, the send button's held-state note rides its styled tip instead of re-adding a native title, and `wireTip` strips native titles for every styled tip by construction.
- **The black "Token usage analytics" slab** in light: raw `#2d2d2f` → `--btn-bg` (gear census 13→12).
- **Desktop sizing**: the new-session picker's directory dropdown caps at the dropdown measure `min(520px, 100% − 28px)` (it was 872px wide for a one-column folder list) and wears the shared menu vocabulary; the file browser overlay stays GIANT per the 2026-08-24 ruling, but its listing reads at a centered `min(720px, 100%)`.
- **Select → type → ⌘⏎**: a printable keystroke nobody claimed drops the cursor into the composer (native insertion, never synthesized), so select-to-quote flows straight into typing and staging.
- **The stuck « opening "…" » tab**: the first full session frames can land before the 1.4 MB bundle attaches its listener; `echat` advances on send, so the pane then gets only deltas it dropped baseless, forever. `ready` now resets the client's whole chat base (the `needFull` repair, client-wide), and every delta shape treats no-base as the desync it proves (`requestFullSession`), with parked asks cleared on reconnect.

2683 webview tests + 6544 pytest, all green; each fix pinned.

henrypinkard:
Merging on the owner's word, as a squash (the fourth-round commit body quotes the user, and a merged quote cannot be scrubbed), with one commit folded onto your branch (`484462aa`):

- **Attributions paraphrase the user.** Six comments carried verbatim quotes (three from the recovered round-two commit, three new in round four: gear.js, styles.css, gear.test.ts). The repo's privacy rule paraphrases and never quotes, so each is reworded to what the user wanted rather than what they said.
- **Type-to-focus stays out of the pane's own modals.** The new handler focused the composer on any unclaimed printable key, including keys typed inside the settings, analytics and shortcuts modals and the meta menus, and treated a `<select>` as a non-typing target — so a letter typed to jump in a settings dropdown blurred it and landed silently in the hidden draft. `SELECT` is a typing target now and those four surfaces are gated, pinned beside the existing gate pins.

Merged as-is otherwise, on the owner's call, including the classic-visible changes (bubble links and quotes, the bell's OFF slash, the gear chart's tokens, the picker-menu shadow and the analytics button) and the Android viewport mode; if any of those needs to revert, it will be a follow-up rather than a send-back. Two small things for a later round: the legend swatch in the analytics chart is still a literal `#7d8590` while its bar moved to `--text-faint` (they no longer match in classic), and the queued context refresh is dropped when the in-flight request fails (the rerun check sits after the early return). Suites green on the branch (6544 pytest, 2683 node); the four-round series landed a lot of real fixes — thank you.

PR #887, opened 2026-09-02
Three queue-delivery fixes: a recovered send reaches the live queue, the rename ping's gate is atomic, and its feed-hold counts as a hold

ksarma:
Three fixes in the SDK backend's queue delivery, one commit each, all in `kernel/sdk_backend.py`. Each commit carries its own tests, and each test failed against the tree before its change.

## 1. A recovered send enters the live session's queue; a resumable reconnect only flags

**What breaks.** `_mark_dropped_echoes`' redeliver arm (a provably lost human send, verified unlanded by `_text_landed`, goes back into the queue) always writes `reg['queue']` on disk. Two of its three callers run with a live `SdkSession` already registered: `SdkSession._run` (the fresh-spawn call; `_ensure` registers the session before `start()`) and `_reconcile_stranded`'s resumable branch. There the in-memory `_pending` is authoritative, and its next `_persist_queue` snapshot (any enqueue, unqueue, or feed) overwrites the reg. The recovered text sits on disk unfed, its echo un-flagged, until a future kernel boot happens to re-read the reg.

**Trace.** A session restart inside a live kernel: `_ensure` builds the `SdkSession` (seeding `_pending` from the reg's queue) and registers it; `_run` calls `_mark_dropped_echoes(sid, self.pending())`; the redeliver arm writes `reg['queue'] = have + add`; the CLI connects and the drain pops, or the user enqueues; `_persist_queue` writes `reg['queue'] = list(self._pending)`, which never held the recovered text.

**Fix.** When a live session is registered for the sid, deliver through `SdkSession.enqueue`: dedupe against `pending()` at the write moment and set `send()`'s `/compact` and `/clear` enqueue-time flags. The boot caller (`_reseed_echoes`, before any session exists) keeps the reg write unchanged.

The same commit fixes the resumable-reconnect caller. `_reconcile_stranded`'s resumable branch is documented flag-only (the fed atom may have landed, so re-feeding risks a real duplicate), but the redeliver arm re-fed there whenever the bounded transcript-tail scan missed. `_mark_dropped_echoes` gains `refeed: bool = True`, and that caller passes `refeed=False`: the loss surfaces as the dropped flag and is never re-fed. One behavior change to review: a human send lost across a resumable reconnect now shows as never-delivered instead of being re-queued. It is what the branch's docstring already promised; if you prefer the re-feed there, reverting is one argument.

**Tests.** `tests/test_sdk_echo_durability.py`: `RedeliveryFeedsTheAuthoritativeQueue` (the live path reaches `_pending` and survives the next mirror; a text already pending is not duplicated) and `ReconnectStrandIsFlagOnly` (the resumable reconnect flags, leaves `reg['queue']` empty, and never re-feeds; the fresh-conversation branch still re-heads). `tests/test_sdk_spawn_pin_send.py`'s fake backend accepts the new keyword (required: without it the existing resumable-reconnect test raises `TypeError`) and asserts `refeed=False` there. `tests/test_restart_redelivery.py` is untouched and stays green: its stub has no `sessions` attribute, so it exercises the reg path by construction.

Relation to #878: independent and complementary. The boot path is unchanged here, so #878's fresh re-read in `_boot_reconcile` still addresses the boot-listing staleness, and its `BootDeliversARefedSend` test builds a real backend whose `sessions` is empty at boot, which is the reg path this PR keeps. No textual overlap.

## 2. The rename ping's empty-queue check and its enqueue share one lock hold

**What breaks.** `_deliver_rename_ping` checks emptiness under `s._lock`, releases, writes `renameNote=None`, then enqueues in a second hold. A send landing between the two holds queues ahead of the ping, and the CLI batches every message that arrives before a turn starts into one user record: the user's words render inside the ping's machine bubble, the fold the 2026-08-25 settle-delivery redesign exists to prevent.

**Fix.** `SdkSession.enqueue_if_empty(text) -> bool` does the check and the append in one hold. The gate uses it and spends the note only after the ping is queued, so a kernel death between the two steps re-pings a true fact at a later settle instead of losing the note.

**Tests.** `tests/test_sdk_rename_ping.py::test_a_send_racing_the_gate_cannot_queue_ahead_of_the_ping`: a deterministic interleave on a real, unstarted session. The reg write is the window, so a hook on `_update_reg` plays the racing send there; before the change the racing text landed at the head of the queue. The file's stub session gains the same atomic gate so the two existing delivery tests keep driving the real `_deliver_rename_ping`.

## 3. `queue_recallable` treats the rename ping's feed-hold as a hold

**What breaks.** `queue_recallable` returns False while a turn is in flight unless a hold is on. The ping's feed-hold (`_ping_feeding`: set when the drain feeds a `RENAME_PING_HEAD` item, cleared by that turn's first streamed message) holds every feed but was not counted, so the queued bubble lost its cancel affordance for the window in which a recall would still win.

**Fix.** Add the flag to the hold set beside the interrupt and rewind holds.

**Tests.** `tests/test_sdk_backend.py::PendingQueue::test_queue_recallable_during_the_ping_feed_hold`: a real session with `inflight=1`, toggling `_ping_feeding`. Before the change the gate returned False under the hold.

## Tests

All fixtures are synthetic (placeholder uuids, invented texts) and SDK-free (a `/bin/true` backend; unstarted real sessions or the files' existing stubs).

- Targeted: `python -m pytest -q tests/test_sdk_backend.py tests/test_sdk_echo_durability.py tests/test_sdk_rename_ping.py tests/test_sdk_spawn_pin_send.py tests/test_restart_redelivery.py tests/test_sdk_compacting_signal.py tests/test_sdk_api_retry_detail.py tests/test_model_fallback_card.py tests/test_sdk_lifecycle_hardening.py tests/test_sdk_boot_stagger.py tests/test_injected_voice.py`: 300 passed, 30 skipped.
- Full suite (`python -m pytest -q tests/`, run with pytest-xdist `-n 8`): 6533 passed, 44 skipped, 1 failed. The failure, `tests/test_postal_relay_honesty.py::PresenceBlinkHonesty::test_disk_twin_carries_the_cache_across_a_bus_restart`, fails the same way on the base commit 31d8731b (alone and under xdist; it passes when its file runs in order), and its module loads only `bin/romp-postal-service`. Pre-existing and unrelated.
- No shell or UI surface changed, so bats and the extension's tests were not run.
- No injected copy changed; `test_injected_voice` is green.

One test-setup detail: the two new echo-durability classes create the state dir's `sdk/` listing dir before constructing the backend. `list_regs` treats a missing dir as a scan fault and serves every cached row, so a fresh state dir without that dir reseeded earlier tests' regs into the new backend's boot. That looks like a separate finding in `list_regs` (its comment says a missing dir enumerates fine, but `os.scandir` raises on one) and is not addressed here.

### DIFF
--- a/kernel/sdk_backend.py
+++ b/kernel/sdk_backend.py
@@ -2017,8 +2017,13 @@ class SdkSession:
           after our receive loop was cancelled — repeats a kickoff message, which beats silently losing
           it.)
         - A RESUMABLE conversation: the atom may genuinely have landed, so re-feeding risks a real
-          duplicate. Surface the loss instead: the drop-mark flips the echo to "never delivered" NOW,
-          rather than hours later at the next thread spawn."""
+          duplicate. Surface the loss instead, FLAG-ONLY (refeed=False, 2026-08-26): the drop-mark flips
+          the echo to "never delivered" NOW, rather than hours later at the next thread spawn.
+          _mark_dropped_echoes' redeliver arm must not run here: its _text_landed scan is a bounded
+          transcript-tail read whose miss would land the message TWICE in the resumed conversation, the
+          exact duplicate this branch is documented to refuse. The re-feed lives only where the
+          conversation is NOT resumable — the re-head above, and the boot/dead-spawn callers
+          (_reseed_echoes, _run), where no client survives to have landed it."""
         if not self.inflight:
             return
         stranded = list(self._inflight_texts)      # fed to the abandoned client, never resulted
@@ -2035,7 +2040,7 @@ class SdkSession:
                 self._pending[0:0] = stranded
             self._persist_queue()                  # the fresh inputs() drains _pending on its first pass
         elif stranded:
-            self.backend._mark_dropped_echoes(self.sid, self.pending())
+            self.backend._mark_dropped_echoes(self.sid, self.pending(), refeed=False)
         self.backend._poke()
 
     # ---- async internals (run inside the quarantined loop) ----
@@ -5323,7 +5328,7 @@ class SdkBackend:
             if self._live.get(reg["sid"]):
                 self._mark_dropped_echoes(reg["sid"], reg.get("queue") or [])
 
-    def _mark_dropped_echoes(self, sid: str, queued_texts) -> None:
+    def _mark_dropped_echoes(self, sid: str, queued_texts, refeed: bool = True) -> None:
         """A fresh CLI is spawning for this sid, or the kernel just booted: whatever process held any
         earlier send is gone. An input echo whose text is neither in the surviving queue (about to be
         delivered to the new CLI) nor landed in the transcript has no holder left — its send is provably
@@ -5333,7 +5338,14 @@ class SdkBackend:
         history). Event-based — keyed on the spawn/boot that orphaned the send, never on age — and
         self-correcting: an echo whose text actually LANDED still prunes by text on the next build, so a
         premature flag can never stick to a delivered message. The flag rides the registry mirror
-        (_persist_echoes), so it survives further restarts."""
+        (_persist_echoes), so it survives further restarts.
+
+        `refeed=False` (2026-08-26, honoring _reconcile_stranded's documented policy): the RESUMABLE-
+        reconnect caller takes the flag path ONLY — the redeliver arm's _text_landed scan is a bounded
+        tail read whose miss would land the message a second time in a conversation that genuinely
+        kept the first, exactly the duplicate that branch refuses by design. The loss still surfaces in
+        full (the dropped flag); only the queue re-add is withheld. Boot and dead-spawn callers keep the
+        default: no client survived there to have landed anything."""
         d = self._live.get(sid)
         if not d:
             return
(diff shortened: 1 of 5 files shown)


henrypinkard:
Merging on the owner's word, as shipped. One note on the policy question your body flagged: the review agreed the asymmetry is unprincipled in the abstract — the resumable-reconnect caller is now flag-only while the fresh-spawn caller still re-feeds, and both rely on the same bounded `_text_landed` tail scan — but your reading of the reconnect branch's own docstring is the documented contract, and flipping either side is one argument, so it lands as you wrote it. If the owner wants one policy everywhere later, it is a one-line change on either side. Two small notes: `test_live_session_redelivery_never_duplicates_a_pending_text` also passes on main (a fair guard, but not red-first), and the `else:` branch in `test_a_send_racing_the_gate_cannot_queue_ahead_of_the_ping` is unreachable, so a future gate that wrongly holds would pass it — the assertion could be unconditional. The three fixes themselves check out, and the live-path redelivery closes a real limbo.

PR #888, opened 2026-09-02
Move a live session's working directory: SdkBackend.move over the CLI's set_cwd control request, with a tab-menu row, POST /move and `romp move`

ksarma:
## Summary

A live session's working directory can now be changed. Right-click a tab and choose **Move to folder…**, or run `romp move <session> <dir>` (`POST /move`). The conversation, name, mailbox, goals and history stay with the session; from its next turn on, the agent works in the new folder and reads its `CLAUDE.md`. Terminal (tmux) sessions refuse with a reason, since there is no relocation primitive to drive. The case that motivated it: a subproject becomes its own repository and the session working on it should move with it.

Until now the directory was fixed at creation, and `_cwd_of`'s docstring said there was no SDK call to relocate a session. There is one: Claude Code's `set_cwd` control request, the headless twin of the interactive `/cd`. This PR wraps it and corrects the docstring.

## Two dependencies to know about

**A private SDK sender.** The Python `claude-agent-sdk` has no public wrapper for `set_cwd`: none in 0.2.132 (the version a `bin/romp-sdk-setup` venv had installed when this was built) through 0.2.151 (PyPI's newest as of 2026-09-02). The request goes through `client._query._send_control_request(request)`, whose signature `(request: dict, timeout=60.0)` is unchanged across those versions. The call is isolated in one helper, `SdkBackend._set_cwd_request`, with a note to swap in the public wrapper when one lands; a client without the sender fails with a named reason and changes nothing. For the shape to expect: the TypeScript SDK's `Query` implements `setCwd(path, {trustAccepted, trustedDirectory})` and sends exactly this request (`sdk.mjs` in `@anthropic-ai/claude-agent-sdk` 0.3.259), but `sdk.d.ts` does not declare it, so it is not a public API there yet either.

**The CLI's protocol.** Verified against Claude Code 2.1.257 in a hermetic `CLAUDE_CONFIG_DIR`; the protocol strings are also present in 2.1.258. Response arms: `ok {cwd, changed, transcript_relocated}` (a same-dir request is `ok, changed: false` and emits nothing); `needs_trust {directory, trust_root?}`; `rejected {reason, message}` with reason one of `not_found`, `not_a_directory`, `unsafe_path`, `blocked_by_rule` (a `Cd(...)` permission rule), `busy`; a relocation failure comes back as a control error with the CLI's chdir rolled back. On `ok` the CLI moves the current `<lastSid>.jsonl` and its `<lastSid>/` folder into the new cwd's project slug (a file already at the destination is set aside as `.superseded-<ts>`), appends a `{type: relocated, sessionId, relocatedCwd}` record plus an isMeta notice telling the model its environment block is stale, re-resolves project settings, hooks, skills and MCP servers, fires the `CwdChanged` hook (not `SessionStart`; Romp registers no `CwdChanged` hook, so nothing Romp-side re-runs), and emits, with no query sent, an init `SystemMessage`, a turn-less `ResultMessage(num_turns=0)` and two `commands_changed` messages. A `set_cwd` sent right after a `ResultMessage` was accepted in testing, so the bounded busy retry described below is a backstop for a case that did not reproduce.

## The backend: `SdkBackend.move`

In order:

1. Canonicalise the target: realpath plus true on-disk casing, an existing directory only, never created. The stored string is what every transcript path derives from.
2. Revive a dormant session first, in its old folder. A `--resume` from another cwd adopts the project directory it finds the transcript in and keeps writing there.
3. Answer `"busy"` before writing anything when a turn is in flight; the kernel parks the op and retries.
4. Write a two-phase `cwdPending` flag into the registry, then send the request. Answer `needs_trust` by re-sending with trust accepted: the user's folder pick is the trust decision, and the CLI records it per git root.
5. On `ok`: update the live object, the registry (`cwd` plus a durable `movedFrom`), the `names/` discovery file, and rename the session's earlier transcripts (`/clear` episodes, resume forks) into the new slug, since the CLI moves only the current file. Nothing is copied: the same fsid under two slugs makes every future `--resume` fail with "No conversation found".

A lost reply (control error, timeout) and a kernel death between the CLI's `ok` and the registry write are settled the same way: the transcript's location under exactly one slug decides. Both or neither is logged as a problem and left for a person, and a second move is refused until it is cleared.

`SessionBackend.move` is a concrete ABC default that refuses with a backend-neutral reason, so a backend without a relocation primitive never reads as movable by omission; `TmuxBackend.move` spells out the terminal case.

## The kernel around it

- A guard keeps the move's turn-less `ResultMessage` from running the turn-settle side effects (armed before the request, spent on `num_turns == 0`). Without it, a false turn-completed, a redundant `waiting` write and an early parked reconnect all fired on a turn that never was.
- Boot reconcile heals registries left with `cwdPending`.
- The drive-op FIFO gets a `("cwd", path, retries)` kind. A mid-turn move parks as a plain-words "move to …" chip and fires at turn end. `_moving` holds the queue while the blocking round trip runs on its own thread. A CLI-side `busy` retries a few passes, then waits on the backend's turn counter, because a turn the CLI started itself still ends with a `ResultMessage`.
- One move per session at a time (the flag is claimed under the registry lock). Comment threads keep their own folder in this version.
- The system card and lane read the registry's cwd first. The per-record `cwd` stamp is the CLI's tracked cwd (a shell `cd` inside a Bash call moves it too) and lags a move until the next turn.
- `POST /move` runs the move synchronously on the handler thread. A dormant session is revived first, so the handler can block up to 60 s (connect) plus 30 s (control); `romp move` uses `curl -m 90` for that reason.

Claude Code keys three things by folder, so they stay behind: the old project's auto-memory, its entry in `~/.claude.json` (allowed tools, MCP approvals, trust), and the old repository's `.claude/settings.local.json`. The reference docs list them.

## Before and after

Before: the only way to put a session in another folder is to end it and start a new one there, losing the conversation. After: `romp move web /srv/notes-api/web` on an idle SDK session prints `romp move: "web" now works in /srv/notes-api/web`; the transcript sits under `~/.claude/projects/-srv-notes-api-web/` only; the next turn's environment names the new folder. On a session mid-turn the CLI prints that the move is queued, and the chip in the chat fires when the turn ends.

## Tests

- `tests/test_session_move.py` (backend, against a scripted fake control channel): every response arm, the two-phase flag, the names rewrite, prior-transcript relocation, boot heal, lost reply, a concurrent claim, the `num_turns=0` guard, and a source pin that the private sender is spoken in one place.
- `tests/test_session_move_kernel.py`: park, fire, hold and retry, plus `POST /move` over the real handler on loopback.
- `tests/test_event_model_relocated.py`: the `relocated` and isMeta records yield no atoms, the parent chain stitches across them, and no turn is left open.
- `tests/romp.bats`: `romp move` argv plumbing, relative-dir resolution against the caller's cwd, queued and refused replies, usage and no-token exits.
- `ui/webview/tab-move.test.ts`, `fast-toggle.test.ts`, `vscode-extension/src/pipe-intent.test.ts`: source pins for the row, the dialog and the typed replies; `moveSession` is an intent op.
- `tests/test_session_move_live.py`: one live integration test, opt-in (`ROMP_MOVE_LIVE=1` plus an SDK interpreter, a `claude` binary, and `ANTHROPIC_API_KEY` or a helper command in `ROMP_MOVE_LIVE_API_KEY_HELPER`). It skips in CI. It bills two short turns and checks the response shapes, that the conversation survives the move, and that the transcript lives under the new slug only.

The full Python suite, `tests/romp.bats`, and the extension's typecheck, tests and build pass.

## Adjacent open PRs

This touches `_boot_reconcile` (where #878 edits the docstring tail) and the gap between `rename` and `set_model` plus `SdkBackend.__init__` (where #882 adds lines). Whichever lands second needs a mechanical rebase at those spots; nothing semantic overlaps.

### DIFF
--- a/bin/romp
+++ b/bin/romp
@@ -657,6 +657,7 @@ EOF
     _romp_cmd "romp sessions [--json]"     -            "List the fleet with each session's state, colours and backend"
     _romp_cmd "romp fork <parent> <name>"  -            "Branch a session into a new parallel one (parent untouched; --at <uuid> picks the cut)"
     _romp_cmd "romp rename <session> <name>" -          "Rename a session (uuid-keyed: mailboxes, goals and history all follow)"
+    _romp_cmd "romp move <session> <dir>"   -           "Move an SDK session's working directory (its conversation follows; mid-turn it waits for the turn to end)"
     _romp_cmd "romp color <session> [<#hex|slot>]" -    "Recolor a session's identity swatch (palette hex or slot number; no color arg reads bg/fg)"
     _romp_cmd "romp tag [<name>] [--add ...]" -         "Session tags: bare lists them; <name> with --add/--remove/--color/--delete merges one (a tagged session leaves the no-tags view); --host <kernel> edits an attached kernel's tag; --rename renames"
     _romp_cmd "romp watch-pr <n> [--repo o/r]" -        "One mail here when the PR lands or fails — the kernel owns the watch, so it survives restarts (replaces the mortal shell loop)"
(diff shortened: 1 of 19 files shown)


henrypinkard:
Merging on the owner's word. Your branch now carries a merge of current main (`1e27c135`): the collision with the per-session env change was six additive unions (the parked-op kinds gain both `env` and `cwd`, the env chip and the move chip render side by side, main's learned-version `_learn_model(..., raw=)` call keeps its argument beside your move-settle branch, the two new backend attributes coexist, and the env test's tuple pin lists both kinds). Suites on the merged head: 6853 pytest, 2731 node, typecheck, build — all green.

The review found the design careful and event-keyed: canonicalisation before anything is written, the dormant session revived in its OLD folder first, `busy` answered before any write, the two-phase `cwdPending` claimed under the reg lock (so two concurrent moves cannot both send), the lost-reply case decided by where the transcript actually is (and the both/neither case left loudly for a person), and the turn-end counter as the retry event. Two items the owner is aware of, not blockers: the private SDK sender (isolated in one helper with a named failure, exactly right), and that a move auto-accepts the CLI's trust prompt for the new folder and latches trust durably in the CLI's config — the folder pick as the trust decision is defensible, but worth one sentence in the docs so nobody is surprised later. One nit: `true_case` duplicates the kernel's `_true_case` (the comment says why; fine). Thanks for the protocol verification against the real CLI — it made this reviewable.

PR #889, opened 2026-09-02
Per-session env vars: `romp new --env NAME=VALUE` rides POST /new into the per-session flag-settings payload

ksarma:
## Problem

Two SDK sessions in the same directory cannot run with different environments. Env reaches a session only through the directory-scoped `.claude/settings*.json`, which applies to every session in that repo and outlives all of them. The per-session mechanism already exists: `flag_settings_path` hands each CLI a per-sid settings file (the `ultracode`/`fastMode` carrier), and settings files accept an `env` block.

Reproduce today: start two SDK sessions in one repo and try to give only one of them `FEATURE_FLAG=1`. There is no door — `romp new` has no env flag, POST /new ignores an `env` key, and the settings-file route reaches both sessions.

## What ships

Spawn-time slice: no dashboard surface, no separate live-mutation op.

- **CLI.** `romp new --env NAME=VALUE <name>`, repeatable, split on the first `=` (values may contain `=`; an empty value is meaningful). A missing `=` or a NAME outside `[A-Za-z_][A-Za-z0-9_]*` is a usage error (exit 2). `romp new --no-env <name>` sends the explicit empty declaration and clears a running session's per-session env. Both refuse loudly with `-t` (a tmux session's CLI reads the tmux server's environment). The `/new` leg now calls curl without `-f`, so the kernel's 400 body reaches the user instead of reading as "kernel not reachable"; a real connection failure still says so. The older-kernel warning is per asked key: a kernel that echoes model but not env warns about `--env` specifically.
- **POST /new** takes `"env": {NAME: value}`. It is validated before anything is created: non-object payloads, bad names, non-string values, NUL bytes, and falsy junk (`false`/`0`/`""`/`[]`) all answer 400 naming the offender; only absent/`null` means "not asked". On a fresh SDK spawn the reg is born with the env, so the eager first connect already carries it. On the idempotent `existing:true` open it re-asserts through the park-aware `_set_env_or_park` (same FIFO as `/model` and `/effort`, replace-in-place on repeat). Echoed back like model/effort. A live tmux target, or `backend: tmux`, refuses with `ok:false` rather than dropping the env.
- **Backend.** Reg key `env` beside model/effort. `flag_settings_path(..., env=, log=)` folds it in beside `ultracode`/`fastMode`; the return-`""`-when-no-keys contract stands, and an unwritable file now logs a problem row naming the dropped keys (the session still launches). `_options` rewrites the file from the session at every connect, so reconnects re-assert the env by construction. `set_env` mirrors `set_effort`: validate, locked persist via `_update_reg`, reconnect to apply; an unchanged re-assert skips the reconnect; never a remembered default (`write_sdk_default` is not touched). Forks inherit env like model/auth.
- **Replace, not merge.** A re-run declares the session's full env; vars not re-named drop. `{}` is that contract's limiting case and is how a spawn-time var is removed from a running session.
- **Reserved names.** `ROMP_SID` and `ROMP_SESSION_NAME` are refused in both validator copies: `options.env` owns them, and a user var of either name would shadow or be shadowed with nothing pointing at the cause. A reg written before the rule still launches, with just those names skipped (loudly, in the Log); a fork drops them from the inherited copy, also loudly.

## Documented boundaries

- The rank of this layer's `env` against a same-named directory `.claude/settings*.json` var is **not verified** live; the docstring and `docs/reference.md` say so. Distinct names, the toggle case this is for, need no ranking.
- Values are copied into per-sid files and the session registry, so this is for toggles (`FEATURE_FLAG=1`, `CLAUDE_CODE_*`), not secrets; the docs say where credentials go instead.
- Two validator copies exist by necessity (`kernel._env_error` at the /new door, `sdk_backend.env_request_error` behind the lazy `_sdk()` import). `ValidatorLockstep` pins them verdict-for-verdict, messages included.

## Tests

- `tests/test_session_env.py` grows from the existing identity-env pins (kept unchanged at its end) to: the validator table (`EnvRequestError`), the flag-settings merge and loud OSError path (`FlagSettingsEnv`), spawn persistence and refusals (`SpawnEnv`), the per-connect rewrite that makes reconnects re-assert (`OptionsThreadsEnv`), `set_env` (change/unchanged/replace/clear/dormant), fork inheritance, the legacy reserved-name skip at `_options` and `fork` (`LegacyReservedEnv`), the lockstep pin, and source pins on the park/drain plumbing.
- `NewRouteEnv` in `tests/test_new_route_prefs.py`: the 400s, null-means-not-asked, born-into-spawn with echo, existing-session re-assert, `{}` clears, both tmux refusals (including that no spawn thread starts).
- `tests/romp.bats`: the seven `--env`/`--no-env`/help cases, the per-key ack warning, and the 400-body-vs-not-reachable pair (the curl stub honours `-f` the way real curl does, so the test proves the flag, not just the message).
- `ui/webview/fast-toggle.test.ts`: the kernel park-tuple pin gains `"env"`.
- Fake `spawn()` doubles and the `_create_sdk_session` source pins in three other test files take the `env=` parameter.

Red-first: with the code hunks reverted, 46 pytest cases, 7 bats cases and the webview pin fail; all green with them. Full suite: pytest 6581 passed / 40 skipped, bats 102/102, vscode-extension 2674/2674, typecheck clean. Fixtures are synthetic (`FEATURE_FLAG`/`UI_THEME`, placeholder sids).

## Note on #882

#882 (model pickers) edits the docstring of `_apply_new_session_prefs` and the `--model` help/comment lines in `bin/romp`; this PR adds adjacent lines (a first-line clause and appended sentences in that docstring; `--env` lines after the `--effort` ones). Whichever lands second may need a trivial docstring merge; the rest auto-merges.

### DIFF
--- a/bin/romp
+++ b/bin/romp
@@ -635,6 +635,8 @@ EOF
     _romp_cmd "romp new -m <text> <name>"  -            "Start it AND deliver <text> as its first prompt (one command, idea to working)"
     _romp_cmd "romp new --model <id> <name>" -          "Model for the SDK session, as a FULL id (e.g. claude-fable-5); re-asserted if <name> already runs"
     _romp_cmd "romp new --effort <level> <name>" -      "Reasoning effort for the SDK session (e.g. high, ultracode); re-asserted if <name> already runs"
+    _romp_cmd "romp new --env NAME=VALUE <name>" -      "Per-session env var for the SDK session (repeatable; toggles, never secrets); a re-run on a running <name> replaces the whole set"
+    _romp_cmd "romp new --no-env <name>"   -            "Clear a running SDK session's per-session env (declares the empty set)"
     _romp_cmd "romp resume [<id>]"         -            "Resume a past conversation (full-screen picker; or an exact UUID)"
     _romp_cmd "romp status"                romp-manager "Show manager + kernel status"
     _romp_cmd "romp refresh"               romp-manager "Restart the postal bus + all romp kernels (picks up new code — run after an update)"
@@ -1647,6 +1649,8 @@ detach=false
 dir_arg=""
 model_arg=""
 effort_arg=""
+env_args=()
+no_env=false
 use_tmux=false
 msg_arg=""
 tag_arg=""
@@ -1660,7 +1664,7 @@ case "${1:-}" in
                 --detach)  detach=true; shift ;;
                 -d)        shift
                            if [[ $# -eq 0 || "$1" == -* ]]; then
-                               echo "usage: romp new [-t] [-d <dir>] [-m <text>] [--tag <label>] [--model <id>] [--effort <level>] <name>" >&2; exit 2
+                               echo "usage: romp new [-t] [-d <dir>] [-m <text>] [--tag <label>] [--model <id>] [--effort <level>] [--env NAME=VALUE] [--no-env] <name>" >&2; exit 2
                            fi
                            dir_arg="$1"; shift ;;
                 -m|--message) shift
@@ -1669,7 +1673,7 @@ case "${1:-}" in
                            # last two-step on the idea-to-working path). Empty text is a usage
                            # error, not a silent no-op.
                            if [[ $# -eq 0 || -z "$1" ]]; then
-                               echo "usage: romp new [-t] [-d <dir>] [-m <text>] [--tag <label>] [--model <id>] [--effort <level>] <name>" >&2; exit 2
+                               echo "usage: romp new [-t] [-d <dir>] [-m <text>] [--tag <label>] [--model <id>] [--effort <level>] [--env NAME=VALUE] [--no-env] <name>" >&2; exit 2
                            fi
                            msg_arg="$1"; shift ;;
                 --tag)     shift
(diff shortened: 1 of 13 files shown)


henrypinkard:
Merging on the owner's word, with one commit folded onto your branch (`01b0baa7`) on secret handling, which the body did not raise: env values can be tokens, and they now live in three places — the reg, the per-sid flag-settings file the CLI reads, and (while parked) the chat chip. The settings file was written with the default umask (world-readable on a shared host); it is created 0600 now, the serve-token treatment, and a pre-existing file is tightened too. The parked chip rendered `NAME=VALUE` to everyone looking at the dashboard; it renders the sorted names only (`/env A_FLAG B_TOKEN`, `/env (cleared)` for the empty declaration). Both pinned in `tests/test_session_env.py`. Everything else stands as shipped: the mirrored validators with the lockstep pin, the reserved identity names skipped loudly at the apply seam, replace-not-merge, fork inheritance, and the honest note that this layer's precedence against a directory `.claude/settings` `env` of the same name is unverified. Suites green on the branch (6588 pytest, 2674 node; one order-dependent timing flake in `tests/test_ship_reship.py` passed three times in isolation). Clean, careful work.

PR #895, opened 2026-09-03
The workflow-progress parser says so, once, when the CLI's list is a shape it cannot read

lczh:
Follow-up to #884, covering the two suggestions from its merge note.

## Shape diagnostic

The retirement parser keys on the field names a probe of CLI 2.1.257 recorded: an entry typed `workflow_agent`, its `index`, its `state` from `start`/`progress`/`done`/`error`, and an `agentId` on every started slot. A rename inside that list would bring the ever-growing live count back silently. So a list this build cannot read is now reported once per process on stderr, naming what arrived and what the user will see go wrong, on the same pattern the api_retry payload already uses (`_retry_shape_warned`).

The shapes the CLI ships on an ordinary run stay silent, so the latch is never spent on a false alarm:

- a list of only phase (and log) entries, which a run announces before any agent is queued;
- a queued slot with no `agentId` yet;
- an error slot with neither `agentId` nor `startedAt`, written for an agent blocked or thrown before its id was minted.

Only an entry of a type this build has never seen counts as a rename when no agent entry is present, and "started" means `startedAt` or a `progress`/`done` state, never an error state. The consequence line states the worst case truthfully: under a type or id rename the run's roster stays empty, so a run's end retires nothing and the agents stay until the session reconnects. The `workflow_progress` key itself is not checked, since a payload without the list is indistinguishable from a throttled tick; the docstring says so.

## Wording

The docstrings no longer claim the CLI fires `SubagentStop` only for clean finishes. What the probe showed is narrower: a workflow agent whose model did not exist got a start hook and no stop hook. The bundle does attempt the stop hook on an interrupted query, so the observed gap is failures, and the comments now say exactly that.

## Tests and verification

- `WorkflowProgressShapeIsLoud` in `tests/test_sdk_backend.py`: a renamed list is named once with the worst-case consequence; every recorded ordinary shape is silent; each missing field and an unknown entry type is named; an unknown state word is named; a throttled tick is not a shape.
- Full Python suite and node suite green on the branch.
- A two-lens adversarial review with one refuter per finding confirmed three defects in the first draft (the phase-only list, the id-less error slot, the consequence line); all three are fixed here, and the refuter for the phase-only case reproduced the shape against the installed CLI before the fix.

Unrelated observation while running the suites: `tests/test_kernel_update.py::CheckLoop::test_two_cadences_one_loop_and_a_crash_never_kills_the_thread` fails about one run in three on pristine `main` as well as on this branch, always with the drift probe never having run, so the loop is exiting on the two mocked naps before reaching it. Likely the unmocked converge check between them. Not touched here.

### DIFF
--- a/kernel/sdk_backend.py
+++ b/kernel/sdk_backend.py
@@ -1863,9 +1863,10 @@ class SdkSession:
         self._sub_lock = threading.Lock()            #   hooks mutate on the loop thread; snapshot() reads from the kernel thread
         #                                                (guards _subagents AND _bg_tasks)
         self._wf_agents: dict[str, set] = {}         # LIVE Workflow runs: task_id -> the agent ids its progress
-        #   lists have named (the roster). The CLI fires SubagentStart for every workflow agent but SubagentStop
-        #   ONLY for the ones that finish cleanly (probe-verified on CLI 2.1.257, 2026-09-02: a failed agent
-        #   leaves no stop hook), so a run's own per-agent list is what retires the rest — _reconcile_workflow_agents.
+        #   lists have named (the roster). The CLI fires SubagentStart for every workflow agent but NOT
+        #   SubagentStop for every one of them (probe-verified on CLI 2.1.257, 2026-09-02: a workflow agent whose
+        #   model did not exist got a start hook and no stop hook), so a run's own per-agent list is what retires
+        #   the rest — _reconcile_workflow_agents.
         self._wf_slots: dict[str, dict] = {}         # task_id -> {slot index: the agent id it last held}: a retried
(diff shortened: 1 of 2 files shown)


henrypinkard:
Merging on the owner's word. This is exactly the follow-up asked for on the subagent-count fix, done carefully: the shape diagnostic keys only on the fields the parser reads, the CLI's ordinary shapes (phase-only lists, a queued slot without an id, an error slot with neither id nor start) are proven silent so the once-per-process latch is never spent on a false alarm, the consequence line states the worst case truthfully, and the docstrings now claim only what the probe observed. Suites green on the branch merged with current main (6995 pytest, 2738 node, build); the one red check was the known runner hang, green on rerun. Thank you for the quick turnaround.

PR #896, opened 2026-09-03
Let SDK sessions request thinking summaries, and render a summary instead of the "Thinking…" placeholder

ksarma:
## What breaks

Every thinking block an SDK session writes back is signature-only. The chat shows "Thinking…" for each one, and the transcript on disk carries no reasoning summary, so an agent's reasoning cannot be read after the fact.

The cause is in the CLI, read in the 2.1.257 binary and re-read at 2.1.258. Its thinking-display resolver uses an explicit `--thinking-display` when one is passed, consults `settings.json`'s `showThinkingSummaries` only in interactive mode, forces `omitted` only for `--print` text/json output, and on the SDK's `stream-json --verbose` path requests no display at all. The API default then applies, and for current models that is signature-only (claude-agent-sdk's `types.py` says so for Opus 4.7+). romp's `_options` passes `effort=` and no thinking config, so nothing ever asks for summaries.

## Reproduction

On CLI 2.1.257 or 2.1.258, open any SDK session's transcript under `~/.claude/projects/`: every `thinking` block has a `signature` and an empty `thinking` string, and in the chat every thinking turn reads "Thinking…". A counts-only census of local transcripts (no content read out) found zero thinking blocks with both a signature and text on CLI 2.1.221 through 2.1.247 (15,563 signed blocks in total); such blocks appear only on the two versions where the flag below was passed.

## The fix, in three commits

1. **`kernel/sdk_backend.py`**: a connect-time setting read on every connect, like `effort`. When `STATE/thinking-summaries.json` says enabled, `_options` passes the SDK's typed `ThinkingConfigAdaptive` field, `thinking={"type": "adaptive", "display": "summarized"}` (claude-agent-sdk 0.2.132; the transport emits `--thinking adaptive --thinking-display summarized`), never `extra_args`. Off passes nothing, so the CLI's default stands unchanged. The decision is a pure helper (`thinking_kw`), so the standard test runner, which has no `claude_agent_sdk` and skips the `_options` pins, still exercises it; `tests/README.md` says how to run the gated classes with the SDK venv on `PYTHONPATH`.

2. **`kernel/kernel.py` `build_session` and `ui/webview/render.ts`**: `encrypted` now means opaque, a signature AND no text. A signed block with text is a summary and renders through the existing two-line clamp (click to expand). The renderer re-checks the text itself, so a bundle served by an older kernel, whose flag still means "signed", shows what it is handed. Judges do not read thinking text: their readers key on `type == "text"`, and a test pins that.

3. **The setting**: per-install, OFF by default, its own json under `STATE`, a gear checkbox "Thinking summaries" between Conserve memory and File editing, reported by `/version`. It is deliberately not in federation's `KERNEL_SETTING` set: each kernel keeps its own answer and nothing propagates (whoever reads the summaries turns it on where they read them). For the same reason it sits top-level in `/version` and not in the `settings` sub-dict, whose mixed marks promise a cross-machine write. It follows the same gesture ordering as the other kernel-side settings, because two dashboards on one kernel can race: the post carries `gt`, the store applies under `_SETTINGS_LOCK` in gesture order, a stale stamp stands down and answers the delivering socket with `settingStale` (the gear toasts it under the row's name via `STALE_LABELS`), an equal stamp carrying the stored value is a silent echo (the `_gesture_echo` check the other five gt-gated setters have), and a failed write is a loud `None` that applies nothing.

## Two limits, both stated in the gear copy

- A running session is not switched live. The python SDK wraps no runtime thinking-display control (the CLI's `set_max_thinking_tokens` control request carries a `thinking_display`; 0.2.132 does not expose it). The setting applies at the session's next reconnect: an effort, billing, per-session-env or bypass-mode switch, the first fast-mode opt-in, a file rewind, a kernel restart. A model switch applies live over the control channel and does not reconnect; the copy says so.
- The flag is not display-only. The CLI resolves `--thinking adaptive` before a `MAX_THINKING_TOKENS` in its environment and before `alwaysThinkingEnabled: false` in settings (read in the binary), so on an install that had thinking off, asking for summaries also turns adaptive thinking on, at that cost. The SDK has no display-only field. `_options` logs one line per backend when a cap is present in the CLI's environment; the settings half is not readable from the backend, so the gear copy carries it.

## Verified behavior

CLI 2.1.257 and 2.1.258, claude-agent-sdk 0.2.132 (`ThinkingConfigAdaptive.display`): on a running install with the toggle on, new SDK sessions write thinking blocks carrying both a signature and summary text into their transcripts, and the chat shows the clamp. With it off, blocks stay signature-only exactly as before. Counts only, from local transcripts: 10 signed+text blocks on 2.1.257 and 271 on 2.1.258, beside 5,174 and 1,493 signed-only blocks on the same versions; zero on every earlier CLI.

## Default and version floor

Shipped OFF. Default-ON is one boolean in `_thinking_summaries_on`'s fallback plus the fresh-install test.

`bin/romp-sdk-setup` installs `claude-agent-sdk` unpinned; the typed `thinking` field and `ThinkingConfigAdaptive.display` are present in 0.2.132. On a venv that predates the field, turning the toggle on makes `ClaudeAgentOptions(**kw)` raise `TypeError` at connect, a loud failure rather than a silent one; `effort=` has the same exposure today.

## Tests

- `tests/test_sdk_backend.py`: `ThinkingKw` (the SDK-free decision and the override note); `OptionsAssembly` (the typed field, the transport flags, one override log line per backend, none without a cap). Run with the SDK on `PYTHONPATH`, the 5 new tests pass; the one remaining failure in that class (`test_ultracode_…`) is pre-existing at the base and unrelated to this change.
- `tests/test_thinking_summary_events.py`: the opaque rule through `build_session` on a synthetic transcript; the judges' readers skip thinking.
- `tests/test_setting_gesture_order.py`: `ThinkingSummariesSetting` (store, stale stand-down, silent echo, a pre-field file reads as gt 0, the WS branch answers the socket and propagates nothing, write failure, `/version`); the `settingStale` cases-table row; the WiringPins counts now cover five toggles.
- `tests/test_kernel_settings_sections.py`: section placement of the new row.
- `ui/webview/thinking-summaries.test.ts`: the render rule, the kernel's matching expression, the gear row and copy, the stamped post, the `STALE_LABELS` entry, and that it is not a `KERNEL_SETTING`.

Each new test was red against the base before its code landed. Full runs: 6876 Python tests pass (`pytest -n 8`), 2735 node tests pass, `tsc --noEmit` clean, bats 375/375. All fixtures synthetic.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -20597,8 +20597,13 @@ def build_session(sid, now, tmux=None, path_override=None, tail_cap_t=None):
                     if not isinstance(b, dict):
                         continue
                     if b.get("type") == "thinking":
+                        # `encrypted` means OPAQUE — nothing to show but the placeholder: a signature and
+                        # no text. A block with a signature AND text is a summary (the SDK's display
+                        # "summarized", requested when the thinking-summaries toggle is on) and renders
+                        # its text; the old signature-only rule hid every summary (2026-09-01).
                         events.append({"kind": "thinking", "text": b.get("thinking", ""),
-                                       "encrypted": bool(b.get("signature")), "uuid": a.get("uuid"), "ts": ts})
+                                       "encrypted": bool(b.get("signature")) and not (b.get("thinking") or "").strip(),
+                                       "uuid": a.get("uuid"), "ts": ts})
                     elif b.get("type") == "text" and b.get("text", "").strip():
                         txt = b["text"]
                         if a.get("command"):             # a slash command's OUTPUT (command:True on the synthetic
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -573,6 +573,9 @@ def _version_info():
             "login": _login_state(),         # the in-dashboard login flow's state/url/err (T157) — never a secret
             "acctLabel": _claude_account_label(),   # which login this box holds ("" = none) — the gear's Billing row
             "fileEditing": _file_editing_on(),   # dashboard raw-mode editing opt-in → gates the viewer's Edit
+            # per-install: SDK sessions ask for reasoning summaries (gear checkbox). Top-level only — not
+            # in the "settings" sub-dict below, whose mixed marks promise a cross-machine write this never makes
+            "thinkingSummaries": _thinking_summaries_on(),
             "updateMode": _update_mode(),    # ask|auto|off (the boot release check) → the gear dropdown
             "updateAvail": _UPDATE_AVAIL[0],   # newer release the boot check found ("" = none/unknown)
             "judgeModel": jd._triage_model(), "indexModel": jd._index_model(),      # current per-tier judge models → the gear dropdowns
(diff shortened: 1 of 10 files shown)


henrypinkard:
Merging on the owner's standing word for your PRs, as shipped — nothing to fold. Reviewed inline: the connect-time read mirrors effort exactly (re-asserted on every connect, picked up at the next reconnect, honest that a model switch is not one), the gt-gated setter carries the echo check the other five setters have, the toggle is per-install and correctly kept out of the federation set and out of `/version`'s settings sub-dict, `_options` builds its `env` before the thinking block reads it, and the render half's re-check of the text keeps an older kernel's flag honest. The gear copy stating that asking for summaries turns adaptive thinking on where a cap had it off is the right kind of honesty. Suites green on the branch merged with current main (6995 pytest, 2738 node, build).

PR #897, opened 2026-09-03
The compaction suggestion claims its latch before it sends, and every nudge-ledger writer holds the lock

ksarma:
## What breaks

`_compact_suggest_tick` can inject the `/compact` suggestion twice into one session. It reads the per-session `compactSuggested` latch under `_NUDGE_LOCK`, releases the lock, runs its fire-time gates (roster tag, thread fork, open turn, and the idle gate through `_settle_event_key`), sends, and latches only after the send. The tick rides `_auto_nudge_tick`'s alive walk, which has two concurrent entry points on main: the pusher's periodic pass and the `setCompactSuggest` WS handler's act-now re-tick (`run_dead_wait=False`). Two entries that pass the same unlatched check both send.

The latch is also only as durable as the ledger's other writers. `_fire_debt_reminder` held a snapshot of the whole auto-nudge blob across a live send and wrote it back whole; `_debt_reminder_outcomes`, `_debt_backstop_tick`, `_set_intr_blocked`, `_nudge_deferred_ok`, and `_deferral_sweep_tick` ran unlocked read-then-write spans over the same file. A whole-blob write is last-writer-wins over every key, so any of these straddling a latch write erases it, and the next tick re-derives the crossing as due and sends again.

## Reproduction

Both are deterministic two-thread tests in `tests/test_compact_suggest.py`, gated on events (no sleeps), in the same idiom as `TwoFlushRace` in `tests/test_setting_gesture_order.py`:

- `ConcurrentSendRace.test_two_concurrent_entries_inject_once`: the first entry is held inside `_settle_event_key` (between its latch check and its send) until a second entry runs to completion. On main: two sends (`2 != 1`).
- `DebtWriterRace.test_a_debt_writer_straddling_a_claim_never_erases_it`: a debt reminder is held inside its send while a compact suggestion latches; the reminder's record write then lands. On main: the latch is gone (`[] != [400000]`) and the next tick sends the suggestion a second time.

## The fix, in two commits

**1. The latch write becomes the claim** (`_compact_suggest_tick`). Under `_NUDGE_LOCK`, re-read the blob, re-derive what is still due against the fresh latch, and stand down (`return False`) if a concurrent entry already claimed it; otherwise latch, then send. A failed send rolls the claim back under the lock so the next tick retries the same crossing, which keeps the existing rule that a fire that never went out is never left latched. The post-send latch block is deleted.

**2. Every remaining ledger writer holds `_NUDGE_LOCK` and mutates only its own keys.** The writers take one of two shapes. A span that is pure dict work holds the lock across the whole span (`_set_intr_blocked`, `_nudge_deferred_ok`; the idiom the settings writers use since #879). A writer with IO between evidence and write (`_fire_debt_reminder`'s send, `_debt_escalate`'s goal-store writes in the two debt sweeps, `_deferral_sweep_tick`'s transcript reads) does that IO outside the lock, then re-reads the blob fresh under it and pops or sets only the keys it decided. This finishes the migration the lock's comment deferred "as touched"; the comment now describes the two shapes. No new lock ordering: `_atomic_write` takes only its own short counter lock, and the episode writers already hold `_NUDGE_LOCK` across the same call.

The second commit is in this PR because the first is not durable without it: the debt writers can erase the claim, and can also resurrect a rolled-back claim after a failed send (`DebtWriterRace.test_a_debt_writer_straddling_a_rollback_never_resurrects_the_latch` is red against commit 1 alone, the first point at which there is a claim to resurrect).

## Tests

`tests/test_compact_suggest.py`: the 18 existing tests are unchanged (their class becomes a shared `_Fixture`); 22 total.

- `ConcurrentSendRace` (2): inject once under the race; a failed send unlatches and the next tick retries.
- `DebtWriterRace` (2): the erasure straddle and the resurrection straddle, each also asserting that the debt writer's own `debtNudged` key still lands, so neither writer's output is lost.

Red-first progression: on main, inject-once and erasure fail and the other two pass (the old code never latched a failed send). After commit 1 alone, inject-once passes; erasure and resurrection fail. After commit 2, all four pass. The race tests ran green 10 of 10 in a loop; their waits are 10 s event timeouts, so a starved runner surfaces an assertion rather than a hang.

Also run: `test_kernel_debt_nudge`, `test_kernel_deferral_sweep`, `test_kernel_nudge_last_resort`, `test_kernel_interrupt_blocks`, `test_kernel_stalled_blocks`, `test_nudge_unblock_hold`, `test_auto_nudge_every_kernel`, `test_setting_gesture_order`, `test_injected_voice`, `test_kernel` (these patch the ledger accessors directly and drive every migrated writer): 620 passed. Full suite: 6860 passed, 45 skipped. Synthetic ids only.

## Reviewer notes

- `_set_auto_nudge` and `_set_compact_suggest` are untouched; #879 already put them under `_NUDGE_LOCK`.
- A lock audit (nearest enclosing `with _NUDGE_LOCK:` for every `_write_auto_nudge(` call) reports zero unlocked sites after this change. Main has eight, in the six functions above.
- A send failure now writes the blob twice (claim, then rollback) where the old code wrote nothing. Both are atomic publishes.
- Not claimed fixed: two concurrent `_fire_debt_reminder` calls for one debtor can still both pass the pre-read and both send. That race pre-exists; this change only stops the reminder's record write from erasing other writers' keys.
- The comment inside `_auto_nudge_tick` attributing the `run_dead_wait=False` re-tick to `setAutoNudge` is pre-existing and left alone; on main the re-tick is `setCompactSuggest`'s.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -5350,6 +5350,29 @@ def _compact_suggest_tick(sid, tm, now):
     if not _st or now - _st < idle_s:
         return False                                   # not settled long enough — the crossing stays
     #                                                    armed; a later tick re-checks the same gate
+    with _NUDGE_LOCK:                                  # CLAIM BEFORE SEND (the double-send race,
+        # 2026-09-01): the check above releases the lock before the fire-time gates, and this tick
+        # has two concurrent entry points (the pusher's periodic pass and the setCompactSuggest
+        # handler's act-now re-tick) — latching only after the send let two entries pass the same
+        # unlatched check and inject the suggestion twice into one session. The latch write IS
+        # the claim: a fresh read under the lock re-derives what is still due (a concurrent entry
+        # may have claimed it in the gap), the winner latches and sends, the loser stands down
+        # right here. Same direction as the nudge machinery's own bookkeeping (_mark_auto_nudged
+        # records only after its send, so a failed send retries), inverted at this seam because
+        # the send must be single-shot — and the failure arm below rolls the claim back under the
+        # lock, so "never latch a fire that never went out" still holds durably: the next tick
+        # retries the same crossing.
+        d = dict(_auto_nudge_data())                   # fresh read — a concurrent writer's blob is
+        cs = dict(d.get("compactSuggested") or {})     # not clobbered
+        held = [t for t in [int(x) for x in cs.get(sid, [])] if tokens >= t]
+        due = [t for t in due if t not in held]
+        if not due:
+            return False                               # a concurrent entry already claimed this fire
+        cs[sid] = held + due
+        #                                                BOTH thresholds latch when found past both —
+        #                                                one message, never two in a tick
+        d["compactSuggested"] = cs
+        _write_auto_nudge(d)
     try:
         # the marker tail its sibling injectors carry (T207, the user 2026-08-31, who saw the
         # bare send render as their own blue bubble and ruled it must read system-injected):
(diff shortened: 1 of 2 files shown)


henrypinkard:
Merging on the owner's standing word for your PRs, as shipped — nothing to fold. Reviewed inline: the claim-before-send latch with the failed-send rollback under the lock is exactly right, and finishing the ledger-writer migration in the two shapes (pure dict spans under the lock; IO outside, then a fresh re-read mutating only the writer's own keys) is what makes the claim durable — the debt writers' pre-send whole-blob write was a real side door. The two-thread race tests are event-gated with no sleeps, in the TwoFlushRace idiom. The one red check on this head was the bats job hitting `Address already in use` on a runner (a fixed-port collision class, rerun requested); the local bats run on the branch merged with main is in the merge note's follow-up. Suites green on that merged tree (6995 pytest, 2738 node, build).

PR #901, opened 2026-09-03
_api_error reads a tail window instead of the whole transcript per append

christianazinn:
## Why

`db479cd3` ("Appends fold through the parse instead of re-emitting the whole session") amortized the event-model parse. `_api_error` sits beside it on the same hot path, with the same `(mtime,size)` cache in front of the same whole-file read — and was left behind. Every append busts that key, so `build_feed`/`build_session` re-read and re-`json.loads`'d the **entire** transcript, per session, per push cycle.

Profiled on a live 11-session install (transcripts 20–116 MB, ~475 MB total):

- `Thread-5 (_pusher)` held **88% of the kernel's CPU** — 7.2M of 8.15M ticks, ≈20 CPU-hours in a 22h uptime.
- 9 of 12 `py-spy` samples of that thread sat in `_api_error` / `raw_decode` / `read_text`.
- `/version` on the same kernel reported `parse: {full: 927, fold: 14213, fallback: 0}` — the fold was working at **93.9%** and buying nothing visible, because this scan was next to it.

Symptoms were a pusher that couldn't finish a cycle: panes minutes stale, and a reconnected pane never receiving the non-keepalive frame that retires the stale-connection banner.

## What

The verdict depends only on the records from the **last deciding one** onward — an `isApiErrorMessage` assistant record, fresh assistant output, or a genuine user prompt — because anything earlier is overwritten. So the scan starts near the tail and widens 4× until a pass reports it actually saw such a record.

`decided` is the window's own proof of sufficiency rather than a heuristic: it's set at the three sites that assign the verdict, so `False` means "started too late, widen" and `True` means this window's answer **is** the whole file's answer. A transcript whose tail genuinely carries no verdict still reads in full, one widen at a time.

`model_refusal_*` records need no special casing — they land after the error they annotate and only mutate an `err` already set, so with their error out of window nothing is assigned and the caller widens.

## Reviewing this

The classification is unchanged **by construction**. `git diff -w` removes no loop-body line — only relocated scaffolding plus the three `decided = True` markers. The loop is the original, dedented one level (it loses the `try:` nesting).

## Tests

`tests/test_api_error_tail.py` (13 tests) pins the part that's new: the windowed read equals the whole-file read for every transcript shape, including the widen cases.

Red first — an earlier draft nested the loop inside the seek guard, so the `start == 0` pass scanned nothing. It compiled clean, and 7 of these 13 tests fail on it.

Also run green: `test_kernel_apiretry`, `test_judge_apierror_no_caption`, `test_judge_apierror_no_fabricated_done`, `test_judge_orphan_replies`, `test_judge_nudge_wedge_moot_retire`, `test_idle_queue_drive`, `test_event_model_{incremental,assembly,golden}`, `test_kernel_{send_park,ops_gate_busy,limit_queue}`.

Unrelated note: `tests/test_judge_auth_billing.py` fails 3 tests on clean `main` on my machine, identically with and without this change (5 runs) — flagging in case it's not already known.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -17495,7 +17495,7 @@ def _api_error(path):
         # TAIL-FIRST: the pusher calls this per session per push, and the old whole-file read cost
         # O(transcript) on EVERY append — the same unamortized shape the assembly fold retired for the
         # event-model parse, left behind here. Widen 4x until a pass reports it saw the deciding record.
-        win = _API_ERR_TAIL_WINDOW
+        win = max(1, _API_ERR_TAIL_WINDOW)   # a knob of 0 or less would never widen (0*4 stays 0): clamp, don't spin
         while True:
             start = size - win if win < size else 0
             err, decided = _api_error_scan(path, start)
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -17375,6 +17375,102 @@ def _spend_dialog_showing(pane):
             and ("What do you want to do?" in pane or "Enter to confirm" in pane))
 
 
+# _api_error's scan window. Its answer depends ONLY on the records from the LAST DECIDING one
+# onward, so the scan starts near the TAIL and widens until it provably saw one — a session's whole
+# transcript is re-read only when its tail genuinely carries no verdict. Overridable for tests.
+_API_ERR_TAIL_WINDOW = int(os.environ.get("ROMP_API_ERR_TAIL_WINDOW", "262144"))
+
+
+def _api_error_scan(path, start):
+    """One pass of _api_error's classification from byte `start` to EOF → (err, decided).
+
+    The loop is _api_error's original whole-file scan, UNCHANGED (one dedent) except that each of its
+    three `err = ...` sites now also sets `decided`. That flag is the window's own proof of sufficiency:
+    True iff this pass saw a record that ASSIGNS the verdict (an isApiErrorMessage assistant record,
+    fresh assistant output, or a genuine user prompt) — exactly the record the whole-file scan's result
+    depends on, since everything before it is overwritten. False means the window started too late and
+    the caller must widen; True means this window's answer IS the whole file's answer.
+
+    A non-zero `start` lands mid-line, so the first partial line is dropped. model_refusal_* records
+    land a few records AFTER the error they annotate and only mutate an err already set, so they need no
+    special handling: with their error out of window nothing is assigned, and the caller widens."""
+    err = None
+    decided = False
+    with open(path, errors="replace") as f:
+        if start > 0:
+            f.seek(start)
+            f.readline()                          # the window cut this line in half — drop it
+        for line in f:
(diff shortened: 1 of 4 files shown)


christianazinn:
Correction to the "Why" section above — the attribution is wrong, though the fix isn't.

I credited `build_feed`/`build_session` as the hot callers. Both are actually **guarded**:

```python
aerr = _api_error(path) if not (open_now or awaiting_why) else None     # :19275, :21270
```

so they skip a session with an open turn entirely — i.e. exactly the working sessions whose transcripts are being appended to. They were never the cost.

The unguarded hot caller is the spend-cap sweep, from `_auto_pause_on_spend_limit` on the tick (`:28548`):

```python
def _spend_capped_session(now, tmux):
    for s in _alive_sessions(now, tmux):
        p = s.get("path")
        if p:
            e = _api_error(p)
            if e and e.get("spendLimit"):
                return s          # early-outs only on a HIT
    return None
```

It early-outs only when it *finds* a capped session, so the healthy state — no cap anywhere — is the worst case: every alive session, each read to the end. That's what the profile was actually catching, and it's why the cost is proportional to (sessions × transcript size × append rate) rather than to anything error-related. On the profiled install that was 11 sessions and ~475 MB per sweep whenever the appends had busted the cache.

Nothing about the patch changes: making `_api_error` itself tail-first makes every one of its ~10 call sites cheap, including this sweep. But if you'd prefer the sweep to also stop asking about sessions it can't be blocked on, that's a separate and complementary change and I'm happy to write it.

henrypinkard:
Thanks for this, and for the correction on the caller. Verified against main: `_spend_capped_session` runs `_api_error` unguarded for every alive session on the tick, while `build_feed` / `build_session` are guarded by the open turn. The value of the fix is unchanged.

One commit pushed onto your branch (e050bdbe) so this can land without a round trip:

- The differential tests compare the windowed answer to the whole-file answer, which a scan that always starts at byte 0 also satisfies, so nothing pinned the partial read itself. Three tests now record each pass's start offset: one nonzero start for a decided tail, strictly earlier starts across a widen, and a non-None answer (refusal included) from a mid-file window.
- `ROMP_API_ERR_TAIL_WINDOW=0` (or negative) made the widen loop spin forever (0*4 stays 0). The driver now clamps to one byte; a test drives 0 and -7 through a recorder that fails instead of hanging.
- The test docstring's word for a group of sessions is on the repo's banned list (CLAUDE.md, vocabulary rule); replaced with "install". The first commit message carries it too, so this will go in as a squash with a clean message.

Full suite green locally on the folded head (7008 passed, 44 skipped).

Separately: you mention three tests in `tests/test_judge_auth_billing.py` failing on clean main. We cannot reproduce that (27/27 pass under a hermetic HOME, Python 3.10 through 3.13, with `claude`, `node` and `tmux` hidden from PATH). Could you post the three test ids, the assertion or traceback text, and your OS and Python version? Someone on our side will own the fix.

PR #902, opened 2026-09-03
Static bundles get a validator and a gzip, so no-cache can revalidate cheaply

christianazinn:
## Why

`/dist` and `/media` go out with `Cache-Control: no-cache` and **no `ETag` or `Last-Modified`**. `no-cache` means *revalidate*, but with no validator the browser has no conditional to send — so it re-fetches the whole bundle on every open. And the server offers no compression at any `Accept-Encoding`: a request with `Accept-Encoding: gzip, br` comes back with all 773,663 bytes of `render.js`.

Measured on the served dist:

```
dist js+css total   2,249,027 B   per cold dashboard load, uncompressed, every time
render.js             773,663 B
editor-chunk.js       640,298 B
feed.js               350,246 B
timeline-main.js      133,093 B
styles.css            131,456 B
GET /dist/render.js with "Accept-Encoding: gzip, br"  →  773,663 B
GET /dist/render.js                                   →  Cache-Control: no-cache, no ETag
```

On a tailnet-reached kernel (the `tailscale serve` path in the guide) that is the load time, and it's paid again on every open even though the bytes never changed.

## What

Two changes, neither touching the freshness rule:

1. **`ETag` = `(mtime_ns, size)`, and a matching `If-None-Match` answers `304`.** That tag moves on every rebuild exactly like the `?v=` stamp the URLs already carry, so the existing guarantee — "a same-name rebuild can never serve a stale body" — still holds. The browser just gets a 304 instead of the body when it already has that build.
2. **gzip when the client offers it**, for text suffixes over 1 KB. Measured **3.2× overall**:

```
render.js         773,663 → 238,631   (3.2x)
editor-chunk.js   640,298 → 221,814   (2.9x)
feed.js           350,246 → 110,877   (3.2x)
styles.css        131,456 →  24,180   (5.4x)
TOTAL js+css    2,249,027 → 699,873   (3.2x)
```

The gzip is made **once** per `(path, mtime, size)` and held beside the plain bytes. The kernel is a single process serving every pane, and on a busy install it's already CPU-bound on one core, so compressing ~750 KB per request would relocate the cost rather than remove it. The two encodings carry different etags and both responses send `Vary: Accept-Encoding`, so nothing can hand a gzip body to a client that never asked for one. A compression failure falls back to the plain asset — fail toward serving.

## Tests

`tests/test_dist_etag_gzip.py` (12 tests): gzip round trip, per-encoding etags, rebuild invalidation (both the tag *and* the served bytes, so the cache can't pin an old body), the skip rules for small and already-compressed files, compress-once, the failure fallback, and the route's 304 branch.

Unrelated note: `tests/test_judge_auth_billing.py` fails 3 tests on clean `main` on my machine, identically with and without this change (5 runs) — flagging in case it's not already known.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -11,7 +11,7 @@ zero protocol change at switchover. WS is hand-rolled on the stdlib socket (no d
 
 Run:  bin/romp-kernel   → opens http://127.0.0.1:29855
 """
-import json, os, queue, random, re, signal, socket, sys, time, threading, traceback, base64, bisect, errno, hashlib, hmac, struct, subprocess, shutil, shlex, http.client, uuid, tempfile, stat
+import json, os, queue, random, re, signal, socket, sys, time, threading, traceback, base64, bisect, errno, hashlib, hmac, struct, subprocess, shutil, shlex, http.client, uuid, tempfile, stat, gzip
 from pathlib import Path
 from datetime import datetime, timezone
 from importlib.machinery import SourceFileLoader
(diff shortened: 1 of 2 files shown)


christianazinn:
Reframing this one honestly: it's **efficiency hygiene, not a fix for a slow dashboard load.** I originally presented it as the latter and that was overstated.

Two things I should have checked before writing that framing:

- The server is `ThreadingHTTPServer` (`:35109`), so asset requests don't serialize behind each other. Any implication of head-of-line blocking in my description was wrong.
- On the install I measured, the tailnet path is **direct**, not DERP-relayed (`active; direct <addr>`). So there's no relay throughput penalty, and 2,249,027 bytes over direct WireGuard is a second or two — not the multi-second "forever to load" that sent me looking.

The load symptom on that install almost certainly shares a root cause with #901: the dashboard's *content* arrives over the WS from `_push`, on the thread that profiling showed holding 88% of the kernel's CPU. The bundles land fine and then the page waits on a payload the starved pusher can't produce. Fixing #901 is what addresses that.

What this PR still stands on, independent of any of that:

- `no-cache` currently can't do the revalidation it asks for, because there's no validator to match. That's a straightforward defect regardless of link speed — the ETag makes the conditional answerable, and the freshness rule is unchanged because the tag moves with every rebuild exactly like the `?v=` stamp.
- 3.2x on the wire for clients that offer gzip, computed once per build rather than per request.

Worth having, but please weight it accordingly against #901 — and feel free to treat it as low priority or to reject the gzip half if you'd rather not hold compressed copies in the kernel's memory.

henrypinkard:
Thanks, and the reframing is appreciated; the validator defect stands on its own, and the compress-once design is the right call for a single-process kernel.

One commit pushed onto your branch (3ef0e987): the three `RouteWiring` tests asserted that the 304 branch's lines exist in the kernel source, which a route that has them and still misbehaves also satisfies. They are replaced by six tests that drive the real `do_GET` over a fake socket (the pattern `tests/test_kernel_auth_hardening.py` uses) against a bundle written into a temp dir: tag, Vary and no-cache on the first fetch; a 304 with no body on a matching conditional; a stale tag and a same-name rebuild both get the body again under a new tag; the gzip form revalidates under its own tag while a plain-form tag from a gzip-offering client never yields a 304 for the wrong form; and the traversal guard still answers 404.

Full suite green locally on the folded head (7006 passed, 44 skipped).

Two nits left alone, for the record: the 304 carries `Content-Length: 0`, which RFC 9110 says a 304 should omit unless it equals the 200's length (browsers tolerate it), and the `"gzip" in accept_encoding` check does not honor a `gzip;q=0` refusal (browsers never send one).

PR #904, opened 2026-09-03
Parked ops deliver on the turn's settle, not at the tail of a judge pass

lczh:
## Parked ops deliver on the turn's settle, not at the tail of a judge pass

**The bug.** The kernel parks a user's input while its session is busy, compacting, or limit-held, and delivers the FIFO once the session is quiet (`_apply_pending_ops`). The only caller of that function was the tail of the judge producer's pass, after an untimed join on the judge tiers; a turn-end poke only set `_producer_wake`, which the producer reads between passes. So every parked op in every session waited for the judges to finish.

A judge pass can run for hours. On 2026-09-03 one session's closer sweep was alarm-killed turn after turn (192 calls at 120 s each) and held a pass for 6h22m; no judge ran for any session in that window, and a typed `/clear` parked mid-turn sat as a queued chip until the user cancelled it. Sending to an idle session worked because an idle send never touches the queue. The judge-side cause (the closer's unbounded rider menu and its continue-past-failure sweep) is a separate PR.

**The fix.** Delivery rides the pusher cycle, the kernel's existing event-woken at-settle job runner, as its first job; the producer's tail call is removed (not kept as a backstop: `_apply_pending_ops` is not re-entrant). Both backends poke the kernel through `_wake_kernel`, which sets the producer and the pusher wakes; the Codex backend pokes once more after its `turn_id` clears. A no-op cycle no longer re-saves or re-wakes. Two places that leaned on the producer's 3 s cadence without saying so, a move the CLI answered `busy` and a turn-opening delivery to a tmux session (no authoritative `busy()`), now hold the sid's queue for an explicit 3 s (`_hold_drain`), named as a clock; SDK and Codex need no hold. The SDK input generator counts the fed turn in flight under the same lock that popped it.

**Visibility.** `_send_or_park` returns whether it parked; `POST /send` answers `{"ok": true, "queued": <bool>}` for text sends, parked `/model` `/effort` `/fast`, and remote answers; `romp send` prints `queued` vs `ok`; a cancelled parked op is logged (sid and op kind only).

**Verification.** Full Python suite green (7039 passed, 0 failed on the head; 7048 on the maintainer's scratch merge with #905), bats green (382). The behavioral tests fail on the parent commit; two boundary pins pass there by design, and some of the new tests hit the parent's missing names (`_wake_kernel`, `_drain_hold`) before their behavioral assertion, so the headline liveness test pokes both wakes by hand when `_wake_kernel` is absent and fails at the delivery assertion itself. The first commit went through a four-lens adversarial review with two skeptics per finding; its eight confirmed findings are fixed in it (the move-retry spacing, the tmux self-wake, the parked meta-command answer, the remote `queued`, the SDK pop-to-inflight gap, stale comments, and two test-quality points).

**Review round.** The second commit takes the four nits: the `busy` re-park and its hold now land before `_moving` releases the queue; the pusher cycle memoizes sid-to-path on its thread-confined scope so a held sid's transcript path resolves once per cycle instead of once per gate; a far kernel's `ok: false` rides back verbatim; and `romp new -m` prints "first prompt queued" when the kernel parked it. The follow-up PR you asked for, a `TmuxBackend.busy()` read from the hook-maintained session state that also retires the tmux half of `_hold_drain`, is being prepared separately.

### DIFF
--- a/bin/romp
+++ b/bin/romp
@@ -1526,6 +1526,13 @@ print(json.dumps(d))' "$_who" "$_when")"
                   -H 'Content-Type: application/json' -d "$_payload")" \
         || { echo "romp $_verb: kernel not reachable on :$_kport (is romp running?)" >&2; exit 1; }
     if [[ "$_resp" == *'"ok": true'* || "$_resp" == *'"ok":true'* ]]; then
+        if [[ "$_verb" == "send" && ( "$_resp" == *'"queued": true'* || "$_resp" == *'"queued":true'* ) ]]; then
+            # the kernel PARKED it (a turn is open / compacting / the account is limit-held) and delivers it
+            # when the session is quiet. Said out loud so a sender inside that very turn — an agent sending
+            # itself a slash command — knows the command has NOT run yet.
+            echo "romp send: queued ($_who) — delivers when the session is quiet"
+            exit 0
+        fi
         echo "romp $_verb: ok ($_who)"
         exit 0
     fi
--- a/bin/romp
+++ b/bin/romp
@@ -2016,7 +2016,12 @@ print(json.dumps(d))' "$session_arg" "$msg_arg" "$tag_arg" "$_resp")"
                           -H 'Content-Type: application/json' -d "$_mpayload")" \
                 || { echo "romp new: session is up but the message did NOT land (kernel send failed). Retry: romp send $session_arg '<text>'" >&2; exit 1; }
             if [[ "$_mresp" == *'"ok": true'* || "$_mresp" == *'"ok":true'* ]]; then
-                echo "romp new: first prompt delivered"
+                if [[ "$_mresp" == *'"queued": true'* || "$_mresp" == *'"queued":true'* ]]; then
+                    # the kernel PARKED it (the fresh session is not quiet yet): it delivers when it is
+                    echo "romp new: first prompt queued — delivers when the session is quiet"
+                else
+                    echo "romp new: first prompt delivered"
+                fi
             else
                 echo "romp new: session is up but the kernel refused the message: $_mresp. Retry: romp send $session_arg '<text>'" >&2
                 exit 1
--- a/kernel/codex_backend.py
+++ b/kernel/codex_backend.py
@@ -1300,7 +1300,8 @@ class CodexBackend:
                 s.turn_id = None
                 s.state = "waiting"
                 s.since = time.time()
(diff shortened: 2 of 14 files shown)


henrypinkard:
Reviewed (own read plus a three-lens adversarial pass with two skeptics per finding): the design holds on every path traced. The settle poke wakes both loops; the drain runs first in the pusher cycle, which runs with zero clients too; `inflight` moves under the same lock as the pop; a no-op cycle no longer re-wakes its own thread; the SDK and Codex `busy()` gates close synchronously inside `send()`. Full suite green here on the head (7037 passed), and green on a scratch merge with #905 (7048 passed). No fold from our side.

One confirmed finding, pre-existing rather than introduced here, recorded so it gets its follow-up: for a tmux session `_working_now` falls to `_parse_cached`, which returns None whenever the transcript's (mtime, size) moved since the last parse, and `None or {"turns": []}` reads as idle. A tmux session mid-turn whose transcript grew since the last parse therefore reads as quiet, and the drain can fire a parked op into the open turn (a parked slash command lands as text). Main had the same predicate at the producer's 3 s cadence; this PR polls it every cycle and before `_push_all` refreshes the cache, and `_TMUX_PROMPT_HOLD_S` covers only the 3 s after the drain's own delivery. The doctrine-shaped fix is the one your `_hold_drain` docstring already names: a `TmuxBackend.busy()` from the hook-maintained `@claude-state` the cycle snapshot already carries (UserPromptSubmit flips it to working), which also retires the tmux half of the hold. Happy to take that as a follow-up PR from you.

Nits, none blocking:
- `_move_now` discards `_moving` in its `finally` before it re-parks the cwd op and calls `_hold_drain`; a cycle landing in that gap can fire the op behind the move or burn a retry. Re-park and hold before the discard closes it. Same ordering existed on main.
- The drain gate resolves `_path_of` (a discover fingerprint) two to three times per parked sid per cycle (`_compacting_now`, `_working_now` on tmux, `_limit_hold`); `_compacting_now` already accepts a `path=` hoist, so a sid-to-path map built once per cycle would remove the repeat.
- The remote-forward arm still answers `ok: true` for a far kernel's `ok: false`, on a line this PR rewrote; and `romp new -m` still says "first prompt delivered" when the reply was `queued`.
- Body says every new test fails on the parent; two pass there, and eight of eighteen fail only with AttributeError on the new names.

PR #905, opened 2026-09-03
The closer caps its riders per call and a failed call ends the session's sweep for the pass

lczh:
## The closer caps its riders per call and a failed call ends the session's sweep for the pass

**The bug.** `_close_session` walks every end-known unclosed turn of a session and used to `continue` past a failed `_close_turn`; `run_close` awaits one such walk per session with no timeout. On 2026-09-03 one session's closer calls were alarm-killed 192 times in a row inside a single walk: 6h22m during which every judge for every session was silent. It recurred twice more the same day.

The kills were healthy but slow calls, not hangs. Served closer duration tracks output tokens (r≈0.99; input r≈0.07) at roughly 87 tok/s, so about 11.5k output tokens fits under `CALL_ALARM_S`; every served call with 10k+ output took 100 to 118 s, a distribution truncated at the alarm. The output came from the menu: `_close_turn` appended every rider with no cap, 24 riders on most of the backlog. It was self-sustaining because riders retire only through the `closerLookT` stamp a landed reply writes, so a killed call stamps nothing and the same menu rides the next turn.

**The fix.** `CLOSE_RIDER_CAP = 6` bounds the re-nominating riders per call (lifted, steps-finished, starved), never the turn's own menu and never the status riders (one-shot per status turn; a cut one would be lost, not deferred). Cut riders ride a later landed call, so the backlog drains one landed call at a time; this is a drain, not a fairness cap, and `CLOSE_FAIRNESS` stays `None`. `_close_session` ends the session's walk at the first failed call with a loud `sweep-cut` row carrying the menu's shape (counts only); a loop `break`, so the store still saves, and `_death_finalize` is told the session is not settled so a dead session's marker is never finalized off a walk that left turns unswept. The discriminator is `_judge_ctx.last_call_fail` being a dict, not `res is None`: parse rejects and pause-skips still walk on, and the safeguards tombstone keeps its arm. `_judge_run`'s failure exits leave the same traces on both engines (the stash, the model-health latch, and `model= chars= ms=` on the `call` row); the Codex branch previously wrote none of them.

**Unchanged on purpose.** `CALL_ALARM_S`, `CLOSE_FAIRNESS = None`, the safeguards tombstone. If a capped call still dies, the cost is one call per pass per such session, said loudly every pass; a degrade step for that case is a follow-up.

**Verification.** Full Python suite green (7034 passed, 0 failed on my head; 7048 on the maintainer's scratch merge with #904). Every behavioral test fails on the parent commit; the two boundary pins (a parse reject and a pause-skip still walk every turn) pass on the parent by design, since they fence the discriminator rather than exercise the fix. A four-lens adversarial review with two skeptics per finding confirmed six findings (status riders cut permanently, the Codex engine never engaging the cut, a cut walk finalizing a death marker, an overstated drain rate, a docs kinds list, a fixture leak), all fixed in the first commit.

**Review round.** The maintainer's commit on this branch gives the Codex served path the same recovery edge as the Claude one (`_mark_call_served`, cleared stash), lets never-looked riders take the room first so a top that keeps receiving filings cannot pin the same room-full every call, and rotates a cut dead session's marker to the back of the death drain. A docs bullet under "When a judge fails" now explains riders, the cap, and the cut. Left as a follow-up, as agreed: a give-up after repeated cuts on the same (turn, fingerprint), mirroring the safeguards tombstone.

### DIFF
--- a/docs/judge-pipeline.md
+++ b/docs/judge-pipeline.md
@@ -171,6 +171,6 @@ cards: one in each tree, linked, and the link checks itself off.
 | why is this card in this state? | the node's `log` in `goals/<sid>.json`: every event has source, kind, reason, time |
 | why was this work filed here? | `placements` in the store, plus the node's `why` and `trail` |
 | why did this mint at top level? | the node's `why`: "declared in the agent's own to-do list" means the plan-sync mirror — a step declared while serving a dispatch carries `serving` and folds into the sender's ask card at render (T137); the grouper may merge a duplicate; anything else is the planner |
-| did a judge fail or get skipped? | `judge-errors.jsonl`: every row carries judge, session, kind (parse, call, give-up, cite-miss, rate-limited, task-store, history-unreadable, drift-skip), and the evidence (reply tail, API message, re-arm event) |
+| did a judge fail or get skipped? | `judge-errors.jsonl`: every row carries judge, session, kind (parse, call, give-up, sweep-cut, cite-miss, rate-limited, task-store, history-unreadable, drift-skip), and the evidence (reply tail, API message, re-arm event) |
 | what did a judge call cost, and when? | `judge-usage.jsonl`, per judge and session |
 | what happened to a peer message? | `timeline/messages.jsonl` by message id, plus the delivered markers in the transcript |
--- a/docs/judges.md
+++ b/docs/judges.md
@@ -341,6 +341,24 @@ While the account usage window is exhausted, the rate gate skips every call
 across every session and logs one `rate-limited` row per window; gate skips count
 toward nothing.
 
+- **The closer bounds its own call and its own sweep.** Behind a turn's own
+  goals the closer carries *riders*: open work nominated from elsewhere (a goal
+  whose recorded steps are all finished, a starved leaf, a lifted card, and on
+  a status-report turn every open top). One session's closer calls were once
+  killed by the call timer 192 times in a row inside a single per-session
+  sweep, silencing every judge for six hours: the menu carried 24 riders, the
+  reply did not fit under the timer, and a killed call stamps nothing, so the
+  identical menu rode the next turn. Two bounds, neither a fairness cap.
+  `CLOSE_RIDER_CAP` limits the re-nominating riders per call (never the turn's
+  own goals, never the status riders, which are one-shot per status turn),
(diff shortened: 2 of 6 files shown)


henrypinkard:
Reviewed (own read plus a three-lens adversarial pass with two skeptics per finding). The cap and the cut do what the body says: the turn's own menu and the status riders are never cut, cut riders stay unstamped because the stamp loop covers the filtered menu, the walk `break`s so the store still saves, and `sent` is defined before every failure exit that prints the call shape. Full suite green here (7034 on your head, 7048 on a scratch merge with #904).

One commit pushed onto your branch (380abdab), three review finds each with a test that fails on your head:

- The codex failure exits now arm the model-health latch and the sweep-cut stash, but the codex served path never called `_mark_call_served` nor cleared `last_call_fail` (the claude branch does both at its served exit). Under `romp engine codex` one failed call left the model in `degraded` for the process's life, `stats[model]["fails"]` counted all-time rather than consecutive failures so three spread over any span read as "N in a row", and the `recovered` edge never fired. The served exit now leaves the same recovery edge on both engines.
- Under the cap the re-nominating riders were ordered by mint alone. A landed reply's own filing re-arms the riders an earlier call stamped (`_filed_since`), and those are older-minted, so while a top kept receiving filings the same room-full rode every call and riders past twice the room never rode once (the incident's 24 exceed 2x6). Unstamped riders take the room first now; channel and mint order are unchanged within each group. The test files one done under the top per landed call and requires all 20 riders to ride within ceil(20 / room) calls.
- A cut walk leaves a dead session's marker pending, and `_death_pending` takes the oldest marker first, so a turn whose call dies the same way every pass pinned the head of the drain queue, and `DEATH_DRAIN_PER_PASS` such sessions would starve every newer dead session of its sweep and its `ended` settle. `_death_rotate` touches the marker on a cut so it waits at the back, one call per pass behind everyone else; the `DEATH_DRAIN_PER_PASS` comment says so.

Left as the follow-up your body already names: a give-up after repeated cuts on the same (turn, fingerprint), mirroring the safeguards tombstone, so a live session whose one turn always dies gets its later turns swept again. Two nits: the docs name the `sweep-cut` kind but never explain the cut or the rider cap; and the body's red-first claim overstates for two boundary tests that pass on the parent.

PR #910, opened 2026-09-03
Repeated kills on one turn give it up loudly, and the session's later turns get swept again

lczh:
## Repeated kills on one turn give it up loudly, and the session's later turns get swept again

Follow-up to #905, as requested in its review: "a give-up after repeated cuts on the same (turn, fingerprint), mirroring the safeguards tombstone, so a live session whose one turn always dies gets its later turns swept again."

**The problem.** After #905 a failed closer call ends the session's sweep for the pass and the turn keeps the retry-next-pass contract. If the same turn's call dies the same way every pass, that live session pays one doomed call per pass forever, and its later end-known turns are never reached because the walk is cut at the head turn every time. A dead session at least rotates to the back of the death drain; a live session's later turns simply stay unswept.

**The fix.** A KILL is struck against the turn it died on at its current atom count, in `closeFails[tid]` as `{"fp", "fails", "kind": "kill"}` under `DISTILL_FAIL_CAP`, the way the safeguards arm strikes refusals. At the cap the turn is adopted exactly as the safeguards give-up adopts one (swept, `closedSig` at fp, no verdicts), one `give-up` row names the turn, the kill count and the re-arm event (the turn growing re-judges it through the existing `closedSig` growth check), and the walk continues to the next turn in the same pass. A landed call retires the record.

**Only kills strike, and a kill is the timer.** The call runner stamps `"kill": True` on its failure record only when the call ran to the timer: the perl alarm wrapper's SIGALRM landing on the child (`returncode == -signal.SIGALRM` at the two empty-output exits, on either engine) or a `TimeoutExpired` from the subprocess backstop. Everything else is the API or the process answering with an error in place of a reply, and says nothing about the prompt: an error envelope (a 529, an auth error, a 429), a clean exit of any code with empty output (exit 0 is a failed exec under the wrapper, exit 1 a startup crash or, on Codex, which has no envelope exit, a usage-limit or auth refusal), any other signal, any other exception. Those still cut the sweep for the pass with a loud `sweep-cut` row, leave no strike, keep the plain retry, and leave an existing kill streak untouched. Without that line a short 529 storm or a broken install would have adopted live turns without verdicts, which the safeguards tombstone, justified by determinism per prompt, never does. The streak rule: `closeFails` holds one record per turn, so the latest kind wins and nothing adds up; the count continues only for the same kind at the same size, and a strike of another kind (a kill after a refusal, a refusal after a kill, a parse reject after either) or at another size starts over at 1. A parse reject means the model answered, a kill means it never did.

**Unchanged.** `CALL_ALARM_S`, `CLOSE_FAIRNESS = None`, the rider cap, the dead-session rotation, the safeguards arm's own behavior.

**Accepted trade-off, per review.** A kill streak cannot tell a doomed prompt from an API slow patch: three slow passes would adopt every live session's head turn without verdicts, permanently. That cost is bounded (one turn per session per three passes) and loud (a give-up row for each). A later gate could strike only while the model is otherwise serving, using the model-health latch, which would separate the two.

**Review round.** The six nits are folded: the streak rule stated precisely with direct `_close_strike` unit tests, the Codex flag test's other-signal case, the docs' two exceptions and re-wrap, `_CALL_HEALTH` restored after the failure-contract tests, and the `DISTILL_FAIL_CAP` comment.

**Verification.** Full Python suite green (7093 passed, 0 failed on top of today's main). Tests in `tests/test_judge.py`: repeated kills give the turn up and the walk goes on to the next turn in the same pass, including one where the later turn lands with a real verdict in that same pass; a kill strike resets when the turn grows; kill and parse strikes do not add up; a landed call retires the record; a transient failure cuts the walk but never strikes; a transient cut leaves a kill streak untouched; the flag is stamped for the alarm signal and a timeout and for nothing else (exit 0, exit 1, another signal, on both engines). `tests/test_judge_close_riders.py` drives the cut with the real alarm shape. Two adversarial review passes with two skeptics per finding ran on this change: the first caught that every failure class was striking, the second that the kill was still keyed on empty output rather than on the timer; both are fixed here.

### DIFF
--- a/docs/judges.md
+++ b/docs/judges.md
@@ -321,7 +321,11 @@ event).
 - **An empty reply never counts.** `parse` means the model's own text was
   rejected, and the row carries the reply tail so the log says why. Empty
   replies (the rate gate, a failed call) log nothing at the caller and
-  never burn a retry cap.
+  never burn a retry cap — with two exceptions, both the closer's, both
+  adopting a turn loudly on their own event: a safeguards refusal of the
+  turn's content, and a *killed* call (one the timer ended; never an API
+  error or a process that ended any other way), each counted against the
+  turn it happened on. The kill streak is described below.
 - **Every judge is capped.** Three genuine parse rejects on the same work
   item (`JUDGE_FAIL_CAP`) and the judge gives up loudly, one `give-up` row
   naming the re-arm event, instead of retrying every pass forever:
@@ -331,7 +335,7 @@ event).
 | opener, live re-plan | hard-places at card level immediately | (no retries needed) |
 | planner (work run) | 3 tries, then hard-place a user message / drop a non-user segment | on its next segment |
 | placer | files at the card immediately | (no retries needed) |
-| closer | 3 tries, then skips the turn | when the turn gains atoms |
+| closer | 3 parse rejects, or 3 killed calls, then skips the turn | when the turn gains atoms |
 | archiver | 3 tries, then keeps serving the old headline | when the session gains a turn |
 | grouper, consolidator | 3 tries, then leaves the board shape as is | when the top set changes |
 | courier | 3 tries, then resolves from the sender's declared kind | (terminal; never orphans a message) |
@@ -356,8 +360,13 @@ toward nothing.
   subprocess error, an API error envelope, on either engine) ends that
   session's sweep for the pass with one `sweep-cut` row naming the turns left
   behind and the shape of the menu that died; parse rejects and pause-skips
-  still walk on. A dead session cut this way keeps its marker pending and
-  waits at the back of the death drain, so it cannot starve the others.
+  still walk on. Three *killed* calls on the same turn at the same size (the
+  timer ending the call; an API error, or a process that ended any way other
+  than the timer, cuts the walk too but leaves no strike) give that turn up
+  loudly (one `give-up` row; the turn growing re-arms it) and the walk moves
+  on, so a session whose one turn always dies still gets its later turns
+  swept. A dead session cut this way keeps its marker pending and waits at
+  the back of the death drain, so it cannot starve the others.
 
 ## Billing, and when the credential itself is broken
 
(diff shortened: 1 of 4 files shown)


henrypinkard:
Reviewed (own read plus a three-lens adversarial pass): no material findings. The kill flag is stamped at exactly the four timer sites on both engines and read as the flag alone; the wrapper's exit was verified by running the actual `perl -e 'alarm 1; exec @ARGV'` shape (returncode -14 through a direct exec, `sh -c` and `node`; a missing binary exits 0 with empty output, as your comment says). Streak kinds never add up, the give-up adopts like the safeguards tombstone and the walk continues in the same pass, `_death_finalize`/`_death_rotate` are unchanged for real cuts, and the latent `dict + 1` on a parse reject after a safeguards record is fixed in passing. Full suite green on your head merged with today's main (7105 passed).

Nits, none blocking, in case you want to fold them:
- `closeFails` holds ONE record per turn, so "kept as separate streaks" means "the latest kind wins, and nothing adds up" — conservative and fine, but the body overstates it; a direct unit test of `_close_strike` for kill-vs-safeguards and for a legacy `{fp, fails}` record (read as safeguards) would pin it.
- The codex flag test lacks the "another signal" case the claude test has (`-SIGKILL` → no flag); the code is symmetric.
- `docs/judges.md` says "with one exception" while the safeguards tombstone is a second, pre-existing one; and the reflowed paragraph has a stray 45-character line.
- The new FailureContract tests push `_CALL_HEALTH` past the cap for `model-x` and `gpt-5-test` without clearing.
- The `DISTILL_FAIL_CAP` definition comment still names only the distiller/briefer; it now also bounds the closer's kill streak.
- One design note rather than a defect, for the maintainer to accept consciously: a kill streak cannot tell a doomed prompt from an API slow patch (the 2026-07-27 shape in the `CALL_ALARM_S` note); three slow passes would adopt every live session's head turn without verdicts, permanently, since an end-known turn of a live session never grows. Bounded and loud, and the mechanism you were asked for; a later gate could strike only while the model is otherwise serving.

PR #924, opened 2026-09-04
PR #886 follow-ups: review nits + the jump chip moves bottom-left

yatharth:
The two nits from #886's merge note, plus one live-board ask:

- **Legend swatch matches its bar**: `var(--text-faint, #7d8590)` — the bar moved to the token in #886 and they split in classic.
- **Queued context refresh survives a failed attempt**: the None-payload early return sat before the rerun tail, so an ask queued behind a refresh that then errored was dropped; the rerun now fires on both exits (fail-then-work test).
- **The jump chip moves bottom-left, says "go to bottom"** (the user's word for it, in the styled tip), and its border reads through `--menu-border` — the raw white-alpha border made it a cream-on-cream ghost in the light theme, which is why the user never noticed it existed. Shape/size/behavior unchanged.

2782 webview tests + 7180 pytest green.

### DIFF
--- a/kernel/sdk_backend.py
+++ b/kernel/sdk_backend.py
@@ -2610,6 +2610,12 @@ class SdkSession:
         finally:
             self._ctx_refreshing = False
         if not isinstance(cu, dict):
+            # a queued ask survives THIS attempt's outcome: this early return used to sit before the
+            # rerun tail, so a failed/None payload dropped a switch-time ask on the floor and the old
+            # model's number stood until the next turn (PR #886 review) — same tail, both exits
+            if self._ctx_refresh_again:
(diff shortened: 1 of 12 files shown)


yatharth:
Two more from the live board, same branch: the file browser presents as the viewer's centered card over a dim (the user superseded their 2026-08-24 full-pane ruling), and the third hit of the `.md`-ties-specificity trap — inline/fenced code and bold inside the user's bubble fell onto page-tuned inks over the saturated fill (the reported red text in light). Fenced code now sits in a `var(--bg)` well where the hljs palette is correct in both themes. 2781 node / 7180 pytest green.

PR #933, opened 2026-09-05
The views rebuild on exact inputs, the watched tab is served while nothing changed, and the per-cycle store readers answer from a stat

lczh:
## What this fixes

Long-lived sessions grow transcripts without bound, and the kernel's push
cycle grew with them: on a seventeen-session board with an 80 MB transcript
the warm cycle took about 0.7 s with no dashboard open and 1.0 to 1.2 s with a
chat pane watching the largest session, most of it rebuilding views whose
inputs had not changed and re-reading small stores whose files had not moved.

Upstream has since folded the per-push transcript and states readers and the
chat payload build (97c0159e, c84165d4, 8925efad), which covered the first
part of this work; the readers here defer to those folds. What remains is the
other half, the decision of WHEN to rebuild at all:

- The judge generation advances only when a producer pass changed a
  judge-written store (goals, captions, archive, episodes), keyed on a
  fingerprint taken before and after the pass, instead of on every pass. A
  quiet pass leaves every view's signature alone.

- The board-view signature carries every input its rulings read: the judge
  run counter, the SDK snapshot facts (subagents, background tasks,
  interrupting, model and effort pending, auth pending, retry count,
  connected, spawning), and the auto-nudge, nudge-event, judge-auth and
  judge-limit stores. The time bucket stays, documented: removing it needs
  client-side age tints and a due-time heap for the build-time caps, Stage 1
  work.

- The ACTIVE chat tab, rebuilt on every cycle by design so the watched session
  stays live, is served from its last build while an exact key is unchanged:
  the chat signature, the snapshot digest minus its clock, the backend's live
  revision (bumped at seven mutation sites), its queue, compacting and
  clearing flags, fifteen side files, the names hash, the clock predicates
  resolved to the booleans they decide, and the external inputs the build
  reads (watches, PR watches, usage, the CLAUDE.md chain, both git trees'
  HEADs, running tasks' output files, the hour mark).

- Per-cycle readers of small stores answer from a stat while their files are
  unchanged: captions fold appends, the thread registry and names snapshot
  and task store memoize per entry on (mtime, size, inode), the backend's
  ownership check memoizes on the registry's stat, and the awaiting-lift gate
  skips a session whose inputs (goals, overrides, transcript, messages, SDK
  registry stats, live background-task ids) are unchanged. Two mechanisms are
  deliberately not gated: the awaiting-wake outcomes and the nudge walk carry
  a dead-man clock in their rulings.

- The version route counts how often each view was rebuilt versus served, so
  the next measurement does not need a replay harness.

Measured on an offline replay of the real board, seventeen sessions with a
41 MB transcript watched in the chat pane, against current upstream main with
its own folds in place: the warm push cycle 0.49 to 0.87 s on main, 0.27 to
0.31 s on this branch once the active tab is served from its last build (on
the older base the same change was 1.0 to 1.2 s down to 0.30 s; upstream's
folds took the rest). The cold first cycle, about 15 s of boot parse on both,
is Stage 2's subject; the timeline's O(all atoms) rebuild is Stage 1's.

Two adversarial review passes reproduced seven defects and two residuals in
the first cut, all fixed: the lift gate omitted its ruling's clock arm and
subagents and recorded itself before ruling; the clock-predicate expiry arm
was dead; the judge generation missed between-pass store writes; the active
key missed six side files and the CLAUDE.md chain; stat keys lacked the inode;
the task-store reader returned None on a mid-scan unlink; the tests' open
counter missed pathlib reads.

A third review, of the squash onto current main, found that the served active
tab keyed only the SDK backend's live tail: a tmux session's composer echo and
a Codex session's queue live in their own stores, which the build renders but
no file records, so the watched tmux tab served its last build with the send
missing until some file moved. The key now takes the OWNING backend's tail,
digested from exactly what the build reads. It also found that background
tabs had lost their implicit three-second refresh: the judge generation no
longer advances every pass, and nothing else keyed the snapshot facts a
background tab's chips render, so the background key now folds the same
per-session snapshot digest the board key uses. Smaller: the spend hold file
and the env file whose key line decides the Billing selector join the active
key, every writer of the spend hold marks the views dirty, the pre-pass judge
fingerprint is computed only on the first pass where it is consulted, and the
active key reuses the pusher's already-computed base instead of a second stat
pass.

Tests: tests/test_stage0_log_readers.py, with a private state sandbox per
class so the shared kernel singletons cannot leak between modules; the
redelivery stub models the live-revision hook.

Branch: one commit on current main, rebased today; three adversarial review rounds folded; full Python suite 7369 passed.

### DIFF
--- a/kernel/event_model.py
+++ b/kernel/event_model.py
@@ -494,7 +494,9 @@ def _read_jsonl(path):
 # whole-parse cache contract. The cached list itself is never extended in place — a grown file stores a
 # NEW list — so a concurrent reader holding the old list is never surprised mid-iteration.
 _JSONL_CACHE = {}                 # path -> (mtime, size, offset, tail_bytes, records); dict order = LRU, hits reinsert
-_JSONL_CACHE_MAX = 384            # bounds MEMORY only — past the cap, evict the least-recently-USED entry, one per
+_JSONL_CACHE_MAX = 1024           # bounds MEMORY only (384 → 1024 on 2026-09-03: the per-session states,
+                                  # captions and the postal/nudge logs became tenants — a few KB each — and
+                                  # must never evict a live transcript's slot) — past the cap, evict the least-recently-USED entry, one per
                                   # insert, never clear(). The old clear-at-cap was sized to the session count, but
                                   # the working set is FILES, not sessions (every subagent writes its own transcript):
                                   # once more distinct files than slots passed through one push cycle, the clear
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -850,12 +850,21 @@ def pass_watermark(tier, fsid):
     return _PASS_DONE.get((str(tier), str(fsid)))
 _active_lock = threading.Lock()
 _active_seq = [0]
+_active_change = [0]      # bumped on EVERY begin and end — the exact "the set of in-flight judge calls
+#                            changed" event the kernel's view signature keys on (2026-09-03), so a
+#                            judging swirl lights and clears on the call itself, not on a pass cadence
+
+
+def active_change():
+    """Monotonic count of changes to the in-flight judge-call set (begins + ends)."""
+    return _active_change[0]
 
 
 def _active_begin(judge, fsid, sent):
     """Mark a judge call as running; returns a run id to pass to _active_end on completion."""
     with _active_lock:
         _active_seq[0] += 1
+        _active_change[0] += 1
         rid = _active_seq[0]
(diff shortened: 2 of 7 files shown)


lczh:
Tier: `fix` — a fork author cannot add labels here, so asking a maintainer to apply it. The change removes pathological per-cycle work (views rebuilt on unchanged inputs) with tests that fail before it; no new capability.

PR #934, opened 2026-09-05
A chat tab click no longer waits behind the other panes' work

lczh:
## What this fixes

On a board of seventeen sessions every click on a chat tab lagged noticeably. The four dashboard panes are same-origin iframes, so they share one browser main thread, and a headless-browser replay of the real board's payloads measured what the click waited behind: median input-to-paint latency of 160 to 224 ms, worst 560 ms, against about 60 ms for the chat pane alone. After these two commits, on the same replay: median 64 to 72 ms, worst about 100 ms; main-thread busy time from the other panes 75 percent down to 14.

## Commit 1

On a board of seventeen sessions every click on a chat tab lagged noticeably.
The four panes are same-origin iframes, so they share one browser main thread,
and a headless browser replaying the real board's payloads put a number on what
the click waited behind: median input-to-paint latency of 160 to 224 ms with a
worst case of 560 ms, against about 60 ms for the chat pane on its own, with the
main thread busy three quarters of the time on work nobody was looking at.

The work, and what each change does about it:

- The timeline skeleton frame carries its clock nested under `data`, where the
  per-client dedup never looked, so an unchanged skeleton crossed on every
  pusher cycle and the timeline rebuilt its whole SVG each time. The signature
  strips the nested clock for that frame type.

- The timeline's live-follow loop woke on every animation frame, read layout to
  check its visibility, and rebuilt the SVG once the edge had crept 0.15 px,
  about twice a second at a one-hour window. It now sleeps exactly as long as
  the edge needs to move one pixel at the current zoom, bounded to 0.1 to 2 s,
  and touches nothing in between. Inside each rebuild the prompt-dot pass
  compared every turn against every message; processed messages are indexed by
  recipient once per draw instead.

- The feed rewrote all cards on every frame, and its card-flight animation read
  each card's rectangle between transform writes, forcing a layout of the whole
  document per card, 155 times per frame. The flight now reads every rectangle
  before writing any transform, runs at all only when a card changed column,
  arrived or left, and a card whose data and display state did not change is
  not repainted.

- The outline pane rebuilt its list on every feed frame while hidden in a
  display:none iframe. It defers until it is on screen.

- In the chat pane a background tab that received any tail was marked stale,
  so the click rebuilt its whole eighty-unit window; it is now repainted from
  the changed point like the active tab. Inbound tails repaint the tab strip
  once per animation frame instead of once per tail, the idle pre-build checks
  its budget before each tab, and its fallback deadline counts down instead of
  reporting twelve milliseconds forever.

Coalescing the skeleton and bars frames into one client-side draw was tried and
reverted: the timeline's tests rely on synchronous draws, and the kernel-side
dedup makes it unnecessary.

Measured on the same replay, four panes under whole-payload traffic: median
click latency 64 to 72 ms, ninetieth percentile 72 to 96, worst 88 to 104; the
main thread busy 13 to 15 percent instead of 75.

Tests: the skeleton dedups across a clock step and still crosses on a real
change; the live loop sleeps by pixel rate and the index answers the one-second
question exactly against a brute-force scan; the flip gate on pure inputs plus
pins that the read pass precedes the write pass and that the paint key covers
the display state; the outline gate; the chat's incremental branch, the
coalesced strip, and the idle budget order. Three existing pins updated to the
new mechanics.

## Commit 2 (review round)

A three-lens review of the previous commit reproduced what its speedups had
cost, each in a headless browser against the real board, and this closes them.

- With the skeleton frame deduped, a quiet board sends the timeline no clock
  for up to the kernel's sixty-second repost. The live edge's free-running
  clock was capped thirty seconds past its last anchor, so it glided, froze
  for half a minute, then jumped forward when the repost landed. The cap now
  outlasts the repost, and a bars frame re-anchors the edge the way the
  skeleton does, so whichever frame arrives keeps the edge honest.

- The paced live loop died when the pane was hidden and nothing restarted it
  on reveal, where the old per-frame loop had come back within a cycle. It now
  keeps a long sleep while hidden and resumes on its own. A zoom mid-sleep kept
  the sleep computed for the old zoom; a restart now drops a pending sleep and
  looks at once. Under a held pointer it no longer polls every hundred
  milliseconds: the release event, which already flushes the buffered redraw,
  restarts it.

- The feed's paint key covered the item and a few focus states but not the
  view preferences, the working, awaiting and unknown status sets the session
  dots and delegation lines read, the self host, or the clock behind the
  durations a card renders, so a grouped-mode toggle or a peer changing state
  left cards stale until their own data changed; and a button latched into its
  in-flight label relied on the next paint to re-enable when the kernel refused
  the action without changing the card. The key now carries an epoch bumped by
  those inputs and a fifteen-second clock, and a card with a disabled button
  always repaints.

- The flip gate compared columns only, so a card that moved within its column
  snapped instead of gliding, against the rule that in-column shifters glide.
  The gate now sees position too.

- The outline pane deferred even its first content while hidden, so revealing
  it showed the pane loader fading over nothing. The first content paints
  through; only later rebuilds wait. Its new identifiers no longer use the word
  the vocabulary rule reserves.

- A background chat tab whose window the user had scrolled away from the tail
  could receive an in-window change the incremental path would not repaint; it
  falls back to the window rebuild in that case.

Tests extend the previous commit's: the cap outlasts the repost and the pure
functions glide through a sixty-one second silence without a stall or a jump,
the bars frame anchors, the hidden and held-pointer paths, the re-pacing
restart, the widened paint key and the latched-button rule, the in-column
reorder case, the first paint through, and the scrolled-away window rule.

Rebased onto current main today (no conflicts). Full Python suite 7337 passed, node 2799 passed. Independent of #914's kernel changes; merges clean against the Stage 0 branch as well.


lczh:
Tier: `fix` — a fork author cannot add labels here, so asking a maintainer to apply it. A lag bug (tab clicks waiting behind other panes' work), with tests that fail before it.

PR #935, opened 2026-09-05
Open any subagent's whole transcript from the dashboard, live while it runs

yatharth:
## What

A Claude Code subagent (the `Agent` tool; older transcripts say `Task`) used to show two points of its life on the parent's tool head: the kickoff prompt and the final report. Everything between was invisible, a running background agent had nothing to say for minutes, and a background agent's report fold showed the launch acknowledgement forever. Now:

- **Level 0, on the Agent head** (and on agent rows of the background-tasks box): an open-transcript arrow whenever the launch names an agent; while the agent runs, a preview of its last three tool calls in the head's own `<tool> <desc>` vocabulary, the last row trailing `· N tool calls · elapsed`. Inside the tool turn, so a collapsed compact run hides it. Gone when the agent finishes.
- **Level 1, the viewer:** the arrow opens the agent's WHOLE conversation as a peek tab wearing the parent's colour, rendered through the chat's own renderers (Compact transcript, folds, day dividers), read-only. Header: "subagent of <parent> · <type> · running|finished", the parent name a link back to the launch, a pin that keeps the tab, an "earlier part not shown" note when the tail was cut. Live frames update it in place with the chat's follow-only-at-bottom rule; the romp loader holds the pane until the first frame; a missing file shows its error sentence in the pane.
- **The report fold is now the real report** for a background agent: once its `<task-notification>` lands, the notification's `<result>` fills the head's output (and `isError` follows the status) instead of the launch ack. The notice card in the transcript is unchanged.

Plan doc: `plans/subagent-transcripts.md` (problem, verified data facts, decisions, wire contract, caps, and the slice-2 scope — the Awaiting box grouped by kind — which is NOT in this PR).

## Why read the file rather than un-drop the stream

The SDK backend deliberately drops live stream messages tagged `parent_tool_use_id` (`tests/test_sidechain_atoms.py` pins it) — that stays. Claude Code writes every subagent its own complete JSONL beside the parent transcript (`<proj>/<fsid>/subagents/agent-<agentId>.jsonl` + a `.meta.json` sidecar whose `toolUseId` is the parent's tool_use block id; the background task's `/tmp` output file is a symlink to it). Reading that file keeps the parent stream clean, works for dormant/revived and tmux sessions, and reads the authoritative store. Liveness is event-based: the SDK's SubagentStart/Stop set (now shipped with agent ids) and its task-lifecycle set for SDK sessions, the CLI-epoch ghost gate otherwise — the same gate the bg-tasks box already applies, so head, preview, viewer and box agree.

## Wire contract

- Every `kind:"tool"` event carries `toolUseId` (the tool_use BLOCK id; `uuid` was and is the record's). Agent/Task events add `agentId` (or `null`), `agentAsync` (a background launch); while a background agent runs: `agentRunning: true`, `output: ""`, `agentGist: {recent: [{tool, desc, ts}] (≤3, newest last), calls, since, last}`; once its notification lands: `output` = the `<result>` text, `isError` from the status, no gist.
- bg-tasks rows: `agentId` on agent rows only.
- Client → kernel `{type:"openSubagent", id:<sid>, agentId}` / `{type:"closeSubagent", id, agentId}`. Kernel → client `{type:"subagent", id, agentId, meta:{agentType, description, spawnDepth, toolUseId}, running, events:[…chat events…], truncated}` or `{…, error:"<sentence>"}`. First frame on open, re-pushed by the pusher's chat pass only when the agent file's (mtime,size) or the running flag moved; stops on close/disconnect. Federation: routed by the parent `id` like a comment-thread request; the client's tab id is `<parentId>/agent/<agentId>` (colon-free suffix, so `hostOf()` keeps working).
- The chat fold never seals a running agent's launch turn (the undecided-seam rule's twin); a sealed pending agent demotes when its report lands or its sidecar appears. `TUR_CONSUMED_KEYS` widened with `agentId`/`isAsync` (event model + SDK mirror, in step).

## Verified

- `uv run … pytest tests/` — 7357 passed, 40 skipped (new: `tests/test_subagent_transcripts.py`, 24 tests: sidecar/ack join, tool-event stamps, gist shape, landed report + failed status, liveness from live sets and the ghost gate, viewer build/cap/error/forked-dir lookup, open/close/re-push, bg rows, fold hold + equivalence).
- `cd vscode-extension && npm run typecheck && npm test && npm run build` — 2796 passed (new: `ui/webview/subagent-view.test.ts`, 15 tests; two `peek-tab.test.ts` pins moved deliberately with a comment saying why; `render-agent.test.ts` untouched and green; `tests/test_sidechain_atoms.py` untouched and green).
- Headless screenshots (both themes, synthetic notes-api fixtures, not committed): the Agent head with the running preview + arrow beside a finished Agent head with `prompt + report`; the viewer as a peek tab with header, transcript and the single dim read-only footer line.

### DIFF
--- a/kernel/event_model.py
+++ b/kernel/event_model.py
@@ -42,11 +42,12 @@ MESSAGES_LOG = STATE / "timeline" / "messages.jsonl"
 # Mid-turn the model stops with `tool_use` (a tool cycle) — that does NOT end the turn.
 END_STOPS = ("end_turn", "stop_sequence")
 # The toolUseResult keys a consumer actually reads (Edit's structuredPatch → diffRows,
-# AskUserQuestion's answers map → the answered box). The atom carry is GATED on one of these
-# being present: an unconditional dict carry held every Read result's full file bytes in the
-# parse cache by reference — ~a fifth of transcript bytes on read-heavy sessions — for shapes
-# nothing reads. Widen this set when a new consumer appears; never revert to carry-all.
-TUR_CONSUMED_KEYS = frozenset(("answers", "structuredPatch"))
+# AskUserQuestion's answers map → the answered box, the Agent tool's agentId / isAsync → the
+# subagent join + the background-launch flag, plans/subagent-transcripts.md). The atom carry is
+# GATED on one of these being present: an unconditional dict carry held every Read result's full
+# file bytes in the parse cache by reference — ~a fifth of transcript bytes on read-heavy sessions —
+# for shapes nothing reads. Widen this set when a new consumer appears; never revert to carry-all.
+TUR_CONSUMED_KEYS = frozenset(("answers", "structuredPatch", "agentId", "isAsync"))
 # romp's own postal marker, injected into a delivered message body. It is the ONLY
 # postal signal — never the generic "Stop hook feedback:" prefix (any blocking Stop
 # hook produces that). The sender rompUuid is resolved from timeline/messages.jsonl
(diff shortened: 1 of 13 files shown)


yatharth:
Second cut on the Agent head (140a0f8a): the fold now carries everything in ONE click. Label reads `prompt · 12 tool calls` while running and `prompt · 12 tool calls · report · 3 lines` once finished; inside — all open, nothing nested — the prompt as markdown, every tool call the agent has made (one dim row each, newest last, live while running, `N earlier calls not shown` past the 200 cap), then the report. The three-row preview under a running head stays but only while the fold is closed. Kernel side: `agentSteps` + `stepsTotal` ship on every Agent event, running or finished, from the same per-file fold cache; `agentGist` keeps `calls/since/last` and drops `recent`. The open-transcript arrow stays on a finished head (pinned). Fixture `tools/ui-verify/fixtures/agent-head.html` renders the three states.

PR #937, opened 2026-09-05
The bell opens a popover with separate switches and a test button, and a turn end can buzz the phone

yatharth:
Stacked on #940 (notify-click-open): merged in so the two push PRs no longer collide; land #940 first.

## The problem

One tap on the notification bell did two jobs at once: it flipped the kernel-wide master switch and it subscribed or unsubscribed this device's Web Push. So turning the bell off on a phone silenced every device; a denied permission threw a line into the Log while leaving the master on, with no way back; nothing ever told the user whether the push service had accepted a subscription; and a turn finishing never buzzed unless the judges happened to move its card.

## What is now true

- Tapping either bell (desktop rail or mobile tab bar) opens a small anchored popover in the house menu dress. The shell now defines the menu tokens in its own `:root` / `body.theme-light` blocks (it loads no sheet), and the popover's rules read them with the dark literals as `var()` fallbacks. Escape and an outside tap close it like every other shell panel.
- Rows: **All devices** (the master, `/notify-all`), **This device** (this browser's subscription, permission still requested inside the tap's own stack; when the site is blocked the row is disabled and says how to re-enable; where the Push API is missing it says so), **Also when a turn finishes** (new, `/notify-turns`, off by default), and **Send a test notification** (acknowledges at once, self-restores, shows the push service's answer as one sentence under itself).
- The bell glyph reflects THIS device: lit only when the master is on AND this browser is subscribed; slashed otherwise. A browser without the Push API has no subscription half to reflect, so the master alone paints it there. The tooltip names which half is off.
- `"*turns"` is a second reserved key beside the master in `notify-cards.json` (same fire path, same cache, same `__ncards__` watch; prune keeps both reserved keys).

## `/push/test` contract

`POST /push/test {endpoint}` sends ONE notification (title `romp`, one plain sentence) to that subscription only and returns the push service's answer as `{ok, status, detail}`: `ok` is a 2xx; `status` is the HTTP status, or `0` when the request never got an answer (with the error in `detail`); a refusal carries the service's own reason in `detail`. An endpoint nobody subscribed answers `ok:false` with `detail` "this device isn't subscribed yet". A dead subscription (404/410) is pruned exactly as the fan-out prunes it, and `detail` says so. Missing crypto is the same loud 500 `/push/subscribe` gives. Underneath, `_push_post` is the one HTTP path returning `(status, detail)`; `_push_send_one` keeps its prune semantics (False on 404/410 only, everything else keeps the subscription) on top of it.

## The turn-complete rule

`_turn_notify_tick` runs in the pusher cycle right after the feed build. The event is the session's turn-end key moving: the Stop hook's `lastStopAt` stamp (SDK sessions), else a STOPPED `states/` transition (`waiting`/`idle`, so a turn starting never reads as an end). It fires only with the master AND the turn switch on and the session unmuted, pushing `{title: <session name>, body: first line of the reply clipped to 120 chars, sid}` through `_push_notify` and forwarding to trusted peers the bell-event way. The first sighting of a session is a silent baseline, and the memo advances while the switch is off, so turning it on later never replays old ends.

## The no-double-buzz rule

One phone buzz per session per turn end. The bell event (a card entering needs_input/completed) and the turn-finished push both claim `(sid, turn-end key)` through `_buzz_claim`; whichever files first buzzes and the other yields. Within one cycle the feed builds first, so when the judges have already ruled the more informative bell event wins; across cycles the turn push usually lands first and the later bell event yields on the phone. Bell events never suppress each other (two cards of one session moving in one build still buzz twice), the desktop notification and the badge are outside the rule, and only what buzzed locally is forwarded to peers, so their phones hear each turn end once too.

## Verification

- Python: full suite green, 7369 passed / 40 skipped (run in two halves under the tool's time cap). New `tests/test_kernel_notify_popover.py` covers the store and `/notify-turns`, `/push/test` on the real handler with each outcome, `_push_post` against a loopback push-service double, the prune rule, the tick (baseline, both switches, mute, fallback key, wiring order), the claim in both orders, the relay, and the shell markup/tokens. Three existing pins updated deliberately (federation forward, notify-bells forward, RailBell model), each with a comment.
- TypeScript: `npm run typecheck`, `npm test` (2782 passed; `menu-theme-tokens.test.ts` gains the popover as a surface plus a shell-defines-tokens test), `npm run build`.
- Screenshots (headless, synthetic fixture): desktop dark and light show the card anchored above the rail bell with two lit pills, one off, and the accepted sentence under the button; phone-width dark and light show it above the tab bar with the This-device row disabled and its blocked-site sub-line, and a red refusal line under the button. Tokens resolve to the dark card and the cream card respectively; text and pills line up. The first desktop render caught the popover CSS sitting inside the mobile media block, which is now fixed (the rules are global).

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -418,6 +418,39 @@ internet and your agents, with no device check in front of it.
     standing between other members and your agents. Either keep the tailnet to
     your own devices, or write an ACL restricting the kernel machine to them.
 
+#### Notifications on your phone
+
+Romp can buzz your phone when a session needs you or finishes a task, so you can
+put the phone down while the sessions work. On an iPhone, first add Romp to the
+Home Screen (share sheet, then **Add to Home Screen**) and open it from there:
+iOS only lets an installed app receive notifications, so in a plain Safari tab
+the option stays off and says so. On Android and on a desktop browser the page
+itself can receive them.
+
+Then tap the bell. On a phone it sits in the bar along the bottom; on a desktop
+it is in the bottom-right cluster. A small card opens with three switches and a
+button:
+
+- **All devices** is the main switch. It turns notifications on for every task,
+  on every device you have set up and on the desktop of every machine you have
+  attached. The bells on individual sessions and cards then work as mutes.
+- **This device** turns them on for the phone or browser you are holding. The
+  first time, the browser asks for permission. If you refuse, the row goes grey
+  and tells you where to allow it again (on an iPhone, Settings, then
+  Notifications, then Romp; in a desktop browser, the site permission beside the
+  address). Turning this off silences only this device.
+- **Also when a turn finishes** adds a notification every time any session
+  finishes a turn, with the session's name and the first line of what it said.
+  With many sessions running this is a lot of buzzing, so it is off unless you
+  want it. A turn that ends by asking you something buzzes once, not twice.
+- **Send a test notification** sends one notification to the device you are
+  holding and prints the push service's answer under the button, so you can see
+  at once whether the phone is set up or why it is not.
+
+The bell itself shows the state of the device you are looking at: lit when the
+main switch is on and this device is set up, and crossed out otherwise. Its
+tooltip says which of the two is off.
+
 ## Security and trust
 
 Romp drives agents that run tools and shell commands as you, so reaching its API
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -428,24 +428,31 @@ the option stays off and says so. On Android and on a desktop browser the page
 itself can receive them.
 
 Then tap the bell. On a phone it sits in the bar along the bottom; on a desktop
-it is in the bottom-right cluster. A small card opens with three switches and a
-button:
+it is in the bottom-right cluster. A small card opens with a main switch, two
+switches indented under it, and a button:
 
-- **All devices** is the main switch. It turns notifications on for every task,
-  on every device you have set up and on the desktop of every machine you have
-  attached. The bells on individual sessions and cards then work as mutes.
-- **This device** turns them on for the phone or browser you are holding. The
-  first time, the browser asks for permission. If you refuse, the row goes grey
-  and tells you where to allow it again (on an iPhone, Settings, then
-  Notifications, then Romp; in a desktop browser, the site permission beside the
-  address). Turning this off silences only this device.
-- **Also when a turn finishes** adds a notification every time any session
-  finishes a turn, with the session's name and the first line of what it said.
-  With many sessions running this is a lot of buzzing, so it is off unless you
-  want it. A turn that ends by asking you something buzzes once, not twice.
+- **Notifications** is the main switch. Off silences every device and the
+  desktop of every machine you have attached; the bells on individual sessions
+  and cards are mutes under it. While it is off, the two switches beneath it are
+  dimmed but still work, so you can set a phone up first and switch everything
+  on when you are ready.
+- **This device**, under it, turns them on for the phone or browser you are
+  holding. The first time, the browser asks for permission. If you refuse, the
+  row goes grey and tells you where to allow it again (on an iPhone, Settings,
+  then Notifications, then Romp; in a desktop browser, the site permission
+  beside the address). Turning this off silences only this device. With the
+  main switch off, the row says the device is set up but nothing arrives until
+  the main switch is on.
+- **Also when a turn finishes**, also under it, adds a notification every time
+  any session finishes a turn, with the session's name and the first line of
+  what it said. With many sessions running this is a lot of buzzing, so it is
+  off unless you want it. A turn that ends by asking you something buzzes once,
+  not twice.
 - **Send a test notification** sends one notification to the device you are
-  holding and prints the push service's answer under the button, so you can see
-  at once whether the phone is set up or why it is not.
+  holding, whatever the switches say, and prints the push service's answer under
+  the button, so you can see at once whether the phone is set up or why it is
+  not. With the main switch off, the answer adds that real notifications will
+  not arrive until it is on.
 
 The bell itself shows the state of the device you are looking at: lit when the
 main switch is on and this device is set up, and crossed out otherwise. Its
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -451,8 +451,11 @@ switches indented under it, and a button:
 - **Send a test notification** sends one notification to the device you are
   holding, whatever the switches say, and prints the push service's answer under
   the button, so you can see at once whether the phone is set up or why it is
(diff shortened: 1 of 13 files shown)


yatharth:
Follow-up on the popover's hierarchy, from the feedback that "All devices" beside "This device" read as a scope choice rather than the main switch it is — and that with the master off, the device row still painted ON (it reads only this browser's subscription).

What changed:
- The top row is now **Notifications**, with a two-sentence sub-line: it is the main switch, off silences every device and every attached machine's desktop, and the session/card bells are mutes under it.
- **This device** and **Also when a turn finishes** are nested under it (indent + the popover's own `--menu-border` hairline). While the master is off both rows dim in the existing `.off`/`.busy` wash, toggled by `paint()` from the same kernel bit that paints the master's pill — so it follows the boot fetch and every `notifyAll` shell push, no polling. They stay operable so a phone can be set up first; the device sub-line then says it is set up but nothing arrives until the main switch is on.
- The test button still ignores the switches on purpose; with the master off its result adds one sentence saying real notifications won't arrive until the main switch is on.
- Guide section updated to the new labels and nesting; ShellPopover/RailBell pins moved deliberately; new tests cover the nesting markup, the master-off class + CSS rule, the operable-while-dimmed contract, the device sub-line branch, and the appended test-result sentence. Menu tokens unchanged (the token sweeps cover the new rules).

Verified: 154 Python tests across the notification files, TS typecheck + 2782 tests + build green, and headless shots of the on and master-off states in both themes.

yatharth:
Heads-up for whichever of #937 and #940 lands second (I own both branches and will do the reconciliation then; verified on a local merge of the two, all push test files green):

- `_turn_notify_tick`: the push call becomes `_push_notify(title, body, sid, kind="turn")` and the forwarded row gains `"kind": "turn"`, so a peer's tap lands the same way.
- `_push_test`: the payload goes through `_push_payload("romp", "Test notification — this device is set up.", kind="test")` so the tap on a test notification just brings romp forward.
- `_cached_feed`: keep #937's `_buzz_claim` gate and `_buzzed` list, with #940's four-tuple (`_iid`) and `kind="card", card_id=_iid` on the push, and the forwarded row carrying `kind`/`cardId`.
- Pins in `tests/test_kernel_notify_popover.py`: the test payload now carries `tag` + `data` (kind test, no sid); the feed-path index looks for the call with its kind arguments; the relay assertion accepts `host` as a kwarg; the fired turn row includes `"kind": "turn"`.

yatharth:
**Restacked on #940 + the test notification now round-trips to a session** (tip 91d7d81a)

- **Merge with #940** (870d56f3): `_cached_feed` keeps #937's `_buzzed` list and the `_buzz_claim(_sid, _turn_end_key(_sid), "bell")` gate, iterates #940's four-tuple `(_t, _b, _sid, _iid)`, pushes `kind="card", card_id=_iid`, and forwards only `_buzzed`. `_turn_notify_tick` pushes and forwards `kind="turn"`; `_push_test` builds its payload through `_push_payload(..., kind="test")`. Pins in the federation and popover tests follow the same shapes as the maintainer's local integration preview.
- **The test notification is addressed to the session you were looking at** (91d7d81a). Press "Send a test notification" on one session, switch to another session and another browser tab, tap the notification, and you land back on the first.
  - Shell: the test handler reads the chat pane's active tab at the press — `#tabs .tab.active[data-id]` off the same-origin chat iframe, the DOM the mobile `#mcur` chip already mirrors, so no new channel — and POSTs `/push/test {endpoint, sid, host}` (host = the prefix of a `host:sid` federated id). No tab in front → no sid, the old probe.
  - Kernel: `_push_test(endpoint, sid="", host="")` → `_push_payload("romp", "Test notification — tap to come back to <name>.", sid, kind="test", host)`; the name is the local registry's, a federated session falls back to its short id (its name lives at the origin kernel). The answer echoes `sid` + `name`; a non-string sid/host is a 400. Without a sid, body and payload are unchanged.
  - Tap routing needed no code: the SW passes the routing block through without branching on kind and the shell's `land()` POSTs `/reveal` for any sid (card scroll only for `kind: "card"`) — now pinned for `kind: "test"` in both node-executed harnesses.
  - Popover result line adds "Tapping it brings you back to <name>." on success with a session; the guide's test-button bullet says the test is addressed to the session you were looking at.
- Tests: targeted push set 176 passed; extension typecheck/build clean, 2785/2785; full Python suite 7395 passed, 40 skipped.

Not armed for auto-merge: land #940 first.

yatharth:
Two fixes from the first real-device try (test button pressed on a remote session, notification tapped) — tip `cf0a55fd`.

**1. The tap did nothing on the phone.** Root cause in the service worker: `clients.matchAll({type:'window'})` lists the dashboard's same-origin pane iframes as window clients too (`frameType: 'nested'`), most recently *focused* first. After switching sessions in the chat pane's mobile picker, the chat iframe was `ws[0]`, and only the top-level shell listens for the worker's message — so the routing block was posted into a frame that never heard it. The worker now hands the tap to the most recently focused **top-level** client, and opens the deep link only when there is none.

Two more holes the phone makes likely (sockets that die without a close, which the ping timeout has up to 30 s to notice), fixed at the kernel's reveal aiming: the deep-link arrival POSTs `boot:true` so the kernel parks for the pane about to connect instead of "delivering" into the previous page's dead same-`wid` socket; and a live tap to a socket with a ping outstanding is delivered *and* kept parked until that socket's pong (or a redial's `ready`) — never replayed once proven. In the chat pane, a focus now retires the persisted-tab restore (which used to jump back over the reveal when the remembered tab was a remote session arriving later), and a federated id whose host has not relayed yet shows the romp loader rather than the no-sessions copy.

**2. A remote session's test notification named its short id.** The tunnel supervisor already polls each attached host's `/sessions` (id + name); the poll now files names beside sids, and `/push/test` names a federated session from that snapshot, host-prefixed as the dashboard shows it. If the kernel has no name, the shell's tab label stands in (a string, flattened, clipped to 80 chars — else 400); the short id is last. The popover's result line echoes the same name.

**Real-device check:** on the phone, open a session hosted on another machine, press *Send a test notification* (the result line should say it brings you back to `host:name`, not a short id), switch to another session, then tap the notification: the chat pane opens on that session, its chip names it, and the notification body carries the same name. Also worth one try after leaving the app in the background for a minute before tapping.

Tests: worker harness with nested pane clients; reveal script posts `boot` on the deep link only; `_reveal_request` parks past a stale twin on boot and keeps/retires the unproven copy; `/reveal` passes `boot`; `/push/test` name from snapshot → label → short id, non-string label → 400; poll rows + supervisor filing; render.ts pins. Full Python suite 7405 passed / 40 skipped; TS 2787 passed; bundles build. Auto-merge not armed.

PR #938, opened 2026-09-05
The Awaiting box lists what a session waits on as rows grouped by kind, and the chip opens it

yatharth:
**Stacked on #935 (slice 1); the first four commits are that PR's and this diff should be read after it merges.** Slice 2 of `plans/subagent-transcripts.md` — the five commits from `8336e6a2` on, plus one fixture-note fix.

## What and why

A session can wait on three different things — background agents, background commands (`run_in_background` Bash / Monitor), and armed kernel watches — and the dashboard used to fold them into one word chosen by which source spoke first. A background agent plus a background shell command read "Awaiting 2 tasks" with the agent silently absorbed; an armed watch read "job", a word nothing else on screen explained; and the statusline chip was the one status word on the pane you could not click. This PR makes the three kinds separate rows, grouped, and makes the chip the way in.

- **Kernel: every awaited thing is its own row.** `_session_awaiting` now COMBINES its live sources (SDK subagent hooks, pending background launches, armed watches) into `items` — `[{kind, id, label, since, agentId?, detail?, watchId?}]`, kind in `agents | commands | watches | peer | timer` — and derives the legacy `kind` / `count` / `why` / `tasks` from that union. One kind present → the word and sentence it always wore (byte-identical); several → kind `"mixed"`, count = every row, a why that names each group. The old `all(...)`-collapse-to-"task" is gone. A background agent seen by both the hook set and the task stream is one row. `items` ships wherever `awaitingKind`/`awaitingCount` ship: the chat status, the timeline lane, the goal card and the placeholder card. The stamp / overlay / owned-yield arms ship `items: []`; peer arms list their peers as rows. A `cancelWatch` WS door reaches the same `cancel_watch` the CLI does (loud `warn` on a miss). `AWAIT_KINDS_JUDGED` keeps `"mixed"` out of what a closer may file.
- **Words.** `task` → "command", `job` → "watch" (`KIND_WORD`, the timeline's `tlKindWord` twin, `kindWord` pluralizes "watches"); `mixed` has no word, so the badge reads "Awaiting 4". One shared label rule (`awaitWord`) for the chip, the box gist and the feed pill: one row → its word, several of one kind → count + word, several kinds → the number alone with the breakdown ("2 agents · 1 command · 1 watch") in the tooltip.
- **Chat pane.** The Awaiting chip is a `<button data-act="awaitingChip">` delegated once on the stable `#statusline` (click-safe across the per-push rebuild, `.romp-acted` acknowledges); click opens the `#bg-tasks` box and scrolls to it, `setTip` carries the breakdown and the why. The box lists rows grouped by kind under small dim headers (only when 2+ groups): agent rows keep slice 1's open-transcript arrow + Stop while their launch is a live tracked task; command rows keep Stop + the command/output-tail fold; watch rows wear the await-green dot, "armed", a since clock, and Cancel only where a cancel path exists (generic `romp watch` rows carry `watchId`; PR watches have no early-retire path today, so no button). A row with nothing to unfold is not a toggle.
- **Feed.** The pill shows for ANY wait with rows (or an older kernel's `tasks` read as rows of the legacy kind's group), reads by the same `awaitWord`, and its expansion lists the labels grouped under `.ftask-group` headers. `spinFor`'s say-it-once rule stands the caption down under rows of any kind.
- **Card movement timing is unchanged** — `_session_chip`'s formula is untouched; only what the surfaces say and what is clickable changed.

## How verified

- `vscode-extension`: `npm run typecheck` clean, `npm test` 2810 pass / 0 fail, `npm run build` clean.
- Python: `pytest tests/` (minus the live codex smoke) 7372 passed, 40 skipped, 0 failed.
- New tests: `tests/test_awaiting_rows.py` (all three sources at once → mixed; shell + agent is two rows of two kinds, never "task"; the hook/launch merge; single-kind sentences byte-identical to before; watch rows' handles; the mixed kind's enum/parse rules; the cancel door) and `ui/webview/awaiting-rows.test.ts` (the label rules; the three word maps agree on every kind × count; chip-as-button + delegate; grouped rows and per-kind affordances; Cancel → `cancel_watch`; the feed pill for any wait; the vocabulary). Existing pins updated deliberately, each with a note: `test_awaiting_count.py` (also pins that every surface ships the rows), `test_awaiting_card.py`, `test_awaiting_since.py`, `test_awaiting_box_sync.py`, `test_kernel.py`, `test_kernel_awaiting_merge.py`, `test_kernel_awaiting_stamp.py`, `test_kernel_bg_tasks.py`, `test_kernel_owned_yield_awaiting.py`, `test_kernel_timeline_awaiting.py`; UI: awaiting-state, awaiting-box-sync, awaiting-peer-name, bg-tasks-layout, feed-awaiting-swirl, spin-caption, timeline-awaiting, src/bg-tasks.
- Headless screenshots via `tools/ui-verify/shot.sh` from the two synthetic fixtures added here (`tools/ui-verify/fixtures/awaiting-rows-{chat,feed}.html`), dark and light, eyeballed:
  - **Chat:** the expanded box headed "Awaiting 4 · 2 agents · 1 command · 1 watch"; three groups under dim caps headers (AGENTS / COMMANDS / WATCHES); the first agent row with the open-transcript arrow, "· 12m", RUNNING, Stop and a fold caret, the second (hook-only) with the arrow and RUNNING alone; the command row with Stop + caret; the watch row with a green dot, "· 31m", ARMED, Cancel and a caret; the closing note beneath; the statusline below with the green "Awaiting 4" chip rendered as a button. Light theme identical in its own palette.
  - **Feed:** a Working-column card with the pill pressed open reading "Awaiting 4 · 12m", and beneath it the four labels grouped under AGENTS / COMMANDS / WATCHES headers. (The swirl glyph is absent in the headless shots only because the sheet's `url(../media/…)` cannot resolve from the harness's temp dir.)
  - The chat fixture's usage note first asked for a 330px-tall shot, which clipped the box at its 50vh cap; the last commit corrects the note to 540px and says why.

## Left for later (as the plan records)

- An early-retire path for PR watches (then their rows get Cancel too).
- The nested-subagent chain in the viewer header.
- The owned-yield arm words every owned dispatch as a "command" since it carries no type.

### DIFF
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -2876,7 +2876,7 @@ def _parse_plan(raw, menu_len, allow_extend=False):
         elif do in ("done", "block", "awaiting"):
             g, r = _int(o, "goal"), _int(o, "ref")
             ak = str(o.get("kind") or "").strip().lower() if do == "awaiting" else ""
-            ak = {"kind": ak} if ak in AWAIT_KINDS else {}             # garbage/absent → kindless (legacy)
+            ak = {"kind": ak} if ak in AWAIT_KINDS_JUDGED else {}      # garbage/absent/"mixed" → kindless (legacy)
             if g and 1 <= g <= menu_len:
                 ops.append({"do": do, "why": why, "goal": g, **ak})
             elif r and r >= 1:
@@ -9171,7 +9171,15 @@ CLOSER_SYS = (
 # background task/watcher; an external job (cluster/CI/build); a peer session; a scheduled check-back.
 # The judge files one per awaiting verdict; a stamp without one (older judges, legacy stores) is
 # kindless and behaves exactly as before the enum existed.
-AWAIT_KINDS = ("agents", "task", "job", "peer", "timer")
+#
+# Two tuples since 2026-09-05 (plans/subagent-transcripts.md, slice 2): AWAIT_KINDS_JUDGED is what a
+# closer may FILE — the specific thing one goal waits on; AWAIT_KINDS adds "mixed", which only the
+# kernel's LIVE session read (_session_awaiting) produces when a session waits on several kinds at
+# once (an agent AND a background command AND a watch). A stamp is never "mixed": the parse sites
+# validate against the judged tuple, so an LLM emitting the word degrades to kindless like any other
+# off-enum kind, and every lift/supersede rule keyed on a stamp's kind keeps seeing a specific one.
+AWAIT_KINDS_JUDGED = ("agents", "task", "job", "peer", "timer")
+AWAIT_KINDS = AWAIT_KINDS_JUDGED + ("mixed",)
 
 # The PEER-kind write gate's evidence (the user 2026-08-24, after three reports of idle sessions
 # reading "awaiting a peer"): a kind=peer stamp requires an un-answered kind=question THIS session
@@ -9277,7 +9285,7 @@ def _parse_close(raw, menu_len):
                 why = " ".join(str(it.get("why", "")).split())[:300]
                 if kinds:
                     k = str(it.get("kind") or "").strip().lower()
-                    out[n] = {"why": why, "kind": k if k in AWAIT_KINDS else None}
+                    out[n] = {"why": why, "kind": k if k in AWAIT_KINDS_JUDGED else None}   # a closer files a specific kind, never "mixed"
                 else:
                     out[n] = why
         return out
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -17527,6 +17527,13 @@ def _awaiting_kind_phrase(kind, n):
     return "%d timer%s" % (n, "" if n == 1 else "s")
 
 
+def _awaiting_join_items(agents, commands, watch):
+    """The awaited rows in display order — agents, then commands, then the watch dict's rows — the ONE
+    concatenation both the idle read (_awaiting_from_items) and the turn-agnostic read
+    (_session_background_items) use, so the two can never list a different set."""
+    return list(agents) + list(commands) + list((watch or {}).get("items") or [])
+
+
 def _awaiting_from_items(agents, commands, watch):
     """The awaiting answer DERIVED from the live rows — agents (source 0 + the pending agent launches),
     commands (the pending shell/monitor launches) and `watch` (_watch_awaiting's dict, or None). None when
@@ -17535,7 +17542,7 @@ def _awaiting_from_items(agents, commands, watch):
     count = every row, and a why that names each group ("waiting on 2 background agents, 1 background
     command and 1 armed watch"). `since` is the oldest row's own time; `tasks` the labels, so the
     payload's awaitingTasks fallback lists every awaited thing for consumers that still read that."""
-    items = list(agents) + list(commands) + list((watch or {}).get("items") or [])
+    items = _awaiting_join_items(agents, commands, watch)
     if not items:
         return None
     kinds = [k for k in _AWAIT_ITEM_KINDS if any(it["kind"] == k for it in items)]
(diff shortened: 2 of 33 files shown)


yatharth:
Fix pushed (`cf20600c`): a background agent reported by both the SubagentStart hook and the CLI's task stream is now ONE Awaiting row, and the chip counts it once.

**What was wrong.** Seen live 2026-09-06: two background agents + one background shell command read "Awaiting 5 · 4 agents · 1 command" — each agent once by type (with the open-transcript arrow) and once as "Running <description>" (no arrow). Confirmed on the live kernel via `/classify`: the hook rows carried `agentId`; the stream rows carried a `toolu_…` id, a "Running "-prefixed label, and no `agentId` at all. The slice-2 join keyed on `agentId`, but the stream side never had one: `_live_bg_tasks` dropped the lifecycle `task_id` (which IS the agent id for an Agent task, probe-verified and already what `_on_task_event` retires the subagent by); the launch ledger is Bash/Monitor-only and its `agentId` is the ACTING agent; and the sidecar meta map is keyed on the ORIGINAL launch's toolUseId, which a resumed agent's task no longer carries — exactly the agents that duplicated.

**The fix (kernel-side only; nothing moves differently).**
- `sdk_backend._live_bg_tasks` ships `taskId` on every `bgTasks` row (the CLI's own lifecycle key).
- `kernel._bg_live_norm` derives an agent row's `agentId` from that key (a designed field, not a scrape), keeps the ledger's acting agent only for shell rows, and on the transcript-scan path copies the async ack's `agentId` (it used to drop it). The sidecar map stays the last fallback.
- Agent rows drop the CLI's "Running " prefix from the lifecycle description (`_agent_task_label`) — the status word beside the row already says running. Shell rows keep the user's own words untouched (a command described "Running the suite" must stay that).
- A joined row's `since` is the earlier of the launch time and the hook stamp, as the design said (the code only filled a MISSING one; `test_awaiting_rows` pin updated with a note).

**Tests.** New class `OneRowPerAgentAcrossTheHookAndTheStream` in `tests/test_kernel_bg_tasks.py` drives the REAL `_bg_live_norm` from a synthetic snapshot: two agents in both sources + one shell → 3 rows, description labels, both `agentId`s, count 3, "waiting on 2 background agents and 1 background command"; hook-only agent → type label; stream-only agent → description; the tmux scan path carries the ack's `agentId`. `test_sdk_backend` pins `taskId` on every live row; `test_awaiting_count` pins the once-per-agent count. All six new tests failed before the fix. Full Python suite: 7382 passed / 40 skipped; extension typecheck clean, 2817 tests, build ok. No client-side label transformation exists (`awaitRowSpec` renders `it.label` and the arrow off `it.agentId`), so no UI change.

Auto-merge is NOT armed.

yatharth:
Follow-up on this branch (commit after cf20600c): **the box no longer flaps at a turn boundary.**

**Seen live:** with two background agents and a background command, the idle session showed the grouped rows ("Awaiting 3 · 2 agents · 1 command" + the idle note). Sending a message made that view vanish; it came back when the turn ended. Cause: the rows shipped only through `_session_awaiting`, which is idle-only by design (the chip's Awaiting is correct and unchanged), so mid-turn `awaitingItems` was empty and `renderBgTasks` fell to the legacy "N background tasks" list — two presentations of one set of facts.

**Kernel**
- The live-source assembly (hook subagents ⋈ pending agent launches on `agentId`, pending commands, armed watches) is now `_awaiting_live_rows`, shared by `_session_awaiting` (idle read) and the new turn-agnostic `_session_background_items`; `_awaiting_join_items` is the one concatenation.
- `_awaiting_items_payload(aw, sid, path, tmux)` is what the chat status and the timeline lane ship as `awaitingItems` in **both** turn states, same key: the wait's own rows when idle-awaiting, everything in flight otherwise. The read runs under the build's own liveness map (`_serve_live` lends it to `_live_scope`), so the working path still takes no fresh liveness read (`test_kernel_pusher_snapshot` caught the first cut forking tmux twice per build).
- `awaitingWhy` / `awaitingKind` / `awaitingCount` / `awaitingTasks` / `awaitingTaskIds` stay idle-gated exactly as before — nothing about when the chip or a card moves changed.
- Retired the 2026-08-30 mid-turn arm that re-ran `_watch_awaiting` alone into `awaitingWhy` while the turn was open: it kept watches visible mid-turn but made the box read "Awaiting" under a Working chip. Watches ride `awaitingItems` mid-turn like everything else now; `awaitingWhy` means idle-and-waiting everywhere.

**Client**
- One renderer: `renderAwaitWhy` folded into `renderBgTasks`, legacy count-headed branch gone, no "background tasks" string left in the renderer.
- Header follows the wait: idle → "Awaiting 3 · 2 agents · 1 command" (as today) + await-green dot + the idle note; working → "In the background · 2 agents · 1 command" (singular forms fall out of `awaitBreakdown`), worst tracked status as the dot, no note. Rows and their affordances are identical in both states; `s.bgTasks` still lends command rows the output tail and Stop.
- Tracked tasks the wait does not name list under "Also running" (was "Background tasks"). `bg-awaited` outline = a wait or an id in `awaitingTaskIds`, never the chip state. `bgFoldOpen` is only written by the header toggle and the chip click, so the fold survives the idle↔working flip.

**Tests:** +7 kernel tests (`RowsDoNotDependOnIdleness`, the count/kind idle-only pin, the real-normalizer join with the turn open, the re-pinned mid-turn watch test), +3 UI tests ("one presentation" section in `awaiting-rows.test.ts`); re-pinned awaiting-state / awaiting-box-sync / awaiting-peer-name / bg-tasks-layout / src/bg-tasks with notes. Full Python suite 7389 passed; UI 2820 passed; typecheck + build clean. The chat fixture now renders both states one above the other.

PR #941, opened 2026-09-06
The boot orphan reaper matches under systemd (orphaned = parent is not a live kernel, not ppid 1) and reads ps with -ww

ksarma:
Two bugs made `find_orphan_clis` match nothing, with no error, and the second one also blinded the interrupt escalation's process scan. Both live in `kernel/sdk_backend.py`; each commit fixes one and carries tests that fail before it.

## What breaks

**1. Under `systemd --user`, the boot reaper never matched.** `find_orphan_clis` called an SDK CLI orphaned only when its ppid was 1, which describes launchd on macOS. Under the Linux install (`romp-service` writes a `systemd --user` unit) the user manager is a child subreaper, so a CLI left behind by a kernel that died without a drain (a crash respawn, a SIGKILL at the service's stop timeout, a kernel killed by hand) re-parents to the user manager's pid, never to 1. The reap has matched nothing on Linux under the service since it was written. The one thing that ended such a CLI was a full service restart: the unit sets no `KillMode`, so systemd's default empties the whole cgroup. Until then the orphan kept writing the transcript the resumed session was also writing.

**2. Both process scans ran `ps -axo`, which truncates.** procps cuts every output line to an exported `$COLUMNS` (BSD ps does so by default), and the sid an SDK CLI carries sits about 2 KB into its argv, behind `--append-system-prompt`. A kernel started from a shell that exported `COLUMNS` neither reaped orphans nor found its own child to signal (`_session_cli_pid`, the interrupt escalation).

## Reproduction

Both observed on a Linux install under the service (procps-ng 4.0.4):

- `ps -axwwo pid=,ppid=,command= | grep -- '--input-format stream-json'` lists every live SDK CLI with the kernel's pid as ppid. After a kernel dies without a drain (the manager respawns it), the same listing shows the leftover CLIs with the `systemd --user` pid as ppid, not 1, so the old `ppid != 1` test skipped them and the new kernel's boot reconcile logged `reaped 0 orphaned CLI(s)`.
- `COLUMNS=80 ps -axo pid=,ppid=,command= | grep -c -- '--input-format stream-json'` prints 0 while `-axwwo` on the same box prints the live count (20 at the time); every `-axo` line is exactly 80 characters.

## The fix

- `find_orphan_clis` builds a pid -> (ppid, command) map from the listing and calls an SDK-marked CLI holding one of our sids orphaned when its parent is absent from the listing or is not a romp kernel. `_is_kernel_cmd` decides on argv tokens (a token with basename `romp-kernel`, or a `kernel/kernel.py` path), never on substrings, so a `tail -f .../romp-kernel.log` parent is not a kernel. The parent check stays load-bearing: a CLI parented to any live kernel, another instance's included, is left alone (the 2026-07-06 duplicate-backend reap of live sessions stays prevented), and tmux-run interactive CLIs are untouched as before (no stream-json mark). The caller, `_boot_reconcile`, is unchanged.
- One constant, `PS_ARGV = ["ps", "-axwwo", "pid=,ppid=,command="]`, serves both scans. `-ww` is unlimited width on procps and on BSD ps (verified on procps; read from the man page for BSD, not run).

## Tests (`tests/test_sdk_lifecycle_hardening.py`)

Commit 1, `FindOrphanClis`: an orphan under launchd; an orphan under the systemd user manager; an orphan whose parent is absent from the listing; a live CLI under four kernel spellings, never reaped; a CLI under a different live kernel left alone while its neighbour under the user manager is reaped; the token-vs-substring pin for `_is_kernel_cmd`. The two existing fixtures that encoded "ppid != 1 means live" gain the parent kernel's own line. Four of the six new tests fail on the base (the launchd case and the four-spellings case hold under both definitions and stay as regression guards).

Commit 2, `PsArgv`: pins the argv; both call sites through mocks (the escalation finds a child whose sid sits 2 KB into the argv; the reap test asserts the argv it ran); a source scan that no `"-axo"` remains and both sites share the one constant (deliberately brittle: a third ps site in the module must update it); and two Linux-only tests that run the real `ps` with `COLUMNS=80` against a process whose argv carries a marker past its 3000th character. With `PS_ARGV` the marker survives; with the old `-axo` argv the line is cut at 80 columns. The second is the control (procps only) and passes before and after by design; every other PsArgv test fails on the base. The real-ps tests skip where `ps` is absent or not procps, so the macOS job skips them and the ubuntu job runs them.

## Boundaries worth a look

- A live SDK CLI whose direct parent is not a kernel process is reaped at the next boot. The Agent SDK spawns the CLI directly, so a stock install has no such process; a non-exec wrapper script installed as `claude` would create one.
- `_is_kernel_cmd` recognizes what `romp-serve` runs (`exec <python> .../bin/romp-kernel`) and a hand-run `python3 kernel/kernel.py`. A kernel launched another way (`python -m kernel.kernel`) is not recognized, which matters only with two kernels running at once. A process whose command text merely contains such a token (a shell whose command line names the kernel path) counts as a kernel and shields its children; that is the safe direction, since a false "kernel" prevents a reap and never causes one.
- The reaper still sends SIGTERM only, as before.
- The separate `ps -axo pid=,command=` in `kernel/kernel.py` (ssh tunnel cleanup) is untouched: short argv, different matcher. It could route through the same constant in a follow-up.
- No docs change: the reference does not describe the reaper.

Tier: fix

### DIFF
--- a/kernel/sdk_backend.py
+++ b/kernel/sdk_backend.py
@@ -1239,26 +1239,53 @@ def _cli_carries_sid(cmd: str, sids) -> bool:
     return False
 
 
+def _is_kernel_cmd(cmd: str) -> bool:
+    """Does this `ps` command text name a romp kernel? The manager runs `romp-serve`, which execs
+    `<python> <repo>/bin/romp-kernel` (bin/romp-serve's last line), so a kernel shows as
+    `/usr/bin/python3.12 …/bin/romp-kernel`; a hand-run one is `bin/romp-kernel` or
+    `python3 kernel/kernel.py`. Matched on argv TOKENS — the basename `romp-kernel`, or a
+    `kernel/kernel.py` path — never on a substring, so `tail -f …/romp-kernel.log` is not a kernel."""
+    for tok in cmd.split():
+        base = tok.rsplit("/", 1)[-1]
+        if base == "romp-kernel":
+            return True
+        if tok in ("kernel.py", "kernel/kernel.py") or tok.endswith("/kernel/kernel.py"):
+            return True
+    return False
+
+
 def find_orphan_clis(ps_lines: list[str], lastsids: list[str]) -> list[int]:
     """PIDs of ORPHANED SDK-driven `claude` CLIs holding one of OUR sessions (--resume/--session-id
-    in either flag spelling, + the stream-json mark — see _cli_carries_sid). Orphaned = re-parented
-    to launchd (ppid 1): a live SDK CLI is always a child of the kernel that spawned it, so only a
-    dead kernel's leftover — a zombie writer that would fight the resume for the transcript —
-    reaches ppid 1. The parent check is load-bearing: matching on the command line alone let a
-    duplicate backend's reconcile reap freshly-resumed LIVE sessions mid-turn (2026-07-06). Pure
+    in either flag spelling, + the stream-json mark — see _cli_carries_sid). Orphaned = its PARENT
+    is not a live romp kernel (absent from the listing, or a process that is not a kernel): a live
+    SDK CLI is always a child of the kernel that spawned it, so only a dead kernel's leftover — a
+    zombie writer that would fight the resume for the transcript — has any other parent. The parent
+    check is load-bearing: matching on the command line alone let a duplicate backend's reconcile
+    reap freshly-resumed LIVE sessions mid-turn (2026-07-06); and a CLI parented to a DIFFERENT live
+    kernel is that kernel's, never reaped here.
+
+    Until 2026-09-05 "orphaned" meant ppid 1 (macOS: an orphan re-parents to launchd). Under
+    `systemd --user` that never happens: the user manager sets PR_SET_CHILD_SUBREAPER, so a CLI left
+    behind by a kernel that died without a drain (a crash respawn, a SIGKILL at the service's stop
+    timeout) re-parents to the `systemd --user` pid (verified on a Linux box that day: ppid = the user
+    manager, never 1), and this reaper had matched nothing on Linux under the service all along; only
+    a full service restart, which empties the unit's cgroup (systemd's default KillMode=control-group),
+    ever ended such a CLI. The definition now names the property the ppid-1 check approximated. Pure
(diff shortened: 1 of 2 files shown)


ksarma:

Reviewed inline: `find_orphan_clis` now maps pid to (ppid, command) over the whole listing. It reaps an SDK-marked CLI holding one of our sids only when the parent is absent or fails `_is_kernel_cmd`. That predicate matches argv tokens (basename `romp-kernel`, or a `kernel/kernel.py` path), not substrings. On a Linux install under the service, I confirmed the live kernel's line matches through the `romp-kernel` token (`bin/romp-serve` execs it as `<python> bin/romp-kernel`). I also confirmed that all 32 live stream-json CLIs are parented to it and that the new predicate reaps none of them. The user manager is a subreaper, so the old ppid 1 test matched nothing there. On procps-ng 4.0.4, `ps -axo` cuts every line to an exported COLUMNS; `-axwwo` prints full width. The longest live CLI line here is 2447 characters, so the sid sat past an 80-column cut. `PS_ARGV` now serves both the reap and `_session_cli_pid`. Privacy and vocabulary scans of the 197 added lines are clean. The body's claims check out: KillMode, the untouched ps site in `kernel/kernel.py`, and the absence of reaper docs.

- `kernel/sdk_backend.py:1295`: the current kernel's own children are protected only when `_is_kernel_cmd` recognizes its launch spelling. The old `ppid != 1` test protected them unconditionally. Passing `os.getpid()` into `find_orphan_clis` and treating `ppid == own pid` as live would restore that.
- `tests/test_sdk_lifecycle_hardening.py:19`: the header still says ps is patched in every test. The two real-ps PsArgv tests run the real ps and spawn a sleeper child.

I verified the tests red-first per commit in scratch worktrees. The head test file against `origin/main` fails 9: the four FindOrphanClis cases, four PsArgv cases, and the reap test's `PS_ARGV` pin. Commit 1's file fails 4 at main. The head file against commit 1 fails 5. At head all 39 pass, with both real-ps tests running. The tests catch both a substring `_is_kernel_cmd` and an `-axo` `PS_ARGV`. All six required checks are green at head.

Review run by the maintainer's automated review session; verified against head 509c3f11.

PR #942, opened 2026-09-06
To-do card: a TaskCreate the CLI rejected is not a checklist item

ksarma:
## What breaks

A session whose only `TaskCreate` calls were rejected by the CLI showed the To-do card's error state on every build ("Can't read Claude's task store (~/.claude/tasks) for this session, so the to-do list can't be shown accurately"), although it had no to-do list at all.

## Cause

`_fold_tasks` (kernel/kernel.py) treated every `TaskCreate` tool_use as a checklist create. The CLI rejects a malformed call: the paired `tool_result` carries `is_error` and an `InputValidationError` naming the missing `subject`. Two such shapes appear in real transcripts, `{agent_hint, prompt}` with no `subject` and a `{tasks: [...]}` batch. No task was created and nothing was launched. The earlier revision of this PR called the first shape a background-agent launch; no installed Claude Code build has one, and that diagnosis was wrong. The fold still minted a pending item with an empty subject and a creation-order id (`c0`). The card then saw open tasks in the transcript and no readable store, the condition its fail-loudly branch reports.

`event_model.declared_plan`, the judge's mirror of the same fold, had the same hole: the rejected create folded to a keyless item with empty text, which `_sync_declared_plan` minted as a standalone open card reading "(declared step)" that no `TaskUpdate` could ever close.

## Reproduction

Fold a transcript whose one `TaskCreate` is `{"agent_hint": "...", "prompt": "..."}` or `{"tasks": [...]}` and whose paired result has `is_error` set, with an `InputValidationError` naming the missing `subject`:

```
>>> _fold_tasks(rejected_only)
[{'id': 'c0', 'subject': '', 'activeForm': None, 'status': 'pending'}]
>>> [nd["text"] for nd in store["nodes"].values()]   # after _sync_declared_plan over the same transcript
['(declared step)', '(declared step)']
```

Through `build_session` with the store unreadable, the phantom item produces a `todo` event carrying the error. The tests reproduce each of these.

## Fix

Both folds record, in the result pass they already make, the `tool_use_id` of every `tool_result` with `is_error` set, and the `TaskCreate` branch skips a create whose id is in that set. The skip keys on the CLI's rejection, the event that says no task exists, so every rejected shape is covered, including the `{tasks}` batch. An accepted create (its result carries `Task #N`) folds as before, a create whose result has not landed yet still folds under its creation-order id, `TaskUpdate` handling is unchanged, and the card's error branch is untouched: with no phantom task, it fires only when the transcript shows outstanding checklist work.

This revision answers the review hold, which found the earlier diagnosis not real and two sibling paths open. The four items folded:

1. `kernel/kernel.py`, `_fold_tasks`: the skip is keyed on the paired result's `is_error`, replacing the guard keyed on `prompt`/`agent_hint` in the input.
2. The comment there now says a `TaskCreate` the CLI rejected is not a checklist item. The launch and the bgTasks/TaskStop rendering it described do not happen.
3. `kernel/event_model.py`, `declared_plan`: the same skip, so `_sync_declared_plan` mints no placeholder card, with tests in `tests/test_judge_authoritative_plan_sync.py`.
4. `tests/test_kernel.py`: the fixture results carry `is_error` with validation-style text, and a `{tasks}` case pins the re-keyed guard.

## Tests

`tests/test_kernel.py`, `ViewBuilder.test_fold_ignores_a_rejected_taskcreate`, next to the existing todo-store tests. Fixture data is synthetic.

1. `_fold_tasks` over a rejected-only transcript (`{agent_hint, prompt}`) returns `None`.
2. `_fold_tasks` over a rejected `{tasks: [...]}` batch returns `None`.
3. A mixed transcript folds to only the accepted create, `[("1", "vet the pairs")]`.
4. Through `build_session` with `_read_task_store` patched to return `None` and the real fold driven over each transcript: both rejected-only transcripts yield no `todo` event, and the mixed transcript still yields the error card (the control that proves the card path is exercised).

`tests/test_judge_authoritative_plan_sync.py`, beside the existing `declared_plan` and `_sync_declared_plan` tests:

5. `DeclaredPlanAdapter.test_a_rejected_taskcreate_is_not_a_declared_step`: a rejected-only session folds to `[]`; with one accepted create the fold is `[("1", "Design v3")]`.
6. `SyncFindOrCreate.test_a_rejected_taskcreate_mints_no_placeholder_card`: a rejected-only session mints no node and returns `False`; with one accepted create only key `1` is mirrored and no node reads "(declared step)".

Red first, each against the code it pins:

- Without any guard (`_fold_tasks` at the base commit): (1) fails with `[{'id': 'c0', 'subject': '', ...}] is not None`.
- Under the earlier revision's guard keyed on input names: (1) passes and (2) fails with the same `c0` item. The `{tasks}` case is what a guard keyed on input names lets through.
- `declared_plan` without the skip: (5) fails with two items, `{'key': 'c0', 'text': ''}` and `{'key': 'c1', 'text': ''}`; (6) fails with `['(declared step)', '(declared step)'] != []`.

All six pass with the skips.

Verification: `pytest tests/test_kernel.py -n 8` 462 passed; `pytest tests/test_judge_authoritative_plan_sync.py -n 8` 28 passed; `pytest tests -n 8` 7370 passed, 50 skipped.

## Out of scope

The chat drops every raw `TaskCreate`/`TaskUpdate` tool_use before rendering (they are folded into the one checklist card), and a `tool_result` with no tool event to attach to is dropped with it, so a rejected `TaskCreate` and its error never appear in the chat: the user has no sign the model's call failed. That predates this PR. Surfacing a rejected create is a separate change if wanted.

Tier: fix

### DIFF
--- a/kernel/event_model.py
+++ b/kernel/event_model.py
@@ -2768,12 +2768,14 @@ def declared_plan(session):
     [{key, text, activeForm, status}] — the FALLBACK behind task_store_plan for a session with no
     live task store, so downstream (the judge's plan-sync) sees a generic 'declared plan' shape
     instead of raw tool calls. Mirrors the kernel's _fold_tasks, with the same blind spots: only the
-    MAIN agent's TaskCreate/TaskUpdate calls, and only those on the transcript's live chain.
+    MAIN agent's TaskCreate/TaskUpdate calls, and only those on the transcript's live chain — and the
+    same skip: a TaskCreate the CLI rejected (its paired tool_result carries is_error) is not a step.
     `key` is the stable `Task #N` id lifted from TaskCreate's
     result text (a creation-order `cN` fallback if the result is unreadable); `status` rides each
     TaskUpdate. Only TaskCreate/TaskUpdate are folded — plain TodoWrite (no durable ids) is not
     used by romp. Empty list if the session declared no plan."""
     results = {}                                           # tool_use_id → result text (carries 'Task #N')
+    rejected = set()                                       # tool_use_ids whose result came back is_error
     for turn in session["turns"]:
         for a in turn["atoms"]:
             if a.get("type") != "user":
@@ -2782,6 +2784,8 @@ def declared_plan(session):
                 if isinstance(b, dict) and b.get("type") == "tool_result" and b.get("tool_use_id"):
                     c = b.get("content")
                     results[b["tool_use_id"]] = c if isinstance(c, str) else json.dumps(c)
+                    if b.get("is_error"):
+                        rejected.add(b["tool_use_id"])
     tasks, order = {}, 0
     for turn in session["turns"]:
         for a in turn["atoms"]:
@@ -2792,6 +2796,13 @@ def declared_plan(session):
                     continue
                 inp = b.get("input") or {}
                 if b.get("name") == "TaskCreate":
+                    # A TaskCreate the CLI rejected (a malformed call — no `subject`, or a {tasks: [...]}
+                    # batch — answered by an is_error tool_result naming the missing field) created no task,
+                    # so it is not a declared step. Folded, it became a keyless item with empty text that
+                    # _sync_declared_plan minted as a standalone open "(declared step)" card no TaskUpdate
+                    # could ever close. Keyed on the result's is_error, as the kernel's fold is.
+                    if b.get("id") in rejected:
+                        continue
                     m = re.search(r"Task #(\d+)", results.get(b.get("id"), "") or "")
                     key = m.group(1) if m else "c%d" % order
                     af = inp.get("activeForm")
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -18640,6 +18640,7 @@ def _fold_tasks(session):
     order, or None if there were none. The webview renders this as a todo card (kind:'todo') and hides the
     raw Task* calls (ACK_TOOLS) — so the kernel emits the folded card and skips the raw tool events."""
     out = {}                                              # tool_use_id → result text (carries 'Task #N')
+    rejected = set()                                      # tool_use_ids whose result came back is_error
     for turn in session["turns"]:
         for a in turn["atoms"]:
             if a.get("type") != "user":
@@ -18648,6 +18649,8 @@ def _fold_tasks(session):
                 if isinstance(b, dict) and b.get("type") == "tool_result" and b.get("tool_use_id"):
                     c = b.get("content")
                     out[b["tool_use_id"]] = c if isinstance(c, str) else json.dumps(c)
+                    if b.get("is_error"):
+                        rejected.add(b["tool_use_id"])
     tasks, order = {}, 0
     for turn in session["turns"]:
         for a in turn["atoms"]:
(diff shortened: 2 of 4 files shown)


ksarma:
Holding this one: the guard suppresses the shape that was seen, but the cause given in the body and the code comment is not real, and two sibling paths stay open. Four things to fold, then I merge.

Reviewed inline: the diff is one guard at the top of the TaskCreate branch of `_fold_tasks` plus one ViewBuilder test. `_read_task_store`, the store join, and the card's fail-loudly branch are unchanged, so the store stays authoritative. The guard narrows what the fallback transcript fold counts as a checklist item. It suppresses the observed phantom item and cannot drop an accepted checklist create. But no installed Claude Code build defines a background-agent form of TaskCreate. The binaries contain no `agent_hint` key and no result text of the kind the fixture uses. The one real record with that input shape is a tool_result with `is_error` set and an InputValidationError naming the missing `subject`. The CLI created nothing and launched nothing. The guard keys on two input key names rather than on that error result, which is the event that says no task was created. A `{tasks: [...]}` create, also present in real transcripts and rejected the same way, still folds to a phantom item at head. The judge's mirror fold in `event_model.declared_plan` did not get the guard.

To fix before merging:

- `kernel/kernel.py:18602`: the comment describes a launch that does not happen and says bgTasks/TaskStop render it; nothing does. Say instead that a TaskCreate the CLI rejected is not a checklist item.
- `kernel/kernel.py:18609`: key the skip on the paired tool_result's `is_error`, or on a result that carries no `Task #N`. The fold's first pass already walks every result. That also covers the rejected `{tasks: [...]}` shape.
- `kernel/event_model.py:2794`: `declared_plan` still folds the same create into an open item, which `_sync_declared_plan` mints as a standalone open card with the placeholder text (declared step) that no TaskUpdate can close. Apply the same guard, with a test beside `tests/test_judge_authoritative_plan_sync.py`.
- `tests/test_kernel.py:405`: mark the fixture results `is_error` with validation-style text, and add a `{tasks}` case so the test survives a re-keyed guard.

Red-first holds for the current test: over `origin/main` it fails at its first assertion with the message the body gives. The whole file at main is 1 failed, 461 passed, and 462 passed at head across three xdist runs. All six required checks are green at head. The hold is about the diagnosis, not the tests.

Review run by the maintainer's automated review session; verified against head 764089c8.

ksarma:

Reviewed inline: both folds record the tool_use id of every tool_result carrying `is_error`, in the result pass they already make. The TaskCreate branch skips a create whose id is in that set, before the Task #N lookup and before the creation-order counter moves. This replaces the earlier guard on the prompt and agent_hint input keys. The kernel comment now says a TaskCreate the CLI rejected is not a checklist item, and no longer mentions a launch or bgTasks/TaskStop. `declared_plan` in `kernel/event_model.py` carries the same three lines, so `_sync_declared_plan` mints no placeholder card and the two folds stay identical. An accepted create whose result carries Task #N still folds in both. A create whose result has not landed still folds under its cN id. A TaskUpdate naming a rejected create's id is a no-op: that branch is a `tasks.get` followed by an in-place status write. Nothing else in either fold changed. The fixtures mark the rejected results `is_error`, with InputValidationError text naming the missing subject. They include the `{tasks}` batch, the case the input-keyed guard let through. Added lines are clean against the private-strings list, the hostname and home paths.

- Follow-up, out of scope: `kernel/kernel.py:18682` and `kernel/event_model.py:2812` still apply a TaskUpdate whose paired result carries `is_error`. The same rejection-keyed skip fits there.
- Follow-up, expectation only: stores already holding pre-fix "(declared step)" mirrors will see them move to completed once rather than disappear. The move comes from the agent-done verdict at `kernel/judge.py:5074`, the existing heal for any watched-open item that leaves the plan.

Red first at origin/main (ef789966) with the PR's two test files checked out. `test_fold_ignores_a_rejected_taskcreate` fails with the c0 item. `test_a_rejected_taskcreate_is_not_a_declared_step` fails with two keyless items. `test_a_rejected_taskcreate_mints_no_placeholder_card` fails with two "(declared step)" nodes. At head both files pass, 490 in all (462 + 28). CI: all six required checks pass at head.

Review run by the maintainer's automated review session; verified against head 9b974149.

PR #943, opened 2026-09-06
Turning Auto Nudge on acts at once again: the setAutoNudge handler re-ticks on a real apply

ksarma:
## What breaks

Turning Auto Nudge on from the gear no longer acts in the same handler call. Since #574 the `setAutoNudge` WS arm called `_auto_nudge_tick(int(time.time()), _tmux_sessions(), run_dead_wait=False)` right after the setter, so the first nudge pass ran at once. At tip the arm only applies the setting (and answers a stale gesture), and the first pass waits for the pusher's next periodic tick.

## How it was lost

#846 (T208) added the `setCompactSuggest` arm by inserting its two lines between `_set_auto_nudge(bool(msg["enabled"]))` and the act-now comment and tick. The tick line itself did not change, so it stayed in the diff as context and became the new arm's body; a line-based blame shows no removal, and #846's description mentions no setAutoNudge change. #879 then re-derived the gt-gated arm as `if _set_auto_nudge(...) is None: _tell_stale_gesture(client)`, keeping that shape.

The tree still describes the missing caller. `_auto_nudge_tick`'s one-observer comment says setAutoNudge's WS-handler tick passes `run_dead_wait=False`, its ack-fast comment says the WS setAutoNudge op calls it directly on turn-on, and `DeadWaitOneObserver.test_the_setautonudge_call_site_skips_the_sweep` has been passing because its whole-file substring is satisfied by the setCompactSuggest line.

## Reproduction

Drive `Handler._dispatch_ws` with `_auto_nudge_tick` stubbed to record calls and send `{"type": "setAutoNudge", "enabled": true, "gt": <fresh stamp>}`. At tip the stub records nothing. `AutoNudgeTurnOnActsNow` in `tests/test_setting_gesture_order.py` is that reproduction: it fails on the base commit and passes with the fix.

## The fix

The setAutoNudge arm now mirrors the setCompactSuggest arm below it. When the setter returns a stamp (a real apply), the arm calls `_auto_nudge_tick(int(time.time()), _tmux_sessions(), run_dead_wait=False)`; when it returns None (a stale stamp, the gesture's own echo, or a failed write) the arm fires no tick, and the stale case still answers the delivering socket with `settingStale`. `run_dead_wait=False` keeps the death transition's one observer, as before. `_compact_suggest_tick`'s claim-before-send comment now counts three concurrent entry points (the pusher and both WS re-ticks); #897's claim under `_NUDGE_LOCK` already makes the third entry safe.

Two notes for review:

- Turn-off no longer re-ticks: the second commit gates the arm's tick on `msg["enabled"]`, the one-line change offered in the first revision. The setCompactSuggest arm still ticks on any real apply.
- The two `_auto_nudge_tick` comments that describe the setAutoNudge caller are unchanged: they are accurate again with this fix, and #936 is editing that region.

## Re-entry guard (from review)

The branch is two commits, rebased onto main after #945. The first is the change above; its stale branch carries the two-argument `_tell_stale_gesture(client, msg)`.

The second commit makes `_auto_nudge_tick` single-flight. The tick has two concurrent entry points, the pusher's periodic pass (0.5 s backstop) and the act-now pass the WS arms run on the handler thread, and the nudge send has no claim-before-send, so a WS pass overlapping the pusher's injected the same nudge twice into one session. The tick now takes a module-level `Lock` with a non-blocking acquire and stands down whole when it is held. The body moves to `_auto_nudge_pass` (the `_pusher_cycle` / `_pusher_cycle_jobs` split), so the diff is the guard rather than a re-indented walk. The loser loses nothing: the flag write precedes the turn-on's tick, so a pass found in flight already reads the turned-on world, and a stood-down pusher pass re-runs on its next cycle. It is a plain `Lock`, not an `RLock`, so a same-thread re-entry stands down the same way. I kept the direct tick rather than setting `_pusher_wake`: the reviewed shape and the one-observer pins keep their meaning, and the guard also covers the setCompactSuggest arm's re-entry at main.

Also from the review: the arm ticks only on turn-on (`msg["enabled"]`), so a turn-off no longer forks the session listing on the WS thread for a tick that returns at the enabled gate, and both arms' "don't wait 4s" comments now name the 0.5 s cadence.

## Tests (red first)

With both commits' test changes applied on the base commit, seven tests fail (five from the first commit, two from the second); with the fix, all pass.

- `WiringPins.test_auto_nudge_branch_gates_on_the_stamp_and_skips_the_tick_on_stand_down` (replaces `test_auto_nudge_branch_passes_the_stamp`): the source pin flips from `is None` to `is not None`.
- `AutoNudgeTurnOnActsNow` (new, `tests/test_setting_gesture_order.py`): drives the real `_dispatch_ws` with the tick stubbed. A fresh stamped turn-on ticks once with `run_dead_wait=False` and the handler's `_tmux_sessions()` listing; a stale stamp and then the gesture's own echo add no tick; an unstamped toggle (an older dashboard) ticks once.
- `DeadWaitOneObserver.test_the_setautonudge_call_site_skips_the_sweep` (`tests/test_dead_wait_block.py`): slices the setAutoNudge arm out of `_dispatch_ws`'s source and requires the tick inside it, and requires `run_dead_wait=False` at every WS call site (the count of `_auto_nudge_tick(int(time.time())` calls equals the count carrying `run_dead_wait=False`).
- `AutoNudgeTickIsSingleFlight` (new, second commit, `tests/test_setting_gesture_order.py`): drives the real tick with the per-session walk replaced by a recorder that holds the first pass mid-send, then runs the other entry point against it in both interleaves: the pusher's pass in flight when a turn-on arrives through the real `_dispatch_ws`, and the WS pass in flight when the pusher's pass arrives. Each asserts exactly one send; the second also asserts that the stood-down pusher pass runs on its next cycle. Both failed on the first commit with two sends. A third test pins that a raising pass releases the guard; it failed against a variant without the `finally`.
- `AutoNudgeTurnOnActsNow.test_a_real_turn_off_applies_and_fires_no_tick` (second commit): failed on the first commit with two ticks.

Full `pytest tests` at the two-commit head: 7374 passed, 50 skipped under xdist. No shell or UI file changes.

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -5546,20 +5546,21 @@ def _compact_suggest_tick(sid, tm, now):
     with _NUDGE_LOCK:                                  # CLAIM BEFORE SEND (the double-send race,
         # 2026-09-01): the check above releases the lock before the fire-time gates, and this tick
         # has three concurrent entry points (the pusher's periodic pass, setAutoNudge's and
-        # setCompactSuggest's WS re-ticks) — latching only after the send let two entries pass the
-        # same unlatched check and inject the suggestion twice into one session. The latch write
-        # IS the claim: a fresh read under the lock re-derives what is still due (a concurrent
-        # entry may have claimed it in the gap), the winner latches and sends, the loser stands
-        # down right here. Same direction as the nudge machinery's own bookkeeping
-        # (_mark_auto_nudged records only after its send, so a failed send retries), inverted at
-        # this seam because the send must be single-shot — and the failure arm below rolls the
-        # claim back under the lock, so "never latch a fire that never went out" still holds
-        # durably: the next tick retries the same crossing. The claim (and the rollback) is
-        # durable against every OTHER ledger writer too: all of them hold _NUDGE_LOCK across
-        # their read-modify-write, and the ones with IO mid-span re-read fresh under the lock
-        # and mutate only their own keys — before that migration (2026-09-01), a debt reminder's
-        # pre-send snapshot written whole erased a concurrent claim (or resurrected a rolled-back
-        # one) and re-opened the double send through the side door.
+        # setCompactSuggest's WS re-ticks; single-flight since PR #943 via _AUTO_NUDGE_TICK_LOCK,
+        # but the latch is the SEND's own single-shot guarantee, so it stays) — latching only after
+        # the send let two entries pass the same unlatched check and inject the suggestion twice
+        # into one session. The latch write IS the claim: a fresh read under the lock re-derives
+        # what is still due (a concurrent entry may have claimed it in the gap), the winner latches
+        # and sends, the loser stands down right here. Same direction as the nudge machinery's own
+        # bookkeeping (_mark_auto_nudged records only after its send, so a failed send retries),
+        # inverted at this seam because the send must be single-shot — and the failure arm below
+        # rolls the claim back under the lock, so "never latch a fire that never went out" still
+        # holds durably: the next tick retries the same crossing. The claim (and the rollback) is
+        # durable against every OTHER ledger writer too: all of them hold _NUDGE_LOCK across their
+        # read-modify-write, and the ones with IO mid-span re-read fresh under the lock and mutate
+        # only their own keys — before that migration (2026-09-01), a debt reminder's pre-send
+        # snapshot written whole erased a concurrent claim (or resurrected a rolled-back one) and
+        # re-opened the double send through the side door.
         d = dict(_auto_nudge_data())                   # fresh read — a concurrent writer's blob is
         cs = dict(d.get("compactSuggested") or {})     # not clobbered
         held = [t for t in [int(x) for x in cs.get(sid, [])] if tokens >= t]
@@ -5599,6 +5600,20 @@ def _compact_suggest_tick(sid, tm, now):
     return True
 
 
+# The tick is SINGLE-FLIGHT (PR #943 review): it has two concurrent entry points — the pusher's
+# periodic pass (0.5 s backstop) and the setAutoNudge / setCompactSuggest arms' act-now pass on the
+# WS handler thread — and the nudge send has no claim-before-send (_mark_auto_nudged records AFTER
+# the send so a failed send retries; _compact_suggest_tick's latch covers only its own seam), so two
+# passes that overlapped each derived the same due goal from the same store and injected the same
+# nudge twice into one session — reproduced from two threads at the 0.5 s cadence. A non-blocking
+# try-acquire, never a wait: the loser stands down whole, and loses nothing. The flag write precedes
+# the turn-on's tick, so a pass found in flight already reads the turned-on world (before the write
+# it would have returned at the enabled gate), and a pusher pass that meets the WS pass re-runs on
+# its own cadence — at worst an act-now degrades to the next cycle. A plain Lock, not an RLock: a
+# same-thread re-entry (a WS tick nested in a pass) must stand down the same way.
+_AUTO_NUDGE_TICK_LOCK = threading.Lock()
+
+
 def _auto_nudge_tick(now, tmux, run_dead_wait=True):
     """One Auto-Nudge pass (from the periodic pusher). For each ALIVE, IDLE session (its turn ended) that
     isn't awaiting/compacting/api-error, isn't WAITING ON A LIVE PEER (a wait isn't a stall — the human's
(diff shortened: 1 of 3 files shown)


ksarma:
The review passes; holding only for a rebase. #945 landed first and changes the `_tell_stale_gesture(client)` call on these same lines to `_tell_stale_gesture(client, msg)`, so this branch now conflicts in `kernel/kernel.py`. Once it is rebased onto main and CI is green, I merge. One follow-up to file.

Reviewed inline: the `setAutoNudge` arm in `Handler._dispatch_ws` now branches on whether the setter returns a stamp. A real apply runs `_auto_nudge_tick(int(time.time()), _tmux_sessions(), run_dead_wait=False)` on the handler thread. A stale stamp, the gesture's own echo, or a failed write fires no tick. The stale case still answers the delivering socket. This matches the `setCompactSuggest` arm directly below it. It restores the form the arm had from #574 until #846 inserted the new arm between the setter and the tick line. I confirmed that at the #574, #846 and #879 merge commits. The tick keeps `run_dead_wait=False`, so the death transition still has one observer. The rewritten dead-wait pin slices the `setAutoNudge` arm out of the method source instead of matching the whole file. That is why the old pin stayed green while the tick was missing. The `AutoNudgeTurnOnActsNow` tests drive the real handler with the tick stubbed. They cover a fresh apply, a stale stamp, an echo, and an unstamped toggle. The updated `_compact_suggest_tick` comment correctly counts three entry points. Added lines carry no private strings, hostnames, or home paths. No injected text changed.

- `kernel/kernel.py:35716`, follow-up. The nudge send has no claim-before-send, and `_auto_nudge_tick` has no tick-level lock. So this WS-thread pass can overlap the pusher's pass (0.5 s backstop) and inject the same nudge twice into one session. Three independent reproductions from two threads confirmed the double send. The bug is pre-existing: the `setCompactSuggest` arm re-enters the same way at main. This change re-opens it for the toggle people actually flip, and turn-on is when the most goals are due. Either fix closes it: a try-acquire guard around `_auto_nudge_tick`, or having the arm set `_pusher_wake` so the pusher stays the single owner. The arm's 4 s comment predates the 0.5 s pusher cadence. Gating the tick on the enabled flag can go in the same change.
- `kernel/kernel.py:35718`, the rebase: #945 pins the two-argument `_tell_stale_gesture(client, msg)` form. Carry that call in this arm's stale branch.

Red-first verified. With only the two test files on origin/main, 5 failed / 72 passed. The failures were the three `AutoNudgeTurnOnActsNow` tests, the `WiringPins` is-not-None pin, and the sliced `DeadWaitOneObserver` pin. At head, 77 passed, also under xdist. All six required CI checks are green at head.

Review run by the maintainer's automated review session; verified against head ba56632f.

ksarma:
Merging as is. One PR body edit and two small items go to follow-ups.

Reviewed inline: the rebased `setAutoNudge` arm ticks only on a real apply (the setter returns a stamp) with `enabled` true. Its stale branch calls the two-argument `_tell_stale_gesture(client, msg)` that #945 pins. The head has 15 two-argument calls and no one-argument form, and #945's pins pass. The single-flight guard is a module-level plain `threading.Lock`, taken with `acquire(blocking=False)` and released in a `finally`. Nothing ever blocks on it, so it cannot deadlock against `_NUDGE_LOCK`. A same-thread re-entry returns the same way. The guard sits inside `_auto_nudge_tick`, the only caller of `_auto_nudge_pass`, so the pusher and both WS arms (`setAutoNudge` and `setCompactSuggest`) pass through it. A pusher pass that the guard rejects re-runs on its 0.5 s cadence. A pusher pass found in flight by a WS turn-on must already have passed the enabled gate after the flag write. A due nudge is therefore delayed by at most one cycle. The new tests gate on `Event.wait` and `Thread.join` with backstops and no sleeps. They use a placeholder sid and touch only a private state directory. Privacy and vocabulary scans of the added lines, commit messages and body are clean.

Follow-ups:

- PR body, Tests (red first): the sentence saying five tests fail at the base is the first revision's count. With the head's two test files at `origin/main`, seven fail. The two remaining single-flight tests are green at main and red only against the first commit, as the per-test bullets already say. Body-only edit.
- `tests/test_dead_wait_block.py:447`: the nested WS-shaped tick in `test_a_mid_pass_ws_tick_leaves_the_transition_alone` now returns at the guard before `run_dead_wait` is read. The assertion is therefore satisfied by the lock rather than by the sweep skip the comment describes. The sequential sibling still pins the skip; only the comment is out of date.
- `kernel/kernel.py:35937`, pre-existing: the WS-thread tick is unwrapped, unlike the pusher's `try/except`. The reader loop re-raises `OSError` as a socket failure, so an `OSError` from the pass head would close the dashboard connection with no log line. The `setCompactSuggest` arm has been unwrapped the same way since #846. One shared wrap around both arms would fix both.

Test evidence: with the head's test files in a scratch worktree at `origin/main` ef789966, 7 failed / 81 passed across `tests/test_setting_gesture_order.py` and `tests/test_dead_wait_block.py`. The seven are the WiringPins pin, the four AutoNudgeTurnOnActsNow tests, the pusher-meets-WS single-flight test, and the DeadWaitOneObserver call-site pin. At head, 88 passed. The second commit's three new assertions fail against the first commit's kernel and pass at head. All six required checks pass at c334e4f0.

Review run by the maintainer's automated review session; verified against head c334e4f0.

PR #944, opened 2026-09-06
Tests: every temp directory the suite makes is removed when the run ends

ksarma:
## Summary

The test suite never removed the temp directories it created. This PR records every `tempfile.mkdtemp` call the test process makes and removes those directories when the run ends. Tests and test infrastructure only; no product code changes.

## What breaks

`tests/__init__.py` and `tests/conftest.py` each create a state root per process with `tempfile.mkdtemp(prefix="romp-tests-state-")` and never remove it, so every pytest process (the controller and each xdist worker) leaves two behind. About 490 test modules carry the same module-level `XDG_STATE_HOME = tempfile.mkdtemp()` preamble, and 293 of the 503 modules that call `mkdtemp` have no cleanup anywhere in the file. A full run leaves thousands of directories in the system temp directory.

On one developer machine that runs the suite many times a day, about 1.9 million of these directories accumulated. Listing the temp directory took minutes and tens of gigabytes of memory, and the out-of-memory kills that followed took down a running romp kernel.

## Reproduction

At current `main`, with a fresh private temp root:

```
export TMPDIR=$(mktemp -d)
python -m pytest -q -n 4 tests/test_kernel.py tests/test_judge.py
ls "$TMPDIR" | wc -l    # 101
```

Ten of the 101 entries are `romp-tests-state-*` roots (five processes, two each); the rest are per-test `mkdtemp` directories. With this PR the same commands leave 1 entry, pytest's own `pytest-of-<user>` base for `tmp_path`, and repeating the run keeps it at 1.

## The fix

`tests/__init__.py` wraps `tempfile.mkdtemp` once (the wrapper is marked, so a second import does not double-wrap) to record every directory the process creates, and registers an `atexit` sweep that removes what was recorded. `tests/conftest.py` calls the same sweep from `pytest_sessionfinish`, which fires in the controller and in every xdist worker. The sweep is idempotent and removes a recorded path only if its realpath is strictly under the realpath of `tempfile.gettempdir()`, it is a directory, and it is not a symlink. `tempfile.TemporaryDirectory` goes through the same `mkdtemp` and cleans itself first, so the sweep finds nothing of it.

The ordering the design rests on: pytest imports the conftest as `tests.conftest` through the package (the repo has no pytest config file, so the default `prepend` import mode applies), so the package installs the hook before conftest mints its own root, and that root is recorded too. A bare `python -m unittest tests.test_x` is covered through the package import. A direct script run (`python tests/test_x.py`) is not, which is the boundary the existing state floor already has.

## Tests

`tests/test_tempdir_hygiene.py` makes three checks:

- the hook is installed (`tempfile.mkdtemp.romp_tracked`);
- a child `python -m pytest` session that runs the in-process test removes the directory it made and its `XDG_STATE_HOME` root (conftest's session-end sweep);
- a child `python -m unittest` run does the same (the package's atexit sweep).

The child writes the two paths to a marker file named by `ROMP_HYGIENE_MARKER`; the parent asserts both are gone once the child exits. All three fail on `main`: the first because the hook is absent, the two child runs because the child exits non-zero on that same assertion.

`tests/test_state_isolation_order.py` still passes; both floor lines are kept (the conftest line gains a trailing comment).

Verification: full suite with `-n 8` (7325 passed, 50 skipped, one failure in `test_postal_relay_honesty` that is a pre-existing xdist-ordering flake and passes alone) and the CI-style serial run of `test_tempdir_hygiene.py`, `test_state_isolation_order.py` and `test_kernel.py` (466 passed). The full run left no `romp-tests-state-*` or per-test directories in the private temp root.

## Not covered by this PR

- Temp files. `mkstemp` and `NamedTemporaryFile(delete=False)` create files, which the hook does not record. Most sites clean up in-file; the `_set_live` helper in `tests/test_postal_live_only.py` and `tests/test_postal_relay_honesty.py` does not, so a full run still leaves about 40 small `.json` files. That is a separate, smaller fix.
- Temp use by child processes and by shell `mktemp`. At tip none of it lands in the system temp directory: kernels spawned by tests write only inside the state root, and the bats suites `rm -rf` in teardown.
- `tests/manager-drain.test.js`, whose module-level `mkdtempSync` has no removal: one directory per `node --test` run.
- A run ended by pytest-timeout's `--timeout-method=thread`, which exits through `os._exit` and so skips both `atexit` and `pytest_sessionfinish`. A hung CI run leaks that run's directories, which is acceptable on an ephemeral runner.

A follow-on that would cover children and files as well is to point `TMPDIR` at one private root per run and remove it at session end. This PR takes the smaller, source-level step because it also covers bare unittest runs, which a pytest-only redirect would not.

Tier: tests-only

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -4,8 +4,53 @@ resolve their state root at import time. conftest.py covers only pytest; without
 unittest run operated on the REAL ~/.local/state/romp (tests/test_kernel.py overwrote the real
 remotes.json that way today). Does NOT cover `cd tests && python -m unittest test_x` or a direct
 script run — the per-module preambles do, and tests/test_state_isolation_order.py enforces them."""
+import atexit
 import os
+import shutil
 import tempfile
+
+# Temp-directory hygiene (2026-09-06): the suite used to leak every directory it made. This floor and
+# conftest.py's each minted a state root per process and never removed it (two per pytest process:
+# eighteen for a `pytest -n 8` run), and about three hundred test modules call tempfile.mkdtemp with
+# no cleanup of their own. On a developer machine that ran the suite many times a day that left about
+# 1.9 million directories under the system temp directory, enough that listing it took minutes and
+# tens of gigabytes of memory, and the out-of-memory kills that followed took a running romp kernel
+# down. The fix is at the source: every tempfile.mkdtemp call made in this process is recorded, and
+# everything recorded is removed at interpreter exit (and, under pytest, at session end; see
+# conftest.py). Only directories this process created, and only under the system temp directory, are
+# ever removed. tempfile.TemporaryDirectory goes through the same mkdtemp and cleans itself first, so
+# the exit sweep finds nothing of it. This package imports before conftest.py (pytest imports
+# `tests.conftest` through the package), so conftest's own root is recorded too.
+_MADE_DIRS = []
+_REAL_MKDTEMP = tempfile.mkdtemp
+
+
+def _tracked_mkdtemp(*a, **k):
+    p = _REAL_MKDTEMP(*a, **k)
+    _MADE_DIRS.append(p)
+    return p
+
+
+_tracked_mkdtemp.romp_tracked = True          # tests/test_tempdir_hygiene.py asserts the hook is installed
+if not getattr(tempfile.mkdtemp, "romp_tracked", False):
+    tempfile.mkdtemp = _tracked_mkdtemp
+
+
+def remove_made_dirs():
+    """Remove every directory this process minted through tempfile.mkdtemp. Idempotent; registered
+    with atexit here and called again from conftest's pytest_sessionfinish."""
+    root = os.path.realpath(tempfile.gettempdir())
+    for p in reversed(list(_MADE_DIRS)):
+        try:
+            rp = os.path.realpath(p)
+            if rp != root and rp.startswith(root + os.sep) and os.path.isdir(rp) and not os.path.islink(p):
+                shutil.rmtree(rp, ignore_errors=True)
+        except Exception:
+            pass
+    _MADE_DIRS.clear()
+
+
+atexit.register(remove_made_dirs)
 
 os.environ["XDG_STATE_HOME"] = tempfile.mkdtemp(prefix="romp-tests-state-")
 os.environ.pop("ROMP_STATE_DIR", None)  # a live kernel's export outranks the XDG floor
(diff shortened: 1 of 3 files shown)


ksarma:
Merging as shipped. Two follow-ups below; neither holds the merge.

Reviewed inline: the mkdtemp wrapper in `tests/__init__.py` passes arguments and the return value through unchanged. It is marked, so a second import cannot double-wrap. The diff touches only `tests/`, so the tests-only label fits. No Python under `bin/`, `kernel/`, `cli/`, `postal/`, `claude/`, `scripts/` or `hooks/` calls mkdtemp or TemporaryDirectory. Spawned kernels never import the tests package. Nothing under test can observe the hook. The sweep iterates only paths this process got back from mkdtemp. It removes one only when its realpath is strictly under the temp root's realpath, is a directory, and is not a symlink. The hardcoded /tmp socket dir in `test_kernel_socket_deliver.py` is minted outside that root and is skipped. The failure direction is a leak, never a deletion outside the root. With no pytest config file at the repo root, the default prepend mode imports conftest through the package. Conftest's own state root is recorded, as the body says. `tests/` has no os.fork or multiprocessing, so the atexit sweep cannot run in a forked child against the parent's list. The body's reproduction holds: `-n 4` over `test_kernel.py` and `test_judge.py` leaves 101 entries at main and 1 at head. That entry is pytest's own tmp_path base, not a mkdtemp directory.

- `tests/test_tempdir_hygiene.py:21` (follow-up): a direct `python tests/test_tempdir_hygiene.py` run never imports the tests package. The hook assertion then fails, and the two `romp-hygiene-*` scratch directories from `_run_child` are left behind. Confirmed at head: `FAILED (failures=1)`, two directories. Importing `tests` at module top for the `__main__` case, or dropping the block, fixes it.
- `tests/conftest.py:17` (follow-up): the plain hookimpl runs before pytest's runner `pytest_sessionfinish`, which performs deferred teardown after an interrupted run. A strict class-level rmtree of a mkdtemp'd directory would then raise. No current teardown does. `@pytest.hookimpl(trylast=True)` closes it.

Red-first verified: with only the new test file checked out onto origin/main, all three tests fail. The hook is not installed, and both child runs exit non-zero on that same assertion. At head, the file plus `test_state_isolation_order.py` pass 5/5 serially. The new file passes 3/3 under `-n 4` and under `python -m unittest`, leaving 0 entries in a fresh private temp root. CI: all six required checks green at e95ce457.

Review run by the maintainer's automated review session; verified against head e95ce457.

PR #945, opened 2026-09-06
The #879 follow-ups: gesture stamps climb above every stamp seen, a stood-down pick can be applied anyway, and the federation queue's edges are journaled, cleared per entry, and toasted once per refused flush

ksarma:
Follow-up to the review note on #879, built as proposed there: stamp gestures as `max(Date.now(), lastSeenGt + 1)` with the gear learning each store's current `gt` from `/version`, an explicit "apply anyway" action on the toast, and copy that does not claim another device acted. The three smaller notes from the same review ride as a second commit: the queue-time breadcrumb, the per-entry clear in `flushPending`, and one host-named toast per refused flush. The `setCompactSuggest` note is a comment fix only: it stays out of `KERNEL_SETTING` by design (per-install), and the comment above the per-install toggles in gear.js now says so.

## What breaks

**Clock skew locks a setting.** Gesture stamps were the device's bare wall clock. A laptop ten minutes ahead picks a triage model; every kernel stores a stamp ten minutes in the future. For the next ten minutes every pick of that setting from a correctly-clocked phone is refused (`gt <= applied_gt`), the phone's gear toasts that the setting "changed more recently somewhere else" (which the kernel cannot know; it only knows it holds a larger number), and no gesture from the phone can win.

**Queue edges.** A setting queued for a host whose socket was down left nothing in the client-diag journal (only the non-setting drop path did), so a later disagreement between machines was not attributable. `flushPending` sent every queued entry and then cleared the map once, so a send that threw mid-flush left the delivered entries in the map and the next open replayed them. A stale flush refused by N kernels drew N identical toasts naming no host.

## Reproduction

All synthetic, in the existing harnesses:

- `tests/test_setting_gesture_order.py::ASkewedClockCannotLockTheStore`: a store holding a future stamp refuses a correctly-clocked pick; on the base commit nothing lets the dashboard climb above it.
- `ui/webview/federation-send-queue.test.ts`: a `FlakySocket` whose second send throws; on the base commit the redial re-sends the delivered entry and the queue still holds both.
- `ui/webview/setting-stale-fold.test.ts`: three frames for one gesture, from `web`, `api` and the local kernel; on the base commit, three toasts and no host text.

### Commit 1: the gesture clock and Apply anyway

**Kernel.** `/version` gains `settingsGt`: every gt-gated store's last-applied stamp, keyed by the store names the `settingStale` frame already uses (14 keys, epoch-ms integers; nothing path-shaped, so still fine on the auth-exempt route). The `settingStale` frame gains `gesture`: the delivering socket's own refused message echoed back without its stamp, so the dashboard can re-issue it as a new gesture; the stamp is dropped on purpose so a re-issue can never reuse the stale one. `_tell_stale_gesture(client)` becomes `_tell_stale_gesture(client, msg)` at its 14 call sites.

**Webview clock.** A small shared CJS module, `ui/webview/gesture-clock.js`, keeps a per-store logical clock seeded by the wall clock: `stamp(store) = max(Date.now(), seen[store] + 1)`, recording its own result. Every emitter (the gear's fifteen literals, the feed's judge-limit switch, the file viewer's consent) mints its stamp through it at the click, still inside the message literal; the queue's deliver-unchanged contract is untouched. The gear learns each store's stamp from `/version` on every open (it fetched it already), the viewer from the `/version` read it already makes before asking consent, and the `settingStale` listener from the frame's `storedGt`. Side effect: two clicks on one setting within one millisecond no longer collide as an equal-stamp refusal.

**Toast.** It names the setting and the kept value, says the pick was not applied because a later pick is already in place, no longer says where or by whom, and offers **Apply anyway**: the echoed gesture re-issued with a fresh stamp above everything this page has seen. A new user gesture is legitimate new information, the event the ordering rule asks for. The action is offered only when the echo's type matches the setting the frame names (`STALE_TYPE`), so a frame from any linked kernel can re-issue that one setting and nothing else.

### Commit 2: the queue's edges

- `sendRemote` emits a `sendqueue` breadcrumb `{host, msgType, gt, rs, superseded?}` on the queue branch: `rs` is the socket's `readyState` at queue time (-1 if not dialed), `superseded` the replaced entry's `gt`. Nothing mints a timestamp in the send path; the kernel stamps the journal line.
- `flushPending` deletes each entry as its send succeeds; a throw halts the loop, leaves the rest queued for the next open, and posts a `hostconn` `flush-halt` breadcrumb naming flushed and held types. The open breadcrumb on a complete flush is unchanged.
- The `settingStale` frame gains the refused gesture's own `gt`; `prefixInbound` stamps `host` on a remote kernel's `settingStale` as it does for `cardMoveAck`; the gear folds frames by `(setting, gt)`, the refused gesture's identity (an event key, never a time window), into one toast that lists the refusing kernels: "Triage model: not applied on web, api, this machine. A later pick (fable) is already in place." A frame without `gt` (an older kernel) keeps one toast per frame. Liveness of the folded toast is checked at lookup, so a dismissed or expired toast never absorbs a later frame; the dismissal treatment and timings are untouched.

## Compatibility

Additive on both sides. An older kernel sends no `settingsGt`, no `gesture` and no `gt` in the frame: the clock falls back to the wall clock, the toast shows without the action, and refusals are toasted one per frame. An older dashboard keeps stamping `Date.now()` and is refused exactly as today. No store format changes. Two gear documents on one page (feed pane and chat pane) each keep their own clock; each learns on its own open and from the frames it receives, and either can apply anyway. The judge-limit banner in feed.ts has no `/version` read of its own: with the gear never opened in that document it stamps with the wall clock and relies on the refusal plus Apply anyway.

## Tests

Every new test fails on the base commit without the code hunks (checked by reverting the source files and re-running: 10 pytest and 11 node failures for commit 1, 2 pytest and 10 node failures for commit 2, all in the new or changed tests).

Kernel, `tests/test_setting_gesture_order.py`: `VersionReportsEveryStoredStamp` (fresh stores read 0, each setter's stamp is reported, an unstamped apply reports its arrival time, the frames' setting names equal the report's keys), `ASkewedClockCannotLockTheStore` (refused at the skew, applied and propagated at `storedGt + 1`, no frame), the frame's `gesture` echo without `gt` and its `gt` on every gt-gated setting, and the wiring pin on the new signature.

Webview: `gesture-clock.test.ts` drives the module with `Date.now` stubbed (monotonic per store, learned + 1, a stepped-back clock, garbage ignored, the skew scenario end to end). `gear.test.ts`'s completeness pin now requires every `KERNEL_SETTING` emitter to stamp through the clock under its own store name and rejects a bare `Date.now()`. `setting-stale.test.ts` pins the new copy and the absence of the old claim, the learn, the whitelist, the action and its re-issue shape, and the fold's three-file wiring. `setting-stale-fold.test.ts` lifts the gear's toast block the way `gear-models-frame.test.ts` lifts its models block and drives frames through it. `federation-send-queue.test.ts` covers the breadcrumb and the mid-flush throw. `multi-kernel-merge.test.ts` covers the host stamp. The verbatim literal pins in `thinking-summaries.test.ts`, `file-view.test.ts`, `feed-limit-banner.test.ts` and `tests/test_kernel_update.py` follow the emitters.

Full suites on the branch: pytest 7334 passed (46 skipped), `npm test` 2808 passed, `npm run typecheck` and `npm run build` clean, bats 389 passed.

Not in scope: `/tunnels` rows do not lift peers' stamps; a remote's newer stamp reaches the dashboard through the refusal frame, and Apply anyway is the designed path from there.

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -633,6 +633,11 @@ def _version_info():
                          "commentModel": jd._state_str("comment-model", "session"),
                          "commentEffort": jd._state_str("comment-effort", "session"),
                          "commentFast": jd._state_str("comment-fast", "session")},
+            # every gt-gated store's last-applied gesture stamp (epoch-ms ints, nothing path-shaped):
+            # the gear stamps its next gesture above these instead of trusting the device clock.
+            # Top-level, not lifted into /tunnels rows — a remote's newer stamp reaches the dashboard
+            # through its settingStale frame, and Apply anyway is the designed path from there.
+            "settingsGt": _settings_gt(),
             "defaultDir": _tilde(_default_create_dir()),   # the resolved default new-session dir → the gear "Default directory" field
             "nativeDialogs": _native_dialogs()}   # whether Browse… can draw a dialog HERE → the gear drops the button when it can't
 
@@ -28070,7 +28075,44 @@ def _setting_kept_value(name):
     return jd._state_str(name, "")   # the judge-tier stores are bare value files
 
 
-def _tell_stale_gesture(client):
+# Every gt-gated store, by the name _setting_stale is called with — the vocabulary the settingStale
+# frame and the gear's STALE_LABELS already share, so /version's settingsGt speaks the same one.
+_GT_STORES = ("auto-nudge", "compact-suggest", "file-editing", "update-mode", "thinking-summaries",
+              "judge-model", "index-model", "judge-effort", "index-effort",
+              "distill-model", "distill-effort", "comment-model", "comment-effort", "comment-fast")
+
+
+def _setting_stored_gt(name):
+    """The last-APPLIED gesture stamp a store holds — _setting_kept_value's switch, for the stamp
+    beside the value. Absent file, absent field or garbled content all read 0 (nothing to outrank),
+    the same defensive shape the setters use for `prev`."""
+    if name == "auto-nudge":
+        return _gt_int(_auto_nudge_data().get("gt"))
+    if name == "compact-suggest":
+        return _gt_int(_auto_nudge_data().get("compactSuggestGt"))
+    if name == "update-mode":
+        return _update_mode_gt()
+    if name in ("file-editing", "thinking-summaries"):
+        try:
+            d = json.loads((jd.STATE / (THINKING_SUMMARIES_FILE if name == "thinking-summaries"
+                                        else "file-editing.json")).read_text())
+            return _gt_int(d.get("gt")) if isinstance(d, dict) else 0
+        except Exception:
+            return 0
+    return _judge_state_gt(name)   # the judge-tier stores keep theirs in a `<name>.gt` sidecar
+
+
+def _settings_gt():
+    """/version's `settingsGt`: every gt-gated store's last-applied gesture stamp, keyed by the
+    store name the settingStale frame uses (PR #879 follow-up). Gesture stamps are minted on the
+    DEVICE, so a clock running ahead used to stamp every store into the future and lock every
+    correctly-clocked device out until the skew elapsed; a dashboard that reads this on open stamps
+    its next gesture at max(Date.now(), gt + 1) instead of trusting its own wall clock
+    (ui/webview/gesture-clock.js). Epoch-ms integers only — fine for the auth-exempt route."""
+    return {n: _setting_stored_gt(n) for n in _GT_STORES}
+
+
+def _tell_stale_gesture(client, msg):
     """A gt-gated setter just refused: if that refusal was a stale STAND-DOWN (recorded on this
     thread by _setting_stale), answer the DELIVERING socket with a small settingStale frame — the
     same targeted _reply idiom the saveFile acks use, never a broadcast. Without it the refusal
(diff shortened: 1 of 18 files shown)


ksarma:
Merging as shipped. Two follow-ups and two nits below; none blocks.

Reviewed inline: The gesture clock (`ui/webview/gesture-clock.js`) stamps max(Date.now(), highest stamp seen for the store + 1). It learns from the new `/version` `settingsGt` field and from each `settingStale` frame's `storedGt`. All 17 emitters mint through it, under `gear.test.ts`'s completeness pin. The kernel's `settingsGt` report reads the same fields the setters compare against. All 14 stand-down branches call the two-arg `_tell_stale_gesture`, so the frame carries the refused gesture's `gt` and its echo without the stamp. Apply anyway is whitelisted by `STALE_TYPE`, so a frame can only re-issue the setting it names. The re-issue is a fresh click stamped above everything the page has seen, which is the event the ordering rule asks for. The fold keys on (setting, gt) and reads liveness from `parentNode` at lookup. `prefixInbound` stamps host on a remote kernel's `settingStale`; a local frame reads as this machine. The button sits on a node the fold never rebuilds. Its click bubbles to the container's delegated dismiss, so it is click-safe and acknowledged. `flushPending` deletes each entry as its send succeeds and halts with a flush-halt breadcrumb on a throw. `sendRemote` journals a sendqueue breadcrumb at queue time. Both directions stay compatible with older peers. Privacy, hostname, home-path and vocabulary scans of the 860 added lines are clean. The body's per-commit failure counts and file descriptions match the diff.

Findings:

- `ui/webview/gear.js:868`, follow-up: liveness is `parentNode` only. A same-gesture refusal arriving between the fade at 11 s and the removal at 12 s is written into a toast already at opacity 0 and never shown. Reproduced by firing the 11000 ms timer before a second frame. Fix: treat a toast carrying `.fade` as not live, or drop the `staleOpen` entry when the fade fires, with a test.
- `ui/webview/gear.js:837`, follow-up: neither the copy nor the Apply anyway label shows the refused value, only the kept one. In the frozen-tab case the pick is hours old, and one click sends it to every linked kernel. `m.gesture` carries the value, so render it and update the exact-string pins in `setting-stale.test.ts` and `setting-stale-fold.test.ts`.
- `ui/webview/gear.js:838`, nit: the copy calls the kept pick a later one. The kernel cannot make that recency claim under the skew this PR fixes; it holds only a larger stamp. Same copy follow-up.
- `ui/webview/gear.js:820`, nit: the Apply anyway button inherits the toast's click-to-dismiss title.

Red-first verification ran per commit in scratch worktrees. Commit 1's tests at `origin/main` give 10 pytest failures (`test_setting_gesture_order.py`, `test_kernel_update.py`); `gesture-clock.test.ts` fails to bundle because its module does not exist there; the other five webview files give 10 failures. Commit 2's tests at commit 1 give 2 pytest and 10 node failures, all in the changed tests, matching the body. At head the two pytest files pass 97/97 (also twice under `-n 4`) and the nine webview files 157/157. CI: all six required checks green at head.

Review run by the maintainer's automated review session; verified against head a167c31b.

PR #947, opened 2026-09-06
File viewer: the GitHub button says why when there is no link

ksarma:
## What breaks

When the kernel decides a viewed file has no GitHub link (not in a repository, not committed, no origin remote, or an origin that is not on GitHub), the `fileGitLink` reply carries an empty `url` and the viewer hides the button. From the page, an uncommitted file is indistinguishable from a broken feature. Two smaller defects sit on the same path:

- `git rev-parse --abbrev-ref HEAD` spells a branch that shares its name with a tag as `heads/<branch>`, and GitHub 404s the URL built from it.
- A repository with no `origin` remote takes the same branch as a non-GitHub origin, so the only verdict available would name a remote the repository does not have.

## Reproduction

In a repository with a GitHub origin:

1. Open an untracked file in the viewer. No GitHub button appears and nothing says why.
2. On branch `main`, run `git tag main` and open a tracked file. The button links to `.../blob/heads/main/...`, which 404s.
3. Open a tracked file on a branch that was never pushed. The link 404s with no hint that a push would fix it.

## The fix

**Kernel.** `_file_github_link(raw, sid, check_origin=True)` returns `(url, reason)`, and the `fileGitLink` reply carries `reason` beside `url`. An empty url comes with one of a fixed set of plain phrases: `not in a git repository`, `not committed (untracked file)`, `not committed (no commits yet)`, `not committed (staged only)`, `no origin remote`, `the origin remote is not on GitHub`, and, for a relative path that no session directory resolves, `relative path with no session directory to resolve it against` (no repository check ran, so the verdict does not claim one). A file that is in the index but not in HEAD's tree gets the staged-only verdict from `git cat-file -e HEAD:<path>`, since `ls-files` reads the index and the URL would name a path GitHub has on no ref. `_file_github_url` keeps its url-only contract and never queries the network.

A branch that is not on origin keeps its url (it is the right one once pushed) and carries `branch X is not on origin`. `_origin_has_branch` reads the local tracking ref `refs/remotes/origin/X` first, which costs nothing; only its absence pays one `git ls-remote --heads origin refs/heads/X`, bounded at 3 s with `GIT_TERMINAL_PROMPT=0`. `_git_net_out` runs that query in its own session and kills the whole process group at the deadline: `subprocess.run`'s timeout kill reaches git alone, and the ssh git had spawned kept its TCP connect open for minutes after the viewer had been told "could not check". The pattern is the full ref and the match is exact, because ls-remote patterns match a ref's tail (`main` also matches `refs/heads/x/main`).

The answer is memoized per (repository, branch) with no clock, and the memo keeps only what a local event can later contradict: a False, and only where the clone's fetch refspec maps the branch onto `refs/remotes/origin/X`. `_origin_tracks` reads `remote.origin.fetch` for that: the default `+refs/heads/*:refs/remotes/origin/*` qualifies, a `--single-branch` clone's exact refspec qualifies for its one branch alone, a negative refspec excludes, the first positive match decides (git's push updates that refspec's tracking ref alone), a source-only refspec is skipped, and any shape it does not understand reads as untracked. Where the branch is tracked, a push from this clone or a fetch writes the tracking ref, the free check sees it and drops the entry. A True with no tracking ref has no drop event (a deletion on GitHub writes nothing locally, and `fetch --prune` has no ref to prune), and a False in a `--single-branch` clone would outlive the push, which writes no tracking ref there; neither is stored, and each open asks, with concurrent opens of one key still sharing one query. "could not check" is not memoized either. The verdict is therefore as fresh as the clone's own refs: a branch deleted on GitHub reads as present until `git fetch --prune`, and one pushed from another clone reads as absent until a fetch. The docstring and the guide state the same rule.

The branch name comes from `git symbolic-ref -q HEAD`, so a same-named tag no longer changes the ref to `heads/<branch>`, and the query works on any git: `branch --show-current` needs 2.22, and below that its failure silently gave every file a commit-sha URL. A repository with no origin remote reads `no origin remote`.

**Viewer.** The `github-link` registry entry mounts one unit. Until the kernel answers it holds a placeholder: a disabled button titled `Checking GitHub…` with the pane's three pulsing loader dots where the caption will go, and `aria-busy` on the unit, so a check that waits on `ls-remote` reads as a wait rather than as the old no-button state. A socket drop while the ask is out loses the reply, so the shim's reconnect event re-asks with the same reqId. With no url it renders a real disabled `<button>`, with the reason as a visible caption beside it and in the tooltip (`No GitHub link: ...`). With a url and a reason it renders a dashed anchor with the reason as its caption and aria-label. With a url alone it renders the anchor as before. The control is a disabled button because an hrefless anchor has no role for assistive tech to read; the reason is a caption as well as a tooltip because touch has no hover and a disabled button takes no focus. The caption wraps inside a bounded width (18em) rather than truncating: nothing in the unit takes a tap, click or focus, so a cut sentence would have had no way to finish. Both stylesheets carry the same rules.

**Compatibility.** The wire change is additive. An older viewer ignores `reason`. An older kernel's reasonless empty reply renders as a disabled button with the caption `this kernel predates link reasons; restart it after updating`.

**Docs.** The guide's "Reviewing a document" paragraph describes the three states, lists the reasons a file can have no link, and states what the branch check can and cannot see.

## From review

Folded before the merge, from the review at 7c064ac4:

- `_origin_has_branch` memoized a True while the tracking ref was absent, with no event to drop it, and a False that a `--single-branch` push could not contradict. Only a False is memoized now, and only under a refspec that maps the branch onto its tracking ref (`_origin_tracks`); everything else is asked on each open. The docstring, the guide and this body state the refresh rule as it is. Three tests cover the two states and the refspec shapes.
- The caption wraps instead of truncating at 18em with the full text in `title`; the exact CSS pin is updated and the caption's redundant tooltip is gone. Measured in headless Chromium at a 13px base: the two branch-bearing phrases and the 60-character fallback take two lines of a 192px box, none clipped; the short phrases stay on one line.
- `git symbolic-ref -q HEAD` replaces `git branch --show-current`. A test stands in an old git with a PATH wrapper that rejects `--show-current`, and checks the branch URL and the detached-HEAD sha URL under it.
- An unresolvable relative path is captioned `relative path with no session directory to resolve it against`, not `not in a git repository`; the guide's list includes it.
- The red-first count for the two UI test files below was wrong in the earlier body (8 of 12); it is 10 of 12.

Second round, from the review at fbfad515, folded by file into the same three commits:

- Kernel: a file staged on a branch with commits got a live URL and an empty reason, so the viewer drew an enabled button that 404s. After the sha check, `git cat-file -e HEAD:<path>` decides, and the verdict is `not committed (staged only)` (`GH_STAGED`). A test mirrors the unborn-branch one and checks that the committed neighbour still links and that the commit flips the verdict.
- Kernel: `_origin_tracks` accepted any matching fetch refspec, but git's push updates the tracking ref of the first match alone, so a `refs/remotes/other/*` line ahead of the default kept a memoized False the push could not contradict. The first positive match decides now; a source-only refspec is skipped as git skips it; a negative match still excludes wherever it sits. One test runs the two-line config end to end (the fixture's push writes `other/wip` and not `origin/wip`, and the next open sees the push); the refspec table gains both orderings and the source-only case.
- Viewer: the unit was hidden until the reply, so an open that waited on `ls-remote` showed no GitHub control for up to 3 s. It mounts a placeholder now (a disabled button, the loader dots, `aria-busy`), and the reply replaces it; the `[hidden]` rule and its pin are gone, both sheets gain `.fileview-gh-dots`, and the reconnect event re-asks a question whose reply a socket drop lost, so the placeholder cannot outlive its wait. The DOM test checks the placeholder before every state and the re-ask.
- Guide: the reasons list names a file staged but in no commit, and the paragraph says the button waits with pulsing dots while the check runs.

## Tests

`tests/test_file_github.py` (38 tests): a reason for every no-link verdict, the relative-path verdict among them; a file staged on a branch with commits and its committed neighbour; the same-named-tag case on both paths; the no-origin verdict; an old git without `branch --show-current` still names the branch and still links the sha when detached; the origin check served by a local bare repository through a stand-in ssh, so no test reaches the network; a pushed branch asks origin nothing; a missing tracking ref falls to ls-remote; a slashed branch is matched whole; an unreachable origin reads "could not check"; a slow origin is cut off and the spawned ssh dies with the cut; a detached sha is never checked; the memo, the in-flight sharing, the tracking-ref flip and the unanswered re-ask; a branch on origin that this clone never fetched is asked each time and its deletion on origin is seen on the next open; a `--single-branch` refspec is asked each time and the push is seen on the next open; a second namespace ahead of the default takes the push as git does and the next open sees it; the refspec shapes `_origin_tracks` reads, in both orders; the url-only caller never asks; the wire op carries `reason`. For the origin-backed tests the kernel's git runs with `GIT_CONFIG_GLOBAL=/dev/null` and `GIT_CONFIG_NOSYSTEM=1` (git >= 2.32; one test self-skips below that), and one test shows the pin is load-bearing against a global `url.insteadOf` rewrite. The process-group test reads `/proc` to treat a zombie as gone, so that check is Linux-specific; the pytest job runs on ubuntu.

`ui/webview/github-link.test.ts` (12 tests): the placeholder at mount, the three states, the reasonless fallback, the stale-reply guard and the reconnect re-ask run against a small DOM stand-in; source pins cover the CSS (the caption rule, and that it carries no `nowrap`, `ellipsis` or `overflow: hidden`) and the kernel side. `fileview-parity.test.ts` covers the new rules in both sheets.

Before the change, 29 of the 38 Python tests fail against the old kernel (`_file_github_link` absent, the `heads/main` URL, the no-origin phrase, the memo's module state, the staged-only verdict), and 12 of the 13 tests in the two UI test files fail against the old viewer (measured in a scratch tree at fd2135c0 with the PR's test files).

Results: `pytest tests/test_file_github.py` 38 passed (with and without xdist); full `pytest -n 8` 7407 passed, 46 skipped, 0 failed. vscode-extension: `npm run typecheck` clean, `npm test` 2838 passed.

Tier: feature

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -28,7 +28,20 @@ passage, and repeat — each staged note remembers its quote and its place. **�
 everything you staged along with whatever is in the box, so the session applies the lot
 in one pass, and you never copy a line out of the document by hand. The line in each
 label is checked against the file at the moment you select, so numbers that moved under
-you are caught rather than quietly carried.
+you are caught rather than quietly carried. The title bar's **GitHub ↗** button opens the
+file on GitHub. While the check runs, the button waits dimmed with pulsing dots beside it.
+When there is nothing to open, the button stays in place, dimmed, and a caption beside it
+says why (the file is not in a git repository or not committed — untracked, staged but in no
+commit, or on a branch with no commits yet — the repository has no origin remote, its origin
+is not on GitHub, or the path is relative and no session's directory resolves it); the
+button's tooltip repeats the reason. A file on a branch that is not on origin keeps its link,
+drawn with a dashed border, and the caption says the branch is not on origin yet. That check
+trusts your clone's own refs: a branch deleted on GitHub reads as present until `git fetch
+--prune`, and one pushed from another clone reads as absent until a fetch. A branch that has
+never been pushed is asked of origin once and the answer kept until a push or fetch from this
+clone writes its tracking ref; where nothing local could refresh the answer (a
+`--single-branch` clone, or a branch on origin this clone has not fetched) origin is asked on
+each open.
 
 ### The feed
 
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -28618,6 +28618,41 @@ def _git_out(args, cwd, timeout=5):
         return None
 
 
+def _git_net_out(args, cwd, timeout, env):
+    """_git_out for the one query that leaves the machine (ls-remote): git runs in its OWN session, and
+    the deadline kills the whole process group. subprocess.run's timeout kill reaches its direct child
+    alone, and the ssh git had spawned sat in its TCP connect for minutes after the viewer had already
+    been told "could not check" (reproduced 2026-09-05). A fresh session also has no controlling
+    terminal, so an ssh that wants a passphrase fails instead of waiting for one."""
+    try:
+        p = subprocess.Popen(["git", "-C", cwd] + args, stdin=subprocess.DEVNULL, stdout=subprocess.PIPE,
+                             stderr=subprocess.DEVNULL, text=True, start_new_session=True, env=env)
+    except OSError:
+        return None
+    try:
+        out, _ = p.communicate(timeout=timeout)
+    except subprocess.TimeoutExpired:
+        try:
+            os.killpg(p.pid, signal.SIGKILL)         # start_new_session: the child's pid is its pgid
+        except (ProcessLookupError, PermissionError):
+            pass
+        try:
+            p.communicate(timeout=5)
+        except Exception:
+            try:
+                p.kill()
+            except Exception:
+                pass
+        return None
+    except (OSError, subprocess.SubprocessError):
+        try:
+            p.kill()
+        except Exception:
+            pass
+        return None
+    return out.strip() if p.returncode == 0 else None
+
+
 # origin remote → the https://github.com/<owner>/<repo> base, for the spellings git actually writes:
 # scp-like ssh, ssh:// (with an optional port, and GitHub's own documented SSH-over-HTTPS host
 # ssh.github.com), https (with an explicit :443), .git suffix and all. Anchored, so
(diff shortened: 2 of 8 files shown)


ksarma:
Recommend merging after the comment window. Two findings need follow-up; neither blocks.

Reviewed inline: `_file_github_link` returns a (url, reason) verdict from a fixed phrase set. The `fileGitLink` reply carries `reason` as an additive field, which an older viewer ignores. `_file_github_url` stays url-only with no network query. The branch name now comes from `git branch --show-current`. A shell probe with a tag named after the branch confirmed the old `heads/` spelling is gone. A repository with no origin gets its own verdict. `_origin_has_branch` reads the tracking ref first. Otherwise it runs one `ls-remote` with a full-ref exact match, a 3 s cap, `GIT_TERMINAL_PROMPT=0`, and a process-group kill in `_git_net_out`. The per-(repo, branch) memo shares one in-flight query and never stores an unanswered one. The viewer mounts one unit per open: a disabled button, a dashed anchor with caption, or a plain anchor. No click handler exists for a re-render to lose. The parity test pins both sheets byte-equal. Privacy and vocabulary scans of the 740 added lines and the commit messages found nothing. The label fits: a self-contained capability with an additive wire change, plus two small fixes that land with tests.

- `kernel/kernel.py:28520`, follow-up. A True answer memoized while the tracking ref is absent has no drop event. The ref is absent when the branch was pushed from another clone, or in a single-branch clone. The only pop runs when `refs/remotes/origin/<branch>` is present. So after the branch is deleted on GitHub, `fetch --prune` prunes nothing. The memoized True then serves a 404 link with no caption until the kernel restarts. The docstring, `docs/guide.md:37` and the body say `fetch --prune` refreshes it. That holds only when the tracking ref existed at some point. Reproduced with a local bare origin. In a single-branch clone the same gap keeps a memoized False after the push, so a working link is captioned as not on origin. Memoize only what the tracking ref can later contradict, or state the limit and the restart recovery, with a test for the state.
- `ui/webview/styles.css:3298`, follow-up. The caption is nowrap at 18em, with the full text only in `title`. The five fixed phrases fit. Both branch-bearing phrases and the 60-character reasonless fallback truncate: 263px and 306px in a 192px box in headless Chromium. Nothing in the unit takes a tap, click or focus, so touch and keyboard users see a cut sentence. That is the dead end the progressive-disclosure rule forbids, and the PR's own reason for adding the caption. Let it wrap, or make the unit tap-to-expand.
- `kernel/kernel.py:28574`, nit. On git below 2.22, `branch --show-current` exits non-zero and `_git_out` returns None. Every file then silently gets a commit-sha URL with the origin check skipped, where main gave the branch URL. `git symbolic-ref -q HEAD` gives the same tag-proof answer on any git.
- `kernel/kernel.py:28544`, nit. An unresolvable relative path is captioned as not in a git repository, a check the kernel did not run. The guide's list of reasons at `docs/guide.md:33` omits that verdict.

Locally: `tests/test_file_github.py` 32 passed, alone and six times under xdist. `github-link.test.ts` plus `fileview-parity.test.ts` passed 12 of 12. Against `origin/main` the same files fail 23 of 32 and 10 of 12 (the body says 8 of 12). CI: all six required checks green at head.

Review run by the maintainer's automated review session; verified against head 7c064ac4.

ksarma:
Recommend merging after the comment window. One follow-up remains to fold, before or after the merge at the author's call.

Reviewed inline: all five folds from the review at 7c064ac4 are in the diff, each with a test that fails without it. The memo keeps only a False, and only when `_origin_tracks` finds a fetch refspec mapping the branch onto `refs/remotes/origin/<branch>`. The free tracking-ref check drops the entry once the ref appears. A True with no tracking ref is re-checked on each open (two opens, two ls-remotes); a deletion on origin shows on the next open. I ran `_refspec_match` and `_origin_tracks` against the default wildcard, a single-branch exact refspec, several lines, a negative refspec, a source-only refspec, a wildcard not at the end, a mirror refspec and a two-star pattern. The match is git's own prefix, suffix and length test. Every shape either tracks correctly or reads as untracked, and an untracked read asks origin. The caption wraps at 18em, right-aligned, with no nowrap, ellipsis or overflow hidden in either sheet, and carries no title. `git symbolic-ref -q HEAD` replaces `branch --show-current`, tested under a PATH wrapper that rejects the option. An unresolvable relative path returns GH_NO_BASE before any git call. The per-open ls-remote for a branch on origin with no tracking ref is the right trade-off. It is bounded at 3 s, off the recv loop, and shared across concurrent opens. A memoized True had no local event to drop it.

- `kernel/kernel.py:28769`, follow-up: a file that is staged but in no commit, on a branch with commits and a tracking ref, gets a live URL and an empty reason. The viewer then renders a plain enabled button that 404s on GitHub. Only the index is checked. It is pre-existing at main, but it is the state this PR set out to make legible. The guide also says the caption covers a file that is not committed. One local call after the sha check fixes it: a failing `git cat-file -e HEAD:<rel>` gives a staged-only verdict, with a test mirroring the unborn-branch one.
- `kernel/kernel.py:28683`, nit: `_origin_tracks` accepts any matching refspec, but git's tracking-ref update on push takes the first match. With a `refs/remotes/other/*` line before the default one, a memoized False survives the push until the next fetch. The config is pathological; returning on the first non-negative match would mirror git.
- `ui/webview/file-view.ts:171`, nit: the unit stays hidden until the reply, so an open that waits on ls-remote shows no GitHub control for up to 3 s. A dimmed placeholder or the loader would distinguish a slow check from the old no-button state.

Locally at head: `tests/test_file_github.py` 36 passed (single process and under xdist), the two UI test files 12 of 12, vscode-extension npm test 2834 passed. In a scratch worktree at origin/main with the PR's test files, 27 of 36 Python tests and 10 of 12 UI tests fail, as the body states. All six required checks are green at this head.

Review run by the maintainer's automated review session; verified against head fbfad515.

ksarma:
Recommend merging after the comment window. One item to fold, before the merge or as a follow-up.

Reviewed inline: the three folds since fbfad515 and the earlier-round invariants. The staged-only verdict comes from `git cat-file -e HEAD:<path>`, after the sha check and before the URL is built. Its test covers the verdict, the committed neighbour, and the flip on commit. `_origin_tracks` takes the first positive fetch refspec covering the branch, skips source-only lines, and honours a negative line anywhere, as git's query_refspecs does. The two-line config test shows the push landing in the other namespace and the next open seeing it. The viewer mounts a disabled button with the pane's own loader dots and aria-busy, replaces it once on the reply, and hangs no click handler on any node. It is built once per open; no per-push render touches it. The re-ask rides romp:wsup, which the shim dispatches once per reconnect from onopen with the same reqId. gitHooks is cleared on reply, close and replace-open, so a settled open asks nothing more. The one possible duplicate, an ask queued during the outage, is absorbed by the kernel's in-flight dedupe and the reqId guard. The memo still keeps only a False under a tracking refspec; a True with no tracking ref is asked per open. The caption wraps, symbolic-ref names the branch, and the relative-path verdict returns before any repo check.

- `kernel/kernel.py:28778`, follow-up: the ls-remote env sets only `GIT_TERMINAL_PROMPT=0`, so the child inherits `GIT_ASKPASS`, `SSH_ASKPASS` and `core.askPass`. If the kernel was started from a terminal exporting an askpass program, a credential-wanting private origin pops a GUI prompt on every open of that file. The 3 s kill removes the prompt (reproduced against a local 401 stand-in). The `_git_net_out` docstring says ssh fails instead of waiting; that does not hold on this path. `GIT_ASKPASS=""` plus `SSH_ASKPASS_REQUIRE="never"` in the same env dict closes it, with a test on the env. VS Code's askpass returns a token silently for a signed-in user, so that path would then read "could not check". The docstring should say which behaviour was chosen.
- `ui/webview/github-link.test.ts:182`, nit: the re-ask assertion accepts any number of asks per reconnect, because each test body's initFileView adds another listener. A second listener inside initFileView would pass it.

Locally at head: `tests/test_file_github.py` 38 passed; `github-link.test.ts` plus `fileview-parity.test.ts` 13 of 13. At origin/main with the PR's test files, 29 of 38 and 12 of 13 fail, matching the body. All six required checks are green at this head.

Review run by the maintainer's automated review session; verified against head a5191671.

PR #948, opened 2026-09-06
Index judges: a bare alias is checked against the model the CLI actually served, and the tier-pick pin that bound nothing is gone

ksarma:
This picks up the two no-gate notes from the #880 review: the bare `haiku` alias pinned to 4.5 in `_ALIAS_HEAD` with no drift signal if the catalog moves it, and `test_index_tier_with_no_model_resolves_the_tier_pick` passing against main too. Two commits, one per note.

## What breaks

`_adaptive_thinking` places a bare family alias (`haiku`, `sonnet`, ...) by `_ALIAS_HEAD`, a table copied from the 2.1.257/258 binaries, and nothing checks the table against what the CLI serves. The CLI at 2.1.263 resolves alias targets at run time from a published catalog and falls back to its compiled table (its own log lines name a `[publishedCatalog]` fallback to compiled behavior), so the head can move with no CLI upgrade. When it does, the index tier keeps choosing the lever for the old version, with nothing in the log. For `haiku` that is the failure #880 fixed, returning through the alias: a Haiku the CLI treats as adaptive gets no `--effort`, and if it also rejects `thinking: disabled`, runs full-cost thinking silently. The fold narrowed this (the variable rides regardless), but the effort default and the lever line still answer from the table.

## Reproduction

Stub `subprocess.run` to return `{"result": "a caption", "usage": {}, "modelUsage": {"claude-haiku-4-6-20260301": {"outputTokens": 15}}}` and call `_judge_run("haiku", ..., tier="index")` twice. Before this change: no `--effort` on either call, nothing on stderr beyond the lever line, and `_adaptive_thinking("haiku")` still False. `AliasHeadDrift.test_a_moved_alias_is_announced_once_and_the_lever_follows` is that reproduction; it fails before the change.

## The fix

Every judge call already receives the CLI's answer. The `-p --output-format json` envelope carries `modelUsage`, keyed by the model id that actually served, beside the `usage` and `total_cost_usd` that `_log_judge_usage` reads; `_judge_run` ignored it. Confirmed live on 2.1.263: a `--model haiku` call returns one `modelUsage` key, `claude-haiku-4-5-20251001`, with `outputTokens` in it.

`_note_served_model(model, wrap)` runs after every successful claude-engine call. When the call's model was a bare alias, it reads the same-family entry of `modelUsage` (the one with the most output tokens: the model that wrote the reply), records the served version in `_ALIAS_SERVED`, and `_adaptive_thinking` consults that ahead of `_ALIAS_HEAD` from then on. When the served version differs from the table, one stderr line per (alias, served id) names both versions and says the lever follows the served one; the existing per-model lever line then re-announces the new lever on the next call. Versioned ids are never remapped. Nothing changes while the table agrees, which it does on 2.1.263 (`latest_per_family`: fable 5.1, opus 5, sonnet 5, haiku 4.5).

Design points a reviewer may want to weigh:

- The first call on a moved alias still runs under the table's answer, because the envelope arrives after the call; the second call is corrected.
- An envelope naming no same-family model is quiet. Absence carries no evidence of drift, so the table's answer stands exactly as before; a test pins this. A once-per-process line there is a three-line change, but `IndexTierLever`'s empty-stderr assertion would then need its stub to carry `modelUsage`.
- If you prefer signal only, with no self-correction, the `_ALIAS_SERVED.get(fam) or` lookup in `_adaptive_thinking` is the one line to drop; the drift line stands on its own, and the lever-follow assertions go with it.
- Every tier feeds `_ALIAS_SERVED` (a triage call on `sonnet` reveals its head just as well); only the index lever reads it.
- The check is best-effort like `_log_judge_usage`: a malformed entry cannot cost the reply.

## The second note

`_judge_env` reads no model since the fold made the variable unconditional, so a test named for `_judge_env("index")` resolving the tier's pick has no behavior to bind: it asserted the default pick's name and the unconditional variable, both true on any main. It is removed rather than rewritten; the tier-pick contract is pinned where it lives (`ModelTiers` for the captioner/archiver calling `_index_model()`, `IndexTierLever` for what `_judge_run` does with the model).

The prose around it still described the shape the review reversed. `_judge_env`'s docstring said the variable is NOT set on adaptive models and that `None` resolves the tier's pick; the `JudgeEnv` and `IndexTierLever` class docstrings said effort replaces the variable; the stranger's stderr line said the variable is left unset. All four now state the unconditional contract: the variable rides every index call, honored where the CLI takes `thinking: disabled` and a harmless no-op where it drops it, with `--effort` landing beside it. The `model` parameter stays, since `_judge_run` and the existing pins pass it and dropping it would rewrite the fold's own tests for no behavior gain; say the word and a follow-up drops it.

## Tests

`tests/test_judge.py`, new class `AliasHeadDrift` (subprocess stubbed; argv and env captured, read with `.get()` so a failure never prints the environment):

- an agreeing served version is quiet, and recorded
- a moved alias is announced once; the first call runs under the table, the second carries `--effort low` with the variable still riding and a fresh lever line, the third is silent
- the reply writer decides between two same-family ids
- another family's entry never speaks for `haiku`
- a versioned pin is never remapped
- every tier feeds the record; only the index lever reads it
- an envelope naming no model is quiet
- a malformed entry cannot cost the reply

Red before the change: all eight error on the missing `_ALIAS_SERVED`; with only the two module-level names added, the five behavioral cases fail (no drift line, nothing recorded, no `--effort` on the second call) and the three quiet pins pass. `tests/test_judge.py` ran three times under `-n 8` (the new class clears `_ALIAS_SERVED` in setUp and tearDown, so `EffortCapability`'s `haiku -> False` pin never sees a drifted head in a shared worker). Full pytest suite green twice. No shell or UI surface is touched.

## Scope

First-party auth, as in #880. Bedrock/Vertex alias targets are not modeled; the served id would still be read correctly there, but the denylist and floors remain first-party. A CLI older than the `modelUsage` field yields no signal and the previous behavior.

Tier: fix

### DIFF
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -194,8 +194,14 @@ _EFFORT_FLOOR = {"fable": (5, 0), "mythos": (5, 0), "opus": (4, 6), "sonnet": (4
 #   family -> first (major, minor) the CLI treats as adaptive; below it is its denylist. The catalog knows
 #   no Haiku past 4.5, so haiku's floor marks where the denylist ends, not a version that exists.
 _ALIAS_HEAD = {"fable": (5, 1), "opus": (5, 0), "sonnet": (5, 0), "haiku": (4, 5)}   # the catalog's aliases block
+#   as read from the 2.1.257 / 2.1.258 binaries. 2.1.263 resolves alias targets at run time from a published
+#   catalog, falling back to its compiled table, so this can go stale with no CLI upgrade: _note_served_model
+#   checks it against the id each call actually served, and _ALIAS_SERVED outranks it once a call has answered.
 _MODEL_FAMILIES = tuple(_EFFORT_FLOOR)
 _UNKNOWN_MODEL_LOGGED = set()   # ids already announced as unplaceable (one stderr line each)
+_ALIAS_SERVED = {}              # family -> (major, minor) the CLI served for the BARE alias this process, read off
+                                #   result envelopes (_note_served_model); _adaptive_thinking asks it before the table
+_ALIAS_DRIFT_LOGGED = set()     # (family, served id) pairs already announced as off the table (one stderr line each)
 
 
 def _model_family_version(model):
@@ -233,7 +239,8 @@ def _model_family_version(model):
 def _adaptive_thinking(model):
     """True when `model` runs adaptive thinking under CLI 2.1.257 / 2.1.258 — the models whose index-tier
     cost lever is `--effort` (_EFFORT_FLOOR: Fable and Mythos, every version; Opus >= 4.6; Sonnet >= 4.6;
-    a bare alias is the version the catalog resolves it to, _ALIAS_HEAD — `haiku` is Haiku 4.5, so False).
+    a bare alias is the version the CLI has been SEEN to serve for it this process — _ALIAS_SERVED, read off
+    the result envelopes — else the table's head, _ALIAS_HEAD: `haiku` is Haiku 4.5, so False).
     False is the CLI's denylist — Haiku 4.5 and older, Sonnet 4.5 and older, Opus 4.5 and older, every
     claude-3 — the models where MAX_THINKING_TOKENS=0 is honored and is the lever. An id this cannot
     place — no family it knows, or a family with no readable version — gets the CLI's own answer for an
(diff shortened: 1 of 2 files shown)


ksarma:

Reviewed inline: `_note_served_model` runs only in the claude-engine success branch of `_judge_run`, after `_log_judge_usage` and inside try/except. A malformed envelope therefore cannot drop the reply. It reads `modelUsage`, which CLI 2.1.263 emits keyed by the served model id. It takes the same-family entry with the most output tokens and records the version in `_ALIAS_SERVED`. It writes one stderr line per (family, served id) when that version differs from `_ALIAS_HEAD`. `_adaptive_thinking` reads the record ahead of the table. Versioned ids are never remapped. An envelope naming no same-family model leaves the table's answer unchanged. The 2.1.263 compiled table matches `_ALIAS_HEAD` for all four families, so nothing changes today. The second commit's claim holds: at main `_judge_env` branches only on tier and auth and never reads `model`. The removed test asserted two facts true on any main. Both are still pinned: the default pick in `ModelTiers` and `tests/test_kernel_judge_model.py`; the unconditional variable and the model-less call in `JudgeEnv`. The reworded stranger stderr line now matches the code.

Two non-blocking follow-ups:
- `IndexTierLever.setUp` and `EffortCapability` clear `_LEVER_LOGGED` but not `_ALIAS_SERVED`, so their haiku pins depend on `AliasHeadDrift`'s tearDown. One `clear()` each makes them self-sufficient.
- An envelope with no `modelUsage` field is silent, so a stale table would stay trusted with nothing logged. A once-per-process line there would satisfy the loud-failure rule.

Red-first verified: the PR's test file over main's code fails all eight `AliasHeadDrift` tests on the missing `_ALIAS_SERVED`. With only the two module names added, five fail on substance and three pass, as the body says. At head `tests/test_judge.py` is 514 passed serially and twice under xdist, plus an adverse-ordering run of the drifting tests ahead of the haiku pins. All six required CI checks are green at head.

Review run by the maintainer's automated review session; verified against head e6edcab8.

PR #949, opened 2026-09-06
Model pickers: the learned-version scan re-parses a reg only when its file changed

ksarma:
`_learned_versions` re-read and JSON-parsed every session reg under `STATE/sdk` on every call, and it sits on the pickers' hot paths. This change caches the reg scan per file, keyed on the file's content-stat, and keeps the derivation against the catalog per call. Follow-up to the note on #882.

## What it cost

Each call ran one glob plus one read and one JSON parse per reg. The callers:

- `GET /models`: once per picker open and per models-frame re-fetch (every open picker re-reads on `_models_changed`).
- Every pick: `_note_model_pick` asks `_version_family`, which short-circuits only on catalog ids and otherwise falls through to the scan. The bare family alias a family click sends took that path on every click.
- The typed `/model` vouch (`_vouched_model`), the judge tier setters (`_judge_model_values`), and the SDK loop's refusal hook (`_model_pick_refused`).

Reproduction at 2b9db2be: with three regs under `STATE/sdk`, a second `_learned_versions()` call reads all three files again. `test_a_warm_scan_re_reads_no_reg_whose_file_is_unchanged` counts the reads and expects zero; at that tip it sees three.

## The fix (commit 1)

The scan moves into `_reported_model_ids`, which caches each reg's reported `liveModelId` keyed on `(mtime_ns, size, inode)`. That is the key `list_regs`' `_REG_CACHE` already uses for the liveness sweep: `write_reg`'s `os.replace` creates a new inode per write, so a same-size same-instant rewrite still misses. Per call: one `scandir` plus one `stat` per reg.

Only the reported string is cached, never the derived rows. The derivation against the live catalog snapshot (the `_catalog_lock` snapshot of `_VERSION_FAMILY` and labels, the dateless-alias fold, `_note_unknown_model`, the one-time stderr announce) still runs every call, so both T222 properties hold: a catalog that catches up drops the `learned` mark from an id already scanned, and a sighting after an inflight refresh lands still asks for its own refresh (the `_note_unknown_model` docstring relies on the list being re-derived per read).

Failure semantics are unchanged. An unreadable, non-JSON or non-dict reg, or one without `liveModelId`, contributes nothing. A failed read or parse is not cached, so the next call retries it. Regs that vanished leave the cache on the next scan. Scan order stays reg-name order, as the sorted glob was.

Two placement choices. It stays in kernel.py rather than routing through `list_regs`, because the SDK backend module loads lazily inside `_sdk_locked()` and the picker paths (and their tests) run without a backend constructed. The cache is an unlocked module dict: get/set/pop are atomic, two threads racing on a changed file store equal tuples, and holding `_catalog_lock` around file I/O would stall the catalog refresh thread.

## Commit 2, separable

`_version_family` returns `''` before scanning when `_model_id_parts` rejects the value. A learned row's value is always a first-party version id, so an alias, `default` or a typo can match none. The result is identical for every input (a `[1m]`-tagged or dated spelling still passes the parts check, still scans, and still misses the dateless learned values); only the scan is avoided, so the family-click path no longer touches the regs. Drop this commit if you would rather keep the lookup's shape.

## Tests (red first at 2b9db2be)

In `tests/test_model_versions.py`, class `LearnedVersions`:

- A warm second scan performs zero reg reads and returns the same rows (three reads at the previous tip).
- A rewritten reg (new inode, via tmp + `os.replace`, the `write_reg` shape), a new reg and a deleted reg are each reflected on the next call; the deleted one's cache entry goes with it.
- A garbled and a non-dict reg are skipped and `/models` still answers with the learned row; the garbled one is retried on the next call, the parsed one stays warm.
- Commit 2: a bare alias, `default` and a typo resolve to `''` without a call into the learned-version scan; a version-shaped value still consults it.

Existing learned-version, catalog-union and refresh-dedup tests are untouched and green. Full `tests/` sweep: 7327 passed, 50 skipped. bats: 389/389. Synthetic regs only (placeholder sids, session name `web`).

## Not included

`_model_alias_boot_pass`'s own one-time reg pass, `_thread_reg`, and anything in `sdk_backend.py`.

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -893,6 +893,18 @@ _VERSION_FAMILY = {v["value"]: fam for fam, vs in MODEL_VERSIONS.items() for v i
 MODEL_PICKS_FILE_NAME = "model-picks.json"   # {family: full-id} — a viewer pref all surfaces read, like colormap
 _model_picks_lock = threading.Lock()   # its read-modify-writes run on the WS-handler, producer AND SDK-loop threads
 _learned_announced = set()   # ids already announced on stderr as outside the catalog (once per process)
+# The learned-version scan runs on the pickers' hot paths — every /models read (one per picker open and per
+# models frame), every pick (_note_model_pick → _version_family falls through to it for any value the
+# catalog does not list, the bare alias a family click sends included), the typed /model vouch, the judge
+# tier setters and the SDK loop's refusal hook — and re-opening and JSON-parsing every reg per call was the
+# cost. The file is the truth, so the key is exactly what changes when the file does: (mtime_ns, size,
+# inode), the key list_regs' _REG_CACHE uses (write_reg's os.replace mints a new inode, so a same-size
+# same-instant rewrite still misses). Only the one reported string is cached, never the derived rows: the
+# derivation must run against the LIVE catalog snapshot every call (a catalog that catches up drops the
+# mark; _note_unknown_model must see each sighting). Unlocked on purpose: dict get/set/pop are atomic, two
+# threads racing on a changed file both parse it and store equal tuples, and holding _catalog_lock around
+# file I/O would stall the catalog refresh thread.
+_learned_reg_cache = {}   # str(path) -> ((mtime_ns, size, ino), liveModelId or "") — see _reported_model_ids
 # Bumps on every pick-memory change and every catalog growth; rides the models frame (_models_changed) AND
 # the /models payload, so a picker can drop a response older than one it has applied. Seeded from the clock
 # rather than 0: the counter is per process, and a kernel restart must never hand a page that kept its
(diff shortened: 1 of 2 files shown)


ksarma:

Reviewed inline: the new `_reported_model_ids` keys each reg on `(st_mtime_ns, st_size, st_ino)`, the triple `list_regs`' `_REG_CACHE` already uses. Every production reg writer (`write_reg`, `_atomic_write`, the codex registry) publishes through `os.replace`. A rewrite creates a new inode, so a same-size same-second rewrite still misses. Deleted regs are pruned on every scan. A truncated or garbled file changes size and fails the parse. A failed read or parse pops the entry, so the next call retries it. Each call stats before it reads. A race can leave an older key with newer content, corrected by the next stat. It can never leave a current key with older content, so no stale parse is served. Only the reported `liveModelId` string is cached. The catalog snapshot, the dateless-alias fold, `_note_unknown_model` and the one-time announce still run per call. The existing catalog-union test pins that on a warm cache. Commit 2's early return in `_version_family` is return-value equivalent: every learned value is a version id `_model_id_parts` accepts. The scandir filter and name sort reproduce the old sorted glob. The probe that checked this included a dotfile, a bare `.json`, a directory and a `.tmp`.

One follow-up nit, `tests/test_model_versions.py:462`: the cold `_learned_versions()` call runs outside `_counting_reg_reads`. So nothing in that test proves the counter sees reads at all. Wrapping the cold call and asserting the three names would close it.

Red-first verified: with the PR's test file checked out at origin/main, four LearnedVersions tests fail. They fail on three reads instead of zero, 3 != 0 on the alias path, and no `_learned_reg_cache` attribute. At head, `tests/test_model_versions.py` runs 41 passed and `tests/test_model_catalog.py` 31 passed. The Python 3.11 codex client-factory failure is not caused by this PR. That test imports only `codex_backend.py` and `session_backend.py`, and passed 3 of 3 at head and 3 of 3 at main under Python 3.11 here. The rerun is green, so all six required checks pass at head.

Review run by the maintainer's automated review session; verified against head 433ff4cf.

PR #950, opened 2026-09-06
Document what a typed /model or /effort does on each backend, and pin it

ksarma:
Follow-up to the merge note on #882, which named two consequences of routing typed commands through the kernel's setters and asked for a line about each: the composer's `/effort` reconnects an SDK session, and a typed `/model` on a tmux session is auto-confirmed. Nothing at tip recorded either. This PR writes both down where they will be read, and pins them. No behavior changes.

## Why document rather than align

The SDK's control channel (claude-agent-sdk 0.2.132) offers `set_model` and `set_permission_mode` but no effort request, so a reconnect is the only way to apply an effort change that the registry, `sdk-defaults.json` and the next connect all agree on, which is the property #882 exists to hold. On tmux, the CLI's `/model X` opens a confirmation; the dashboard cannot press Enter in the pane, and the next send's clear-guard would refuse to paste over the open dialog, so the kernel accepts the confirmation for the user, as the lane menu's pick has since the model park was built. Both behaviors are right; they were undocumented.

## What changes

One commit, three parts:

- **The contract and the router say what each backend does.** `SessionBackend.set_model` and `set_effort` (`kernel/session_backend.py`) were bare `...` with no docstring, while the neighbouring `set_fast` spells out its per-backend mechanics. Both gain docstrings in that register: what the SDK does, what tmux does, when False comes back. `_route_meta_command`'s docstring (`kernel/kernel.py`) gains the sentence that a typed command and a picker click are one gesture, and that the landing is the backend's. Docstrings and one trailing comment only; every source-pinned line (the `_set_effort_or_park(be, sid, value)` park line, the `sendCommand` call the UI test pins) is untouched.
- **User-facing docs.** `docs/reference.md` gains "Model and effort, from the statusline or a typed command" beside "Fast mode": a typed `/model X`, `/effort X` or `/fast on|off` is a setting change (a bare `/model` or a value the kernel cannot vouch for stays the CLI's); an SDK session switches model live but reloads for effort ("Reloading session…", switching-dots, deferred to turn end while busy); a terminal session gets the CLI's own command typed into its pane with the `/model` confirmation accepted on the user's behalf.
- **Guard tests.**
  - `tests/test_session_api.py`: the ABC's control setters must carry a docstring naming their mechanism. Red on the unfixed base (`SessionBackend.set_model carries a docstring` fails with `''`), green with the docstrings.
  - `tests/test_sdk_kernel.py` (KernelWiring): a typed `/model opus` or `/effort high` on a tmux-owned sid reaches `TmuxBackend.set_model` / `set_effort`, never `send()`. The existing routing test covered SDK sids only.
  - `tests/test_kernel_pane_clear.py`: `TmuxBackend.set_model` calls `_tmux_send` with `model_cmd=True` and `set_effort` without it. A recorder stands in for `_tmux_send`, so the test pins the call shape and never enters the timed confirm wait.

## Verification

- `pytest -n 8 tests`: 7329 passed, 46 skipped. One failure in `tests/test_postal_relay_honesty.py` (xdist ordering, unrelated to this change); the module passes alone (30 passed).
- `vscode-extension`: `npm run typecheck` clean, `npm test` 2781/2781, `npm run build` ok.
- `git diff --check` clean.

## Out of scope

Claude Code 2.1.263 marks its non-interactive `/effort` descriptor `supportsNonInteractive`, so a literal `/effort` into the SDK stream would execute. A live-effort path that avoids the reconnect is possible in principle, but it is the registry-bypassing shape #882 closed, so it would be its own feature PR. The second Enter's timed wait predates this PR and is unchanged.

Tier: tests-only

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -179,6 +179,26 @@ ghostty --working-directory={dir}   # Linux: Ghostty
 code {dir}                          # VS Code instead
 ```
 
+### Model and effort, from the statusline or a typed command
+
+Typing `/model X`, `/effort X`, or `/fast on|off` into the chat composer, or
+sending one with `romp send`, is the same setting change as a pick from the
+statusline's model and effort dropdowns: the kernel takes it through its own
+setters, so what it remembers (the value a reconnect relaunches with, the
+defaults new sessions start from) follows the switch. A bare `/model` (the
+CLI's own picker), a value the kernel cannot vouch for (a typo), or a longer
+message that merely opens with the command goes to the CLI verbatim, and the
+chat shows the CLI's own reply.
+
+The two backends apply the change differently. An SDK session switches model
+live but reloads to apply a new effort: the chat shows "Reloading session…"
+and the effort badge shows switching-dots until the reload completes, and a
+session that is mid-turn reloads when the turn ends. A terminal (tmux) session
+gets the CLI's own command typed into its pane. `/model` there asks for a
+confirmation, which the kernel accepts on your behalf so the pane is never
+left waiting on a keystroke the dashboard cannot send; `/effort` and `/fast`
+apply in place.
+
 ### Fast mode, from the chat statusline
 
 The statusline's badges — permission mode, model, effort — are each a small
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -11345,7 +11345,7 @@ class TmuxBackend(sb.SessionBackend):
         return True, ""
 
     def set_model(self, sid, value):
-        _tmux_send(_name_of(sid) or sid, "/model " + value, model_cmd=True)   # /model opens a confirm → 2nd Enter
+        _tmux_send(_name_of(sid) or sid, "/model " + value, model_cmd=True)   # /model opens a confirm → the 2nd Enter accepts it for the user
         return True
 
     def set_mode(self, sid, mode):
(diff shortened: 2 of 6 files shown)


ksarma:

Reviewed inline: The tests-only label fits. Both hunks in `kernel/kernel.py` are non-code. Line 11348 changes only the trailing comment. The statement before the hash is byte-identical at base and head. Lines 20489 to 20499 sit inside the `_route_meta_command` docstring, between its opener at 20485 and its closer at 20503. In `kernel/session_backend.py`, the abstract `set_model` and `set_effort` replace a bare ellipsis body with a docstring. Both compile to the same instruction stream and stay abstract. Nothing in `kernel/`, `cli/` or `postal/` reads their `__doc__` at runtime. Compiling `kernel.py` at base and head gives identical bytecode, names and non-string constants for all 1632 code objects. That one docstring is the only difference. I checked the docstring and docs claims against `SdkBackend.set_model` and `set_effort`, `TmuxBackend.set_model`, `set_effort` and `set_fast`, `_tmux_send`, and the three `_route_meta_command` callers (composer, lane menu, POST /send). Each mechanism claim holds. Added lines contain no private strings, hostnames or home paths.

- `kernel/session_backend.py:127` and `:142` (follow-up): the new contract promises `False` for an unknown sid or bad value. `TmuxBackend.set_model` (`kernel/kernel.py:11347`) and `set_effort` (`:11362`) return `True` unconditionally. Only `SdkBackend` behaves as written. A clause saying the tmux side cannot tell, so callers validate first (as `_route_meta_command` does), would make it accurate.
- `docs/reference.md:184` (follow-up): the sentence places a typed `/fast` under the model and effort dropdowns. The fast toggle is the separate badge the next section describes.

The three changed test files pass locally at head (64 passed, also under xdist `-n 4` twice). With the PR's test files checked out onto a scratch worktree at `origin/main`, the new docstring test fails with the empty `__doc__` the body describes. The two guard tests pass there, as expected for tests that pin existing behavior. All six required checks are green at head.

Review run by the maintainer's automated review session; verified against head 6a049cad.

PR #951, opened 2026-09-06
Tests: bats suites that reach tmux get a private socket directory, and the ones that bind ports pick free ones

ksarma:
## What breaks

Three test-only problems in the bats suites; no product code is touched.

**1. Real managers run `tmux start-server` on the developer's default socket.** `tests/romp-manager-ensure.bats` and `tests/romp-manager-origin.bats` start a real `bin/romp-manager`, whose `startManager()` runs `tmux start-server ; set -g exit-empty off` before it binds the control port. Only the tmux-bootstrap test put a fake tmux on PATH, so on any machine with tmux the other tests' managers ran that against the default socket. Usually a no-op, because a server is already up. With none up, the test's manager created the machine's tmux server, and for the rest of the day that server carried the test's environment (`BATS_*` variables, a fake-serve path under a temp dir already deleted). CI does not see it: both runners are ephemeral, so a server a test leaves behind dies with the job.

**2. Fixed loopback ports collide.** ensure.bats's tmux-bootstrap test used 7571/7572, the same literals as origin.bats's control and serve ports. A manager SIGTERM'd in origin.bats outlives the kill by `shutdownAll`'s exit grace, so `bats tests/romp-manager-origin.bats tests/romp-manager-ensure.bats` finds `up` exiting "already running" without ever calling tmux (2 of 3 runs here). Any literal also collides when two checkouts run bats at once on one machine (reproduced: both runs red at the base).

**3. The postal squat port sits inside the ephemeral range.** The red Shell check on #897's first CI attempt (2026-09-03) was not a manager-suite collision. `tests/romp-postal.bats` "the bus spawn survives a taken port" binds its squatter at `ROMP_POSTAL_PORT + 7000` = 34200+N, inside the 32768-60999 range the file's own header says to stay below, and the squatter itself failed to bind (`Address already in use`) when a transient source port held it.

## The fix

Three commits.

**`tests/free-port.bash`** (commit 1). `free_port VAR...` assigns each named variable a distinct loopback port nothing is bound to, probed by a plain bind (no `SO_REUSEADDR`, so a TIME_WAIT port reads busy and is skipped), chosen at random from 20000-24999: below Linux's (32768-60999) and macOS's (49152-65535) ephemeral ranges, so a transient outgoing connection cannot hold the pick as a source port between the pick and the subject's bind (the rule romp-postal.bats's header already records), and clear of the project's own defaults (7432, 29855, the postal tests' band). It fails loudly without python3 or a free port: an empty port would reach the manager as `Number('') || 7432`, the default control port, where a developer's live manager may be listening, and bats fails the test at the pick. Every fixed bind in romp-manager-ensure, romp-manager-origin and romp.bats now picks (including romp.bats's dead-port probe, which wants exactly an unbound port). ensure.bats's tests share setup's per-test picks, so teardown's `/stop` reaps every manager the same way and the per-test literals are gone.

**`tests/tmux-private.bash`** (commit 2). `tmux_private_socket_dir "$TEST_DIR"` exports `TMUX_TMPDIR` under the test dir and creates it first: tmux 3.4 silently uses the default socket directory when `TMUX_TMPDIR` names a missing one (`make_label()` expands the candidates with errors ignored; the same code is in 3.5a; verified on 3.4 via `#{socket_path}`), so an export without the mkdir isolates nothing. `tmux_private_kill && rm -rf "$TEST_DIR"` ends teardown: it kills every server whose socket sits under the private directory, addressing the socket with `-S` (so it can never resolve to the machine's server) through the first tmux on PATH outside the test dir (a PATH mock answered `kill-server` with exit 0 and leaked a server on the first try), fails when the directory is already gone (a server started under it has then leaked), and refuses when the helper was never armed. It has to be teardown's final status: bats runs teardown with errexit and the ERR trap off, so a failing command mid-teardown is swallowed and the test passes. Wired into the four suites whose subject shells out to tmux: romp-manager-ensure (its recording fake tmux now covers the whole file, since the detached manager that `ensure` spawns inherits PATH), romp-manager-origin (a no-op tmux), romp, and tmux-status-hook. `tests/README.md` records both rules.

**`tests/romp-postal.bats`** (commit 3). The squat moves to +2000 (29200+N), which keeps it and both retry candidates (+500, +1000) below 32768 and clear of the per-test band and the kernel's default port; an arithmetic pin holds the band.

## Tests

- `tests/tmux-private.bats` (6, none runs tmux): the directory is created and exported; the kill fails loudly when the directory is gone; it is a silent no-op when never armed; the real-tmux lookup refuses when unarmed and skips a tmux under the test dir; and under a nested bats, the prescribed teardown line fails a test whose `rm -rf` ran before the kill and passes one where it did not.
- `tests/free-port.bats` (5): a pick is in band and bindable; picks are distinct; a caller's variable is assigned even when it is named like one of the helper's locals; loud failure without python3; usage error with no names.
- Two pins in `tests/romp-manager-ensure.bats`: the `ensure`-spawned detached manager's tmux is handed a `TMUX_TMPDIR` under the test dir that already exists (fake tmux; runs on every CI OS); and with the real tmux, the server socket appears only under the private directory, answers `show -g exit-empty` with `off`, and is gone after the teardown kill. That one skips where tmux is absent; the ubuntu runner image ships one, so on the Linux bats job it ran as `ok 144` rather than skipping; `HOME` and `XDG_CONFIG_HOME` point at the test dir so the developer's tmux.conf is not sourced (`/etc/tmux.conf` is still read; nothing short of `-f` avoids that, and the manager's call cannot carry one).
- Red-first: with the helper's `mkdir` removed, both ensure pins and the helper's create test fail (the real-tmux pin at its `[ -d "$TMUX_TMPDIR" ]` guard, before any tmux runs), and so does every test in the four wired suites, at teardown's `tmux_private_kill` guard, which reads the never-created directory as gone; with `rm -rf` moved before the kill, all six ensure tests fail with the leak message. The 7571 collision reproduced at the base (2 of 3 reverse-order runs; both of two concurrent runs) and not after (0 of 3; 0 of 2). The squat pin fails with +7000 and passes with +2000.
- Full `bats tests/*.bats`: 402/402, real-tmux pin `ok`; no tmux server or manager left behind after any run. Full pytest and `node --test tests/manager-*.test.js`: unchanged surface (no `.py` or `.js` touched).

## Not in this PR

Installing tmux on the Linux bats job (the workflow deliberately installs only bats, and the ubuntu runner image already ships one, so the real-tmux pin runs there anyway; the fake-tmux pin carries the coverage on a runner without tmux). Two pre-existing flakes seen while running romp.bats, both present at the base and untouched here: its curl mock exits without draining the `--config -` stdin that `bin/romp` pipes in, so under load a kernel-API test fails with "kernel not reachable" (passes alone); and six temp dirs per run survive teardown with `home/.local/state/romp/serve-token` written after the `rm -rf`.

One risk to note: macOS caps Unix socket paths at 104 bytes, and `mktemp -d` there plus `/tmux/tmux-<uid>/default` lands near 94. If the manual macOS bats job ever fails in the real-tmux pin on socket creation, shorten the private dir.

Tier: tests-only

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -16,6 +16,26 @@ Every bug fix or feature change lands with a test (repo rule). Four suites:
 - **`*.bats`** — the shell surfaces: `bin/romp`, the launch chain, hooks,
   postal CLI. Keep them GNU/BSD-portable (CI runs bats on ubuntu).
   Run: `bats tests/*.bats`.
+  Any test whose subject shells out to tmux must isolate the tmux socket
+  directory: `load tmux-private`, `tmux_private_socket_dir "$TEST_DIR"` in
+  setup (it exports `TMUX_TMPDIR` under the test dir and creates it first;
+  tmux 3.4 silently uses the machine's default socket directory when
+  `TMUX_TMPDIR` names a missing one), and `tmux_private_kill && rm -rf
+  "$TEST_DIR"` as the last line of teardown (the kill fails when the
+  directory is already gone, since a server started under it has then
+  leaked; it has to be teardown's final status, because bats swallows a
+  failing command mid-teardown). A tmux mock on PATH
+  covers only the tests that install one: on 2026-09-06 a full bats run
+  ran `romp-manager-ensure.bats` while the machine's default tmux server
+  was down, the real manager it starts ran `tmux start-server` on the
+  default socket, and for the rest of the day the machine's tmux server was
+  the test's, carrying the run's environment inside the service's cgroup.
+  Any test whose subject binds a loopback port picks it with `load
+  free-port` + `free_port VAR...`, never a literal: a literal shared by two
+  files collided within one run (`romp-manager-ensure.bats` once used
+  `romp-manager-origin.bats`'s control port), and any literal collides when
+  two checkouts run bats at once on one machine. The helper picks below
+  the ephemeral range, so a transient source port cannot hold the pick.
 - **node suites** — live beside their sources in `ui/webview/*.test.ts` and
   `vscode-extension/src/*.test.ts`, run with `npm test` from
   `vscode-extension/`. Many pin lines of `kernel/kernel.py` as strings — run
--- a/tests/free-port.bash
+++ b/tests/free-port.bash
@@ -1,0 +1,68 @@
+# tests/free-port.bash: loopback TCP ports nothing is bound to, for the bats suites whose subject
+# binds one (the ones that start a real bin/romp-manager).
+#
+# Use from a bats file: `load free-port` at the top, then `free_port CPORT MPORT` in setup() or a
+# test (`local cport mport; free_port cport mport` inside a test). Each named variable is assigned a
+# distinct port. The function fails, with a `free-port:` line on stderr, when python3 is missing or
+# no port is found, and bats then fails the test right there: an empty port would otherwise reach
+# bin/romp-manager as `Number('') || 7432`, the default control port, where a developer's live manager
+# may be listening.
+#
+# Why not a literal: a literal shared by two files collides within one run (romp-manager-ensure.bats
+# once ran `up` on romp-manager-origin.bats's control port, and a manager SIGTERM'd there outlives
+# the kill by shutdownAll's exit grace, so `up` exited "already running" and never called tmux), and
+# any literal collides when two checkouts run bats at once on one machine.
+#
+# Ports come from 20000-24999: below Linux's ephemeral range (32768-60999) and macOS's (49152-65535),
+# so a transient outgoing connection cannot hold the pick as its SOURCE port between the pick and the
+# subject's bind (the lesson tests/romp-postal.bats's header records), and below the project's own
+# defaults (the postal bus's 25302, the postal tests' 27200-28300, the kernel's 29855; the manager's
+# 7432 sits under the band). The probe is a plain bind with no SO_REUSEADDR: a port in TIME_WAIT from
+# an earlier test reads busy and is skipped (node's listen would bind it anyway, so skipping is the safe
+# side). What remains is the window between the pick and the subject's bind, milliseconds long, while
+# bats runs files one after another; a second bats run on the same machine competes on a random pick
+# out of 5000, where a literal collided every time.
+#
+# The locals carry a prefix because bash scopes dynamically: with plain names, `local name; free_port
+# name` would assign the helper's own local and hand the caller an empty string, the exact failure the
+# helper exists to prevent.
+
+free_port() {   # free_port VAR...: assigns each VAR a distinct free loopback port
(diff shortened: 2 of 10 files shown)


ksarma:
Merging as shipped. Two follow-ups below, and neither holds the merge.

Reviewed inline: the diff touches ten files under `tests/` and changes no runtime path, so the tests-only label fits. `tests/free-port.bash` probes a plain bind on 127.0.0.1 from 20000-24999. That band sits below the Linux and macOS ephemeral ranges and clear of 7432, 25302 and 29855. The probe skipped two unrelated listeners in that band on the review box. `tests/tmux-private.bash` creates the directory before exporting TMUX_TMPDIR. That order is the fix: tmux 3.4 with a missing TMUX_TMPDIR falls back to the default socket directory, confirmed with a read-only probe. The kill-before-rm teardown depends on bats 1.10 failing a test only on teardown's final status, confirmed against bats-exec-test. The red-first claims reproduce. The reverse-order origin plus ensure run fails at origin/main on the 7571 collision. It passes at head in sequential and concurrent runs. Removing the mkdir, moving the rm before the kill, and restoring the +7000 squat each go red. The real-tmux pin ran on the Linux CI job rather than skipping.

- `tests/free-port.bash:45`, follow-up (body only): the PR body lags the final revision. The band is 20000-24999, not 20000-26999. free-port.bats has five cases, not four. The suite is 402, not 401. The real-tmux pin ran on the Linux CI job as ok 144 rather than skipping, as commit 297c1de already says. The mkdir mutation also fails every wired suite at the teardown guard, not only the two pins and the create test. Commit messages are correct.
- `tests/romp-manager-ensure.bats:47`, follow-up: teardown can still leak a manager if the detached manager from `ensure` has not bound CPORT within the 4 or 5 s wait. In that case teardown has no pid, and tmux_private_kill finds no socket. The rm then removes the fake tmux and the private directory before the manager runs `tmux start-server`. The real tmux falls back to the default socket directory, and the manager leaks. This needs a severe stall. The base did this on every run, but it is the one gap left in what this PR closes. A teardown that polls /status for a bounded window would cover it, failing loudly and leaving TEST_DIR in place when the manager never answers. #953 stacks on this PR and could carry it.
- `tests/tmux-private.bash:78`, nit: `kill-server ... || true` discards a connect failure for a socket whose server is mid-start, so teardown proceeds to the rm. The case is narrow, but the helper's stated stance is loud failure.

Locally, under a private TMPDIR and TMUX_TMPDIR with the real tmux 3.4, all seven touched suites pass with no skips: free-port 5, tmux-private 6, tmux-status-hook 44, romp-manager-origin 1, romp-manager-ensure 6, romp-postal 26, romp 98. No tmux server or manager was left behind. CI: all six required checks are green at head, Shell job 1..402.

Review run by the maintainer's automated review session; verified against head cdc6f748.

PR #952, opened 2026-09-06
A dropped dashboard socket is loud, the stale prompt keys on the heartbeat, and the Outline pane's ages run on the kernel's clock

ksarma:
Three small fixes on top of the view-delta protocol that #914 and the #928 fold shipped. That protocol is what this builds on and nothing here changes it: the shim still dials `?delta=1`, `_send_slot`/`_send_slot_delta` still carry the feed and the bars, and the shim still reassembles every `{type:"delta"}` frame before a bundle sees it. What was left after it landed were three things that leave no trace or run on the wrong clock.

## 1. A client dropped for falling behind is silent

`_mk_ws_send` drops a client that has fallen `WS_QUEUE_BYTES` behind by raising, and every caller — `_send_client`, the tail of `_send_slot_delta`, `_send_to_app`, `_send_to_view`, the keepalive — catches it with a bare `c["alive"] = False`. `_drop_dead_ws_client` writes a `ws: dropping` line for the ping-timeout and supersession drops, but the budget drop writes nothing, so a dashboard dropped and reconnected every few minutes reads as a flaky network.

**Fix.** The raise site logs the drop (`_note_ws_drop`), once per client, with the same `ws: dropping` prefix the other two drops use, so one grep finds every kind: the pane, the dashboard id, the backlog, the frame that tipped it, and — for a push-path send — the slot whose frame it was. `_client_send` leaves the slot on the client for the length of its call, so the line reads `slot=feed`, `slot=chat:<sid>`, `slot=taborder`; a direct one-shot `client["send"]` reply reads `slot=-`. Every push and keepalive send goes through `_client_send`; the ~60 direct replies are covered by the raise site itself. The same drop rides the dashboard's bell through `_sdk_problem_rows` (the existing kernel-problems channel): at most five drop rows, none older than an hour, merged by time with the backend rows so they never crowd a backend problem out. Bell text is plain language for the person at the dashboard.

**Reproduce (before):** a client with the kernel's own `send` and `qbytes` ten bytes under the budget receives a 100-byte frame → `OSError`, `alive=False`, and nothing on stderr. `tests/test_ws_drop_loud.py` drives this on `_send_client`, on a direct `client["send"]`, and through `_send_slot` on a delta client whose second (delta) frame tips the budget.

## 2. The stale prompt is armed on a timer

After a reconnect the pane shim arms the "what you see may be stale" prompt on a one-second `setTimeout`. On a board of several hundred cards the connect push is a multi-megabyte frame that can take longer than a second to land over a forwarded or tunnelled link, so the prompt raises on reconnects that are resyncing normally, and a hidden pane's reconnect storm multiplies it.

**Fix.** The arm keys on two events. The prompt raises on the **second** keepalive arriving on the reconnected socket before any resync frame — one heartbeat period bracketed by two beats from a kernel that is alive, talking to this socket, and not resyncing it — or on the reconnected socket closing again before its resync. One keepalive is not the event: the heartbeat thread can enqueue a beat the instant a socket is accepted, ahead of the connect push. The first non-keepalive frame retires the arm as before (a delta the shim reassembles counts — the bundle receives the full message); the hidden-pane suppression, the announced-restart skip and the foreground path are unchanged in behaviour (the foreground path now hands its reason to the reconnect it forces, whose arm follows the same two events). Every close of a socket that opened leaves a `wsclose` clientDiag (code, reason, wasClean, socket age, quiet gap) — a kernel-side drop (1006, no close frame), a clean restart, a proxy timeout and the watchdog's own close were indistinguishable before; the redials an outage refuses are counted and reported as one `wsconnfail` row on the next open; at most 20 breadcrumbs wait in the shim's queue, other queued messages untouched.

**Design point to weigh.** A genuinely stale reconnect is now reported after 10–20 s of no resync rather than after one second. The trade is against a prompt that was raising on healthy reconnects.

## 3. The Outline pane's ages run on the browser's clock, and only when a frame lands

`fleet.ts` computed every "(Xm ago)", the current goal's elapsed time, the recency cutoff and the slider's range from `Date.now()` inside `render()`, against timestamps the **kernel** wrote — so every age is off by whatever skew sits between the browser's clock and the kernel's (minutes on a phone or a laptop back from sleep). And `render()` runs only on a frame, a settings event or a click; on the delta path a quiet board sends the pane one frame every 60 s (`_DEDUP_REPOST_S`), so the ages move in minute steps.

**Reproduce (before):** the new `fleet-live-clock.test.ts` runs `fleet.ts` for real under mock timers with the kernel clock five minutes behind the browser's. Two cards 20 s and 40 s old render as **no rows at all** — the browser clock reads them five minutes old and the recency cutoff drops them.

**Fix.** The pane anchors on the frame's `now` paired with the moment that frame arrived from the wire (`feed-age.ts` `liveNow`: the kernel's clock plus the local time elapsed since — only the local clock's deltas enter, never its offset), and re-renders every 15 s while visible. A whole `render()` rather than a per-element repaint, because the cutoff filter must be able to drop a row as it ages out. A hidden document or a zero-size iframe (the shell's `display:none` pane) skips the tick and catches up once when shown (`visibilitychange`, or the `resize` the iframe gets), not on every flip.

`federation.ts` stamps each host's feed frame with its wire arrival (`perHostFeedAt`) and `mergeHostFeeds` puts it beside `now` on the merged frame as `nowAt` — both from the local frame, or from the newest remote arrival when no local frame exists, so the pair always describes one frame. The merged frame is re-emitted on a view-order write, on every remote host's frame and on a detach; anchored on the emit instead, every age would move back by however long the local board had been quiet. `nowAt` is additive: `feed.ts` ignores it and is unchanged, and a frame without it (the VS Code pipe hands frames straight to the pane) anchors at its arrival.

A `{type:"delta"}` frame reaching the bundle unreassembled — the shim rebuilds every one before a bundle sees it, and federation's remote sockets never dial for deltas, so this means a host handed the pane a kernel frame raw — is loud (`console.error`, a clientDiag breadcrumb, a `needSlot` for the whole slot, the shim's own message for a delta it cannot apply) instead of falling through the `type !== "feed"` return with every later update dropped in silence.

## Tests

- Python: `tests/test_ws_drop_loud.py` (new, 7) — one stderr line per drop with its fields; the row reaches `_sdk_problem_rows`; a direct send's drop is loud; a push-path drop names its slot; a drop on the view-delta path is loud and named; drop rows never hide a backend problem; every send path goes through the helper (source pin). Shim pins in `test_kernel_disconnect_banner`, `test_kernel_ws_heartbeat` (the keepalive-branch pin now holds the branch's `return` explicitly) and `test_restart_seam` follow the new text.
- Node: `ui/webview/pane-shim-stale.test.ts` (new, 10) runs the shim's JavaScript, lifted from `kernel.py` and rendered as the feed page renders it, in a vm sandbox with fake sockets, timers and clock, and drives each rule (resync first; one keepalive; two keepalives; a reassembled delta as the resync; close before resync and the foreground path; announced restart; wsclose; wsconnfail; the queue cap). `fleet-live-clock.test.ts` (new, 5) runs `fleet.ts` against a small DOM stand-in under mock timers. `feed-clock-anchor.test.ts` (new, 4) drives the real `FederationManager` with a stub clock. `feed-age.test.ts` (new, 2) pins `liveNow`. One pin added to `fleet.test.ts`; `view-order-wiring.test.ts` follows the new `mergeHostFeeds` call.
- Every new or changed test was run against the unfixed code first and fails there (details in the commit messages). All fixtures are synthetic (placeholder ids, `TESTHOST`, the notes-api demo world).

Docs: one bullet each in `docs/read-side.md` for the keepalive-keyed prompt and the loud drop, and one for the Outline pane's clock; one sentence in `docs/guide.md`'s remote-access subsection.

Known follow-ups, not in this PR: the feed pane's own 15 s age tick still reads `Date.now()` (the same skew, bounded by the 60 s repost); `mergeHostFeeds` concatenates remote hosts' ledgers without rebasing their times onto the local clock, so a remote row's age is off by that host's skew.

Tier: fix

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -339,6 +339,9 @@ own a direct path; the dashboard is then a plain URL on your tailnet.
     when a panel closes, so it never leaks connections, but under Remote or
     Tunnels the extension still relays each whole view payload to the local
     window as it changes. It does not yet take the deltas the browser panes do.
+    A pane that falls 16 MB behind is dropped and reconnects on its own; the
+    drop is logged in the kernel log and shows in the dashboard's bell, so a
+    link that cannot keep up reads as what it is rather than as a flaky network.
 
 ### From your phone
 
--- a/docs/read-side.md
+++ b/docs/read-side.md
@@ -54,6 +54,31 @@ completed); the feed just paints columns. (Reflected in `docs/judges.md`.)
   **same served UI** over the **same protocol**, so the two front ends stay
   consistent by construction. Keeping the WS protocol stable is the compatibility
   contract.
+- **Liveness is a keepalive, and staleness is event-keyed.** The kernel sends a
+  `ka` frame to every socket every 10 s; the pane shim force-closes a socket that
+  has gone 30 s without any frame. After a reconnect the shim raises the "what you
+  see may be stale" prompt only on the SECOND `ka` arriving before the resync
+  frame — one full heartbeat period, bracketed by two kernel heartbeats on that
+  socket with no resync between them (a single `ka` can be a beat that was already
+  queued when the socket was accepted) — or on the reconnected socket closing
+  again before its resync; the first non-keepalive frame retires the prompt, never
+  a timer. A client that falls 16 MB behind is dropped by the kernel, loudly: one
+  `ws: dropping` line in the kernel log, written where the drop is decided so every
+  send path is covered, naming the pane, the dashboard, the backlog and the push
+  slot whose frame tipped the budget when a push did; and a row in the dashboard's
+  bell (at most five drop rows, none older than an hour, so they never crowd out a
+  backend problem). Every close of a socket that opened leaves a `wsclose`
+  breadcrumb (code, reason, socket age) in `client-diag.jsonl`; the redials an
+  outage refuses are counted and reported as one `wsconnfail` row on the next
+  open, and at most 20 breadcrumbs wait in the shim's queue for it.
+- **The Outline pane's ages run on the kernel's clock.** Its timestamps are the
+  kernel's, so the pane never reads the browser's clock against them: it anchors
+  on the frame's `now` paired with the moment that frame arrived from the wire
+  (`feed-age.ts`; federation stamps the merged frame it re-emits with `nowAt`, so a
+  re-emit on a view-order write or a remote host's frame never moves an age), and
+  every "(Xm ago)", the current goal's elapsed time and the recency cutoff move on
+  a 15 s refresh — skipped while the pane is hidden, with one catch-up render when
+  it is shown again — rather than only when a frame lands.
 - **Both judge tiers run continuously for any live session — no connection gate.**
   The kernel runs the index tier (captioner + archiver) AND the triage tier
   (planner → closer → courier → grouper → consolidator → distiller) in parallel,
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -30214,6 +30214,8 @@ def _shim(app, v=0):
     # pane sat silent through rebuilds).
     return """
 (function(){var queue=[],ws=null,everConnected=false;
+var queuedDiag=0,DIAG_QUEUE_MAX=20;   // clientDiag rows waiting in `queue` for a reconnect, capped (an outage must not pile up breadcrumbs); other queued messages are untouched
+var failedConnects=0,firstFailT=0;   // handshakes that never OPENED since the last open: reported as ONE wsconnfail row on the next open, never one wsclose per redial
 // This pane's DASHBOARD id. ?wid= when the host supplies one (the VS Code extension builds its own pane
 // URLs); otherwise the shell's per-tab id from sessionStorage, which is scoped to the browsing context and
 // shared with same-origin iframes — so every pane of one window agrees, a second window differs, and no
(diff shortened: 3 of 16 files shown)


ksarma:
Merging as shipped. One follow-up to fold, at `kernel/kernel.py:30405`.

Reviewed inline: the budget drop in `_mk_ws_send` logs at the raise site through `_note_ws_drop`, once per client. Every push and keepalive send goes through `_client_send`, so no bare `alive=False` catch remains. The row reaches the bell through `_sdk_problem_rows`, capped at five and merged by time with the backend rows. The shim's stale prompt raises on the second keepalive on the reconnected socket before any resync frame. It also raises when that socket closes before its resync. The first non-keepalive frame retires it. The announced-restart skip, hidden-pane suppression and foreground path keep their behaviour. The Outline pane computes ages from `liveNow(now, nowAt, Date.now())`, so only the local clock's deltas enter. It re-renders every 15 s while visible, with one catch-up render when shown. Federation stamps each host's frame arrival as `nowAt`; this is additive, so `feed.ts` is unchanged. A raw delta frame reaching the pane is an error rather than a silent return. Privacy scan of the 1126 added lines is clean. One gap: the shim's silence watchdog abandons an armed socket without running the close rule. A reconnected socket that gets neither a keepalive nor a resync therefore never raises the prompt, where main's timer did.

Findings:

- `kernel/kernel.py:30405` (follow-up): the watchdog branch calls `abandon()`, which nulls `onclose` before closing. The close rule at 30352 never runs, and the next open re-arms. A socket that stays silent after a reconnect flaps the loader every 30 s and never raises the prompt. Reproduced over three cycles; main raised on the first. Fix: retire `stalePending` and call `raiseStale(why+"-quiet")` before `abandon()`, in the foreground abandon too. Pin it in `pane-shim-stale.test.ts` by capturing `setInterval`. Same fold: `docs/read-side.md:70` and the comment at 30342 say every opened socket's close leaves a wsclose row. Abandoned sockets leave none.
- `kernel/kernel.py:27446` (nit): the bell text formats the raw app id, so the Outline pane's drop is named by its internal id. Map through `_PANE_ORDER`.

Red-first held. At origin/main with only the PR's test files checked out, pytest ran 15 failed / 29 passed: all 7 in `test_ws_drop_loud.py`, 5 disconnect-banner pins, 1 heartbeat pin, 2 restart-seam pins. The Node build could not resolve `./feed-age` for two files. The remaining four ran 16 failed / 33 passed: pane-shim-stale 9 of 10, the Outline live-clock file 5 of 5, one pin each in the Outline pane test and view-order-wiring. At head: 44 passed for the four Python files, 55 of 55 across the six webview files. CI: all six required checks green.

Review run by the maintainer's automated review session; verified against head a39e4297.

PR #953, opened 2026-09-06
Session CLIs and the manager's tmux server run in their own systemd scopes, so a service restart no longer kills session work

ksarma:
#941 and #951 have merged, so this PR sits on `main` as its own commits, rebased onto the current tip: the three that make the change (b570ef49, 4975d38d, 5e48c1b5) and three from review (947e2720, ad1cac11, 590ba7dc; see "From review" below). #941 is a prerequisite: the scopes remove the cgroup kill that had been terminating the orphans the ppid-1 reaper missed on Linux.

## What breaks

On Linux a service restart kills every session's detached work along with the kernel. The manager runs as a systemd user service with systemd's default `KillMode=control-group` (the unit sets no KillMode line, on purpose: a wedged kernel or postal bus still holding its port must die on a service restart). Every process the service tree starts lands in the service's cgroup: each session's `claude` CLI, its tool shells, their `setsid` children and any tmux server a session started. `systemctl --user restart romp-manager` empties that cgroup. Kernel-only restarts (`romp refresh`) never touch the service, so they leave the detached work running.

To reproduce on a Linux box with the service installed: from any SDK session's shell, run `tmux new-session -d -s sweep 'sleep 600'`, then `systemctl --user restart romp-manager`. `tmux ls` no longer lists `sweep`.

## The fix

1. `bin/romp-cli-scope`: an exec-in-place POSIX sh wrapper around `systemd-run --user --scope --quiet --collect --unit=romp-session-<sid8>-<pid>-<start-ns>`. The SDK spawns the wrapper as `cli_path`, the wrapper execs systemd-run, and systemd-run in `--scope` mode execs the CLI in place, so the CLI keeps the pid, parent and argv the SDK spawned; `find_session_cli` and `find_orphan_clis` are unchanged. A pre-flight `systemd-run … -- true` (bounded to 10 s when coreutils' `timeout` is on PATH; the SDK's initialize timeout is 60 s) that fails falls back to a direct launch with one `romp-cli-scope: fallback: …` stderr line, which the kernel logs at once as a problem naming the session (a launch that succeeds never drains the stderr tail, so nothing else would read it). `ROMP_CLI_REAL` unset is refused with exit 127 and a `romp-cli-scope: refused: …` line, which the launch-error card reports. With no `ROMP_SID` (the SDK's per-connect `claude -v` probe) the wrapper runs the CLI directly.
2. `cli_scope_supported` (kernel/sdk_backend.py) decides once per backend, at construction: on with `ROMP_CLI_SCOPE=1`, or under `ROMP_SUPERVISED` unless `ROMP_CLI_SCOPE=0`; then only on Linux, only with `systemd-run` on PATH, and only if a probe scope starts. The verdict is logged as `cli scope: on` or `cli scope: off` with the reason. When the switch was wanted on Linux and the box cannot honour it (no `systemd-run`, or the probe failing), the off verdict is also filed as a problem, so the error center shows that sessions are running inside the service cgroup; off by request, or off because the platform has no systemd, stays a plain log line. When on, the kernel also exports `ROMP_CLI_REAL` into its own environment: the SDK's minimum-version check spawns `[cli_path, "-v"]` with the kernel's environment rather than `options.env` (claude_agent_sdk 0.2.132), so the wrapper has to find the real CLI there too.
3. `bin/romp-manager` starts its default tmux server through the same switch (`tmuxStartArgv`, a pure function with its own node suite) in a `romp-tmux-<ms>` scope, and falls back to the bare call with a distinct log line. That line carries the first stderr line of the tool that failed, bounded to 200 characters: systemd-run's own (its logger prefixes every line `Failed to …`), or tmux's when tmux failed inside a scope that did start. With nothing on stderr it names systemd-run and says how the call ended (exit status, timeout, signal). A box without tmux logs nothing, as before.
4. The lost-background-tasks notice (`task_death_notice`) says the tasks were cut off when the claude process that started them ended, and asks the session to check whether each is still running before relaunching it: under scopes a task's shell can outlive the CLI.
5. `launch_failure_text` drops the wrapper's fallback notice from a failed launch's stderr tail so the CLI's own reason fits on the card.
6. `tests/conftest.py` floors `ROMP_CLI_SCOPE=0` so no test can start a real scope on a developer's user manager: the verdict defaults to on under `ROMP_SUPERVISED`, which a tool shell under a self-hosted install inherits. For bats, `tmux_private_socket_dir` (the helper every suite that starts the real manager already calls) exports the same floor.

Docs: a "What survives a restart" section in docs/reference.md (with the error-center entry) and a row for the wrapper in bin/README.md.

## macOS

macOS is unchanged by construction. The kernel answers "not Linux" before it looks for `systemd-run` (the launchd plist sets `ROMP_SUPERVISED` too; with the checks the other way round, every macOS start would log "systemd-run is not on PATH"), and that verdict is a plain line, not a problem entry (otherwise every macOS kernel start would open with an error-center entry for a feature the platform does not have). The manager picks the bare tmux argv on darwin. Both paths are covered on Linux CI by platform-injected tests, since the macOS matrix runs only on a manual dispatch: `Supported.test_not_linux_is_off_before_which_or_probe` over darwin, freebsd13 and win32 asserts no PATH lookup, no probe, no problem entry and an off verdict that names the platform; the node suite's darwin case asserts the bare argv whatever the environment. The bats wrapper suite skips its 10 s hang-bound check where `timeout(1)` is absent, and the manager suite skips its two scoped cases off Linux.

## Tests

- `tests/test_cli_scope.py`: the truth table with injected platform, `which` and `run`, with the problem flag asserted on every branch; the exact probe argv; one cached verdict per backend, and the wanted-but-unavailable verdict landing in the backend's `problems()`; the `ROMP_CLI_REAL` export; `_options` wiring; a missing or non-executable wrapper degrading loudly once; the fallback notice logged as a problem naming the session; the refusal line not logged as a fallback.
- `tests/romp-cli-scope.bats`: a fake `systemd-run` first on PATH records argv; verbatim argument pass-through; unit naming; pre-flight ordering, fallback and the 10 s bound; exec-in-place pid checks on the scoped path and every direct path; a clean stdout; the two refusals.
- `tests/romp-manager-tmux-scope.bats`: the real manager started with a fake `systemd-run` first on PATH and `ROMP_CLI_SCOPE=1`. The fake records its argv and either execs the command after `--`, as `--scope` mode does, or exits 1 with a bus error on stderr; the fake tmux can fail the same way with a connect error. Four cases: the scoped call (one `systemd-run` call with the `--user --scope --quiet --collect --unit=romp-tmux-<ms>` argv, tmux run once from inside it, the "in its own transient scope" line); `systemd-run` failing (tried once, then the bare call; the failure line names systemd-run and carries its first stderr line and not its second; the "without a scope" line); tmux failing inside the scope (`systemd-run` called once, tmux twice, inside the scope and then bare; the log names tmux and carries its line); and the floor under an inherited `ROMP_SUPERVISED` (no `systemd-run` call, the bare start, the plain line). Every absence check runs grep and tests its status, the form `romp-cli-scope.bats` uses. Nothing reaches the real `systemd-run` or the user manager.
- `tests/manager-tmux-scope.test.js`: 19 cases, collected by the existing `node --test tests/manager-*.test.js` step; seven cover `scopedStartFailure` (systemd-run's line, tmux's line, the first non-blank line only, the 200-character bound, nothing on stderr, a timeout and a signal, a spawn error).
- `tests/test_cli_scope_floor.py`: pins both halves of conftest's `ROMP_CLI_SCOPE=0` floor on its source, the module-level statement and the autouse fixture that re-asserts it. A test that reads the value cannot pin the floor: `test_cli_scope.py` sets `ROMP_CLI_SCOPE=0` itself at import, and pytest collects every module on every worker before any test runs, so with both floor lines removed the value still read "0" under a full run while the two source pins failed.
- Updated: `tests/test_sdk_launch_error.py` (the notice is dropped from the tail, the refusal stays), `tests/test_injected_voice.py` (the reworded notice, pinned in full in its singular, plural and reconnect forms), `tests/test_kernel_bundled_echo_prune.py` (its fixture literal), `tests/tmux-private.bats` (the helper's floor).

Against the base without the code: `test_cli_scope.py` collects and 27 of its 30 tests fail (the three that pass exercise behaviour the base already has: an ordinary stderr line is only buffered, the refusal line is not logged as a fallback, and `_options` with the switch off leaves `cli_path` alone); `test_sdk_launch_error.py` fails at collection (it reads `CLI_SCOPE_FALLBACK_PREFIX` in a class body); `romp-cli-scope.bats` fails on the missing wrapper; `romp-manager-tmux-scope.bats` fails its three scoped cases (the floor case pins behaviour the base has); `manager-tmux-scope.test.js` fails 19 of 19 on the missing exports; the injected-voice test fails on the old wording; `test_cli_scope_floor.py` fails 2 of 3 against a conftest without the floor under a full run (`test_cli_scope.py`'s import-time set satisfies the value check), and 3 of 3 run alone. With it, on Linux: pytest 7417 passed and 50 skipped; the six bats suites this PR touches or that start the real manager 135 of 135 (the previous revision's full bats run was 425 of 425, and this revision adds one case); `node --test tests/manager-*.test.js` 59 of 59. The extension is unchanged since the previous revision, where `npm test` was 2830 of 2830 and `tsc --noEmit` clean.

## From review

The review comment on this PR named three follow-ups and two nits. All are folded in, as three commits on top of the original three.

- The boot verdict is a problem entry when it should be (947e2720). `cli_scope_supported` hands its log callback `problem=<bool>`: true when the switch was wanted on Linux and `systemd-run` is missing or the probe fails, so the error center shows that sessions are running inside the service cgroup, as the launch-time fallbacks already did. The not-Linux and off-by-request verdicts stay plain lines. `test_cli_scope.py` asserts the flag on each branch of the truth table and, through a backend built on a supervised Linux box without `systemd-run`, the entry in `problems()`. docs/reference.md notes the entry.
- `startTmuxServer` has a bats suite (590ba7dc): `tests/romp-manager-tmux-scope.bats`, described under Tests.
- The `ROMP_CLI_SCOPE=0` floor for bats moved from three `setup()` blocks into `tmux_private_socket_dir` (same commit), so any suite that isolates tmux inherits it. `tests/tmux-private.bats` pins the export and `tests/README.md` records it beside the tmux and port floors. The new suite turns the switch back on after that call, behind the fake.
- The lost-tasks notice reads "N background tasks you had running were cut off when the claude process that started them ended (cause)" (ad1cac11): the session is named once, as "you", and the clause says whose process ended. The reconnect cause is reworded to read as that parenthesis ("a settings switch or a rewind restarted it"). The voice test pins the singular, plural and reconnect forms in full.
- The earlier body said `test_cli_scope.py` fails at import at the base. It collects; the per-test counts are above.

Second round. The review of 9fa583ac named three test follow-ups. All are folded by file into the commits where they belong, and the branch is rebased onto the current `main`.

- `tests/romp-manager-tmux-scope.bats` (590ba7dc): every absence check runs grep and tests its status (`log_lacks`), the form `romp-cli-scope.bats` uses. A bare `! grep -q` that was not a test's last command was exempt from bats' errexit and asserted nothing. Verified by mutation: a manager that logged the forbidden fragments on every path passed the old suite 3 of 3 and fails the new one on each of those three cases.
- `bin/romp-manager` (4975d38d): `startTmuxServer` pipes stderr on the scoped call and logs its first line, bounded, naming the tool that wrote it (`scopedStartFailure`, exported). With stdio `'ignore'` the line could carry only the "Command failed" argv, and it named systemd-run for a tmux failure inside the scope. Seven node cases cover the helper; the bats suite asserts the first stderr line in the log and not the second, and gains the case where tmux fails inside the scope. Against the previous revision's manager the bats suite fails 2 of 4 and the node suite 7 of 19.
- `tests/test_cli_scope_floor.py` (590ba7dc): pins both halves of conftest's floor on its source, described under Tests. With either half removed the matching pin fails; `tests/README.md` points to it.

## Points to weigh

- `ROMP_CLI_REAL` in the kernel's own environment: every session and its tool shells inherit it. It does no harm today; a later SDK that passes `options.env` to the version probe would make the export unnecessary, and a session's shell that sets `ROMP_SID` and runs the wrapper by hand gets a scope.
- The exec-in-place property relies on `systemd-run --scope` executing the command in the calling process, which every systemd since scopes existed (v205) does. Were systemd-run to fork instead, the CLI's pid would differ from the SDK's, `find_session_cli` would miss it, and the interrupt escalation would fall back to its existing behaviour. The bats exec-in-place test uses a fake systemd-run, so it proves the wrapper's `exec`, not systemd's.
- The bats hang-bound test needs coreutils' `timeout(1)`. It skips where that is absent, and the macOS matrix runs only on manual dispatch, so a regression there would surface late. `assert_direct_exec_in_place` also uses `$BASHPID` (bash 4 or newer; brew's bats-core brings bash 5).
- The lost-tasks notice is a user-visible injected string. `test_kernel_bundled_echo_prune.py` held it as a literal and is updated; `test_injected_voice.py` pins the sentences in full; `test_sdk_backend.py` and `test_sdk_boot_stagger.py` compare against the function's output and needed no change.
- The manager bats suite starts a real manager per case (four, on free ports, with a fake kernel launcher), like the two existing manager-driving suites. The scoped call pipes stderr for the failure line; the pipe closes when tmux's client exits (its server daemonizes onto /dev/null), so it holds nothing open, and the bare call keeps stdio `'ignore'`. The line attributes the failure by systemd-run's `Failed to …` log prefix, in both directions: a tmux error that began the same way would be attributed to systemd-run, and a systemd-run line without that prefix to tmux; the line itself is carried either way.

## Known and stated

The first service restart after this lands still empties the current cgroup, tmux servers included: the running manager and its tmux server predate the change. The guarantee holds from the following restart on. A scoped CLI the kernel's drain could not reach (kernel SIGKILLed at the stop timeout) outlives the restart and is terminated by the boot reaper from #941. Not in this PR: a postal bus autostarted from a session's MCP server starts inside that session's scope and keeps the `romp-session-*` unit loaded until the bus exits.

Tier: feature

### DIFF
--- a/bin/README.md
+++ b/bin/README.md
@@ -23,6 +23,7 @@ no separate implementation to point at.
 | `romp-node-launch` | POSIX sh | Runs the manager under a romp-owned copy of node (`romp-node`) so macOS TCC permissions can be granted to romp alone. |
 | `romp-manager` | Node | The kernel **supervisor**: spawns kernels via `romp-serve`, respawns on crash, `up/ensure/restart-all/status/down`. Reached via `romp up` / `romp refresh` / `romp status`. |
 | `romp-serve` | Bash | The manager→kernel seam: maps the manager's spawn contract onto the kernel's env, picks the python, then `exec`s the kernel (PID preserved for the supervisor; the kernel self-builds stale UI bundles). |
+| `romp-cli-scope` | POSIX sh | Linux + systemd only: the kernel spawns each session's `claude` CLI through it, and it `exec`s `systemd-run --user --scope` in place, so the CLI (and every tool shell, `setsid` child and tmux server it later starts) runs in a transient scope of its own instead of the manager service's cgroup, which a service restart empties. PID, parent and argv preserved; `ROMP_CLI_SCOPE=0`, or no `ROMP_SID` (a probe, not a session), runs the CLI directly, and so does a failed pre-flight scope, with one stderr line saying why, which the kernel logs as a problem at once. |
 | `romp-sdk-setup` | Bash | Provisions the Agent SDK venv for the SDK backend. Run by `install.sh`. |
 
 ## Symlinks → `kernel/` (the always-on core)
--- a/bin/romp-cli-scope
+++ b/bin/romp-cli-scope
@@ -1,0 +1,88 @@
+#!/bin/sh
+# romp-cli-scope — run one session's `claude` CLI inside its own transient systemd scope.
+#
+# WHY: on Linux the manager runs as a systemd user service that relies on systemd's default
+# KillMode=control-group (the unit sets no KillMode line, on purpose: it is what guarantees a wedged
+# kernel or postal bus still holding its port dies on a service restart). Every process the service
+# tree starts lands in the service's cgroup, and that includes each session's CLI, its tool shells,
+# their setsid children and any tmux server a session starts — so a service restart on 2026-09-05
+# (`systemctl --user restart romp-manager`) killed sessions' tmux servers and setsid jobs along with
+# the kernel. Only kernel-only restarts (`romp refresh`), which never touch the service, had left
+# them running. `systemd-run --scope` moves the CLI, and so everything it later spawns, into a scope
+# of its own outside the service cgroup.
+#
+# WHY exec-in-place matters: the kernel's interrupt escalation (find_session_cli in sdk_backend.py)
+# finds the CLI as a DIRECT CHILD of the kernel, and the boot orphan reaper (find_orphan_clis) reads
+# `--input-format stream-json` and the sid from the argv and treats a CLI whose parent is no live
+# romp kernel as orphaned. Both hold only if the CLI ends up with this wrapper's pid, its parent and
+# the argv tail the SDK passed: `exec` here hands our pid to systemd-run, and systemd-run in --scope
+# mode moves itself into the new scope and then execs the command in place, so the CLI keeps the pid
+# the SDK spawned and its parent stays the kernel. (An orphan re-parents to the `systemd --user`
+# subreaper, never to pid 1, scope or no scope — which is why the reaper keys on the parent not
+# being a kernel rather than on ppid 1.)
+#
+# The scope name carries the first 8 characters of the session id, this pid (the pid the CLI will
+# run as) and a start time, so `systemctl --user list-units 'romp-session-*'` doubles as a pid
+# record. systemd refuses a duplicate unit name ("already loaded"), and a previous CLI's scope can
+# outlive that CLI (a tmux server it started keeps the scope loaded), so once pids wrap a pid alone
+# could collide; the time component makes the name unique.
+#
+# Two direct-exec paths, no scope: no ROMP_SID (a process with no session identity is a probe — the
+# SDK's per-connect `claude -v` version check — not a session, and must not get a scope of its own),
+# and a failed pre-flight (`systemd-run … -- true`: the user bus gone, scopes refused), where one
+# stderr line names the reason and the CLI runs directly rather than not at all.
+#
+# Every stderr line this wrapper writes starts with `romp-cli-scope:`, and its second word says which
+# of the two messages it is, so the kernel tells them apart without reading the prose
+# (SdkSession._on_cli_stderr):
+#   `romp-cli-scope: fallback: …` — the failed pre-flight above; the CLI runs directly. The kernel logs
+#     this line the moment it arrives, as a problem naming the session. On this path the CLI starts,
+#     so no launch failure ever drains the stderr tail; without that the line would never have been
+#     read and the boot verdict would keep saying scopes are on. The launch-error card drops these
+#     lines from a failed CLI's stderr tail, since the log already has them.
+#   `romp-cli-scope: refused: …` — ROMP_CLI_REAL unset, exit 127, no CLI at all. The kernel logs no
+#     fallback for it: the launch fails, and the launch-error path reports the line on its own.
+# The rest of the CLI's stderr the kernel only buffers, for a launch that fails.
+#
+# Environment: ROMP_CLI_REAL (required) — the real CLI; ROMP_SID — the session id (kernel-set);
+# ROMP_CLI_SCOPE=0 — run the CLI directly, no scope. Arguments pass through verbatim.
+set -u
+
+if [ -z "${ROMP_CLI_REAL:-}" ]; then
+    echo "romp-cli-scope: refused: ROMP_CLI_REAL is unset or empty; it must name the real claude CLI, and this wrapper does not guess one" >&2
+    exit 127
+fi
+
+sid="${ROMP_SID:-}"
+if [ "${ROMP_CLI_SCOPE:-}" = "0" ] || [ -z "$sid" ] || ! command -v systemd-run >/dev/null 2>&1; then
+    exec "$ROMP_CLI_REAL" "$@"
+fi
+
+# Pre-flight: can this user manager start a scope right now? The kernel probed once at its own start,
+# but the bus can go away later, and a launch that fails outright is worse than one in the wrong cgroup.
+# Bounded to 10 s when `timeout` (coreutils) is on PATH: a bus that accepts the connection and never
+# answers holds systemd-run for its own 90 s default, past the SDK's 60 s initialize timeout, so the
+# fallback below would never engage in the case it exists for. Any non-zero exit, 124 (the bound hit)
+# included, is a failed pre-flight. Without `timeout` the pre-flight runs unbounded, as it did before.
+if command -v timeout >/dev/null 2>&1; then
+    err="$(timeout 10 systemd-run --user --scope --quiet --collect -- true 2>&1)"; rc=$?
+    [ "$rc" -eq 124 ] && err="the pre-flight scope did not finish within 10 s"
+else
+    err="$(systemd-run --user --scope --quiet --collect -- true 2>&1)"; rc=$?
+fi
+if [ "$rc" -ne 0 ]; then
+    nl='
+'
+    err="${err%%"$nl"*}"
+    echo "romp-cli-scope: fallback: systemd-run cannot start a transient scope (${err:-no detail}) — running the CLI directly, outside a scope; a service restart will take its background work down" >&2
+    exec "$ROMP_CLI_REAL" "$@"
+fi
+
+t="$(date +%s%N 2>/dev/null)"
+case "$t" in
+    ''|*[!0-9]*) t="$(date +%s)" ;;    # a date without %N prints the letter; seconds still separate pid reuse in time
+esac
+unit="romp-session-$(printf '%.8s' "$sid")-$$-$t"
+
+exec systemd-run --user --scope --quiet --collect --unit="$unit" \
+    --description="romp session $sid" -- "$ROMP_CLI_REAL" "$@"
--- a/bin/romp-manager
+++ b/bin/romp-manager
@@ -549,11 +549,95 @@ function runManager() {
 // prompt "Visual Studio Code wants to access…" — even with VS Code closed (its helpers linger). Starting
 // the server here at login fixes it permanently. `exit-empty off` keeps the (initially empty) server alive
 // until real sessions join; it's a no-op if a server is already running (the user 2026-06-16).
+//
+// Under the Linux service the server goes into a transient systemd scope of its own (2026-09-05). The
+// manager runs as a systemd user service that relies on systemd's default KillMode=control-group (the
+// unit sets no KillMode line), so a server started here lands in the service's cgroup and `systemctl
+// --user restart romp-manager` kills it, with every session's tmux jobs inside — a service restart that
+// day killed two sessions' tmux servers and setsid jobs along with the kernel. `systemd-run --scope`
+// starts the daemon outside the service cgroup; a session's later `tmux new-session` joins that server
+// and survives a service restart. The tmux CLIENT still exits at once (it only starts the daemon), so
+// execFileSync's semantics are unchanged, and when a server already runs the call is the same no-op and
+// the scope collects itself. Any failure of the scoped path falls back to the bare call. The macOS path
+// is as it was: launchd owns the lineage there and the TCC rationale above still applies.
+const TMUX_START_ARGV = ['tmux', 'start-server', ';', 'set', '-g', 'exit-empty', 'off'];
+
+// Is `name` an executable on the env's PATH? (Node has no `which`.)
+function commandOnPath(name, env = process.env) {
+  for (const dir of (env.PATH || '').split(path.delimiter)) {
+    if (!dir) continue;
+    try { fs.accessSync(path.join(dir, name), fs.constants.X_OK); return true; } catch { /* next dir */ }
+  }
+  return false;
+}
+
+// The PURE choice of how to start the server — exported for unit tests. null when tmux is not on PATH:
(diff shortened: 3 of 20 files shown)


ksarma:
Recommend merging after the comment window, once #941 and #951 have landed. Three follow-ups, none blocking.

Reviewed inline: `cli_scope_supported` answers "not Linux" before the PATH lookup and the probe. `tmuxStartArgv` returns the bare tmux argv off Linux. The macOS path is therefore unchanged by construction, and platform-injected tests that run on Linux CI pin both. `bin/romp-cli-scope` is mode 100755 in git and shellcheck-clean. It execs through `systemd-run --scope`, which runs the command in place, so the CLI keeps the pid, parent and argv the SDK spawned. `find_session_cli` and `find_orphan_clis` need no change. The `ROMP_CLI_REAL` export into the kernel's own environment matches the installed claude_agent_sdk 0.2.132, whose version probe inherits `os.environ` rather than `options.env`. `tests/conftest.py` floors `ROMP_CLI_SCOPE=0` at import and again per test. The three bats suites that start the real manager export it in `setup()`. The transient-unit count on this machine's user manager was identical before and after every local run, so no test started a real scope. The reworded lost-tasks notice passes the injected-voice check and agrees in number. It asks the session to check whether each task still runs before relaunching it. The 1332 added lines carry no private strings, hostnames, home paths or the banned word.

Follow-ups:

- `kernel/sdk_backend.py:147`: file the boot-time wanted-but-unavailable verdict with `problem=True` when the switch was wanted on Linux, as launch already does. Today that verdict (supervised, Linux, systemd-run missing or the probe failing) is a plain log line and no problem entry. The error center therefore never shows that sessions are running inside the service cgroup.
- `bin/romp-manager:593`: add a bats case with a fake `systemd-run` first on PATH and `ROMP_CLI_SCOPE=1`. It covers both branches without touching the user manager. `startTmuxServer`'s scoped call, its fallback to the bare call and both log lines have no test. Only the pure chooser is covered, and the manager-running bats suites floor the switch off.
- `tests/README.md:19`: record the new bats floor (export `ROMP_CLI_SCOPE=0` in any suite that starts the real manager) beside the tmux and port floors, or carry it in `tmux_private_socket_dir`. It lives only in `setup()` comments of three suites.

Two nits:

- The body says `test_cli_scope.py` fails at import at the base. It collects, and 25 of 28 tests fail per test; the red-first requirement is met either way.
- The notice says "session" twice in one clause, and "were cut off when its claude process ended" reads the same.

Locally at head: the PR's five Python files 118 passed. The full pytest suite 7373 passed, with 5 xdist-ordering failures in two untouched files that pass alone. The bats suite `romp-cli-scope` passed 19 of 19; the manager-driving suites 186 of 186 under a private socket directory. Node manager suites 52 of 52; the new file 12 of 12. At the stack base the new Python, node and bats tests fail as claimed. CI: all six required checks green.

Review run by the maintainer's automated review session; verified against head 9fe25325.

ksarma:
Recommend merging after the comment window. There is one test follow-up to fold; nothing is blocking.

Reviewed inline: the five folds from the previous review are in the diff as described. `cli_scope_supported` passes `problem=True` to its log callback only when the switch was wanted on Linux and systemd-run is missing or the probe fails (`kernel/sdk_backend.py:154-159`). The flag is asserted on every branch of the truth table, and the entry is checked in `problems()`. `task_death_notice` says the tasks were cut off when the claude process that started them ended. It names the session once and passes `tests/test_injected_voice.py`. `tests/romp-manager-tmux-scope.bats` starts the real manager with a fake systemd-run first on `PATH`. It covers the scoped argv, the bare fallback and both log lines. `tmux_private_socket_dir` exports the `ROMP_CLI_SCOPE=0` floor (`tests/tmux-private.bash:36`); `tests/tmux-private.bats` pins it and `tests/README.md` records it. The PR body's base-line count for `test_cli_scope.py` (27 of 30 fail) matches a scratch run at origin/main. `bin/romp-cli-scope` is 100755 in the index. No test started a real scope: the `romp-` scope units on the user manager were identical (33) before and after the full pytest, node and bats runs. That check ran on this systemd-hosted machine, from a shell that inherits `ROMP_SUPERVISED`. Privacy and vocabulary scans of the added lines and commit messages are clean.

Follow-ups:

- `tests/romp-manager-tmux-scope.bats:94` (also 95, 110, 125): a bare `! grep -q` that is not the last command of a bats test is exempt from errexit and asserts nothing. A manager mutated to log the forbidden lines on each path still passes 3 of 3. Lines 96, 111 and 126 assert only because they come last. Use the run-then-status form this PR already uses at `tests/romp-cli-scope.bats:149`.
- `bin/romp-manager:605`: with stdio `'ignore'` the failure line can only carry the Command failed argv, never systemd-run's or tmux's stderr. It also names systemd-run when tmux itself fails inside the scope. Pipe stderr and log its first line (the PR body notes this gap).
- `tests/conftest.py:121`: no test fails if the pytest floor is dropped, because `tests/test_cli_scope.py:36` sets `ROMP_CLI_SCOPE=0` itself at import. Pin the floor with a test that does not set the value.

Tests:

- At head: pytest 81 passed with 190 subtests across the four changed Python files (full suite 7413 passed, 40 skipped); `node --test tests/manager-*.test.js` 52 of 52; bats 36 of 36 across the five changed shell suites (425 in full).
- At origin/main with the PR's tests checked out: `test_cli_scope.py` fails 27 of 30, `test_sdk_launch_error.py` errors at collection, the voice test fails on the old wording, `manager-tmux-scope.test.js` fails 12 of 12, `romp-cli-scope.bats` fails 19 of 19, and the manager bats suite fails its two scoped cases.
- At the last pre-fold commit dc09ee55: the six fold-specific tests fail.
- CI: all six required checks green at 9fa583ac.

Review run by the maintainer's automated review session; verified against head 9fa583ac.

ksarma:
Recommend merging after the comment window. Nothing to fold.

Reviewed inline: all three folds from the last round hold. `tests/romp-manager-tmux-scope.bats` has no bare negations left and sixteen `log_lacks` run-then-status checks. A manager mutated to log every forbidden fragment fails the suite 4 of 4. The unmodified manager passes 4 of 4. `startTmuxServer` pipes stderr on the scoped call only. A real tmux start through an exec-in-place fake `systemd-run` returned in 13 ms with the server's stderr on `/dev/null`. A grandchild holding the pipe costs a bounded 5 s ETIMEDOUT and survives under stdio ignore too. The pipe therefore adds no hang and no leak. `tests/test_cli_scope_floor.py` fails 3 of 3 with both conftest floor lines removed, 2 of 3 beside `test_cli_scope.py`, and passes at head. The earlier rounds' items are still in place. `cli_scope_supported` files `problem=True` for the wanted-but-unavailable verdict. The injected-voice test passes on the reworded notice. `tests/README.md` records the floor. `bin/romp-cli-scope` is 100755 and shellcheck-clean. The `romp-` scope count on a live systemd install was 14 before and after every run. The 1,777 added lines contain no private strings, hostnames or home paths.

Two nits, neither blocking:

- `bin/romp-manager:639` logs the server as unscoped after an ETIMEDOUT on the scoped call. tmux may already have daemonized inside the scope, so the line can predict a loss that will not happen.
- `tests/romp-cli-scope.bats:185` spends a real 10 s and asserts a 9 to 16 s wall-clock window that a loaded runner could miss. A fake timeout exiting 124 would cover the same branch in milliseconds.

Locally at head, the five touched pytest files report 84 passed with 190 subtests. Node manager suites pass 59 of 59. Bats: romp-cli-scope 19 of 19, tmux-private 7 of 7, romp-manager-tmux-scope 4 of 4. All six required checks are green at this head.

Review run by the maintainer's automated review session; verified against head 590ba7dc.

PR #954, opened 2026-09-06
The parked-op queue is mutated under one lock that no backend call ever holds

lczh:
Since the drain moved onto the pusher cycle (#904) it runs every 0.5 s and on every wake, on the pusher thread, while the WS and HTTP handler threads park and cancel ops in the same dict and the move thread re-inserts a head. Nothing serialized them. Adversarial review confirmed two races:

- A cancel, or the move thread's re-insert, could change a sid's list between the drain's head read and its pop. The drain then popped a list the cancel had just emptied (an `IndexError` its blanket `except` turned into a dropped queue), or the cancel's own check-then-pop landed on a list the drain had shifted and removed the op *behind* the one clicked.
- The disk mirror was written through one fixed temp path from any thread, so two writers could tear it and the next boot restored a wrong queue.

## What changes

**One RLock around every mutation** of the queue: the park, the cancel, the move thread's head re-insert, the drain's reads and pops, and the mirror write (now through `_atomic_write`'s per-writer temp name, so every snapshot is whole).

**The lock never spans a backend call.** `CodexBackend.send` can synchronously spawn and initialize the app server with no request timeout, and the SDK's `busy` and `turn_seq` take the backend's own lock, so a lock held across any of them would hold every handler's park behind it and could nest inside a backend lock. The drain takes the lock only to read a sid's head and to pop what a step delivered, releasing before the send, the setter or the `turn_seq` read. A raise still drops that sid's queue once, logged, exactly as before.

**The head stays the visible head while the backend has it.** Every handler gate keys on queue presence and otherwise on `busy`, which flips only once the backend has registered the op (the SDK backend runs `_ensure` under its own lock before it enqueues; the Codex backend may spawn the app server first). Popping before the call would leave the queue empty for that window, and a pane send landing in it would hand over *ahead* of a parked slash command. So a command, compact or settings op is read under the lock and recorded in flight, the call runs with the lock released and the op still at the head, and afterwards the head is popped only if it is still that very op. A cancel on the in-flight head is refused as too late rather than reported as a cancellation of something the backend already has; the refusal keys on the head slot, not identity alone, because a compact parks one interned tuple per code object. A move op is not recorded in flight (its only mid-call action is the `turn_seq` read), so a cancel on a waiting move still succeeds. The send run alone is popped before delivery, as the parent already did.

**Handler side:** the expensive gates (tmux fork, discover sweep, usage read, `busy`) run outside the lock as they always did; only the cheap queue-presence check and the park run under it, as one step (`_gate_or_park` / `_park_behind_queue`), so an op never slips past a queue another handler filled between the two reads. The top-of-walk acquire is non-blocking, so a cycle never stalls every session's push behind a handler.

Unchanged: the failure contract, park order and the sequential delivery rule, every gate and hold and where the hold is armed relative to the pops, and the rule that only a real mutation saves the mirror and wakes the pusher.

## Tests

`tests/test_kernel_parked_ops_lock.py` (17 tests) drives each race and each invariant: a cancel arriving while the backend has the head (in-flight one too late, the one behind cancelled, nothing dropped); a cancel paused between its body check and its pop while a drain cycle lands (event-keyed, no sleeps); a pane send issued from *inside* the backend's own call that must park behind the in-flight command (passes on `main`, fails when the head is popped before the call); a same-kind pick replacing the in-flight head; a second compact press cancelled while the first is with the backend; a waiting move cancelled from inside the `turn_seq` read; the locked re-check parking behind a queue the expensive gates missed; a backend that asserts from inside every call the drain makes that the drain does not own the lock and another thread can take it; and eight threads hammering the mirror. Source mutants that drop the locked re-check, the identity pop, the in-flight record or the head-slot refusal each fail exactly one test. Two setter source pins follow the setters' new call. The four test modules an earlier cut had rewritten are byte-identical to `main`.

Full suite 7346 passed / 44 skipped / 0 failed with nothing deselected; `bats tests/*.bats` 389/389.

## Notes for review

- **#923** (a model pick applies live mid-turn) edits the same lines of `_set_model_or_park`: it widens the gate but still reads the queue unlocked and hands over directly, which is the handler race this closes. Whichever lands second rebases; the shape here (gate function outside the lock, queue check + park inside) takes #923's wider gate unchanged.
- **#955** stacks on this one: it moves the drain's per-sid transcript-parse refresh out from under the lock to sit among the other gates that already run outside it, and reads the usage file once per drain instead of once per op. Its diff shows both commits until this merges.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -19858,6 +19858,26 @@ def _session_chip(sid, path, session, tm, now):
 # the moment compaction ends. A repeated model/effort pick REPLACES its earlier parked op in place (last
 # pick wins — one queued chip, not a pile), since only the final setting can matter.
 _PENDING_OPS_FILE = Path(jd.STATE) / "pending-ops.json"
+# ONE lock around every MUTATION of _pending_ops (2026-09-05): the WS/HTTP handler threads park and cancel,
+# the move thread re-inserts a head, and the pusher thread drains — all on this dict, and since the drain
+# moved onto the pusher cycle (#904) it runs every 0.5 s and on every wake. Unlocked, two races were
+# confirmed by review: (1) a ✕ — or the move thread's head re-insert — changed a sid's list between the
+# drain's head read and its pop, so the drain popped a list the ✕ had just emptied (an IndexError the
+# blanket except turned into a dropped queue) or the ✕, re-locating by index+body and then popping,
+# removed the op BEHIND the one clicked once the drain's pop had shifted the list; (2) two writers tore the
+# disk mirror through one shared temp path (a kernel death then restored a wrong queue). The lock guards the
+# mutations ONLY: no backend call ever runs under it (CodexBackend.send can synchronously spawn and
+# initialize `codex app-server` with no request timeout; SdkBackend.busy/turn_seq take the backend's own
+# lock), and a handler evaluates only the queue-presence check under it (_park_behind_queue) — its
+# expensive gates (_compacting_now / _working_now / _limit_hold: a tmux fork, a discover sweep, the usage
+# file) run outside. So a holder owns it for a dict read, a list append or pop, and the mirror write —
+# microseconds. Re-entrant, because _park_op and _save_pending_ops take it and are called from under it.
+# The drain takes it NON-BLOCKING at the top of its walk (_apply_pending_ops). No backend thread ever takes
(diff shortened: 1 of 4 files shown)


lczh:
Tier: **fix** (a bug fix with tests that fail before it). A fork author cannot add labels here (`AddLabelsToLabelable` is refused), so could a maintainer apply the `fix` label? Same applies to the stacked #955.

PR #957, opened 2026-09-06
The pane shim raises the stale prompt when it abandons a silent reconnected socket, and the drop bell names the pane by its label

ksarma:
## What breaks

Follow-up from the review of #952, which found this gap and the bell-row nit below (that PR shipped as reviewed; both are folded here).

The stale prompt has two raising events: the second keepalive landing on the reconnected socket before its resync, and that socket closing before its resync. `abandon()`, which the watchdog and the foreground path call on a socket quiet past 30 s, nulls `onclose` before closing it, so the close rule never runs for an abandoned socket, and the redial's open re-arms from zero. A kernel that accepts the reconnect and then never speaks on it (no keepalive to count, no close to rule on) flaps the loader every 30 s and never raises the prompt. The 1 s timer this rule replaced raised on the first cycle.

Reproduction: `ui/webview/pane-shim-stale.test.ts` now captures the shim's `setInterval` and ticks the watchdog by hand. Against the shim before this change, three silent cycles produce three `watchdog-close` rows, three redials and zero raises.

## The fix

`abandon()` runs the close rule itself: an armed socket it gives up on retires the arm and raises with the arming path plus `-quiet` (`reconnect-quiet`, `foreground-quiet`). The review suggested running the rule before `abandon()` at its two call sites; it lives inside `abandon()` instead, a deliberate departure: abandon's contract is to do what the disowned `onclose` would have done, so every caller is covered by construction (the call-site pins in `test_kernel_disconnect_banner.py` are unchanged), and running the raise after `ws` is nulled lets its breadcrumb queue and ride the reconnect instead of going down the quiet socket.

The `onclose` comment and `docs/read-side.md` said every close of an opened socket leaves a `wsclose` row. An abandoned socket leaves none: the watchdog's own `watchdog-close` row goes down the quiet socket before the abandon (the foreground path's abandon sends no row of its own), and for an armed socket the `-quiet` raise, queued for the redial, is the record that survives. Both now say so.

Second commit: `_note_ws_drop` formatted the raw app id into the bell text, so a dropped Outline pane was named by its internal id. The row now maps the id through `_PANE_ORDER`; an app with no rail label keeps its id. The stderr line keeps the id on purpose: it is the log's word for the pane, shared with `_drop_dead_ws_client`.

## Tests

Red first: every new or changed assertion was run against the unchanged code before the fix.

- `ui/webview/pane-shim-stale.test.ts`: one silent cycle raises once (`reconnect-quiet`), the breadcrumb rides the redial instead of going down the quiet socket, the abandoned socket leaves no `wsclose` row, and the redial's resync retires the prompt; three silent cycles raise once each, with one `watchdog-close` row each and no `wsclose`; the foreground path abandoning an armed socket raises too, and its redial arms as `foreground`. Against the unchanged shim all three fail at the first raise count (expected 1, got 0).
- `tests/test_kernel_disconnect_banner.py`: pins the rule's text in `abandon()`; fails against the unchanged shim.
- `tests/test_ws_drop_loud.py`: drops a client of every `_PANE_ORDER` app and checks each bell row carries the label, the Outline row never carries the id while its log line still does, and an unlabeled app keeps its id. Three existing assertions follow the labels. Four failures against the unchanged code.
- Runs at head: `npm test` 2833 passed and `npm run typecheck` clean; `pytest tests/test_ws_drop_loud.py tests/test_kernel_disconnect_banner.py tests/test_kernel_ws_heartbeat.py tests/test_restart_seam.py tests/test_injected_voice.py` 52 passed; the full `pytest tests` 7372 passed, 46 skipped.

Tier: fix

### DIFF
--- a/docs/read-side.md
+++ b/docs/read-side.md
@@ -55,22 +55,29 @@ completed); the feed just paints columns. (Reflected in `docs/judges.md`.)
   consistent by construction. Keeping the WS protocol stable is the compatibility
   contract.
 - **Liveness is a keepalive, and staleness is event-keyed.** The kernel sends a
-  `ka` frame to every socket every 10 s; the pane shim force-closes a socket that
-  has gone 30 s without any frame. After a reconnect the shim raises the "what you
-  see may be stale" prompt only on the SECOND `ka` arriving before the resync
-  frame — one full heartbeat period, bracketed by two kernel heartbeats on that
-  socket with no resync between them (a single `ka` can be a beat that was already
-  queued when the socket was accepted) — or on the reconnected socket closing
-  again before its resync; the first non-keepalive frame retires the prompt, never
-  a timer. A client that falls 16 MB behind is dropped by the kernel, loudly: one
-  `ws: dropping` line in the kernel log, written where the drop is decided so every
-  send path is covered, naming the pane, the dashboard, the backlog and the push
-  slot whose frame tipped the budget when a push did; and a row in the dashboard's
-  bell (at most five drop rows, none older than an hour, so they never crowd out a
-  backend problem). Every close of a socket that opened leaves a `wsclose`
-  breadcrumb (code, reason, socket age) in `client-diag.jsonl`; the redials an
-  outage refuses are counted and reported as one `wsconnfail` row on the next
-  open, and at most 20 breadcrumbs wait in the shim's queue for it.
+  `ka` frame to every socket every 10 s; the pane shim abandons a socket that has
+  gone 30 s without any frame and redials at once. After a reconnect the shim
+  raises the "what you see may be stale" prompt only on the SECOND `ka` arriving
+  before the resync frame — one full heartbeat period, bracketed by two kernel
+  heartbeats on that socket with no resync between them (a single `ka` can be a
+  beat that was already queued when the socket was accepted) — on the reconnected
+  socket closing again before its resync, or on the shim abandoning it as quiet
+  before its resync (a kernel that accepts the reconnect and never speaks on it
+  sends nothing to count and nothing to close); the first non-keepalive frame
+  retires the prompt, never a timer. A client that falls 16 MB behind is dropped
+  by the kernel, loudly: one `ws: dropping` line in the kernel log, written where
+  the drop is decided so every send path is covered, naming the pane, the
+  dashboard, the backlog and the push slot whose frame tipped the budget when a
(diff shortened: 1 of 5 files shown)


ksarma:

Reviewed inline: `abandon()` now runs the stale close rule itself, guarded by `stalePending` and `openSock===d`. The rule runs after the socket's handlers are detached, the socket closed and `ws` nulled. The rule belongs in `abandon()`, not at its two call sites. That is the one place that disowns `onclose`, so the watchdog tick, the foreground path and any future caller are covered by construction. Raising after `ws` is nulled makes `send()` queue the breadcrumb (capped by `DIAG_QUEUE_MAX`) for the redial's `onopen` to flush. The row is therefore not written to a quiet socket that might never deliver it. No double raise is possible: both callers abandon only a `readyState` 1 socket, `onclose` is nulled before `close()`, and every raise path clears `stalePending` first. The redial re-arms through `armStale(pendingWhy||"reconnect")`, and the first non-keepalive frame retires the prompt through `clearStale`. A foreground abandon of a reconnect-armed socket therefore raises `reconnect-quiet`, and its redial arms as foreground. The `onclose` comment and `docs/read-side.md` now say only browser-reported closes leave a `wsclose` row, which matches `send()` writing straight to an open socket. The bell row maps the app id through `_PANE_ORDER` at call time, with the id as fallback. The stderr line keeps the id.

- `ui/webview/pane-shim-stale.test.ts:274`: the body, docs and shim comment name `foreground-quiet`, but every `-quiet` why the file asserts is `reconnect-quiet`. A test that arms a redial as foreground and ticks it silent would pin that path.
- `kernel/kernel.py:31264`: the shell's bell JS keeps its own map of the same four pane labels (`PN`), so the drop row and the connection-lost row name panes from two lists. `PN` could derive from `_PANE_ORDER`.

Red-first at origin/main with the PR's test files checked out: pytest on `test_kernel_disconnect_banner.py` and `test_ws_drop_loud.py` gave 5 failed, 17 passed (the new `-quiet` source pin plus four label assertions). `pane-shim-stale.test.ts` failed tests 8, 9 and 10 at the first raise count (expected 1, actual 0). At head: 52 passed across the five pytest files the body names, 13 of 13 in the shim test, npm test 2833 passed. All six CI checks are green.

Review run by the maintainer's automated review session; verified against head 51eaec7a.

PR #958, opened 2026-09-06
A same-origin markdown link opens in the file viewer, rendered, with its figures resolved against the document

alowet:
## What

A chat message linking a markdown file on the dashboard's **own origin** — a published report or evidence doc served beside romp, e.g. `https://<host>/figs/<run>/evidence.md` — used to open as raw text in a new tab. It now opens in the pane's existing file viewer, rendered, exactly like a local `.md` path already did: same modal, same Rendered ⇄ Raw preference, same loader-first wait, same `mdBlock` (marked + DOMPurify + hljs).

- **One viewer for all markdown.** `openUrlView(href)` is a URL mode of the file viewer; the chat's anchor delegate routes an unmodified click on a same-origin `.md`/`.markdown` link there. Bar: dimmed `host/dir/` + basename, Rendered / Raw, `Open ↗` (the raw URL, new tab), `Copy URL`, ✕; Esc closes.
- **Same origin only — cross-origin links are unchanged.** The browser fetches the URL itself, so a cross-origin fetch would be a CORS guess; a `.md` on any other origin keeps today's new tab exactly (no try-then-fall-back, no regression for `github.com/…/blob/…/README.md` pages that merely end in `.md`). Ctrl/⌘/Shift-click still opens the tab.
- **Zero new kernel surface.** No route, no relay, no server-side fetch: the kernel's `/file` relay stays a preview relay (`_remote_file`'s docstring).
- **Relative references resolve against the document**, in both modes: `fig` beside a URL document fetches from the document's directory (and the *response* URL, so a redirected alias resolves correctly); beside a local file it rides `fileUrl()` to the sibling on disk. A sibling `.md` link opens in the same viewer; in-document `#fragment` links land on `md-`-prefixed heading ids inside the modal instead of spawning a tab.
- **Loud failures, bounded reads.** `HTTP <status> from <host>`, could-not-load, and the kernel's 2 MB text cap mirrored — refused from `Content-Length` when declared, else by streaming the body and counting bytes as they arrive (cancelled one chunk past the cap). Closing or replacing the modal, or refusing a response, aborts the transfer.

## Why

Sessions publish their results as `.md` reports next to their figures; reading them meant a raw-text tab. Since the local-path case already rendered in the viewer, extending that one surface (rather than a new tab page or a kernel-side renderer) keeps one code path, one set of styles, and the user in the dashboard.

## Verification

- `npm run typecheck`, `npm test` (2838 pass), `npm run build`; `tests/test_file_view.py` + `tests/test_kernel_preview.py` (42 pass; no kernel change).
- Executed unit tests for every pure helper (`isMarkdownUrl`, `resolveDocRelative`, `joinDocPath`, `urlTitleParts`, `headingSlug`/`uniqueSlugs`, `readTextCapped`, `settleUrlResponse`, `overCapWords`); source pins for the DOM wiring in `md-url-view.test.ts`.
- Two rounds of adversarial review, each finding fixed in its own commit (streamed byte-counted cap with abort on teardown; `Response.url` as the document base; modifier clicks; refused responses abort the transfer; linear slug dedupe).
- Headless Chromium smoke against a throwaway kernel serving the built bundle: same-origin `.md` renders with figures resolved, sibling and fragment links navigate in place, cross-origin and modified clicks reach `window.open`, redirect resolves from the response URL, and a real chunked no-`Content-Length` server showed the reader disconnecting one 64 KiB chunk past the cap and refused responses disconnecting after 131 KB.

## Not in this PR

- Cross-origin `.md` links (a peer kernel's host, raw.githubusercontent.com) still open raw in a tab. Rendering them needs either CORS on the file server or a kernel relay — the latter is an SSRF trade-off the repo has so far declined, so it is left as a separate decision.
- The VS Code webview path is untouched (`openLink` → `openExternal`).

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -46,6 +46,13 @@ clone writes its tracking ref; where nothing local could refresh the answer (a
 `--single-branch` clone, or a branch on origin this clone has not fetched) origin is asked on
 each open.
 
+**Opening a markdown document.** A markdown link in the chat opens in the file viewer,
+rendered, with **Raw** one click away — a path on the session's machine, or a link to a
+file served from the dashboard's own address (a published report, an evidence doc). Figures
+and links inside the document resolve relative to the document, so a `![fig](fig.png)`
+beside it shows, and a link to a sibling document opens in the same viewer. Links to files
+on other sites open in a new tab, as before.
+
 ### The feed
 
 The feed is Romp's task-management layer: a card for each task. Romp's
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -51,7 +51,8 @@ rendered, with **Raw** one click away — a path on the session's machine, or a 
 file served from the dashboard's own address (a published report, an evidence doc). Figures
 and links inside the document resolve relative to the document, so a `![fig](fig.png)`
 beside it shows, and a link to a sibling document opens in the same viewer. Links to files
-on other sites open in a new tab, as before.
+on other sites open in a new tab, as before — and a ctrl- or ⌘-click still opens the file in
+a tab.
 
 ### The feed
 
--- a/ui/webview/capped-read.test.ts
+++ b/ui/webview/capped-read.test.ts
@@ -4,7 +4,7 @@
 // these run the real reader. Fixtures are invented bytes; nothing here is recorded data.
 import { test } from "node:test";
 import * as assert from "node:assert/strict";
-import { readTextCapped, humanSize, overCapWords } from "./capped-read";
+import { readTextCapped, humanSize, overCapWords, settleUrlResponse } from "./capped-read";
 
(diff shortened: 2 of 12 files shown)


alowet:
Screenshots from the headless-Chromium smoke (throwaway kernel serving the built bundle; synthetic fixture document at a loopback URL):

**Dark — a same-origin `.md` link opened in the pane's file viewer** (title bar: dimmed `host/dir/` + basename; Rendered / Raw; Open ↗; Copy URL; ✕):

markdown viewer, dark

**Light theme**, same document:

markdown viewer, light

**In-document fragment link landed** (`[results](#results)` scrolls the modal to the `md-results` heading, 10px below the bar, instead of opening a tab):

fragment landed

PR #959, opened 2026-09-06
A PDF opens in its own browser tab, in the browser's viewer, with the in-app view kept as the fallback

lczh:
## What

When a session produced a PDF, the dashboard showed it in a small window inside itself: the chat's PDF card opened a lightbox with an iframe, and a `.pdf` link or file-browser row opened the in-pane viewer. This makes a PDF open **in its own browser tab, in the browser's viewer**, the way a paper opens from OpenReview or HotCRP: full size, kept open beside the dashboard while the work goes on.

Those sites do nothing more than link the PDF's URL with a new-tab target and serve it inline as `application/pdf`; the browser renders it from its own cache and nothing lands on disk unless the browser is set to download PDFs. The kernel's `/file` route already serves exactly that, same-origin and cookie-authed, with a federated session's file relayed through the same URL builder.

- **`ui/webview/preview.ts`** — `openPdfTab(path, sid)`: one synchronous `window.open(fileUrl(path, sid), "_blank")` inside the click gesture, so a popup blocker reads it as user-initiated (a fetch-then-open would lose the gesture). The kind check is the opener's own, by extension, the only fact available inside the gesture; the viewer's media branch keeps keying on the kernel's `Content-Type` verdict. The opener is severed on the handle (`w.opener = null`) rather than with the `noopener` feature, because that feature returns `null` even on success and the null handle is the one signal that the popup was **blocked**; on that signal the card falls back to the lightbox and the viewer to its in-pane frame, so the PDF is never unreachable. `openPdf()` = tab, else lightbox; the PDF card's click uses it.
- **`ui/webview/file-view.ts`** — `openFileView` starts with `if (openPdfTab(path, sid ?? null)) return;`, so every path into the viewer (a path link, a file-browser row, a relayed open) takes the tab first.
- **`kernel/kernel.py`** — a PDF is served with `Content-Disposition: inline; filename="<basename>"` on GET and HEAD, locally and through the `/remote/<host>/file` relay (derived from the requested name, never mirrored from the remote), so the tab is titled after the file and a save names it. Images are unchanged.
- **Not a dead end when the file is too large.** A PDF over `_PREVIEW_MAX_BYTES` used to reach the viewer, whose 413 handling offered a Download button; in a tab of its own it would have shown only the refusal. A *navigation* (`Sec-Fetch-Dest: document`) to such a PDF now gets a small page with the same sentence and a link to the download half of the same route (local or relay); fetches, iframes and HEAD probes keep the plain-text 413 they parse. The directory listing's `viewable` verdict is now size-aware (media cap / text cap), so the file browser routes an oversize row to the download up front, and its tooltip says "opens as a download" rather than "not viewable".
- **Relayed error verdicts are prose.** `_remote_file` used to hand a remote's 404/413/415 text back under the local media mime; an `` merely failed on that, a tab shows a corrupt-PDF error. Every non-success relayed verdict is `text/plain` now.
- **`docs/guide.md`** — an "Opening a PDF" paragraph.

Web dashboard only: the VS Code webview cannot open a window and never rendered previews (routing a PDF to the OS browser from there would put the serve token in an OS-level URL, so it is deliberately left out).

## Review

Two adversarial passes with a refuter per finding, then a skeptic pass over the fixes. Five real problems in the first cut, all fixed here and each pinned by a test that fails on the old code: the live opener (reproduced in a real Chromium: a link inside the PDF navigates the same tab to a foreign site, which could then replace the dashboard with a look-alike); the oversize-PDF dead end; the relay's missing inline header; the relay's mislabelled error prose; the browser tooltip's wrong reason.

## Tests

`ui/webview/pdf-new-tab.test.ts` runs the opener against a stubbed window (permitted popup, blocked popup, webview origin, non-PDF path, federated relay URL, severed opener) and pins the wiring of the card, the viewer, the kernel header, the oversize page, the listing's verdict, the relay and the guide. `tests/test_kernel_preview.py` pins the inline disposition on GET and HEAD, its absence on images, and the oversize page with its escaping against a navigation only. `tests/test_listdir.py` pins the size-aware verdict. `tests/test_kernel_remote_file_relay.py` pins the relay's locally derived header against a remote that lies, its error verdicts as prose, and the oversize page on the relay route. The inline-preview and file-view pins follow the new click and import.

Full suites on this tree: Python 7393 passed / 40 skipped; node 2842; bats 402.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -55,10 +55,12 @@ on other sites open in a new tab, as before — and a ctrl- or ⌘-click still o
 a tab.
 
 **Opening a PDF.** A PDF the session mentions, or one you click in the file browser, opens
-in a new browser tab in the browser's own viewer, the way a paper opens from OpenReview:
-full size, and it stays open beside the dashboard while you keep working. If the browser
-blocks the new tab, the PDF opens in the in-app viewer instead; a PDF too large to show
-offers a download in its place.
+inside the dashboard like an image: the chat's PDF card opens it full-view, a path or a
+file-browser row opens it in the file viewer. Cmd-click it instead (Ctrl on Windows and
+Linux), or middle-click, and it opens in a new browser tab in the browser's own viewer, the
+way a paper opens from OpenReview: full size, and it stays open beside the dashboard while
+you keep working. If the browser blocks that new tab, the PDF opens inside the dashboard
+instead; a PDF too large to show offers a download in its place.
 
 ### The feed
 
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -54,6 +54,12 @@ beside it shows, and a link to a sibling document opens in the same viewer. Link
 on other sites open in a new tab, as before — and a ctrl- or ⌘-click still opens the file in
 a tab.
 
+**Opening a PDF.** A PDF the session mentions, or one you click in the file browser, opens
+in a new browser tab in the browser's own viewer, the way a paper opens from OpenReview:
+full size, and it stays open beside the dashboard while you keep working. If the browser
+blocks the new tab, the PDF opens in the in-app viewer instead; a PDF too large to show
+offers a download in its place.
+
 ### The feed
 
 The feed is Romp's task-management layer: a card for each task. Romp's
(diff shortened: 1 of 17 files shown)


lczh:
Tier: `feature` — a fork author cannot add labels here, so asking a maintainer to apply it. A self-contained new capability inside the existing model (a PDF opens in its own browser tab); the design points are in the body.

PR #965, opened 2026-09-07
Tests: the relay-honesty disk-twin test seeds its own state directory, and six follow-ups the #941–#957 reviews named

ksarma:
Seven tests-only commits, one per item, under `tests/` and `ui/webview/*.test.ts`; no product code changes. The first fixes a test whose result depended on which tests ran before it. The other six answer follow-ups named in the merge reviews of #941, #946, #947, #949, #953 and #957.

## The ordering-dependent test

`tests/test_postal_relay_honesty.py::PresenceBlinkHonesty::test_disk_twin_carries_the_cache_across_a_bus_restart` fails whenever it runs without an earlier test in the same process having created the postal state directory: selected alone by node id or `-k`, as its class alone (1 failed, 4 passed), or on a pytest-xdist worker that did not first run one of the relay ack tests (4 of 5 `-n 8` runs of the module at the base commit). The module in file order passes (30 passed). This is the pre-existing ordering failure the verification notes of #887, #944 and #950 mention.

Why: the presence twin lives at `STATE/local-presence-good.json`. `_local_presence` writes it inside a `try/except` that swallows every failure (deliberately: the twin is best-effort), and the directory is created only by `serve()` and `ensure()`, which these tests never call. Every explicit mkdir in the module (the session-flags parent, `names/`, `sdk/`, the timeline log) lands beside `postal/`, not inside it. In file order the directory appears as a side effect of three `IdShapedAddressingRoutes` tests whose `_relay_in` reaches the ack branch and calls `peer_seen_add`, which mkdirs `PEER_SEEN`'s parent, the same directory. So the disk-twin test's first answered read silently wrote nothing, and the second read found no twin.

Fix: `PresenceBlinkHonesty.setUp`, which already resets the in-memory cache and unlinks the twin, now also creates the twin's directory, the one thing `serve()` would have done. The test additionally asserts the twin exists after the answered read, so a future regression of the write is reported at the write with its own message rather than at the second read. Removing the seed while keeping that assertion fails at "an answered read writes the disk twin" (checked, then restored). The swallowed write in `_local_presence` is correct for a running bus, where the directory exists before the first listing, so no product code changes.

## The six follow-ups

- **#941**: the lifecycle-hardening header no longer claims ps is patched in every test; `PsArgv`'s two real-ps tests (Linux-only, against a sleeper child they spawn and kill) are named as the exception. Comment-only.
- **#946**: a comment at the child-returncode assertion in `test_session_move_live.py` says the failure message embeds the child's stderr, which can carry the helper's own diagnostics. An always-run `ApiKeyHelperPrecedence` class covers the explicit `$ROMP_MOVE_LIVE_API_KEY_HELPER` winning, the command borrowed from `$CLAUDE_CONFIG_DIR/settings.json`, the `~/.claude` default, and `""` when nothing names one, with `HOME` and `CLAUDE_CONFIG_DIR` under a temp dir throughout, so nothing reads the operator's own settings. The live class stays opt-in.
- **#949**: the cold `_learned_versions()` call in `test_a_warm_scan_re_reads_no_reg_whose_file_is_unchanged` now runs inside `_counting_reg_reads` and asserts the three reg names, so the warm scan's zero is proven against a counter that sees reads. A counter that never matches fails the new assertion (checked, then restored).
- **#957**: `pane-shim-stale.test.ts` gains a scenario that forces a reconnect from the visibility path (a quiet socket abandoned on foreground, no raise since it never armed), lets the redial arm as `foreground`, and ticks it silent: the watchdog's abandon raises `foreground-quiet`, the spelling `docs/read-side.md` and the shim comment name alongside `reconnect-quiet`. Changing the shim's `raiseStale(qw+"-quiet")` to `raiseStale(qw)` fails this test and the three `reconnect-quiet` pins (checked, then restored).
- **#947**: `github-link.test.ts` binds the viewer's poster once per file, so the reconnect re-ask is asserted as exactly one. Every `initFileView` call adds another `romp:wsup` listener on the shared window, which is why the count could only be asserted as "at least one" before. A duplicated `romp:wsup` listener inside `initFileView` now fails the test with 2 != 1 (checked, then restored). `tsc --noEmit` accepts the `typeof import("./file-view")` type.
- **#953**: the 10 s pre-flight test in `romp-cli-scope.bats` is replaced by a fake `timeout` on PATH that exits 124: same wrapper branch, same single stderr line, in milliseconds instead of a 20 s sleeper with a 9 to 16 s wall-clock window. The fake is the `timeout(1)` the wrapper finds, so the test also runs on a box without coreutils timeout (macOS), where the old one skipped. The file runs in 3.5 s instead of 13.5 s. What goes away: the only executable check that a real coreutils timeout cuts a hanging pre-flight tree. The argv pin that runs the real timeout on a fast pre-flight stays, and `_preflight_sleeps` stays for the no-timeout test. Changing the wrapper's `-eq 124` to `-eq 125` fails the new test at the "did not finish within 10 s" pin (checked, then restored).

## Not in this PR

The other follow-ups from the same reviews come separately: the `__main__` import in `test_tempdir_hygiene.py` (#944), the `_ALIAS_SERVED.clear()` setUps in `test_judge.py` (#948), and the dead-wait mid-pass comment (#943). Also out: the `_set_live` NamedTemporaryFile leftovers noted in #944, and any change to the twin's write path.

## Tests

Targeted, at head: `test_postal_relay_honesty.py` 30 passed in order, the single test and its class green alone, `-n 8` on the module 5 of 5 green (base: 4 of 5 red); `test_session_move_live.py` 6 passed, 1 skipped (was 2 and 1); `test_model_versions.py` 41 passed; `test_sdk_lifecycle_hardening.py` 39 passed; `test_state_isolation_order.py` 2 passed; `romp-cli-scope.bats` 19/19 (the replaced test runs, not skipped); `pane-shim-stale.test.ts` 14/14 (was 13); `github-link.test.ts` 12/12; `npm run typecheck` clean.

Full: `bats tests/*.bats` 429/429; `npm test` 2839/2839; `pytest -n 8` 7583 passed, 46 skipped, 8 failed in two untouched modules (`test_kernel_deferral_sweep.py`, `test_kernel_nudge_last_resort.py`) that pass alone (11 and 34 passed): xdist ordering, not this change. CI runs pytest serially, so it never saw the relay-honesty failure and will not see those. CI proves Linux only (the macOS jobs are dispatch-only); the bats fake is POSIX sh and the two webview tests are platform-independent node tests.

Tier: tests-only

### DIFF
--- a/tests/romp-cli-scope.bats
+++ b/tests/romp-cli-scope.bats
@@ -154,8 +154,9 @@ SH
 # The pre-flight is bounded (2026-09-05): `timeout 10 systemd-run … -- true` when coreutils' timeout is
 # on PATH. A user bus that accepts the connection and never answers would otherwise hold systemd-run
 # for its own 90 s default, past the SDK's 60 s initialize timeout, so the fallback never engaged in the
-# case it was added for. The fake systemd-run below sleeps on the pre-flight only; the wrapper must fall
-# back well inside that sleep, and the CLI's own launch must not run through timeout.
+# case it was added for. The fake systemd-run below sleeps on the pre-flight only, so the no-timeout test
+# can show an unbounded pre-flight that completes; the bound itself is proven with a fake timeout that
+# exits 124.
 _preflight_sleeps() {   # $1: seconds the fake systemd-run sleeps on a `-- true` pre-flight, then exits 0
     cat > "$BIN/systemd-run" <<SH
 #!/bin/sh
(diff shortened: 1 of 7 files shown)


ksarma:
Merging as shipped. Two wording nits below; neither blocks.

Reviewed inline: the flake diagnosis holds. At origin/main the disk-twin test fails alone and whenever PresenceBlinkHonesty runs before IdShapedAddressingRoutes. It fails because only the ack branch's `peer_seen_add` creates the postal STATE directory and `_local_presence` swallows the twin write. The setUp mkdir removes that dependency. The new twin-exists assertion catches a missing seed at the write. Each of the six follow-ups does what the source comment asked. Each new assertion fails under the mutation it guards. A never-matching reg-read counter fails the new cold-scan check; the old test passed vacuously. A duplicated `romp:wsup` listener fails the exact re-ask count; the old `>= 1` passed. `raiseStale` without the `-quiet` suffix fails the new foreground-quiet test. A hardcoded reconnect-quiet fails only that test, so the coverage is new. A wrapper checking for 125 fails the fake-timeout bats test at the pre-flight message. The precedence class opens no settings outside a private temp dir and spawns no helper. The live class stays opt-in. Every touched file is under `tests/` or is a webview `.test.ts`, so the label fits.

Follow-ups:
- `tests/test_session_move_live.py:178` (follow-up): the docstring says `HOME` and `CLAUDE_CONFIG_DIR` are patched throughout. The first two tests patch only `CLAUDE_CONFIG_DIR`. Isolation holds today because `~/.claude` is consulted only when `CLAUDE_CONFIG_DIR` is empty. Patch `HOME` in both or reword.
- `ui/webview/pane-shim-stale.test.ts:283` (follow-up): the comment cites 'the body', which has no referent inside the file.

Locally, the four Python files: 116 passed, 1 skipped (the opt-in live class). The disk-twin test is green alone, in adverse order, and in 16 of 16 xdist runs of the module. `romp-cli-scope.bats` passed 19/19 in about 4 s. npm typecheck clean, 2839/2839. CI: all six required checks green.

Review run by the maintainer's automated review session; verified against head 7ae11680.

PR #966, opened 2026-09-07
Codex judge child: strip every ANTHROPIC_* name, not only the API key

ksarma:
## What breaks

`_judge_run`'s codex branch launches `codex exec` with the judge environment minus `ANTHROPIC_API_KEY` (the #885 fold). Every other variable in the Anthropic namespace that the kernel's environment carries still reaches another vendor's process. Since #962 the billing names are no longer among them: a codex call resolves `auth = "codex"`, so `_judge_env` withholds `ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN` and `CLAUDE_CODE_OAUTH_TOKEN` and injects none back. What still rides:

- `ANTHROPIC_BASE_URL` and `ANTHROPIC_CUSTOM_HEADERS`: a proxy or gateway URL and the header that usually carries the gateway's own credential. The kernel reads `ANTHROPIC_BASE_URL` itself (`sdk_backend._fetch_key_fast_org`), so an installation that sets it is an ordinary one.
- `ANTHROPIC_MODEL`, `ANTHROPIC_SMALL_FAST_MODEL`, `ANTHROPIC_LOG` and any deployment-specific `ANTHROPIC_*` name.

## Reproduction

With the judge engine set to codex and `ANTHROPIC_BASE_URL=https://proxy.example.invalid` exported to the kernel, the codex child's environment contains that variable. The new test `test_the_real_judge_env_and_the_strip_compose` does this in-process: it runs the real `_judge_env` and fails on main with `ANTHROPIC_BASE_URL reached the codex child`.

## The fix

The filter in the codex branch becomes a prefix test:

```python
cenv = {k: v for k, v in env.items() if not k.startswith("ANTHROPIC_")}
```

A prefix rather than a longer list, because the namespace is the vendor's: a list has to be maintained as the CLI grows names, and a new one would land in the child silently. The prefix also drops harmless names (`ANTHROPIC_LOG` and other debug knobs) from the codex child, on purpose; that child has no use for them either. The claude-engine path is untouched: that child is the vendor's own CLI, and `_judge_env` still injects the key for a key-billed call.

Composition with #962: its codex bypass lives in `_judge_env` (the `auth = "codex"` short-circuit in `_judge_run` and the pops of the three billing names), and this change is on the branch's own `cenv` line, about 220 lines below. Neither replaces the other. `_judge_env` decides billing and withholds credentials for every engine; the codex branch removes whatever else the vendor's namespace carries. The unmocked test runs both.

## Tests

`tests/test_judge_engine.py`, `CodexChildEnvIsVendorScoped` (the home of the #885 pin):

- `test_the_codex_subprocess_env_lacks_every_anthropic_name`: `_judge_env` mocked to hand over `ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_BASE_URL` and `ANTHROPIC_CUSTOM_HEADERS`; asserts that none of them, and no other `ANTHROPIC_`-prefixed name, reaches the child, and that `ROMP_SUMMARIZING` still rides. Fails on main: `ANTHROPIC_AUTH_TOKEN reached the codex child`.
- `test_the_real_judge_env_and_the_strip_compose`: the same fixtures placed in `os.environ`, `_judge_env` real. Fails on main: `ANTHROPIC_BASE_URL reached the codex child` (the two billing names were already gone, which is #962 doing its part).

Assertions name the variable and never print the mapping; a failed `assertNotIn` on a dict prints the developer's whole environment, so the existing assertion is rewritten the same way. Fixture values are not credential-shaped. The module's `if __name__ == "__main__"` guard sat above this class and moves to the end, so a direct run collects it too.

`docs/codex.md` gains one sentence in "Running the judges on Codex".

Full `pytest` run locally: 7584 passed, 50 skipped, 0 failed. The change is a dict comprehension with no platform branch, and the tests replace `subprocess.run`, so the Linux CI cell covers every changed line. The widened filter has been in a Linux installation's kernel since 2026-09-05; that installation runs the Claude judge engine, so the codex path is exercised by the tests, not by live traffic.

## Design point

`CLAUDE_CODE_OAUTH_TOKEN` is Claude Code's login token and sits outside the prefix. `_judge_env` withholds it for a codex call since #962, so the codex child does not see it today; it is left out of this filter to keep the rule one vendor namespace. If you would rather the branch's own filter cover it too, it is one more exact name in the same comprehension.

Tier: fix

### DIFF
--- a/docs/codex.md
+++ b/docs/codex.md
@@ -48,7 +48,9 @@ one-shot `codex exec` call — ephemeral (no session files), using romp's custom
 `romp_judge` permission profile, and billing the machine's Codex login. The
 profile grants only Codex's minimal runtime files plus read access to the fresh
 empty scratch workspace supplied by `-C`; it denies writes, network access, and
-host-wide reads. Verified live: the real caption and gist
+host-wide reads. Every `ANTHROPIC_*` variable is removed from the child's
+environment, so a key, proxy URL or custom header configured for Claude never
+reaches Codex. Verified live: the real caption and gist
 judges answer correctly in ~5–6s per call.
 
 Honest caveats while this is new:
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -1528,9 +1528,14 @@ def _judge_run(model, sys_prompt, user, effort=None, judge=None, tier="triage", 
             outp = os.path.join(JUDGE_SCRATCH, "codex-%d-%d.out" % (os.getpid(), rid))
             try:
                 try:
-                    # another vendor's process has no use for the Anthropic key (_judge_env re-injects
-                    # it for key-billed sessions); strip it from the child's environment (PR #885 review)
-                    cenv = {k: v for k, v in env.items() if k != "ANTHROPIC_API_KEY"}
+                    # another vendor's process has no use for anything in the Anthropic namespace.
+                    # _judge_env already withholds the billing names for a codex call (auth "codex":
+                    # nothing is injected back), but the kernel's environment can carry more under the
+                    # same prefix — ANTHROPIC_BASE_URL and ANTHROPIC_CUSTOM_HEADERS for a proxy (the
+                    # header usually carries its credential), ANTHROPIC_MODEL, debug knobs — and a list
+                    # would need maintaining as the CLI grows names. Strip the prefix; the harmless
+                    # names go too, on purpose (PR #885 review, widened from the one key)
+                    cenv = {k: v for k, v in env.items() if not k.startswith("ANTHROPIC_")}
                     p = subprocess.run(_judge_cmd_codex(model, _codex_effort(effort, tier), outp),
                                        input=(sys_prompt or "") + "\n\n" + (user or ""),
                                        capture_output=True, text=True, cwd=JUDGE_SCRATCH, env=cenv,
--- a/tests/test_judge_engine.py
+++ b/tests/test_judge_engine.py
@@ -154,13 +154,17 @@ class JudgeEngine(unittest.TestCase):
         self.assertIn("--output-format", cmd)
 
 
-if __name__ == "__main__":
-    unittest.main(verbosity=2)
-
-
 class CodexChildEnvIsVendorScoped(unittest.TestCase):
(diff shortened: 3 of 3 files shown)


ksarma:

Reviewed inline: the codex branch of `_judge_run` now drops every ANTHROPIC_-prefixed name from the child environment (kernel/judge.py:1538), where it previously dropped only the API key. A proxy URL, custom header, model pin or debug setting configured for Claude therefore no longer reaches the codex exec process. The claude branch still passes env unchanged (kernel/judge.py:1606), and `tests/test_judge_auth_billing.py` pins that a key-billed claude child keeps its key. Composition with #962 holds: `_judge_run` resolves auth codex for the codex engine (line 1436). `_judge_env` pops ANTHROPIC_API_KEY, ANTHROPIC_AUTH_TOKEN and CLAUDE_CODE_OAUTH_TOKEN and injects a key only for auth key (lines 1279, 1305). Nothing is put back before the prefix strip runs. Nothing the codex child needs carries the prefix: `_judge_cmd_codex` reads no environment and passes `--ignore-user-config --strict-config`. `_judge_env`'s own additions (ROMP_SUMMARIZING, DISABLE_PROMPT_CACHING, MAX_THINKING_TOKENS) survive the strip, as the tests assert. On the design point, the branch filter stays one vendor namespace. `_judge_env` withholds CLAUDE_CODE_OAUTH_TOKEN for a codex call, and the #962 test pins it absent (tests/test_judge_auth_billing.py:262). No extra name is needed. Privacy and vocabulary scans of the added lines are clean, and the body matches the diff.

Red-first verified. With the PR's `tests/test_judge_engine.py` on origin/main (a3fe9a2f), two tests fail as the body states: `test_the_codex_subprocess_env_lacks_every_anthropic_name` and `test_the_real_judge_env_and_the_strip_compose`. ANTHROPIC_AUTH_TOKEN and ANTHROPIC_BASE_URL reached the codex child; 2 failed, 7 passed. At head the file passes 9 of 9. CI at f094e3fd: all six required checks pass.

Review run by the maintainer's automated review session; verified against head f094e3fd.

PR #967, opened 2026-09-07
Review follow-ups from #941, #942/#943/#957, #945, #947, #948 and #953: one fix per merge comment

ksarma:
Eight follow-ups the merge comments on #941, #942, #943, #945, #947, #948, #953 and #957 named as out of scope for those PRs, one commit each, each with a test that fails on the code before it. Nothing here changes a frame, a store or a wire format; the commits are independent and separable.

## 1. A TaskUpdate the CLI rejected moves no item in either to-do fold (from the #942 comment)

What breaks: `_fold_tasks` (kernel) and `declared_plan` (event_model) record the tool_use ids whose tool_result came back `is_error` and skip a TaskCreate in that set, but a TaskUpdate in the same set was still applied. A status value the CLI refused (an `InputValidationError` on `status`) folded as the item's status, and on the sync side a refused completing update flipped the mirrored agentTask node to done for a task the store still holds open.

The fix: both TaskUpdate branches skip a rejected update before the lookup, keyed on the result's `is_error` exactly as the TaskCreate branch is, so the two folds stay identical. An update whose result has not landed still applies.

Tests: `test_fold_ignores_a_rejected_taskupdate` (tests/test_kernel.py); `test_a_rejected_taskupdate_moves_no_step` and `test_a_rejected_taskupdate_leaves_the_mirrored_node_open` (tests/test_judge_authoritative_plan_sync.py). All three fail before the change: the refused status lands, and the sync returns True and flips the node. Fixtures are synthetic, with invented `InputValidationError` text.

## 2. The WS-thread act-now tick logs a failure instead of closing the socket (from the #943 comment)

What breaks: the setAutoNudge and setCompactSuggest arms run the auto-nudge tick on the WS handler thread with no wrap. The reader loop re-raises OSError as a socket failure and its outer handler tears the connection down silently, so an OSError out of the pass head (the session-listing fork, a store read) closed the dashboard's socket with no log line: since #846 for setCompactSuggest, and for setAutoNudge once #943 restored its tick.

The fix: one module-level wrap, `_ws_act_now_tick()`, beside `_auto_nudge_tick`, holds the `run_dead_wait=False` call in a try/except that logs the traceback (`auto-nudge (ws act-now): ...`) the way the pusher's own wrap does; both arms call it. Catching Exception there is safe because the tick never writes to the delivering socket (its only push is `_push_soon()`, a wake flag), so nothing caught is that socket's own failure; the helper's docstring says so. `_tell_stale_gesture` stays outside the wrap. The single-flight lock and the one-observer sweep skip are unchanged (the same call text, inside the helper).

Tests: `WsActNowTickIsWrapped` (tests/test_setting_gesture_order.py) drives the real dispatch for both arms with a tick that raises OSError, then RuntimeError; before the change the exception escapes `_dispatch_ws`. After: the client stays alive, the flag write stands, stderr names auto-nudge and the traceback. The dead-wait call-site pin (tests/test_dead_wait_block.py) is re-pointed: both arm slices call the helper, `_dispatch_ws` has no direct `_auto_nudge_tick(` call, and the helper's source carries the exact sweep-skipping call. The mid-pass test's comment now names the single-flight lock as what holds its transition (the nested same-thread tick returns at the try-acquire before `run_dead_wait` is read), not the sweep skip. If you prefer two inline wraps over the helper, that pin needs no change; say so.

## 3. The bell names panes from `_PANE_ORDER` (from the #957 comment)

What breaks: `_LANDING_ERRS_JS` kept its own four-entry `PN` map beside `_PANE_ORDER`, a second hand-written list.

The fix: `PN` is `json.dumps(dict(_PANE_ORDER))` spliced into the script constant at import, so the rail, the mobile tabs, the drop row (#957) and the connection-lost row name panes from one list. `_PANE_ORDER` and its comment move above `_LANDING_ERRS_JS` because the string is built from it at import; no module-level reference precedes the new position, and `_rail_buttons_html` / `_mtab_buttons_html` stay where they are. The constant remains one complete string, so `test_error_center`, which runs it in node, keeps proving the script executes (its chat-drop entry still reads "Chat").

Tests (tests/test_pane_order_parity.py): `test_labels_match_the_pn_map` now parses the literal as JSON (the old parser read bare keys and single-quoted labels, so against JSON it read nothing); `test_the_pn_map_is_built_from_the_constant` pins the derivation and the definition order. Both fail before the change.

## 4. The boot orphan reaper treats this kernel's own children as live by pid (from the #941 comment)

What breaks: since #941 `find_orphan_clis` decides "live" from the parent's `ps` text alone, so the calling kernel's own children are shielded only when its launch spelling is one `_is_kernel_cmd` knows (`.../bin/romp-kernel`, `kernel/kernel.py`). The old `ppid != 1` test shielded them unconditionally. The exposure is not only a second kernel: the reconcile runs on a thread after the backend is constructed, concurrent with the kernel serving requests, so a session started before the thread's `ps` is this kernel's child carrying one of the lastsids, and a kernel run as `python3 ./kernel.py`, through a `-c` runner or from a renamed launcher would SIGTERM its own live session at boot.

Reproduction (synthetic, the pure function before the change): a listing with the user manager at 901, the kernel at 31337 spelled `python3 ./kernel.py`, its child 707 and an orphan 708 under 901 returns [707, 708]; a listing holding only the child, with the kernel's own line absent, returns the child.

The fix: `find_orphan_clis` takes the calling kernel's pid as a required third positional (`own_pid`, the shape `find_session_cli` already has with `parent_pid`) and treats `ppid == own_pid` as live before the parent lookup, so a listing missing the kernel's own line still protects its children; `_boot_reconcile` passes `os.getpid()`. A CLI parented to any other kernel is judged by that kernel's text as before; the caller's existing `pid == os.getpid()` skip stays. An `own_pid=None` default is a two-line change if you prefer the ten existing calls untouched.

Tests (tests/test_sdk_lifecycle_hardening.py): the ten FindOrphanClis calls gain the pid; three new pure tests (own children live under three spellings `_is_kernel_cmd` does not know, asserted False for each so the test rests on the pid path; own children live when our line is absent; another kernel's child under an unknown spelling is still an orphan); one BootReconcile test runs the reconcile with `ps` and `os.kill` patched and the kernel spelled `python3 ./kernel.py` under this process's pid, and asserts only the orphan under the user manager is signalled. Before the change: the amended and new pure tests fail on the signature (TypeError) and the BootReconcile test on the property (the child is killed too).

## 5. A judge envelope with no `modelUsage` map is announced once per process (from the #948 comment)

What breaks: since #948 `_note_served_model` checks the bare-alias table against the model id the CLI's result envelope reports under `modelUsage`. When the envelope carries no such map (a CLI older than the field, or a shape change) the check returned without a word, so the index tier kept choosing its lever from a table nothing had verified, with nothing in the log.

Reproduction: stub `subprocess.run` to return `{"result": "a caption", "usage": {}}` and call `_judge_run("haiku", ..., tier="index")`: stderr carries only the lever line and `_ALIAS_SERVED` stays empty.

The fix: at the existing `if not isinstance(mu, dict): return` site, after the bare-alias gate (a versioned pin is never checked, so it stays silent), one stderr line per process names the alias whose call revealed it, says the served version cannot be checked, and that the index tier's lever answers from the alias table until a CLI that reports the field. The latch is a module set beside `_UNKNOWN_MODEL_LOGGED` and `_ALIAS_DRIFT_LOGGED`, check-then-add without a lock like them. Design points: once per process, not per alias (the field's presence is a property of the CLI); the same line fires for a malformed non-dict value (same site; the malformed-entry test still proves the reply survives); a present map that names no same-family model stays quiet as before, and the tests pin the two cases apart; the text avoids the `_ALIAS_HEAD` token so the test class's drift-line filter keeps meaning drift lines. Default picks are bare aliases, so on such a CLI a default install logs one line at its first successful judge call; on a CLI that emits the field nothing changes. Per-alias instead of per-process, or excluding the malformed case, is a one-line change either way.

Tests (tests/test_judge.py, AliasHeadDrift): `test_a_missing_model_usage_map_is_announced_once_per_process` (one line on the first bare-alias call, none on a second alias, none for a versioned pin, one again after the latch is cleared), and the present-map case pinned quiet separately. Before the change, with the tests in place: every AliasHeadDrift test errors on the missing latch; with the latch alone, the new test fails on the missing line and the other eight pass.

A second, tests-only commit is the comment's other follow-up: `IndexTierLever.setUp` and a new `EffortCapability.setUp` clear `_ALIAS_SERVED` (and the new latch), so their `haiku -> False` pins no longer depend on `AliasHeadDrift`'s tearDown having run in the same xdist worker. Shown load-bearing by a scratch run with that tearDown reset disabled and the moved-alias test run first: four assertions across the two classes fail without the setUps and pass with them. The `IndexTierLever` stub is unchanged: its empty-stderr pin is on a second call within one test, which the per-process latch keeps quiet in any order (the #948 body assumed the stub would need `modelUsage`; it does not).

## 6. The GitHub button's origin check never pops an askpass prompt (from the #947 comment)

What breaks: the origin check's `ls-remote` ran with `GIT_TERMINAL_PROMPT=0` alone, so the child inherited `GIT_ASKPASS`, `core.askPass` and `SSH_ASKPASS`. For an https origin that answers 401, git asks for a username and a password through the first of those it finds; with a GUI askpass exported in the kernel's shell a prompt appeared on every viewer open of a file under such an origin, then died at the 3 s deadline (reproduced against a loopback 401 stand-in). ssh's own `SSH_ASKPASS` use (DISPLAY set, no tty) was open as well, and `_git_net_out`'s docstring claimed the sessionless ssh would fail rather than wait, which did not hold on this path.

The fix: `GIT_ASKPASS=""` and `SSH_ASKPASS_REQUIRE="never"` join `GIT_TERMINAL_PROMPT="0"` in the same env dict. The empty value is load-bearing: git tests the variable's presence before falling to `core.askPass` and then `SSH_ASKPASS`, so set-empty overrides both while unset would not (on git 2.43 each of the three sources ran the askpass twice under `GIT_TERMINAL_PROMPT=0` alone; zero runs with `GIT_ASKPASS=""`, failure in about 10 ms). `SSH_ASKPASS_REQUIRE=never` is honoured by OpenSSH 8.4 and later and ignored by older ssh. The chosen behaviour, now in the docstrings: no prompt of any kind; an origin that wants a credential reads "could not check whether branch X is on origin", including where an editor's askpass would have answered silently for a signed-in user. A configured `credential.helper` is not disabled and may still answer silently; out of scope here.

Tests (tests/test_file_github.py, `NoPromptOnOrigin`, both red before the change): the stand-in ssh dumps the transport child's environment and the test asserts set-empty `GIT_ASKPASS` (pinned as `${GIT_ASKPASS+set}:` so a later pop() reads as the regression it is), `SSH_ASKPASS_REQUIRE=never` and `GIT_TERMINAL_PROMPT=0`; a loopback 401 server plus a marker-writing askpass shows the kernel's query runs no askpass and returns unchecked, after a control run proves the stand-in does elicit the askpass under `GIT_TERMINAL_PROMPT=0` alone (skips on a git without the https helper). OS-neutral: POSIX sh, `http.server` on 127.0.0.1:0, env vars only.

## 7. romp-manager: after a failed scoped start, the log reads where the tmux server sits (from the #953 comment)

What breaks: `startTmuxServer`'s fallback logged "tmux server ensured without a scope — under the service, a service restart will take it down" whenever the bare `tmux start-server` succeeded after a scoped failure. That predicts a loss in two cases where none happens. A scoped call that times out: `execFileSync` SIGTERMs the tmux client (systemd-run exec'd it in place), the server daemon that client forked stays alive inside the romp-tmux scope, and the bare retry no-ops onto it (the shape the #953 review measured). And after a service restart the previous manager's server is still up in its old scope, so if systemd-run fails on this start (bus not up, a refused unit) the bare call again no-ops onto a scoped server.

The fix: after the bare call succeeds the manager reads the server's pid (`tmux display-message -p '#{pid}'`, which does not start a server and exits 1 with "no server running" when there is none; tmux 3.4) and its `/proc/<pid>/cgroup`, and words the line from that: a `romp-tmux-*.scope` on the path names the unit and says a restart leaves it alive; anything else is the previous line verbatim; an unreadable answer (probe failure, non-numeric pid, missing cgroup file) says it could not be read and predicts no loss. cgroup v1 (one line per controller) and v2 are matched per line. The probe is bounded by the start's timeout, runs only on the fallback path, never throws, and runs with the same scrubbed env as the two starts (so `test_op_credential_boundary`'s exec-count pin goes from two to three, with the reason). Wording the line as uncertain only on ETIMEDOUT would miss the second case; `systemctl is-active` on the unit probes the scope, not the server. Unchanged: the first failure line, the scoped-success line, and the darwin path (the edited branch is unreachable when `tmuxStartArgv` returns `scoped:false`).

Tests: tests/manager-tmux-scope.test.js covers `placementFromCgroup` (v1 and v2 samples, a romp-session scope not counted, an empty file), `tmuxServerPlacement` (the pid is what the cgroup read receives; every probe failure is unknown with its first line, bounded) and `bareEnsuredLine` (the unscoped string byte-identical to the old one); nine tests that fail before the change on the missing exports. tests/romp-manager-tmux-scope.bats gains the probe in the fake tmux (pid 1, whose cgroup is never a romp-tmux scope; `FAKE_TMUX_PID=none` for no server) and a case for the unreadable outcome; case 2 now pins the bare call followed by the probe. Both fail before the change.

## 8. Stale-setting toast: fade liveness, the refused value, no recency claim, the button's own title (from the #945 comment)

Two commits, gear.js and its two test files only; no kernel, frame or CSS change.

What breaks: (a) a stale flush refused by N kernels folds into one toast keyed by (setting, gt), and liveness read only the node's parentNode; the toast fades at 11 s and is removed at 12 s, so a refusal arriving in that second was written into a toast at opacity 0 and never shown. (b) Neither the copy nor the Apply anyway label showed the refused value, only the kept one; in the frozen-tab case the pick is hours old and one click sends it to every linked kernel, so the user could not see what they were about to apply. (c) The copy called the kept pick "a later pick ... already in place", a recency claim the kernel cannot make under the clock skew #945 fixed: it holds only a larger stamp. (d) The Apply anyway button set no title, so it inherited the toast's "click to dismiss" tooltip.

The fix: (a) `staleLive(t)` reads the fade class beside parentNode at lookup, so a dismissed, expired or fading toast is not live and the frame gets a fresh toast; still event-keyed, no cleanup rides the timers (the fold block holds no setTimeout, which the source test pins). (b, c) The frame's gesture echo already carries the value: every emitter posts `{type, <one value field>, gt}` and the kernel echoes it without gt, so `staleRefused` reads the echo's one key beside `type`, worded through `staleWord` (booleans as on/off), and only for the setting the frame names (the whitelist Apply anyway uses; an echo for another setting is not this setting's value). An echo of any other shape, or none from an older kernel, reads as no value and the copy and label take their value-less form. The copy becomes "Triage model: opus was not applied on web, api, this machine. Keeping fable." (value-less: "Triage model: not applied on web. Keeping fable."), with no ordering word; the label becomes "Apply opus anyway". A folded toast keeps the value its first frame carried. A boolean toggle reads "Auto Nudge: on was not applied on web. Keeping off." under the same uniform wording; a "Turn on anyway" special case is a reasonable alternative if you prefer it. (d) The button gets its own title: "apply opus on every linked kernel, replacing the value they kept". An older kernel's frame (no gesture, no gt) behaves as before.

Tests: setting-stale-fold.test.ts fires the 11000 ms timer between two frames of one gesture and expects two toasts (one before the change); its exact strings are rewritten to the new copy, and it gains the label, the button title, a mismatched echo (no value shown) and a two-field echo (value-less copy, plain label); setting-stale.test.ts pins the new lookup, the helpers, the title line and the retired claims (its fold-region sanity bound widens from 3000 to 6000 characters, since the helpers grew the region). Before the changes: 2 tests fail for (a), 11 for (b) to (d).

## Test runs

Per commit, red first as stated above, then green. Whole suites at the head, on Linux: `pytest -n 8` over tests/: 7599 passed, 46 skipped, and one failure, `test_op_credential_boundary`'s exec-count pin, raised to three with the reason and folded into commit 7 (that module re-run alone: 5 passed; no other file changed after the sweep). `bats tests/*.bats`: 430 ok. In vscode-extension: `npm test` 2840 pass, `npm run typecheck` and `npm run build` clean. None of the surfaces is platform-specific: the Python and node changes have no platform branch, the bats scoped-path cases are Linux-only by construction as before, and the manager edit sits in a branch unreachable on darwin.

Tier: fix

### DIFF
--- a/bin/romp-manager
+++ b/bin/romp-manager
@@ -586,8 +586,10 @@ function withoutOpCredentials(env) {
 // starts the daemon outside the service cgroup; a session's later `tmux new-session` joins that server
 // and survives a service restart. The tmux CLIENT still exits at once (it only starts the daemon), so
 // execFileSync's semantics are unchanged, and when a server already runs the call is the same no-op and
-// the scope collects itself. Any failure of the scoped path falls back to the bare call. The macOS path
-// is as it was: launchd owns the lineage there and the TCC rationale above still applies.
+// the scope collects itself. Any failure of the scoped path falls back to the bare call, and the line after
+// it READS where the server actually sits (its cgroup) rather than assuming the bare call placed it: a
+// timed-out scoped call kills only tmux's client, the daemon it forked stays inside the scope. The macOS
+// path is as it was: launchd owns the lineage there and the TCC rationale above still applies.
 const TMUX_START_ARGV = ['tmux', 'start-server', ';', 'set', '-g', 'exit-empty', 'off'];
 
 // Is `name` an executable on the env's PATH? (Node has no `which`.)
@@ -641,6 +643,66 @@ function scopedStartFailure(e) {
   return { tool, detail };
 }
 
+// After a failed scoped start the bare call runs, and where the server then sits is READ, not predicted
+// (the #953 review): the old fixed line said the server was unscoped and a service restart would take it
+// down, which is wrong in two cases. A scoped call that times out — execFileSync SIGTERMs the tmux CLIENT
+// (systemd-run exec'd it in place) while the server daemon that client forked stays alive inside the
+// romp-tmux scope, so the bare retry is a no-op onto a scoped server. And after a service restart the
+// previous manager's server is still up in its old scope, so if systemd-run fails on THIS start (bus not
+// up yet, a refused unit) the bare call no-ops onto a scoped server again. `#{pid}` is the server's pid;
+// display-message does not start a server (exit 1, "no server running on …", when there is none — tmux 3.4).
+const TMUX_SERVER_PID_ARGV = ['tmux', 'display-message', '-p', '#{pid}'];
+
+// PURE: where a /proc/<pid>/cgroup text places the server. cgroup v2 is one `0::/…/<unit>` line; v1 is one
+// line per controller, each ending in the unit — so every line is tried. Only the manager's own romp-tmux-*
+// scopes count: a romp-session-* scope is a session CLI's, not this feature's claim.
+function placementFromCgroup(text) {
+  const lines = String(text || '').split('\n');
+  for (const line of lines) {
+    const m = /\/(romp-tmux-\d+\.scope)\s*$/.exec(line);
+    if (m) return { placement: 'scoped', unit: m[1] };
+  }
+  if (lines.some((l) => l.trim())) return { placement: 'unscoped' };
+  return { placement: 'unknown', why: 'empty cgroup file' };
+}
+
+function defaultServerPid() {
+  return execFileSync(TMUX_SERVER_PID_ARGV[0], TMUX_SERVER_PID_ARGV.slice(1),
+                      { encoding: 'utf8', stdio: ['ignore', 'pipe', 'pipe'], timeout: TMUX_START_TIMEOUT_MS,
+                        env: withoutOpCredentials(process.env) });
+}
+function defaultReadCgroup(pid) { return fs.readFileSync('/proc/' + pid + '/cgroup', 'utf8'); }
+
+// Never throws — a throw here would abort startManager before the control port binds — so every failure
+// of the probe is an 'unknown' placement carrying the failure's first line, bounded like the failure
+// line's detail. `serverPid` and `readCgroup` are injectable for the node suite; the defaults ask tmux
+// (bounded by the same timeout as the start) and read /proc.
+function tmuxServerPlacement({ serverPid = defaultServerPid, readCgroup = defaultReadCgroup } = {}) {
+  const why = (e) => {
+    const first = String((e && e.stderr) || '').split('\n').map((s) => s.trim()).find(Boolean)
+      || String(e && e.message || e).split('\n')[0];
+    return first.length > FAILURE_LINE_MAX ? first.slice(0, FAILURE_LINE_MAX - 1) + '…' : first;
+  };
+  let pid;
+  try { pid = String(serverPid()).trim(); } catch (e) { return { placement: 'unknown', why: why(e) }; }
+  if (!/^\d+$/.test(pid)) return { placement: 'unknown', why: 'tmux reported no server pid' };
+  let text;
+  try { text = readCgroup(pid); } catch (e) { return { placement: 'unknown', why: why(e) }; }
+  return placementFromCgroup(text);
+}
+
+// PURE: the line the bare call's success logs, worded from where the server sits. The unscoped wording is
+// the line as it always was (its pins stay); the other two predict no loss.
+function bareEnsuredLine({ placement, unit, why }) {
+  if (placement === 'scoped') {
+    return `tmux server ensured — already running inside ${unit} (the scoped start had begun it, or an earlier one still holds it); a service restart leaves it and its sessions' jobs alive`;
+  }
+  if (placement === 'unscoped') {
+    return 'tmux server ensured without a scope — under the service, a service restart will take it down';
+  }
+  return `tmux server ensured — whether it sits in a scope could not be read (${why}); under the service, a restart takes it down only if it is not in one`;
+}
+
 function startTmuxServer() {
   const pick = tmuxStartArgv();
   if (!pick) return;   // no tmux on PATH: nothing to start, and nothing to log (as before the scopes)
@@ -656,7 +718,8 @@ function startTmuxServer() {
     return;
   } catch (e) {
     if (!pick.scoped) return;   // a server is already up, or tmux itself failed — nothing more to do
-    // Say what failed and where the server will land: this may be a terminal run with no service cgroup.
+    // Say what failed; where the server sits is read from its cgroup after the bare call (bareEnsuredLine),
+    // not predicted here.
     const { tool, detail } = scopedStartFailure(e);
     log(tool === 'tmux'
       ? `tmux failed inside the scope systemd-run started for it (${detail}) — starting it the plain way instead`
@@ -664,7 +727,7 @@ function startTmuxServer() {
   }
   try {
     execFileSync(TMUX_START_ARGV[0], TMUX_START_ARGV.slice(1), { stdio: 'ignore', timeout: TMUX_START_TIMEOUT_MS, env: withoutOpCredentials(process.env) });
-    log('tmux server ensured without a scope — under the service, a service restart will take it down');
+    log(bareEnsuredLine(tmuxServerPlacement()));
   } catch (e) { /* a server is already up, or tmux itself failed — nothing to do */ }
 }
 
@@ -777,6 +840,7 @@ function ensureUp() {
 module.exports = { restartGate, quietGate, quietTick, loadSpecs, specEnv, fileStamp,
                    kernelToken, fetchBusy, resetDrainNotices, queueQuietRestart, clearPendingQuiet,
                    tmuxStartArgv, scopedStartFailure, commandOnPath, TMUX_START_ARGV,
+                   placementFromCgroup, tmuxServerPlacement, bareEnsuredLine, TMUX_SERVER_PID_ARGV,
                    withoutOpCredentials, serviceEnvHasRef };
 if (require.main === module) {
   const cmd = process.argv[2];
--- a/kernel/event_model.py
+++ b/kernel/event_model.py
@@ -2769,7 +2769,8 @@ def declared_plan(session):
     live task store, so downstream (the judge's plan-sync) sees a generic 'declared plan' shape
     instead of raw tool calls. Mirrors the kernel's _fold_tasks, with the same blind spots: only the
     MAIN agent's TaskCreate/TaskUpdate calls, and only those on the transcript's live chain — and the
-    same skip: a TaskCreate the CLI rejected (its paired tool_result carries is_error) is not a step.
+    same skips: a TaskCreate the CLI rejected (its paired tool_result carries is_error) is not a step,
+    and a rejected TaskUpdate moves none.
     `key` is the stable `Task #N` id lifted from TaskCreate's
     result text (a creation-order `cN` fallback if the result is unreadable); `status` rides each
     TaskUpdate. Only TaskCreate/TaskUpdate are folded — plain TodoWrite (no durable ids) is not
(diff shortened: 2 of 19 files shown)


ksarma:

Reviewed inline: Each of the ten commits touches only the files its message names. The `tests/test_op_credential_boundary.py` edit is the exec-count pin going from 2 to 3. That pin covers the new scrubbed tmux pid probe in b566c0e2, not the askpass change. The rejected TaskUpdate set is built in a first pass over all turns, so both `_fold_tasks` and `declared_plan` skip it soundly. The act-now tick now runs through `_ws_act_now_tick`, which catches and logs. That matters because `Handler._ws` re-raises an OSError out of `_dispatch_ws` as a socket failure. The bell's PN map is `json.dumps(dict(_PANE_ORDER))`. The constant moved above `_LANDING_ERRS_JS`, which is built from it at import. `find_orphan_clis` takes `own_pid` and treats `ppid == own_pid` as live before the command-text check. The only caller passes `os.getpid()`. The ls-remote env sets `GIT_ASKPASS` empty and `SSH_ASKPASS_REQUIRE=never`. Set-empty is the right override because git tests `askpass && *askpass` before falling to `core.askPass`. The judge's no-modelUsage line latches once per process. With the latch disabled, only the two tests that assert quiet-on-second-call fail, so no other stderr pin is order-dependent. On tmux 3.4, `display-message -p '#{pid}'` prints the server pid on a zero-session server and exits 1 with no server, matching the manager's probe. The stale toast reads the refused value as the echo's single non-type key. That holds because every emitter posts `{type, one field, gt}` and `_tell_stale_gesture` echoes msg minus gt.

- `bin/romp-manager:700`: `placementFromCgroup` returns unscoped for any cgroup not ending in `romp-tmux-N.scope`. A default-socket server started from the user's own terminal (login-session scope) therefore reads as unscoped. The line then claims a service restart will take it down, which it will not. Comparing the server's cgroup with `/proc/self/cgroup` would read the answer instead of predicting it.
- `ui/webview/gear.js:848`: `staleWord('')` returns `''` for the effort selects' Default option. A refused Default pick therefore renders the value-less copy and a plain Apply anyway. That is the frozen-tab case this commit set out to show. Mapping `''` to Default for the effort settings would close it.

Red-first checked per commit in a scratch worktree (each commit's tests at its parent, then at the commit): 66cb03cd 3 failed then 493 passed; 19d0e7cb 3 then 90; 6778ca2d 2 then 5; d51c3452 14 then 43; 412de8db 9 then 515; 826b33db 2 then 40; b566c0e2 1 pytest plus 9 node plus 2 bats then all green; d9a51ebe and 74db19a3 2 and 11 npm failures then 2839 and 2840 pass. At head: pytest over the nine changed Python files 1191 passed, npm test 2840 pass, node manager-tmux-scope 28 pass, bats 5 ok. CI at 74db19a3: all six required checks pass.

Review run by the maintainer's automated review session; verified against head 74db19a3.

PR #968, opened 2026-09-07
pre-push hook: scan each ref's tip and each new commit's added lines, not every inherited tree

ksarma:
## What breaks

`.githooks/pre-push` greps the full tree of every commit in a ref's update range (`remote_sha..local_sha`; for a new ref, everything no remote-tracking ref has). That is the wrong unit once a denylisted string has landed on `main` and been redacted forward: every branch cut from `main` in between inherits the string in the trees of its own commits although none of its diffs adds it, and a branch that merged `main` also carries `main`'s own commits in its range. The hook then refuses those pushes, naming the branch's own commit over a file the branch never touched, and the only way through is `--no-verify` on every such branch — the habit the hook exists to prevent.

## Reproduction

With a synthetic denylist (`zzsynthuser`) and a bare remote:

1. On `main`, commit `leak.txt` containing `/home/zzsynthuser/code`; push (the string escaped before the denylist knew it).
2. Cut `feature` from that commit and commit `web.txt` (clean).
3. On `main`, remove `leak.txt`; push.
4. On `feature`, merge `main`. Its tip is clean; its earlier commit's tree still holds `leak.txt`.
5. `git push -u origin feature` — refused: `commit <branch-work sha> would publish a personal identifier in: leak.txt`.

`tests/pre-push-hook.bats` (new) scripts this against the hook; the case "a branch that only INHERITED main's leak in its tree passes once its tip merged the redaction" fails on the current hook with status 1.

## The fix

Two rules replace the per-commit tree scan, both about what THIS push changes on the remote:

1. **The tip tree of each pushed ref must be clean.** That is the exposure a push creates; a string anywhere in it refuses the push, named as the tip (`the tip of refs/heads/<branch> (<sha>) would publish a personal identifier in: <path>`).
2. **Each commit new to the remote must add no banned line** in its diff against its first parent. A string introduced in an intermediate commit and removed by a later one is still caught, and the commit named (`commit <sha> ADDS a personal identifier in: <path>`) — the property #23 established. A commit whose tree only inherits an older string is not refused.

"New to the remote" is `<tip> --not <remote_sha> --remotes=<remote>`: anything reachable from the pushed-to remote's refs as last fetched is excluded, scoped to that remote — a commit only another remote has is still scanned, where the old new-ref form (`--not --remotes`) excused it. With no remote-tracking refs for the remote (never fetched, or a push by URL, where git passes the URL as `$1`) the exclusion matches nothing and the range is what the remote ref lacks, as before.

The added-lines pass is `git diff-tree -p -r -M` against the first parent (`--root` for a root commit) through a short awk that takes the path from the `+++ b/<path>` header only between a `diff --git` line and the first `@@` hunk, so an added content line that itself begins with `++ b/` (rendered `+++ b/...` in the patch) is content, attributed to the file it is in. `-M` keeps a rename from re-adding every line of the moved file. A merge is diffed against its first parent, so a string that exists in neither parent (a conflict resolution) counts as the merge's own addition.

A string already on the remote is fixed forward on `main` (redact, push `main`), not by refusing every branch cut since; the refusal footer now says so: `If the tip only INHERITS a string an older commit added and main has since redacted, merge main.`

One semantics change to accept explicitly: a branch whose intermediate commits inherit a string already on the remote is no longer refused — only its tip and its own added lines are checked. The tip rule still refuses re-exposing that string in any pushed tip.

## Tests

`tests/pre-push-hook.bats` (new, 11 cases) drives the hook by hand — ref lines on stdin, remote name as `$1` — with a synthetic denylist and a bare remote, so it can model remote-tracking refs the way a clone has them:

- basic: a clean commit passes; a string in a regular file is refused; a string in an intermediate commit is refused naming the commit; no denylist file means no scan;
- a branch that inherited `main`'s string passes once its tip merged the redaction; the same branch is refused while its tip still carries it, named as the tip, with the `merge main` remedy and no `ADDS`;
- a string added by one commit and removed by the next is refused naming the commit; a fresh string after merging `main` is refused naming `api.txt` without re-flagging `leak.txt`;
- a commit only another remote has is still scanned (per-remote scoping); with a real `remote_sha` and no remote-tracking refs, the fallback range still catches an intermediate addition;
- an added line shaped like a diff header (`++ b/decoy.txt`) is content: the hit is attributed to the real file, not to `decoy.txt`.

Eight of the eleven fail against the current hook (five on behavior — the inherited-leak pass, tip naming, per-remote scoping, and two whose refusal changes — and three more that pin the `ADDS` wording). The header-shaped case fails against a header parser that reads `+++ ` anywhere in the patch.

The eight existing `tests/install-sh.bats` hook tests (through a real `git push`) are unchanged and stay green: the intermediate-commit case is caught by the added-lines pass, and the only-new-commits case passes because the tip is clean and the range excludes what `origin/main` holds. `CLAUDE.md`'s backstop sentence and the `install-sh.bats` header comment describe the new rule.

Verified: `bats tests/pre-push-hook.bats` 11/11; `bats tests/install-sh.bats` 29/29; `bats tests/*.bats` 440/440; `shellcheck .githooks/pre-push` clean at warning level (the pre-existing SC2001 style notes on `sed` remain); full pytest unaffected (no Python changed). The reproduction above was also run through a real `git push` with the old and the new hook: refused before, accepted after; a new banned line on the branch is then refused naming that commit; a push by URL still catches it through the fallback range.

## Portability and limits

bash 3.2-safe (arrays are expanded only after the non-empty check, so `set -u` holds; `${var:0:10}`, `local`, here-strings) and POSIX awk/grep/sed/cut/sort. The test runs the hook through a `cd` helper rather than GNU `env -C`, so the suite stays GNU/BSD-portable per `tests/README.md`. Neither pass scans binaries (`git grep -I`; `diff-tree` prints `Binary files differ`), as before. `git grep` reads regular-file blobs only, so a symlink whose target holds a string is caught by the added-lines pass when the link is new to the remote, not by the tip scan when it is older. The exclusion is only as fresh as `refs/remotes/<remote>/*`: a clone that never fetched excludes nothing, which is stricter, never looser.

Tier: fix

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,19 +1,32 @@
 #!/usr/bin/env bash
 # romp pre-push hook: refuse to publish a commit that carries a personal identifier.
 #
-# The repo may go public and history is permanent, so this scans the COMMITS
-# BEING PUSHED (every new commit's full tree, not the working tree) for the
-# strings in ~/.config/romp/private-strings.txt — one banned string per line,
-# `#` comments and blank lines ignored, matched case-insensitively. With no such
-# file it exits 0 and stays out of your way, so on a clone that never set one up
-# (any contributor's machine) it is a no-op.
+# The repo may go public and history is permanent, so this reads the COMMITS
+# BEING PUSHED (not the working tree) for the strings in
+# ~/.config/romp/private-strings.txt — one banned string per line, `#` comments
+# and blank lines ignored, matched case-insensitively. With no such file it
+# exits 0 and stays out of your way, so on a clone that never set one up (any
+# contributor's machine) it is a no-op.
 #
-# Why pushed trees rather than a working-tree scan: the hook is symlinked into
+# Two checks per pushed ref, both about what THIS push changes on the remote:
+#
+#   the tip's tree    the content the push exposes. A banned string anywhere in
+#                     it refuses the push.
+#   each new commit   the lines it ADDS against its first parent. A string
+#                     introduced in an intermediate commit and removed by a
+#                     later one is still caught, and the commit named; a commit
+#                     whose tree merely INHERITS a string an older commit put
+#                     there is not. Every commit on every branch cut from main
+#                     between a string landing there and its redaction inherits
+#                     it that way, and a per-commit TREE scan refuses them all,
+#                     over files the branch never touched, when the only
+#                     exposure any of them could add is its tip.
+#
+# Why pushed commits rather than a working-tree scan: the hook is symlinked into
 # the SHARED git hooks dir (by install.sh), so it fires from every worktree —
 # but a working-tree scan is only as good as the tree you happen to push from,
 # and an untracked scanner script exists in only one worktree. Reading the
-# denylist from $HOME and grepping the pushed shas arms every worktree equally,
-# and catches a string buried in an intermediate commit whose tip is clean.
+# denylist from $HOME and reading the pushed shas arms every worktree equally.
 #
 # Bypass a false positive with `git push --no-verify`.
 set -euo pipefail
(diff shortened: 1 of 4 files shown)


ksarma:
Merging as shipped. Two follow-ups to file; neither blocks.

Reviewed inline: the tip-tree `git grep` still refuses a string anywhere in a pushed tip and names it as the tip. The added-lines pass over each new commit's first-parent diff catches a string one commit adds and the next removes, and names the commit. The awk reads the path only from the header between a `diff --git` line and the first hunk, so header-shaped content is attributed to the real file. A string already in the base is not re-flagged once the tip is clean, and the body states that change plainly. A deleted ref, empty stdin, a new branch with no remote-tracking refs, and a push by URL all exit without error; the fallback range still scans. No bash 4 constructs were added, and the userland calls are POSIX. The test uses a cd helper rather than `env -C`. The `CLAUDE.md` and `install-sh.bats` prose match the hook for a single-remote clone and quote no one. The added lines carry no private strings, hostnames, or real home paths.

- `.githooks/pre-push:66` (follow-up): the exclusion covers only the pushed-to remote's refs. In this repo's push-to-fork flow, a clean branch pushed to the fork as a new ref is refused. The refusal names an upstream main commit no fork ref reaches; the hook at main passed that push. Syncing fork main is refused the same way. The footer at line 121 and `CLAUDE.md` line 66 say merge main, which does not help there. Also exclude the upstream remote's refs (or all remotes), add a two-remote test, and qualify both sentences.
- `.githooks/pre-push:80` (follow-up): a merge is diffed against its first parent only, so a merge of main taken while the leak was live is refused as its own addition. The content is in the second parent and the tip is clean. The old hook refused it too, but the body says only conflict resolutions count for merges. Diff merges with `-c` or `--cc`, counting lines present in no parent, and add a merge-while-live case.
- `tests/pre-push-hook.bats:21` (nit): `install-sh.bats` exports a temp `HOME` and `GIT_CONFIG_GLOBAL`; `setup()` here does not. A developer's global `merge.ff=only` or `commit.gpgsign` therefore breaks the fixtures locally.

Red-first verified: with the PR's test files over the hook at main, `tests/pre-push-hook.bats` fails 8 of 11 (cases 3, 4, 6, 7, 8, 9, 10, 11), as the body states. At head it is 11/11, `tests/install-sh.bats` is 29/29, and shellcheck is clean at warning level. CI at head: all six required checks pass.

Review run by the maintainer's automated review session; verified against head 641b236d.

PR #969, opened 2026-09-07
Analytics cost view: count each API response once, read every subagent transcript, and price Fable 5.1's cache reads at its own rate

ksarma:
The analytics modal's Sessions figure over-counts the main transcript, misses every subagent transcript, and prices Fable 5.1 cache reads at four times the catalog rate. This PR makes `_session_tokens` count each API response once across the main transcript and its subagents', adds a `claude-fable-5-1` price row, and has `_price_for` match a model's own version before falling back to its family. Two commits, two files (`kernel/kernel.py`, `tests/test_token_usage.py`); independent of #956, see below.

## What breaks

On one installation the Sessions figure (`/analytics`, `_token_analytics`) read about 1.5x the spend ledger over a day. Three causes, all in `kernel/kernel.py`:

1. **Every transcript record was summed.** The CLI writes a response with several content blocks as several `assistant` records sharing one `message.id`, each carrying the response's `usage` block. `_session_tokens` appended a row per record, so one response counted 2.3 to 3.0 times (measured on that installation).
2. **Only the main transcript was read.** The CLI writes each spawned agent's conversation under `<projects>/<slug>/<sid>/subagents/`: Task agents at the top as `agent-<id>.jsonl`, and since Claude Code 2.1.261 (checked on 2.1.263) Workflow agents one level down at `workflows/wf_<id>/agent-<id>.jsonl`. None of those tokens reached the view. On that installation they were a quarter of the sessions' tokens and 62% of the output tokens.
3. **Fable 5.1 cache reads were priced 4x.** `DEFAULT_MODEL_PRICES` has no `claude-fable-5-1` row, so `_price_for`'s family fallback returned the `claude-fable-5` row, whose cache read is 1 $/Mtok. Claude Code's baked-in catalog prices Fable 5.1 at `tier_10_50_cache_read_0_25` (10 / 50 / 12.5 / 0.25 $/Mtok). The remote feed could not correct it: `_refresh_remote_prices` matches only exact (family, major, minor) signatures, and the table had no fable-5-1 signature to look for.

## Reproduction

- Write a transcript with three `assistant` records sharing one `message.id`: `_session_tokens` returns three times the usage.
- Put a transcript under `<sid>/subagents/` beside the main one: it does not count at all.
- `_price_for("claude-fable-5-1", _model_prices(now))["cache_r"]` is 1e-6 (Fable 5's rate); the catalog says 0.25e-6.

Each is one of the new tests below, and each fails against the base commit.

## The fix

**Commit 1: `_subagent_transcripts`, `_session_tok_rows`, `_session_usage`**

- One row per `message.id`, keeping the record with the largest output count. The largest-output rule handles a difference between the two transcript kinds: a main transcript's same-id records repeat one usage block, while a subagent transcript's carry the stream-start snapshot (a few output tokens) on every record but the last, which holds the final tally. Measured over 30 days of one installation's subagent files, 94% of multi-record groups differ and the last record holds the maximum in every one; a first-seen fold keeps about a tenth of the subagents' output. An id-less record stands alone. This is an observed regularity of the CLI's writer, not a documented contract; the docstring says so, and a writer that logged per-block usage deltas under one id would be under-counted.
- `_subagent_transcripts` walks the session's `subagents/` tree recursively with `os.walk` (`followlinks=False`) and skips a symlinked `subagents` directory and symlinked files, so the reader never opens anything outside the session's own tree. The CLI writes no symlinks there.
- The row cache is per file, not per session: `_session_tok_cache` maps each transcript path (the main one and every subagent's) to `((mtime, size), rows)`, so a subagent file landing or growing re-parses that file alone while the main transcript's rows stay cached, and the main file moving leaves every subagent's rows in place. Commit bb0c40b8's message describes a per-session fingerprint over every file's `(path, mtime)` instead; that was the design rejected on the way, since it re-parses the whole tree whenever any one file moves, which on a session with Workflow agents running is every build. The per-file cache is what shipped (`_transcript_tok_rows`, and `_session_tok_rows`' docstring names the alternative as rejected). `_session_tokens` and `_session_cost` were the cache's only readers; both now go through `_session_usage`, which makes one pass per session for the tokens and the priced estimate, and `_token_analytics` calls it once per session instead of twice. Payload shape unchanged.

**Commit 2: prices**

- A `claude-fable-5-1` row: 10 / 50 / 12.5 / 0.25 $/Mtok. The catalog's 1h cache-write rate (20) folds into `cache_w` at the 5m rate; the table has one cache-write column for every model, so this is the simplification it already makes.
- `_price_for` tries an exact id, then a priced model with the same `_price_sig` signature, then the family. One behaviour change to name: a dated or minor-versioned id now lands on its own version's row rather than on whichever same-family row the dict lists first (`claude-sonnet-4-6-2099...` lands on `claude-sonnet-4-6`). The existing `test_price_for_exact_then_family_then_none` still passes, since `claude-opus-4-8-20990101` signs as the opus-4-8 row either way.

Cost: one directory walk per session per `/analytics` build (behind the existing 15 s memo) plus one stat per contributing file.

## Relation to #956 and #935

#956 (`romp spend-rebuild`) recounts the ledger's token columns from the transcripts under the same two rules, once per `message.id` with subagents included. It edits `cli/spend_rebuild.py`, `_spend_windows` and the chat-body hover; it does not touch `_session_tokens`, `_token_analytics`, `_price_for` or `DEFAULT_MODEL_PRICES`, so with #956 merged the analytics view still over-counts. The two PRs are complementary and share no hunks; whichever lands second rebases. Two things I would raise on #956 rather than fold in here: its `subagents/*.jsonl` glob is flat and misses the `workflows/wf_<id>/` nesting, and its first-seen dedupe keeps the stream-start snapshot in subagent transcripts.

#935 adds `_subagents_dir(path)`, which computes the same directory `_subagent_transcripts` derives. If #935 lands first, this PR should call it instead.

## Tests

`tests/test_token_usage.py`, synthetic data only (a placeholder sid, invented counts):

- `test_records_sharing_a_message_id_count_once`: three same-id records count once; id-less records stand alone; a split whose last record carries the larger output keeps the larger figure.
- `test_subagent_same_id_records_keep_the_final_tally_whatever_their_order`: the largest-output rule in both record orders, so neither a first-record nor a last-record fold can quietly replace it.
- `test_subagent_transcripts_beside_the_main_one_count_and_refresh_the_cache`: top-level and `workflows/wf_*/` files count; a non-`.jsonl` file, a symlinked directory and a symlinked file do not; a later subagent file and a nested file's growth refresh the cache while the main file's mtime stands.
- `test_each_transcript_caches_on_its_own_stamp_so_one_file_moving_parses_only_itself`: the cache is per file (`path -> ((mtime, size), rows)`): a second subagent landing parses that file alone while the main transcript's and the first subagent's cached rows stay; the main transcript growing re-parses it while the subagents' rows stay; a same-mtime rewrite that changes a file's size still refreshes it.
- `test_a_symlinked_subagents_directory_is_not_walked_either`: `os.walk` follows its top argument even when it is a symlink, so that case is closed separately.
- `test_fable_5_1_has_its_own_row_with_the_quarter_rate_cache_read`: the row, Fable 5's own rate kept, and a dated id landing on the fable-5-1 row by signature.

All six fail on the base commit (the two id tests sum three times, the subagent and cache tests hit the missing walk, the price test gets 1e-6). The full pytest suite, `npm test` and `npm run typecheck` pass locally. The change is file and JSON logic plus `os.walk(followlinks=False)` and `os.path.islink`, whose semantics are the same on macOS; no shell surface changed.

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -26386,6 +26386,11 @@ DEFAULT_MODEL_PRICES = {   # $/token: input, output, cache write (5m), cache rea
                            # the LiteLLM refresh overwrites these with the live feed (which carries the
                            # exact ids), so they just need to be sane when offline / before the first fetch.
     "claude-fable-5":            {"in": 10e-6, "out": 50e-6, "cache_w": 12.5e-6, "cache_r": 1e-6},
+    "claude-fable-5-1":          {"in": 10e-6, "out": 50e-6, "cache_w": 12.5e-6, "cache_r": 0.25e-6},   # its
+    #   own row: without one the family fallback priced Fable 5.1's cache reads at Fable 5's $1/Mtok — 4x
+    #   the list rate. Rates read from Claude Code's baked-in model catalog (2.1.261, checked on 2.1.263),
+    #   where claude-fable-5-1 carries the tier `tier_10_50_cache_read_0_25` (10 / 50 / 12.5 / 0.25 $/Mtok;
+    #   1h cache writes are 20 and are folded into cache_w here at the 5m rate, a known simplification)
     "claude-opus-4-8":           {"in": 5e-6, "out": 25e-6, "cache_w": 6.25e-6, "cache_r": 0.5e-6},
     "claude-sonnet-5":           {"in": 3e-6, "out": 15e-6, "cache_w": 3.75e-6, "cache_r": 0.3e-6},
     "claude-sonnet-4-6":         {"in": 3e-6, "out": 15e-6, "cache_w": 3.75e-6, "cache_r": 0.3e-6},
@@ -26461,11 +26466,17 @@ def _model_prices(now=None):
 
 
 def _price_for(model, prices):
-    """The price row for a model: exact id, else any priced model of the same family
-    (fable/opus/sonnet/haiku) so a differently-dated id still gets a sane rate, else None
-    (uncounted — defensive)."""
+    """The price row for a model: exact id, else the priced model with the same (family, major, minor)
+    signature (a dated `claude-fable-5-1-2026…` lands on the fable-5-1 row, not on whichever fable row
+    the table lists first — the two differ 4x on cache reads), else any priced model of the same family
+    (fable/opus/sonnet/haiku) so an unknown id still gets a sane rate, else None (uncounted — defensive)."""
     if model in prices:
         return prices[model]
+    sig = _price_sig(model)
+    if sig:
+        for k, v in prices.items():
+            if _price_sig(k) == sig:
+                return v
     m = (model or "").lower()
     for fam in ("fable", "opus", "sonnet", "haiku"):
         if fam in m:
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -26140,7 +26140,34 @@ def _derive_judging(sid, caps, goals, t0, out, seg_ends=None):
         pass
 
 
-_session_tok_cache = {}   # path -> (mtime, [(t, in, out, cache_w, cache_r, model), ...]): per-message token rows
+_session_tok_cache = {}   # transcript path -> ((mtime, size), [(t, in, out, cache_w, cache_r, model), ...]): one
+#                           token row per API RESPONSE (message.id) in THAT file. The main transcript and each
+#                           subagent transcript cache separately, so a file that moved re-parses itself alone
+
+
+def _subagent_transcripts(path):
+    """The subagent transcripts beside a session's main one, sorted; [] when none. The CLI writes each
+    spawned agent's conversation under `<sid>/subagents/` next to `<sid>.jsonl`: Task agents as
+    `agent-<id>.jsonl` at the top, and (Claude Code 2.1.261) Workflow agents one level down at
+    `workflows/wf_<id>/agent-<id>.jsonl` — its path parser takes any extra segments — so the walk is
+    RECURSIVE (a flat listing misses every nested file; measured on one installation, those held a
+    quarter of the sessions' tokens and 62% of their output tokens). Bounded to the session's own
+    subagents tree: no symlink is followed — not a directory (os.walk's followlinks=False), not a FILE
+    (os.walk lists a symlinked file like any other and the reader would open it wherever it points; the
+    CLI writes none, so one is a user's, and it is skipped), and not the subagents directory itself.
+    Cost: one directory read per directory under it plus one lstat per file."""
+    base, ext = os.path.splitext(str(path))
+    if ext != ".jsonl":
+        return []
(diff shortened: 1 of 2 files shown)


ksarma:
Merging as shipped. The write-up owes one correction. No code needs folding in.

Reviewed inline: `_transcript_tok_rows` folds a file's assistant records by message.id and keeps the largest-output record. That fits both transcript kinds: a main transcript repeats one usage block per response; a subagent transcript holds the final tally only in its last record. Id-less records stand alone. `_subagent_transcripts` walks `<sid>/subagents/` recursively with followlinks off. It refuses a symlinked subagents directory and symlinked files. It picks up both `agent-<id>.jsonl` at the top and `workflows/wf_<id>/agent-<id>.jsonl` below it, matching what the installed CLI writes. A missing main transcript gives zeros. A subagent file gone between listing and read is skipped without a crash. Each file caches on its own (mtime, size) stamp, so a subagent landing or growing re-parses only itself. The claude-fable-5-1 row (10 / 50 / 12.5 / 0.25 $/Mtok) matches the tier_10_50_cache_read_0_25 tier in the installed Claude Code catalog. claude-fable-5 keeps its own cache-read rate. `_price_for` tries the exact id, then the same (family, major, minor) signature, then the family. Dated, [1m] and minor-versioned ids each land on their own row with no cross-family hit. Ids with no row still fall to the family row as before. #956 and #935 are both open with nothing on main, and #956's kernel.py hunks are in different functions. Nothing here re-implements or conflicts with landed work.

Findings:

- `kernel/kernel.py:26143`, follow-up, prose only. The body and commit bb0c40b8 describe a per-session (fingerprint, rows) cache keyed on every file's (path, mtime). The code caches per file on (mtime, size), and the `_session_tok_rows` docstring names the per-session design as the rejected one. The body lists five new tests where the diff adds six; the per-file cache test is unlisted. The comment at `tests/test_token_usage.py:137` still refers to a fingerprint. The shipped mechanism is the better one, and the description should say so.
- `kernel/kernel.py:26257`, nit. A subagent transcript that exists but fails to open is dropped from the sum with no trace, indistinguishable from a deleted one. The main-transcript path already behaved this way. A stderr line on the non-ENOENT case would make the undercount visible.

Red-first met: with the PR's test file checked out at origin/main, the six new tests fail (the same-id fold, the subagent order fold, the subagent walk, the symlinked directory, the per-file cache stamp, and the Fable 5.1 row) and 21 pass. At cb14b937 `tests/test_token_usage.py` is 27 passed, stable over three xdist runs and five serial repeats. CI at head: all six required checks pass.

Review run by the maintainer's automated review session; verified against head cb14b937.

PR #970, opened 2026-09-07
File viewer: a session chip in the title bar, and a selection quote seeds the composer from the feed-hosted viewer too

ksarma:
Two changes to the shared file viewer (`ui/webview/file-view.ts`), one commit each. Both are browser-dashboard only; the VS Code extension's behavior is unchanged.

## 1. The title bar names the session the file was opened from

**What is wrong at tip.** The viewer carries a sid (for the GitHub link, the file browser and `fileUrl`'s federation routing) but never shows it. With several sessions on one repository, or on worktrees of it, a file opened over one chat looks the same as the same file opened over another.

**Reproduction.** Two sessions, `web` and `api`, on a `notes-api` checkout. Click `app.py` in `web`'s transcript, then the same path in `api`'s. The two viewers are indistinguishable; the tab color behind the dimmed backdrop is the only cue, and the modal covers it.

**The change.** A session chip sits in the title bar between the path and the actions: a pill in the session's identity color (the color its tab wears), with the `host:` token quiet inside for a remote (federated) session and marked while that host's link is down (`hostNameNodes`), tooltip "Opened from the api session". The chip is fixed width; only the directory half of the path yields room.

The identity is resolved from the sid inside `file-view.ts` rather than threaded through arguments, because the openers mostly know only a sid (the relay branch `initFileView` keeps and the conflict Reload live in the module; the file browser hands over a bare sid). Each hosting document registers a lookup once beside its `initFileView` boot:

- `render.ts`: `sessions`, then `tabMeta` (the ladder `renderTabs` uses, so a dormant session keeps its name and color).
- `feed.ts`: `sessionsMeta`, then the asks list. Never `sessionColors`, which is keyed by name.

A sid neither list can name falls to the same 8-character stub the kernel's `_peer_identity` uses, uncolored. No sid, or no registered lookup, means no chip: an identity is looked up, never invented. `openFileView` and `initFileView` keep their signatures.

CSS lands in both mirrored sheets. `.fileview-sess .host-prefix` sets `color: inherit`: the global `.host-prefix { color: var(--dim) }` would otherwise beat the pill's inline foreground, and a remote session's `host:` token sat at about 1:1 contrast on a saturated pill.

One sentence in the guide's "Reviewing a document" paragraph.

## 2. A passage selected in the feed-hosted viewer reaches the chat composer

**What is wrong at tip.** The guide says any passage selected in the file viewer lands in the composer as a quote chip. The viewer opened from the feed pane's file browser posts its `editorSelection` message to its own window, where nothing listens (`feed.ts` has no composer and no handler). The gesture is dead, and each selection still pays a fresh read of the file. `file-view.test.ts` recorded this as by design.

**Reproduction.** In the feed pane, Browse files for `api`, open a file, select a line. Nothing appears in any composer.

**The change.** `composerWindow()` resolves where the seed goes: this window when it holds `#composer-input` (the chat-hosted viewer, unchanged); else the same-origin shell when it frames this document (detected by the shell's `#chat-pane` node, the idiom `inRompShell` in `render.ts` already uses); else `null`, and the gesture stands down before the fresh read (a cross-origin parent such as the VS Code webview host, or a standalone pane). The shell's message listener in `_LANDING_SETTINGS_JS` gains one arm that forwards an `editorSelection` message whole into the chat iframe. The chat's existing handler seeds the chip for `m.sid`, so the quote lands in the browsed session even if the active tab changed. The feed gains no listener; `render.ts` and the sheets are untouched.

Design points to rule on:

- The new arm checks the message shape only, not `e.origin`, matching every other arm in that listener. An origin check is a one-token change; say so if you want it and the TS pin follows.
- On the phone layout the chip lands in the chat tab without switching tabs, since reviewing a document is a select-several-passages loop.
- A standalone `/feed` page or a cross-origin-framed page now skips the fetch on selection instead of fetching into dead air.

## Tests

- `file-view.test.ts`: source pins for the resolver export, the chip build and both registrations; an executed replica of the resolver ladder (named session, host-prefixed 8-character stub, no chip) on synthetic rows; pins for `composerWindow`'s body, the gate order and the shell arm (kernel.py read from the TS suite, as the relay test already does); `composerWindow` lifted from the source and run against shimmed window/document pairs for six hosting cases; the media-gate order pin moved behind the target gate.
- `fileview-chip.test.ts` (new): `openFileView` runs against the DOM stand-in `github-link.test.ts` uses; asserts the chip's text, title, inline colors and position for a registered identity, the `host:` token for a remote one, the uncolored stub for an unknown sid, and no element for no sid or no resolver.
- `fileview-parity.test.ts`: the two new rules are byte-equal in both sheets, `color: inherit` included.
- `tests/test_kernel_mobile.py`: the shell arm and the `f-chat` iframe id, from pytest.

Red first: with the source hunks reverted, 9 UI tests fail for the chip (4 pins, 4 executed, 1 parity) and 4 UI plus 1 Python fail for the seed; dropping `color: inherit` alone fails the CSS pin. Typecheck, the file-viewer test files and the two Python modules that read the shell constant are green locally; CI runs the full suites.

## Notes

Both changes have run on a self-hosted instance since 2026-09-03. The build machine is headless, so the phone layout deserves a real-browser glance. #958 and #959 touch `file-view.ts`'s import block, `closeFileView`'s neighborhood, both sheets and the same guide paragraph; the overlap is adjacent lines only (#958's URL viewer intentionally carries no chip: a URL has no sid), and this rebases after them. The file browser's own bar stays unlabeled; a possible follow-up on the same resolver.

## From review

Two nits from the review of #970, folded into the first commit (both sheets and the CSS pin in `file-view.test.ts`):

- The chip is `font-size: 0.82em`, the size of the bar's buttons, instead of `11px`. The px value was a third size on a bar whose text is `0.86em` and `0.82em` of `--fs`, and it did not scale with its neighbors.
- The chip is `display: block` instead of `inline-flex`, which makes `text-overflow: ellipsis` take effect: the property acts on block containers only (a flex container's text sits in an anonymous flex item the property cannot reach), so an over-long name hard-clipped at `max-width`. `align-items: center` goes too; the bar's own `align-items` centers the pill, and its single nowrap line centers within it. Truncation is made real rather than the declaration dropped because an ellipsis tells the reader there is more (the tooltip carries the full name); a clip through a glyph reads as a rendering fault.

The pin asserts `display: block`, `font-size: 0.82em` and the overflow triple and rejects `inline-flex`, `align-items` and a px size; the parity test keeps the rule byte-equal in both sheets. Red first: with the fold reverted the pin fails for both sheets; with one sheet reverted the parity test fails too.

Tier: feature

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -28,8 +28,11 @@ passage, and repeat — each staged note remembers its quote and its place. **�
 everything you staged along with whatever is in the box, so the session applies the lot
 in one pass, and you never copy a line out of the document by hand. The line in each
 label is checked against the file at the moment you select, so numbers that moved under
-you are caught rather than quietly carried. The title bar's **GitHub ↗** button opens the
-file on GitHub. While the check runs, the button waits dimmed with pulsing dots beside it.
+you are caught rather than quietly carried. When several sessions work in the same
+repository, or in worktrees of it, the viewer's title bar says which one you opened the
+file from: a chip with the session's name, in the same color as its tab. The title bar's
+**GitHub ↗** button opens the file on GitHub. While the check runs, the button waits dimmed
+with pulsing dots beside it.
 When there is nothing to open, the button stays in place, dimmed, and a caption beside it
 says why (the file is not in a git repository or not committed — untracked, staged but in no
 commit, or on a branch with no commits yet — the repository has no origin remote, its origin
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -31984,6 +31984,13 @@ if(m.romp==='browseFiles'){var bf=document.getElementById('f-feed');
     try{window.__rompPaneToggle&&window.__rompPaneToggle('feed',true);}catch(e){}}
   try{window.__rompMobileTab&&window.__rompMobileTab('feed');}catch(e){}   // phone: one pane at a time
   try{bf&&bf.contentWindow&&bf.contentWindow.postMessage({romp:'browseFiles',path:m.path,sid:m.sid},'*');}catch(e){}}
+// A passage selected in a viewer hosted by a pane with NO composer (the feed) posts up in the
+// editorSelection shape the chat already handles (file-view.ts composerWindow); the shell forwards it
+// whole into the chat pane, whose composer seeds the labeled quote chip for the session the file was
+// opened for (m.sid beats the active tab there). The chat-hosted viewer posts to its own window and
+// never gets here.
+if(m.type==='editorSelection'&&typeof m.text==='string'){var fc=document.getElementById('f-chat');
+  try{fc&&fc.contentWindow&&fc.contentWindow.postMessage(m,'*');}catch(e){}}
 // the browser owns the restore: browseClosed alone puts a brought-forward feed back the way it was
 if(m.romp==='browseClosed'&&window.__rompFeedWasOff){window.__rompFeedWasOff=false;
   try{window.__rompPaneToggle&&window.__rompPaneToggle('feed',false);}catch(e){}}});
--- a/tests/test_kernel_mobile.py
+++ b/tests/test_kernel_mobile.py
@@ -260,6 +260,17 @@ class LandingShell(unittest.TestCase):
         self.assertIn("html.ios-standalone #mtabs{padding-bottom:env(safe-area-inset-bottom,0px)}", html)
         self.assertIn("#mtabs{display:flex;position:fixed;left:0;right:0;bottom:0", html)
 
+    def test_the_shell_forwards_a_quote_seed_from_a_composerless_pane_into_the_chat(self):
+        # A passage selected in the file viewer hosted by the FEED pane (the file browser's document)
+        # has no composer in its own document; file-view.ts posts the editorSelection message up to the
+        # shell, which forwards it whole into the chat iframe, whose existing handler seeds the labeled
+        # quote chip for m.sid. The arm sits in the same listener as the browseFiles relay, and the chat
+        # iframe carries the id it keys on.
+        js = km._LANDING_SETTINGS_JS
+        self.assertIn("if(m.type==='editorSelection'&&typeof m.text==='string'){var fc=document.getElementById('f-chat');", js)
+        self.assertIn("try{fc&&fc.contentWindow&&fc.contentWindow.postMessage(m,'*');}catch(e){}}", js)
+        self.assertIn("<iframe id=f-chat class=m-on src=/chat>", km._landing())
+
 
 class TimelineTouchSurface(unittest.TestCase):
     def test_timeline_fits_svg_and_drops_overflow_scroller_on_touch(self):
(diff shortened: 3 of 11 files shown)


ksarma:
Recommend merging after the comment window; nothing to fold.

Reviewed inline: the chip's name and color come from kernel-owned records only. `render.ts` reads sessions, then tabMeta. `feed.ts` reads sessionsMeta from the kernel's sessions frame, then asks. A sid neither names gets the kernel's own 8-character stub, uncolored. Without a sid or a lookup there is no chip. The chip is a non-interactive span; the tooltip is its second level. It uses no accent, and the parity pin holds both sheets byte-equal. `composerWindow` returns this window when a composer exists, else the same-origin shell, else null. The gesture stands down before the fresh read. The viewer trims the selection and returns on empty. The chat handler requires a non-blank string again. `kernel.py` has one editorSelection arm, forwarding only into the chat iframe. The chat-hosted viewer never reaches the shell. `seedEditorQuote` replaces the single labeled chip in place and edits composerCitations only. As a result, nothing double-seeds and the draft is untouched. On design, the shape-only check is fine as shipped, because the chat iframe was already reachable by `postMessage`. The worst case is a visible chip that sends nothing. The cross-origin and standalone stand-down is correct. Staying on the feed tab on the phone is right, but the gesture then has no visible acknowledgement in the viewer. A follow-up could add a short cue in the bar.

Two nits for a follow-up:
- `font-size: 11px` (styles.css:3289, feed.css:1397) adds a third size, in px, to a bar whose text is 0.86em and 0.82em of `--fs`. The chip therefore stops scaling with its neighbors.
- `text-overflow: ellipsis` (styles.css:3290, feed.css:1398) is inert on an inline-flex container, so an over-long name hard-clips at `max-width`.

Tests at head: `test_kernel_mobile.py` 34 pass; file-view, fileview-chip and fileview-parity 42 pass; typecheck clean. Against origin/main with the PR's test files, 8 file-view, 4 chip, 1 parity and 1 Python test fail, matching the body. CI: six of six checks pass at d64ac698.

Review run by the maintainer's automated review session; verified against head d64ac698.

ksarma:
Recommend merging after the comment window; nothing to fold.

Reviewed inline: this second round covers the delta since d64ac698, which is exactly the three files the body names (`feed.css`, `styles.css`, `file-view.test.ts`; 23 insertions, 7 deletions). It folds both nits from the first round. The chip rule is now `display: block` at `font-size: 0.82em`, byte-equal in both sheets. The chip is appended directly to `.fileview-bar` between the name and the actions. So it stays a flex item of the bar's centered row, and `display: block` only changes its inner layout. The pill, its inline identity colors, and the `host:` span with `color: inherit` are unaffected. As a block container with overflow hidden and nowrap, the chip now ellipsizes an over-long name. The inline-flex version hard-clipped it at max-width. No ancestor between body and the chip sets a font-size, so 0.82em resolves to the same size as `.fileview-btn` with no compounding. The bar uses two sizes instead of three. The CSS pin asserts the new declarations within the `.fileview-sess` rule alone and rejects inline-flex, align-items and any px size. Everything the first round confirmed is byte-identical and was not reviewed again: the resolver ladder and its registrations, composerWindow, the shell's editorSelection forward, and the chip tests.

Locally at head, the three viewer test files pass 42/42 and `tests/test_kernel_mobile.py` passes 34/34. With both sheets reverted to d64ac698, the pin fails alone. With only one sheet reverted, the pin and the parity test both fail, as the body states. All six required checks are green at e8da8022.

Review run by the maintainer's automated review session; verified against head e8da8022.

PR #971, opened 2026-09-07
Review follow-ups from #968, #967, #964 and #965: one commit per merge comment

ksarma:
Five commits, one per follow-up the merge reviews of #968, #967, #964 and #965 filed. Each fix landed with its test failing first against the merged code.

## `.githooks/pre-push`: the exclusion covers every fetched remote (the review of #968)

The added-lines pass scanned every commit the pushed-to remote lacked (`--not --remotes=<remote>`). In a clone with a fork and the project it forked from, a clean branch cut from the project's main is pushed to the fork as a new ref; no fork ref reaches the project's newer commits, so the range held them all, and one that added a string the project had since redacted was named as the branch's own addition. Nothing on the branch could clear that: the footer's merge-main remedy addresses a tip that inherits a string, not a range that holds the project's history. Syncing the fork's main was refused the same way.

The exclusion is now every remote-tracking ref (`--not --remotes`). A commit any fetched remote already holds was published there, and the fix for a string it carries is forward on that remote. A commit no remote has is still scanned on the way to either remote. The footer names the main that carries the redaction and adds a fetch step for a refusal that names a commit already on some remote; `CLAUDE.md` and the `install-sh.bats` header say "new to every fetched remote". The suite's `setup()` also exports a temp `HOME` and `GIT_CONFIG_GLOBAL` (the review's nit), so a developer's global `merge.ff` or `commit.gpgsign` cannot reach the fixtures.

Tests: the per-remote case in `tests/pre-push-hook.bats` is replaced by three two-remote cases (origin as the fork, upstream as the project): a clean branch cut from the project's main passes to the fork as a new ref; syncing the fork's main passes; a leak local to the clone is still refused without re-flagging the project's commit. The tip case pins the qualified remedy. Against the merged hook 4 of 13 fail (cases 7, 10, 11, 12); 13/13 after.

## `.githooks/pre-push`: a merge's additions are the lines in none of its parents (the review of #968)

Every commit was diffed against its first parent, a merge included, although the body of #968 said only a conflict resolution would count as a merge's own addition. A branch that merged main while main carried a string, then merged the redaction, was refused over the first merge: its first-parent diff added the string, but the content came in through the second parent, which the remote already had, and the branch's tip was clean.

A merge is now read through the combined diff against all its parents (`git diff-tree -c`), keeping the lines that carry a `+` in every parent's column: content in no parent, a conflict resolution or a line typed into the merge. `-c` omits outright a file that matches any parent, which holds no such line. One `diff-tree` form serves every commit: on a single-parent commit `-c` is the ordinary diff, `--root` covers a root commit, and the awk sizes the column prefix from `rev-list --parents`. The path header is read on `diff --combined` as well as `diff --git`; rename detection stays.

Tests: a merge of main taken while the leak was live passes once the tip merged the redaction (refused by the first-parent diff, naming the merge); a string typed into a merge's conflict resolution is refused naming the merge. Against the hook after the first commit 1 of 15 fails; 15/15 after. `install-sh.bats` 29/29 throughout; shellcheck clean at warning level.

## `bin/romp-manager`: the server's cgroup is compared with the manager's own (the review of #967)

`placementFromCgroup` returned `unscoped` for every cgroup that was not a `romp-tmux-N` scope, and the line then said a service restart would take the server down. A default-socket server the user started from a terminal sits in the login session's scope, which is neither a romp scope nor the service's cgroup; a restart leaves it alone, so the line predicted a loss that would not happen, the fault the #953 follow-up had just removed for the scoped cases.

The placement now reads the answer: the server's `/proc/<pid>/cgroup` is compared with the manager's own `/proc/self/cgroup`. The same cgroup, or one under it, is the service's (`KillMode=control-group` takes the subtree) and stays `unscoped`, the line as it always was; a romp-tmux scope stays `scoped`; any other cgroup is `outside`, and the line names it and predicts no loss. `systemdCgroupPath` picks the hierarchy systemd manages from either cgroup version (the `0::` line, or v1's `name=systemd` controller). `tmuxServerPlacement` reads the server's file, then `self`; a failure of either read is `unknown` with its first line, as before.

Tests: `tests/manager-tmux-scope.test.js` drives the classification with synthetic `/proc/<pid>/cgroup` texts: a login-session server against a service-run manager is outside and named; the manager's own cgroup and a child of it are unscoped; a romp-tmux scope is scoped wherever the manager sits; v1's `name=systemd` hierarchy is the one compared; an empty file on either side is unknown, saying which; the outside line names the cgroup. Against the merged manager 5 of 31 fail; 31/31 after, 75/75 across the manager suites. `tests/romp-manager-tmux-scope.bats`: the pid-1 case now reads as outside the service's cgroup (pid 1's cgroup is never the manager's own outside a container, which the case skips on); it fails against the merged manager at the outside-line grep; 5/5 after.

## `ui/webview/gear.js`: a refused Default effort pick reads as Default (the review of #967)

The effort selects' Default option is the empty value (no effort flag), and `staleWord('')` read it as no value: a refused Default pick drew the value-less copy and a plain Apply anyway, in the frozen-tab case the copy exists for.

`STALE_WORDS` maps, per setting, the sentinel options whose value is not the word the select shows: the effort selects' `''` and the distilling pair's `none` read as Default, `triage` as Follow triage, the comment pair's `session` as Same as the session and `default` as Default. `staleWord` takes the setting and words the refused and the kept value through it; any other string still reads as itself, booleans as on/off. Apply anyway re-issues the echoed gesture as sent: the value, never the word.

Tests: `setting-stale-fold.test.ts` drives a refused Default triage effort ("Triage effort: Default was not applied on web. Keeping high.", an "Apply Default anyway" button whose click posts the empty value), the distilling and comment sentinels by name, and a plain level as itself with a kept Default named. `setting-stale.test.ts` pins the helper's shape and, generically, that every literal sentinel option `paintChoices` renders under another name has that name in `STALE_WORDS`. Against the merged gear 2 of 2842 fail; 2842/2842 after, typecheck clean; the pytest modules that read `gear.js`: 717 passed.

## Wording follow-ups (the reviews of #964 and #965)

- `kernel/session_backend.py`: the `set_effort` docstring names Codex beside the SDK as refusing an unknown sid, as the `set_model` paragraph already does.
- `tests/test_kernel_postal_isolation_routes.py`: `PolicyPins` pins the skill's `kind` bullet (the hook's was pinned in `romp-postal-context.bats`, the skill's nowhere): the required parameter, its three values, no retired body prefix. Against a `SKILL.md` without the bullet it fails at the first assertion.
- `tests/test_session_move_live.py`: the first two `ApiKeyHelperPrecedence` tests patch `HOME` as well as `CLAUDE_CONFIG_DIR`, so the class docstring holds.
- `ui/webview/pane-shim-stale.test.ts`: the foreground-quiet comment no longer cites "the body".

## Verification at head

`bats tests/pre-push-hook.bats` 15/15; `tests/install-sh.bats` 29/29; `tests/romp-manager-tmux-scope.bats` 5/5; `tests/romp-postal-context.bats` 3/3; `node --test tests/manager-*.test.js` 75/75; `npm test` 2842/2842 and `npm run typecheck` clean; `shellcheck -S warning .githooks/pre-push` clean. Full pytest: 7616 passed, 46 skipped (a first run had three xdist-ordering failures in `tests/test_postal_read_receipts.py`, a module this branch does not touch, which passes 16/16 alone and in the rerun).

Tier: fix

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -44,26 +44,28 @@ while IFS= read -r line; do
 done < "$strings_file"
 [ ${#grep_args[@]} -gt 0 ] || exit 0
 
-# $1 is the remote being pushed to: a name, or a bare URL when pushing by URL.
-remote="${1:-}"
 zero=0000000000000000000000000000000000000000
 
-# The commits THIS push publishes for one ref: what the remote does not already
-# have. "Already has" means reachable from ANY of the remote's refs as this clone
-# last fetched them (refs/remotes/<remote>/*), not only from the ref being pushed.
-# The difference is a branch that merged main: its range holds main's commits,
-# every one of them on the remote already through main. Scanning those refuses
-# the push over a string that main itself published, and a string already on the
-# remote is fixed FORWARD, on main (redact, push main) — not by blocking every
-# branch that merged main before the redaction. The exclusion is scoped to the
-# remote being pushed to: a commit only another remote has is still new to this
-# one. With no remote-tracking refs for the remote (never fetched, or a push to a
-# bare URL) the exclusion matches nothing and the range falls back to everything
-# the remote ref lacks — for a new ref, the whole history.
+# The commits THIS push publishes for one ref: what no remote already has.
+# "Already has" means reachable from the remote ref's current value or from ANY
+# remote-tracking ref this clone holds (refs/remotes/*, as last fetched) — every
+# remote's, not only the pushed-to remote's or the ref being pushed. The
+# difference from the ref's plain update range is a branch that merged main: its
+# range holds main's commits, every one of them published already through main.
+# Scanning those refuses the push over a string that main itself published, and a
+# string already on a remote is fixed FORWARD, on main (redact, push main) — not
+# by blocking every branch that merged main before the redaction. Every remote
+# counts because that remedy exists only for a commit no remote has yet: in a
+# clone with a fork and the project it forked from, a clean branch cut from the
+# project's main is pushed to the fork as a new ref, and an exclusion scoped to
+# the fork left the project's own commits in the range — one named as adding a
+# string the project had since redacted, which no rewrite of the branch could
+# fix (the #968 review). With no remote-tracking refs at all (never fetched) the
+# exclusion matches nothing and the range falls back to everything the remote
+# ref lacks — for a new ref, the whole history.
 rev_range() {   # <local_sha> <remote_sha> → a rev-list argument list
-    local args="$1 --not"
+    local args="$1 --not --remotes"
     [ "$2" = "$zero" ] || args="$args $2"
-    [ -z "$remote" ] || args="$args --remotes=$remote"
     printf '%s' "$args"
 }
 
@@ -118,7 +120,8 @@ if [ "$blocked" -ne 0 ]; then
     echo "romp pre-push: BLOCKED — a banned string from $strings_file is in a pushed commit." >&2
     echo "  Replace it with synthetic data (see CONTRIBUTING.md), then push again." >&2
     echo "  If the string is only in an INTERMEDIATE commit, squash or rewrite the branch." >&2
-    echo "  If the tip only INHERITS a string an older commit added and main has since redacted, merge main." >&2
+    echo "  If the tip only INHERITS a string an older commit added, merge the main that has since redacted it." >&2
+    echo "  If the named commit is already on some remote, this clone has not fetched it: fetch, then push again." >&2
     echo "  To bypass for one push (you are sure it is fine): git push --no-verify" >&2
     exit 1
 fi
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -12,15 +12,17 @@
 #
 #   the tip's tree    the content the push exposes. A banned string anywhere in
 #                     it refuses the push.
-#   each new commit   the lines it ADDS against its first parent. A string
-#                     introduced in an intermediate commit and removed by a
-#                     later one is still caught, and the commit named; a commit
-#                     whose tree merely INHERITS a string an older commit put
-#                     there is not. Every commit on every branch cut from main
-#                     between a string landing there and its redaction inherits
-#                     it that way, and a per-commit TREE scan refuses them all,
-#                     over files the branch never touched, when the only
-#                     exposure any of them could add is its tip.
+#   each new commit   the lines it ADDS: against its parent, or for a merge the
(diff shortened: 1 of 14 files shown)


ksarma:
Merging once the checks are green. Nothing to fold.

Reviewed inline: each of the five commits against the merge comment it answers. Each touches only the files its message names. The hook's added-lines pass excludes every commit a remote-tracking ref reaches. The tip-tree grep is unchanged. So the only exempt strings are those added and removed inside the pushed range that some fetched remote already holds. In a scratch two-remote repo with an invented string, a clean branch cut from an upstream that holds a redacted commit pushes to the fork. A leak no remote holds is refused, naming its commit. A tip leak is refused. A merge's additions are the combined diff against all parents with a plus in every column, sized from `rev-list --parents`. The same `diff-tree` form serves single-parent and root commits. romp-manager compares the server's cgroup with its own: the `0::` line on v2, the `name=systemd` controller on v1. A failed read or an empty file on either side yields unknown, with the reason. The outside placement names the cgroup. gear.js maps a sentinel value to the word its select shows through a guarded per-setting lookup. The lookup cannot throw on an absent setting or an unmatched value. Apply anyway still posts the value. The fifth commit is text only.

- `.githooks/pre-push:112`: the comment above the rev-list loop still says new to the remote. The rule is now every fetched remote.
- `ui/webview/setting-stale.test.ts:185`: the sentinel pin searches the whole `STALE_WORDS` source for each value and label pair. A pair filed under the wrong setting would pass. Key the check by setting.

Red-first per commit: `pre-push-hook.bats` fails cases 7, 10, 11, 12 against the base hook and 13 against the first commit's. `manager-tmux-scope.test.js` fails 5 of 31 and `romp-manager-tmux-scope.bats` case 2 against the base manager. The stale ts tests fail 2 against the base gear.js.

At head: `pre-push-hook.bats` 15/15, `install-sh.bats` 29/29, `romp-manager-tmux-scope.bats` 5/5, manager suites 75/75, `npm test` 2842/2842, typecheck and shellcheck clean, touched pytest modules 15 passed.

CI: Shell and vscode-extension pass. The four Python jobs fail on one test in `tests/test_stage0_log_readers.py`, a file this PR does not carry. main's own run at 2e3e8151 fails on the same test, so the failure is main's.

Review run by the maintainer's automated review session; verified against head 7f238149.

ksarma:
Re-triggering CI against main now that the test_stage0_log_readers failure is fixed there (#972); the head is unchanged.

PR #995, opened 2026-09-07
The converge stands down for a quiet deploy already parked for the code on disk

henrypinkard:
**Bug (T240d).** A peer's p2p apply advanced this box's checkout and asked the manager for a **quiet** restart. The kernel's drift check then saw the checkout ahead of the running kernel and posted an **immediate** restart-all, so the quiet window the peer asked for never ran. Ledger evidence from 2026-09-07: a p2p-update quiet row at 09:23:39 PT, then a main-converge/now row at 09:27:27 PT whose cut stopped ten sessions, a running review workflow among them. The same shape at 01:42 PT two seconds apart on the pull side, where origin read ahead of the just-reset checkout.

**Fix, event-based.** Before converging in auto mode, the drift check reads the pending request row. When the newest restart-requesting audit row is a quiet p2p-update (or a quiet main-converge) for the sha the checkout holds, no cut row has consumed it (`auditT`), and it is inside the window a quiet request stays pending, the check **stands down**: one log line per sha ("already parked as a quiet deploy — leaving it to the quiet window"), the sha left unmarked so the pass after the row is consumed or expires re-evaluates. The pending window is the far manager's backstop bound, exactly what `_recent_restart_audit` already gives such a row for cut attribution, so a park lost with its manager self-heals there rather than holding forever. The pull side stands down too: the parked restart delivers the code on disk first, and any pull follows on the pass after it lands. One checkout read feeds both the verdict and the match.

**Unchanged.** The cool-down, the immediate policy for a genuinely new pull, and the manual-mode banner (a click is the user's explicit choice to restart now).

**Writers.** The converge's audit row now carries the sha it deploys, so a quiet `romp refresh` is matched rather than guessed at; a legacy quiet row without a sha matches nothing. The p2p apply script writes its quiet row right after the reset and before the owner check, whose manager status call was a window in which the checkout was already ahead with no row on disk. When no owning manager answers, the fallback's immediate row is newer and supersedes it for every reader.

**Review folds (second commit).** Adversarial review, two lenses, twelve findings each verified hermetically; five confirmed and folded:
- The pull-side stand-down applies only while the kernel does not yet run the checkout. A quiet deploy that landed but lost its cut row leaves its row unconsumed for the rest of the window, and a pull must not wait on a park that already delivered.
- The once-per-sha latch clears, and the end of a stand-down is said once ("no longer pending") instead of silently proceeding.
- The pending-row reader walks a 200-row tail. Fifty session self-closes inside a parked window had pushed the live park's row out of a 50-row tail.
- `romp refresh --quiet` wrote a row with no `when` and no `sha`, so the CLI's own quiet park was pre-empted just the same. The CLI row now carries both under `--quiet` (pinned by a new bats file against fake manager and postal stand-ins), and the reader accepts that action-less row. The main-converge quiet branch has no production writer today; the test comments now say so.

Stated, not folded: the pending window is the 20-minute bound the cut-attribution reader ties to the manager's default 15-minute backstop. A box that raises the manager's backstop past it by environment override converges over the tail of its own park. Refuted: short-sha matching hazards, the test's unconditional unlink, a future-stamped row, and the landed-with-lost-cut-row delay (masked by the default cool-down and removed by the first fold).

**Tests, red first.** The 09:23 PT ledger shape posts the restart on main and stands down here, with one line and no second line on the next pass; the pull shape stands down; consumed, other-sha, immediate and expired rows proceed; a quiet converge row with a sha stands down and a legacy one proceeds. Source pins hold the apply script's reset → row → owner-check order and the sha on the converge row.

### DIFF
--- a/bin/romp
+++ b/bin/romp
@@ -1714,13 +1714,21 @@ case "${1:-}" in
                    RA_PPID="$PPID" \
                    RA_PARENT="$(ps -o command= -p "$PPID" 2>/dev/null | head -1)" \
                    RA_TTY="$(tty 2>/dev/null || true)" \
-                   RA_SID="$_ra_sid" RA_NAME="$_ra_name" \
+                   RA_SID="$_ra_sid" RA_NAME="$_ra_name" RA_WHEN="${2:-}" \
+                   RA_SHA="$(git -C "$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)" rev-parse --short=8 HEAD 2>/dev/null || true)" \
                    python3 - >> "$_ra_dir/restart-audit.jsonl" <<'PYEOF'
 import json, os, time
-print(json.dumps({"t": int(time.time()), "ppid": int(os.environ.get("RA_PPID") or 0),
-                  "parent": os.environ.get("RA_PARENT") or "", "sid": os.environ.get("RA_SID") or "",
-                  "name": os.environ.get("RA_NAME") or "", "tty": os.environ.get("RA_TTY") or "",
-                  "tmux": os.environ.get("TMUX_PANE") or ""}))
+row = {"t": int(time.time()), "ppid": int(os.environ.get("RA_PPID") or 0),
+       "parent": os.environ.get("RA_PARENT") or "", "sid": os.environ.get("RA_SID") or "",
+       "name": os.environ.get("RA_NAME") or "", "tty": os.environ.get("RA_TTY") or "",
+       "tmux": os.environ.get("TMUX_PANE") or ""}
+if os.environ.get("RA_WHEN") == "--quiet":
+    # a PARKED restart: the kernel's drift check reads when + sha to stand down for it (T240d) —
+    # without them an immediate converge pre-empted the quiet window this flag asked for
+    row["when"] = "quiet"
+    if os.environ.get("RA_SHA"):
+        row["sha"] = os.environ["RA_SHA"]
+print(json.dumps(row))
 PYEOF
                } 2>/dev/null || true
                "${ROMP_POSTAL_BIN:-romp-postal-service}" restart >/dev/null 2>&1 || true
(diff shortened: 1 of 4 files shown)


ksarma:
Recommend approval at 5f20e187. One stale sentence remains as a follow-up; nothing else is left to fold. The second commit landed while this review ran, so the verification below was redone against it.

Reviewed inline: the stand-down runs in the auto branch of the drift check, after the slot mark and before the cool-down. Manual mode, the cool-down, and the immediate policy for a new pull are unchanged. The pending window is the 20-minute bound that `_recent_restart_audit` already applies to a quiet row for cut attribution (the manager's 15-minute backstop plus margin). Consumption is the newest cut row's `auditT`, so the stand-down ends on the cut that lands the deploy. A park lost with its manager delays a needed converge by at most that window plus one drift pass. The converge then reports this once. Row matching holds: a sha prefix matches in either direction. The newest row wins, so the fallback's immediate row supersedes the quiet one. A quiet row with no sha matches nothing. The relaxed action guard admits only `bin/romp`'s refresh row; remote-restart and service rows carry actions. The once-per-sha line clears on an in-sync pass. Both new lines are kernel stderr, not text injected into a session, so the injected-voice rule does not apply. The pull side stands down only while the kernel does not yet run the checkout. The apply script writes its quiet row after the reset and before the owner check.

- Follow-up, `tests/test_main_drift_notice.py:382` and the body's Writers paragraph: both still say the converge row's sha matches a quiet romp refresh. At this head the CLI row's own `when` and `sha` do that. The converge's quiet branch has no production caller, as the Review folds section says.

Red first at origin/main: the two new tests in `tests/test_main_drift_notice.py` fail (2 failed, 32 passed). The first fails on the missing new global before its behavioral assertion; a `getattr` default would fix that. The first case of `tests/romp-refresh-audit.bats` fails because the refresh row carries no `when` or `sha`. At 5f20e187: 34 passed, bats 2 of 2, and 83 related restart and remote-update tests pass. CI: all seven checks green at 5f20e187; `mergeStateStatus` read UNKNOWN at query time.

Review run by the maintainer's automated review session; verified against head 5f20e187.

PR #1008, opened 2026-09-07
Session scopes: opt-in memory limits and an OOM-score adjustment, so a runaway process dies inside its own session

ksarma:
Depends on #1002 (the env-file warning PR), which must merge first; GitHub shows its commits in this diff until then.

Stacks on the env-file warning PR (branch envfile-warning-offer): review and merge that first; this PR's own diff is 026923a53..9f4891c7e980cf02f7a959cf6275d110bb8acf86.

## What this adds

Four opt-in service-environment variables put a memory limit and an OOM-score adjustment on each session's scope (the per-session scopes from #953), so a runaway process dies inside its own session instead of taking the kernel down. Nothing is set by default: with the variables unset, every launch is byte for byte what #953 does today (pinned by the bats case "none, or empty ones, bring no -p at all" and the unchanged exact-argv tests).

## Why

A session's shell expanded a glob over a large temp directory, grew past 30 GB, and the machine's userspace OOM killer took the largest process it saw: the romp kernel, which cut every session. With each CLI in its own scope, a memory limit on the scope makes the runaway die inside its own session. The cgroup's OOM killer SIGKILLs the largest process in the scope and nothing else, the session sees a failed tool call (exit status 137), and the kernel and the other sessions are untouched.

## The variables

- `ROMP_CLI_SCOPE_MEMORY_MAX` becomes `MemoryMax=` on the scope (the hard limit).
- `ROMP_CLI_SCOPE_MEMORY_HIGH` becomes `MemoryHigh=` (throttle; the limit itself kills nothing. The docs note that `systemd-oomd`, where it is set to kill on the user manager's memory pressure, can still kill a throttled scope, and say how to check that setting).
- `ROMP_CLI_SCOPE_MEMORY_SWAP_MAX` becomes `MemorySwapMax=`. Without it a scope over `MemoryMax` swaps instead of dying, for as long as swap lasts (verified: a 300 MB hog under `MemoryMax=64M` survived on a machine with a swapfile; with `MemorySwapMax=0` it was killed).
- `ROMP_CLI_SCOPE_OOM_SCORE_ADJ` is written by the wrapper to its own `/proc/self/oom_score_adj` before the exec, on the fallback path too, so the CLI and everything it spawns inherit it and the kernel keeps its own; machine-wide killers then prefer a session over the kernel.

Whenever a memory property is set the wrapper adds `OOMPolicy=continue`: a scope's default is `stop` (systemd 255), under which one OOM kill makes systemd stop the whole scope, tmux servers included, the opposite of what the limit is for.

## Validation at both ends

The kernel checks the values once at start (`cli_scope_limits`). A value that fails its rule is a problem line naming the variable and the rule, and it reaches the wrapper EMPTY, so an unparsable value never makes `systemd-run` refuse to start the CLI; an unset variable is not sent at all. The wrapper re-checks on every launch and reports a value it refuses, a property systemd rejects, or an adjustment Linux will not let it write as a third stderr form, `romp-cli-scope: ignored: …`, which the kernel logs at arrival as a problem naming the session (the CLI still starts, in its scope), and which `launch_failure_text` drops from a failed launch's card. The pre-flight scope carries the properties, so a systemd that rejects them (`OOMPolicy=` on scopes needs 253) fails there; the wrapper retries bare, then once more with them, and only a second failure drops them, quoting the failure that decided.

A value that passes the syntax can still be refused by the machine (an old systemd; an adjustment below the inherited floor; a read-only `/proc`), and the wrapper would then report it on every launch while the boot log called it in force. So with the scopes on, the kernel runs the wrapper's own steps once at its start: a probe scope with the properties, a memory-controller check inside it (systemd accepts `MemoryMax=` from a user manager without the controller and applies nothing; the only sign is no `memory.max` in the cgroup, and the verdict is a marker the probe command prints, so a scope that never started reads as unsettled, not as "no controller"), and the adjustment write in a throwaway child, in two steps so the exit status says whether the file would not open or the write was refused. A refusal lands once, as a problem line, and goes down empty. A check that does not answer settles nothing: a plain line says so, and the boot line lists that value as "set but not settled", never as "in force". Boot cost with limits set: two to four extra subprocesses at backend construction, each bounded at 10 s.

## Deliberate choices

- The size rule is narrower than systemd's: digits with one optional K/M/G/T suffix, or `infinity`. `50%`, `1.5G`, `16 G`, `16P`, `1G 512M` and `16g` are refused with the rule in the line. One shared corpus pins that the kernel's regex and the wrapper's shell function agree on every value.
- The adjustment refuses a leading zero: Linux parses `oom_score_adj` base 0, so `0400` would be octal.
- `OOMScoreAdjust=` on the manager unit was tried and rejected: an unprivileged user manager cannot set a unit below its own `oom_score_adj` (a unit asked for -500 came up at 100), and sessions inherit the kernel's value anyway, so that lever moves both tiers together. The docs record this so it is not retried.
- Two coverage gaps, documented: work handed to the manager's default tmux server runs in that server's scope, and a `systemd-run --scope` job a session starts is a sibling scope outside the limits (a plain `tmux -L` server is inside).
- Reporting is log and problem lines only; this PR keeps no runtime tally of `ignored:` lines. A follow-up that adds a health endpoint can carry the values, the rejected names and a count from the backend attributes this PR introduces (`cli_scope_limits`, `cli_scope_rejected`, `cli_scope_memory_delegated`, `cli_scope_unsettled`).

## macOS

Unchanged by construction. `cli_scope_supported` is False off Linux before any PATH lookup or probe, so `SdkBackend.__init__` calls `cli_scope_limits(scope_on=False, run=subprocess.run)`, whose `in_force and scope_on and run is not None` guard runs no probe; with a variable set it logs one plain "apply to nothing" line, with none set it logs nothing, and `_options` never reaches the overlay. Linux CI proves that path with injected platforms (`Supported` pins `platform='darwin'`; `LimitRules` covers `scope_on=False` with good and bad values). The wrapper's `size_ok` and `adj_ok` use only POSIX constructs (`case`, `${#1}`, `${1#-}`), so the shared-corpus test runs them under `sh` on macOS too; the `/proc`-dependent bats cases skip where `/proc/self/oom_score_adj` is absent. The docs subsection ends by saying none of it applies on the launchd path.

## Tests

- `tests/test_cli_scope.py` (+48; 30 to 78): `LimitRules` (the shared corpus, the rule lines with and without the scopes), `LimitsOnTheBackend` (the once-per-backend read; the `_options` overlay: vetted values down as themselves, refused ones down empty, unset ones not sent; the `ignored:` line logged at arrival), `LimitsSettledAtBoot`, and `SettleTable`, which enumerates every cell of the boot probe from its axes (198 subtests: argvs run, values handed down, rejected, the controller verdict, unsettled, every log line, the exact boot line).
- `tests/romp-cli-scope.bats` (+16; 19 to 35): properties on both scopes with `OOMPolicy=continue`, none without a limit, the size rule, the retry chain and the deciding failure quoted, the bounded pre-flight carrying the properties, the adjustment on every path including the fallback, and `apply_adj`'s two-step wording. A fake `systemd-run` on PATH; the `romp-session-*` unit count is the same before and after the suite (15 and 15 here).
- `tests/test_sdk_launch_error.py` (+1): the `ignored:` line dropped from the launch card's tail.
- `tests/conftest.py` floors the four variables to unset the way it floors `ROMP_CLI_SCOPE` (a tool shell under a self-hosted install inherits them).

Red-first against the base with only the test files checked out: `tests/test_cli_scope.py` does not collect (module-level references to the new constants), so there is no per-test count to give; `tests/test_sdk_launch_error.py` 1 failed / 34 passed; `bats tests/romp-cli-scope.bats` 16 not ok / 19 ok (all 16 new cases fail, the `/proc` ones included). Green: full pytest 7641 passed / 50 skipped; `bats tests/*.bats` 445 ok; `node --test tests/manager-*.test.js` 63/63; `shellcheck bin/romp-cli-scope` clean (the unquoted `$props` uses are word-split on purpose; an `SC2086` directive says so).

Verified on systemd 255 in earlier work, cited rather than re-run here: the `OOMPolicy=stop` default on scopes, the swap behaviour under `MemoryMax` alone, the `OOMScoreAdjust=` floor, and that a scope started with the wrapper's properties gets `memory.max`, `memory.high` and `memory.swap.max` applied with `memory.oom.group=0`.

Tier: feature


ksarma:
Closing as landed: this PR's three commits merged upstream inside #998 (the API-health PR was stacked on this branch; merge dd85c49d carries them as 30be5840, 34e71c7c and ac0f1808, plus the review commits that extend the scope surfaces with the cliScope block on GET /api-health). A rebase onto main leaves nothing to merge here.

PR #1010, opened 2026-09-07
Session tabs carry an emoji: set from the tab menu, by the session itself, or from the CLI

ksarma:
Depends on #1004 (the PR-links PR), which must merge first; GitHub shows its commits in this diff until then.

Stacks on the PR-links PR (branch `prlinks-offer`): review and merge that first; this PR's own diff is fd83e5876..f3f916ba1.

## What it does

A session's tab can carry one emoji before its name, so sessions tell apart at a glance by role or state: a moon on the overnight run, a flag on the release manager. The session itself can change it, so a long run can raise a label when it goes unattended and lower it when it needs someone.

Three doors, one validator, one store:

- **The tab menu.** Right-click a tab and choose **Emoji…**. The dialog sends `setSessionEmoji {id, emoji}` over the WebSocket. The kernel answers `emojiSet {id, emoji}` once the store has it (the strip changes on that confirm, as a rename changes on `renamed`), or a `warn` with the reason. The dialog stays open for the verdict and shows a refusal under the input with the typed value kept.
- **The session.** A postal tool `set_emoji(emoji)` beside `set_working`; `''` clears.
- **The shell.** `romp emoji <session> <emoji>`, `romp emoji <session> --clear`, and `romp emoji <session>` to read. The tool and the command go through `POST /emoji {target, emoji}`: `""` clears, a body without the key is a 400, never a clear, and so is a non-string value. `GET /emoji?target=` is the read half. A dormant id is read from the names registry with the kernel stopped.

A sid an attached host owns forwards over its tunnel like `/working` does, for the set and the read. An older remote kernel's 404 is reported as version skew (`that host's kernel (<host>) predates tab emoji`) rather than as a tunnel fault.

## Try it

```
romp emoji web 🌙          # romp emoji: "web" now shows 🌙
romp emoji web             # 🌙
romp emoji web ab          # romp emoji: refused — not an emoji: "ab"   (exit 1)
romp emoji web --clear     # romp emoji: "web" shows no emoji now
```

## The store

The emoji is the names record's fifth field (`name\tcwd\tbg\tfg\temoji`), written only while set, so a four-field record stays byte-identical. Every writer carries it: the kernel's name, color and palette writers (through `_names_line`), the SDK and Codex backends' `write_name` (the one hunk outside the feature's own files: a caller that passes no emoji carries whatever the record already holds), and the shell's record, rename-hook, title-freeing and resume paths. A colorless record is colored before it gets a fifth field. bin/romp's readers split on single tabs (`_romp_split_record`), because bash's IFS-tab read folds an empty run of fields and would read the emoji as the color.

## Changes to existing behavior

1. The kernel's names-registry writers (`_set_name`, `_set_session_color`, `_set_palette`, and the new `_set_session_emoji`) share one RLock and no longer publish over a record that reads with no name. Every writer puts the name first, so an empty first field is another writer's window or a damaged file, and overwriting it erased the session's name and cwd. The writer re-reads once after 50 ms, then leaves the record as it is and reports the sid to the kernel log and the dashboard's error center. An install with damaged records will see those reports where it saw silent overwrites before.
2. bin/romp's names-record writer is atomic (a temp file in the same directory, then `mv -f`). Its truncate-then-write opened the window in point 1.
3. `_remote_forward` keeps its contract (parsed JSON on a 200, else None) and now wraps `_remote_forward_status`, which returns `(status, json)` and takes `method="GET"`. One side effect changed: a 200 whose body is not JSON used to raise inside the try, demand a redial and return None. It is an answer, so it now returns `(200, None)` with no redial.

## Validation

`_emoji_check` accepts exactly one emoji per UTS #51 as far as the property tables reach: a base with an optional U+FE0F, a skin tone only on an Emoji_Modifier_Base and directly after it, at most four ZWJ parts, a keycap, a two-letter flag, tags only on the black flag. It refuses anything textual, a lone modifier or joiner, a non-string, an unpaired surrogate and a stray invisible character, each with a one-line reason naming the code point, and it never raises. The tables are Unicode 16.0's emoji-data.txt, regenerated by hand per Unicode release, and they decide what is invisible: Python 3.12 and 3.13 ship an older Unicode database that calls every 16.0 emoji unassigned. RGI membership is not checked, so a well-formed ZWJ chain of up to four emoji passes even when no font joins it; the reference says so. Values are capped at 48 bytes.

Three review passes on this change before offering it fixed the resume-path field drop, the colorless-record misread, a 500 on an unpaired surrogate, a null read as a clear, the interpreter-database misfire on Python 3.12/3.13, the naming of a stray after a flag or before a keycap mark, the atomic shell write, and the dialog staying open on a late confirm.

## Rendering

A span before the label on real, placeholder and federated tabs, exposed to assistive technology: session names are not unique on a host, and two tabs both named `web` may differ only by the moon one wears. The menu row shows the current emoji as its icon (hidden there; the row text is the label). This PR ships a text input; a picker dialog is a natural follow-up.

## Riders

- Two bats suites (tmux-status-hook, romp-headless) and two cases in romp.bats rewrote a mid-test bare `!` negation, which checks nothing under bats' errexit, as `run` + `[ "$status" -ne 0 ]`; `tests/test_bats_bare_negation.py` guards the form. The new emoji cases needed the checked form.
- The first commit routes bin/romp's resume-time `picker-check` through `ROMP_POSTAL_BIN` (as `mail` and `refresh` already are), and romp.bats stands in a no-op there: the resume case in this PR detaches that call and raced the real service without it. The commit is independent of the feature and can land on its own.

## Platform

The shell changes use parameter expansion, `$'\t'`, indexed arrays, `IFS= read -r`, `[[ =~ ]]` and `mv -f`: no `readarray -d`, no `${var,,}`, no associative arrays, no `local -n`, so bash 3.2 (macOS) runs them by construction. PR CI runs bats on Linux only; a macOS `workflow_dispatch` run is available on request.

## Tests

- `tests/test_session_emoji.py` (48): the validator, the store, the backend writers, the three tab_meta frames, POST and GET `/emoji` over the real Handler, the WS op.
- `tests/test_postal_emoji.py` (8), `tests/test_bats_bare_negation.py` (6); pins in `test_kernel_send.py` and `test_kernel_tabs_first.py` carry the field.
- `tests/romp.bats` (+12): set, clear, read, refusal, dormant id, forwarded read, kernel-stopped read, the names-map and resume shapes, the picker-check seam.
- `ui/webview/tab-meta.test.ts` (15), `ui/webview/federation-tab-emoji.test.ts` (2), `vscode-extension/src/tab-emoji.test.ts` (10 source pins: the strip, the menu row, the dialog, the warn router, the pipe, the guide); five existing pins follow the new spelling.

Synthetic ids and names throughout. Full runs on the branch: pytest 7709 passed, 46 skipped; bats 441/441; extension typecheck clean, 2911 tests, build clean.

Tier: feature


henrypinkard:
Closing this one for now, on the owner's call after a discussion with the author.

The outcome of that discussion: a per-tab emoji is really a user-configurable status that lives in the tab title, and the same slot already carries the working/awaiting dots and the context bars. Rather than add the emoji as its own feature, the plan is to abstract those indicators together with a configurable status into one system. The work here (the tab menu path, the session-set path and the CLI path) is a good basis for that and can come back as part of it; the review fixes pushed to this branch stay on it.

Separately noted from the same discussion: session names should be freely changeable, with a stable id underneath, so that renaming never breaks references between sessions.

PR #1013, opened 2026-09-07
T244: a ✕ on a provisional tab's sending bubble forgets the send everywhere it lives; every missed cancel leaves evidence

henrypinkard:
## Summary
The user pressed ✕ on a message still showing "sending…"; the text returned to the composer but the sending bubble stayed.

**Reproduced by reading, after the harness cleared the obvious path.** A served repro (hermetic kernel, kernel frozen with SIGSTOP so nothing acknowledges the send, ✕ on the bare bubble, then thaw) showed the parked-send path clean: bubble gone, `cancelResult ok:true`, nothing re-injected. The long-lived "sending…" is a **provisional tab's**: a send into a session still being created goes into `provisionalQueue` and wears the bare group for the whole creation window (up to 90 s). The ✕ dropped only the optimistic entry and posted a cancel for a sid the kernel never had; adoption then re-sent every queued text, the cut one included, and re-registered its bubble under the real tab, while the ✕'s restore had already put the same text in the composer.

**Fix.** A ✕ on a provisional-tab bubble also forgets the text from `provisionalQueue` (one press, one send) and posts no kernel cancel; the restore stays. Adoption re-sends only what remains.

**Evidence for next time.** A client-diag row for every `cancelResult ok:false` (sid, body length, whether a restore was in play, the refusal text; never the body) and for every provisional-stage ✕; a kernel log line for a body-only cancel that finds nothing in either queue.

## Tests (red first)
`ui/webview/provisional-cancel.test.ts`: the qx handler's provisional branch; the executed `forgetProvisionalSend` helper; the diag breadcrumbs; the kernel log pin. `tests/test_kernel_send_park.py`: the log line.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -11143,6 +11143,10 @@ def _drive(msg, client):
             err2 = _cancel_backend_queued(be, sid, -1, md)
             if err2 is None:
                 err = None
+        if err:
+            # evidence for the next report (T244): a ✕ whose cancel found nothing in either queue — the send
+            # had already gone through, or never reached this kernel. sid only: the body is the user's text
+            sys.stderr.write("queued-cancel miss: %s (body-only)\n" % sid)
         client["send"](json.dumps({"type": "cancelResult", "ok": not err, "id": sid,
                                    "md": md, "text": err or ""}))
         _push_soon()
--- a/tests/test_kernel_send_park.py
+++ b/tests/test_kernel_send_park.py
@@ -428,6 +428,13 @@ class QueuedBubble(unittest.TestCase):
         self.assertIn('_cancel_backend_queued(be, sid, int(msg["idx"]), str(msg.get("md") or ""))', src,
                       "the backend-queue cancel goes through the drift guard now")
 
+    def test_a_body_only_cancel_that_finds_nothing_is_logged(self):
+        # T244 (the user 2026-09-07): a ✕ on a "sending…" bubble left no evidence of which way the cancel went.
+        # The body-only arm now logs a miss (sid only — the body is the user's text) so the next report carries it.
+        src = open(os.path.join(BIN, "romp-kernel")).read()
+        arm = src.split('elif t == "cancelQueued" and msg.get("md"):')[1].split("\n    elif ")[0]
+        self.assertIn('sys.stderr.write("queued-cancel miss: %s (body-only)\\n" % sid)', arm)
+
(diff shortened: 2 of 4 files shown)


ksarma:
Reviewed after the merge: nothing needed folding, and the fix is right as shipped. Two wording follow-ups and one nit are listed below for a later pass.

Reviewed inline: `forgetProvisionalSend` removes the first exact match from `provisionalQueue` with `indexOf` and `splice`, so one press forgets one queued copy. The existing `pendingSent` removal in the `qx` handler uses the same first-match rule, so display and queue stay consistent when two identical texts are queued. On a provisional tab the handler posts no `cancelQueued`, keeps the `restoreToComposer` call, and keeps no `pendingCancelRestores` stash. The text lands in the composer exactly once, and adoption re-sends only what `dropProvisional` slices from the trimmed queue. The cancel-miss row carries `sid`, `mdLen` and `hadRestore`; the cancel-provisional row carries `mdLen` and `queuedLeft`. The kernel writes that data verbatim to `client-diag.jsonl`, so no body or substring reaches it; the second commit removed the `text` field the first commit had. The kernel line writes `sid` only to stderr, a kernel log, not text injected into any session. The glyph stays on the `document.body` delegate, acknowledged by `flash` plus the synchronous bubble removal. Privacy and vocabulary scans of the 91 added lines were clean.

Follow-ups:
- `ui/webview/render.ts:14399` (follow-up): the comment and the first commit message say a pre-fix provisional cancel would have produced the too-late toast. It never reached the cancel arm. `_drive` refused it at the `_kernel_knows` guard, so the client showed a not-delivered modal plus a bell entry and an `undelivered.jsonl` row. This fix removes that outcome too. Correct the comment and note the removed modal.
- PR body, Evidence paragraph (follow-up): it still lists the refusal text among the cancel-miss row's fields. The second commit removed that field. Drop the phrase.
- `ui/webview/render.ts:14403` (nit): `forgetProvisionalSend`'s boolean result is discarded, so the cancel-provisional row cannot say whether the forget hit. Recording it would make the row self-evidencing.

Red-first verified. With the PR's test files on origin/main, `tests/test_kernel_send_park.py` failed 1 of 45 (the queued-cancel miss line absent) and `provisional-cancel.test.ts` failed 3 of 3. At head, 45 of 45 and 3 of 3 pass, with no sleeps or timers in either. CI at head: all 7 checks green.

Review run by the maintainer's automated review session; verified against head f65305f4.

Issue #1012, opened 2026-09-07
Proposal: a command key source, with the 1Password reference (#962) as its built-in default


ksarma:
## Summary

I propose that romp's runtime key source be a configured command that prints `NAME=VALUE` lines, with the 1Password reference merged in #962 kept as that command's built-in default. The kernel runs the command, holds the set it prints in memory, and injects the values into each child it starts: session CLIs, judge calls, and the model catalog fetch. `ROMP_API_KEY_REF=op://vault/item/field` keeps working unchanged, because the kernel treats a reference as a command it already knows how to run. Any other secrets manager, a vault CLI, a hardware token, or a plain script is configured with one line and handled the same way. This is the contract Claude Code's own `apiKeyHelper` offers at the CLI level, applied at the kernel level. The design has been on my fork's main since 2026-09-06 (ksarma/romp#229), where it sits beside the file source. This issue reframes it as a generalization of #962, so that adopting it changes nothing for an installation that configured a reference. Below: what #962 does, what a command source does, where the two differ and why I chose what I chose, what the code change would be, and the decisions that are yours.

## What #962 does today

#962 (which carried #932 onto main) adds one runtime key source, a 1Password secret reference, and optimizes for one line of configuration, a known-good secrets manager, and no API key on disk (the op bootstrap token in the service env is the one credential it does keep there). The mechanics, as they stand on `main`:

- **One source kind per manager.** `kernel/keysource.py` `KeySource` has the kinds `op`, `file`, `environment`, `none` and `error` (`KeySource.validate`). A `ROMP_API_KEY_REF=op://vault/item/[section/]field` line in `service.env`, or in a foreground manager's environment, selects the `op` kind (`parse_source`, `select_source`).
- **Resolution per operation, no cache.** `KeySource.resolve` runs `op read --no-newline <ref>` with stdin closed, a 15 s deadline (`OP_TIMEOUT`) and a whitelisted environment (`op_subprocess_env`, `OP_ENV_PASSTHROUGH`), at each session launch or reconnect (`kernel/sdk_backend.py` `_options` through `_work_key_and_source`), each key-billed judge call (`kernel/judge.py` `_judge_env` through `_resolve_work_key_gated`) and each catalog refresh (`kernel/kernel.py` `_models_api_credential`). The module docstring states the rule: the value is handed to that one operation and never written to disk or cached for a later one.
[...]

PR #1014, opened 2026-09-07
No romp key source means Claude Code's own credential, for every billing pick

lczh:
## What

After #932/#962, a session whose Billing pick is the API key crashes at launch when romp holds no key source: `_options` raises `KeySourceError("API key billing selected but no API key source is configured")`, the thread dies, and every key-billed judge call fails for the rest of its pass. Before #932 the same launch started the CLI without an injected key and Claude Code's own credential resolution applied, on many boxes its `apiKeyHelper`. That is how an install that had always billed a key through the helper looked when it took the update: `service.env` with `ROMP_EXPECTED_AUTH=key` and no key or reference line. The board was down until the operator wrote a reference and token into the file and restarted the service by hand.

This restores the rule from before #932 for that one case: **when no key source is configured, romp injects nothing and Claude Code decides**, whatever the session's pick. A configured source that fails to resolve still refuses the launch as #932 intended.

- **`kernel/sdk_backend.py` `_options`**: an unconfigured source means `launch_keyed` is false for every pick; the CLI launches with no `ANTHROPIC_API_KEY` and the login tokens restored, exactly like an unpicked launch today. When the pick was the API key, one problem row per process says the sessions are running on Claude Code's own credential and names the two lines that would put romp in charge. The existing per-init `apiKeySource` row behaves as before #932.
- **`kernel/judge.py` `_judge_env`**: the same for key-billed judge calls, the pre-#932 keyless child, said once per process on stderr; a configured source keeps the gated retrieval path.
- **`docs/reference.md`**: one paragraph. With no source, Claude Code's `apiKeyHelper` (which can fetch from any secrets manager) or its login applies; a reference or key line is for boxes where romp should manage the key. romp reads no Claude Code settings; it simply stays out of the way.

Deliberately reversed from #932: "explicit API-key billing with no key now fails instead of switching to login." The maintainer's direction (2026-09-07) is that not providing romp a key should mean Claude Code's default, and that romp should stay agnostic to any particular key system.

Two small companions of the rule: a remembered API-key default is not seeded onto a new session on a box with no source (the picker already shows login there; a seeded pick would ring the per-init row for every new session), and the judge's keyless child gets the login tokens a login call gets, as it inherited them before #932.

Known follow-up, not changed here: `romp keyswap --cycle` on a box with no source still answers the old "no API key source is configured" error for key-picked sessions instead of moving them onto the new rule; it did so before this change too.

Not in this PR: hold/self-heal machinery, a preflight, a setup command, or changes to how `op read` runs (those live on a separate branch and are discussed against #1012 instead).

## Tier

- [ ] tests-only
- [x] fix — `tests/test_runtime_key_auth.py`'s unconfigured-source launch tests fail before this change (the launch raises).
- [ ] feature
- [ ] major-feature

## Tests

`tests/test_runtime_key_auth.py` 27 (new `UnkeyedPickLaunch`: the incident's box shape, an explicit pick launches un-injected with a login launch's tokens, one row across launches, the init check quiet on a keyed landing and ringing on a login one, a configured source unchanged, a removed reference still refused, the remembered default not seeded), `tests/test_judge_auth_billing.py` 45 (new `UnkeyedJudgeCalls`), `tests/test_session_auth.py` 40, `tests/test_keyswap.py` 91 (three refusal tests flipped to the rule; marker and error cases still refuse), `tests/test_runtime_keysource.py` 74, `tests/test_sdk_backend.py` 230, `tests/test_expected_auth.py` 32, `tests/test_sdk_launch_error.py` 34, `tests/test_op_credential_boundary.py` 5, `tests/test_supervised_floor.py` 3. Run one module at a time on a shared machine; the full suites are CI's. A read-only adversarial review found five issues, all folded (the judge child's login tokens, the seeded default, the docs criterion, a stale docstring) or noted above (`--cycle`).

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -368,6 +368,16 @@ selected configuration. An empty or invalid reference is an error, not a
 request to use the legacy key. Remove competing plaintext assignments when
 migrating; `romp keyswap` does this automatically when selecting a profile.
 
+With no key source selected — the env file has no `ROMP_API_KEY_REF=` line
+and no `ANTHROPIC_API_KEY=` line with a value, and no reference was selected
+earlier (a removed reference stays an error until another source is
+configured; see below) — Romp injects nothing, and Claude Code's own
+credential resolution applies: its `apiKeyHelper`, which can fetch a key from
+any secrets manager, or its login. A session's API-key Billing pick then only
+takes effect once a source exists; until one does, the kernel says so once in
+its log. A reference or key line in `service.env` is for boxes where Romp
+should manage the key — swap it, fingerprint it, cycle sessions onto it.
+
 Romp runs [`op read --no-newline`](https://www.1password.dev/cli/reference/commands/read)
 for each Claude SDK session launch or reconnect, each API-key-billed judge
 call, and each direct model-catalog refresh. A paginated catalog refresh uses
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -490,9 +490,10 @@ key inherited when the kernel started. After removing a selected provider,
 API-key operations fail until a valid source is selected; choose **Login**
 explicitly to use that mode. Removing a static `ANTHROPIC_API_KEY=` line
 while the kernel runs is different: the file stays authoritative and
-sessions without an explicit Billing pick launch on the login, and the
-kernel log says so once, with the removed key's fingerprint and the file's
-path. Removing a source does not revoke a credential already held by a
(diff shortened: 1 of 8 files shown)


lczh:
Tier: `fix` — a fork author cannot add labels here, so asking a maintainer to apply it. Before this change, `tests/test_runtime_key_auth.py::UnkeyedPickLaunch` fails: a session picked for API-key billing on a box with no key source raises at launch instead of starting on Claude Code's own credential.

PR #1015, opened 2026-09-07
RFC: a credential command as the runtime key source, with the 1Password reference as its built-in default

ksarma:
Discussion: #1012.

Depends on #998 (which depends on #1008 and #1002). The branch is stacked on #998's head; the diff against it is the 17 commits listed below (rebased onto #998's review head 01e9247b on 2026-09-07; one parity commit added).

## Summary

This PR makes romp's runtime key source a configured command that prints `NAME=VALUE` lines, and keeps the 1Password reference from #962 as that command's built-in default. The kernel runs the command, holds the set it prints in memory, and injects the values into each child it starts: session CLIs, judge calls, and the model catalog fetch. `ROMP_API_KEY_REF=op://vault/item/field` keeps working unchanged: an installation that configured a reference and sets no command changes nothing. Any other secrets manager, a vault CLI, a hardware token, or a plain script is one `ROMP_CREDENTIAL_COMMAND=` line and is handled the same way. The set never enters `os.environ`; it reaches children through their environment dicts only.

The design is #1012's; the choices below answer the questions raised there, and the last section restates them. The code was re-derived against `keysource.select_source` on the #998 base: the reference keeps its per-operation `op read`, its fail-closed rule and its whitelisted subprocess environment, and the command kind sits beside it in the same seam.

## What #962 does today

One runtime kind, `op`, selected by a `ROMP_API_KEY_REF` line in `service.env` or in a foreground manager's environment (`keysource.parse_source`, `select_source`). `KeySource.resolve` runs `op read --no-newline` at each launch, key-billed judge call and catalog refresh, with no cache and no fallback; a failure raises `KeySourceError` and the launch is refused. The reference is authoritative once selected: removing the line is an error until another source is configured, remembered across restarts in `service.env.source`. The op bootstrap token lives in `service.env`; the kernel claims the `OP_*` names out of its environment, hands them to the `op read` subprocess alone, and scrubs them and the startup key from the tmux server. `romp keyswap <name>` rewrites the live file from a `service.env.<name>` profile; `/keycycle` answers `keyFp` (empty for op), `sourceFp` and per-session rows. `romp-service status` says nothing about the key source.

## What this adds

- **A `command` kind in `kernel/keysource.py`**, selected by `ROMP_CREDENTIAL_COMMAND` through the same two doors as the reference: a line in `service.env` for any manager, the environment for a foreground manager; a supervised manager reads the file only. Precedence in the file is command, then reference, then key. The marker `service.env.source` holds `op` or `command`, so removing either runtime line is an error until a source is configured again. `write_source` accepts the command kind, so a `service.env.<name>` profile may carry a command line and `romp keyswap <name>` selects it from the reference or a key line. `resolve` for the command kind goes through a module hook (`COMMAND_RESOLVER`) the kernel wires at import, so every existing caller of `resolve()` works unchanged and `keysource.py` never imports the runtime.
- **The runtime, `kernel/envsource.py`** (stdlib only). The command runs as `/bin/sh -c <command> sh <selector>` with stdin closed, in its own process group, under `ROMP_CREDENTIAL_TIMEOUT_S` (default 15 s, at most 300). The output is parsed with the env file's own line rule; `ROMP_*` names and the four CLI authentication and endpoint names (`ANTHROPIC_AUTH_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`, `ANTHROPIC_BASE_URL`, `ANTHROPIC_CUSTOM_HEADERS`) are dropped with one problem line. The set lives in one private dict and leaves through `injection`, `take` and `resolve_key`, whose callers merge it into a child's environment. The cache is keyed on events, never a timer: the source fingerprint, an invalidation, an authentication failure (once per credential), and the selector file's stat identity.
- **Keep the last good set.** A failed run keeps the previous set, marks the record `stale`, says one problem line per failure kind, and re-runs at the next caller; concurrent callers coalesce on one run. A launch is never refused for a failed run. A command that never succeeded injects nothing, and the line says the `apiKeyHelper` or the login bills.
- **Where the values go.** `_options` takes the set once per connect: `ANTHROPIC_API_KEY` is injected only when the billing decision is the key (an explicit key pick, or no pick with a key in the set); the other names are merged under romp's own entries; the reserved names are filtered on the launch's actual billing. Judge calls get the set minus the key, re-added for a key-billed call through the existing `_resolve_work_key_gated`; a Codex-engine call merges nothing. The catalog fetch reads the set's `ANTHROPIC_LP_API_KEY` ahead of the process environment. The judges' default billing (`_WORK_KEY_CONFIGURED_FN`) answers from the set under the command kind, so a keyless set bills the login beside the session instead of refusing.
- **Rotation by name.** `ROMP_CREDENTIAL_SELECTOR_FILE` holds one token, passed as `$1`; `ROMP_CREDENTIAL_NAMES` declares the names an operator may select and decides whether a token is shown by name or by length. `romp keyswap <name>` under a command checks the name, writes the token (atomic, 0600, through a symlink), re-runs the command in the operator's shell, confirms the fingerprint moved, undoes the write when nothing moved, and asks the kernel to re-run.
- **Fingerprint-converging reconnects.** Each connect is stamped with the fingerprint of the credential the session launched on (the injected key, or the `apiKeyHelper` output hashed inside `envsource`, or nothing) and the role set. `cycle_key` under the command kind reconnects only when a stamp differs from what a launch would get now, so a repeated `--cycle-all` converges and helper-billed sessions move after a rotation. The op and file kinds keep #962's `cycle_key` body.
- **A boot verdict** (`key_source_verdict`, pure) computed once at backend construction from the environment, the env file's text, the systemd unit, its drop-ins, the launchd plist and the `apiKeyHelper` command. It logs the source in force and, as problems, a failed first run, dropped names, an `ANTHROPIC_API_KEY` line a runtime source ignores, `ExecStart` through a shell, a declaration that contradicts the set; and, as information, other credential-shaped names (`*_API_KEY`, `*_TOKEN`) found in those places, since under a command they reach every session. Under the reference the `op` names are exempt. Under the file kind the boot log is the base's.
- **`/keycycle`** keeps `ok`, `keyFp`, `sourceFp` and the rows, gains `"refresh": true` (re-run first), and answers `keySource`, `keyKind`, `keyErr`, `setFp`, `selector`, `launched` and a per-row `from`. `GET /api-health` carries the verdict as `keySource`.
- **`romp keyswap`** gains the command arm (a bare report runs the command in the operator's shell and compares fingerprints with the kernel's, saying `MISMATCH` with the cause), `--refresh`, and a mode-mismatch report when the kernel's kind differs from this shell's. **`romp-service status`** prints `key source: file | reference | command (selector ...) | none (...)`, the `ExecStart` shape, and the credential-shaped names the unit, a drop-in, the plist or `service.env` carries, names only.
- **Docs**: `docs/reference.md` "A credential command", "Rotation by name" and "What `romp-service status` shows"; the keyswap and `/keycycle` sections; `docs/judges.md`; `kernel/README.md`.

## The reference as the default command

The equivalence is in the seam. Both runtime kinds are selected by `select_source`, retire the startup key in `work_api_key_source`, are reserved-name sources, carry a `sourceFp`, feed the boot verdict and `romp-service status`, and are dispatched on `source.kind` by every reader. The op kind keeps #962's direct `op read` (never through `/bin/sh`), fail-closed, no cache, whitelisted environment; the command kind gets the cache and keep-last-good. `claim_op_env`, `withoutOpCredentials` and the tmux scrub stay reference-only: they are the reference's bootstrap mechanism. Under a command nothing is claimed or scrubbed, the command runs with the kernel's environment, and `op_consumer()` is false whenever a command is selected.

## Differences from #962 and why

1. **No bootstrap credential is managed for a command.** A command authenticates itself. The verdict names credential-shaped lines it finds as information under a command (they reach every session) and flags a leftover `ANTHROPIC_API_KEY` line as a copy to remove and rotate. Under the reference the `OP_*` names stay claimed and exempt.
2. **Keep the last good set for the command kind; fail closed for the reference.** The failures a cached source sees are store outages, and a refused launch for a brief outage costs more than one possible refused turn on a stale key, which the once-per-credential invalidation then replaces. The reference keeps #962's rule because a per-operation read has no last good set to stand on.
3. **One run per invalidating event, never a timer.** A judging pass of hundreds of key-billed calls is one run.
4. **Rotation by name** through a selector file that never holds a key, so an `apiKeyHelper` reading the same file moves with it.
5. **Reconnects converge on launch fingerprints**, so helper-billed sessions are cycled too.
6. **The verdict reads the whole service definition**, not only the env file, and `romp-service status` names what it finds.
7. **Selection is live per call**, #962's rule. #1012 described a mode pinned at kernel start; the pin is dropped, because the cache identity is the source fingerprint and a swap between a command and the reference re-runs or empties by construction.

## What changes in the code

`git log 01e9247b..HEAD`, oldest first:

- `e9f6dabb` Key source: a command kind beside the reference, selected the same way
- `99e1f7cd` Command source: the runner, parser, selector and event-keyed cache
- `42e4c40c` Sessions, judges and the catalog take the command's set; the key rides the billing decision
- `81b6fd1d` Boot verdict: the key source in force, and credential-shaped names in the service definition
- `f676241d` Cycle converges by launch fingerprint; /keycycle gains refresh and the key-source fields
- `ee8c1329` romp keyswap: the command arm, --refresh, and a mode mismatch
- `94f9ece1` romp-service status names the key source; the launcher strips one layer of quotes
- `ba4912e6` Docs: the credential command beside the reference
- `62cc1f3d` Docs: a leftover reference line still shapes the manager's tmux server
- `809dd3f4` Review fixes: test_command_source loads bin/romp-kernel at import, not in setUpClass
- `3042b229` Review fixes: romp-service status reads the removed-source marker
- `e2204fb0` Review fixes: the alias hint says the installer reads ROMP_SERVICE_ENV_FILE only
- `3929d3ea` Review fixes: the judges' default billing follows the set's key under the command kind
- `99832aad` Review fixes: the courier link tests remove their override journal in tearDown
- `9790d415` Review fixes: JudgesDefaultBilling takes the module's kernel load
- `71b3ec79` Review consolidation: the docs match the readers, and three parity leftovers
- `34bad116` Review parity: the env-file remedy covers a command line, and api-health status reads keep the key-source block

Files: `kernel/keysource.py`, `kernel/envsource.py` (new), `kernel/sdk_backend.py`, `kernel/judge.py`, `kernel/kernel.py`, `cli/keyswap.py`, `bin/romp`, `bin/romp-service`, `bin/romp-node-launch`, `docs/reference.md`, `docs/judges.md`, `kernel/README.md`, and 15 test files. 27 files, +10055/-210; 6409 of the added lines are tests.

## Compatibility

An installation with `ROMP_API_KEY_REF` in `service.env` needs no change. What it sees is additive: `/keycycle` carries the new fields; `romp-service status` prints `key source: reference`; a boot line may appear for a leftover `ANTHROPIC_API_KEY` line or a credential-shaped name in the unit, a drop-in, the plist or `service.env`; `work_key_fp()` returns empty under the reference instead of resolving `op` (the route already answered `keyFp` empty for op). An installation with an `ANTHROPIC_API_KEY=` line keeps the file kind, its write path and its boot log. Two behaviour changes reach every installation: `bin/romp-node-launch` strips one layer of matching quotes from every `service.env` value on macOS, not only the command line (parity with systemd's `EnvironmentFile=` and the kernel's own read); and `bin/romp-manager` still scrubs the tmux server whenever a `ROMP_API_KEY_REF` line exists, command line or not, which the docs state. One case needs a manager restart and is documented: a manager that started under the reference claimed `op`'s variables, and a swap to a command on that running manager leaves them claimed.

Opting into a command is one added line, plus `ROMP_CREDENTIAL_NAMES` and the selector file if rotation by name is wanted. `ROMP_CREDENTIAL_NAMES`, `ROMP_CREDENTIAL_SELECTOR_FILE` and `ROMP_CREDENTIAL_TIMEOUT_S` are tuning, not secrets, and are read from the manager's environment first, then `service.env`.

## What is not proposed

The file kind's write path (`keysource.write_source` for a key, `write_key`) and the `service.env.<name>` profiles are as #962 ships them; the only change in the file kind's write is that a competing `ROMP_CREDENTIAL_COMMAND` line is removed alongside a competing reference line. No refusal of `romp keyswap <name>` in file mode, and no documentation stance that `service.env` should hold no credential: the docs present the command as an option beside the reference and the file. No claim-and-whitelist for an operator's command (#1012 question 3): a command that needs `OP_*` finds them in the kernel's environment because nothing is claimed for it.

## Tests

266 pytest tests and 18 bats cases across 15 files; every credential command in the tests is a fixture script printing synthetic `NAME=VALUE` lines, and no test runs `op`. `tests/conftest.py` removes `ROMP_CREDENTIAL_COMMAND`, `ROMP_CREDENTIAL_NAMES` and `ROMP_CREDENTIAL_TIMEOUT_S` at import and per test, floors `ROMP_CREDENTIAL_SELECTOR_FILE` to a nonexistent path with a collection-time refusal, and points the unit, plist and Claude settings directories at empty temp dirs, so no test runs a real command or a real `apiKeyHelper`.

- `tests/test_envsource.py` (+97): runner, parser, selector, event-keyed cache and coalescing, source identity, helper fingerprint, a `NothingLeaks` class asserting `os.environ` is unchanged after every reader and no fixture value reaches a record.
- `tests/test_key_source_verdict.py` (+49): the pure verdict, line by line per kind.
- `tests/test_command_source.py` (+56): a real backend and the real routes under the command kind: `_options`, the judges' default billing, cycle convergence, `/keycycle`, `/api-health`.
- `tests/test_keyswap.py` (+45), `tests/romp-service.bats` (+13), `tests/romp.bats` (+4), `tests/romp-node-launch.bats` (+1): the CLI arms, the status lines, the launcher.
- `tests/test_judge_auth_billing.py` (+8), `tests/test_model_catalog.py` (+4), `tests/test_api_health.py` (+2), `tests/test_expected_auth.py` (+3), `tests/test_session_move_live.py` (+1), `tests/test_courier_link.py` (+1).

The base's keysource, op-boundary and keyswap tests pass unchanged. Red-first counts from the review rounds, each recorded in its commit: the judges' default-billing class failed 5 of 5 against the unported kernel and a probe showed `_judge_env` raising for a keyless set while the session ran; the status marker case failed at its first `none` assertion; the alias-hint test carried the old claim in its output; the import-time load fix was shown by two probe modules (2 failed, then 6 passed) and a serial run of the 214 modules through the deferral sweep (3836 passed) that had failed 7 tests in CI order; the consolidation's two keyswap cases and the envsource cache case failed against the unported code. Full suite on the final head: 8355 passed, 51 skipped, 0 failed; `tests/romp.bats` 104/104; `tests/romp-service.bats` 32/32.

## Questions for the maintainer

1. The command as the general mechanism, with the reference as its built-in default, so `ROMP_API_KEY_REF` is one spelling of a command source?
2. Failure semantics as built: keep-last-good for the command kind, fail-closed for the reference. Keep both, or one rule for both?
3. Bootstrap credentials: the reference keeps its claim-and-whitelist; a command claims nothing, and the verdict names credential-shaped lines as information. Acceptable, or should an operator's command be able to name variables it needs the same way?
4. Codex-engine judge calls merge nothing from the set (the branch's answer to #1012's question 4).
5. May the set carry several variables (`ANTHROPIC_LP_API_KEY` for the kernel's direct calls, other names for the sessions' tool shells), and is the dropped list right?
6. Naming: `ROMP_CREDENTIAL_COMMAND`, `ROMP_CREDENTIAL_NAMES`, `ROMP_CREDENTIAL_SELECTOR_FILE`, `ROMP_CREDENTIAL_TIMEOUT_S`, or a family closer to `ROMP_API_KEY_REF`?
7. May the verdict read the unit, its drop-ins, the plist and the kernel's own environment, and should `romp-service status` report the `ExecStart` shape?
8. Selection is live per call, #962's rule. Agreed?
9. Review found five follow-ups that this PR leaves as they are; each can be a later change, or fold in here on your word:
   - `SdkBackend.work_key_configured` (the dashboard's billing label, `default_auth`, and `effective_auth` with no key argument) still answers from the descriptor, so under a keyless command set the display says the key bills while launches bill the login or the helper. Aligning it needs a read-only accessor so a UI snapshot never runs the command.
   - The same accessor would let status reads (`/api-health`, a bare `romp keyswap`) serve the stale record while a command is failing instead of re-running it, and let `/keycycle` take the set once per request instead of once per row.
   - `judge._env_auth_ok` re-arms the once-per-credential invalidation on served login-billed calls too.
   - `keysource.runtime_reserved_names` with no `keyed` argument (the per-session env door) is stricter than the launch under a keyless set.
   - `bin/romp-service` and `bin/romp-node-launch` resolve the env-file path from `ROMP_SERVICE_ENV_FILE` only, while the kernel also reads the alias `ROMP_SERVICE_ENV`; the keyswap hint says so.

Tier: major-feature


lczh:
Thanks for turning this into a PR so quickly. Two thoughts from my side.

First, a heads-up that this overlaps with #1014, which I opened for the outage we hit today. That fix makes "no key source configured" mean romp injects nothing and Claude Code's own auth takes over, for every billing pick. I think this branch does the same thing for the command kind, but keeps the raise for the reference and file kinds, which is the path that actually took our board down when we updated. Whichever one lands second will probably need the other's rule carried over; #1014 is the smaller one, so maybe it goes first?

Second, more of an open question than an objection. I've been wondering whether romp should just be agnostic to keys altogether and let the underlying agent's own config predominate, i.e. `apiKeyHelper` for Claude Code, whatever the equivalent is for other agents. It seems like it might scale better as more agents get added, and it would mean romp never has to hold or fetch a key itself. This PR moves in the other direction (a second in-kernel source, a selector file, a verdict that reads the unit and the plist), so it might be worth settling that question first, before the numbered ones at the end. I'm not an expert here, so genuinely curious what you and the maintainer think.

ksarma:
Converting to draft while the direction question is settled on #1012 (whether romp holds a key source at all, and #1014 landing first). Nothing here needs review until then.

ksarma:
Closing in favour of the direction settled on #1012 and now merged as #1014: romp injects a key only where an operator configured one, and otherwise stays out of the way and lets Claude Code's own resolution apply. apiKeyHelper already covers the command-that-prints-a-credential case, so a second in-kernel source is not needed. The credential-shaped-names boot line and the romp-node-launch quote-strip from this branch follow as two small standalone fixes, and a doc paragraph on the apiKeyHelper route for secret-manager keys.

PR #1016, opened 2026-09-07
Returning to the dashboard's tab no longer redials every pane and replays the world

lczh:
Follows #934 (the chat tab-click lag fixes, merged 2026-09-07), which this change was developed on top of and rebased onto current main; the same render paths are involved.

## What

Switching back to the dashboard's browser tab after it sat in the background froze the page. Each pane iframe holds its own socket to the kernel, and the pane shim judged that socket by the age of its last received frame. A tab Chrome had frozen ran no script, so on return every pane found a "stale" socket, abandoned it, redialed, and the kernel re-sent the whole state to all four panes at once: about 15 MB on a seventeen-session board, each frame parsed and painted on the one main thread the same-origin panes share. A shorter absence replayed its queued frames instead, one full paint per frame.

- **Honest silence (kernel.py `_shim`).** The Page Lifecycle `resume` event, which Chromium fires before `visibilitychange` on a thaw, stamps the last-received clock on an OPEN socket, so a paused renderer is no longer read as a dead link. A socket the kernel really dropped closes in the same burst; the first close after a foreground redials **now** (a second one inside the window waits 250 ms so a kernel that is down is not hammered), instead of the blind 1.5 s. `federation.ts` stamps its remote sockets the same way. Browsers without `resume` keep today's behaviour.
- **One ordered queue per socket, sliced.** Frames are handed to the bundle through a FIFO flushed from a `MessageChannel` task in ~8 ms slices: a newer whole-state frame (`feed`/`bars`/`data`/`tabOrder`/`working`/`globalRetryPaused`) replaces its older queued twin, chained frames keep wire order, `applyDelta`/`needSlot` stay synchronous, and a click can land between slices. Measured: the first cut delivered 25 queued chat tails in one 796 ms task; sliced, the longest task is 96 ms.
- **Apply state always, paint when visible.** New pure `ui/webview/paint-gate.ts`. The feed and board panes apply every payload while hidden and paint once on `visibilitychange`/intersection, with cards that moved while away snapping (no FLIP replay: 788 held frames in 45 s, 31,044 in ten minutes, are gone). The timeline parks bars that outrun their skeleton, holds its draw while hidden, and its live tick stops instead of polling a layout every 2 s. The chat's active tab paints once per animation frame however many tails a burst lands.
- **An honest cue and a breadcrumb.** The pane's corner badge now waits for the first fresh frame (`romp:wsfresh`), not the socket opening, so a real resync over a tunnel never looks like a frozen dashboard. Each return files a `clientDiag` row naming the browser state it found (`page-load` with `wasDiscarded`, `return` with `decision keep|redial-closed|redial-stale`, `frozenMs`, `quietAtResumeMs`, and `return-fresh` with the wait), re-filed onto the redial when the kept socket turns out to have been dead, so the regime a user actually hits can be read from `client-diag.jsonl`.

## Measured

Offline replay of a seventeen-session board in headless Chromium, hidden/frozen tabs emulated in-page (Chrome's own freeze cannot be driven headless), medians of three runs per row, same harness before and after:

| regime | redials | MB after the switch | busy ms, first 2 s | longest task ms | time to current state |
|---|---|---|---|---|---|
| frozen 45 s, socket alive | 4 → 0 | 17 → 2 | 532 → 354 | 135 → 96 | 1231 → 475 ms |
| frozen 45 s at 6× cadence | 4 → 0 | 28 → 13 | 427 → 452 | 214 → 300 | 2195 → 585 ms |
| hidden 45 s | 0 → 0 | 0 | 58 → 51 | 83 → 74 | paints while hidden 14–15 → 0 |
| hidden 10 min | 0 → 0 | 5 → 5 | 168 → 124 | 123 → 105 | paints while hidden 127–192 → 0 |
| socket already dropped by the kernel | 4 → 4 | 15 → 15 | 483 → 424 | 118 → 135 | new socket in 72–98 ms (was 114–215); current state 1.3–1.4 s (was 1.1–1.2 s: the panes' deferred paints now land ahead of the resync) |

The dropped-socket row is the next change's subject (an active-tab-only reconnect push with skeleton tabs); this PR makes it no worse and instruments it.

## Review

A design critique (three critics, a defender, a synthesizer) reshaped the first proposal before any code: the `hello` probe was dropped for the `resume` stamp, per-slot deferral became one ordered FIFO, coalescing moved off `requestAnimationFrame` (held while hidden) onto `MessageChannel`, and the harness was rebuilt after its chat numbers were shown to be a refusal loop of its own making. Two adversarial passes with a refuter per finding on the implementation, plus a skeptic pass on the two late refinements: one finding survived (the keep-decision breadcrumb written into an already-dead socket), fixed and pinned.

## Tier

- [ ] tests-only
- [x] fix — a bug fix with tests that fail before it (the shim's return-to-tab and reconnect behaviour).
- [ ] feature
- [ ] major-feature

## Tests

`tests/test_pane_shim_return.py` (36 tests) runs the real shim under node with a faked browser: the resume stamp and the watchdog, the immediate redial on the first close after a foreground and the 250 ms wait on the next, the ordered queue with its whole-state replacement and time slicing, the held return row re-filed onto a redial, the breadcrumb rows and the badge cue. `ui/webview/paint-gate.test.ts` and `feed-hidden-paint.test.ts` pin the pure gate and the feed's hold, release and snap; `ui/timeline-hidden-hold.test.ts` the parked bars and the held draw; `ui/webview/chat-tail-repaint.test.ts` the coalesced repaint; `ui/webview/federation-reconnect.test.ts` the remote sockets' resume stamp; the heartbeat, banner, restart-seam, loader, outline-visibility, feed-flip, follow-move, awaiting-box, live-tick and bars-loader pins follow the new lines deliberately. Full suites on this tree: Python 7368 passed / 40 skipped; node 2836; bats 402.

### DIFF
--- a/docs/read-side.md
+++ b/docs/read-side.md
@@ -56,7 +56,13 @@ completed); the feed just paints columns. (Reflected in `docs/judges.md`.)
   contract.
 - **Liveness is a keepalive, and staleness is event-keyed.** The kernel sends a
   `ka` frame to every socket every 10 s; the pane shim abandons a socket that has
-  gone 30 s without any frame and redials at once. After a reconnect the shim
+  gone 30 s without any frame and redials at once, the silence measured from the
+  later of the last frame and a Chromium `resume` (the thaw of a frozen tab) on a
+  still-open socket. A socket kept on the strength of that stamp is provisional:
+  if no frame confirms it within 15 s (1.5 keepalive periods) the watchdog puts
+  it down there instead, a socket already 30 s overdue when the tab froze is not
+  stamped and is redialed at the return, and a browser without `resume` behaves
+  as before. After a reconnect the shim
   raises the "what you see may be stale" prompt only on the SECOND `ka` arriving
   before the resync frame — one full heartbeat period, bracketed by two kernel
   heartbeats on that socket with no resync between them (a single `ka` can be a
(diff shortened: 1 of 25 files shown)


lczh:
Tier: `fix` — a fork author cannot add labels here, so asking a maintainer to apply it. Before this change the pane shim judged its socket by the age of the last frame it received, so a frozen tab always read as stale and every pane redialed on return (`tests/test_pane_shim_return.py` fails before it). Stage 2 of the same work is the stacked PR that follows.

PR #1017, opened 2026-09-07
A reconnecting dashboard page gets its active session and a strip of skeleton tabs, not every transcript

lczh:
Tier: feature

Stage 2 of the return-to-tab work. Stage 1 (#1016) is merged. This PR is the stage-2 commit, the two review folds, two merges of `main` (one from the Update-branch button, one resolving `docs/reference.md` against the Billing tooltip rewrite, both texts kept) and one fold the screenshots below uncovered (the strip's repaint gate, see Review).

## What

When a pane's socket had been dropped while the dashboard tab was away (a long freeze, a laptop lid, a network change), the page redialed and the kernel served the new socket as a client that held nothing: every session's transcript tail, ~9 MB of the ~15 MB a seventeen-session board pulls after the switch, for one tab on screen. Stage 1 keeps a live socket alive; this stage makes the unavoidable reconnect cheap.

- **The reconnect signal is the page's.** The pane shim appends `&reconnect=1` to its `/ws` URL when it has connected before (`everConnected`). The kernel's own signal, a superseded twin socket, misses the cases where the kernel had already dropped the old socket itself (its no-pong rule, or a socket the page found closed).
- **Kernel (`kernel.py`).** For such a client the first tab strip resolves a skeleton set (`_resolve_reconnect` / `_skeleton_for`: every listed session except the active one, transcript-less ones and any already sent on this socket, smallest transcript first) atomically under the client's slot lock. The ONE tab-order frame builder (`_tab_order_frame`, via `_send_tab_order` for all four senders) attaches `skeleton: [...]` on every strip once a set has existed, an empty list after it empties, and prunes the set to the live strip. `_send_chat_or_status` sends a skeleton session a deduplicated `{type:"status"}` frame on its own slot instead of the full or a tail, so chips stay honest without transcripts. A full send, a `needFull`, or `activeTab` releases a session; `ready` (a fresh page) clears everything, so a first load is byte-for-byte unchanged.
- **Pane shim.** Besides the DOM event, the reconnect is enqueued as a `{type:"wsup"}` frame in the shim's ordered dispatch queue, behind the dead socket's still-draining frames and ahead of the new socket's, so a bundle scoping "loaded on this socket" sees the flip in the right place.
- **Chat pane (`render.ts`, new pure `ui/webview/skeleton-tabs.ts`).** A skeleton tab wears its name and color, the shared status chip (the unknown ring until a status frame lands), no swirl; draggable and closable like any tab; title "Not loaded yet — click to load". A click shows the loader at once and asks for the full (`needFull` now carries a `why`); the idle prebuild asks for one skeleton at a time, cheapest first, never while the tab is hidden. Every active-tab display path reads through `liveSession()`, which returns nothing for a skeleton, so the transcript the page held before the outage is never shown as current; a delta for a skeleton session asks for the full rather than splicing onto a stale base; a strip that releases the active a frame before its full arrives leaves the loader up until the full lands (`skeletonLoading`). The client adopts no id the strip did not list. One `clientDiag {what:"skeleton"}` row per reconnect names the regime on the user's machine.
- **Federation (`federation.ts`).** `skeleton` joins the host-prefixed array ids; a per-host list (array replaces, absence keeps pruned to the new order, `closed` removes, detach drops); the merged strip always carries the union.

## Screenshots

From a synthetic replay (invented sessions, `TESTHOST`): the chat pane after its socket was dropped and the page redialed.

| The tab you were reading is back in full; the others are skeleton tabs | A skeleton tab clicked: the loader stands in before any kernel frame | The clicked tab, loaded |
|---|---|---|
| skeleton tabs after the reconnect | the loader on a clicked skeleton tab | the clicked tab loaded |

## Measured

Stage-1 harness (17 sessions, tab frozen 45 s then returned, 3 runs per row, medians; dashboard-wide "current" = every pane shows the marker planted during the absence). Frame identity is exact, not statistical:

| after the reconnect | stage 1 | this PR |
|---|---|---|
| session frames in the connect push | 17 (all transcripts, ~9 MB in one burst) | 1 (the active) + 16 `status` frames |
| further fulls | 0 | 16, one at a time on idle `needFull` asks, all `why: prefetch`, no `nobase` / `gap` |
| skeleton tabs when the active tab is current / at the end of the 12 s window | – | 16 / 0 |
| chat bytes before the active tab is current | not measured (the burst above) | 0.54 MB |
| dashboard current, frozen tab whose socket got a FIN | 1317 ms | 996 ms |
| dashboard current, socket the kernel dropped for silence | 1445 ms | 1111 ms |
| chat pane's own time to current | ~226 ms | ~242 ms (unchanged: the active was already first) |
| chat bytes over the whole 12 s window | 9.0 MB | 9.0 MB (the prefetch fills every skeleton within the window) |
| click on a skeleton tab: loader / session shown | – | 17 ms / 145 ms |
| chat current on a paced 25 Mbit / 5 Mbit link | not measured | 736 ms / 1.8 s (dashboard bound by the timeline's ~4 MB of bars) |

Plainly: this PR changes WHEN the transcripts arrive (one cheap frame first, the rest in idle time, never while hidden) and what the strip shows meanwhile; it does not reduce the total pulled by a page that stays on the tab. The dashboard-wide time to a current screen in the dead-socket regime is now bounded by the timeline's ~4 MB of bars, not the chat; that is the next step on this ladder and is not claimed here. Two caveats: one of three runs of the dropped-socket row saw the shim dial twice for one return (the tab came back while a redial was still connecting; the visibility handler aborts a CONNECTING socket as stuck), costing one extra 0.5 MB full; the harness preflight now reports that case, unfixed here. And the matrix ran on the tree BEFORE the five review folds below; those are covered by unit tests and were not re-measured.

## Review

A written spec first (from a read of the current code, with three findings that reshaped the plan: the shim's flag as the signal, the unused client `status` handler as the honest chip channel, and the timeline as the true bound). Then a two-lens adversarial review with a refuter per finding on the implementation: five confirmed, none refuted, all folded with tests. One high: a click on a skeleton tab whose releasing strip preceded its full left the loader up (now the `skeletonLoading` latch). One medium: the set was resolved outside the client's slot lock (now atomic, and a sid already sent on the socket is never skeletoned). Three low: a session that left the strip lingered in the set (pruned in `_send_tab_order`); the client adopted ids the strip did not list (array path filters to the order); the socket-up handoff ran at `onopen` ahead of the dead socket's still-draining frames (now the queued `wsup` frame).

**Fold after the screenshots (26dce39c):** main's tab-strip repaint gate (`tabStripSig` in `renderTabs`) hashed every input of a loaded or placeholder tab and none of a skeleton's, so a strip whose sessions had not changed skipped its rebuild when the skeleton set arrived and no skeleton tab painted. The signature now carries a skeleton tab's own reads (the kernel's list, its status frame, the peek state); `tab-strip-skip.test.ts` and `skeleton-tabs-wiring.test.ts` pin it.

## Tier

- [ ] tests-only
- [ ] fix
- [x] feature — a self-contained capability inside romp's existing model (the pane socket, the tab strip, the pusher's per-client slots); design points above.
- [ ] major-feature

## Tests

New: `tests/test_chat_skeleton_reconnect.py` (18: helpers, the frame builder, the status slot, releases, atomic resolution, pruning, every touch of the set under the lock), `ui/webview/skeleton-tabs.test.ts` (8, pure state), `skeleton-tabs-wiring.test.ts` (14, render.ts pins incl. the latch), `federation-skeleton.test.ts` (12). Pins moved: `test_kernel_delta_send`, `test_kernel_tabs_first`, `test_pane_shim_return` (the `wsup` marker's place in the FIFO), `test_kernel_disconnect_banner`, `test_pane_loader_reconnect`, `test_timeline_views`, `awaiting-state.test.ts`, `chat-delta-resync.test.ts`. Full suites on the final tree: Python 7944 passed (one browser drag test flaky under machine load, green alone), node 3009, bats 447.


lczh:
Tier: `feature` — a fork author cannot add labels here, so asking a maintainer to apply it. A self-contained capability inside the existing model (the pane socket, the tab strip, the pusher's per-client slots): a reconnecting page gets its active session plus skeleton tabs instead of every transcript; design points and measurements in the body. Stacked on #1016; please merge that first.

lczh:
Rebased onto main e76bef9e (the merge of #1064); your two commits are kept as they were, only the stage-2 commit's content changed to resolve conflicts. The one substantive resolution: main's T258 `_tab_order_frame(order, tabs, live)` is now the single builder and takes an optional trailing client, which appends the `skeleton` list when that client has had a set, so a strip carries both the affirmed-live sids and the skeleton list and `_send_tab_order(c, tab_order, tab_meta, live)` passes the liveness map through under the client lock. federation.ts keeps `perHostLive` and `perHostSkeleton` side by side (ARRAY_ID gains `live` then `skeleton`), render.ts keeps the T249b held-transcript guard beside the skeleton gate, and the source pins were re-anchored rather than dropped. Modules run one at a time: test_chat_skeleton_reconnect 17, test_kernel_tabs_first 3, test_timeline_views 44, test_pane_shim_return 41, test_live_session_unresolved_transcript 7, plus the skeleton, federation, tab-order and session-views node bundles, all green. One interplay left untested that a reviewer may want: a sid that is skeleton on a client and also omitted from `order` while live (T258's retained path) leaves the skeleton set on that frame and renders as a plain placeholder until the next strip re-lists it; the kernel-side T258 fix keeps such a sid in `order`, so this is defense in depth only.

lczh:
Rebased again onto main 51acdc60 after #1128 and the tab-strip work that followed; your two commits are unchanged, only the stage-2 commit's content moved. Resolutions: the single `_tab_order_frame(order, tabs, live, c=None)` keeps main's new `selfHost` field and `**_views_payload()` carrier and appends the client's `skeleton` list; the T264b per-group tab copies (`draggedEl`, `data-copy`, the copy passed to the context menu) and the T262g dot slot are folded into the shared `wireTabDrag` / `applyTabStatus` helpers verbatim, so the skeleton tab wears both; federation keeps `selfHost` beside `skeleton`; the retired-provider changes are untouched. Main's own pins were re-anchored, none dropped, including the `tab-strip-skip-exec` harness, which now lifts the helpers for real. Modules run one at a time: the skeleton, tab-order, live-session, credentials and session-auth Python modules plus the pusher-change, wire-once and perf pins, and the tab-strip, tab-groups, tab-drag, federation and skeleton node bundles, all green; the static pin scan reports nothing lost. Ready for review on this head.

lczh:
CI's extension job caught one more main-side executed harness that lifts `chatTail` (`chat-exact-tail-exec.test.ts`, T262i): it now supplies the skeleton state the handler reads and gains one test, a tail for a skeleton tab asks for the full and touches nothing. Folded into the stage-2 commit; your two commits are unchanged in content. Every executed harness over render.ts was then run one at a time (awaiting-rows, chat-exact-tail-exec, comment-reply-landed, gear-models-frame, models-rev, file-view, provisional-cancel, setting-stale-fold, tab-groups, timeline-awaiting, spend-order-twin, thread-selection-scope, tab-strip-skip-exec), all green. Ready for review on this head.

lczh:
Since the Update-branch merge (260cf004): `main` merged again (`docs/reference.md` conflicted with the Billing tooltip rewrite; both kept, our not-loaded sentence ahead of main's paragraph), then one fold, 26dce39c: main's tab-strip repaint gate (`tabStripSig`) carried no skeleton inputs, so on an otherwise unchanged strip the skeleton tabs never painted; the signature now carries them, pinned by `tab-strip-skip.test.ts` and `skeleton-tabs-wiring.test.ts`. Screenshots of the regime (synthetic replay) are in the body's Screenshots section and in the guide's chat section. The head moved, so the feature needs the owner's approval on it.

henrypinkard (approved):
Approved by the maintainer (feature tier).

PR #994, opened 2026-09-07
Requests from sessions: what a session needs from you, held until you or it says otherwise

ksarma:
Discussion: #993

A session can now file a request with the person it works for (a decision, a credential, a look at a draft) and keep working on everything else. The request stays in view until you answer it, you dismiss it, or the session withdraws it; nothing that reasons by inference clears one. It shows on the card at the bottom of the session's transcript (a **Waiting on you** section with **Reply** and **Dismiss** on each row), as a flag after the session's name in the tab strip, and as a quiet marker on the session's feed cards; once the session has gone idle on a request it marked as blocking, its card moves to Blocked and the app badge counts the request. After a resume, a compaction or a clear the session gets its open requests back, so it can withdraw the ones that are met. Everything sits behind a per-machine switch that ships off (the gear's **Requests from sessions** row, under Settings, Sessions, Requests); an install that never turns it on sees no change in its payloads.

## Design of record

The RFC's questions are decided as follows, and the code does each of these.

1. **The switch ships off, and off is inert.** Only a literal `true` in `user-todos-enabled.json` turns it on; an absent file reads off silently, a malformed one reads off and is said once per file version. The reader is memoized on the file's stat. Off, the routes answer 409, the bus leaves the two tools out of `tools/list`, the one gated store read returns nothing (every surface goes quiet from that one read), the chat's delta frames carry no `userTodos` key, the boot pass over lost answers returns after one store read, and the SessionStart hook exits on a stat of the switch file. The store is untouched: rows filed while the switch was on stay on disk, and the kernel log counts them at boot.

[...]

PR #1018, opened 2026-09-07
The auto-nudge and retry-suppress ledgers are never rewritten from a fabricated default after a read fault

lczh:
The auto-nudge and retry-suppress ledgers are never rewritten from a fabricated default after a read fault

Two automatic ledgers under STATE — auto-nudge.json and retry-suppressed.json —
are each read by one reader and written by one writer, with many
read-modify-write sites between them (`d = dict(reader()); d[k] = v; writer(d)`).
Both readers answered ANY fault with the fresh-install default and cached it
under the file's real stat key, so the next writer persisted the fabrication.

auto-nudge.json: `_auto_nudge_data` returned {"enabled": True, "nudged": {}} on
any stat OSError (not just ENOENT) and on any read or decode fault, then cached
it. After one EIO on a tick the dedupe map read as empty, so every stalled goal
across every session was nudged again, every tick; and a user's explicit
auto-nudge OFF came back ON.

retry-suppressed.json: `_retry_suppress_data` did the same with {}; the arm in
`_suppress_session_retry` then wrote {this sid} over every sibling session's
"stop retrying" — a retry storm the user had explicitly halted, resumed. The
arm's write also had no OSError handling: an ENOSPC out of the interrupt handler
reached the WS reader loop, which reads any OSError as a socket failure and tore
the dashboard's connection down with the suppression neither on disk nor
anywhere. The arm and the pusher's re-arm sweep rewrote the whole blob unlocked,
and a non-numeric floor raised TypeError into the sweep and killed it.

The fix, shared by both ledgers (`_ledger_read` / `_ledger_write_proved`):

- Missing is the ONLY fault that reads as the default — a file that is not
  there IS the fresh-install state (ENOENT behavior unchanged byte for byte:
  uncached, no log). Any other stat/read fault yields a COPY of the last
  snapshot this process proved (else the default) tagged "_unproved" with the
  fault text; nothing unproved is cached, so the file's next successful read
  ends the episode — an event, not a clock. Every proved state — a read, a
  cache hit, absence, a write — resets BOTH once-per-episode registries, so the
  same fault returning after a heal is said again. Fault texts carry no
  destination path and no stamp, so a fault that persists has one text.
- The writer is the one choke point: it refuses any snapshot carrying the tag
  (one stderr line per fault episode, keyed on the fault text) and returns
  False; the file keeps what it holds. A write that lands refreshes the cache
  with what it read back under the file's stat key, so "the last snapshot this
  process proved" includes this kernel's own last write. The tag travels IN the
  snapshot because every RMW site copies it with dict(): a flag beside the
  cache would not survive the copy, and a per-path "last read failed" bit races
  a sibling thread's proved read. The ~20 auto-nudge RMW sites are covered
  unchanged.
- Bytes that do not parse (invalid JSON, a non-object value, not UTF-8 — the
  ledgers are read as UTF-8, never the locale's encoding) are moved aside as
  <file>.corrupt-<utc stamp> (a -n suffix within one second) — the judge's
  goal-store quarantine shape — evidence kept, never deleted, one stderr line,
  and the ledger then reads as absent. The move is guarded on the file still
  being the one read (same inode and generation); a file that cannot be moved
  (a read-only state dir) rides the deduped fault text as errno + strerror,
  never a line per read and never a text per second.
- The auto-nudge pass stands down whole on an unproved snapshot (nudge, wake,
  compaction suggestion, debt ladder, sweeps — each records what it fired in
  this ledger, and a refused record re-fires every tick), once-loud on stderr
  and in the dashboard's error center, and resumes on the first tick whose read
  proves. The debt reminder and the compaction-suggestion claim check the tag
  themselves for a fault that lands mid-pass. The interrupt→blocked tick, which
  runs outside that pass, files no block under a fault (its once-per-episode
  marker could not be minted, and the lift is gated on the marker); a block it
  did file — a proved goal-store write the feed must hear — pushes whatever the
  marker write's fate; the lift on re-engagement runs whatever the ledger's
  state, and pushes only when its marker clear lands (a refused clear with an
  idempotent lift is not a push per cycle).
- `_set_auto_nudge` / `_set_compact_suggest` report a refused write as
  not-applied (None) and record a `refused` notice on the thread, a sibling of
  `_setting_stale`'s stand-down; the WS branch answers the delivering socket
  with the settingStale frame plus `why` (the gear re-fills the checkbox and
  says not applied, keeping the stored value — `kept` is the last snapshot this
  process proved, named only when there is one) and files the fault in the
  error center.
- retry-suppress: `_suppress_session_retry` returns None when the stop is on
  disk, else one sentence naming why (an unproved read, or the write's
  OSError, now caught); the interrupt handler warn-toasts it while the
  interrupt itself still happens. Arm and clear hold a new
  `_RETRY_SUPPRESS_LOCK` across their read-modify-write only. A non-numeric
  floor is read as no floor, said once per session.

Unchanged: `_auto_nudge_on` (display); the fresh-install path; every RMW
site's code; the OSError arm of the setting toggles.

Tests: 32 new, run against a `git archive` copy of origin/main f06e6678. 27
fail there, each on its defect's assertion (a rewritten file, a fired message,
a raised OSError or TypeError, a silent socket) — never on a missing name.
Five pass on main, by construction or by design: the two ENOENT pins
(FreshInstall and retry-suppress: that arm is unchanged), the two mid-pass
controls (they prove the path is reached on a healthy ledger), and
test_no_compaction_suggestion_when_the_ledger_faults_mid_pass — which passes
on main for an unrelated reason (main's fabricated default lacks
compactSuggestEnabled, so `_compact_suggest_tick` returns at its own gate) and
earns its place as the sole killer of the claim-latch mutant.
- tests/test_ledger_unproved_reads.py (new): the seeded OFF + dedupe record,
  Path.read_text raising EIO; `_set_auto_nudge`, `_set_compact_suggest` and
  `_mark_auto_nudged` leave the bytes unchanged and log one refusal; a stat
  fault; the tagged copy of the last proved snapshot; once per episode across
  six same-text episodes ended by a write, a read, a cache hit and ENOENT;
  corrupt bytes / a JSON array / non-UTF-8 quarantined in the judge's shape
  with the -n suffix; a file replaced under a corrupt read is not moved aside;
  an unmovable corrupt file has one fault text across three seconds; ENOENT;
  the interrupt→blocked tick under a fault (no block filed, no push storm, the
  lift honored, a mid-tick fault pushes the block it filed exactly once).
- tests/test_wake_deadman_toggle.py: the pass fires nothing under EIO, logs
  once, posts one error-center row, resumes on the first proved read; an
  explicit OFF is never read as ON; the debt reminder and the compaction
  suggestion send nothing when the fault lands mid-pass and the head gate
  stayed open (with controls).
- tests/test_session_retry_suppress.py: a read fault leaves {s1, s2} byte for
  byte, the clear reaches the writer's refusal, the click is told; ENOSPC is
  reported, not raised; the interrupt handler toasts through the real
  dispatcher; a non-numeric floor; corrupt bytes; ENOENT; lock and toast pins.
- tests/test_setting_gesture_order.py: a toggle under EIO through the real
  dispatcher — file unchanged, the socket hears the frame with the kept value
  and the fault, no act-now tick, one error-center row; the kept value includes
  this kernel's own last write; with no proved snapshot the frame names none.

Measured on this commit: the four touched modules (118 passed) and the 34
ledger-touching modules in one sequential run (1156 passed, 25 subtests). The
full suite is left to CI (not run on the development machine).

## Tier

**fix** — a bug fix with tests that fail before it. A fork author cannot apply labels here; asking a maintainer to apply `fix`.

## Notes for review

- Tests were run per module on the development machine (the four touched modules and the 34 ledger-touching modules, one process at a time); the full suite is left to this PR's CI.
- **#994** (user todos) reads `_auto_nudge_data` at two new sites; it does not touch the reader or writer this changes, so whichever lands second should rebase cleanly.
- The quarantine sidecar shape (`<file>.corrupt-<UTC stamp>[-n]`, `os.replace`, an inode re-check) matches the goal-store PR opened alongside this one, so the two stores wear one convention.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -871,7 +871,11 @@ parse: the file is renamed `<name>.corrupt-<UTC stamp>` beside the original
 error center says so, and the ledger reads as a fresh install until you
 restore it from that file. Nothing is deleted. Any other read fault leaves the
 file untouched: the kernel serves the last copy it read, writes nothing to it,
-and says so once, until the file reads again.
+and says so once, until the file reads again. A write that fails (a full or
+read-only disk) is told to the gesture that asked, the gear's toast or the
+stop button's warning, and said once per fault episode in the error center;
+the automatic pass sends nothing whose record could not land, and the file
+keeps what it holds.
 
 ## Switches
 
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -863,6 +863,16 @@ entry under the same kind says so. A file that cannot be read at all keeps
 showing its last-read values until it can. Nothing here needs a restart; the
 sidecars are yours to inspect or delete.
 
+Two small ledgers there, `auto-nudge.json` (the auto-nudge switch and its
+per-goal records) and `retry-suppressed.json` (the sessions whose auto-retry
+you stopped), are moved aside rather than overwritten when their bytes do not
+parse: the file is renamed `<name>.corrupt-<UTC stamp>` beside the original
+(`-1`, `-2`, ... when a second one lands in the same second), the dashboard's
+error center says so, and the ledger reads as a fresh install until you
+restore it from that file. Nothing is deleted. Any other read fault leaves the
+file untouched: the kernel serves the last copy it read, writes nothing to it,
+and says so once, until the file reads again.
+
 ## Switches
 
 Effective immediately, no restart.
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -4106,8 +4106,10 @@ def _setting_stale(name, gt, applied_gt):
     alone left the refusing kernel's verdict invisible to the dashboard that made the gesture
     (the open gear kept showing the refused pick, and with the mesh AGREEING on the kept value
     the mixed marks showed nothing). Every check clears the previous verdict first, so a popped
-    notice is always the CURRENT call's."""
+    notice is always the CURRENT call's — the refused-write verdict (_note_refused_gesture) included,
+    so a setter called outside a WS arm (nothing pops there) never leaves one for an unrelated gesture."""
     _stale_seen.last = None
+    _stale_seen.refused = None
     if gt is None or gt > applied_gt:
         return False
     _stale_seen.last = {"setting": name, "storedGt": applied_gt, "gt": gt}
@@ -4138,7 +4140,7 @@ def _pop_stale_notice():
     return d
 
 
-def _note_refused_gesture(name, gt, enabled, snap):
+def _note_refused_gesture(name, gt, enabled, snap, why=None):
     """A gt-gated toggle's write was REFUSED because its ledger read was unproved (the snapshot carries
     UNPROVED — the file could not be read, so the writer would not persist a copy of it). Recorded on this
     thread the way _setting_stale records a stand-down, so the WS branch that delivered the gesture can tell
@@ -4147,9 +4149,13 @@ def _note_refused_gesture(name, gt, enabled, snap):
     the file healed, ON is what ran. A different notice from the stale one on purpose: nothing about
     ordering is claimed, and the frame names the fault. `known` says whether a kept value can be NAMED at
     all: the tagged copy is the last snapshot this process proved (its last read or its own last write)
-    only when one is cached; otherwise it is the default, and the frame must not present that as kept."""
+    only when one is cached; otherwise it is the default, and the frame must not present that as kept.
+    `why` given: the WRITE itself failed (the snapshot was proved; the publish raised) — the same frame, the
+    fault named, and `write` set so the reply files no second error-center row: the writer already filed the
+    episode's one (the maintainer's fold on PR #1019: a refused write is told, once per episode)."""
     _stale_seen.refused = {"setting": name, "gt": gt, "refused": "on" if enabled else "off",
-                           "why": str(snap.get(UNPROVED) or "the ledger could not be read"),
+                           "why": str(why or snap.get(UNPROVED) or "the ledger could not be read"),
+                           "write": why is not None,
                            "known": str(jd.STATE / "auto-nudge.json") in _autonudge_cache}   # kept: None → the
     #                                                                                gear drops "Keeping …"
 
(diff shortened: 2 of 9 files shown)


lczh:
Tier: **fix** (a bug fix with tests that fail before it). A fork author cannot add labels here, so could a maintainer apply the `fix` label?

lczh:
Follow-up commit c6f547e8, on top of your review fixes: it carries the save-step lesson from #1019 to these ledgers. A publish that fails at the write door (a full disk, a read-only state dir) is now said once per fault episode and answered on the gesture's socket through the `settingStale` + `why` frame your fold added, with the gear naming the write as the cause; the retry-suppress sweep's clear reports a failed publish instead of raising into the pusher; and a quarantine that declines because the file changed re-reads the peer's bytes (bounded) instead of serving an unproved copy. The pre-existing pin that a full-disk toggle "sends nothing" is changed deliberately, with the reason in the test.

lczh:
The follow-up is amended in place to d8fc5e01: same behavior, but the gear helper's write-failed branch is compacted to your three-line footprint so `setting-stale.test.ts`'s slice bound holds (the first cut overran it and turned the extension job red). Every webview test that reads gear.js was run singly and is green.

lczh:
Rebased onto main (bebe6b35) after #1019 and #1021 landed; your review-fixes commit rides along unchanged. One conflict, in `_interrupt_block_tick`'s lift arm: #1019 made the lift return whether the marker is spent, this branch made the marker write report whether it landed — the resolution takes both, so a store fault keeps the marker for the next tick and a refused marker write pushes nothing. Touched modules and #1019's boundary module are green; CI is re-running.

lczh:
Rebased again onto main (1860a254) after #1020 landed. Two textual overlaps (the import line; a docs paragraph), both kept as the union; your review-fixes commit rides along unchanged. One alignment: the follow-up commit now uses the `_errno_text` helper #1020 brought to main instead of its own identical copy. Ledger modules plus #1020's flag and bell modules are green; CI is re-running.

PR #1019, opened 2026-09-07
A goals file that cannot be read is never republished as an empty store: read faults raise and unparseable bytes are quarantined aside

lczh:
A goals file that cannot be read is never republished as an empty store: read faults raise and unparseable bytes are quarantined aside

One session's goals file that cannot be read or parsed is preserved and reported,
never replaced by an empty store, and never lets one session's fault take the board
down for every other session.

## Tier

- [x] `fix`: a bug fix, with a test that fails before it.

## The defect

Every reader in the goal store's save path answered a read failure with the
ABSENT-file value: load_goals swallowed any exception into a fresh store at
_baseRev 0, _disk_rev read 0, _matches_disk read "no match", _rebase_onto_disk
had "nothing to rebase onto". So on an I/O error, a permission fault, or a
torn/half-written file, save_goals found base 0 == disk 0 and published the
empty store over the user's goals file, destroying the only copy of everything
it held. ENOENT, EIO, EACCES and corrupt content all took the same branch.

## The fix

1. One strict reader, `_read_store_json(path, *, quarantine=False)`: an absent
   file reads as None (a session with no store yet); a read FAULT (EIO, EACCES,
   a directory at the path) RAISES, always; UNPARSEABLE content (invalid JSON,
   bytes that are not UTF-8, a top-level value that is not an object) is moved
   aside to `<file>.corrupt-<utc stamp>` when `quarantine` is set and read as
   None, and raises otherwise. load_goals reads with quarantine on, so a session
   starts fresh only after the bad bytes are preserved, with one stderr line and
   one `store-quarantined` judge-errors row recording the move. _disk_rev,
   _matches_disk and _rebase_onto_disk read strictly, so a fault or a corrupt
   file at save time raises out of save_goals and the file is left byte for byte
   as it was.

2. A per-session boundary, `jd.load_goals_or_fault(fsid)` -> `(store, None)` or
   `(None, exc)`. load_goals stays strict; this is where its raise is caught,
   filed and CONTAINED to the one session it is about. The first fault files one
   `store-unreadable` judge-errors row for that sid (the surface every
   per-session failure already reaches: `romp judges`, the feed's debug join); an
   identical repeat files nothing; a fault with different text is a new episode;
   a successful read of that sid ends the episode (the event, not a timer), so a
   later fault files again. It never hands back an empty store: a caller that
   gets `(None, exc)` skips that session's goal-derived work and writes nothing.

   Without it the strict raise reached nothing outside the judge passes: the
   feed builder's session loop (`_feed_goals`, its live fallback for any sid the
   pre-pass snapshot skipped), build_timeline, build_session's two reads, and the
   kernel's ticks all ran under the pusher's single outer try, so ONE faulting
   file made the pusher log a line and send nothing to any client, every cycle,
   for as long as the fault lasted; run_courier and run_propagate had no
   per-session catch, so a raise there skipped the grouper, consolidator and
   distiller for every session that pass; and a raise inside a WS gesture
   handler would have reached the receive loop's catch, which re-raises an
   OSError to the outer handler, where it is SWALLOWED and the client dropped
   (`except (BrokenPipeError, ConnectionResetError, OSError): pass` +
   `finally: client["alive"] = False`) - the dashboard disconnected without a
   word. A raise from load_goals is now contained per session at every kernel
   and judge site, not only in the pass wrappers:
   - routed through the boundary: `_feed_goals` + build_feed's loop (the
     faulting session renders with no goal-derived content: no cards, so no
     floors and no Analyzing swirl, which land only on cards; the provisional
     card is gated off, since "the planner has not placed this yet" is an
     inference from placements we could not read; a render-only stand-in that is
     never saved), build_feed's peer-store read for the origin badge (a sender
     whose store faults reads absorbed), build_timeline (the lane renders without
     goal-derived data and `_bars_complain(sid, "goals", ...)` names it), both
     build_session reads (seams off, an empty ledger tree, the tab still builds),
     `_followup_body`, `_record_interrupt_block`, `_lift_interrupt_block` (which
     returns False on a fault so `_interrupt_block_tick` KEEPS the intrBlocked
     marker and retries the lift next tick - erasing it would have left romp's
     own block on the card with no tick ever looking again) and `_intr_block_stands`
     (an unreadable store reads as "the block still stands", so a SECOND stop
     during the fault keeps the marker too instead of popping it and failing to
     re-block through the same fault), `_lift_spent_awaiting`
     (the gate is forgotten so the next cycle retries), `_auto_nudge_session`,
     `_compact_goal_store` (+ the sweep does not mark an unreadable store as seen,
     since its mtime never moves), `_mark_nodes_cleared` and `_restore_goal_archive`
     (the faulting session is skipped and RETURNED as `{sid: fault}`; `_clear_all`,
     `_clear_ask` and `_undo_clear` pass that up, and the WS `askClear`, `clear`,
     `clearAll` and `undoClear` ops answer the socket that made the gesture with
     the feed's `err` dialog - `_gesture_store_refusal` - because a refusal of a
     user gesture must reach the user, not only the judge-errors log. The dialog
     says what happened, in `text`, with no promise, and three different things
     can have happened: a whole-card clear (askClear, Clear-all; one dialog per
     session, worded for any number of cards) DID take the cards off the board -
     cleared.jsonl hides top cards on its own - but the durable flag was not
     written into a goals file romp could not read; a sub-goal drop (the modal's
     Drop, sub-task-only) changed NOTHING, because a sub row renders from the node
     flag alone, the very write the fault refused, so the user is told to try
     again (a retry appends a fresh row and sets the flag); an undo restored
     nothing and the next Undo retries it. It sends no `copy`, which by contract
     is the USER'S undelivered text - the feed renders it as a "Copy my text"
     button and folds it into the bell entry. `_undo_clear` keeps a faulting
     session's ids OWED in every fault shape: a compacted card's restore read
     faults, so its ids are never journaled as undone; an id that was journaled
     (an uncompacted card, whose restore step has nothing to read; or a store that
     faults between the restore and the flag step) whose flag step then could not
     run is re-journaled as cleared. Either way those ids stay the newest cleared
     batch and the next Undo restores exactly them. Journaling every id first
     consumed the batch and the nodes stayed in the archive for good; journaling
     last is not an option, since the reopen verdict's gate needs the undo row on
     disk before the flag step; and a node left restored-but-flag-cleared would be
     hidden exactly like the clear, with its undo spent and no modal op able to
     reach it (resolve and clear are the only two). Found by the Clear-all test:
     the WS clearAll handler indexed `d["items"]`, a payload key build_feed no
     longer emits, and raised before _clear_all ran - Clear-all cleared nothing and
     only a stderr line knew; it now reads `d.get("items") or []`), `_goal_segments`, `_cards_for_segments`, propagate's `_ref_goal` (a
     faulting recipient joins to nothing this pass, the tracker stays open; it
     files a `store-unreadable` row through the boundary, not `pass-crash`), and
     the index tier's `tasks_for` (no caption tasks for that session this pass);
   - run_courier and run_propagate: a per-session try around every store read
     files the same `pass-crash` row the other passes file and continues - the
     courier's own-session read, a relayed message's sender, a pending message's
     recipient and its sender, and the root walk (a faulting peer on the walk
     skips that message this pass rather than minting an "unrooted" top);
     propagate's recipient scan and the sender's store;
   - every remaining direct load_goals call sits inside an existing per-session
     catch: the planner/grouper/consolidator/closer/unblocker/distiller pass
     wrappers and the rewound-reconcile loop (`pass-crash` / `rewound-reconcile`
     rows), the courier's link-repair try, the kernel's per-session tick and
     sweep catches (dead-wait sweep, awaiting-wake outcomes, deferral sweep, debt
     escalation, episode boundary, message-summary scan, nudge-failed, interrupt
     marker check, wake, fork seeding, stamp read, background-task tops), the WS
     gesture handlers that answer the client with an error (resolve, decision),
     `_drop_goals_after` (a `revert-failed` row), the follow-up reopen (stderr),
     best-effort decorations that return None/{}/[] on any failure, and CLI-only
     entry points (--ab-close, --ab-classify, the echo backfill).

## The old test asserted the defect

tests/test_judge_store_noop_publish.py's
test_an_unreadable_file_falls_through_to_a_real_write wrote "{not json" over a
published store and expected the file to be REPLACED by a fresh one. It is
rewritten as UnparseableStore: the bad bytes must be in a .corrupt-* sidecar,
verbatim, before the fresh store is published to the emptied path.

Unchanged: an absent file is still a fresh store and a create, with no
quarantine and no error; no file lock is taken (this codebase takes none), so
the quarantine rename shares save_goals' documented window with a concurrent
republish; the kernel's pre-pass feed snapshot (which reads the files itself) is
not touched here. The cleared-goal ARCHIVE keeps its swallow-then-blind-write
shape (load_goal_archive reads any fault as an empty archive; save_goal_archive
republishes), so the same fault class can still empty an archive - a follow-up,
not touched here.

## Tests

Fail on main unless marked; the boundary module also fails with the strict
reader alone (the raise escapes build_feed, _push, run_courier, run_propagate
and tasks_for).

  tests/test_judge_store_noop_publish.py::UnparseableStore
    test_an_unparseable_file_is_quarantined_aside_then_a_fresh_store_is_written
    test_every_unparseable_shape_is_quarantined_not_replaced
    test_an_absent_file_is_a_fresh_store_with_no_quarantine_and_no_error (passes
      on main: pins that ENOENT is unchanged)
  tests/test_judge_store_cas.py::ReadFaultCas
    test_a_read_fault_raises_from_load_instead_of_reading_as_an_empty_store
    test_a_read_fault_at_save_publishes_nothing_and_the_file_is_untouched
    test_a_file_corrupted_after_load_is_not_overwritten_by_the_save
    test_disk_rev_reads_zero_only_for_an_absent_file
    test_matches_disk_and_rebase_raise_on_a_fault_or_a_corrupt_file
  tests/test_goal_store_fault_boundary.py
    FeedBoundary: one faulting store costs that session only and files one row
      per episode (repeat files nothing, a successful read re-arms; the origin
      badge whose sender faults reads absorbed and comes back live); the
      faulting file is never written (passes on main: pins containment/no-write);
      the provisional card is not inferred from a store that faulted (and is back
      once it reads); the timeline frame ships with one lane faulting and
      `_bars_complain` names it; the chat tab builds with its store faulting
      (seams off, empty ledger tree, one row)
    PushBoundary: one faulting store does not stop the push (a feed payload with
      the healthy session's card reaches the client)
    InterruptLiftBoundary: a fault at the re-engage tick keeps the intrBlocked
      marker and writes nothing; the next healthy tick lifts the block and
      spends the marker; a second stop during the fault keeps it too, and once
      the file reads the re-engage lifts romp's own block
    TriagePassBoundary: run_courier and run_propagate continue past a faulting
      session and file its pass-crash row; the courier files a faulting relayed
      sender, a faulting re-read recipient and a faulting pending sender each
      as its own row and moves on; a root walk that hits a faulting peer skips
      the message (no recipient top minted, the message stays pending); propagate
      leaves a faulting sender's tracker for the next pass and joins a faulting
      recipient to nothing (one store-unreadable row, the tracker unwritten); the
      index tier captions the healthy session when one store faults
    GestureRefusal (through the real WS dispatcher): an undo whose session faults
      answers the socket with an `err` whose text names the fault, says the cards
      were not restored and that the next Undo retries them, and carries no
      `copy`; it writes nothing and keeps the archive; the per-card clear
      (askClear) answers with what happened (card off the board, flag not written,
      no `copy`); Clear-all across two cards of one session answers once, plural-
      safe and true, and clears (the `items` crash); a sub-goal drop on a REAL child
      node gets its own account, the row still renders un-cleared afterwards, and
      a healthy retry clears it; a compacted card's undo skipped for one session
      journals only the other session's undo rows and the next Undo restores
      exactly the skipped ids; an uncompacted card's undo, and an undo whose fault
      first appears at the flag step, both stay owed and are restored by the next
      Undo; a gesture that reaches every session says nothing (passes on main: a
      pin)
  tests/test_kernel_goal_compaction.py: an unreadable store is not marked seen,
    so the next sweep retries it and archives its cleared top
  tests/test_kernel_awaiting_lift.py: a faulting store forgets the inputs gate
    (_lift_seen) so the very next tick retries and lifts
  tests/test_timeline_bars_resilience.py: pins `_bars_complain(sid, "goals", gfault)`
    among the per-lane stages

Each per-session catch was checked against the mutant that restores the bare
read (every courier and propagate site, the provisional gate, the compaction
seen-mark, the _lift_seen forget, the timeline complain, both build_session
reads, the origin-badge peer read): each mutant is killed by the test named for
it above.

tests/test_kernel_rewind.py's source pin on build_session's ledger read is
updated to the boundary-routed shape (the hold filter still applies).

Ran, one module at a time (the full suite is not run on the maintainer's devbox
per the 2026-09-07 ruling; CI runs it): against the final tree,
tests/test_goal_store_fault_boundary.py (23 passed) and tests/test_kernel.py (464
passed); the wider sweep of the seven test modules the commit touches plus the 25
modules that pin or exercise the changed functions (32 single-module runs, 1643
passed, 0 failed) ran one revision earlier, before the final _undo_clear and
dialog-text shape. No .bats exercises this change (the only one naming judge.py
does so in comments). The five webview tests that read judge.py were re-checked
against the new source with their own JUDGE/JUDGE_FLAT transforms (21
match/includes targets, all hit; node_modules is absent on that machine, so they
were not executed via npm; judge.py is unchanged since). Against an origin/main
copy, the boundary module fails 21 of 23 (the two marked above pass), and the
compaction, awaiting-lift and timeline-bars additions each fail. Deleting the
re-journaled clear for a flag step that could not run fails exactly the
uncompacted-undo and flag-step-fault tests.

Every new test uses a private synthetic sid and cleans its override journal in
tearDown.

## Tier

**fix** — a bug fix with tests that fail before it. A fork author cannot apply labels here; asking a maintainer to apply `fix`.

## Notes for review

- Tests were run per module on the development machine (the boundary module, the judge-store modules and the modules that pin the changed functions, one process at a time); the full suite is left to this PR's CI.
- This also fixes a pre-existing bug the new dispatcher test exposed: the WS `clearAll` handler indexed a payload key the feed no longer sends, so Clear-all had silently done nothing. It is a one-token change with its own test; say if you would rather have it as a separate PR.
- **#1003** and **#1009** (perf counters, pane telemetry) add `load_goals` call sites; they do not touch the store readers this changes, so a rebase should be clean either way.
- The cleared-goal ARCHIVE readers keep their old swallow-then-write shape (named in the body); a follow-up.

### DIFF
--- a/docs/judges.md
+++ b/docs/judges.md
@@ -463,7 +463,15 @@ permission/API-error floors: one interrupt at a time, the present event first.
   `STATE/judge-errors.jsonl` (the row contract above; kinds are parse,
   call, give-up, sweep-cut, cite-miss, rate-limited, task-store, history-unreadable,
   task-key-collision — a duplicated to-do mirror key, reconciled per node
-  and surfaced loudly),
+  and surfaced loudly, store-unreadable: a goals file that cannot be read,
+  filed once per fault episode and ended by the next successful read,
+  store-unwritable: a goals file whose publish failed under a user gesture,
+  and store-quarantined: a goals file whose bytes did not parse, moved aside).
+  A file that does not parse is never deleted: it is moved beside its path as
+  `<file>.corrupt-<utc stamp>` (a `-n` suffix when two land in the same second)
+  before a fresh one is written, so the bytes survive for inspection, and the
+  `*.json` globs that enumerate stores skip it; the same sidecar convention
+  applies to any other state file romp moves aside as unparseable.
   `STATE/judge-auth.json` (the per-session judge-auth-down latch — see
   "Billing" above).
 - Debugging: run the judge's own code against the live store
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -124,6 +124,7 @@ def _rebind_state(path):
     CODEXDIR = STATE / "codex"
     EPIDIR = STATE / "episodes"
     _lastsid_memo.clear()   # sdk-registry reads are mtime-memoized per sid — a rebind must not serve the old root's values
+    _STORE_FAULTS.clear()   # unreadable-store episodes belong to the old root's files
     _episode_memo.clear()   # ...and so are the episode-log reads
     _head_memo.clear()      # transcript heads are immutable per path, but a rebind swaps the whole world of paths
     _namefp_memo.clear()    # names-entry content is memoized per SID against same-second mtimes — across a
@@ -839,7 +840,11 @@ def _log_judge_error(judge, fsid, err, note=None, goal=None, seg=None):
              "scratch" (the judge scratch cwd can't be made private — call skipped, never rerouted
              to a world-writable directory; see _ensure_judge_scratch), "sweep-cut" (the closer ended a
              session's walk for the pass at a FAILED call — _close_session; the note names the turns left
-             behind and the shape of the menu that died)
+             behind and the shape of the menu that died), "store-quarantined" (a goals store that
+             could not be parsed was moved aside to <file>.corrupt-<stamp> and the session started fresh),
+             "store-unreadable" (a goals store that cannot be READ — EACCES, EIO — filed once per fault
+             episode by load_goals_or_fault; inside a judge pass the raise instead reaches the pass
+             wrapper's "pass-crash" row)
       note   the evidence — reply tail, error message, exception name, or the give-up scope + re-arm
              event. Callers must pass it; an empty note means the caller has nothing at all to show.
       goal   the node id (or list of node ids) the judge was ruling on, when one exists — the feed's
@@ -2350,8 +2355,11 @@ def tasks_for(fsid, leaf, files, now):
     except Exception:
         pass
     session = parsed_session(fsid, files, now)
-    tasks = []
-    for t in _ready_tasks(session, load_goals(fsid)):
+    store, fault = load_goals_or_fault(fsid)
+    if fault is not None:
+        return []                                      # its row is filed; this session captions nothing this
+    tasks = []                                         # pass and the other sessions' captions proceed
+    for t in _ready_tasks(session, store):
         kind = t.get("kind", "work")
         text = _prompt_text(t["atoms"]) if kind == "prompt" else _unit_text(t["atoms"])
         task = {"kind": kind, "text": text, "writes": t["writes"]}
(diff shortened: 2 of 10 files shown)


lczh:
Tier: **fix** (a bug fix with tests that fail before it). A fork author cannot add labels here, so could a maintainer apply the `fix` label?

PR #1020, opened 2026-09-08
Flag, order and bell writes are refused when their state file can't be read, never written over an empty

lczh:
Flag, order and bell writes are refused when their state file can't be read, never written over an empty

Three kernel state stores shared one unsafe shape. Each reader folded ANY stat/read/JSON
fault into an empty value the same way a genuinely-missing file reads empty, and two of
them CACHED that fabricated empty under the file's real (mtime_ns, size) key. Every
read-modify-write writer then copied the empty, applied its one edit, and atomically
wrote it back -- so a transient EIO/EACCES on read, or torn bytes, erased the whole store
under a success ack, with no notice:

- session-flags.json: `_set_session_flag`/`_set_notify_session` -> every session's flags,
  including the postalServiceOff postal-isolation boundaries.
- session-order.json: `_merge_session_order` (a drag) and `_ordered` (every push) -> the
  user's saved lane/tab order, replaced with plain discovery order under no gesture.
- notify-cards.json: `_set_notify_all`/`_set_notify_turns`/`_set_notify_card` -> every
  per-card, per-session and master bell override.

(The fourth store, timeline-views.json, has the same defect and the same fix in the next
commit, kept apart because open PRs rewrite its write path.)

The fix. ONE strict reader, `_read_state_json(path, st=None)`, distinguishes the fault
classes the old `except Exception: {}` conflated: a MISSING file returns None (a fresh
install has no files, byte-for-byte as before); an UNREADABLE existing file (a stat or
read fault) raises `_StateUnreadable`; torn or non-JSON bytes are QUARANTINED aside and
then read as None, so the store starts empty ONLY after the evidence is preserved. It reads
bytes (a non-UTF-8 torn file quarantines rather than escaping as a UnicodeDecodeError) and
never caches.

The quarantine wears the shape the goal-store and ledger quarantines wear, so one
convention reads across all three: `<file>.corrupt-<utc stamp>`, a `-n` suffix for a
second collision in the same second, an `os.replace`, and a same-file re-check (inode,
mtime, size against the stat taken before the read) so an atomic publish that landed
before the re-check is never moved aside -- the caller gets a transient `_StateUnreadable`
and its next read is a fresh one. Bytes that cannot be moved (a read-only state dir) leave
the store UNREADABLE, so a writer refuses, rather than read as empty over evidence that
was not preserved. The fault text is errno + strerror only (never a path, never the
per-second stamp), so the once-per-fault-text registries below dedupe a disk that stays
broken.

Decision A: `_StateUnreadable` is a plain Exception, NOT an OSError. The WS receive loop
re-raises `(BrokenPipeError, ConnectionResetError, OSError)` as a genuine socket failure
and tears the connection down; every other exception falls to its logging arm and the next
message still processes. The review found a handler path where a read fault could escape
(the views heal, fixed in the next commit); as an OSError that escape would have dropped
the dashboard's socket, as an Exception it costs one logged line. The class carries `path`
and `fault`, and its `str()` is plain ("session-flags.json could not be read (read failed:
[Errno 5] Input/output error)"), so every surface that shows the reason shows the same
words.

The mutation path is proved. Each store gets a `_X_proved()` snapshot that reads through
the strict reader and lets `_StateUnreadable` propagate; every RMW writer snapshots
through it and refuses on a fault instead of writing over an empty. That includes the
value a click is JUDGED against when it lives in the other store: `_set_notify_card`
takes the card's default from the session's own bell (session-flags.json) and
`_set_notify_session` compares against the master (notify-cards.json). Read through the
display readers, a fault on the other store folded that default -- a mute matching the
fabricated default was then DELETED under the success path, the user's override erased
-- so both now read it proved, and a fault on either store refuses the write.

The display path never raises: build_feed / build_timeline / build_session all run under
_push's single outer try, so a reader that raised would abort every client's push and
wedge the board. Each display reader instead serves the last value this kernel read (its
cache entry; for the order, a last-known-good copy latched on every clean read or write)
or, on a cold start with nothing read yet, the empty default -- UNPROVED either way, never
cached and never latched as the new known-good, so no writer persists it -- and files
exactly ONE visible error per fault EPISODE (one stderr line + one dashboard
sync-notice); a repeated identical fault files nothing new, and a clean read clears the
episode (event-based, no timer). The sync-notice ring is the surface the views store's
stale-writer guard already uses (the shell's bell files it under its sync kind); a
dedicated kernel-fault surface is a follow-up. `_write_session_order`'s audit read is the
display read, which never raises, so it needs no guard.

Decision B: refusals answer in each surface's EXISTING contract, never a frame the pane
cannot render.

- HTTP. POST /notify-all and /notify-turns answer 200 `{"ok": false, "retryable": true,
  "error": ...}` -- the routes' own refusal shape -- never a 5xx: `romp tag` posts with
  `curl -sf`, which turns any 5xx into "kernel not reachable (is romp running?)", the wrong
  reason, and `_remote_forward` treats every non-200 as a tunnel hiccup; both read a 200
  ok:false body as the refusal it is. The shell's bell popover posts its switches through
  one `post()` helper that resolved on any 2xx and would have run the SUCCESS arm on an
  ok:false body, flipping the switch to the state the kernel had just refused; it now
  rejects on ok:false with the route's `error`, so the switch stays put and the reason
  toasts through the same `fail()` the transport errors use.

- WS. The setSessionFlag, cardNotify and reorderTabs/writeOrder arms catch the fault and
  answer the DELIVERING socket with one frame, `{"type": "settingRefused", gesture, sid,
  itemId, flag, value, text}` (`_refuse_setting`, through `_reply` -- the targeted idiom
  the settingStale stand-down and the saveFile acks use, never a broadcast). `gesture`
  names the kind ("flag" / "bell" / "order"; "views" in the next commit) so no pane
  infers it from which fields are empty; `value` is what the kernel's display path still
  paints for that flag or bell -- the value the next push carries -- so the pane repaints
  the refused toggle to it on THIS event rather than to a value it recorded at the click
  (two clicks before the first refusal made such a record wrong until the next push). The
  `warn` frame the first draft sent did neither: only the chat page renders `warn`, so a
  refused bell on the feed page and a refused lane flag on the timeline page stayed
  painted as if they had landed, their sticky latches never released, until a reload.
  Every pane that posts one of these gestures handles the frame:
  - the feed (a card bell) drops that card's `pendingNotify` latch and repaints (the paint
    key reads the latch, so the bell returns to the payload's value), toasts the text -- a
    soft refusal, nothing typed was lost, so never the must-dismiss dialog -- and files it in
    the shell's bell. The page still has NO `warn` handler; undelivered-err.test.ts's pin of
    that stays true on purpose.
  - the timeline (a lane-gear flag): both boots (timeline-boot.ts and the kernel's inline
    twin) dispatch the frame to the panel; `settingRefused` drops the sid+flag entry of
    `_pendingFlags`, repaints the lane to the frame's `value` (on the current frame's
    session and on the copy the open gear built from), shows a dismissible refusal row in
    the gear -- rebuilt in place while open, shown on the next open otherwise, retired by a
    retry -- and files it in the shell's bell.
  - the chat (a tab-menu flag) repaints the session's local copy to the frame's `value`,
    toasts the text and files it in the shell's bell.
  - reorderTabs/writeOrder: no shipped pane posts it today (the strip and the lanes write
    the viewer's own arrangement), so the frame has no latch to release; the contract holds
    for the op all the same.
  The shell's bell gets a kind of its own for these, `refused` ("not saved"; its tooltip
  says what it means; the warning-yellow chip): filing them under `warn` would have let a
  user who muted the judge's anomaly stamps mute refusals of their own changes with them.
  The error center's filter row grows from 13 kinds to 14.
  Internal callers (the order GC on every push, the bell prune on a card leaving) skip the
  pass on a fault, loud once per episode, instead of writing.

Unchanged: the ENOENT case reads empty exactly as before; a clean read reads and caches
exactly as before; every route's success shape and every success push are untouched.

Tests (module runs on this commit; the full suite is left to CI, not run on the
development machine):

- tests/test_kernel_session_flags.py (21): a flag toggle under a read fault leaves the
  file byte-for-byte unchanged and raises (on main the file is left holding only the new
  toggle, the isolation boundary gone); the display reader serves the last cached value
  over a fault and caches nothing under the file's NEW stat key (the key is bumped by a
  second write first, since a primed cache under the same key is a hit and never reads),
  serves {} uncached on a cold cache, and re-reads the real file once the fault clears; one
  notice per episode, re-armed by a clean read; the WS arm answers `settingRefused` with
  gesture, sid, flag and the painted value for both the plain setter and the tri-state
  bell, the file untouched, nothing marked dirty, nothing escaping; a session-bell click
  with the BELLS file unreadable is refused and the flags file unchanged (before: the mute
  popped and the file rewritten as {}), directly and through the WS arm; `_StateUnreadable`
  is an Exception and not an OSError.
- tests/test_kernel_order.py (31): a drag under a fault is refused with the file unchanged
  (main drops the untouched lane); `_ordered` renders the last-known order and persists
  nothing; the display reader serves the known-good on every faulted read and latches
  nothing when there is none; the WS drag arm refuses on the socket; the quarantine's `-n`
  suffix, its republished-before-the-re-check case (the good file is never moved), the
  cannot-move case (unreadable, not empty), the `os.replace`, and the stamp-free fault text.
- tests/test_notify_bells.py (44): the master bell toggle under a fault leaves the file
  unchanged (main leaves just {"*": true}); the display reader's fault behaviour as for the
  flags; the WS bell arm refuses on the socket addressed to the card, with gesture and the
  painted value; a card-bell click with the FLAGS file unreadable is refused and the bells
  file unchanged (before: the override deleted), directly and through the WS arm; POST
  /notify-all and /notify-turns under a fault answer 200 ok:false retryable with the file
  untouched and no dashboard told the switch flipped; the shell's `post()` rejects an
  ok:false body.
- tests/test_error_center.py (15): the filter row's kind count and label order take
  "not saved".
- tests/test_kernel.py: the SessionOrderStable seam injects the persisted order at
  `_session_order_proved`, the mutation snapshot `_ordered` reads.
- ui/webview/setting-refused.test.ts: the kernel frame's fields and every store-fault arm
  answering through it with its gesture (none with a `warn`); both boots dispatching the
  frame (the dispatch exercised for real); the timeline's latch release, repaint to the
  frame's value, gear row and bell filing; the feed's latch release, repaint, toast (no
  dialog); the chat's repaint, toast and filing; the `refused` kind listed, labelled,
  explained and worn, and every pane filing under it.

The tests drive the setters directly (`_set_session_flag`, `_set_notify_session`,
`_merge_session_order` + `_write_session_order`, `_set_notify_all`, `_set_notify_card`) --
internal entries, the code the WS arms and routes call -- and the real `_dispatch_ws` with
a fake client capturing `send`. Mutants checked one module at a time, every one red: each
display reader caching the fault value (or the order reader latching [] as its known-good);
each of the three WS arms with its catch deleted (the fault escapes `_dispatch_ws`); the
notify routes answering 503; the refusal frame reverting to `warn`, dropping `value`,
dropping `gesture`; the card default or the session master read through the display
reader again; `_StateUnreadable` back to an OSError; the quarantine's re-check removed; its
`-n` suffix loop removed. The extension suite ran once on the two-commit tip (typecheck
clean); the files it flagged were folded and re-run singly, as were the files the review
fold touched.

## Tier

**fix** — a bug fix with tests that fail before it.

## Notes for review

- Tests were run per module on the development machine (the touched Python modules one process at a time; the touched TS test files singly after a tests build; `tsc --noEmit` clean); the full suite is left to this PR's CI.
- This PR deliberately leaves the **timeline-views** store (tags, lenses, tag order) untouched: it has the same defect, but **#1001** (tag edits as acknowledged writes with a write sequence) and the PRs stacked on it (#997, #1007) rewrite that store's write path. A follow-up carrying the views half is ready on this branch's parent and will be re-derived onto #1001 once it lands, so the two do not collide.
- The new `refused` bell kind ("not saved") gives store refusals their own kind instead of borrowing the judge-anomaly `warn`; a dedicated kernel-fault surface for the once-per-episode state-file notices (today they ride the `sync` kind) is a follow-up.
- **#1019** (goal store) fixes a pre-existing `KeyError` in the WS `clearAll` handler that a test here also ran into; this PR does not touch that line, so the two merge cleanly in either order.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -633,6 +633,19 @@ State is written under `${XDG_STATE_HOME:-~/.local/state}/romp/`. Transcripts
 are read in place from where Claude Code writes them (`~/.claude/projects/`)
 and never copied.
 
+Three small files there hold settings you set by hand: `session-flags.json`
+(per-session flags, the postal isolation switch among them), `session-order.json`
+(the saved tab and lane order) and `notify-cards.json` (the bell overrides). A
+change to one of them is refused, never written over an empty, when the file
+exists but cannot be read; the refusal reaches the dashboard's error center
+under the `not saved` kind, with the reason, and the same change can be tried
+again. A file whose bytes cannot be parsed (a torn write) is moved aside, never
+deleted, to `<file>.corrupt-<UTC stamp>` in the same directory (a `-1`, `-2`
+suffix when two land in the same second), the store starts over empty, and an
+entry under the same kind says so. A file that cannot be read at all keeps
+showing its last-read values until it can. Nothing here needs a restart; the
+sidecars are yours to inspect or delete.
+
 ## Switches
 
 Effective immediately, no restart.
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -2278,10 +2278,22 @@ def _state_quarantine(p, st, reason):
         return "could not be moved aside: %s" % _errno_text(e)
     sys.stderr.write("romp-kernel: %s could not be parsed (%s); moved aside to %s, the store reads as empty\n"
                      % (p.name, reason, aside.name))
+    # The dashboard hears it too (review find, 2026-09-08): a quarantine resets the store to EMPTY, so
+    # every bell override, lane flag (postal isolation included) or saved lane order it held reads as
+    # a default from here on, and the stderr line alone left that looking like settings resetting
+    # themselves. One notice, under the bell's `refused` kind (a setting that did not hold, not a
+    # machine sync); no dedupe registry: the next read of this path is an ENOENT, so a corrupt file
+    # speaks exactly once. Guarded like _note_state_fault: a notice never turns a successful move
+    # into a raise.
+    try:
+        _sync_notice("%s could not be parsed and was moved aside to %s; the settings it held start over "
+                     "empty until you set them again" % (p.name, aside.name), ok=False, kind="refused")
+    except Exception:
+        pass
     return None
 
 
-def _read_state_json(path, st=None, _tries=3):
+def _read_state_json(path, st=None, expect=None, _tries=3):
     """The ONE strict reader for a small JSON state file (session-flags, session-order, notify-cards;
     timeline-views follows in its own change). Distinguishes the outcomes the old `except Exception: {}` conflated:
       - a MISSING file is legitimately empty            -> returns None (a fresh install has no files)
@@ -2293,6 +2305,11 @@ def _read_state_json(path, st=None, _tries=3):
         rename is left alone and read afresh here, bounded by `_tries` (the maintainer's fold on
         PR #1019), so the caller gets what is there now -- only a file that keeps changing under
         every read ends _StateUnreadable.
+      - valid JSON of the WRONG top-level type           -> quarantined the same way, when the caller
+        names the store's shape in `expect` (dict for the flags and bells, list for the order). Read
+        as empty instead, a list where a dict belongs was overwritten by the next writer, and a file
+        none of our writers produce is exactly the evidence worth keeping (review find, 2026-09-08).
+        None skips the check.
     `st` is the stat the caller already took (the display readers key their cache on it); without
     one the reader stats first, so the quarantine can tell the file it read from one published
     since. Reads bytes (json.loads accepts them and auto-detects the encoding), so a non-UTF-8 torn
@@ -2313,14 +2330,23 @@ def _read_state_json(path, st=None, _tries=3):
     except OSError as e:
         raise _StateUnreadable(path, "read failed: %s" % _errno_text(e))
     try:
-        return json.loads(raw)
+        data = json.loads(raw)
     except (ValueError, UnicodeDecodeError) as e:
         why = _state_quarantine(path, st, "invalid JSON: %s" % e)
         if why is None:
             return None                              # the bytes are preserved; the store starts empty
         if why == _STATE_REPLACED and _tries > 1:    # a peer published since our stat: read what is there now
-            return _read_state_json(path, None, _tries=_tries - 1)
+            return _read_state_json(path, None, expect=expect, _tries=_tries - 1)
         raise _StateUnreadable(path, "invalid JSON; %s" % why)
+    if expect is not None and not isinstance(data, expect):
+        why = _state_quarantine(path, st, "wrong shape: %s where %s was expected"
+                                % (type(data).__name__, expect.__name__))
+        if why is None:
+            return None
+        if why == _STATE_REPLACED and _tries > 1:
+            return _read_state_json(path, None, expect=expect, _tries=_tries - 1)
+        raise _StateUnreadable(path, "wrong shape; %s" % why)
+    return data
 
 
 # Last-known-good session order (session-order.json has no mtime cache): served over a transient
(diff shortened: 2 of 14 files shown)


lczh:
Follow-up commit ad57da06: the save-step lesson from #1019 applied here. Before it, an `OSError` from the setter's write (a full disk) escaped every WS arm to the receive loop's teardown and dropped the client, and the three routes answered 500. Now a `_write_state_json` door files once per fault episode, every arm answers the same `settingRefused` frame with a "could not be written" text, the routes answer 200 ok:false retryable, the push-path writers contain the fault, and a declined quarantine re-reads the new bytes. The TypeScript pins in `setting-refused.test.ts` now anchor on the two-exception catch, with a length guard so a fallback to a far-away catch cannot pass silently.

PR #1022, opened 2026-09-08
Boot names credential-shaped variables in the kernel's environment that reach sessions

ksarma:
The SDK transport merges the kernel's `os.environ` under `options.env` and hands the whole thing to each session's CLI, and romp takes only the login tokens it claims at boot (`startup_auth_env`) out of it. Since #1128 romp holds no key of its own, and a retired provider name in the environment stops the kernel before a backend exists (`credentials.check_boot_environment`). So any other variable ending `_API_KEY` or `_TOKEN` in the manager's environment, a second provider's key or a leftover token, is inherited by every session and every shell it spawns, with nothing in the log to say so. Under #1128, where every box lets Claude Code resolve its own credential, this is a place a stray credential reaches sessions unseen.

Rebased onto main after #1128 merged (2026-09-09): the boot line now excludes only romp's own control token, and the tests stage variables directly. The boot line's copy is unchanged.

## What this adds

- `env_credential_names(environ)`, a pure helper returning the credential-shaped names in a given environment (excluding only romp's own control token: a login token still present when it runs did reach sessions and is named, and the call's place after the boot claim is what keeps one off the line), sorted, names only.
- A once-per-process boot notice (`_note_env_credential_names`, called from `SdkBackend.__init__` after the problem ring exists and after `startup_auth_env` has claimed the login tokens, so it names what sessions actually inherit) that logs those names for the kernel's own environment. It is an informational line, not a problem-ring entry: an installation may put such a key there on purpose. No value is logged (each is read only to skip empty variables), and a box whose environment carries none says nothing.

## Scope

The process environment only. Reading the systemd unit, its drop-ins, or the launchd plist for the same names (the open question from #1012) is a larger change and is deliberately left out here; it can follow if you want it. Adjacent to #1128's boot check, which stops the kernel on a retired provider name or one of 1Password's names in `service.env` or in the environment; this names the other credential-shaped variables, the ones the kernel tolerates and passes on. The helper's shape still includes 1Password's names through the same classifier the boot check uses (`credentials.is_op_env_name`), so the two agree on what an op name is and a backend built outside `kernel.main()` reports them too. The two do not overlap in code.

## Tests

`tests/test_env_credential_names.py`: the pure helper (which names are named, the control-token and empty-value exclusions, names-not-values); the notice in isolation (says once, informational, no value); and the `__init__` wiring on a constructed backend, including that 1Password's names are named (romp claims none of them; the boot check, not the helper, is what keeps them out of a kernel started by `kernel.main()`) while the login tokens `startup_auth_env` claims are not (the call order in `__init__`; moving the notice above the claim turns the test red). Variables are staged directly in `os.environ`; the module reads nothing of the retired provider code. Each headline claim fails when its mechanism is removed (verified by mutation). Test values are plain synthetic strings, no credential-shaped literals.

Tier: feature


ksarma:
Ordering with #1128: this PR conflicts with it in kernel/sdk_backend.py (three regions) and docs/reference.md, and its test module stages the provider variables through keysource, which #1128 deletes. #1128 first is the cheaper sequence: I rebase this the same day, the exclusion set shrinks to romp's own control token (the login tokens and a key are then named if they ever reach the kernel's environment, which is what the line exists to report), and the tests stage variables directly. The other order would leave #1128 with the same conflicts plus a runtime break in this helper, so please take #1128 ahead of this one.

henrypinkard (approved):
Approved on head a703c839 on the maintainer's word (feature-review session).

PR #1023, opened 2026-09-08
Docs: a key from a secret manager through apiKeyHelper, with none held by romp

ksarma:
Since #1014, a session with no key source configured uses Claude Code's own credential resolution. That makes it possible to supply a secret-manager key entirely through Claude Code's `apiKeyHelper`, with romp holding nothing. This documents that route in the reference.

## What this adds

A `####` subsection after "Existing API keys and Claude login", under "Service environment and credentials", in `docs/reference.md`:

- What romp shows on such a box, so the setup is not mistaken for a failure: the Billing picker offers no API-key choice, a session's Billing row reads `Login (CLI reports API key)`, and `romp keyswap --cycle` skips every session as billing the login.
- A helper script that prints the key (1Password shown as one example, with `op read --no-newline`; any secret manager whose CLI prints a key works the same way), and the credential that CLI needs under a headless service: for 1Password, `OP_SERVICE_ACCOUNT_TOKEN=` in `service.env`, which romp leaves in the sessions' environment when no reference is configured, since the helper needs it there.
- The `settings.json` entry.
- A key-free `service.env` with `ROMP_EXPECTED_AUTH=key`, and the `service.env.source` marker removed if a 1Password reference was ever selected on the box (with the marker in place the removed reference stays an error, not an absent source).
- One restart. Rotation is then a change in the secret manager alone, since Claude Code re-runs the helper on its own refresh interval.

One caveat is called out: the kernel's own model-catalog refresh, the one direct API call romp makes on such a box, reads only a key source romp holds (or an `ANTHROPIC_AUTH_TOKEN` in the service's environment), so on a helper-only box it has no credential: romp serves its built-in list, or the last one it cached, and the kernel log says so at each refresh attempt.

Docs only; no code or test change. `mkdocs build --strict` passes, and the rendered section is one ordered list of four steps.

Tier: tests-only

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -510,6 +510,60 @@ Plaintext keys remain supported for compatibility, but do not meet policies
 that require secrets to be fetched from 1Password at runtime. File permissions
 do not change that distinction.
 
+#### A key from a secret manager, with none held by Romp
+
+If your key lives in a secret manager and you would rather Romp not hold it,
+configure no key source in Romp and point Claude Code's
+[`apiKeyHelper`](https://code.claude.com/docs/en/settings-reference#apikeyhelper)
+at the secret manager instead. Romp passes a key to a session only when it has
+one itself, so with no source configured every session and every API-key-billed
+judge call uses Claude Code's own authentication. Romp cannot see that key: the
+Billing picker offers no API-key choice, a session's Billing row reads
+`Login (CLI reports API key)`, and `romp keyswap --cycle` skips every session
+as billing the login.
+
+1. Write a script that prints the key from your secret manager, and make it
+   executable. With 1Password, for example:
+
+        #!/bin/sh
+        exec op read --no-newline "op://<vault>/<item>/credential"
+
+    The helper runs inside each session's Claude Code process (and each judge
+    call), a child of the service, so the secret manager's CLI needs a
+    credential that works without a desktop app: for 1Password,
+    `OP_SERVICE_ACCOUNT_TOKEN=` in `service.env`, scoped to the one vault. With
+    no reference configured Romp leaves that token in the sessions'
+    environment, where the helper needs it and where an agent's shell can read
+    it. Any secret manager whose CLI can print the key works the same way.
+
+2. Point Claude Code at the script in `~/.claude/settings.json`:
+
+        { "apiKeyHelper": "/path/to/anthropic-key.sh" }
+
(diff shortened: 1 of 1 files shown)


ksarma:
Merging as shipped. One wording fix goes into a follow-up.

Reviewed inline: The diff touches only `docs/reference.md`, 54 added lines of prose plus two placeholder snippets, so the tier fits. Each claim was checked against main. With no key source configured, `_options` injects nothing and restores the startup login env. The judge's `_key_source_unconfigured` path does the same. Sessions and key-billed judge calls therefore run on Claude Code's own credential resolution. `_auth_avail` offers the API-key choice only when `work_key_configured`. `cycle_key` returns login for an unconfigured source, and the CLI prints the skipped line. `claim_op_env`, `strip_op_env`, and the manager's `withoutOpCredentials` leave `OP_SERVICE_ACCOUNT_TOKEN` in the sessions' env when romp holds no reference. `select_source` turns a lingering `op` marker into an error source, and that memory is process-local. The single restart is therefore the right step. `ROMP_EXPECTED_AUTH=key` quiets `_note_auth_source` for a keyed landing. `_models_api_credential` reads only romp's key source or `ANTHROPIC_AUTH_TOKEN`. `_refresh_model_catalog` logs the seed/cache fallback at boot and once per unknown model id. Claude Code's docs confirm the apiKeyHelper anchor and the TTL variable. The diff has placeholders only, no private strings, and no `fleet`.

- `docs/reference.md:522` (follow-up): the Billing row does not read `Login (CLI reports API key)`. Since T124, `render.ts` renders the intent/live disagreement as a row that opens with a ⚠ glyph. The row says Login was picked but the CLI reports the API key, and that the session bills that. `ROMP_EXPECTED_AUTH=key` quiets only the kernel log line, not the row. The same stale wording sits at line 253 (from before T124) and in the PR body. Correct both lines to describe the ⚠ row and say it is the expected steady state for an unpicked session on a helper-only box.
- `docs/reference.md:555` (nit): "the one call Romp makes itself" is broader than the point being made. The kernel also fetches the price feed and posts web push without a credential. The following sentence, that nothing else on such a box needs a key, is the accurate one.

`mkdocs build --strict` at c150d23f exits 0. The rendered section is one ordered list of four items, and both internal anchors resolve. No test file in the diff. CI: all six required checks are green at c150d23f.

Review run by the maintainer's automated review session; verified against head c150d23f.

PR #1026, opened 2026-09-08
T245: scrollbar message notches follow a turn that grows after the paint, and read the real scrollbar when everything is rendered

henrypinkard:
## Summary
A blue user-message notch sat well below the scrollbar thumb while the message it marked was on screen, in a session whose last turn renders dozens of inline figures.

**Cause.** The notch and comment-rail frame is a virtual prefix-sum over the rendered units' measured heights, refreshed only inside a paint, and the paint ran only on scroll, resize and a few render-side callers. Inline figures are lazy, async-decoded images with no intrinsic size before load, so every figure loading after the paint grew its turn and `scrollHeight` with no repaint: the thumb moved, the notches kept the stale frame. Reproduced in the served harness: a rendered turn grown by 1500 px after the paint left the notch at 78 px while its message's true position moved to 31 px.

**Fix, keyed on the missing event ("a rendered unit changed height").**
- One `ResizeObserver` per view element re-runs the shared rAF paint; notches and comment-rail ticks share the frame, so one observer covers both. Disconnected when a view is dropped. No timer, no per-image hook.
- With every unit rendered (no spacers) the frame is the scrollbar's own truth: each unit's real middle in scroll space over `content.scrollHeight`. The unit-height sum omits the gaps between turns, so it sat 27 px off even on a fully rendered conversation. The virtual prefix-sum stays the fallback while spacers hide units; pure scrolling still moves nothing.

**Side issue checked.** A landed slash command reaches the chat as a `cmdGesture` event, not a user event, so the notch painter never sees its expanded wrapper form; the `SLASH_CMD_RE` skip covers the raw echo. No change.

## Tests
- `tests/test_scroll_marks_growth.py` — drives the real /chat page; red first (before growth off by 27 px, after growth off by 47 px and unmoved); skips loudly without a browser
- `ui/webview/scroll-marks.test.ts` — the observer, its disconnects, the exact fully-rendered frame; the virtual branch's no-scrollTop pin scoped to that branch

### DIFF
--- a/tests/test_scroll_marks_growth.py
+++ b/tests/test_scroll_marks_growth.py
@@ -11,9 +11,16 @@ smaller frame. Nothing keyed the paint on "a rendered unit changed height".
 
 The executed guard here drives the real /chat page: after the first paint a rendered turn GROWS (a block
 appended to the last turn — the same thing a late figure does), and the user message's notch must follow
-the scrollbar's truth (red on main: the notch stays where the stale frame put it). Skips LOUDLY without the
-extension deps or a Playwright browser (CI installs none); the CI-safe pins ride
-ui/webview/scroll-marks.test.ts. All fixtures synthetic.
+the scrollbar's truth (red on main: the notch stays where the stale frame put it).
+
+Second executed guard (review of the fix, 2026-09-08): with every unit rendered the frame reads the real
+scrollbar, and that must hold when ONE UNIT OWNS SEVERAL .turn NODES — compact mode tags an expanded tool
+group's child turns, and any absorbed cue, with their unit's data-unit. The first cut gated the exact frame
+on the NODE count equalling the unit count, so opening a tool group silently dropped the frame back to the
+virtual sum (measured 11px off on a five-unit page) and the notch changed basis on every expand/collapse.
+
+Both skip LOUDLY without the extension deps or a Playwright browser (CI installs none); the CI-safe pins
+ride ui/webview/scroll-marks.test.ts. All fixtures synthetic.
 """
 import json
 import os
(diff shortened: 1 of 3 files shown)


ksarma:
Recommend one fixture fix before approval. The runtime change is correct.

Reviewed inline: the ResizeObserver is created in `ensureView`'s create-once branch, so each view element gets one observer for its lifetime, not one per render. Both view-removal sites (the only two `views.delete` calls in `render.ts`) disconnect it. Its callback is `scheduleRailSticky`, gated by the `railStickyPending` rAF flag. The paint writes only body-attached overlays plus visibility on absolutely positioned time markers. It never changes the observed element's box, so it cannot re-trigger itself. The exact-frame branch reads `scrollTop` plus each unit's client-rect offset over `content.scrollHeight`. `#content` has padding and no border, so that is the scroll space `scrollHeight` measures. In a browser the notch lands within 3 px of the geometric oracle before and after a 1500 px growth. The virtual prefix-sum remains the fallback when spacers hide units. The diff has no `SLASH_CMD_RE` hunk, so that side issue is a claim only. The privacy scan of the 233 added lines is clean.

- `tests/test_scroll_marks_growth.py:111`, blocker. The fixture writes the transcript under `cwd.replace("/", "-")`. The kernel resolves it with `re.sub(r"[^A-Za-z0-9]", "-", realpath(cwd))` (`kernel/judge.py:5663`, `kernel/sdk_backend.py:284`). Any dot or underscore in the temp path puts the transcript where the kernel never looks. No notch renders, the driver times out, and on a browser-equipped machine the test fails rather than skips, at head and at main alike. Both mkdtemp suffixes in the lab path draw from `a-z0-9_`, so about 35% of runs hit this even under `/tmp`. With that one line on the kernel's encoding, the test fails at main with exactly the stated 27 px and 47 px offsets. It passes at head. Three sibling tests at main carry the same line; a follow-up sweep is welcome.
- `ui/webview/render.ts:2148`, follow-up. The guard `nodes.length === unitTotal` fails whenever a unit owns more than one `.turn` node. That happens with an expanded tool group (`.tg-child`), an expanded retry group, or an absorbed cue. Opening a group in compact mode drops the frame from exact to virtual and shifts every notch (reproduced: 101 to 74 px and back on toggle). A session carrying one cue never gets the exact frame. Key the branch on no spacers instead, and take each unit's first node or first-to-last span.
- `tests/test_scroll_marks_growth.py:78`, nit. The driver sleeps 400 ms after growing the turn and measures once. A bounded `waitForFunction` on the notch's `style.top` would key the wait on the repaint.

Tests: node red-first confirmed. With the PR's `scroll-marks.test.ts` against origin/main, exactly the two T245 pins fail (8 tests, 6 pass, 2 fail). At head 8/8 pass, and the full vscode-extension suite is 3090/3090. The browser test ran here (Playwright chromium present) and failed at head as shipped, because of the fixture's path encoding. With the encoding corrected it is red at main with the body's numbers and green at head in under 3 s. CI: all six required checks are green at head. The browser test skips on CI because no browser is installed, so CI covers this fix through the source pins only.

Review run by the maintainer's automated review session; verified against head 66f09764.

PR #1030, opened 2026-09-08
Chat: a queued message is editable until the session takes it

alowet:
## What

A message sent to a busy session waits as a dashed bubble — parked while the session compacts, held in the SDK backend's queue behind a running turn, or at the optimistic "sending…" stage. Until now the only way to change it was the ✕: pull it back into the composer and send again, at the back of the queue. It is still the user's message while it waits, so it is now **editable in place**.

## Design points

- **The control.** The queued bubble wears a ✎ beside its ✕, at the same three stages and behind the same recall gate (`cancelable`); romp's own notices and slash commands are not edited (a command is cancelled and typed again). Delegated through the body delegate like the ✕, so a mid-press rebuild cannot eat the click.
- **The editing surface is the composer**, under an "Editing queued message — send replaces it in the queue" pill — the rewind edit's grammar, one level deeper. Esc or the pill's ✕ leaves the message as it was. A rewind edit and a queued edit yield to each other; one edit at a time.
- **In place, kernel-side.** Send posts `editQueued {md, text, idx|park}` (body-only at the optimistic stage). `_edit_parked` swaps the send's words in its FIFO slot under the queue lock — same echo author, the in-flight head refused as too late, exactly as `_cancel_parked` does. `_edit_backend_queued` → `SdkBackend.edit_queued` → `SdkSession.replace_queued` re-verifies the exact old text under the session lock (a shifted index re-locates by body) and re-words the message's optimistic echo so the live tail shows the edited words.
- **A follow-up keeps its context.** `_replace_followup_body` keeps the goal quote and the trailing romp markers (the inverse of `_split_followup`'s body read), so an edited follow-up still files under its goal.
- **Authoritative result frame.** Every arm answers `editResult`; `ok:false` (the message left the queue meanwhile) reverses the optimistic repaint, hands the typed words back to the composer, and repaints the queue from the kernel. Nothing is lost and nothing is sent twice. `editQueued` joins `ID_OPS` so it routes to the owning kernel across linked machines.

## Tests

- `test_sdk_backend.py::PendingQueue`: in place + order kept, `expect` re-locates under the lock, a miss touches nothing, the echo is re-worded.
- `test_kernel_send_park.py::QueuedEdit`: `_edit_parked` in place, body re-locate, gone and in-flight refusals, command-chip and empty-body refusals, follow-up wrapper kept end to end; `_edit_backend_queued` drift guard; the three drive arms answer `editResult`; `ID_OPS` pin.
- `ui/webview/queued-edit.test.ts`: the ✎ and its gate, the `qedit` delegate, the pill, the send path, `editResult` handling, the optimistic repaint, CSS.
- Guide: "A message that has not gone yet" under The chat.

Targeted Python suites (send-park, pending-queue, kernel queue/cancel, parked-ops lock + liveness), the extension typecheck and all node tests pass locally.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -51,6 +51,16 @@ A pull request number in a message, a card, or a note (`#123`, `PR #123`, or
 directory has as its `origin` remote; when that remote is not on GitHub, the number stays
 plain text.
 
+**A message that has not gone yet.** Send to a busy session and your message waits as a
+dashed bubble under an hourglass until the session takes it — while it compacts, while a
+turn runs, or in the beat before the kernel confirms the send. Until then it is still
+yours: the **✕** in its corner pulls it back into the composer, and the **✎** beside it loads
+the text into the composer under an editing pill, so you can change your mind without
+losing your place in the queue. Send replaces the message where it was, a follow-up keeps
+its context, and Esc or the pill's ✕ leaves it as it was. If the session takes the message
+before the edit lands, romp says so and gives your edited words back to the composer
+rather than sending them twice.
+
 **Opening a markdown document.** A markdown link in the chat opens in the file viewer,
 rendered, with **Raw** one click away — a path on the session's machine, or a link to a
 file served from the dashboard's own address (a published report, an evidence doc). Figures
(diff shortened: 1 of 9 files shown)


henrypinkard (approved):
Approved on head ad5c5405 on the maintainer's word (feature-review session).

henrypinkard (approved):
Approved on head 2507093b on the maintainer's word (feature-review session).

henrypinkard (approved):
Approved on head 308c8ea1 on the maintainer's word (feature-review session), after the branch update onto main.

PR #1038, opened 2026-09-08
Kernel and bus request bodies are authorized, bounded and type-checked before they act

lczh:
Three shapes of the same defect on the kernel's HTTP door and the postal bus, verified on `main`:

- **The body was read before the caller was authorized.** `do_POST` read whatever Content-Length the request declared, inside a bare `except: pass`, and only then ran `_authorize`. A token-less loopback or tailnet client could make the kernel allocate and pin a handler thread per oversize POST; a duplicate or non-decimal Content-Length was silently read as no body.
- **Non-object bodies were 500s with tracebacks.** The session-management routes did `(b or {}).get(...)` on the parsed body. A JSON string, number or non-empty array is truthy, so `.get` raised into the catch-all and the client got a traceback naming absolute paths for what is a malformed request. The bus's `/send`, `/peer` and `/peer-exchange` had the same shape.
- **Request flags were coerced with `bool()`.** `bool("false")` is True, so a string enabled auto-nudge, file editing, the master bell and auto-update, paused retries across every session, made a directory, or deleted a tag.

## What changes

**Authorize, then read, within bounds.** `Handler._read_post_body` runs only after `_authorize` passed. Transfer-Encoding is refused with 411 (the body is delimited by Content-Length alone; chunk framing would otherwise be read as a zero-length body and its bytes as the next request); more than one Content-Length, or a value that is not a plain decimal of at most twenty digits, is 400; a declared length past 1 MiB is 413 before a byte is read; a body that arrives shorter than announced (a dead client, a half-close) is 400 naming the shortfall, never "no body". Every refusal closes the connection and says so in the header, because the unread bytes would be parsed as the next request on a keep-alive socket. An absent Content-Length is an empty body, which is what the bodiless POSTs (`/tick` from the hook, `/restart`) send. A 403 now closes too, since the body is unread.

**A body is an object or a 400.** `_json_object_body` answers `(body, error)`: empty is `{}` (the route then names its missing fields), undecodable is "body is not JSON", anything that is not an object is refused naming what arrived through a bounded echo that clips inside the quotes so the error text is always well formed and short. Fifteen routes use it: `/update-dismiss`, `/notify-all`, `/notify-turns`, `/new`, `/fork`, `/rename`, `/move`, `/color`, `/watch-pr`, `/watch`, `/tag` and `/group`, `/working`, `/deliver`, `/picker-check`, `/walk-root`, `/redial`. The bus's `_body` is the kernel gate's twin (411, 400, 413, short read, non-object) and closes on refusal.

**A flag is a JSON boolean or a refusal.** `_as_bool` passes `true`/`false`, keeps the handler's previous default when the field is absent, and refuses anything else naming the field. HTTP sites answer 400; WebSocket sites answer on the socket that sent the frame: the settings ops with a `warn` frame, `editTag` with its own `tagEditFailed`, and `setSessionFlag` and `cardNotify` with the `settingRefused` frame the feed and timeline pages render. On the bus, `tracked` on `/send` and the MCP `send_message` tool, and `up` on peer updates. Every shipped sender already posts real booleans; the refusal reaches only a client that did not.

## Judgment calls

- The 1 MiB cap: every kernel POST is a small JSON control request. Attachments ride the WebSocket, whose own limit is 80 MB. Each shipped client was read: the hook's `/tick` is bodiless, the CLI posts `curl -d` bodies bounded by argv, postal and the kernel's own forwards set Content-Length through urllib and http.client, and the extension's only POST is the manager's bodiless `/ensure`.
- Transfer-Encoding is 411 rather than 413: the request is not too large, it is unframed.
- Absent flags keep their old defaults; only a present non-boolean is refused. The one default that shifts is `mcpAction`, where an explicit `null` now reads as the default (enabled) instead of False; the UI never sends null.
- MCP `send_message`'s `tracked` is the one gated site whose sender is a model rather than shipped code. The refusal text names the fix, so one call recovers.
- The bus caps at 1 MiB too. Nothing shipped produces a mail body that large; a producer-side cap on the `/deliver` banner (the bus concatenates a recipient's pending mail) is a follow-up.

## Unchanged, and residuals

`/restart`'s body parse is owned by #1027 (`_restart_scope_from_body`); its `fleet` flag is on that PR's side, and the two compose. `/judge-settings`' `propagate` is kernel-to-kernel and forwarded bodies strip it. The request-scoped hints left as truthiness are listed in the commit body (`floating`, `nudge`, `manual`, `boot`, `ok`/`keep`, `live`, `generate`, `hidden`); nothing persists from them. `do_POST`'s catch-all 500 is unchanged; the routes above no longer reach it for these inputs.

## Tests

`tests/test_kernel_auth_hardening.py` drives the real `do_POST` through a fake socket and a real loopback server: an unauthenticated 10 MB POST answers 403 with zero reads; an oversize declared length is 413 before any read and the next in-bounds request is served; non-decimal and duplicate Content-Length, Transfer-Encoding, a 5000-digit length, a short body over the wire, and the bodiless `/tick` (a guard of the branch a mutant escaped). `tests/test_rename_route.py` posts `[]`, `"x"`, `1`, `null`, `[1]`, malformed JSON and a 100 KB string to all fifteen routes. `tests/test_tag_route.py` and `tests/test_kernel_session_flags.py` send non-boolean flags to every gated HTTP and WebSocket site and assert the stub behind each never ran and the right frame came back. `tests/test_postal_token.py` and `tests/test_postal_peers.py` cover the bus gate and its flags. `ui/timeline-flags.test.ts` pins the boolean sender and the kernel gate. Mutants (the read moved back before auth, each gate check removed, `bool()` restored, a route's guard reverted, the bus's refusal swallowed) each fail a test.

Module counts on this tree, one runner at a time: `tests/test_kernel_auth_hardening.py` 38; `tests/test_rename_route.py` 7; `tests/test_tag_route.py` 99; `tests/test_kernel_session_flags.py` 39; `tests/test_notify_bells.py` 53; `tests/test_postal_token.py` 16; `tests/test_postal_peers.py` 26; `tests/test_new_session_dir.py` 47; `tests/test_setting_gesture_order.py` 72; `tests/test_tracked_delegation.py` 13; `tests/test_perf_stats.py` 41; `tests/test_op_credential_boundary.py` 5; `ui/timeline-flags.test.ts` 10 of 10. Full suite left to CI.

Rebased over #1027 (merged): its pre-auth read block, which recorded a short read for `/restart` alone, is superseded by the gate, which refuses a short or unparsable body with a 400 before any route runs, so `/restart`'s scope parser now receives no body error. The gate's refusals use #1027's vocabulary ("body could not be read: read N of the M bytes Content-Length announced", "… is an invalid literal for a byte count"), so its `/restart` test holds unchanged.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -52,6 +52,21 @@ These are for scripting and for agents rather than daily use:
 | `romp resume <id> [--name <n>] [--detach]` | Resume one exact conversation by UUID |
 | `romp refresh --quiet` | Refresh at the next quiet window instead — waits for sessions to finish their turns (15-min backstop) |
 
+Raw `POST` callers, anything that talks to the kernel's routes directly rather
+than through `romp`, follow one body contract, and the postal bus's own routes
+share it. The request carries the serve token (`X-Romp-Token`, or `?token=`)
+and is authorized before its body is read. The body is delimited by
+`Content-Length` alone: no `Transfer-Encoding` (411), the header once and a
+plain decimal (400 otherwise), and at most 1 MiB (413 beyond that, refused
+before a byte is read). A body that arrives short of its announced length is
+400, one that stalls for 30 seconds is 408, and every refusal closes the
+connection. The body is a JSON object; an array, string, number or `null` is a
+400 naming what arrived, echoed bounded and well formed. A flag field
+(`delete`, `on`, `mkdir`, `tracked`, and the like) is a JSON boolean: `true`
+and `false` apply, an absent field or an explicit `null` reads as the route's
+default, and anything else (the string `"false"`, `0`, `1`) is a 400 naming the
+field, with nothing acted on.
+
 `--env` gives one session its own environment, so two sessions in the same
 directory can run with different toggles (a `FEATURE_FLAG=1`, a `CLAUDE_CODE_*`
 switch) without editing the directory's `.claude/settings*.json`, which reaches
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -2745,7 +2745,7 @@ def _clear_state_fault(path):
     _state_fault_seen.pop(str(path), None)
 
 
-def _refuse_setting(client, exc, what, gesture, sid="", item_id="", flag="", value=None):
+def _refuse_setting(client, exc, what, gesture, sid="", item_id="", flag="", value=None, log=None):
     """A dashboard gesture (a lane/tab flag, a card bell, a drag, a view change) that this kernel
     REFUSED because the store it edits could not be read -- or, since the maintainer's fold on PR
     #1019, WRITTEN (_StateUnwritable: the publish itself failed): one stderr line, and the refusal answered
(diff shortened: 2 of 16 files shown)


lczh:
The four Python jobs fail on two tests this branch does not touch: tests/test_fleet_restart.py::AnAskStaysOnThatPeer stubs the planner with a one-argument lambda (from #1027) while the sweep now passes `head=` (from #1025), so main itself has failed them since 2e9d4492. #1047 fixes the stub; after it lands I will rebase this branch to re-run CI.

PR #1043, opened 2026-09-08
Serve tokens are minted once under a lock, born 0600, and never rotated by a read fault

lczh:
The serve token is the one credential every romp client holds a copy of: `bin/romp` and the hooks read the file, the bus and each session's MCP process load it at their own start, peers fetch it at attach. Two loaders read it, the kernel's `_load_token` and the bus's `_load_serve_token`, both at import time. Verified on `main`, both treated an existing token they could not read as absent and **minted a new one**: an EACCES or EIO on the read fell through to the mint, a silent rotation that stranded every client holding the old token with nothing to say why. The kernel's mint opened the live file with `O_TRUNC` in place, so a concurrent reader saw an empty file, and it took no lock, so two starters each wrote their own. The bus's mint wrote the file and then chmod'd it, a window at the umask's mode.

## What changes

One shape in both loaders, `_serve_token_read_or_mint(f, who)`:

- **A missing file is the only mint trigger.** Any other read or stat fault raises a `RuntimeError` naming the path and the errno, and saying that romp did not replace the token. At import that refuses to start the daemon, which is visible and repairable, while the token every client holds stays valid. A non-UTF-8 token reads as the same refusal.
- **The mint is atomic and born 0600.** The value is written to `serve-token.<pid>.tmp` opened `O_EXCL` at 0600, length-checked, fsynced and `os.replace`d onto the path. The live path is never opened for writing. Temps left by a crashed earlier starter are removed under the lock.
- **One mint between the kernel and the bus.** A sibling `serve-token.lock` (0600) is `flock`'d around read-or-mint, so the two daemons booting together mint once. When the lock itself cannot be taken, an existing non-empty 0600 token is returned as is, since there is nothing to mint or tighten; every other case is a refusal, because minting or tightening without the lock is the race this closes.
- **A loose token is tightened**, not replaced: a non-empty token that is not 0600 is chmod'd under the lock and said on stderr. An empty or whitespace-only file is a torn earlier mint (no client can hold a token that was never written) and is minted over, said on stderr.

The bus carries its own copy of the function with a keep-in-sync marker, the way the repo already duplicates its kernel/bus routines, because the bus imports nothing from `kernel/` by design; a test compares the two bodies as syntax trees with docstrings stripped, so any divergence fails.

## Judgment calls and residuals

- Fail-closed at import is the right failure: a silently rotated token strands every client and peer; a refusal to start is one message the operator sees.
- `_load_token` is also called at runtime from the check-in handshake, inside a broad `except` that turns any exception into a failed handshake retried each pass. That site is untouched: a refusal there is a retried handshake rather than a loud line, and `main`'s behavior there was worse (a silent mint of a token the kernel itself does not serve). Follow-up material.
- A starter blocked on the lock waits for the holder with nothing said. Blocking `flock` is the event-based mechanism; a non-blocking probe that announces the wait would add visibility. Left as a follow-up.
- `O_EXCL` on the temp and the fsync are pinned by source text and the tree-identity test, not behaviorally.

## Tests

`tests/test_kernel_serve_token_mode.py`: six threads racing the mint read one token and the lock file exists 0600; the token is 0600 from its first byte (an interposed `os.replace` and `os.open` under a 022 umask, and the live path is never opened for writing); an unreadable existing token is a refusal naming the path and its bytes are unchanged; a non-UTF-8 token is the same refusal; a 0644 token is tightened and its value returned; an empty file is minted over and said; a directory at the lock path is a refusal except for a good 0600 token; a crashed other pid's temp is removed by the mint; the two copies are identical as syntax trees. `tests/test_postal_token.py`: the bus twins of birth, race, unreadable, tighten, lock failure and empty file. On `main`, nine of the kernel module's tests and the bus's new tests fail (`6 != 1`, `RuntimeError not raised`, `420 != 384`, the missing helper). Source mutants (the read fault turned back into a mint, the flock removed, the lock-failure arm ignoring the mode, the tighten removed, in-place writes, `O_EXCL` dropped) each fail at least one test, and the postal copy's policy arms now fail their own.

Module counts on this tree, one runner at a time: `tests/test_kernel_serve_token_mode.py` 15 passed (2 subtests); `tests/test_postal_token.py` 13; `tests/test_postal_self_host.py` 9; `tests/test_ledger_unproved_reads.py` 20; `tests/test_token_login_page.py` 8; `tests/test_mention_pins.py` 8. On `main` the kernel module fails 12 of 15 and the bus module 6 of 13, each with the first error the commit body names. Full suite left to CI.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -551,6 +551,15 @@ user account). Local tools (the CLI, hooks, the bus, the editor extension) read
 that file and send it automatically, so you never type it. Only liveness probes
 (`/healthz`, `/version`, `/busy`, and the bus's `/ping`) are exempt.
 
+The kernel and the bus mint that file when it is missing, one mint between them
+under a sibling lock file, `serve-token.lock`. An existing token is never
+replaced: a file left looser than `0600` is tightened at the next start (its
+value is kept, so every client stays valid), and a token that exists but cannot
+be read, or a symlink at that path, refuses to start instead of minting a
+replacement nobody else holds. Under the service that refusal repeats in
+`manager.log` every 10 seconds until you repair the file; the kernel then comes
+back on its own.
+
 A browser cannot read that file, which is why the link `romp` prints carries the
 token in it. The first visit trades it for a year-long cookie, so the bare
 `http://127.0.0.1:29855/` works from then on; `romp url` prints the link again
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -12,7 +12,7 @@ zero protocol change at switchover. WS is hand-rolled on the stdlib socket (no d
 Run:  bin/romp-kernel   → opens http://127.0.0.1:29855
 """
 import math
-import contextlib, json, os, queue, random, re, signal, socket, sys, time, threading, traceback, base64, bisect, errno, hashlib, hmac, struct, subprocess, shutil, shlex, http.client, uuid, tempfile, stat, gzip, collections, functools
+import contextlib, json, os, queue, random, re, signal, socket, sys, time, threading, traceback, base64, bisect, errno, hashlib, hmac, struct, subprocess, shutil, shlex, http.client, uuid, tempfile, stat, gzip, collections, functools, fcntl
 from pathlib import Path
 from datetime import datetime, timezone, timedelta
 from importlib.machinery import SourceFileLoader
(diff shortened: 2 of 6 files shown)


lczh:
The four Python jobs fail on two tests this branch does not touch: tests/test_fleet_restart.py::AnAskStaysOnThatPeer stubs the planner with a one-argument lambda (from #1027) while the sweep now passes `head=` (from #1025), so main itself has failed them since 2e9d4492. #1047 fixes the stub; after it lands I will rebase this branch to re-run CI.

PR #1044, opened 2026-09-08
T252: a message sent mid-turn appears where it was sent and stays there

henrypinkard:
The user's decision (2026-09-07): a message sent while the session is mid-turn must simply appear where it was sent and stay there. The merged mid-turn-sends change explained the bubble's move with a "joined mid-turn" header and a jump/✕ cue; both are gone.

**Change**
- The pending bubble is drawn at its **send position** from the first paint: right after its anchor (the last stable kernel event at the press). Steps that stream in while the CLI holds the send land below it; the absorbed atom the kernel places at the send time replaces it in the same slot.
- `send-pending.ts`: `injectionGroups` (slot per send, grouped per anchor, spliced bottom-up); `Reconciled.unqueue` (a kernel **queued** copy no longer covers our bubble; the copy is hidden instead, one bubble per message; an echo atom covers as before); `cueAnchor` deleted.
- `render.ts`: splice at the slot; `stripOptimistic` before any kernel index is applied (`chatTail`); `hideQueuedCopy` + a held group's reason carried onto our bubble; header/cue code, delegates and CSS removed. The "not confirmed" state and the ✕ on the bubble stay.
- Kernel fields `absorbed` / `landedAt` are untouched (judges read them).

**Tests (red before)**
- `ui/webview/send-pending.test.ts`: placement after the anchor, slot kept while steps stream, landing at the same index, groups per anchor, queued-copy hiding, render/CSS pins (no header, no cue, splice not push, strip before delta).
- `tests/test_pending_in_place.py`: served page, send dropped at the socket, two streamed tool steps, then the CLI's `queued_command` attachment lands the atom in the bubble's slot; position unchanged, reader not moved, no header/cue. Skips without a browser.

**Scroll**: placing in place never moves the viewport; at the press the slot is the tail, so the existing at-bottom reveal is unchanged.

**Served-test hardening (same PR)**: five served tests munged the transcript's project dir with slashes only; the kernel maps every non-alphanumeric character to `-`, and pytest's temp root can carry an underscore, so the kernel found no transcript and drew an API-error turn. That was the "fails only in batch runs" pattern reported on T248. They now munge like the kernel and dump the page state on a timeout.

### DIFF
--- a/tests/test_awaiting_box_sync.py
+++ b/tests/test_awaiting_box_sync.py
@@ -24,6 +24,7 @@ All fixtures synthetic.
 """
 import json
 import os
+import re
 import shutil
 import socket
 import subprocess
@@ -143,7 +144,10 @@ class ServedSync(unittest.TestCase):
              "lastSid": SID, "alive": True, "model": "claude-fable-5-1", "liveModel": "Fable 5.1"}))
         Path(cls.state, "usage.json").write_text(json.dumps({"five_hour": {"pct": 100}, "seven_day": {"pct": 10}}))
         claude = os.path.join(cls.lab, "claude")
-        proj = os.path.join(claude, "projects", cwd.replace("/", "-"))
+        # the kernel finds a session's transcript under Claude's project dir: EVERY non-alphanumeric char of the
+        # realpath becomes '-' (jd._proj_dir). A slashes-only munge missed the '_' pytest's temp root can carry,
+        # so the kernel found no transcript and drew an API-error turn instead (the batch-only flake, 2026-09-08).
+        proj = os.path.join(claude, "projects", re.sub(r"[^A-Za-z0-9]", "-", os.path.realpath(cwd)))
         os.makedirs(proj, exist_ok=True)
         # a CLOSED turn: an open one would invite the boot reconcile to resume it — no real CLI here
         Path(proj, SID + ".jsonl").write_text(
--- a/tests/test_paste_to_focus_browser.py
+++ b/tests/test_paste_to_focus_browser.py
@@ -17,6 +17,7 @@ dev box with the extension installed, which is where ships are gated. All fixtur
 """
 import json
 import os
+import re
 import shutil
 import socket
 import subprocess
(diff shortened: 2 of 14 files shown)


ksarma:
Recommend approval once CI is green on a settled head; the head moved twice while this review ran (reviewed at d1613401, the third commit read; a later push is not covered and gets its own pass). Nothing blocks on the reviewed head. Four follow-ups and one product call for the maintainer to confirm separately.

Reviewed inline: `injectionGroups` groups sends per anchor and returns the highest index first, so the bottom-up splice in `render.ts` keeps lower indices valid. The kernel's `queued` and `apiError` events carry no uuid and cannot anchor. The kernel sorts the echo atom and the absorbed attachment by timestamp into the last turn, so both land in the bubble's slot. At the current head, the gap check in `chatTail` counts kernel events without mutating, and `stripOptimistic` runs only once a delta will be applied. The first cut stripped before the two early returns, so the model lost the bubble while the DOM still showed it. The third commit's tests fail at the second commit on that mismatch. The unqueue path hides one kernel copy per send and carries a held reason onto our bubble. A copy the kernel marked non-cancelable is no longer hidden. No kernel file changes, so `absorbed` and `landedAt` are intact. The not-confirmed state and the bubble's own cancel survive. The second commit is a one-line test-only fix: an ask stub gains `head=None`. It turns two tests red at main green and is here so CI passes. Removing the header and cue reverses part of #996 on a product decision stated in the body. That is a call to confirm separately, not a review finding.

Follow-ups:

- `ui/webview/render.ts:374` and `:3870`: a send the kernel confirms it holds (received, its copy hidden) still shows the sending label and the not-confirmed tooltip for the whole wait, hours under a usage-limit hold. The bare head drops the sends-after-you-answer note. Pass `received` through `mk` and carry the ask note as `held` is carried.
- `ui/webview/send-pending.ts:361` (`placementIndex` at the current head): when the kernel's queued group already lists a message this client does not own, the new send is drawn above it and moves below it at landing. Main merged it into the group in FIFO order. No test covers a foreign text in the group at the press.
- `docs/read-side.md:239`: the published page still describes the removed header and cue. `kernel/kernel.py:26284` carries the same stale comment.
- PR body and commit f1b478b9: only the new served test gained a page-state dump on timeout. The four existing drivers changed by the munge line alone, which does match `kernel/judge.py` `_proj_dir` in all five.

Test evidence: at origin/main, the PR's `send-pending.test.ts` fails to build (no export `injectionGroups`) and 12 pins fail across chat-delta-resync, optimistic-send, queued-indicator and followup-render. There, `tests/test_pending_in_place.py` also fails with header 1 and cue 1 at landing. At the second commit the node suite is 3272/3272 and the six Python files report 36 passed, with the served tests running in chromium. At the current head: 3274/3274, typecheck clean, 36 passed. CI is 7/7 green at d1613401. The current head 289512e3 had no check runs reported at 07:25Z.

Review run by the maintainer's automated review session; verified against head d1613401.

PR #1045, opened 2026-09-08
T248b: kernel-side settings converge across attached machines without a click

henrypinkard:
Follow-up to the Suggest /compact fix: the user's standard is that this setting is consistent always. One click already reached every attached kernel, but a machine attached **after** the click kept its own copy until the next click, shown only by the gear's mixed mark.

**Change** (`kernel/kernel.py`)
- The tunnel supervisor lifts each up peer's `settingsGt` beside the `settings` dict it already lifts on every poll, and calls `_adopt_peer_settings(host, rver)` outside the remotes lock.
- A peer's stamp newer than the local store's, with a different value, is applied through the setting's own gt-gated setter under the **peer's** stamp. Latest-wins on both sides, no ping-pong (the adopter's stamp then equals the peer's; equal stamps are never adopted). Equal stamps or equal values write nothing; no stamps (older kernel) or junk teach nothing; stand-down verdicts are consumed on the supervisor thread.
- **Generalized** to `compactSuggest`, `autoNudge` and `fileEditing`: each is one row of the same table (the judge tiers fan out over `/judge-settings` already; update mode is a per-install boot policy by design).

**Tests (red before)**: `tests/test_mesh_settings_adopt.py` — hermetic two-kernel unit tests at the poll seam (newer → applied; older → stands down; equal → no write; same value → no write; junk/older kernel → nothing; no ping-pong; newest click wins both ways; the three settings generalize) plus source pins on the lift and the outside-the-lock call.

**Review fixes (second commit)**: the poll now returns `settingsGt` (the first cut consumed it without carrying it across, so nothing converged in production while hand-built test dicts passed); a same value under a newer peer stamp now lifts the local stamp too (otherwise the dashboard, which mints above the local kernel's stamps only, could click below the peer's stamp and have the click adopted away one pass later). Tests drive a loopback peer through the real poll.

**Review fixes (fourth commit)**: `/version` now reads each adopted store's value and stamp from one snapshot (`_mesh_settings_snapshot`): read separately, a click landing between the two reads handed a polling peer an (old value, new stamp) pair, which the adopter persisted and the equal-stamp rule then froze on both machines. A NaN/Infinity stamp no longer raises mid-loop (it skips only its own setting).

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -11,6 +11,7 @@ zero protocol change at switchover. WS is hand-rolled on the stdlib socket (no d
 
 Run:  bin/romp-kernel   → opens http://127.0.0.1:29855
 """
+import math
 import contextlib, json, os, queue, random, re, signal, socket, sys, time, threading, traceback, base64, bisect, errno, hashlib, hmac, struct, subprocess, shutil, shlex, http.client, uuid, tempfile, stat, gzip, collections, functools
 from pathlib import Path
 from datetime import datetime, timezone, timedelta
@@ -864,6 +865,7 @@ def _version_info():
                 pass
     except OSError:
         pass
+    _mv, _mgt = _mesh_settings_snapshot()   # value AND stamp of each mesh-adopted store from ONE read (T248b)
     return {"kernel_sha": _kernel_sha(), "kernel_ver": _kernel_ver(), "pid": os.getpid(), "started": int(_STARTED),
             "boot": _BOOT_ID,   # lets a page retire update offers from a previous kernel life (2026-08-15)
             "uptime_s": int(time.time() - _STARTED), "dist_ver": _dist_ver(), "bundles": bundles,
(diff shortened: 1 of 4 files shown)


ksarma:
Not approvable at 35a42389: the mechanism under review does not run in production at that head. The follow-up commits that fix it need a rebase, a CI run and their own review pass, which will follow once the head settles.

Reviewed inline: the convergence argument holds. An adoption writes the peer's value under the peer's stamp, so the two stamps are then equal, and an equal stamp is never adopted. Stamps only rise and are bounded by the newest click. So two or three kernels, or a partition that heals, settle on the newest click without oscillating. The adopted stamp is the originating browser's gesture-clock stamp. The KERNEL_SETTING broadcast already writes that same stamp into every attached kernel, so clock-skew exposure is unchanged. A local click under an adopted future stamp gets a settingStale frame that teaches the gear's clock instead of a frozen control. The adopt call runs after the remotes lock block closes. The setters take their own RLocks and never the remotes lock, and `_pop_stale_notice` runs after every setter. An older kernel with no settingsGt fails the isinstance guard and adopts nothing. The AnAskStaysOnThatPeer stub change in the second commit is test-only and fixes a real red at main. Adoption ignores per-host trust, as the browser broadcast does. fileEditing is the one row where that gates writes; that is a maintainers' call rather than a defect.

- `kernel/kernel.py:16523`, blocker. `_poll_remote_version` returns sha, ver, autoNudge and settings and drops settingsGt. The dict the supervisor hands to `_adopt_peer_settings` therefore never carries a stamp, and the adopter returns `[]` at its first guard on every real poll. Reproduced against a fake /version carrying settingsGt. Through the real poll nothing adopts and the local store stays (False, 0). The same dict handed in raw adopts. The tests pass because `_Kernel.version()` hand-builds the dict and ThePollSeam pins source text. Commit 849f566a adds the key and a ThroughTheRealPoll class. It also changes what a same value under a newer stamp does, so it needs its own pass.
- `kernel/kernel.py:33242`, follow-up. The same-value skip leaves the adopter's stamp stale. Take A ON at 1000, B ON at 5000, C OFF at 3000, with rows polled B then C. A adopts C's OFF and holds it for a full pass before B's ON returns. 849f566a removes the skip. The C-then-B order still shows the intermediate value for the seconds between two row polls, and the suite has no three-kernel heal test.
- `kernel/kernel.py:930`, nit. The /version comment still says settingsGt is not lifted into /tunnels rows. Line 18676 now writes it onto the row, and nothing reads the row field (the adopter reads rver). Drop the row write, or update the comment and give the field a reader.
- `tests/test_mesh_settings_adopt.py:98`, nit. The stand-down test gives the peer the older stamp, so the `pgt <= stored` guard skips before any setter runs. The setter's stand-down path and the `_pop_stale_notice` consume the test names are not exercised.

Red-first at origin/main with the PR's test files checked out: `tests/test_mesh_settings_adopt.py` 10 failed (AttributeError on `_adopt_peer_settings` and `_MESH_ADOPTED_SETTINGS`, plus the lift-line pin), and the second commit's test file 24 passed. At head 35a42389 both files 34 passed. Main's own copy of the AnAskStaysOnThatPeer tests fails 2 at origin/main and passes with the one-keyword change. CI at 35a42389: all seven checks pass. The current head 849f566a has no checks reported. GitHub marks the PR as conflicting with main, where the same stub line already landed.

Review run by the maintainer's automated review session; verified against head 35a42389.

PR #1046, opened 2026-09-08
Usage modal: no unattributed stack from float residue, honest other count, light pressed toggle, loader backstop, dim once, phone door

henrypinkard:
**Fixes (T247b).** Six should-fix items from the adversarial review of the merged usage modal, each reproduced red first.

**Data (`_spend_detail`)**
- **Float residue conjured an unattributed stack.** The recorder rounds a bucket's sum and each sid's sum to six places independently, so a fully attributed bucket carries a few millionths of a dollar the sids do not (28 of 113 live hour buckets did), and the presence test on the unrounded residues hung a hatched "unattributed" chip with no bars on the 8-day view. A residue below the rounding grain (5e-5, anything that rounds to nothing at the payload's four places) is zero, and presence is tested on the rounded values. Test: two sids whose six-decimal sums differ from the bucket by 1e-6 produce no stack and no row, thirty days of such residue sum to nothing, and a sid-less five-cent turn still shows.
- **Keyed "other" counted login-only sessions.** They contribute nothing and the table drops them; only contributors count, so the legend's count is the table's fold (a mixed host with twelve key-billed and five login-only sessions reads "other (2 sessions)", not 7).

**UI**
- **Light pressed toggle.** `.rsp-btn.on` lost the cascade to `body.theme-light .rsp-btn`; the light state rule is written out, so the pressed chip is the accent with `--accent-fg` text and no hairline. Pinned by computed colors in the served test.
- **Loader backstop.** The detail fetch carries an AbortController with a 20 s timer that lands on the existing error and retry path; a close or re-open aborts the in-flight fetch. Pinned by a never-answering fetch in the served test.
- **Dim once.** The row's cells carry the .55, the annotation inside stays at the row's level, and the fold row is a control row with a muted count and a full-opacity "show all".
- **Phone door.** The "By session →" button has its own row at the hover's button size, full opacity, no longer inside the 10px annotation.

**Review folds (second commit).** Adversarial review, two lenses, eight findings, all confirmed and folded: a re-open while the detail fetch was pending painted a false "no answer" over the loader (the superseded fetch's abort rejects before the successor answers, and the guard leaned on the data being present), so ownership is captured before the controller is nulled and a fetch that is no longer current yields on success or failure; the unattributed legend chip dims like its row; the keyed fixture's key total is the sum of its sids with an assertion that nothing is unattributed; the timeout test pins the AbortError text ("no answer from the kernel after 1 s") and a double-open that keeps the loader up; two comments' specificity arithmetic corrected. One verified-clean report covered the cascade, dim-once, the phone button, hermeticity and privacy.

**Screenshots** (served-page harness, synthetic world): light theme with the pressed toggles, the phone's Usage panel with its button, and the modal at 390px.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -37536,8 +37536,11 @@ if(spPending){try{spPending.abort();}catch(e){}}
 var spAbort=new AbortController(),ms=(window.__rompSpendTimeoutMs|0)||20000;spPending=spAbort;
 var spTimer=setTimeout(function(){spAbort.abort();},ms);
 fetch('/spend/detail',{cache:'no-store',signal:spAbort.signal}).then(function(r){if(!r.ok)throw new Error('HTTP '+r.status);return r.json();})
-.then(function(d){clearTimeout(spTimer);if(spPending===spAbort)spPending=null;SP.data=d;if(SP.open)renderSpend();},
-function(e){clearTimeout(spTimer);if(spPending===spAbort)spPending=null;if(!SP.open||spAbort!==spPending&&SP.data)return;
+.then(function(d){clearTimeout(spTimer);var mine=(spPending===spAbort);if(mine)spPending=null;if(!SP.open||!mine)return;SP.data=d;renderSpend();},
+// a SUPERSEDED fetch (a re-open aborted it) yields whatever it carries: its AbortError rejects on the
+// next tick, while the successor is still loading and SP.data is null — a guard that leaned on SP.data
+// painted a false "no answer after 20 s" over the loader (review find). Ownership decides, nothing else.
+function(e){clearTimeout(spTimer);var mine=(spPending===spAbort);if(mine)spPending=null;if(!SP.open||!mine)return;
 SP.err=(e&&e.name==='AbortError')?('no answer from the kernel after '+Math.round(ms/1000)+' s'):String((e&&e.message)||e);renderSpend();});}
 function closeSpend(){SP.open=false;if(spBack)spBack.hidden=true;spTipHide();if(spPending){try{spPending.abort();}catch(e){}spPending=null;}}
 window.__rompCloseSpend=closeSpend;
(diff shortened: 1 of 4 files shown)


ksarma:
Reviewed after the merge: right as shipped, nothing to fold in the code. The author's auto-merge landed it on green CI while this review ran.

Reviewed inline: The residue gate in `_spend_detail` applies only to the bucket-minus-sids dollar remainder, never to a sid's own spend. Its 5e-5 is the half-unit of the payload's four-place rounding. That drops exactly the remainders that would render as 0.0000; token and turn remainders still sum. The keyed other count adds a sid only when it billed dollars or tokens in the range, the same set the table folds. The light pressed toggle reads `var(--accent)` and `var(--accent-fg)`, with hex only as fallbacks; both are defined for `body.theme-light`. The detail fetch's 20 s AbortController lands on the existing error and retry render, and close aborts it. At e4b6972e a re-open while loading painted a false timeout error over the fresh loader (reproduced in a served page). The second commit's ownership guard on both fetch arms fixes it; its double-open pin is red against the old kernel and green at the head. Dimming sits on the row's cells with the annotation at row level; the fold row's show all stays undimmed. The phone door is its own row at the button's existing 11px. The `kernel.py` hunks stay inside `_spend_detail`, its constant, and the modal's inline JS and CSS. Added lines carry no private strings, hostnames, home paths or banned vocabulary.

- PR body, Screenshots paragraph: it names images attached nowhere on the PR; the harness writes them only under `ROMP_SHOTS`. Attach them or drop the paragraph. Follow-up.
- `kernel/kernel.py:29929`: `_SPEND_GRAIN` and the five `round(v, 4)` sites agree by comment only; a single places constant would tie them. Nit.

Red-first against `origin/main` with the PR's test files: 4 failed, 9 passed (residue stack, other count 7 vs 2, light pressed rule pin, pressed toggle's computed colors). The current diff on `origin/main`: 13 passed, with the served Playwright test running here (CI skips it). CI at b9b02c4e: all seven checks pass. The earlier Python red was main's own restart-sweep failure at be6aadc5, cleared by #1047.

Review run by the maintainer's automated review session; verified against head e4b6972e.

PR #1049, opened 2026-09-08
An empty chat build never blanks a session that has content: the kernel keeps its previous build, and the pane never takes an empty session frame as a transcript wipe

henrypinkard:
## What

The frame path behind the recorded scroll snap (romp-on/romp#1036, #1040): a full `session` frame with `events: []` for a session that has content, followed by a content frame one cycle later. The pane took the empty array as the new transcript (a placeholder flash) and read the content frame as a FIRST build, which re-landed the reader through the full-show route.

**Kernel-side cause.** `build_session` parses whatever path the session resolves to. When that path is unreadable for one pusher cycle, `_parse` of a missing file yields zero events, and `_send_chat`'s full path ships whatever the build holds. Two ways a live session's path is briefly unreadable: a resumed SDK session whose registry already names its new leaf while the CLI has not created the file yet (the fingerprint code names this race), and a transcript moved aside. The file-stat cache key is `None` for a missing path, so the empty build is rebuilt and re-sent every cycle until the file returns; the content then diffs against an empty baseline and goes out as a full frame, the client's "first build".

**Fix, both ends, event-based.**
- Kernel: an empty build for a session whose previous push held events is a failed read, not new information about the conversation. The pusher keeps the last cached build (the diff then finds nothing to send) or sends nothing for the sid when nothing is cached; the targeted push skips it. Said once per episode on stderr and on every occurrence as a `romp-perf chatempty` line. A session that never had content is untouched; a `/clear` is never events-empty (its boundary card).
- Pane: a frame that would take a held transcript from content to nothing is status-shaped (chip, name, meta apply), never a wipe; filed once per session as a `clientDiag` row (`empty-session-frame`) so a kernel that sends such frames is seen. The decision is a pure helper (`ui/webview/frame-merge.ts`).

## Tier

- [ ] tests-only
- [x] fix — a bug fix with a test that fails before it
- [ ] feature
- [ ] major-feature

## Tests

New: `tests/test_chat_empty_build_guard.py` drives the pusher with a scripted `build_session` (content, empty, content → a full frame, then nothing, then a tail; the note once per episode and re-armed after content returns; the targeted push; a genuinely empty session untouched; the decision helper) — red on main. New: `ui/webview/empty-session-frame.test.ts` executes the pane's rule and pins upsert's wiring — red on main's render.ts. Suites: node 3270 passed (`npm test`, typecheck clean); Python 8628 passed, 0 failed (`env -u ROMP_API_KEY_CMD`: with that session variable set, 73 key-source/judge/auth tests fail identically on a clean main, an environment effect, not this change).

## Verified live

The two-kernel harness (kernel A serves the page, kernel B attached through A's relay, headless browser, tmux lanes; the trigger is B's transcript file moved aside for two pusher cycles and back) is running on three frozen trees: main before the scroll fix, main today, and this branch. Its result lands here as a comment when the runs finish; the pre-fix tree is expected to reproduce the recorded snap and this branch to show neither the snap nor the placeholder flash.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -22882,6 +22882,44 @@ WIRE_TAIL = 250                                  # events shipped on a full chat
 WIRE_CHUNK = 250                                 # events per loadOlder (chatHead) response
 
 
+# ── an EMPTY chat build never describes a session that has content (T249b, the user 2026-09-07) ─────────────
+# build_session parses whatever path the session resolves to; a path that is unreadable for one cycle — a resumed
+# SDK session whose registry already names its new leaf while the CLI has not created the file, a transcript moved
+# aside by hand — parses to ZERO events, and that build went out as a full {type:"session"} frame with events: [].
+# The pane took it as the new transcript (a placeholder flash), and the content frame that followed a cycle later
+# was a "first build" that re-landed the reader through the full-show route: the recorded scroll snap (T249). An
+# empty build for a session whose previous build had events is NOT new information about the conversation — it
+# is a read that failed — so the previous build stands until content returns, and the episode is said once on
+# stderr (fail loudly, never degrade silently). A session that genuinely has nothing (a fresh one, a never-run
+# SDK session) has no previous events and is untouched; a /clear is never events-empty (its boundary card).
+_EMPTY_BUILD_NOTED = set()      # sids inside an empty-build episode (one stderr line per episode)
+
+
+def _empty_build_regresses(m, prev_events):
+    """Would sending build `m` blank a session the clients hold WITH content? True when the build carries no events
+    while the previous push's build for the sid did."""
+    return not (m.get("events") or []) and bool(prev_events)
+
+
+def _note_empty_build(sid, path, n_prev):
+    """One stderr line per episode naming the sid, what the previous build held and whether the transcript path is
+    even there; a romp-perf `chatempty` line every time, for the harness/perf log."""
+    sid = str(sid or "")
+    _perf("chatempty", sid=sid[:8], prev=int(n_prev or 0))
+    if sid in _EMPTY_BUILD_NOTED:
+        return
+    _EMPTY_BUILD_NOTED.add(sid)
+    exists = bool(path) and os.path.exists(str(path))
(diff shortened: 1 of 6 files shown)


henrypinkard:
## Verified live: the two-kernel harness, three frozen trees, two runs each

Kernel A serves the page; kernel B is attached through A's relay (a check-in peer row, no ssh); a headless browser on A's dashboard; one live tmux lane per kernel with an overflowing synthetic transcript. Per run: click B's tab, scroll to 40% and switch away and back so the tab holds a saved spot X (the stale spot the bug lands on), restart B (the relay reopens in ~1.3 s and B keys the tab live on all trees), then trigger two pushes and read the pane's own scroll trace and frame log.

Two triggers. **Truncate** empties B's transcript file in place for two pusher cycles and restores its bytes: the kernel then builds the session with zero events, which is exactly the frame path this PR is about. **Move** removes the file for two cycles and puts it back: that does something else (below).

| truncate trigger, after B's restart | main before the scroll fix (#1036^) | main before this PR | this PR |
|---|---|---|---|
| a `session` frame with 0 events arrives, then a content frame | yes | yes | **no** |
| placeholder flash in the pane | yes | yes | **no** |
| reader at the bottom | jumps to X | bounces through 0 and back to the bottom | stays |
| reader scrolled up (Y) | jumps to X | yanked to the bottom | stays at Y |

Two runs per tree, identical. The first column is the recorded snap reproduced live: the reader lands on the spot the tab had when last left, whether they were at the bottom or reading. The middle column is what #1036 and #1040 bought on their own; the right column is this PR's guard closing the frame path itself.

The **move** trigger never produced an empty frame on any tree: when a live tmux session's transcript file is gone, the kernel drops the session from the tab list for those cycles, the pane dismisses the tab (the T236 omission path) and shows the neighbouring tab, and the file's return re-lists it. That is a separate behaviour (a live session without a readable transcript leaving the strip), reported to the team, not changed here.

Harness scripts (reusable, tree path as argument): `/tmp/t249b2/harness/run.sh <tree> <runid>` on this machine, with the recipe in `/tmp/t249b2/HARNESS-BRIEF.md`. Trees were `git archive` copies at a09df60c (pre-fix), 6908ad5f (main at the time) and 3e0718f7 (this branch).

PR #1057, opened 2026-09-08
Tools: offline performance benches for the kernel's payload builders and the dashboard panes

ksarma:
Two development tools, both instruments rather than savings. `tools/perf-bench.py` loads a checkout's kernel in-process against a copy of a state directory and times the payload builders on real-sized data, with no live kernel, no network and every write listed. `tools/ui-bench.mjs` records a pane's WebSocket frame stream, replays it into the real pane page in a headless Chromium, and reports where the browser's time goes. Neither touches the kernel, the panes or any payload: the diff is about 5,900 added lines under `tools/`, `tests/`, `CONTRIBUTING.md`, `tests/README.md`, `docs/reference.md` and three CI steps. The kernel performance counters (#1003) and the browser-side telemetry (#1009) are on main; these tools reference nothing from either, and the two `docs/reference.md` sections they added each end with a pointer to the offline instrument for the same layer. They are one PR because the performance PRs that follow cite both tools' rows as their before-and-after evidence, and because the part worth reviewing closely is the same in both: each runs kernel code outside the live kernel, so the guard lists (what is floored, what is tripwired, what may be written where) are the design. The four commits of the offer are separable and ordered so that each stands without the ones after it: the kernel bench alone; the pane bench with its tests; its docs; its CI step last, so the branch is complete without it. A fifth commit, the review commit, sits on top and touches all four, and a sixth, added after review, makes the bench leave its state copy byte-identical (the section Added after review says what it changed); the paragraph after the list says what the review commit changed, and every section below describes the branch as it stands at the sixth commit.

## Commits

1. **An offline bench times the kernel's payload builders against a copy of a state directory, with no live kernel, no network and every write listed.** `tools/perf-bench.py` and `tests/test_perf_bench.py`. Rows: `Sessions.live` and the names snapshot, `discover` cold and warm, `build_session` per largest transcript three ways (cold with every cache a freshly started kernel lacks emptied, each cold sample checked against the event model's assembly counters and the assembly cache; event-model-warm; warm), one parse of every live transcript, `build_feed` with and without warm parses, `build_timeline` bars and skeleton, `load_goals` per largest store, and the pusher's cold cycle (its parses counted inside it), per-app connect and steady cycle with rebuild samples set aside. `--profile` adds cProfile tables per builder; `--repo` picks the kernel so two checkouts run against one copy; `--compare A.json B.json` prints per-row deltas after flagging any difference in the two runs' worlds, a run that stopped on a guard (its `error`) among them. Isolation, all reported in the output: the state root rebound to the copy and the manager port set to a dead port before the import, the catalog, CLI-scope and Claude-binary seams floored, every key-source and credential name `tests/conftest.py` pops dropped (`ROMP_API_KEY_CMD` and `ROMP_API_KEY_REF`, `ANTHROPIC_*`, `CLAUDE_CODE_OAUTH_TOKEN`, `ROMP_EXPECTED_AUTH`, the `OP_*` names and the `OP_SESSION_*` prefix; the test pins the list against the kernel's own constants), `subprocess` in every loaded romp module (the kernel, judge and SDK backend modules and every sibling `kernel/` module one of them loaded, `keysource` among them) replaced by a tripwire that raises on any spawn except the kernel's read-only git queries (`rev-parse`, `ls-files` and the pair `remote get-url`, counted per build; `--no-git` answers even those as failures), network fetches and notifications replaced by recorders that take the kernel's full call shapes (a mismatch there would not raise: `_push` catches its build's exceptions, prints `push build:` to stderr and returns), the background parse asked for and counted rather than started, every thread start counted, `pwd` lookups counted, every write the kernel aims under the copy (`_atomic_write`, the import-time repo-root marker, the order audit log) landing in a shadow directory under the tool's private temp dir, with the kernel's own re-reads of the small JSON state files served from the shadow, every refusal (a spawn, a write aimed outside the copy) recorded before it raises so that a run ending with one on record fails naming the guard even when the kernel's `_push` caught the error and went on, and the copy (files, directories and link targets) fingerprinted before and after on every exit path, byte-identical afterwards. The live default state directory is refused without `--i-know-this-is-live`, and then only a temporary mirror without `sdkvenv` or the kernel's 0600 credential files (`serve-token`, `push-vapid.json`, `push-subscriptions.json`, `remotes.json`) is benched. The guards that decide whether a number is a number are module-level functions, so the test drives each with inputs the fixture cannot produce. Measured with the tool at the main this branch was first built on (3d23d779; not re-measured after either rebase: the first hop's changes to the timed builders were wrapper-level, stores read through `load_goals_or_fault` and a stat before each store read; the second hop's are not, #1060 and #1059 having landed, and the Measurements section says what that does to the table), on a copy of a 31-session state directory (506 MB of transcripts, `--iters 5`, medians): steady pusher cycle 425 ms (rebuild samples 3.0 s), `build_feed` 882 ms, `build_timeline` bars 1061 ms plus a 155 ms skeleton, cold `build_session` 1.3 to 5.3 s on 26 to 75 MB transcripts (the parse over half of it), `load_goals` 5.2 to 7.3 ms on 1.1 to 1.6 MB stores. The full table is in the commit body.

2. **A headless-Chromium bench records a pane's frame stream and replays it into the real page, reporting where the browser's time goes.** `tools/ui-bench.mjs` and `tests/ui-bench.test.mjs`. `--record <app>` connects to a running kernel as the page's shim does (`/ws?app=&delta=1&iid=`, the token as the `romp_token` cookie with a same-origin Origin, the `{type:"ready"}` handshake and nothing else) and writes the frames as JSONL only under the system temp directory (0700 directory, 0600 file, no path inside a git checkout or through a symlink), because a recording is real session data. `--replay <app> --frames FILE` serves the real page, shim and bundles through the kernel's own `Handler` in a python3 subprocess under the floors `tests/conftest.py` applies (the manager variables removed and the manager port set to a dead one, every key-source and credential name conftest pops removed, service env file and model-catalog fetch pointed away, `ROMP_CLAUDE_BIN=/bin/false`, `ROMP_CLI_SCOPE=0`, `ROMP_POSTAL_PEERS=0`, a serve token minted per run, a stdin watchdog), with a Node front server owning `/ws` (the page's own origin, required: an upgrade with a foreign Origin or none at all is refused). The pane shim hands frames to the bundle in its own task (a `MessageChannel` flush that coalesces same-type whole-state frames and stops after 8 ms), so a wire frame and its render are two measurements: the in-page instrument wraps the WebSocket handler (the shim's synchronous work per wire frame; a `needSlot` sent from inside it marks a refused frame), the flush task and the posts that arm it (so the run waits for the queue to drain), and the handoff itself (`window.__rompFed.inbound`, else the `MessageEvent` fallback), and pairs each delivery with the wire frames it carried by the shim's own queue rules. Per frame type the report gives the bytes, `count` and `delivered`, the handler, bundle and settle times (settle end to end: receipt to the main thread free after the delivery that carried it), how many frames coalesced, which seam the deliveries came through, long-animation-frame attribution naming the flush task (the instrument's own entry points, the settle stamp, the observer callback and the collector, are labelled as its bookkeeping, never as a row under its file), the heap before and after a forced GC, DOM and layout counters, console errors and uncaught exceptions, and the messages the pane sent. `--cpu-profile` samples V8 at 500 us, writes a DevTools-loadable profile with the token stripped, and ranks functions by self and total time with source positions from the dist's `.map` files, overall and inside the deliveries that carried the first content frame and the largest frame per type. `--synthesize <app>` writes a byte-stable stream in this tree's wire shapes for the feed, Outline and timeline pages; the chat is recorded, not synthesized. `--gap MS` paces a replay at a fixed interval; `--fast` sends back to back; `--iters N` pools runs; `--cpu-throttle K` slows the page. Measured at this tree on a synthetic 200-card feed stream, `--gap 100`, 3 iterations: the 192 KB full frame handler 1.3 ms, bundle 125 ms, settle 160 ms; the 24 deltas each delivered on their own, handler p50 0.2 ms, bundle 9.9 / 44.3 / 54.6 ms (p50 / p90 / max), settle 28.8 / 66.4 / 89.2 ms; back to back the 25 content frames coalesce into 1.7 deliveries per run (bundle 114 ms for the delivered ones). The same stream through this tool's earlier form, at a main whose shim rendered inside the handler, read handler 118 ms and settle 424 ms for the full frame. The profile puts 93% of `render()`'s self time on the `scrollTop` restore in `ui/webview/feed.ts`, and 99% of the instrument's samples fall inside the handler, flush and delivery windows. For scale, the same tool on a downstream kernel whose shim still rendered inside the handler, on a recorded feed stream of about 800 cards: the first 6.6 MB frame about 90 ms of `JSON.parse` and 870 ms in the handler for 47,000 DOM elements, a 542 KB delta 775 ms, and 97 to 98% of `render()`'s self time on the same line.

3. **CONTRIBUTING documents the pane bench, and tests/README names its node suite.** A "Measuring dashboard pane performance" section (prerequisites, the five commands of a before-and-after measurement with the replay a frame every 100 ms, why the handler and bundle columns are two measurements and what `--fast` coalesces, what a replay and `--cpu-profile` report, how the pages are served, why a recording may land only under the temp directory, which apps can be synthesized, how the node suite runs and skips, POSIX-only) and a "Test environment" heading over the existing notes; `tests/README.md` lists `ui-bench.test.mjs` beside the other node suites; the two performance sections of `docs/reference.md` each end with a pointer to the offline instrument for that layer. Docs only; every command in the section was run as written from the repo root. The review commit extends the section with the pinned playwright Chromium as a way to get a browser (`cd vscode-extension && npx playwright install chromium`, which CI installs), the manager-port and key-source floors as conftest applies them, and `ROMP_UI_BENCH_TIMING=1` for the timing relations; `tests/README.md` names that flag beside `ROMP_UI_BENCH_REQUIRE`.

4. **CI runs the pane bench's node suite after the extension build, with a browser required.** The `vscode-extension` job gains `node --test tests/ui-bench.test.mjs` after Build with `ROMP_UI_BENCH_REQUIRE=1`, so the four headless replays exercise the feed and timeline bundles just built, and a runner without a browser or python3 turns the job red instead of skipping. As posted, the step ran on the runner image's Google Chrome through playwright's channel option with no browser download; the review commit puts two steps before it, an `actions/cache` of playwright's browsers keyed on the extension lockfile and `npx playwright install chromium` (no `--with-deps`), so the required check runs the Chromium the pinned playwright fetches, never the image's unpinned Chrome, and it leaves `ROMP_UI_BENCH_TIMING` unset. Two tests go with the step: the PATH scan that finds a system browser is exercised in a subprocess with a fake `google-chrome` or `chromium` and an empty playwright browsers directory, and a test before the replays asserts a browser was found when the variable is set and prints which one as a diagnostic, so the step's log names the browser the replays ran in. Observed as posted, on ubuntu-latest, in the two CI runs of the pre-amendment tree (6db5b129): 29 and 21 s for the step (42 tests, 0 skipped; the log's browser line read `system google-chrome (channel chrome)`); in the first, 11.8 s was the nested-run test that proves the variable fails rather than skips, and the four replays took 4.7, 1.0, 2.5 and 2.5 s. At the review head, in this PR's CI run (34222961761): the install step 13 s on a cache miss, the bench step 13 s, 45 tests, 0 skipped, no timing diagnostic in the log, the browser line `playwright chromium`, the four replays 3.8, 0.5, 2.2 and 1.4 s.

5. **Review fixes: the bench step asserts nothing timing-dependent, a tripped guard fails the bench, the browser is provisioned by the workflow, the env floor matches conftest.** The review commit, not one of ours; it touches `.github/workflows/ci.yml`, `CONTRIBUTING.md`, `tests/README.md`, both tools and both test modules.

## Review commit

The review commit changed five things, plus two from findings it judged real. (1) The replay tests in the required CI step assert only what holds under any scheduling: frame totals, ordering, presence, and the handoff's accounting identity delivered + coalesced + shim + queued = count; every timing relation (settle margins, the parse cheaper than the render, the share of deltas delivered on their own, the heap freed by the forced GC, the sampler's density, the wrappers' sample share, the CPU-throttle ratio) is an assertion only under `ROMP_UI_BENCH_TIMING=1` and a diagnostic line otherwise, and CI leaves the flag unset; a unit test pins how the instrument attributes its own entry points (the settle stamp, the long-animation-frame observer callback, the collector), so a stretched bookkeeping task never files a row under the instrument's file. (2) `tools/perf-bench.py` records every refused write before its raise and, at the end of `run()`, `check_guards_held` fails the run (rows so far kept, JSON marked partial, `error` set, exit 1) naming the tripwire or the write guard when a refusal is still on record, since `_push` swallows the `BenchError`; `--compare` flags a run with `error` before diffing its rows; two end-to-end tests plant a spawn and an out-of-copy write inside `_push`. (3) `ci.yml` installs the pinned playwright Chromium (`npx playwright install chromium`, cached on the lockfile hash) before the bench step, so the required check never rides the runner image's unpinned Google Chrome; chosen over a loud skip because the bench's only CI run of the real pane bundles is the coverage the step exists for, and a skip in the required job would leave it green with nothing run, while a pinned download is the workflow's own stance for node and bats; no `--with-deps` (apt), no test-count pin. (4) Both tools drop every key-source and credential name `tests/conftest.py` pops (`ROMP_API_KEY_CMD` and `ROMP_API_KEY_REF`, `ANTHROPIC_*`, `CLAUDE_CODE_OAUTH_TOKEN`, `ROMP_EXPECTED_AUTH`, the `OP_*` names and the `OP_SESSION_*` prefix) and set a dead manager port; the ui-bench env test plants all of them, `test_perf_bench` pins the list against `keysource.SOURCE_VARS`, `OP_ENV_NAMES`, `OP_ENV_PREFIX` and `sdk_backend.AUTH_ENV_NAMES` read from source, and the tripwire covers every sibling `kernel/` module a driven module loaded (`keysource` among them), the docstring saying which. (5) `test_perf_bench.py` pins only the caches the cold proof rests on, checks the writes as a superset plus the guard's own record, and asserts the git-per-build relation (at least one `ls-files` per build in a plain directory, fewer in a checkout) instead of exact counts; a CLI test drives `tools/ui-bench.mjs` as a subprocess (`--synthesize`, `--compare`, the usage, each refusal) and a unit test covers `parseArgs`. Also: the front server's `/ws` refuses an upgrade without an Origin header (a page always sends one), and the mirror and the rsync recipe leave `push-subscriptions.json` and `remotes.json` out with the other 0600 credential files. It was reconciled onto the amended head 09896597, keeping that head's rewrite of the refused-delta test.

## Rebased onto main on 2026-09-08

The branch was first built on 3d23d779, was rebased once onto d31e1b1c, and now sits on 66bb933e. The commits at this base, oldest first: e630a5db, 94cec35a, a1a57db7, 8a31befb (the amended head, 09896597 before this rebase) and 75127a4f (the review commit, 0d4454db before it); every author, message and diff is the one reviewed, the one context change being the paragraph below. Of the second hop's 245 commits, one conflicted: `docs/reference.md`, where #998's new API-health section sits between the browser-side telemetry section and "Where things live", the spot commit 3's `ui-bench` pointer was anchored to; the pointer keeps its place as the last paragraph of the telemetry section, ahead of the new section, and the `perf-bench` pointer merged untouched. `.github/workflows/ci.yml` (main's per-merge concurrency group) and `tests/README.md` (the per-session memory-limits docs) merged without conflict, the three CI steps still last in the `vscode-extension` job and the README still counting four suites before this one. Two merges changed the kernel the bench times: #1060 (the pusher and timeline changes) and #1059 (the goal-store memos), the changes the Measurements table's right column was cited for, are on main now, so that table's left column describes the kernel before them and is not this base's numbers (not re-measured; the section says so). At this base: `tests/test_perf_bench.py` 33 passed; `node --test tests/ui-bench.test.mjs` under `ROMP_UI_BENCH_REQUIRE=1` 45 of 45, 0 skipped, on playwright's Chromium, in 12.9 s after `npm ci` and a build; the full pytest under `-n 8` on a loaded machine that self-hosts a kernel had failures in three modules this branch does not touch, `test_rename_route` and `test_tag_edit_ack` green alone and `test_postal_token` red identically at bare main in a scratch checkout of it (an xdist temp-directory failure on that machine). `kernel/`, `ui/` and `vscode-extension/` stay byte-identical to main. Of what main gained in the first hop, three changes touched this work. #1016 moved the pane shim's handoff to the bundle out of the WebSocket handler into a `MessageChannel` flush with same-type coalescing and 8 ms slices, so commit 2's instrument was reworked: it wraps the three seams named above instead of one, the report separates the handler, bundle and settle columns and counts coalesced frames, the CPU-profile windows and the clock alignment use the deliveries, and the feed replay in the tests runs at 100 ms gaps so each delta reaches the bundle on its own (the earlier alignment assertion over handler windows of about a millisecond was quantized to a handful of samples and failed most runs at the tip; it now holds at 99%; since the review commit both are timing relations, asserted under `ROMP_UI_BENCH_TIMING=1` and diagnostics otherwise). Commit 3's section explains the two measurements. #1019 made an unparseable goals file a quarantine (moved aside as `<file>.corrupt-<stamp>`) and a read fault an error; the kernel bench's builders reach stores through `load_goals_or_fault` because they are the live code paths, and on a copy holding such a file the kernel moves it aside inside the copy, which the before-and-after fingerprint lists. #1020 (state readers proved) changed no file the bench's expected write set names; the set (`order-audit.jsonl`, `repo-root`, `session-order.json`) is what a run aims at the copy at the tip, and since the sixth commit those writes land in the tool's shadow while the copy stays byte-identical (the test checks the shadowed set as a floor and the copy's census as 0 changed, 0 new, 0 removed). #1003 and #1009 landed, so nothing here depends on an open PR any more, and the `docs/reference.md` pointers the first draft left out (their anchors were those PRs' sections) are in commit 3. Nothing was dropped.

## Nothing visible changes

No kernel, `ui/webview`, payload or protocol edit; no new route; no runtime dependency; the extension bundles are unchanged. The tests pin the invariants that make the tools safe to run beside a live kernel:

- `perf-bench` refuses the live state directory without its flag and mirrors it even with the flag (the four 0600 credential files and `sdkvenv` excluded, mirror removed); every spawn other than the three read-only git queries raises; the writes the kernel aims at the copy (`order-audit.jsonl`, `repo-root` and `session-order.json`) land in the tool's shadow and are listed by relative path while the copy is byte-identical afterwards (a tree hash before and after), one aimed outside the copy is refused before it runs and recorded, and a refusal still on record when the run ends fails it, even one the kernel's `_push` caught; transcript mtimes are unchanged after a run; no thread is started and none is left running.
- `ui-bench`'s replays run against the real pages served through the kernel's own `Handler` (the front server pipes every non-`/ws` response through unmodified and refuses a foreign or missing Origin or any other path on `/ws`; the tests assert the pages render with no failed resource, no console error and no resync request); the recorder sends the ready handshake and nothing else; a recording lands only under the system temp directory; the page-server subprocess never starts with the manager variables (each planted and asserted absent by name, the manager port set to a dead value as conftest does) or any key-source or credential name conftest pops (each planted and asserted absent) and gets a token minted per run; the written profile carries no token.
- CI: three steps added to the `vscode-extension` job (a cache of playwright's browsers, the pinned Chromium install, the bench), no change to the python or shell jobs.

## Tests

Commit 1: `tests/test_perf_bench.py`, 29 tests as posted, 33 at the review head, 42 at the head (the class setup benches a 400-turn synthetic transcript under cProfile, about 2 s on the CI runners). Red-first, recorded at the first base: (1) the tool's earlier module loader, which needed a loader helper this tree lacks: 7 of 14 red with the tool's own error; re-derived to `SourceFileLoader(...).load_module()`, as `kernel/kernel.py` loads its siblings: green. (2) The constructor-free backend dict without `_owns_memo` (which `SdkBackend.owns()` memoizes on at this tree): 7 red with the BenchError naming the attribute; entry added: green. (3) The cold row: dropping `_ASM_CACHE`/`_ASM_LOCK` from the caches emptied before a cold sample made `test_cold_build_is_a_full_parse` red through the pinned cache set; making the clearing a no-op made 7 red with the tool's own "ran no full assembly" error; restored: green. (4) Dropping `ls-files` from the tripwire's allow list: 8 red, the tripwire naming `_repo_file_index <- _resolve_path_token <- _path_links <- build_session`; restored: green. (5) The recorders called in-process with the kernel's call shapes: red on the tool's first stubs, a `TypeError` on `_push_notify`'s `kind=` keyword, which `_cached_feed` passes on every bell; tolerant stubs: green. Then, at this base, a mutation pass over the claims the earlier tests did not execute added 13 tests and a fixture change (the api session's cwd a plain directory, a placeholder `sdkvenv`, a planted `ANTHROPIC_API_KEY` variable that is not a key, an `age_api_days` knob): the two cold-sample proofs, the snapshot row count (both directions, and `--all-regs-live`), `prepare_env`'s every floor and change, the bench backend's rows (threads skipped, states from the log, `--dormant-rows`, `--all-regs-live`, a missing attribute named), the steady push's rebuild bucketing (the fixture cannot force a rebuild, so this is where that split is executed), a late BenchError leaving the rows printed and the partial JSON written, the mirror tolerating vanished files only and leaving no half-made copy otherwise, an `_atomic_write` outside the copy refused before it runs, the background parse recorded and threads counted, a transcript outside discovery's window backfilled, a kernel lacking a builder refused by name, the git queries per build (as posted the exact counts, one `ls-files` per build of the plain-directory session and none for the checkout's kept samples; at the review head the relation, at least one per build in the plain directory and fewer for the checkout, whose warm-up cached the answer), a connect sample's fresh client receiving what the cold cycle's did, the cold cycle's parses counted inside it, `--sessions` bounding the rows, `repo_head` equal to `git rev-parse HEAD`, `warm_all_parses` over both sessions, the discovery cache cleared before every cold sample, and the planted key dropped. Each was run against a mutation of the guard it covers: 26 mutations, every one red at the mutation and green at the commit. The module passes `tests/test_state_isolation_order.py` and `tests/test_tempdir_hygiene.py`. The review commit adds four tests: two end-to-end runs against a copy of `kernel/` with a spawn or an out-of-copy write planted inside `_push`'s try block (exit 1, the JSON marked partial with `error` naming the guard, the refusal listed, the rows measured before the check kept), the key-source list pinned against `keysource.SOURCE_VARS`, `OP_ENV_NAMES`, `OP_ENV_PREFIX` and `sdk_backend.AUTH_ENV_NAMES` read from the source with `ast`, and `check_guards_held` in-process; the env test plants every name on that list. It also loosens three pins: the cold-cache check asks that `_parse_cache`, `_JSONL_CACHE` and `_ASM_CACHE` were emptied and that the rest came from the tool's own list, the writes are a floor plus the guard's record with nothing removed (since the sixth commit, the shadowed set as a floor beside a copy census of zero), and the git counts are the relation above, so an unrelated kernel cache rename or a new write path no longer fails the tool's test.

Commit 2: `tests/ui-bench.test.mjs`, 40 tests in 10 to 11 s as posted (43 at the review head, which adds a unit test for the instrument's attribution of its own entry points, a `parseArgs` unit test and a CLI test that drives the tool as a subprocess: `--synthesize`, `--compare`, the usage, each refusal), four real headless replays. Red-first at the first base: the tool as it ran on this fork, unchanged: 33 of 34, the feed replay red with a page error from a synthetic card lacking `trgb`; the re-derived tests against the unchanged tool: 6 red; the tool re-derived except the two floors: only the floor test red; floors added: 34 of 34. At the rebase the instrument was reworked for the shim's flush task (above): 36 tests, the alignment assertion moved to the delivery windows. Then the mutation pass added four tests (the front server's `/ws` origin and path check; the instrument on a blank page: the posts that arm the flush counted only from inside a handler or flush, every armed flush fired, a delivery's settle stamped after the second animation frame and not before, a dispatch inside an inbound one delivery and not two, the seam recorded, a throwing inbound recorded and rethrown; a replay of the feed with its last delta's base wrong and a final frame the pane throws on, the refused delta the shim's alone with one `needSlot` sent and the exception collected as uncaught; a replay of the timeline with `--iters 2` and with `--cpu-throttle 4`) and new assertions in the existing ones (a token minted per run, the stripped variables planted and named by the test, two chained frames in one flush giving two deliveries, every delivery explained by a frame and every one through federation's inbound, the first frame's handler time a measurement and below its bundle time, a keepalive's settle at least one frame interval, the heap smaller after the collection than before it, the profile's sampling interval a median under 750 us, a run whose collection did not return saying so). 19 mutations of the tool, every one red (a constant token, a shrunk stripped list, the whole-state rule applied to every kind, the depth guard, the `needSlot` wrap or the `postMessage` count removed, the GC, the sampling interval, the throttle or the Origin check dropped, the inbound trap gone, a handler time of zero, a settle stamped synchronously or after one frame, the iters loop running once, the pageerror listener dropped, a delivery error not recorded, `afterGc` a literal). One claim is not executed by a replay: the shim and the pane bundles log nothing through `console.error`, so the console-error listener is covered only by the fold test that renders collected errors. Two things changed after the mutation pass. When this PR's first CI run went red on the refused-delta replay (a 50 ms-gap replay asserting that no delta ever coalesces; one delta of 23 coalesced on the runner while the same tree passed on the fork), the test was amended (09896597) to pin the refused and the throwing frame by index and to sum delivered + coalesced for the rest. The review commit keeps that rewrite and applies the principle to every replay: what holds under any scheduling stays an assertion (frame totals, ordering, presence, the handoff's identity delivered + coalesced + shim + queued = count, the seams, the token, the profile's shape and source maps), and every timing relation in the list above (the first frame's handler below its bundle time, the settle margins, the keepalive's settle, the share of deltas delivered on their own, the heap smaller after the collection, the sampler's median interval and refined alignment, the wrappers' sample share, the CPU-throttle ratio) is an assertion only under `ROMP_UI_BENCH_TIMING=1` and a diagnostic line otherwise; a mutation caught only by one of those relations (the throttle's ratio, the sampler's density, the heap the collection freed) is red only with the flag set.

Commit 3: docs, no test module. Both insertion anchors were asserted verbatim before the edit; every command in the section was run as written against temp-directory paths (the `--record` command against a loopback WebSocket server standing in for a kernel).

Commit 4: two tests plus the step. The PATH-scan test spawns the module with an empty playwright browsers directory and a PATH holding only a fake `google-chrome` (channel `chrome`), then `chromium` (channel `chromium`), then nothing (the refusal says how to install a browser); removing the scan turns it red. The diagnostic test fails under `ROMP_UI_BENCH_REQUIRE=1` when no browser is found and prints `# browser: <how>` otherwise. Before the step existed, with the browsers hidden and `PLAYWRIGHT_BROWSERS_PATH` at an empty directory, the replay tests under `ROMP_UI_BENCH_REQUIRE=1` fail naming the missing browser, and without the variable they skip and the run exits 0 (a nested run in the suite checks both, now over four replay tests). On the fork's CI at the pre-amendment head 6db5b129 (run 34192675657) the step passed 42 of 42 with 0 skipped in 21 s (29 s in the run of the identical tree before the last message edit), the browser line reading `system google-chrome (channel chrome)`, and every job of both runs was green; this PR's CI run of that same tree failed the refused-delta replay on one coalesced delta, which the amendment and then the review commit answer (Commit 2 above). At the review head this PR's CI run (34222961761) passed the step 45 of 45, 0 skipped, in 13 s after a 13 s browser install, the browser line reading `playwright chromium`.

Suites. `kernel/`, `ui/` and `vscode-extension/` are byte-identical to main on this branch. The full pytest at the branch's head before this pass (the same kernel), twice under `-n 8` on a loaded machine that self-hosts a kernel: 4 failures, none of this branch: `test_paste_to_focus_browser::ServedPaste` (an environment failure on this machine, known), the two `test_scroll_marks_growth::ServedNotchFollowsGrowth` tests (new on main in this hop; the hermetic chat page renders two turns with no overflow on this headless machine, so the scrollbar notch stays hidden; they fail identically at the bare base in a scratch worktree), and `test_ship_reship::ServedWedge::test_restart_between_ship_and_ack...` (browser-driven; it failed 1 of 3 solo runs at the bare base too, load-sensitive). At the head as posted (09896597), the touched modules with the hygiene suites: `tests/test_perf_bench.py`, `tests/test_tempdir_hygiene.py`, `tests/test_state_isolation_order.py` 50 passed (63 subtests); `node --test tests/ui-bench.test.mjs` 42 of 42 with `ROMP_UI_BENCH_REQUIRE=1`, three runs, 10.3 to 11.4 s. At the review head, this PR's CI run (34222961761) is green on every job: the four Python cells (the 3.12 cell 8921 passed, 50 skipped, `tests/test_perf_bench.py` among them), bats, and the extension job with 45 of 45 node tests in 13.0 s.

## Measurements

Kernel builders (`perf-bench`, medians in ms unless noted). The left column was measured at the main this branch was first built on (3d23d779) on a copy of a 31-session state directory with 506 MB of transcripts, `--iters 5`, and not re-measured after either rebase (a fresh copy of that size is what a re-measurement needs). The first rebase's builder changes were wrapper-level; the second's are not: #1060 and #1059 landed, so the left column describes the kernel before the pusher, timeline and goal-store changes those PRs made, not this base. The right column is this fork's kernel, which carried the pusher changes #1060 has since landed, measured with the same tool on a 31-session copy with 386 MB of transcripts; it is cited for scale, not as this tree's numbers.

| Row | This branch's first base | This fork's kernel |
| --- | --- | --- |
| steady pusher cycle | 425 (rebuild samples 3039) | 350 at head, 4350 one revision earlier |
| `build_feed` | 882 | about 900 |
| `build_timeline` bars / skeleton | 1061 / 155 | 800 to 1300 / 200 |
| cold `build_session` | 1277 to 5346 on 26 to 75 MB transcripts, parse over half | 800 to 2600 on 40 to 56 MB |
| `load_goals` | 5.2 to 7.3 on 1.1 to 1.6 MB stores | 6 to 9 per MB |

Feed pane (`ui-bench`, headless Chromium). The synthetic row is this tree; the recorded row is this fork's kernel and its recorded stream, and its shim rendered inside the WebSocket handler, so its handler figures are shim plus bundle.

| Input | First full frame | A delta | Profile |
| --- | --- | --- | --- |
| synthetic, 200 cards, `--gap 100`, 3 iterations (this tree) | 192 KB, handler 1.3 ms, bundle 125 ms, settle 160 ms | handler p50 0.2 ms, bundle p50 9.9 ms (max 54.6), settle p50 28.8 ms (frames up to 3.6 KB) | 93% of `render()` self time on the `scrollTop` restore |
| the same stream, `--fast` (this tree) | coalesced into a delta's delivery in every run; settle 159 ms | 25 content frames in 1.7 deliveries per run, bundle 114 ms for the delivered ones | same line |
| the same stream at a main whose shim rendered in the handler (this tool's earlier form) | handler 118 ms, settle 424 ms | handler p50 6.1 ms | same line |
| recorded, about 800 cards (this fork, handler-rendering shim) | 6.6 MB, `JSON.parse` about 90 ms, handler 870 ms, 47,000 DOM elements | 542 KB, 775 ms | 97 to 98% of `render()` self time on the same line |

The recorded measurement is the basis of a feed-pane incremental-render change offered separately, which the same tool measured at a delta handler 454 to 63 ms, settle 524 to 105 ms and layouts per delta 127 to 5, with the first full frame unchanged at about 600 ms. The CI step took 29 and 21 s on ubuntu-latest in the two runs of the tree as posted, python3 startup included, on the image's Chrome; in the first, 11.8 s was the nested check that the REQUIRE variable fails rather than skips and the four replays took 4.7, 1.0, 2.5 and 2.5 s. At the review head, on the workflow's own Chromium: 13 s to install it on a cache miss and 13 s for the step, the replays 3.8, 0.5, 2.2 and 1.4 s.

## Contracts changed

None in code. Two things the tree inherits: `tests/test_perf_bench.py` pins the set of functions the bench neutralizes against this tree (`keysource.subprocess` among them since the review commit) and the key-source names against `keysource` and `sdk_backend`'s own constants, checks the caches emptied before a cold sample for the three the cold proof rests on and the shadowed writes as a floor beside a copy that must change by nothing (so an unrelated cache rename or a new write path shows in the tool's output without failing its test, while a write that reaches the copy fails it), and the bench's constructor-free backend dict in `make_backend` mirrors the attributes `SdkBackend`'s read-side methods reach; a renamed safety target, a new credential name in the kernel's constants or a new attribute fails loudly and is fixed in that one place (named in commit 1's body). The `vscode-extension` CI job now downloads playwright's pinned Chromium (about 150 MB, cached on the extension lockfile's hash, no `--with-deps`) and needs python3 on the runner; its bench step fails rather than skips without either.

## Not included

- No kernel, `ui/webview` or payload change, and no performance counters: the counters (#1003) and the browser-side telemetry (#1009) are on main and these tools reference nothing from them beyond the two `docs/reference.md` pointers.
- The tools load modules with `SourceFileLoader(...).load_module()`, the idiom `kernel/kernel.py` uses today; a shared loader helper the fork uses is not part of this offer.
- The pane bench speaks this tree's protocol only (a keyed full plus `{type:"delta", slot:...}`). A second feed-delta protocol and a pane-capabilities announcement that exist on the fork are not offered, and neither are synthesizers or docs for panes this tree does not have.
- A CONTRIBUTING subsection on message listeners from other browsing contexts belongs with a later browser-side offer.
- No `--with-deps` on the CI playwright install (that runs apt; the shared libraries Chromium needs come with the image) and no test-count pin on the bench step.
- `tests/bench_chatbuild.py` is untouched.

## Questions

None open. The one question as posted (hard-fail without a browser, or skip with a reason) is settled by the review commit: the step keeps `ROMP_UI_BENCH_REQUIRE=1` and the workflow provisions the browser itself, a loud skip in the required job having been judged worse because it would leave the only CI run of the real pane bundles green with nothing run.

### Added after review

The bench wrote three files into the copy it measured (the import-time repo-root marker, session-order.json and order-audit.jsonl), so an A/B's second run measured a copy the first had changed, and the census ran only when the run returned. Every write the kernel aims at the copy now lands in a shadow directory under the tool's private temp dir, with the kernel's own re-reads of the small JSON state files served from the shadow so the session order's read-modify-write lands once; the copy is byte-identical afterwards (the test pins it with a tree hash), and the census, which now records directories and symlink targets too, prints on every exit path, an exception out of the candidate kernel included. The one known writer the shadow does not take is the kernel's quarantine of an unparseable state file (an os.replace to a `.corrupt-<stamp>` sibling in the same directory), which the docstring names so a census listing such a file reads as that. The no-transcript error follows the 365-day backfill instead of preceding it and names the counts it worked from, and `--cwd-map FROM=TO` (repeatable) rewrites redacted registry cwds before the project-directory derivation, for a copy whose projects/ directory names kept the original path; its per-rule hit counts are read when the run ends, so they cover the builders' derivations beside discovery's. In the JSON, writes.shadowed replaces writes.atomic_writes (schema 3; `--compare` reads no schema, so a JSON of the earlier shape still compares); refused_writes and check_guards_held are unchanged.

Tier: feature

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -161,17 +161,38 @@ jobs:
         run: npm test
       - name: Build
         run: npm run build
+      - name: Cache Playwright's browsers
+        # Keyed on the lockfile, so the download below runs once per playwright version.
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-chromium-${{ runner.os }}-${{ hashFiles('vscode-extension/package-lock.json') }}
+      - name: Install the pinned Playwright Chromium
+        # The pane bench below needs a browser. It is playwright's own download for the version
+        # package-lock pins (1.62.x fetches its matching Chromium build, about 150 MB, cached above), NOT
+        # the runner image's Google Chrome: the image's browser is unpinned and auto-updated, so a Chrome
+        # release the pinned driver cannot speak, or an image without one, would have turned this REQUIRED
+        # check red with no change in the repo (review find, 2026-09-08). The same stance as the node pin
+        # above: the step never rides whatever the image has. No --with-deps: that runs apt, which the bats
+        # step stopped trusting after a mirror outage, and the shared libraries Chromium needs come with
+        # the image. tools/ui-bench.mjs prefers this browser to any system Chrome, and the step's log says
+        # which one ran ("browser: playwright chromium").
+        timeout-minutes: 5
+        run: npx playwright install chromium
       - name: Dashboard pane bench (node --test)
-        # tools/ui-bench.mjs replays a synthetic frame stream into the real feed and timeline pages in a
-        # headless Chrome (the runner image's Google Chrome through playwright's channel option, so no
-        # browser download), served by the kernel's own page route from the dist the Build step just
-        # wrote; python3 comes with the image. The test file lives at the repo root, hence the
-        # working-directory override of the job's vscode-extension/ default. On a contributor's machine
-        # the browser tests skip, and say why, when no browser, python3 or dist is present. Here
+        # tools/ui-bench.mjs replays a synthetic frame stream into the real feed and timeline pages in the
+        # headless Chromium installed above, served by the kernel's own page route from the dist the Build
+        # step just wrote; python3 comes with the image (every Ubuntu image ships one inside the kernel's
+        # 3.10 to 3.13 range, which the Python cells above cover). The test file lives at the repo root,
+        # hence the working-directory override of the job's vscode-extension/ default. On a contributor's
+        # machine the browser tests skip, and say why, when no browser, python3 or dist is present. Here
         # ROMP_UI_BENCH_REQUIRE makes that skip a failure: this is the only CI run of the real pane
-        # bundles, and a runner image that drops Chrome from PATH must turn the job red rather than
-        # green with the coverage gone (the same stance as the cryptography install above, which exists
-        # so its tests RUN).
+        # bundles, and a runner without the browser or python3 must turn the job red rather than green
+        # with the coverage gone (the same stance as the cryptography install above, which exists so its
+        # tests RUN). The replays assert only what holds under any scheduling (totals, ordering, the
+        # handoff's accounting); the timing relations are assertions only under ROMP_UI_BENCH_TIMING,
+        # which this step deliberately leaves unset: a shared runner's scheduler is not what the required
+        # check measures (review find, 2026-09-08).
         working-directory: ${{ github.workspace }}
         env:
           ROMP_UI_BENCH_REQUIRE: "1"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -161,3 +161,18 @@ jobs:
         run: npm test
       - name: Build
         run: npm run build
+      - name: Dashboard pane bench (node --test)
+        # tools/ui-bench.mjs replays a synthetic frame stream into the real feed and timeline pages in a
+        # headless Chrome (the runner image's Google Chrome through playwright's channel option, so no
+        # browser download), served by the kernel's own page route from the dist the Build step just
+        # wrote; python3 comes with the image. The test file lives at the repo root, hence the
+        # working-directory override of the job's vscode-extension/ default. On a contributor's machine
+        # the browser tests skip, and say why, when no browser, python3 or dist is present. Here
+        # ROMP_UI_BENCH_REQUIRE makes that skip a failure: this is the only CI run of the real pane
+        # bundles, and a runner image that drops Chrome from PATH must turn the job red rather than
+        # green with the coverage gone (the same stance as the cryptography install above, which exists
+        # so its tests RUN).
+        working-directory: ${{ github.workspace }}
+        env:
+          ROMP_UI_BENCH_REQUIRE: "1"
+        run: node --test tests/ui-bench.test.mjs
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -93,8 +93,11 @@ not.
 The numbers come from the real pages: the kernel's own HTTP handler serves the
 HTML, the shim, and the bundles from a `python3` subprocess under an isolated
 environment, the pattern of `tests/test_color_route.py` with the floors
-`tests/conftest.py` applies (the manager variables and the API-key variables
-are removed, the manager's key file and the boot model-catalog fetch are
+`tests/conftest.py` applies (the manager variables are removed and the manager
+port set to a dead one; every credential and key-source name conftest pops is
+removed, the API keys, the key reference and command, the token credentials,
+the auth declaration and 1Password's names; the manager's key file and the
+boot model-catalog fetch are
 pointed away, the Claude binary is `/bin/false`, the CLI scope is off, the
 postal peer bus is off, the serve token is minted for the run, and the
 subprocess exits when the bench does). Run state, the browser's profile
@@ -115,9 +118,17 @@ built by `build_session` and is too rich to fake, so record it.
 `tests/ui-bench.test.mjs` (`node --test tests/ui-bench.test.mjs`) covers the
 tool, including the recording client against a local WebSocket server and the
 Handler subprocess's isolation, and replays synthetic feed and timeline streams
-in a real browser. The browser tests skip, saying why, when no Chromium, no
-`python3` or no built `dist/` is available; with `ROMP_UI_BENCH_REQUIRE=1` in
-the environment (CI sets it) that skip is a failure instead.
+in a real browser. The browser tests skip, saying why, when no Chromium (either
+playwright's own, `cd vscode-extension && npx playwright install chromium`,
+which CI installs so the required check never rides the runner image's
+browser, or a system Google Chrome), no `python3` or no built `dist/` is
+available; with `ROMP_UI_BENCH_REQUIRE=1` in the environment (CI sets it) that
+skip is a failure instead. The replays assert
+what holds under any scheduling (frame totals, ordering, the handoff's
+accounting); the timing relations the bench measures (a settle margin, a render
+outweighing a parse, the CPU throttle's slowdown) are assertions only with
+`ROMP_UI_BENCH_TIMING=1`, which a loaded machine can fail and CI does not set;
+without it a relation that did not hold is a diagnostic line in the output.
 
 ## Test environment
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,6 +20,107 @@ The Python and shell suites are also the CI gate, across Python 3.10 to 3.13 on
 Linux; the macOS cells run on demand from the Actions tab (they are billed even
 on a public repo, so they are not part of the per-push matrix).
 
+## Measuring dashboard pane performance
+
+`tools/ui-bench.mjs` replays a pane's frame stream into the real pane page in
+a headless Chromium and reports where the browser's time goes, so a rendering
+change can be measured before and after it lands, on the same input. It needs
+the extension's `node_modules` (`cd vscode-extension && npm ci`), a built
+`dist/` (`npm run build` there), `python3`, and a Chromium: Playwright's own
+(`npx playwright install chromium` in `vscode-extension/`) or a system Google
+Chrome. The tool is POSIX-only (Linux and macOS).
+
+```bash
+# a frame stream with invented content, for a bench that needs no live kernel
+node tools/ui-bench.mjs --synthesize feed --cards 200 --out /tmp/romp-perf/synth-feed.jsonl
+# replay it into the real feed page, a frame every 100 ms, and print per-frame timings
+node tools/ui-bench.mjs --replay feed --frames /tmp/romp-perf/synth-feed.jsonl --gap 100 --json /tmp/romp-perf/before.json
+# change the bundle, rebuild, replay again, then compare the two reports
+node tools/ui-bench.mjs --compare /tmp/romp-perf/before.json /tmp/romp-perf/after.json
+# record 90 seconds of what a running kernel sends the feed page, then replay that
+node tools/ui-bench.mjs --record feed --seconds 90 --out /tmp/romp-perf/frames-feed.jsonl
+node tools/ui-bench.mjs --replay feed --frames /tmp/romp-perf/frames-feed.jsonl --json /tmp/romp-perf/live.json
+```
+
+The pane page's shim does not render a frame inside its WebSocket handler. The
+handler parses the frame, applies a view delta to the state it holds, and
+queues the result; a separate task (a `MessageChannel` message, so it runs
+while the tab is hidden too) hands the queued frames to the bundle, where a
+newer whole-state frame (a full feed, the timeline's bars or skeleton, a tab
+order) replaces an older one still queued, and the task stops after 8 ms and
+re-arms itself so input can land between slices. A wire frame and the
+bundle's render of it are therefore two measurements, and a replay reports
+both, per frame type: the bytes; the handler time (the shim's synchronous
+work per wire frame); the bundle time for each frame the shim delivered on its
+own; and the settle, the time from a frame's receipt until the main thread is
+free again (the second `requestAnimationFrame` after the delivery that
+carried it), each as p50, p90, and max. A `delivered` column beside `count`
+says how many frames of a type reached the bundle as their own delivery; the
+rest were coalesced into a newer frame's (the report counts them) or, for
+keepalives and the resync asks the shim answers itself, never left the shim.
+Frames sent back-to-back with `--fast` queue together, so most of a stream's
+deltas coalesce into one delivery and the bundle column covers only that
+one; for the bundle's cost per frame, replay at the recorded pacing or with
+`--gap 100` (a fixed gap in milliseconds between frames). On a kernel whose
+shim still renders inside the handler, the report says so and the handler
+column includes the bundle's time.
+
+After the table come the long-animation-frame entries with script
+attribution, the JavaScript heap after a forced garbage collection (and before
+it, so the line shows what the collection freed), the DOM size, and every
+console error and uncaught exception. The attribution names each task's entry point
+(the WebSocket message handler, the shim's flush task, a
+`requestAnimationFrame` callback, a timer, a script's evaluation), not the
+function inside the bundle that did the work. For that, add
+`--cpu-profile /tmp/romp-perf/feed.cpuprofile`: it samples the page's
+JavaScript with the V8 profiler across the replay, writes a file Chrome
+DevTools loads (Performance panel), and prints the functions with the most
+self and total time as `bundle.js:function:line` with the source position from
+the dist's `.map` files (a `--production` dist is minified and has none; the
+report says so), overall and inside the delivery that carried the first
+content frame and the largest frame of each type. For the hottest functions
+it also names the lines that hold the time; a forced synchronous layout, for
+instance, shows up as one line of one function owning most of its self time.
+The end-of-run layout, style, script and task counters are cumulative since
+navigation, so they include page load and idle timers (the timeline redraws
+every animation frame while it follows the present), and `--compare` shows
+them without percentages when the two runs differ in pacing or length.
+`--cpu-throttle 4` emulates a machine four times slower (the default is no
+throttling). `--iters 3` pools three runs. With `--fast`, settle times overlap
+(a frame's includes the frames queued behind it); handler and bundle times do
+not.
+
+The numbers come from the real pages: the kernel's own HTTP handler serves the
+HTML, the shim, and the bundles from a `python3` subprocess under an isolated
+environment, the pattern of `tests/test_color_route.py` with the floors
+`tests/conftest.py` applies (the manager variables and the API-key variables
+are removed, the manager's key file and the boot model-catalog fetch are
+pointed away, the Claude binary is `/bin/false`, the CLI scope is off, the
+postal peer bus is off, the serve token is minted for the run, and the
+subprocess exits when the bench does). Run state, the browser's profile
+included, lives under one per-user directory in the temp root; a run killed
+with its whole process group leaves its entry there until the next run sweeps
+it. A Node front server answers the page's WebSocket and proxies everything
+else to the subprocess.
+
+A recording holds real session data. `--record` connects to the running kernel
+as one more pane (the same URL and query, the token as the page's cookie),
+sends the ready handshake and nothing else, and writes only under the system
+temp directory (private to your user: directory 0700, file 0600), refusing a
+path inside a git checkout or through a symlink. Never copy one into the repo;
+the tests use synthetic streams. Apps: `feed`, `fleet` (the Outline pane),
+`chat`, `timeline`. Only the chat cannot be synthesized: its session frame is
+built by `build_session` and is too rich to fake, so record it.
+
+`tests/ui-bench.test.mjs` (`node --test tests/ui-bench.test.mjs`) covers the
+tool, including the recording client against a local WebSocket server and the
+Handler subprocess's isolation, and replays synthetic feed and timeline streams
+in a real browser. The browser tests skip, saying why, when no Chromium, no
+`python3` or no built `dist/` is available; with `ROMP_UI_BENCH_REQUIRE=1` in
+the environment (CI sets it) that skip is a failure instead.
+
+## Test environment
+
 Three things about the test environment are worth knowing, because all have
 produced confusing failures:
 
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1044,6 +1044,12 @@ romp-manager -f | grep romp-perf`; under launchd (macOS), `tail -f
 ~/.local/state/romp/manager.log | grep romp-perf`. Setting `ROMP_PERF=1` in the
 kernel's environment still turns it on at start.
 
+The counters describe a running kernel. To time the same builders offline, on
+a copy of a state directory and with no live kernel, `tools/perf-bench.py`
+loads a checkout's kernel in-process and reports each builder's cost on
+real-sized data; two checkouts can run against one copy for a before-and-after
+comparison. Its module docstring is the reference.
+
 ## Browser-side performance telemetry
 
 The counters above say what the kernel spent. What the browser spent on the
@@ -1171,6 +1177,12 @@ in progress in the same shape, plus a derived `p90_le` per type, `active`
 (the node test stand-ins) gets no telemetry and an unwrapped handler; every
 other browser API is behind a feature check, and nothing in the module throws
 into the pane.
+
+The telemetry describes what the panes did while people used them. To measure
+a pane change before and after on the same input, `tools/ui-bench.mjs` replays
+a recorded or synthetic frame stream into the real pane page in a headless
+Chromium and reports where the browser's time went; the "Measuring dashboard
+pane performance" section of CONTRIBUTING.md describes it.
 
 ## The API-health signal
 
--- a/tests/README.md
+++ b/tests/README.md
@@ -66,7 +66,9 @@ Every bug fix or feature change lands with a test (repo rule). Five suites:
   Run `node --test tests/ui-bench.test.mjs` from the repo root after
   `cd vscode-extension && npm ci && npm run build`. The browser tests skip,
   saying why, when no Chromium, `python3` or dist is present;
-  `ROMP_UI_BENCH_REQUIRE=1` (CI) turns that skip into a failure.
+  `ROMP_UI_BENCH_REQUIRE=1` (CI) turns that skip into a failure, and
+  `ROMP_UI_BENCH_TIMING=1` (a quiet machine, never CI) also asserts the timing
+  relations the replays otherwise report as diagnostics.
 
 **Temp files and git are hermetic, suite-wide.** Two mechanisms, one per half.
 `tests/__init__.py` wraps `tempfile.mkdtemp` so every directory the test process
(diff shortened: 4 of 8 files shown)


henrypinkard (approved):
Approved on head 0ded3dae on the maintainer's word (feature-review session).

henrypinkard (approved):
Approved on head 2b9b3aa7 on the maintainer's word (feature-review session), after the branch update onto main.

PR #991, opened 2026-09-07
Enforce the PR tier policy as a required check ("Tier policy")

henrypinkard:
Discussion: #989 (its ratifications of 2026-09-07 are folded: the `scripts/ci/` guard, the head-floor clock bound to this PR's own runs, and the reversed discussion rule).

## Summary
The maintainers' tier policy of 2026-09-07, enforced as a required status check named **Tier policy**. Rulesets cannot vary by label, so the policy is a check: its rules are a pure function over a PR record (`scripts/ci/tier_policy.py`, pinned by `tests/test_tier_policy.py`), and the workflow (`.github/workflows/tier-policy.yml`) only fetches PR data through the API and posts the verdict as a check run on the head sha. `docs/pr-tiers.md` is the reference.

| Tier | Merges when |
|---|---|
| `docs` (renamed from `tests-only`) | CI green — documentation only: `docs/**` or `*.md` anywhere, never `.github/` or `scripts/` |
| `fix` | the other maintainer's approval, **or** seven consecutive days with the head unchanged and no changes requested |
| `feature` | the other maintainer's approval |
| `major-feature` | the other maintainer's approval **and** a linked issue (`#N`) with a comment by someone other than the author (the opener alone does not count) |

**Standing and approval.** A reviewer's standing is their latest APPROVED / CHANGES_REQUESTED / DISMISSED review; a COMMENTED or PENDING review never displaces it. An approval is a standing APPROVED by a non-author with write/admin/maintain, on the **current** head. A dismissed approval never counts, whoever dismissed it. A dismissed change request clears only when the reviewer dismissed it themselves; dismissed by anyone else it stands, so the author cannot dismiss the peer's objection to reopen the seven-day path (the dismisser and the original state come from the `review_dismissed` issue event; a dismissed review with no event fails loudly).

**Guarded paths.** Any PR touching `.github/` or `scripts/ci/` needs an approval regardless of tier. A renamed or copied file counts under both its old and new path. The files API lists at most 3000 files; a PR with more is flagged and its unseen files are treated as guarded and not documentation. Zero or two tier labels fail here too.

[...]

PR #1062, opened 2026-09-08
Judge: an evidence gate for all six tiers and three correctness fixes at the judge's head

ksarma:
#1059 (the goal-store memos: `_unread`, `load_goals_shared`, the file-identity memos) merged as 9c9a926a, so this diff is its own ten commits, listed below, followed by the review commit and, since the rebase of 2026-09-09, two more: the planner's inner change gate records only a complete pass (12), and the docs (13).

The judge's six tiers (planner, closer, unblocker, grouper, consolidator, distiller) ran every discovered session in full every pass, while on a kernel with about thirty sessions two of them held anything new per pass. This PR gives every tier an evidence gate: a (tier, session) run is skipped only when every input the tier's decision path reads is identical, by identity, to what the tier last judged to completion. Around the gate sit three correctness fixes at the judge's head, found by the design review that preceded it. The ten commits are in dependency order: 1 and 4 pin the parse and its key under the pass frame, which the gate's signature (5) stands on; 8 memoizes the three reads the store tiers repeat, and 9 puts those tiers on the gate 5 built; 10 closes a hole in 9 (a side file the gate stat'd but the stage never read); 2, 3, 6 and 7 are independent of the rest. Each commit is reviewable on its own. About 4,700 changed lines in the ten, about 5,500 with the review commit and the two after the rebase, three in four of them tests. The kernel performance counters (#1003) and the pane telemetry (#1009) are on main; this PR adds no `/perf` rows of its own (see "Not included").

Rebased onto main at c7b175c9 (2026-09-09). Of the 25 PRs merged between 3e4398a3 (the #1166 merge, the base before this rebase) and this base (#1167, #1169 to #1171, #1173 to #1179, #1181 to #1192, #1195, #1199), twelve touch a file this PR changes and five bear on what it does. What moved:

- **The planner gate layering.** #1173 landed a change gate of main's own inside `_plan_session` (`_PLANNER_SEEN`, `_plan_key`: a session whose inputs have not moved since a pass that had nothing to do returns before the store read; its counters are `memos.plannerSkip` on `/perf`), and #1179 added three terms to its key (the leaf's task store, the captions file, each running launch's deadline crossing) and threaded the pass clock through `_bg_unresolved`, `_awaiting_bg_hold` and `_session_settled`, the lines commit 5 threads. Both gates are kept, one inside the other: commit 5's evidence gate around `_plan_session` keys on the inner gate's inputs (the reg by its `spawnedAt` and backend values rather than by identity, so a reg rewrite that changes neither re-arms the inner gate alone) and on `cleared.jsonl`, the death marker and the stall slice besides, and records only a complete run; the inner gate sees only the sessions the outer one ran. Commit 5 no longer adds the `now` parameters (main's since #1179), only rewords their docstrings, and its `tests/test_bg_scan_monitor.py` pin retarget is gone (main made the same change). Commit 12 fixes what the layering broke: the inner gate recorded any pass that placed nothing and left the store's key where it was, a stand-down or an unreadable side file included, so the next pass returned before the sync or the read and the evidence gate stamped that short-circuit as a complete run; three gate cases (the plan-sync stand-down, an unreadable captions file, an unreadable archive) stamped where they must stay due. The inner gate's record rule now also requires the completeness bit unset (`_judge_ctx.stage_incomplete`, which `_gated` resets before every run). Consequence for `memos.plannerSkip`: it counts the sessions the evidence gate let through, not every planner skip; an input only the outer gate keys (a `cleared.jsonl` row alone, a stall record, the death marker) re-arms the outer gate while the inner one returns before the planner runs.
- **The archive read memo.** #1171 gave the archive's readers `load_goal_archive_shared`, one loaded object per file state (`memos.goalArchive`). The review commit's `load_goal_archive` marks the running stage incomplete for a file that exists and does not read or parse, and the memo would have cached that empty answer under the file's key; a permission fix moves no key, so the failure would have been served after the file read again. `load_goal_archive` now sets `_judge_ctx.archive_read_ok` for the calling thread and the shared twin memoizes only when it is set. `tests/test_file_read_memos.py` 16 to 18: an unreadable archive and one that does not parse are the empty shape, never cached, read again on every call, one `archive-unreadable` row per episode, and cached once readable again under the same key; the gate module's archive case now asserts the second pass ran incomplete and unstamped for every tier.
- **The merged docstrings.** Commit 2's `_log_judge_error` kinds text, re-applied over main's reworded `store-unreadable` entry; commit 5's `now` docstrings on the three settle helpers; the review commit's file-memo header comment and `load_goal_archive`'s docstring, merged with #1171's shared-twin paragraph. Content unchanged except as the two points above say.
- **Tests meeting main's fixtures.** `tests/test_kernel_tick_jobs_under_frame.py` restores the judge's state root in tearDown, which main's conftest checks after every test since #1176. The plan-sync stand-down case in `tests/test_judge_stage_gate.py` writes the journal row a clear writes since #1187 beside its `cleared.jsonl` row, so both gates re-arm and the run reaches the sync.
- **Commit 13, docs only.** `docs/reference.md`'s `plannerSkip` and `goalArchive` entries describe the two gates, what the counters count and the readable-only memo; `docs/judges.md` gains a Skips entry under Ops and knobs naming the evidence gate the `gate-stamp` and `*-unreadable` kinds refer to, its signature, its stamp rule and the planner's inner gate.

The range-diff against the previous head: commits 1, 3, 4 and 6 to 10 identical; 2, 5 and 11 differ only as named here. Full suite at commit 12 on this base, `pytest tests/ -n 8`: 10313 passed, 61 skipped, 1671 subtests, 0 failed. At commit 13 the docs pins and the gate modules (thirteen modules): 294 passed.

Rebased onto main at b01fbbe0 (the #1140 merge) on 2026-09-09. One conflict, in `kernel/judge.py`: #1140 (judge concurrency is a knob) resolves every pool's width at call time, `_conc(concurrency)`, on the same lines where commits 5 and 9 replace the six triage pools' submit comprehensions with the evidence-gated loops (planner, closer, grouper, consolidator, unblocker, distiller); both kept, so the gated loops run at the setting's width. `docs/judges.md` merged on its own (#1140's pool-width note sits above this PR's error-row kinds). The eleven commits are otherwise unchanged; the review commit stays last.

Rebased onto main at 0fa3ad5d on 2026-09-09, after #1128 (romp's own API-key providers retired) merged. One conflict, in `begin_pass_frame`: this PR's commit 4 adds the `served` key to the pass frame where #1128 removes the pass-generation counter line beneath it; both kept. `docs/judges.md` merged on its own. The eleven commits are otherwise unchanged; the review commit stays last.

Rebased onto main at 66bb933e (the #1091 merge) on 2026-09-08, directly, now that #1059 is in. Of the forty PRs merged between the old base and this one (#991, #998, #1004, #1033, #1038, #1043, #1044, #1051 to #1053, #1055, #1056, #1058 to #1060, #1063 to #1070, #1072, #1074 to #1076, #1078 to #1083, #1086 to #1089, #1091 to #1093), four touch a file this PR changes. #1059 itself, re-applied on main with its review commit: its `docs/judges.md` kinds list (`frozen-store-write`, `frozen-store-save`) and its `run_propagate` mid-loop publish stand beside this PR's review commit's kinds and commit 3's docstrings. #1068 (the closer's lift horizon at the turn's last event time) edits `_close_turn`'s `apply_close` call, while commit 6 edits the caller that builds the segment index, so both stand in the same function. #1060 edits `kernel/event_model.py`'s `FileAdapter` and `declared_plan`, apart from commit 5's `_bg_expiry_t` and task-store helpers. #1056 (postal rows keyed by the recipient's stable id) is a region of `kernel/judge.py` this series leaves alone. All eleven commits applied without a conflict and the range-diff against the previous head is identical for every one; nothing was dropped, reworked or resolved by hand. On this head, on the same loaded machine: `pytest tests/ -n 8` 9859 passed, 47 skipped, 1670 subtests, 0 failed (the four browser-driven modules that failed at the earlier head passed this time); the 75 judge-side modules alone (`tests/test_judge*.py`, the twelve in the diff and their neighbours) 1507 passed; `npm test` 3455/3455, `npm run typecheck` clean; no shell surface in the diff. The two rebases before this one, both on 2026-09-08, reached db8776fb (the #1031 merge) through #1059's branch; the restack onto #1059's review commit kept the ten commits' content unchanged except two docstring splices that keep both sides (the judge-errors kinds list in `docs/judges.md` in the review commit, `run_propagate`'s in commit 3), and the second rebase carried all ten unchanged (of the twenty-five PRs it crossed, only #1029 touched a file this PR changes, one line in `kernel/judge.py`'s key gate, `KEY_CMD_TIMEOUT`, a region this series leaves alone). #1018 (the auto-nudge and retry-suppress readers proved before a rewrite) is in the base; it reworks the kernel's `auto-nudge.json` reader, and on the judge side commit 10's `stalled_facts` reads that file, so a file #1018 has moved aside reads as absent there, the unmarked path. The first rebase, onto the #1019 merge, is the one that shaped the series. Main's goal-store quarantine (#1019: a goals file that exists and cannot be read raises into a per-session boundary, `load_goals_or_fault`, and unparseable bytes are moved aside) covers the data loss two further commits of this series addressed by refusal, so both came off: the `save_goals` refusal of a store that loaded as a fallback for an unreadable file, `_disk_rev`'s None answer, the archive mark and its refusal, the six stages' stand-downs, the kernel's undo-clear stand-down with its `undoClearResult` frame, and the once-per-episode warn frame. With them gone this PR touches no `kernel/kernel.py` and no `ui/` runtime file (one TypeScript source pin is retargeted). Commit 9 was reworked onto #1019: `_mark_unread` keeps the one fallback load left (a parsed store whose override journal exists and did not read); its docstring and the gate's input inventory say that a store file that cannot be read raises at the stage's load (the run counts incomplete and the runner files its `pass-crash` row) and that an unparseable one is moved aside; and its unparseable-store test became two, a store that cannot be read never stamps, and a store that does not parse is moved aside and the fresh store judged to completion. Commit 4's two new `_PARSE_CACHE.clear()` sites clear the chain memo beside them, the rule main's corpus scan enforces. Two docstring conflicts (the judge-errors kinds list in commit 2, `run_propagate`'s in commit 3) were resolved by keeping both sides. #1016 (hidden panes paint once on return), #1020 (state readers proved before a flag, order or bell write) and #1024 (main-drift converge) touch nothing this PR changes.

## Commits

1. **Judge: pin a cache-hit parse under the pass frame.** `parsed_session` pinned the pass frame's parse only on the `_PARSE_CACHE` miss path; a hit returned the cached parse unpinned, so a warm session (every idle one, most of the time) froze nothing under an open frame, and a turn that ended after the pass's first touch was visible to a later stage and invisible to an earlier one. The hit path now pins into the same frame. At this commit the pin also reached the kernel's six tick jobs that call `parsed_session` while a frame is open (`_frame` is a module global); the review commit scopes the pin to pass threads, so those jobs read the live world at every cycle (see "Review commit"). Accepted lag, for the judge stages: a turn that ends after the pass's first touch is judged whole in the next pass, for warm sessions as the miss path already implied for cold ones. A judgment-consistency fix, no figure. `kernel/judge.py`; `tests/test_judge_pass_frame.py` +3, `tests/test_kernel_tick_jobs_under_frame.py` (1, new, at this commit: the kernel's `_clear_done_working_notes` under an open frame keeps a working note across a mid-pass append and lifts it next pass; the review commit rewrites the module to two cases that assert the opposite for a tick job, a lift at the cycle the turn ends).

2. **Judge: drop the rolledUp marker when a container dissolves, and repair stores published with it.** Umbrella dissolution promoted a container's children to top level without clearing `rolledUp`, and both re-derivers stand aside for a rolled-up node, so a promoted child kept its stale done flag and every rollup appended one settle row and one seam. The dissolution now unrolls each promoted child's whole rolled-up subtree through one helper (`_unroll_node`, the code `_reopen` already ran inline), `_materialize_from_log` repairs stores already published with the marker (one publish at the next rollup, then no-ops), and a top whose diary holds the bug's settle rows gets one romp-authored reopen row plus one judge-errors row (`unroll-heal`) so it is judgeable again. Invariant: the marker means a resolved ancestor's roll-down set these flags, and it is dropped exactly when no resolved ancestor remains. Measured on the self-hosted kernel by inspecting live stores, not benched: one node at `LOG_CAP` with 64 settle rows, and 0.6 to 1.1 MB republished per pass per affected store once its session revives. `kernel/judge.py`, `docs/goal-state.md` (rule 6); `tests/test_judge_dissolution_rolledup.py` (17, new).

3. **Judge: keep a loaded store CAS-protected across every save.** `save_goals` popped the CAS base (`_baseRev`) on every real publish and never restored it, so every later save of the same store object (`_plan_session` saves several times per pass; `_distill_session` saves after titling and again after distilling, with the model call between) ran unconditionally and overwrote concurrent publishes. After a successful publish the object now carries the written revision as its base (the review commit restores the loaded base when the publish did not happen, so a raise inside the loop or at the rename leaves the holder CAS-protected too); the file still never holds the key; stores built without `load_goals` are unchanged. A lost-write fix, no figure; a secondary, unmeasured effect is that later saves in a pass reach the no-op check. `kernel/judge.py` (also rewords `run_propagate`'s docstrings; its drop-after-publish stays); `tests/test_judge_store_cas.py` +3, `tests/test_judge_store_noop_publish.py` +2, `tests/test_judge_distill_early_save.py` (1, new).

4. **Judge: pin the parse key with the parse under a pass frame.** The frame pinned a session's parse but never the (fileset key, pending cut) pair it was read under, so a stamp keyed on the transcript had nothing exact to stand for. `_frame_parse_key(fsid, files)` pins the pair before anything is read (the first toucher wins; a failed stat pins None); `parsed_session` pins hit and miss into the same frame and records the pair each parse was served under; `tasks_for` keys its disk memo on the pinned pair instead of its own stat, which closes a caption-memo race. No figure of its own: it is what makes the gate's stamp exact. One-time cost: the memo key changes shape, so every judge-units-cache entry regenerates once (one parse per session on the first pass). `kernel/judge.py`; `tests/test_judge_pass_frame.py` +12 (one case guards commit 1's hit-path pin by moving the cache slot mid-pass, since commit 1's own case stops depending on it once the key is pinned).

5. **Judge: an evidence gate for the planner and closer.** A (tier, session) run is skipped only when every input the tier's decision path reads is identical by identity to what it last judged to completion: the pinned parse pair and its candidate files, the store, the override journal and the archive, and per tier the captions, episodes, death marker, cleared rows, the registry's `spawnedAt` value, the session's stall records and the leaf's task store. A deferred, failed, empty-reply, raising or cut run leaves no stamp (`_judge_run` is now a belt over `_judge_run_impl` that marks an empty reply incomplete); a background launch's deadline is the one clock input, through one shared expiry helper (`em._bg_expiry_t`); a skip still stamps the pass watermark; the stamps are process state, so the first pass after a restart is a full walk. Invariant: the gate never withholds a verdict the ungated pass would have filed from new evidence, because every writer re-arms by moving an identity. Since the rebase onto c7b175c9 this gate sits around main's own change gate inside `_plan_session` (commit 12 and the dated paragraph above): the outer gate keys on the inner gate's inputs and more (the reg by two values rather than by identity), so the inner gate sees only the sessions the outer one ran. Measured on the self-hosted kernel (31-session state copy, judge paused, sealed placements, two warm-ups then six timed passes, six-worker pools, `process_time` per pass; the harness is ad hoc and not in the repo): planner+closer idle CPU 1.303 s to 0.494 s per pass, store loads and saves 62 to 6 per pass, whole triage pass 2.251 s to 1.437 s; planner skipped 81% of sessions, closer 100%. The six planner sessions that ran every pass held units the paused model answered with an empty reply, which the gate refuses to stamp; live, those place and stamp. `kernel/judge.py`, `kernel/event_model.py`; `tests/test_judge_stage_gate.py` (43, new); two source pins retargeted (`tests/test_distill_tier.py`, `ui/webview/feed-limit-banner.test.ts`).

6. **Judge: build the closer's segment index on the first turn it judges.** The seam-aware index (its only consumer is `_close_turn`'s goal-history block) is built on the first judged turn instead of before the walk, so an all-swept session pays no `_segs` call per turn; the block handed to the model is unchanged because nothing between the store load and the first judged turn writes seams (`rollup_status` stamps them after the walk). About 0.06 CPU-s of an idle pass, measured on the self-hosted live kernel with about thirty sessions. `kernel/judge.py`; `tests/test_judge_close_riders.py` +3.

7. **Tests: the courier dead-wait test removes the journal row and nudge record it leaves.** Its tearDown now removes the override-journal row and the auto-nudge record the test writes under the shared placeholder sid, and a cleanup registered in setUp asserts both are gone; without the removal, sixteen `test_judge.py` distiller and procedural-block tests fail whenever xdist places them after it in a worker, which the gate module of commit 5 does. Reproduced by running the two modules in one process: 16 failed before, 0 after. No behaviour change. `tests/test_courier_link.py`.

8. **Judge: three identity memos on the store tiers' read paths.** `_view_cleared` and `_live_prompt_since` are memoized on their file's `(ino, mtime_ns, size)`, exact because every writer of `cleared.jsonl` and `states/<sid>.jsonl` appends (the commit lists the writers; an unreadable file is never memoized, an absent one drops the entry, `_view_cleared` returns a frozenset); the distiller's parent-to-children map is built once per store (`_kids_map`, threaded to `_distill_due_t` with `kids=None` keeping the kernel's callers unchanged). Measured on the self-hosted kernel (the bench of commit 5): the four store tiers' idle CPU 0.60 to 0.44 s per pass with loads unchanged at 124 (distill 0.210 to 0.108, group 0.117 to 0.079, consolidate 0.109 to 0.077, unblock 0.155 to 0.136), and planner+closer 0.69 to 0.62 s with no code change of their own. Re-measured on this branch with a synthetic micro-bench (min of five, `process_time`): 690 `_distill_due_t` calls over a 104-node store 9.4 to 2.2 ms; 217 `_view_cleared` calls over a 500-row `cleared.jsonl` 202 to 2 ms; 31 `_live_prompt_since` calls over 2000-row states logs 106 to 0.3 ms. `kernel/judge.py`; `tests/test_judge_identity_memos.py` (12, new).

9. **Judge: the unblocker, grouper, consolidator and distiller run through the evidence gate.** Each is keyed on its own decision path's inputs: the unblocker on the pinned parse pair plus the store trio; the grouper and consolidator on the store trio plus `cleared.jsonl`, no parse; the distiller on the store trio, the states file and this session's stall records, no parse. A completeness mark sits at every no-write deferral, and `_mark_unread` marks the one fallback load left: a parsed store whose override journal exists and did not read. A store file that cannot be read raises at the stage's load (#1019, on main), so the run counts incomplete and the runner files its `pass-crash` row; unparseable bytes are moved aside and the fresh store is judged. One rule change, for the planner and closer too: a frame that served no parse for a session now lets a parse-tier stamp stand (the stage's own read would have pinned a pair, so no served pair means the decision depended on the store alone; no frame at all still means no stamp for a parse tier, while the store-only tiers stamp with or without one), which stops the unblocker bypassing about half the sessions every pass, at the cost of re-arming those sessions on every transcript append (about 10 ms per pass on the bench). Measured on the self-hosted kernel (the same bench, before = commit 8): the four tiers' idle CPU 0.436 s to 0.165/0.182 s per pass and their loads 124 to 20; whole triage 1.371/1.769 to 1.130/1.187 s and 199 to 96-98 loads; skipped: grouper 90%, consolidator 97%, distiller 100%, unblocker 48% (a paused-bench floor: 16 sessions hold blocks the paused model cannot examine). `kernel/judge.py`; `tests/test_judge_stage_gate.py` 43 to 68.

10. **Judge: a side-file read that fails after the gate stat'd it marks the run incomplete.** `_live_prompt_since`, `_view_cleared` and `stalled_facts` answered empty on any `OSError` without marking the stage, so a complete-looking run stamped the identity of a file the stage never read, and the session skipped until that file moved, which a permission bit, an EMFILE or an EIO never makes it do. All three now go through `_read_failed`: the stage is marked incomplete and one judge-errors row (`states-unreadable`, `cleared-unreadable`, `stall-unreadable`) is logged per failure episode; `stalled_facts` gains `strict`, so the signature's own failed read runs the stage without a stamp; the review commit puts the four remaining signature files on the same path (see "Review commit"). An absent file ends an open episode as a good read does, so a side file removed while unreadable and recreated unreadable logs a new row. New behaviour: an `auto-nudge.json` that exists and does not parse now writes a row where it was silent. A gate correctness fix, no figure. `kernel/judge.py`; `tests/test_judge_stage_gate.py` 68 to 81.

11. **Review fixes: an unreadable signature is a fault not silence, the warm pin covers only the judge stages, the error rows and docs agree.** The review commit, not one of the ten; what it changed is summarised below. `kernel/judge.py`, `docs/judges.md`; `tests/test_judge_stage_gate.py` 81 to 86, `tests/test_judge_pass_frame.py` 20 to 22, `tests/test_kernel_tick_jobs_under_frame.py` 1 to 2 (rewritten), `tests/test_judge_store_cas.py` (one case rewritten), and since the rebase `tests/test_file_read_memos.py` 16 to 18 (the archive memo, see the dated paragraph).

12. **Judge: the planner's change gate records only a complete pass.** After the rebase onto c7b175c9 (the dated paragraph). `_plan_session`'s record rule, main's inner gate, requires the completeness bit unset besides placing nothing and leaving the store's key where it was: a pass that stood down without a write, or read a side file that exists and did not read, is planned again next pass instead of recorded. Without it a recorded incomplete pass short-circuited the next run before the sync or the read, and the evidence gate stamped that short-circuit as a complete run. The gate's header comment names the two gates and their order. `kernel/judge.py`; `tests/test_planner_skip.py` 16 to 17, `tests/test_judge_stage_gate.py` (one case rewritten; 86), `tests/test_kernel_tick_jobs_under_frame.py` (tearDown).

13. **Docs: the planner's two gates, what plannerSkip counts, and the archive memo's readable-only rule.** `docs/reference.md`, `docs/judges.md`; no code.

**Review commit.** Six changes, in the commit's terms. The pass frame's pin, a warm cache hit too, applies to pass threads only: the thread that opened or joined the frame and this module's pool workers (`_pass_frame`, `_TimedPool.submit`, `begin_pass_frame`, `end_pass_frame`); the kernel's pusher tick jobs read the live world at every cycle instead of the pass-start world for the tiers' whole run, model calls included. The four signature files whose readers answered empty on a read failure go through `_read_failed` like the states file, `cleared.jsonl` and the stall records: `_prompt_gist` (`captions-unreadable`), `_episode_read` (`episodes-unreadable`; a failed read is no longer memoized under the file's mtime), `_death_marker` (`marker-unreadable`), `load_goal_archive` (`archive-unreadable`; since the rebase its shared twin memoizes only a read that succeeded); an absent file ends an episode. The `gate-stamp` judge-errors row wears the judge's name (`_TIER_JUDGE`), not the tier key. `docs/judges.md`'s judge-errors kinds list and `_log_judge_error`'s docstring carry `unroll-heal`, `gate-stamp` and the seven `*-unreadable` kinds. `save_goals` stamps the base again in a `finally`: the written revision after a publish, the loaded base when the publish did not happen, so a raise inside the CAS loop or at the rename keeps the holder's retry CAS-protected; the `_unread` mark is restored with it. In the tests, the `FsCompleteness` allowlist drops `session-flags.json` (no stage reads it; the runners filter their lists on it before any stage runs), and a new case pins the hidden-session eviction the gate's input inventory states: a muted sid's planner, closer and unblocker stamps are evicted, the store tiers keep theirs, and an unmute costs one full run before the skips resume. The sections below describe this head; where the review commit changed a claim, the change is named.

## What changes for a user

Cards, lanes and chat pages render the same rows from the same data, and the gate's tests pin that a skip never withholds a verdict a full pass would have filed. Four things are visible:

- A goal store already published with the stale `rolledUp` marker gets one repair publish at its next rollup, and a top whose diary holds the bug's settle rows gets one reopen row and one judge-errors row; its card moves from a stale done, or a Working that every settled-done reader treated as sealed, to an open top the closer re-rules. Once per node (commit 2).
- A turn that ends after a pass's first touch of a warm session is judged in the next pass, whole, rather than by the later stages of the current one (commit 1). The kernel's tick jobs (a needs-you block, a working-note lift, a nudge decision) are not held to that world: they read the live transcript at every cycle (the review commit).
- The first judge pass after a kernel restart is a full walk; after that an idle session costs each tier a signature check per pass (a few stats and two small value reads), with no store load, parse or save (commits 5, 9). The planner's inner change gate (on main since #1173) sees only the sessions the evidence gate ran, so `memos.plannerSkip` on `/perf` counts those alone (commit 12). Every judge-units-cache entry regenerates once after commit 4 lands.
- Seven new judge-errors rows for signature files that exist and do not read or parse (`states-unreadable`, `cleared-unreadable`, `stall-unreadable` from commit 10; `captions-unreadable`, `episodes-unreadable`, `marker-unreadable`, `archive-unreadable` from the review commit), one per failure episode; each fires where its reader was silent, and the run stamps nothing until the file reads.

## Tests

Red first: every commit's tests were run against the code before its change, then green after it. After that, each commit's finished code was mutated one line at a time (a mark removed, a guard inverted, a key component dropped, a rule reverted) and a test was added for every mutation the existing cases let through; each such case fails under its mutation and passes on the commit. The counts below include those cases.

- Commit 1: the warm-first-touch case fails at its second-touch `assertIs` (a different parse object, with the appended turn ended); the no-frame control passes on both; the concurrent first-toucher case fails when the hit path's pin is an overwrite instead of a setdefault; the kernel tick-job module fails on the code before the change (the job re-parses mid-pass and lifts a working note under the open frame). After: 8/8 and 1/1 in the two modules. The review commit rewrites the tick-job module to two cases: under an open frame the job reads the live file and lifts the note at the cycle the turn ends, and a pass thread's pin does not move it.
- Commit 2: 8 of the first 12 cases fail before (the marker still on the promoted top, settle rows appended through the file, no reopen row, `rev` advancing across rollups); the other 4 pin behaviour that must not change; 5 more pin the heal's edges (the reopen's `ev_t` floor and its latest-evidence moment, a view-cleared top staying sealed, a solid ancestor's stale flag rewritten before the ancestor check, a rolled child with no diary key unrolling to open). After: 17/17.
- Commit 3: the second-save test fails (the other writer's node is gone) and the distiller test fails (the block published during the model call is dropped by the tail save); the re-stamp's edges are pinned in both store modules (a second uncontended save never rebases, a raise inside the CAS loop leaves the object base-less, a hand-built store stays unconditional, a loaded store's second unchanged save is a no-op); the review commit rewrites the raise case: the base is kept at the loop and at the rename, and a retry after a concurrent publish rebases instead of stomping. After: 20 across the CAS and distiller modules; 53 across the CAS and no-op modules.
- Commit 4: 9 of the first 10 cases fail (seven with `AttributeError`, no `_frame_parse_key` or `_parse_key_files`; two on the missing pinned key). The tenth, a warm hit's pin recording its served pair and holding when the cache slot moves, passes at this commit and fails with the hit path returning unpinned. It guards commit 1's pin, which commit 1's own case stops exercising once the key is pinned (an unchanged slot answers the same object either way). Two more pin that a losing parse never overwrites the winner's served pair and that a pair that cannot be computed memoizes nothing. After: 20/20.
- Commit 5: the 43 gate cases fail in setUp (no `_judge_run_impl`), and the two kernel-side pins fail on the old text. After: 43/43 in the module. Ten of the cases each pin one rule against a one-line mutation (the belt, the per-run reset of the completeness bit, the two pending marks alone, the failed-call site alone, the candidate files in the signature, a failed side-file stat and an unlistable task dir bypassing, an unframed runner never stamping, a stamping failure not being a failed pass). The `feed-limit-banner` source pin slices the model-call body and failed on the belt split; it is re-anchored to `_judge_run_impl` in this commit (fails before, 6/6 after).
- Commit 6: the all-swept case counts 2 eager `_segs` calls where 0 are expected; the equality control passes on both; the once-per-walk case fails when the index is rebuilt per judged turn. After: 9/9.
- Commit 7: no new behaviour. The leak reproduced by running `test_courier_link.py` then `test_judge.py` in one process: 16 failed before the change; 533 passed at this head. The setUp cleanup fails the test on its own when either unlink is dropped.
- Commit 8: 6/6 of the first cases fail with `AttributeError` (no `_kids_map`, `_view_cleared_scan`, `_live_prompt_since_scan`); 6 more pin one property each (one kids map per store, a same-tick append seen, a row landing after the read seen next call, an unreadable `cleared.jsonl` not memoized, the rebind clearing the live-prompt memo). After: 12/12.
- Commit 9: 25 fail before (the new store-tier cases plus the eviction and planner/closer signature cases this commit rewrote: the four tiers ungated, so the skip pass still loads and every counter reads 0). After: 68/68. Then fifteen mutations of the finished code, one at a time (a completeness mark removed, the served-pair rule reverted, a file dropped from a tier's inputs, every tier bypassing without a frame, a store tier submitted with the clock input, `pass_done` dropped from the run path), each failing its named test.
- Commit 10: the three unreadable-side-file cases fail before (the tier stamps over the unread file; the gate skips over an unreadable stall file); each case's removal leg (the file removed while an episode is open, then recreated unreadable, a new row) fails without the absent-branch `_read_ok`. The six stage-site tests pass at commit 9, which landed the marks; their red is the mutation set (each mark replaced by `pass`, each test fails). Four more pin a stall read failing after a good signature read, the rebind clearing the failure episodes, and a file vanishing between the stat and the read (cleared and states) ending an open episode. After: 81/81.
- Review commit: `SignatureFileReads` (4 cases: an unreadable captions file never stamps the planner, an unreadable episode log never stamps the planner and is not memoized, an unreadable death marker never stamps the closer, an unreadable archive never stamps any tier; one row per failure episode, and the heal or the run lands once the file reads), `Bounds`' muted-sid case, the `gate-stamp` row's judge name in `Completeness`, two pass-frame cases (a thread outside the pass reads live and pins nothing; a pool worker and a joiner are pass threads), the two tick-job cases, and the CAS case above. `tests/test_judge_stage_gate.py` 81 to 86, `tests/test_judge_pass_frame.py` 20 to 22, `tests/test_kernel_tick_jobs_under_frame.py` 1 to 2. Since the rebase, two archive-memo cases in `tests/test_file_read_memos.py` (16 to 18): an unreadable archive and one that does not parse are never cached and read again on every call, one `archive-unreadable` row per episode; without the `archive_read_ok` check both fail at their first memo assertion (the failed read is memoized under the file's key).
- Commit 12: on the rebased head without the rule, three gate cases (the plan-sync stand-down, an unreadable captions file, an unreadable archive) stamp where they must stay due; with it, 86/86. `tests/test_planner_skip.py` gains one case at the gate's own level: a pass with the completeness bit set is planned again next pass and never recorded, and recorded once complete (fails with the bit dropped from the rule). The plan-sync stand-down case writes the journal row a clear writes since #1187 beside its `cleared.jsonl` row, so both gates re-arm and the run reaches the sync.
- Commit 13: docs only. The docs pins and the gate modules (`tests/test_perf_stats.py`'s memo keys, `tests/test_judge_store_cache.py`'s kinds list, `tests/test_reference_billing_label.py`, `tests/test_postal_hydrate_scope.py`, `tests/test_path_links.py`, the three gate modules and five neighbours): 294 passed.

Full suites at the head of the ten commits, before the review commit, on a loaded machine that self-hosts a kernel: `pytest tests/ -n 8` 8975 passed, 47 skipped, 1403 subtests, 4 failed: `test_paste_to_focus_browser::ServedPaste`, the two `test_scroll_marks_growth::ServedNotchFollowsGrowth` tests and `test_awaiting_box_sync::ServedSync`, Playwright-driven modules that fail identically at the base commit on this machine (the hermetic kernel they start has no Agent SDK) and pass in CI. The twelve test modules in the diff alone: 219 passed (227 cases at the review commit's head; the eight it adds were not part of this run). `vscode-extension`: `npm test` 3274/3274, `npm run typecheck` clean. No shell surface changed (no `.bats` file or script in the diff). At commit 12 on the current base: `pytest tests/ -n 8` 10313 passed, 61 skipped, 1671 subtests, 0 failed.

## Measurements

Every judge-pass figure is the self-hosted deployment's measurement, not re-measured on this branch: the harness is ad hoc (not in the repo) and runs over private copies of a 31-session state directory with the judge paused (no model calls), placements sealed and every end-known turn marked swept, two warm-ups then six timed `run_triage` passes on real six-worker pools, CPU = `process_time` over the pass, legs back to back. The three rows marked "this branch" are a synthetic micro-bench run here. Nothing was re-measured after the review commit or the rebase; the idle-pass figures were taken before main's inner gate (#1173) existed, and that gate runs only for the sessions the evidence gate lets through.

| Change | Before | After | How |
|---|---|---|---|
| Planner + closer, idle pass (commit 5) | CPU 1.303 s; loads 62; saves 62 | 0.494 s; 6; 6 | deployment bench; skipped: planner 81%, closer 100% |
| Whole triage pass (commit 5) | CPU 2.251 s; loads 255; saves 62 | 1.437 s; 199; 6.5 | same legs |
| Closer segment index (commit 6) | one `_segs` call per turn per pass | none for an all-swept session | self-hosted live kernel, about thirty sessions: about 0.06 CPU-s per idle pass |
| Four store tiers, idle pass (commit 8, memos alone) | CPU 0.604/0.573 s; loads 124 | 0.436 s; 124 | deployment bench, two legs; per tier unblock 0.155 to 0.136, group 0.117 to 0.079, consolidate 0.109 to 0.077, distill 0.210 to 0.108 |
| `_distill_due_t` x690, 104-node store (commit 8) | 9.4 ms | 2.2 ms | this branch, synthetic, min of five |
| `_view_cleared` x217, 500-row `cleared.jsonl` (commit 8) | 202 ms | 2 ms | this branch, synthetic, min of five |
| `_live_prompt_since` x31, 2000-row states logs (commit 8) | 106 ms | 0.3 ms | this branch, synthetic, min of five |
| Four store tiers, idle pass (commit 9, over commit 8) | CPU 0.436 s; loads 124 | 0.165/0.182 s; 20 | deployment bench, two pairs in opposite order; skipped: group 90%, consolidate 97%, distill 100%, unblock 48% (paused floor) |
| Whole triage pass (commit 9) | CPU 1.371/1.769 s; loads 199 | 1.130/1.187 s; 96 to 98 | same legs |
| Store with the stale `rolledUp` marker (commit 2) | 0.6 to 1.1 MB republished per pass per affected store; one node at `LOG_CAP` | one repair publish, then no-ops | self-hosted live stores, inspected, not benched |

On the self-hosted deployment where these changes first ran, after they landed together with the store-side and pusher-side changes that went in as separate PRs (#1059 and #1060, both on main now, so not attributable to this PR alone), at the same pass cadence: the judge threads went from 81% to 35% of a core steady (38 to 43% in windows where passes ran at 7 to 9 per 30 s), the planner and closer skipped 97% of sessions steady, and goal-store loads per pass fell from about 455 to 246 plain loads plus 98 served by the shared cache.

## Contracts changed

- `parsed_session` pins a cache hit into the open pass frame as the miss path did, for pass threads only: `_pass_frame()` is the one reader of `_frame`, and answers it for the thread that opened or joined the frame (`begin_pass_frame` marks `_judge_ctx.in_pass`; the creator's `end_pass_frame` clears it) and for this module's pool workers (`_TimedPool.submit` marks every run); any other thread, the kernel's pusher included, reads the live world and pins nothing (the review commit). The frame (`_frame`) gains `"keys"` (the pinned (fileset key, cut) pair per session) and `"served"` (the pair each pinned parse was made under). `_frame_parse_key(fsid, files)` returns `(pair, cut, frame)`; `_pinned_fileset_key` is removed; `tasks_for`'s memo key is the pinned pair, so every judge-units-cache entry regenerates once.
- `_judge_run` is a belt over `_judge_run_impl`, which holds the model call, the billing resolution and the login gate. A test that patches `_judge_run` bypasses the completeness belt; patch `_judge_run_impl` to keep it.
- New module state in `kernel/judge.py`: `_STAGE_STAMP`, `GATED_TIERS`, `PARSE_TIERS`, `_TIER_JUDGE` (tier key to judge name, so the `gate-stamp` row wears the judge's name like every other row), `_TIER_STATS` with `tier_stats()` (ran, skipped, stamped, bypassed, incomplete, due_clock per tier plus the stamp count; read by tests only, not exported), `_VIEW_CLEARED_MEMO`, `_LIVE_PROMPT_MEMO`, `_UNREADABLE_LOGGED`; `_rebind_state` clears them all. `_judge_ctx.stage_incomplete` is the completeness bit; `pass_done` is stamped on a skip.
- The planner's inner change gate (main's `_PLANNER_SEEN`, `_plan_key`, `_plan_session`'s record rule) records a pass only when `_judge_ctx.stage_incomplete` is unset as well (commit 12); its table, key and `planner_skip_stats()` are unchanged in shape. `load_goal_archive` sets `_judge_ctx.archive_read_ok` for the calling thread (True on a read that succeeded or an absent file, False on a file that exists and does not read or parse), and `load_goal_archive_shared` memoizes only when it is True. No signature changes.
- `_bg_unresolved(path, now=None)`, `_awaiting_bg_hold(..., now=None)` and `_session_settled(..., now=None)` take the pass's clock (on main since #1179; this PR's gated planner and closer hand theirs in); `em._bg_expiry_t` is the one expiry definition (`_bg_expired` and `_settle_not_before` share it); `em.task_store_dir` and `em.task_store_fp` are new.
- `_materialize_from_log(nodes, store=None)`; `_unroll_node`, `_unroll_subtree`, `_heal_settle_without_done` and `_resolved_ancestor` are new; `_reopen` uses `_unroll_node`. New judge-errors kinds: `unroll-heal`, `gate-stamp` (a complete run whose stamp could not be written; the session stays due) and seven `*-unreadable` kinds (`states-unreadable`, `cleared-unreadable`, `stall-unreadable`, `captions-unreadable`, `episodes-unreadable`, `marker-unreadable`, `archive-unreadable`); `docs/judges.md`'s kinds list and `_log_judge_error`'s docstring carry them all. `_prompt_gist`, `_episode_read`, `_death_marker` and `load_goal_archive` keep their empty answers for a file that exists and cannot be read or parsed, and now mark the running stage incomplete through `_read_failed`; `_episode_read` no longer memoizes a failed read under the file's mtime, and neither does `load_goal_archive_shared`.
- `save_goals` stamps `_baseRev` again in a `finally` for stores that carried one: the written revision after a publish, the loaded base when the publish did not happen (a raise inside the CAS loop, at the write or at the rename), and the `_unread` mark is restored when nothing was published; a refusal before the pop leaves the object untouched (pinned by `ReadFaultCas`). `_mark_unread(store, reason)` sets the `_unread` mark (`"journal"`, the one reason left since #1019) and marks the running stage incomplete; `_read_failed` and `_read_ok` keep the per-episode set, and an absent read ends an episode as a good one does.
- `_view_cleared()` returns a frozenset (the seven callers test membership; none mutates). `_distill_due_t(store, nid, blocked, kids=None)`; `_kids_map(nodes)`. `stalled_facts(fsid, strict=False)`: strict raises `OSError` for a file that exists and does not read.
- `docs/goal-state.md` rule 6 names container dissolution as the second un-resolver of `rolledUp` children. `docs/reference.md`'s `plannerSkip` and `goalArchive` entries and `docs/judges.md`'s Skips entry describe the gates as they stand (commit 13).

## Not included

- No `GET /perf` or `romp perf` rows of this PR's own. `memos.plannerSkip` (main's, #1173) now counts only the sessions the evidence gate ran, so on a quiet kernel its `skipped` stays low while the evidence gate does the skipping; `tier_stats()` holds the evidence gate's per-tier counters, and the memo counters are module state read by their tests. Wiring them into the counter route #1003 landed is a follow-up PR that wires every judge counter at once. `save_goals`' no-op hash timing and the kernel-side judge wake counters wait for the same PR.
- #1019 (on main) made two further commits of this series moot: a goals file that exists and cannot be read now raises into the per-session boundary, and unparseable bytes are quarantined aside, so the `save_goals` refusal of a fallback store, the kernel's undo-clear stand-down and the once-per-episode warn frame were dropped in the rebase (see the note under the summary). One hole #1019 leaves is taken up here only on the judge's side: since the review commit `load_goal_archive` marks the running stage incomplete and logs an `archive-unreadable` row for a file that exists and cannot be read or parsed, so no tier stamps over an archive it never read, and since the rebase the readers' shared memo does not keep that answer either. Its answer is still the empty archive, and `save_goal_archive` overwrites blindly, so the kernel's compaction pass and undo-clear restore, which read through the same function, would still publish over an unreadable archive. A raise into their paths wants #1019's shape (a read fault raises into a boundary, unparseable bytes are moved aside), in a PR of its own.
- The inner change gate stays as main landed it (its table, key terms and counters); this PR only aligns its record rule with the completeness bit. Retiring it in favour of the evidence gate is Question 1.

## Questions

1. The planner now runs behind two gates, one inside the other. The evidence gate around `_plan_session` keys on the inner gate's inputs and more (the reg by its `spawnedAt` and backend values rather than by identity) and records only a complete run, so in a framed pass the inner gate sees only the sessions the outer one ran and `memos.plannerSkip` counts those alone; outside a pass frame (`romp-judge --plan`) the inner gate is the one that skips, since a parse tier never stamps there. Both are kept here, and commit 12 aligns the inner gate's record rule with the completeness bit. Keep both, or retire the inner gate (its table, key and counters) once the evidence gate's per-tier counters reach `/perf` in the follow-up?
2. Commit 9's served-pair rule applies to the planner and closer too: a run that never called `parsed_session` for a session now stamps where it previously bypassed. The exactness argument is in the commit (the stage's own read would have pinned a pair); the cost is a re-arm on every transcript append for sessions with no blocked candidate. Is that the rule you want, or should the parse tiers keep bypassing?
3. Commit 1's lag for warm sessions (a turn ending mid-pass is judged whole in the next pass) is stated, not measured. The review commit removed it for the kernel's tick jobs, which read live; the question is about the judge stages only. Acceptable as is?
4. Closed by the review commit. Commit 3 left one window documented: a raise inside `save_goals`' CAS loop (after the base is popped, before the rename) left the store object without a base, so the holder's next save was unconditional. The base is now restored in a `finally` when the publish did not happen, so the retry stays CAS-protected, and `ReadFaultCas` pins it at the loop and at the rename.
5. `_materialize_from_log(nodes, store=None)`: the store is optional so callers that pass nodes only keep working. Would you rather it be required?
6. The per-tier counters (`tier_stats()`) arrive ahead of any route that reports them. Keep them for the follow-up that wires the judge counters into #1003's route, or land them with the wiring?

Tier: feature

### DIFF
--- a/docs/goal-state.md
+++ b/docs/goal-state.md
@@ -163,9 +163,11 @@ within a beat it reverts with a toast, never silently.
 5. Every verdict goes through `record_verdict`. A reopen carries `unblock`
    events for blocked descendants and ancestors; every unblock is an
    event.
-6. Reopens un-resolve only auto-rolled (`rolledUp`) children;
-   genuinely-done leaves keep their state. Re-completions re-summarize
-   only the new stretch (`deltaSince`).
+6. Reopens un-resolve only auto-rolled (`rolledUp`) children, and so
+   does the dissolution of a container above them (the marker mirrors a
+   resolved ancestor and is dropped once none remains); genuinely-done
+   leaves keep their state. Re-completions re-summarize only the new
+   stretch (`deltaSince`).
 7. Identity: identical prompts in different turns are different work
    (twins); same-second identical bursts plan once; any seg-id-derivation
    change ships a placements migration (`placementsV`, currently v12);
--- a/docs/judges.md
+++ b/docs/judges.md
@@ -472,8 +472,17 @@ permission/API-error floors: one interrupt at a time, the present event first.
   store-unwritable: a goals file whose publish failed under a user gesture,
   store-quarantined: a goals file whose bytes did not parse, moved aside,
   frozen-store-write: a read-only site wrote to the shared store view, naming
-  the site, after which the shared cache is off for the process, and
-  frozen-store-save: a shared store view was handed to `save_goals`, refused).
+  the site, after which the shared cache is off for the process,
+  frozen-store-save: a shared store view was handed to `save_goals`, refused,
+  unroll-heal: a top left rolled up with settle rows and no done in its
+  diary, given one reopen row so it can be judged again, gate-stamp: the
+  evidence gate could not write a tier's stamp after a complete run, so the
+  session stays due, and the seven `*-unreadable` kinds of the gate's side
+  files, states-unreadable, cleared-unreadable, stall-unreadable,
+  captions-unreadable, episodes-unreadable, marker-unreadable and
+  archive-unreadable: a file the gate stat'd into a tier's signature exists
+  and could not be read or parsed by the stage, so the run is marked
+  incomplete and stamps nothing, one row per failure episode).
   A file that does not parse is never deleted: it is moved beside its path as
   `<file>.corrupt-<utc stamp>` (a `-n` suffix when two land in the same second)
   before a fresh one is written, so the bytes survive for inspection, and the
--- a/docs/judges.md
+++ b/docs/judges.md
@@ -463,6 +463,29 @@ permission/API-error floors: one interrupt at a time, the present event first.
   read fresh each pass; empty = `ROMP_JUDGE_CONCURRENCY` as read at load,
   else 6). Every pool reads it at call time (`_conc`, or `_judge_concurrency()`
   directly); `DEATH_DRAIN_PER_PASS` alone stays on the load-time value.
+- Skips: the six triage tiers (planner, closer, unblocker, grouper,
+  consolidator, distiller) run behind an evidence gate. Before a session is
+  submitted, the runner takes the tier's signature: the identity (inode,
+  mtime, size) of every file the tier's decision path reads, the store with
+  its journal and archive among them, and for the planner, closer and
+  unblocker the pass's pinned parse pair. A session whose signature equals
+  the one the tier stamped after its last complete run is skipped, and its
+  pass watermark is stamped as for a pass that found nothing to do. A run
+  stamps only when it returned normally and set no completeness bit; a
+  deferral without a write, an empty reply, a failed call, a raise, or a side
+  file that exists and did not read leaves no stamp, and the session runs
+  again next pass. The stamps are process state, so the first pass after a
+  restart is a full walk. The planner has a second gate inside
+  `_plan_session`: a session whose inputs have not moved since a pass that
+  placed nothing, left the store's key where it was and ran to completion
+  returns before the store read. The evidence gate keys on the same inputs
+  (the reg by its `spawnedAt` and backend values rather than by identity) and
+  on `cleared.jsonl`, the death marker and the stall records besides, so an
+  idle session stops at the evidence gate; the inner gate's counters
+  (`memos.plannerSkip` on `GET /perf`, see `docs/reference.md`) count only
+  the sessions the evidence gate ran. Outside
+  a pass frame (`romp-judge --plan`) the evidence gate stamps nothing, and
+  the inner gate does the skipping.
 - Logs: `STATE/judge-usage.jsonl` (per-call cost, one name per prompt),
   `STATE/judge-errors.jsonl` (the row contract above; kinds are parse,
   call, give-up, sweep-cut, cite-miss, rate-limited, task-store, history-unreadable,
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -974,14 +974,34 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   while they stand (`served`, `built`); `captions` and `goalArchive` are the
   per-file read memos behind the index tier's caption readers and the re-plan's
   cleared context, each parsed once per file state (`served`, `parsed` or
-  `loaded`); `plannerSkip` is the planner's change gate (`skipped`, `planned`,
-  `recorded`: a session whose parse, store, journal, archive, episode log,
-  its leaf's task store, captions file and reg have not moved since a pass
-  that had nothing to do, and none of whose running background launches has
-  crossed its deadline, is not planned again). The compaction sweep after
-  each judge pass evicts from `pass` and `shared` the entries of stores no
-  session in the discover window owns, so both stay bounded by the live
-  board; the gate memo is bounded by the session count.
+  `loaded`). `goalArchive` memoizes a readable archive only: an archive that
+  exists and cannot be read or parsed is answered empty, marks the running
+  judge stage incomplete, and is not memoized, so the next call reads the file
+  again. `plannerSkip` is the planner's inner change gate (`skipped`,
+  `planned`, `recorded`). The planner runs behind two gates. The outer gate is
+  the judge's evidence gate around `_plan_session` (`docs/judges.md`, "Ops and
+  knobs"): a session whose signature equals the one the planner stamped after
+  its last complete run is skipped before it is submitted. It keys on the
+  inner gate's inputs, the reg by its `spawnedAt` and backend values rather
+  than by identity, plus `cleared.jsonl`, the death marker and the session's
+  stall records. The inner gate
+  sits inside `_plan_session` and sees only the sessions the outer gate ran: a
+  session whose parse, store, journal, archive, episode log, its leaf's task
+  store, captions file and reg have not moved since a pass that had nothing to
+  do, and none of whose running background launches has crossed its deadline,
+  is not planned again. The inner gate records a pass only when it placed
+  nothing, left the store's key where it was, and ran to completion; a
+  deferral without a write, or a side file that exists and did not read,
+  marks the run incomplete, and that session is planned again next pass. So
+  `plannerSkip` counts the sessions the outer gate let through, not every
+  planner skip: an idle session stops at the outer gate and appears in neither
+  `skipped` nor `planned`. Outside a pass frame (`romp-judge --plan`) the
+  outer gate stamps nothing, and the inner gate does the skipping. The
+  compaction sweep after each judge pass evicts from `pass` and `shared` the
+  entries of stores no session in the discover window owns, so both stay
+  bounded by the live board; the courier's and the planner's change-gate
+  tables are pruned to the sessions each pass discovers, and the evidence
+  gate's stamps are cleared at a fixed cap.
 - `judge`: `passes`, `ms_sum`, `ms_last`, `ms_mean` (wall time; a pass waits
   on model calls), `cpu_ms_sum` (CPU time of the judge tier threads and every
   per-session worker they run; the workers' share is `cpu_ms_workers`).
--- a/kernel/event_model.py
+++ b/kernel/event_model.py
@@ -254,12 +254,22 @@ _RESULT_CAP = 16000
 # notification can never arrive (the CLI died with the monitor out, and the transcript never learns).
 # Consumers apply this with THEIR now — baking it into the scan would freeze inside the mtime caches,
 # which an idle transcript never busts. The grace absorbs kill/notify latency.
-def _bg_expired(task, now, grace=120.0):
+def _bg_expiry_t(task, grace=120.0):
+    """The instant a scan row starts reading as expired: deadline + grace, or None for a row with no
+    deadline (it never expires by the clock). One definition for the predicate below and for the judge
+    gate's clock input (judge._settle_not_before), so the two cannot drift."""
     dl = (task or {}).get("deadline")
+    if not dl:
+        return None
     if (task or {}).get("deadlineSrc") == "hook":
         grace = 5.0   # a hook-recorded deadline is the harness's own kill moment (Monitor's
         #               required timeout_ms, journaled at launch) — exact, so only clock skew
-    return bool(dl) and now > dl + grace
+    return dl + grace
+
+
+def _bg_expired(task, now, grace=120.0):
+    x = _bg_expiry_t(task, grace)
+    return x is not None and now > x
 
 
 # The detail an agent/workflow row expands to (the Agent prompt / the Workflow script) is clipped:
@@ -3003,6 +3013,37 @@ def parse_session(leaf_path, rompuuid=None, name=None, color="#888888", dir=None
             "landedTextUuids": sorted(landed)}
 
 
+def task_store_dir(fsid):
+    """Claude Code's task store for one transcript stem: <CLAUDE_CONFIG_DIR or ~/.claude>/tasks/<fsid>,
+    resolved at call time (the env var, as task_store_plan reads it)."""
+    return Path(os.environ.get("CLAUDE_CONFIG_DIR") or str(HOME / ".claude")) / "tasks" / str(fsid)
+
+
+def task_store_fp(fsid):
+    """The task store's identity for the judge gate: None when the stem has no store dir (task_store_plan
+    returns None too, and the caller folds the transcript instead), else the sorted (name, mtime_ns, size)
+    of its item files. An item created, deleted or flipped in place moves it. A dir that exists but cannot
+    be listed raises OSError, as task_store_plan does: the gate then runs the stage and stamps nothing,
+    and the stage surfaces the failure loudly. Item files are stat'd by name, never read: a stat is all
+    identity needs."""
+    if not fsid:
+        return None
+    d = task_store_dir(fsid)
+    if not d.is_dir():
+        return None
+    out = []
+    with os.scandir(d) as it:                              # raises OSError → the caller's bypass
+        for e in it:
+            if not e.name.endswith(".json"):
+                continue
+            try:
+                st = e.stat()
+            except FileNotFoundError:
+                continue                                   # deleted between the listing and the stat
+            out.append((e.name, st.st_mtime_ns, st.st_size))
+    return tuple(sorted(out))
+
+
 def task_store_plan(fsid):
     """The agent's to-do list read from Claude Code's LIVE task store (<config>/tasks/<fsid>/<N>.json,
     honoring $CLAUDE_CONFIG_DIR) — the AUTHORITATIVE state TaskList/TaskGet read, updated by EVERY
@@ -3017,7 +3058,7 @@ def task_store_plan(fsid):
     it loudly, never silently fold (repo policy). A single corrupt item file is skipped."""
     if not fsid:
         return None
-    d = Path(os.environ.get("CLAUDE_CONFIG_DIR") or str(HOME / ".claude")) / "tasks" / fsid
+    d = task_store_dir(fsid)
     if not d.is_dir():
         return None
     items = []
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -2479,8 +2479,11 @@ def _sdk_owned(fsid):
 # goal/caption stores keep flowing through the pass — they are the pipeline's own dataflow (the closer
 # must see this pass's planner verdicts). Shared across tier threads AND their worker pools on
 # purpose: one frame, one world, first touch wins under the lock.
-_frame = None                    # {"parses": {fsid: session}, "keys": {tag: fileset-key}} while a pass runs
+_frame = None                    # {"parses": {fsid: session}, "keys": {tag: pinned value}, "served": {fsid: pair}}
+#                                  while a pass runs: "keys" holds the pass's ("parse", fsid) pairs
+#                                  (_frame_parse_key), "served" the pair each pinned parse was made under
 _frame_lock = threading.Lock()
+_NO_PIN = object()               # "no entry" in a frame's keys: a pinned None (a failed stat) IS an entry
 
 
 def begin_pass_frame():
@@ -2490,7 +2493,7 @@ def begin_pass_frame():
     with _frame_lock:
         if _frame is not None:
             return False                             # a joiner shares the creator's pass — and its key gate
-        _frame = {"parses": {}, "keys": {}}
+        _frame = {"parses": {}, "keys": {}, "served": {}}
         return True
 
 
@@ -2503,22 +2506,69 @@ def end_pass_frame(owned=True):
         _frame = None
 
 
-def _pinned_fileset_key(tag, files):
-    """The fileset key as of this pass's FIRST look (frame-pinned), else the live key. Keeps a
-    memo written mid-pass consistent with the frame's content: a live key over a file that grew
-    mid-pass would stamp stale derived data under the NEW key, and the next pass would trust it."""
+def _parse_key_files(fsid, files):
+    """The files the judge parse of `fsid` reads, as the filesystem shows them now: the candidate
+    transcripts (_judge_candidates over the RAW leaf list every caller hands in) plus states/<fsid>.jsonl
+    when it exists. Takes the raw list on purpose: _judge_candidates over an already-expanded list would
+    append a fork lane's anchor a second time, and a key computed that way would never equal the one the
+    parse cache holds (one spurious parse per fork lane per pass)."""
+    cands = _judge_candidates(fsid, files)
+    states = STATESDIR / (fsid + ".jsonl")
+    return cands, states, list(cands) + ([str(states)] if states.exists() else [])
+
+
+def _frame_parse_key(fsid, files):
+    """The (fileset key, pending cut) pair this pass judges `fsid` under, pinned in the pass frame
+    (2026-09-07).
+
+    Under a frame the FIRST caller pins it (tag ("parse", fsid)) and every later caller in the pass gets
+    the pinned pair back, whatever the files look like now. The pin is taken BEFORE anything is read, so a
+    reader holding the pair sees content at least as new as the pair describes, never older; that
+    ordering is what lets a stamp keyed on the pair stand for the content a stage judged (the evidence
+    gate, _stage_sig): the judged content can be newer than the stamped key, which costs one redundant
+    run next pass, never older, which would be a missed run. The fileset component is what the frame
+    fixes; the CUT the parse runs under stays LIVE (parsed_session reads it at parse time, as before the
+    frame existed), and the pair the parse was actually served under is recorded beside it
+    (fr["served"]) so the gate can withhold its stamp when the two disagree.
+
+    A stat that fails pins None: the tag is PRESENT with value None, the pass parses uncached, and the
+    gate reads a pinned None as run-and-do-not-stamp. Pinning None rather than leaving the tag absent
+    closes the hole where a later caller in the same pass would pin a fresh key over a parse read
+    earlier (a candidate _judge_candidates just saw can be renamed or removed before its stat, so the
+    exists() and the stat can disagree). No frame: the live pair, no pin.
+
+    Returns (pair, cut, fr): the pinned (or live) pair, or None; the LIVE cut this call read, or None when
+    an existing pin answered and nothing was read (the caller reads its own); and the frame dict the pin
+    lives in (None without a frame), so a caller pinning its parse pins into the SAME frame the key went
+    into rather than into whatever frame stands when its parse returns (a parse can span a pass
+    boundary). An existing pin costs one dict read: no stat, no cut read."""
     fr = _frame
     if fr is not None:
         with _frame_lock:
-            hit = fr["keys"].get(tag)
-        if hit is not None:
-            return hit
-    k = _fileset_key(files)
+            hit = fr["keys"].get(("parse", fsid), _NO_PIN)
+        if hit is not _NO_PIN:
+            return hit, None, fr
+    _cands, _states, key_files = _parse_key_files(fsid, files)
+    cut = _pending_cut(fsid)
+    try:
+        key = (_fileset_key(key_files), cut)
+    except OSError:
+        key = None
     fr = _frame
     if fr is not None:
         with _frame_lock:
-            return fr["keys"].setdefault(tag, k)
-    return k
+            return fr["keys"].setdefault(("parse", fsid), key), cut, fr
+    return key, cut, None
+
+
+def _frame_pin_parse(fr, fsid, session, key):
+    """Under _frame_lock: pin `session` as the frame's parse of fsid unless a concurrent first toucher
+    already did (first touch wins), and record the pair the WINNING parse was made under for the gate's
+    served-pair check. Returns the pinned parse."""
+    won = fr["parses"].setdefault(fsid, session)
+    if won is session:
+        fr["served"][fsid] = key
+    return won
 
 
 def parsed_session(fsid, files, now):
@@ -2530,7 +2580,14 @@ def parsed_session(fsid, files, now):
 
     Under an open PASS FRAME (begin_pass_frame) the FIRST parse of a session is pinned and every later
     call in the pass returns it — even after the file grew — so all stages judge one frozen world
-    (the user 2026-07-21); first touch wins across the tier's worker threads.
+    (the user 2026-07-21); first touch wins across the tier's worker threads. The parse's KEY is pinned
+    with it (_frame_parse_key): the fileset component is read before this function reads anything, the
+    cache slot is (pinned fileset key, live cut), and the pair the pinned parse was served under is
+    recorded in the frame. A gate that pinned the key earlier in the pass fixes the fileset component for
+    every parse of the pass; the content read here can then be newer than the key, never older, which is
+    the property the gate's stamp needs. The cache slot stays exact for the cache's own contract: a later
+    lookup hits it only when the live key equals the pinned one, which means the file has not moved since
+    the pin, so the cached content is that key's content.
 
     Passes states/<fsid>.jsonl so REAL idle transitions become idle atoms — without them _session_closed()
     is permanently False and a discharged focus goal never settles to completed (the settled gate, the
@@ -2542,6 +2599,10 @@ def parsed_session(fsid, files, now):
             hit = fr["parses"].get(fsid)
         if hit is not None:
             return hit
+    # The pass's (fileset key, cut) pair, pinned BEFORE this read: a gate that pinned first fixes the
+    # fileset component for the pass; a first toucher pins the live one here. `fr` is the frame the pin
+    # went into, and the parse below is pinned into that same frame.
+    pair, cut, fr = _frame_parse_key(fsid, files)
     states = STATESDIR / (fsid + ".jsonl")
     # A FORKED leaf (SDK /clear: discover hands the lastSid file under the stable romp sid) parses with
     # the session's anchor transcript among the candidates, so a fork whose chain back-links across files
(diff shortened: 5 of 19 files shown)


ksarma:
Heads-up on ordering with #1128: the two conflict in kernel/judge.py (this PR's hunks sit near the judge child's key path, which #1128 removes; docs/judges.md auto-merges). Either order works for me: if #1128 goes first, I rebase this the same day and re-verify the six-tier gate against the new judge child launch; if this goes first, #1128's judge.py hunk rebases over ten commits that leave _judge_env and the key gate untouched.

henrypinkard (approved):
Approved on head 4712c526 on the maintainer's word (feature-review session).

PR #1067, opened 2026-09-08
T252b: an earlier send's landing is a floor for a later one, a send queues below the kernel's foreign texts, and the docs match

henrypinkard:
Follow-up to the merged in-place pending bubble (#1044), from the manager's adversarial review of it.

- **Ordering between two pending sends.** An earlier-registered entry's covering echo and landed atom are now recorded as `floors` on every later entry, whatever their text, and placement sits below them. Before, X's landing was recorded only for same-text entries and Y's bubble was spliced above X's atom until Y landed.
- **A send above a kernel queue holding other texts.** The press records the tail queue's foreign texts (`queuedForeign`); the bubble sits below that group while it holds them and below their atoms once they land. A queue holding only our text keeps the in-place rule. The stale merge-into-the-group comment is rewritten.
- **docs/read-side.md** describes the in-place bubble instead of the removed header and cue.

**Tests (red before)**: `ui/webview/send-pending.test.ts` — the two-send ordering (landing, echo, echo→landed handover, a later send never a floor) and the foreign-queue placement (queued, own copy hidden, landed, own-text-only), plus the comment pin.

### DIFF
--- a/docs/read-side.md
+++ b/docs/read-side.md
@@ -236,8 +236,11 @@ never-delivered verdict, or the user's ✕), and a record the CLI wrote from sev
 back-to-back sends retires one bubble per text block (`blocks` on the user event).
 While the socket is down the bubble is labelled "not confirmed", until a kernel
 copy of the send clears the label. A send that landed mid-turn (`absorbed` and
-`landedAt` on the user chat event) wears a "joined mid-turn" header and, when its
-bubble sat at the tail, leaves a cue where the bubble was. The CLI extracts
+`landedAt` on the user chat event) is placed at its send time, and the chat's
+pending bubble is drawn at that same slot from the press — right after the last
+kernel event at the press, below an earlier send's echo or landing and below any
+texts the kernel already held queued — so the landing replaces it in place; there
+is no header and no cue (T252). The CLI extracts
 no image paths on the stream-json route (its only image-path test belongs to the
 interactive composer's paste handler), so an image path in an SDK send lands as
 typed and the echo's text matches. `_path_bearing` and the extension set it tests
--- a/ui/webview/render.ts
+++ b/ui/webview/render.ts
@@ -319,9 +319,12 @@ const order: string[] = [];           // positional tab order (for cycling)
 const pendingSent = new Map<string, PendingSend[]>();
 const isOptimistic = (e: ChatEvent): boolean => isOptimisticUuid(e.uuid);
 
-// The kernel's own queued group, if one is at the tail. Ours merges INTO it when present: the session is
-// provably holding messages, so this send will queue behind them — no reason to show it as a separate lone
-// bubble (the user 2026-07-16). Tail-scanned; a queued group only ever sits at the bottom.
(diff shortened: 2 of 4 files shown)


ksarma:
Recommend approval as shipped. Two follow-ups; neither blocks.

Reviewed inline: `reconcilePending` records an earlier entry's covering echo and landed atom as `floors`, and only on later-registered entries. Each floor is stored as uuid plus text and ordinal and re-derived from the kernel events each pass. So by the landing route a later send is never an earlier one's floor, and no stale index survives a resync. `placementIndex` follows an echo floor to the landed uuid by text and ordinal after the earlier send's ✕. `stampBase` records the tail queue's texts that match no pending send of this client as `queuedForeign`. The comparison key strips marker comments and leading quote lines, so a nudge's queued body matches its landed text. Duplicates count per text, and the floor is the k-th carrier. Placement is a pure function of the kernel events plus the press-time stamp. Floors are appended only when an echo or a landing is claimed, so a drawn bubble moves only on a kernel event. The render.ts change is comment-only, and the kernel is untouched. The docs paragraph names the three placement legs the code implements. Probes for three interleaved sends, a resync that drops the anchor, and an echo that blinks out all placed correctly.

Follow-ups:

- `ui/webview/send-pending.ts:384`: the echo block runs for an entry that landed in the same scan. When X's absorbed atom and a later same-text send Z's echo first appear in one push, X claims Z's echo. `floorFor` then records it on Y and Z, so Y is drawn below Z's echo and Z below its own. The kernel's echo hide set in `_merge_live_atoms` may keep that frame from the client. On the client the invariant is unguarded, and the test pins only the landing route. Gate on `landedIdx < 0` and add the echo-route case.
- `ui/webview/send-pending.ts:283`: a queued text equal to the send's own text is excluded from `queuedForeign`. No earlier pending entry's floors carry onto a later one. So with another client's "go" queued, A "alpha" then B "go" draws B at the anchor and A below the queue. The two sends stay inverted until A echoes or lands. Give a later entry the placement floors of every earlier entry still pending.

Two nits. `docs/read-side.md:241` says any texts the kernel already held queued; the code floors only texts that are not this client's pending sends. `kernel/kernel.py:26569` still carries the joined-mid-turn header and cue comment named beside the docs in the #1044 follow-up.

Still open from #1044 and out of scope here: `mk` at `render.ts:383` does not pass `received`, so a held send keeps the sending label and tooltip; and the #1044 body's served-test claim. This PR's body was not refreshed after the second commit (foreignKey matching, block matching, per-text ordinals, the third test).

Test evidence: the PR's `send-pending.test.ts` against origin/main fails tests 14, 15 and 16 (`floors` and `queuedForeign` undefined). Tests 31 and 32 fail on the added `queuedForeign: []` pin. A probe of main's module places every fixed scenario at index 1 where head gives 2. At head: 34/34 on the file, typecheck clean, node suite 3287/3287. CI seven of seven green at ff785657, CLEAN.

Review run by the maintainer's automated review session; verified against head ff785657.

PR #1073, opened 2026-09-08
The summarize hook parses on bash 3.2: heredocs never open inside $( )

lczh:
Verified on `main`: `hooks/romp-summarize.sh` opened a heredoc inside a command substitution, and the heredoc's body carried an odd number of apostrophes. macOS system bash 3.2 finds the end of `$( … )` by pairing quotes and parentheses in the raw text, knowing neither heredocs nor comments, so it never found the closing parenthesis and the whole enclosing compound failed to parse. That compound is the detached body of the hook, parsed only after the synchronous prelude had already set the session's tmux spinner to "pending", so on a Mac with system bash the spinner stuck and no summary came. The hook is registered asynchronously, so it did not block prompts; an earlier triage overstated that. The same shape sat in the lab tool and in the bats free-port helper, balanced by luck.

## What changes

Each heredoc moves into a function defined outside any substitution, and only the call is substituted. The Python bodies are byte-identical.

## Tests

`tests/test_shell_bash32_portability.py`: a scanner over every tracked shell file that flags a heredoc opened while a `$(` is unclosed, and a comment inside a substitution that moves the bash 3.2 quote scan, with a bash-3.2 walk of the body as the diagnosis. It flags the three sites on `main` (the summarize hook as the break, the two balanced ones as parsing by accident) and is clean on this branch; `bash -n` runs on every in-scope file; a real-bash-3.2 class runs only when such a binary is present and skips otherwise, never imitated; `ROMP_BASH32` names one explicitly, and a named path that is missing, will not run, or is another version fails the class loudly rather than skipping (a review nit against the first cut, which skipped on a mistyped path). The rule is the shape rather than today's quote count, because a balanced body parses by accident until the next comment apostrophe breaks it silently.

`tests/test_shell_bash32_portability.py`: 14 passed, 2 skipped (the real-bash-3.2 class, no such binary here); `bats tests/free-port.bats` 5/5; `bats tests/romp-summarize-gate.bats` 5/5; on `main` the tree test flags exactly the three sites. Reviewed by seven read-only agents: the heredoc bodies are byte-identical to `main`, the scanner probes behave as designed, and two nits (a silent skip on a mistyped `ROMP_BASH32`; a stale sha in the message) are folded.

### DIFF
--- a/hooks/romp-summarize.sh
+++ b/hooks/romp-summarize.sh
@@ -167,7 +167,15 @@ Reply with ONLY the <=8-word phrase capturing what they asked for. Nothing else.
       #   (nothing)        → the turn produced no text and used no tools; skip.
       #   VERBATIM\n<text> → short, tool-less reply (a direct answer): show as-is.
       #   MODEL\n<turn>    → real work: summarize with the model.
-      excerpt=$(python3 - "$transcript" <<'PY' 2>/dev/null || true
+      # The Python sits in a function and only the CALL is substituted: macOS's system
+      # bash (3.2) finds the end of a $( ... ) by pairing quotes and parens in the raw
+      # text, with no notion of a heredoc, so a heredoc opened inside the substitution
+      # is read as shell. Three apostrophes in the comments below left that scan inside
+      # an unclosed quote, and the whole detached compound failed to parse after the
+      # prelude above had already set the kind to "pending", a spinner that never ended.
+      # tests/test_shell_bash32_portability.py keeps every shell file clear of the shape.
+      _turn_excerpt() {
+        python3 - "$transcript" <<'PY'
 import sys, json
 lu = ""
 texts = []   # assistant text blocks in the latest turn (reset on each human prompt)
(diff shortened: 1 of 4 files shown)


henrypinkard:
Closing this one after review, on the owner's call.

We could not reproduce the parse failure. On a genuine GNU bash 3.2.57 (built from the 3.2.57 tarball with its parser regenerated from its own parse.y), main's `hooks/romp-summarize.sh` passes `bash -n` and executes its Stop arm end to end with output identical to this branch's. The reason is that bash 3.2's `parse_matched_pair` skips `#` comment lines inside a bare `$( ... )`, and all three apostrophes this change blames sit on comment lines. The branch's own `RealBash32` test asserts that the DEFECT snippet fails to parse, and on a real 3.2.57 it parses, so that test goes red exactly on the platform it targets (any Mac with system bash, and the manual macOS matrix). The heuristic scanner models bash 3.2 as counting quotes inside comments, which it does not.

The refactor itself is harmless and output-identical on bash 5 and 3.2, so if a portability lint for heredocs inside command substitutions is wanted, a new PR that presents it as a lint (with the scanner's comment handling corrected and no bug claim) would be welcome.

PR #1076, opened 2026-09-08
T252c: the kernel's queued copies carry an identity, and the chat places by it

henrypinkard:
Follow-up to #1067: the kernel's queued copies now carry an **identity**, and the chat places its pending bubble by it instead of by text and ordinal.

**Kernel**: the SDK backend's `send()` mints the copy's id (its existing echo key) before the enqueue, so the queued copy, the echo atom and the landed atom share one id; each queued text carries `{qid, qts}`; the feed moves the id to a fed ledger and `qid_for_landing` pairs the landed record with it (FIFO per text, at or after the feed); `build_session` ships `qid`/`qts` on queued copies and `qid` on the landed atom. A copy parked in the kernel's own FIFO carries no id until it reaches the backend (the park's op is the three-field record the on-disk mirror and a dozen readers pin; the id is minted where the copy enters the backend's queue) — stated as a gap. tmux copies carry their enqueue stamp and a digest id; nothing pairs a tmux landing (the kernel never sees the CLI take a copy), and a queue restored after a kernel death carries no ids — lossless legacy.

**Client**: a pending send latches its id from the first kernel copy attributed to it; landing, cover and hiding are then decided by id; floors and identified queued copies place exactly. Id-less copies keep today's text path unchanged (an older kernel, the tmux route, a restored queue).

**What came out of `send-pending.ts`**: nothing — the text path is the legacy fallback the lossless-legacy discipline requires, so it stays whole; the id path is added beside it and takes precedence when an identity is present.

**Tests (red before)**: `tests/test_queued_copy_identity.py` (kernel: minting, alignment, legacy copies, the fed ledger and pairing, the chat build, parks, tmux) and three new cases in `ui/webview/send-pending.test.ts` (the two stated limits with ids and their legacy degradation; identity latching; id-following floors).

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -21157,9 +21157,11 @@ def _pending_queued(path):
 
 def _pending_queued_meta(path):
     """_pending_queued's copies with what the CLI's ledger knows about each (T252c): the enqueue record's
-    stamp (`qts`, epoch ms) and an id digested from (stamp, text) (`qid`, stable across folds). The ledger
-    carries no ids and the kernel never sees the CLI take a copy, so nothing pairs a tmux landing with its
-    copy — the landed atom carries no qid on this route; the chat falls back to text there."""
+    stamp (`qts`, epoch ms) and NO id. The ledger carries none, the kernel's tmux echo is minted before the
+    CLI writes the enqueue record, and the kernel never sees the CLI take a copy — so nothing else in the
+    chain could share an id the ledger copy wore, and an id the chat latched from it would make it reject
+    the echo and the landing as another send's (the review of the first cut). The chat reads this route
+    by text; `qid` is None on every copy."""
     def step(pending, o):
         if o.get("type") != "queue-operation":
             return pending
@@ -21181,8 +21183,7 @@ def _pending_queued_meta(path):
             continue
         t = em.parse_z(ts) if isinstance(ts, str) else None
         qts = int(t * 1000) if isinstance(t, (int, float)) and t else None
-        qid = "tq:" + hashlib.sha1(("%s\n%s" % (qts, text.strip())).encode()).hexdigest()[:16]
-        out.append({"md": text.strip(), "qid": qid, "qts": qts})
+        out.append({"md": text.strip(), "qid": None, "qts": qts})
     return out
 
 
@@ -27744,13 +27745,18 @@ def build_session(sid, now, tmux=None, path_override=None, tail_cap_t=None, side
                             # it FIFO per text, at or after the feed — so the chat retires and places its
                             # pending bubble by id, never by text. None on the tmux route or for a record
                             # from before this kernel's life; the chat falls back to text then.
-                            if hasattr(be, "qid_for_landing"):
+                            if hasattr(be, "qids_for_landing") and not a.get("_echo_text") and not str(a.get("uuid") or "").startswith("echo:"):
+                                # per text BLOCK: a record the CLI wrote from several queued sends is one landing per
+                                # block; the kernel's own echo atom is skipped — its uuid IS the copy's id
+                                _tblocks = [b.get("text", "") for b in (blocks or []) if b.get("type") == "text" and (b.get("text") or "").strip()] or [text]
                                 try:
-                                    _q = be.qid_for_landing(sid, a.get("uuid"), text, a.get("t"))
+                                    _qs = be.qids_for_landing(sid, a.get("uuid"), _tblocks, a.get("t"))
                                 except Exception:
-                                    _q = None
-                                if _q:
-                                    ev["qid"] = _q
+                                    _qs = []
+                                if len(_tblocks) > 1 and any(_qs):
+                                    ev["qids"] = list(_qs)
+                                elif len(_tblocks) == 1 and _qs and _qs[0]:
+                                    ev["qid"] = _qs[0]
                             if a.get("absorbed"):
                                 # A mid-turn splice (event_model._absorbed): the CLI queued this send
                                 # behind the running turn and took it at a later tool boundary, so the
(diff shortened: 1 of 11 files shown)


ksarma:
Recommend one fix before approval: a send that latched an id loses its cover after a kernel restart and is drawn twice, a regression against #1067. The rest can follow. The review ran at 73cf43f5; 8de3e561 and d13b0479 landed mid-review and are read below. CI is green at d13b0479.

Reviewed inline: `send()` mints the id before the enqueue, and every `_pending` mutation goes through the `_q_*` helpers. Ids therefore stay aligned with texts, and identical texts get distinct ids. At 73cf43f5 the pairing had two holes. First, `build_session` ran the kernel's own echo atom through `qid_for_landing`. With a build between the feed and the CLI's record, the echo spent the fed entry and the landed atom shipped no id. Second, an identified foreign copy had no text fallback. Its echo frame or an id-less landing placed our bubble above the message it followed. The follow-up 8de3e561 closes both: echo uuids are refused, pairing runs per text block, and the text floor is kept for every copy. A probe of the echo, landed-with-id, landed-without-id and restored-queue frames places the bubble below the foreign message in each. One case is still open. A send that has latched an id is not covered by an id-less same-text copy or a reseeded echo. A kernel restart while a send is queued therefore draws two bubbles for one message. The PR body's claim that a restored queue keeps the text path holds only for a send that never latched. The `landedIdx` gate from the #1067 review is not here. The payload change is additive, and no injected text changes.

- Closed at d13b0479: at 8de3e561 `_mark_dropped_echoes` called `self.forget_fed` and the `BE` double in `tests/test_restart_redelivery.py` did not bind it, so one test failed on all four Python jobs. The third commit guards the call with `hasattr`; `tests/test_restart_redelivery.py` and `tests/test_queued_copy_identity.py` pass 19 at d13b0479.
- Fix before approval, `ui/webview/send-pending.ts:231` and `:467-469`: a latched entry has no text fallback for its cover. After a restart it is injected beside the kernel's restored copy (probe: inject 1, unqueue 0; the never-latched control gives unqueue 1). Fall back to text when the frame's copies carry no ids, or persist the id and echo key across the restart. Add a restart test.
- Follow-up, `ui/webview/send-pending.ts:461-465`: the echo block still runs for an entry that landed in the same scan. The new latch then writes the later send's id onto the retired entry. The `landedIdx < 0` gate and echo-route test from #1067 are still missing.
- Follow-up, `ui/webview/render.ts:3968`: kernel copies now ship `qts` under the field the ✕ handler reads as the client's press-time identity. A ✕ on a kernel-drawn SDK copy therefore drops no entry, where main dropped the first same-text one. Gate `data-qts` on `t.optimistic` or rename the kernel stamp. The comments at render.ts:171 and :15244 are now wrong.
- Follow-up, `ui/webview/send-pending.ts:480`: the id is latched from the copy at the position the text heuristic assigned. The kernel never returns the minted id to the sender, so two dashboards sending the same text both latch the first copy. Returning the id on the sendMessage ack would make attribution exact.
- Follow-up, `kernel/kernel.py:27755` and `:28165`, `kernel/sdk_backend.py:2651` and `:2644`: the new fallbacks (except Exception, the id-to-text mismatch, the 64-entry trim) degrade to the text path with no log line.
- The PR body was not refreshed after 8de3e561. tmux copies no longer carry a digest id, `qid_for_landing` is now `qids_for_landing`, and the two retargeted source pins are unmentioned.

Test evidence: at the PR base with the PR's test files, `tests/test_queued_copy_identity.py` fails 7/7 (no `pending_queued_meta`). The two retargeted pins fail on the moved source strings, and `send-pending.test.ts` fails 6. At 73cf43f5 all pass (96 pytest, 40/40, node 3299/3299, typecheck clean, CI 7/7). At 8de3e561 the PR's Python files pass 101, send-pending 41/41, node 3300/3300, typecheck clean. `tests/test_restart_redelivery.py` gives 1 failed, 6 passed. CI at d13b0479: all seven checks green.

Review run by the maintainer's automated review session; verified against head 73cf43f5.

ksarma:
Reviewed after the merge: the fix asked for in round one is in at ff09d6d8, and the remaining items are follow-ups for a later pass.

Reviewed inline: `_persist_queue` writes `reg['queueMeta']` for every queue position. `queue_meta_from_reg` restores the ids aligned as one block. When no block matches, it restores them FIFO by text. For an older mirror without the field, it restores all None. `_reseed_echoes` keeps the echo uuid. A send that latched an id before a kernel death therefore finds the same id on the restored copy. On the client, `locateId` reads a send by text for any push whose frame shows its id nowhere. As a result, an older mirror's id-less copies still cover a latched send instead of drawing it twice. On the ✕ path, `data-qts` is set only on the client's own bubble, and a kernel copy carries `data-qid`. `dropPending` drops the send that owns that id, else the first id-less same-text entry, else the first by text. Restored and tmux copies therefore cancel as before. The text-path scan now stops at the landing, which closes the echo-after-landing claim from round one. The retargeted pins in optimistic-send, queued-indicator, followup-render, test_sdk_rewind and test_sdk_stream_survives_handler_errors are string moves with no behavior change. The payload stays additive and no injected text changes.

- Follow-up, `ui/webview/send-pending.ts:510`: the echo block still runs for a send that landed in the same scan when the echo precedes the landing in the frame. The latch then writes the later send's id onto the retired entry and its landing floor. The later bubble is placed above the earlier message (probe: index 2 at head, 3 at main). A `landedIdx < 0` gate closes it. The `break` at :500 has no pinning test: reverting it to `continue` leaves send-pending 47/47 green.
- Follow-up, `ui/webview/send-pending.ts:533` and `kernel/kernel.py:13893`: the id is still latched from the copy at the position the text heuristic assigned. The sendMessage handler returns no id. Two dashboards sending the same text within one push therefore both latch the first copy. Unchanged from round one.
- Follow-up, `kernel/kernel.py:27754` and `:28164`, `kernel/sdk_backend.py:2736`: the fallbacks to the text path still log nothing. Unchanged from round one.
- Follow-up, PR body: it still says tmux copies carry a digest id, a restored queue carries no ids, `qid_for_landing` pairs the landing, and send-pending gains three cases. The ✕ change, the `queueMeta` mirror and the retargeted pins are unmentioned. Please refresh it before merging.
- Nit, `ui/webview/render.ts:15245`: the comment still says a kernel bubble's ✕ names no entry, two lines above the `data-qid` read.

Test evidence: at origin/main with the PR's test files, `tests/test_queued_copy_identity.py` fails 20/20 (no `pending_queued_meta`, no `queue_meta_from_reg`). The two retargeted pins fail on the moved strings (22 failed, 87 passed). `send-pending.test.ts` fails 14, all ten new tests among them. At ff09d6d8 the PR's Python files pass 116, send-pending 47/47, optimistic-send 14/14, queued-indicator 24/24, followup-render 5/5, the node suite 3306/3306, typecheck clean. CI: all seven checks green.

Review run by the maintainer's automated review session; verified against head ff09d6d8.

PR #1077, opened 2026-09-08
A send that came back no longer reads as waiting on that peer

lczh:
## What was wrong
Verified on `main`: three readers turn a session's mail log into "waiting on a peer", and none of them read the row the bus writes when a send comes back. The kernel's wait maps feed the awaiting chip, the auto-nudge gate and the clock that supersedes a durable "waiting to hear back" stamp. The judge's ask maps feed the closer's and planner's awaiting-peer gate. The courier plants a "delegated to that peer" tracker from every recent cross-host delegate. A sent row followed by a terminal bounced row for the same message, meaning the send is over and nothing will ever answer it, still read as an open ask in all three. The session was shown waiting on a peer that would never reply, and its card was parked in a state that was not true, hiding that the person might need to act. The review of #1071 filed this as a follow-up.

## What changes
One recognizer in the judge reads a bounced row by the original message id it carries, since the row's other fields differ per writer. A sent row whose id came back makes no open ask in the kernel's wait maps or the judge's ask maps, so the two agree, and the courier plants no tracker for a delegate that already came back. A shared clock ends a pair's wait at the peer's reply or at the bus's return of the newest reply-requiring send, whichever is newer, and it is per message: a live unanswered send keeps the pair open whatever came back for an older one. The moving event is the bounced row's arrival, which changes the log; the maps rebuild once, the edge is gone and the stamp is superseded at that row's time. Nothing is re-derived per build. The bus already tells the sender that a message came back, naming the recipient and the reason, so nothing is added there.

## Deliberately left out
Two pre-existing states stay as they are and are named here so nobody reads them as covered. A handoff tracker planted before its return arrived is not closed: the courier only skips planting one for a delegate that already came back, so a card can still say it is waiting on a peer's result after that peer refused the message, and the nudge that would surface it is suppressed for a handoff; closing that needs a verdict that says the handoff came back, with its writer and rendering. And the "any later word" map is unchanged, so a reply that itself bounced on the local oversize push still counts as the peer's word: the asker's chip clears and a tracker can be marked reported back on a report the asker never received. Recall rows, the other terminal kind, are not read either. Each is a one-recognizer or one-verdict follow-up, not a widening of this fix.

## Tests
Twenty-one new cases across the wait-for, await-release, awaiting-peer-matrix and debt-nudge modules, over a real synthetic mail log through the real readers: a returned send makes no edge, ends the pair at the bounce's time and admits no awaiting-peer stamp; a returned older ask with a live newer one stays open; a delivery acknowledgement is not a return; the kernel and judge twins agree; a delegate that came back plants no tracker; the sent-alone and sent-plus-reply guards hold. The guards that hold on `main` by design pass there; the counts per module are below.

The debt readers are the fourth reader: a reminder's outcome is judged per ask, keyed by the ask's own send time, so a returned ask retires its reminder record on both the turn-end path and the six-hour backstop instead of flipping the asker's card to needs-you over a message that never arrived. That join is per send time, since the record carries only the ask's second; a live question sent in the same second keeps the record open, so a twin that came back never retires the one still owed. The reminder itself is injected into the session, not sent through the bus, so only the ask can come back.

Module runs on this branch, one at a time: wait-for 17, await-release 21, awaiting-peer-matrix 34, debt-nudge 21, `test_judge` 515, the `test_kernel` peer-wait subset 45, and twelve neighbouring reader modules, plus the ten modules and the postal bats that #1058 changed, all green. On `main`: wait-for 4 failed / 13 passed, await-release 5 / 16, awaiting-peer-matrix 3 / 31, debt-nudge 2 / 19, every red on the defect except one that needs the new helper.

## Review
Reviewed by twenty-four read-only agents (two lenses, two skeptics per finding). Confirmed and folded: the debt outcome readers still escalated a reminder whose ask the bus had returned; the commit's claim that receipts closed a pair, when only a peer's own row ever did; a docs sentence saying every reader; a test class below its module's main guard; the bounce-then-late-reply interleaving without a test; and one main-side red that was an attribute error rather than the defect. Four refuted as pre-existing and already stated, two of which are named above.

## Reconciled with #1071
#1071 merged with the maintainer's rule that a refused or destroyed send is neither an ask nor an answer, so a bounced reply answers nothing; that closes the one case this PR had left as pre-existing. The rebase keeps his skip as the membership rule and layers the return clock, the debt retire and the courier plant on top, with both readers sharing one table of returned ids, and the docs now say the row is neither an ask nor an answer.

### DIFF
--- a/bin/romp-manager
+++ b/bin/romp-manager
@@ -845,7 +845,7 @@ function startManager() {
       // SIGTERM respawns blinked every dashboard for an hour, and this handler — the one door every
       // programmatic restart walks through — recorded nothing. The immediate path used to be silent.
       log(`restart-all requested (${url.searchParams.get('when') === 'quiet' ? 'deferred to quiet window' : 'IMMEDIATE'}) from ${req.socket.remoteAddress || '?'}`);
-      if (url.searchParams.get('when') === 'quiet') {            // deploy path: land at the next quiet window
+      if (url.searchParams.get('when') === 'quiet') {            // `romp refresh --quiet`, the one quiet door (T269): land at the next quiet window
         return json(202, { ok: true, deferred: true, coalesced: queueQuietRestart('all') });
       }
       clearPendingQuiet();                                       // this bounce delivers any pending deploy too
--- a/docs/event-model.md
+++ b/docs/event-model.md
@@ -286,7 +286,16 @@ not a session); the recipient's stable id rides the additive `to_sid` key (rows 
 isn't in the log (legacy/rare).
 Two rules the wait readers apply to these rows: only a `question` or a `delegate` opens
 a wait, so the answered clock walks reply-requiring sends and a coordinate-only exchange
-neither opens nor reopens one (a reply of any kind still answers). A row without
+neither opens nor reopens one (a reply of any kind still answers). A `bounced` row
+naming a sent row's id is terminal (the send came back: the peer refused it, the
+recipient exited, it never left) and closes that ask for every reader that opens a
+wait from a sent row — the chip, the stamp clock, the closer's admit gate, the debt
+reminder's outcome, the courier's plant: the row is neither an ask nor an answer, so
+the sender is not waiting on a peer that never got it, a bounced reply answers
+nothing, and a stamp filed before the return is superseded by it as by a reply.
+A handoff tracker planted before the return arrived, and a `recall` row, are not read
+this way. The row is the event, keyed to the message it names, so a newer live ask
+keeps waiting whatever came back for an older one. A row without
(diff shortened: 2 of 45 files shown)


lczh:
Merged main into this branch to clear the conflict that #1095 created when it landed first: main now carries this branch's own commit through #1095, so every conflicting hunk resolved to main's side (this branch's lines as #1095 extended them), and the review test added here for the bounced reply's watermark is kept verbatim in its class. The waitfor, await_release, awaiting_peer_matrix, debt_nudge and the judge's return/recall cases pass on the merge. No commit on the branch was rewritten.

PR #1079, opened 2026-09-08
The usage modal covers every attached kernel's sessions, merged kernel-side by host

henrypinkard:
**Feature (T247c).** The usage modal's per-session table and stacked histogram now include every attached kernel's sessions, not this machine's only. The user's ask (2026-09-08 morning PT): the API spend dashboard looks great but covers only the local sessions, and the federated kernels' sessions matter; the "do not share theirs yet" note was wrong in spirit.

**Design points**
- **Authoritative data, per host.** Each kernel owns its ledger and resolves its own names and colors, so a host's story comes from that host's kernel. `GET /spend/detail` aggregates kernel-side: for every attached host the hover's spend totals include (the same predicate as its rows), it calls that host's `/spend/detail` over the tunnel with the peer's serve token, the transport mirror_trust and the settings mirrors use, in parallel under one bounded timeout, and merges by host.
- **Fail loudly.** A host that is down, times out, refuses the token, or runs an older build without the route is reported in the payload by name and status and shown in the modal as a line ("HOST: not reachable", "HOST: older build, no per-session data"). Never silently omitted; a slow host holds nobody hostage past the timeout.
- **Alignment.** Hours align by absolute time: the local reader ships each key's epoch hour, and a peer's stacks map through its own epochs, or through the offset it reports when an older peer ships none. Never by the spelling of local-hour keys. Days align by each machine's own calendar date, and the footer states the rule when hosts sit in different zones.
- **Presentation.** Totals rows stay the summed windows the hover shows. The table lists sessions from all hosts sorted by dollars, with the host on each row in the tab strip's host-prefix voice when more than one machine contributes; identity colors stay per session. The histogram stacks sessions across hosts in their own colors; "other" counts every non-top contributor across hosts; per-host unattributed folds into one stack whose tooltip names each machine's share. The billing rule is named when every contributing machine shares one, else "each machine's own billing rule". The "this machine only" note is gone. The phone panel's door opens the same merged modal.
- **Top-N across hosts.** A session outside its own host's top-N has at least N host-mates above it, so it cannot enter the merged top-N; every merged stack therefore has its per-session series, and a peer's own "other" fold joins the merged one.

**Tests.** A hermetic two-kernel route test with a stand-in peer behind its serve token: merged rows tagged by host in dollar order, the peer's hour on the same absolute hour though its local key reads three hours later, per-host unattributed with the host named, daily alignment by date; an older peer without epochs aligning through its offset; a down, an older (404), a hanging and a token-refusing peer each reported by name inside one bounded wait; the asked set equals the hover's. The served test drives the two-host page: host prefixes on rows and chips, the older peer's status line, no "this machine only", the timezone footer. It doubles as the screenshot harness.

**Review folds (second commit).** Adversarial review, two lenses, ten findings, all confirmed hermetically and folded. Two were blocking: the route served the merged payload to a peer too, so two kernels attached to each other (the ordinary pairing) asked each other back until the socket timeouts unwound hundreds of nested requests, with the healthy peer reading as timed out; and a peer answering with a merged-shaped payload had every row re-tagged as the peer's, double counting a third host and our own sessions. The peer ask now carries `?local=1`, the route serves the local reader for it and never fans out, and only the rows a peer tags as its own count. Also folded: a tolerant peer call so a real older kernel's plain-text 404 reads as "older build" (the shared transport reported it as a parse error, so it read as "not reachable"); the axes extend, bounded, for a peer's day east of our date line or its hour after ours ticked over, which previously vanished without a trace; the merged "other" count names contributors in that range only; the header names the machine count when several contribute; a duplicate import removed. Each fold has a red-first test, including a stand-in peer that records the path it was asked and one that answers the real kernel's 404 shape.

**Left out.** Nothing from the dispatch. Remote hosts still answer with their own top-N and fold the rest into "other"; the merge is exact for the stacks it draws and for every table row.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -31771,14 +31771,42 @@ def _epoch_of_local_hour(key, off_min):
         return None
 
 
+def _peer_spend_call(row, timeout):
+    """GET /spend/detail?local=1 on an attached kernel — mirror_trust's transport (the tunnel's local
+    forward, THAT machine's serve token), asking for the peer's LOCAL half only: the route never fans out
+    when serving a peer, so a mutual attachment (the ordinary pairing: an ssh attach one way and the
+    check-in row the other) cannot recurse (review find: one modal open ping-ponged between two kernels
+    until the socket timeouts unwound it, hundreds of nested requests deep, and the healthy peer read as
+    timed out). Tolerant of a NON-JSON body: an older kernel's unknown-route answer is a plain-text 404,
+    which the shared transport reports as a parse error — here it is (404, None, None), so the caller can
+    say "older build" rather than "not reachable" (review find). Returns (status, json_or_None, err)."""
+    host = row.get("host") or "?"
+    port, tok = row.get("local_port") or 0, row.get("token") or ""
+    if not port or not tok:
+        return None, None, "no admin path to '%s' (missing forward or token)" % host
+    try:
+        conn = http.client.HTTPConnection("127.0.0.1", int(port), timeout=timeout)
+        conn.request("GET", "/spend/detail?local=1", None, {"X-Romp-Token": tok})
+        resp = conn.getresponse()
+        data = resp.read()
+        conn.close()
+        try:
+            j = json.loads(data.decode() or "null")
+        except Exception:
+            j = None
+        return resp.status, j, None
+    except Exception as e:
+        return None, None, "could not reach %s's kernel: %s" % (host, e)
+
+
 def _spend_detail(now=None):
     """GET /spend/detail — EVERY attached kernel's sessions, merged here (T247c, the user 2026-09-08:
     the per-session breakdown covered this machine only, and the federated kernels' sessions matter).
     Each kernel owns its spend.json and resolves its own names and colors, so a host's story comes
     from THAT host's kernel: for every attached host the hover's spend totals include (the same
     predicate as its rows — a remote reporting spend windows) this kernel calls its /spend/detail over
-    the tunnel with that machine's serve token (mirror_trust's transport), all in parallel under one
-    bounded timeout, and merges by host. A host that is down, that times out, that refuses the token,
+    the tunnel with that machine's serve token (mirror_trust's transport, asking for that kernel's LOCAL
+    half so nothing recurses), all in parallel under one bounded timeout, and merges by host. A host that is down, that times out, that refuses the token,
     or whose older build lacks the route is REPORTED in `hosts` by name and status — never omitted
     (fail loudly) — and the rest renders. Hours align by ABSOLUTE time (each host's epochs, or its
     offset for an older peer), never by the spelling of local-hour keys; days align by each machine's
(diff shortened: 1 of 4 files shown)


ksarma:
Reviewed after the merge at b6886b61: nothing blocked it; four follow-ups are listed below.

Reviewed inline: `_spend_detail` fans out to each attached host over the existing tunnel forward on 127.0.0.1, sending that row's own serve token and nothing else. The route, including the new `?local=1` branch, comes after the `_authorize` gate, so a token-less GET gets 403. The first commit served the merged payload to peers too. Two kernels attached to each other then recursed until the socket timeouts fired, and the healthy peer was reported as unreachable. This was reproduced with two real kernel modules at 6a3bc4e7. The second commit ends it: the peer is asked for its local half, receives exactly one request, and answers in milliseconds. A merged-shaped answer no longer re-tags a third host's rows. Every peer string reaching the modal passes through `esc()` or `textContent`, and the color is regex-checked, so a peer cannot inject markup. Hours align by epoch, with the offset fallback for a peer that sends none; days align by each machine's own date. A peer bucket beyond our newest slot now extends the axis (bounded) instead of vanishing. All peers are waited on together under one 6.5 s join deadline inside the modal's 20 s abort, with the loader up meanwhile. The tier fits: there is no new trust boundary, and the token transport is the same-user one the settings mirrors use. The hosts int-to-list payload change has no consumer outside the kernel's own page.

- `kernel/kernel.py:31849` (follow-up): a peer's HTTP 500, or a 200 without `sessions`, is filed as an older build. Only a 404 means that. A same-build peer whose reader raises is told to upgrade. Give this case its own status with the real reason, plus a stand-in test for a text 500 and a sessions-less 200.
- `kernel/kernel.py:31877` (follow-up): one malformed peer row (no sid, or a non-numeric usd) raises out of the merge. The route then answers 500, and the modal loses this machine's rows too. Fold each peer payload under a per-host guard that demotes that host to a status line.
- `kernel/kernel.py:39692` (follow-up): the shell page loads no stylesheet and defines no `.host-prefix` rule. The prefix therefore renders at the name's own weight and size, not the tab strip's treatment the comment describes. Mirror the rule into the modal's inline CSS and pin a computed style in the served test.
- `tests/test_spend_detail.py:403` (follow-up): no peer fixture has more sessions than top-N or includes its own other stack. The top-N-across-hosts argument and the peer fold branch are therefore unpinned. The local-half test asserts a kernel source string rather than driving the Handler.
- Nits: the range label reads "last 91 days" when a peer's day extends the axis (`kernel/kernel.py:39778`). The token-refusal stand-in sends a bodiless 401, while the kernel sends a 403 text body.

Locally at b6886b61: `tests/test_spend_detail.py` 20 passed; `tests/test_spend_modal_headless.py` 1 passed under Playwright chromium (skipped in CI). The second commit's five changed tests fail against the first commit's kernel and pass at head. CI: all seven build checks green at b6886b61. The Tier policy check is red only because no approval exists yet.

Review run by the maintainer's automated review session; verified against head b6886b61.

PR #1084, opened 2026-09-08
Both tier checks wait a minute for a just-opened PR's label instead of failing at once

lczh:
## What was wrong
Both required tier checks read a PR's labels live from the API, so the order of their runs stopped mattering after #1039. But the run that the `opened` event starts still failed whenever it read before the label landed: `gh pr create --label` fires `opened` first and adds the label a second later, an edit right after creation takes a few seconds, and a hand-labelled PR takes longer. Each such failure was a red required check on a PR that was sorted seconds later, and a failure notice to its author about a state that no longer existed. On 2026-09-08 there were eighteen of them across thirteen branches from the label check alone; since #991 the tier-policy check races the same way, so a PR now gets two. Neither workflow handed its step the event's action, so neither could tell an opened run from a labeled one.

## What changes
The opened and reopened runs, and only those, wait when they find no tier label: the label check's script re-reads the API every five seconds for up to a minute, then judges as before; the policy check's fetcher does the same before the pure policy judges, on the per-PR path only, so the hourly sweep and a manual dispatch never wait, and it refreshes the head after the wait so a push that arrived meanwhile is judged on the current head. A PR that stays unlabelled still fails, a minute later. Two labels never wait, since a doubly sorted PR is a real error rather than a race. Every other event's run judges at once, so removing the last label still fails both checks immediately. The event action reaches each step from the payload's fixed action field; labels never do. The policy itself is untouched. Two knobs let a test run the wait in no time.

## Tests
Twelve new cases across the two modules, in each module's own style: the shimmed-gh workflow script waits then passes after exactly three reads, fails after the bounded retries when no label ever comes, and judges at once on an unlabeled event or on two labels; the policy fetcher's wait counts reads with a fake fetcher and an injected sleep, ends on the alias too, judges a docs-tier label that lands on the second read as docs, and the per-PR entry point waits while the sweep does not; both workflows are pinned to hand their step the event action. On `main`: 3 failed / 11 passed and 9 failed / 82 passed. On the branch: 14 and 91 passed, with every existing pin intact.

Module runs on this branch: the workflow module 14 passed, the policy module 91 passed; on `main` 3 failed / 11 passed and 9 failed / 82 passed, every red one a new case. Reviewed by a read-only agent with no findings.


lczh:
Closing as moot: T273 (8124498c) heals this at the source. The PR body now declares the tier, the tier workflow applies the label itself when the PR carries none and re-runs the label counter on that head, so a just-opened PR no longer sits red between creation and labelling. The grace wait this PR added has nothing left to wait for.

PR #1086, opened 2026-09-08
Follow-the-tail pins the bottom only when there is something new to follow, never on a frame that changed no content

henrypinkard:
## What

A reader near the tail of a busy session was snapped back to the bottom within the first 80 px of wheeling up, frame after frame (the user 2026-09-08: "jumped up slightly on my scroll in many, many sessions").

**The writer, convicted on the two-kernel harness** (kernel B attached through A's relay, a 300-unit transcript, compact mode and normal, wheel-scrolled stop by stop): a stop 60 px above the bottom moved 60 px to the bottom during a quiet window with `scrollHeight` unchanged (compact), and both modes snapped on the next frame; stops at 300 px and at the very bottom never moved, and the continuous wheel scroll through the virtualization window (130 to 248 steps per pass) recorded no anchor-offset violations. `appendActive` treated "within the 80 px follow threshold" as follow mode and pinned the bottom on EVERY frame the pane received, a status-only `chatTail` with an empty suffix included, and a working or awaiting session sends those every pusher cycle. The snap is small (at most the threshold) and lands exactly while the reader scrolls away from the tail, which is what the report describes; the earlier scroll fixes (#1036, #1040, #1049, #1063) are about a different writer and stand.

**Fix.** Follow-mode pins the bottom only when there is something new to follow: the content's height changed across the rebuild, or the reader was already at the very bottom (a no-op pin that keeps the tail visible). A near-bottom reader on a frame that changed nothing stays exactly where they are. The scrolled-up path (anchor restore, raw fallback) and the threshold itself are untouched. Event-based: the content change is the event. Decision in `scroll-keep.ts` `followTail`, pure.

Composes with #1083: the remaining pin goes through the same `writeScroll` helper and stays on the client-diag record as `append-stick`.

## Tier

- [ ] tests-only
- [x] fix — a bug fix with a test that fails before it
- [ ] feature
- [ ] major-feature

## Tests

New `ui/webview/follow-tail.test.ts`: executes the harness case (60 px up, nothing new → no snap; 35 px of new content → follow; at the bottom → pin) and pins `appendActive` measuring before the rebuild and pinning only on `followTail`; the wiring pin is red on the previous render.ts. Suites: node typecheck clean, focused tests green; the full node and Python runs are in flight and CI gates the merge.

### DIFF
--- a/ui/webview/follow-tail.test.ts
+++ b/ui/webview/follow-tail.test.ts
@@ -1,0 +1,36 @@
+// T262 (the user 2026-09-08, "jumped up slightly on my scroll in many, many sessions"): appendActive pinned a
+// reader within the 80 px follow threshold to the bottom on EVERY frame — a status-only chatTail included — so
+// wheeling up from the tail of a working session snapped back within the first 80 px, over and over. The
+// two-kernel harness reproduced it: a stop 60 px above the bottom moved 60 px to the bottom in a quiet window
+// with scrollHeight unchanged (compact mode), and both modes snapped on the next frame. Rule: follow the tail
+// only when there is something new to follow — the content's height changed — or when already at the very
+// bottom (a no-op pin). Executed decision + render.ts wiring pins, red on the previous render.ts.
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import * as fs from "node:fs";
+import * as path from "node:path";
+import { followTail } from "./scroll-keep";
+
+const RENDER = fs.readFileSync(path.resolve(process.cwd(), "..", "ui", "webview", "render.ts"), "utf8");
+
+test("the harness case: 60 px above the bottom, a frame that adds nothing → no snap; new content → follow", () => {
+  assert.equal(followTail(60, 145219, 145219), false, "status-only tail: the reader stays 60 px up");
+  assert.equal(followTail(60, 145219, 145254), true, "35 px of new content: follow the tail (the designed threshold)");
+  assert.equal(followTail(60, 145219, 144945), true, "a height change either way is new information to follow");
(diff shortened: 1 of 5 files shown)


henrypinkard:
## Verified live on a frozen copy of this branch (two-kernel harness, wheel-scroll stops, compact and normal)

Same protocol as the reproduction: kernel B attached through A's relay, a 300-unit transcript, wheel-scrolled to a stop at a chosen distance from the bottom, 3 s quiet, then one transcript record pushed. "moved" is the scrollTop change; "Δheight" the scrollHeight change.

| stop | main (compact) | main (normal) | this branch (compact) | this branch (normal) |
|---|---|---|---|---|
| 300 px up, quiet / push | none / none | none / none | none / none | none / none |
| 60 px up, **quiet** | **moved +60, Δheight 0** (the phantom snap) | none | **none** | none |
| 60 px up, push | moved −274, Δheight −274 | moved +95, Δheight +35 | moved −214, Δheight −274 | moved +95, Δheight +35 |
| at the bottom, quiet / push | none / pinned | none / pinned | none / pinned | none / pinned |
| continuous wheel scroll through the window | 244-251 steps, no anchor violation | 248 steps, none | 251 steps, none | 248 steps, none |

The only thing that changed is the row this PR is about: a reader 60 px above the tail is no longer pulled to the bottom by a frame that added nothing. A push with new content still follows the tail (the designed 80 px threshold), and a reader at the very bottom stays pinned. The compact-mode rows where a push shrinks `scrollHeight` by ~180-270 px are the virtualization's spacer estimate settling on real heights while the reader stays pinned to the tail; no visible movement at the tail, noted here as an observation, not changed.

PR #1091, opened 2026-09-08
Tab strip: a session under several tags appears under every one of them (T264b)

henrypinkard:
There is no home tag (the user 2026-09-08): tags are equivalent, so a session under two tags appears under both. Follows the one-group-per-line strip (#1087).

## Design points

- **The rule.** `sectionTabs` lists, per union, every visible member it holds: a session under N tags has N copies, sections in tag order, the untagged trail holds only sessions in no tag. `homeTag` is retired. The provisional tab (a create in flight) appears under every requested tag from the first paint.
- **Every copy is the full tab of the one session.** Same per-item loop, so each copy wears the identity colour, the state class and dot, and the active highlight; a click on any copy selects the session (the `#tabs` delegate reads `data-id`); the ✕ stays on every copy and ends the one session, its tooltip saying so. `data-copy` names the copy's group for the by-copy readers.
- **Keyboard order.** `planStrip` drops an id from the folded set only when every copy is folded away; one copy on screen keeps the session in ←/→ cycling. Focus, the tab menu and the hover tip land on the first copy (the same session).
- **Drag-and-drop.** The drag tracks the dragged element beside the id (the id alone named every copy, so the first copy moved while the user dragged another); the drop's neighbour rule skips the dragged session's own copy in the group next door (reordering against itself moves nothing). A drop still changes no membership: a copy dragged into another group's row reorders within the global order and re-sections home on the next render. Adding a tag stays with the tab menu's Tags rows.
- **Flagged, not changed here.** The tab menu's "Move to <tag>" row still swaps the session's first holder tag for another; under the equivalence ruling that row's meaning ("move" between equals) is the user's call.
- **Folds, counts and pips** count each group's own copies; a pinned copy shows through its own group's fold only.

## Tests

- `ui/webview/tab-groups.test.ts`: the every-tag rule and the untagged trail executed; the provisional tab under both requested tags; the folded-set rule; render pins (data-copy, ✕ tip, flipTabs per-copy keys, the dragged element in dragover and drop, the own-copy neighbour skip). The pinned-member expectations that encoded the home-tag rule are recomputed under the every-tag rule.
- `tests/test_tab_groups_rows.py`: a two-tag session shown under both groups in the served page, same state dot on both copies, the ✕ tip, a click on the second copy highlighting both; both themes.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -71,8 +71,8 @@ several. Right-click a tab and open **Tags** to add or remove them. Tags filter 
 surface (the tag button in the strip narrows the tabs to the tags you pick), and they group
 the tabs: as soon as any session carries a tag, the strip shows one section per tag, in your
 tag order, each with a header in the tag's color, and the untagged sessions after a divider
-at the end. A session with several tags sits under the first of them in your tag order; its
-other tags still filter. Each header shows a chevron, the tag's color, its name, and a
+at the end. A session with several tags appears under each of them; every copy is the same
+session (click either to open it, and closing either ends it). Each header shows a chevron, the tag's color, its name, and a
 member count. Click a header, or press Enter on it, to fold its section down to the header
 alone; the count then says how many tabs are folded away, and a small dot after it says when
 one of them is working or waiting on you (hover it for their names). To keep one tab visible
@@ -85,7 +85,7 @@ folds (its header says so, and a click there changes nothing), and the `archived
 starts folded. Drag a header to reorder the groups, which
 reorders the tags on every surface (the timeline's tag table shows the same order). To move
 a tab into another group, right-click it and pick **Move to <tag>** under **Tags**: one click
-adds that tag and drops the tab's current group tag, leaving its other tags alone. The row's
+adds that tag and drops the tag of the group you right-clicked it in, leaving its other tags alone. The row's
 **+** adds the tag without moving the tab. **Group tabs by tag**, at the foot of the tag
 button's menu, turns the sections off for this browser.
 
--- a/docs/read-side.md
+++ b/docs/read-side.md
@@ -357,12 +357,12 @@ jobs:
 - **Filtering.** Each surface (chat tabs, timeline lanes, outline) keeps its own
   lens: every session, the untagged ones, or any set of tags.
 - **Grouping.** The chat tab strip sections by tag whenever a session carries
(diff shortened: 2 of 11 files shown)


ksarma:
No objection to the merge as it landed. Nothing to revert; three follow-ups below.

Reviewed inline: `sectionTabs` lists every visible member per tag, and the untagged trail holds only sessions in no tag. Each copy renders as a full tab with `data-copy`, selected through the stable `#tabs` delegate by `data-id`, so a click bubbles once and cannot double-fire. `flipTabs` is keyed per copy. `planStrip` keeps a session in the keyboard order while any copy is on screen. The drag tracks the dragged element, and the drop skips the session's own copy. Pins, counts and folds are computed per group. Against the first commit, the review found the arrow keys dead when the active session's own copy was the nearest tab in the adjacent row. The Show-when-folded and Move-to rows addressed the first holder from every copy. The docs and module comments still stated the home-tag rule. The second commit fixes each of these. `tabInAdjacentRow` gains an own-copy skip and a focused-copy origin. `showTabMenu` receives the right-clicked copy's group. `guide.md` and `read-side.md` are rewritten and pinned. The equivalence ruling is consistent with the inheritance semantics of #1007. `revealIn` still works because the first holder is one of the tags shown. The merge landed under the tier check's rule that a feature by an admin author merges on green.

- `ui/webview/tab-groups.test.ts:1660` (follow-up): the drag-and-drop changes and the copy-scoped menu are covered by source-regex pins only. The second commit itself fixed a drop-walk defect those pins did not catch. An executed drag and menu step in `tests/test_tab_groups_rows.py` would cover both.
- `ui/webview/render.ts:6324` (follow-up): a click on a later copy still leaves focus on the first copy. The next arrow key then measures from a row the user did not click in.
- `ui/webview/styles.css:360`, `render.ts:6095`, `tab-groups.ts:221`, `tab-tags.test.ts:85` (follow-up): residual home-tag wording. The PR body also still says the Move-to row was left unchanged.

Tests: node suite 3435/3435 at 0f2f6f14 and 3439/3439 at cc57e0c2. The served two-tag Playwright test passed at both heads. With the new tests overlaid on the base b6778dcb, 17 node tests and the served test failed. CI green at cc57e0c2, Tier policy pass.

Review run by the maintainer's automated review session; verified against head 0f2f6f14.

PR #1092, opened 2026-09-08
Usage modal second pass: chart first, every session its own stack and row, a scrolling list of tab titles, and an ordering control

henrypinkard:
**Feature (T247e).** The user's second pass on the usage modal (2026-09-08 morning PT): they like it and want these changes.

**Design points**
- **Order.** Totals, then the spend-over-time chart with its range and measure toggles, then the sessions.
- **Scroll pane.** The session list shows every session inside a capped pane with the header row (session / dollars / turns / tokens) sticky at its top; the modal stays a centered card (panel rule) and the list scrolls inside it. The "N more sessions · show all" fold is gone.
- **Every session its own stack.** No top-N and no "other" stack; each session stacks in its own color. A session the registry never colored gets a deterministic swatch of the active palette by sid (flagged `bgDerived`), so every stack and row has a color. Dead sessions keep their identity color; the hatched "unattributed" stack stays. An older peer that still folds beyond a top-N keeps its "other" as that host's own transitional stack, named with the host, so its dollars are not lost.
- **No legend.** The list below the chart is the legend.
- **Rows as tab titles.** No swatch. A row renders like the tab strip renders a tab title: the host prefix in quiet italic gray ("devbox:"), then the name bold in its identity color, under the strip's own class names (`.tab-label.colored` carrying `--chip-bg`, `.host-prefix`). The landing page loads no stylesheet (the panes do), so the two declarations are inlined as the strip's twin, and a test pins the twin against `ui/webview/styles.css` byte for byte so the two cannot drift, the timeline's menu twin being the precedent. Host prefix only when more than one machine contributes; dead sessions stay dimmed with their "not running" annotation.
- **Kept.** The footnote lines (timezone rule, per-host status), the phone door, the toggles' pressed styling and click safety, the residue grain.

**Tests.** Route tests follow the shape: no `topN`, no "other" stack, every session a stack with a color, the colorless session's deterministic swatch, an older peer's fold kept per host, the small session's own hourly stack across hosts, wiring pins for the scroll pane, sticky header and the twin's drift guard. The served test drives a twenty-session two-host world: chart before list, no legend, no swatches, every row present, the first row's title in its identity color at weight 600 with its host prefix, the pane scrolling under a sticky header, the drag-select and dim-once checks unchanged. Screenshots dark and light.

**Ordering control (second commit, T247f).** A third chip pair beside "dollars / tokens": "by spend" (default, descending dollars) and "your order". "Your order" is the order the tab strip and the timeline lanes show, from the one source they share: the kernel's shared seed per host (`session-order.json`, written by a tab drag or a lane drag), hosts local-first then in the remotes listing's order, arranged by this viewer's own drag order read from the strip's `localStorage` key and applied with `view-order.ts`'s own algorithm. Both layers are needed because the strip and the lanes are the seed arranged by the viewer (`mergeHostOrder` then `applyViewOrder`); the seed alone would not match what the user sees. The landing page loads no webview bundle, so the modal carries a twin of `applyViewOrder`, and a node test runs the twin against the module input for input. Sessions the order does not know (dead or archived, or an older peer's) trail in spend order, dimmed as today; the chart's stacks follow the same choice, bottom to top. The choice persists with the range and measure toggles, which persisted nowhere before; all three now ride one `localStorage` entry. Same delegated `data-act` door as the other chips. Tests: route pins for the host-tagged seed; the served pass presses "your order" against a saved order and a viewer arrangement, reads the rows, the chart's bottom stack and the persisted choice; a screenshot with "your order" pressed.

**Review folds (third commit).** Adversarial review of the second pass, twelve findings, ten real, folded: the sticky header was a 55 percent wash (opacity on the header cells composites the background too), so the muting is now the text color and the header occludes; the card itself still scrolled at common heights, so the pane's cap is computed from the room left under the chart; a derived color could be a swatch an identity-colored session in the same payload already wore, so the least-used swatch goes first with ties by the sid's hash; a segment under half a dollar tipped "$0" and now tips "under $1"; resize re-renders coalesce on one animation frame; dead code, a no-op rule, stale comments and an unread stack flag are gone. **Stated, not folded (design limits to weigh):** twelve swatches cannot keep more than twelve sessions apart, so past that the color maps stack to row only through the tooltip; with dozens of sessions active in one hour, per-session segments thin to sub-pixel hairlines; an older peer's transitional "other" stack is named only on hover; the payload is served uncompressed and grows with the session count.

**Review folds, ordering control (fourth commit).** Eight findings, seven real, folded: the seed shipped `session-order.json` verbatim, and the file keeps a recently dead session at its slot while its transcript is in the discover window, so "your order" interleaved a dead session the tab strip did not show; the seed is restricted to what the strip renders (live or kept open), so dead sessions trail as the contract says (this follows the strip, not the timeline lanes, which show recently dead sessions in place). The chip row wraps on a phone and a chip never breaks its label; an order toggle scrolls the pane back to its head rows; the apply-order twin keeps ids that collide with `Object.prototype` keys, as the module's sets do, with that case in the drift test; the persisted choice is read back across the phone reload in the served test; a dead scaffolding line in the driver is gone.

**Left out.** Nothing from the dispatch.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -32596,10 +32596,14 @@ def _spend_detail_local(now=None):
             "tz": time.strftime("%Z", lt), "tzOffsetMin": int((getattr(lt, "tm_gmtoff", 0) or 0) // 60),
             "recordedAt": _spend_recorded_at(),
             "sessions": sessions,
-            # the shared session order (session-order.json — what a tab drag or a lane drag writes): the
-            # modal's "your order" seed (T247f); the viewer's own arrangement is applied client-side, the
-            # way the strip and the lanes apply it (view-order.ts)
-            "order": _session_order(),
+            # the shared session order (session-order.json — what a tab drag or a lane drag writes),
+            # restricted to the sessions the TAB STRIP renders (live, or kept open): the modal's "your
+            # order" seed (T247f). The file keeps a recently dead session at its slot while its transcript
+            # is in the discover window; the strip does not show it, and the modal follows the strip, so
+            # every session no longer running trails in spend order (review find: shipped verbatim, a
+            # session dead an hour interleaved with live rows). The viewer's own arrangement is applied
+            # client-side, the way the strip and the lanes apply it (view-order.ts).
+            "order": [s for s in _session_order() if s in live or s in set(_kept_open)],
             "unattributed": {"usd": round(un[0], 4), "tok": un[1], "turns": un[2]},
             "hours": hrs, "days": _series(days, day_keys)}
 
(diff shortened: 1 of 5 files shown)


ksarma:
Reviewed after the merge: nothing blocked it. Three follow-ups and one nit below.

Reviewed inline: the kernel hunks stay inside `_spend_detail_local`, `_spend_detail`, `_merge_spend_details`, and the modal's inline JS and CSS. Dropping `topN` breaks neither an older landing page nor an older peer. The landing page still reads `d.topN||10`; the peer's fold survives as that host's own stack. The `applyViewOrder` twin is pinned input for input against the module, `Object.prototype` keys included. The seed follows the strip's roster, live or kept open. Ids are host-prefixed the way federation's `prefixId` does it. The arrangement is read from the strip's localStorage key, which the landing page shares with the panes on one origin. A remote `host:sid` arrangement moves the peer's row first. Chips use the existing delegated `data-act` handler on the stable panel and press synchronously. No new font size: 0.86em is the strip's own declaration. Accent appears only on the pressed chip. The sticky header occludes at opacity 1. Resize coalesces on one frame. The card stays centered over the dimmed, unchanged dashboard. The payload measures about 1.1 KB per session uncompressed (17 KB at 3 sessions, 83 KB at 63). The four stated design limits are accurate. One design question for the approver comes from T247c rather than this PR. With multiple hosts, the modal prefixes local rows with the host, while the strip prefixes remote tabs only.

- `kernel/kernel.py:41017`, follow-up. The range/measure/order handler ends with `renderChart()` and never calls `spSizePane()`; the initial render and the resize path both do. When a toggle changes the chart's height, the pane keeps its stale cap and the card scrolls again until a resize. Example: hourly range empty after eight idle days, daily populated. Reproduced headlessly at 1280x820. One call after `renderChart()`, plus a toggle-then-measure step in the served driver, fixes it.
- `kernel/kernel.py:42980`, follow-up. The host-prefix twin's dark fallback `#9aa0a6` is the landing's muted gray, not the strip's `:root` `--dim` fallback `#9a9a9a`. The pin at `tests/test_spend_detail.py:617` substitutes the literal instead of reading it from `styles.css`. A `--dim` change therefore moves the strip but not the modal, and the test stays green. The light twin at 43302 has no pin. Derive both fallbacks from `styles.css` in the test, as `menu-theme-tokens.test.ts` does.
- `kernel/kernel.py:40955`, follow-up. The strip's storage key and the `host:sid` shape are literals with no cross-module pin; the twin test imports `applyViewOrder` only. A `VIEW_ORDER_KEY` rename left every spend test green while your order silently returned the seed order. One assertion reading `VIEW_ORDER_KEY` from `view-order.ts` fixes it.
- `kernel/kernel.py:32606`, nit. The seed's liveness set is `_live_names` (registry-named sids), while the strip renders `_alive_sessions`. A running session with no names entry therefore shows a tab but trails in the seed.

Locally at head, `tests/test_spend_detail.py` 22 passed. `tests/test_spend_modal_headless.py` 1 passed with Playwright chromium; eight screenshots inspected. vscode-extension `npm test` 3441 passed. For information, the PR's test files are red at the PR base: 7 of 22 route tests fail, the served driver times out waiting for the order chip, and the twin test fails. Privacy scan of added lines: no hits. CI: all nine checks green at head; the tier gate waits on approval.

Review run by the maintainer's automated review session; verified against head 069376fa.

PR #1096, opened 2026-09-08
A peer polling a faulting tag store keeps its last reading, and the heal repaints at once

lczh:
Follows #1075, whose review commit made a views-store fault say so on every frame and acknowledgement instead of showing an empty store. Two consumers were still outside that rule.

## What was wrong
Verified on #1075's head: the views route still answered a 200 with the plain blob, and under a fault with nothing good known that blob is the empty default. A peer's poller stores any 200 as this host's tags and keeps its last reading only on a non-200, so during such a fault every peer erased this host's tags in its own view. And when the fault ended, nothing moved the pusher: the fault's start had changed the view signature through the notice, so the marked payloads were built, but the clean read that ended the episode changed nothing, so the marked feed and timeline stood until the next clock bucket or an unrelated change.

## What changes
The views route follows the frames' contract. With nothing good known it answers a retryable 503 in the person's words, naming the step and the errno and no file, and a peer's poller keeps its last reading and stamps nothing. With a last-good blob it answers a 200 carrying that blob marked, which the peer stores once and drops once when the mark departs; a peer reads only the tags and sequence. The first clean read that ends a fault episode marks the views dirty exactly once, so the pusher rebuilds the marked payloads on the heal. The memory-conserving pass was checked and left alone: under such a fault its lens reads as "all", so nothing closes; a comment says so. The CLI's `romp tag` reads the 503 as an unreachable kernel today; the companion change that prints the kernel's words on a 5xx follows this one.

## Deliberately left out
The capabilities frame's views sequence still reads the plain blob, so under a cold-cache fault it announces nothing, and the client's gate stands as before. A peer rendering another host's fault marker would be a feature. The CLI's wording is the companion change's. One pre-existing shape the review named stays as it is and is worth its own change: the peer's pending-tag-edit apply proceeds on the kept reading and, when the faulting host answers its forwarded edit with a retryable refusal, retires the pending edit as final, so a pending delete or rename can be lost to a transient disk fault.

## Tests
Two classes on #1075's test-less harness, driving the real handler over HTTP and the real peer poller dialling the harness's own server: the 503 on the read and stat arms with the poller keeping its reading and the notice filed once, the healed 200 unmarked, the primed-cache 200 marked with the marker the only difference and the peer storing it once; and exactly one dirty mark per episode end across cold, primed and missing-store cases, zero on a hit or a clean miss, with the real cached feed under one fixed signature served as-is until the heal and rebuilt once after. On #1075's head the four new cases are red. A latent harness bug found on the way, a helper rebinding the harness's saved seams so a later teardown restored the wrong function, is fixed.

`tests/test_tag_route.py` 144 passed on this branch; the four new cases are red on #1075's head. Reviewed by read-only agents: the route, the peer and the mark hold; one wording nit folded.


lczh:
Merged main into this branch to clear the conflict in kernel.py: both sides had added the retryable-refusal arm in _apply_pending_tag_edits (yours here in the review fixes, main's through #1094 with the once-per-episode fault log), and the resolution keeps main's arm, which the code right after it already depends on and which is a strict superset of the one here; every other line of both sides stands. One more small commit on top: the review test for the journaled delete on a kept reading read the views file inside the block that makes that very read raise, so its own probe tripped while the kernel assertions held; the store check now sits just after the block. It went unseen because a conflicting PR runs no test workflow. tag_route 148, timeline_views 44, tag_federation_v2 28 pass on the result; nothing of yours was rewritten.

PR #1097, opened 2026-09-08
Dashboard: reload itself on a kernel restart and on a newer served bundle, never mid-gesture, keeping the reader's place

henrypinkard:
## Summary
The dashboard reloads itself after a kernel restart and after a newer served bundle, never mid-gesture, and the chat reader keeps their place. The user's ruling of 2026-09-08 supersedes their 2026-07-13 preference for a banner the reader clicks; the comments that cited the old rule now say so with both dates.

## Design
**One reload core on every kernel-served page** (`_RELOAD_CORE_JS` in kernel.py, `window.__rompReload`), embedded by the shim on every pane page and by the landing ahead of the stale-banner script. Two signals, both exact events:
1. **Kernel restart**: a socket REOPEN whose `/version` answers with a boot id other than the one baked into the page (`checkBoot`). The shell asks on its own socket's reopen; a standalone pane asks on its shim's reconnect; a pane inside the shell leaves it to the shell. The banner's existing 30 s `/version` poll hands the core every reading as the backstop for a page whose socket never dropped.
2. **Build drift**: a `dv` on a keepalive above the page's baked `LOADEDV` (`noteDv`, from the shim's existing check and now the shell's own socket too), or `/version`'s `dist_ver`.

**Never mid-gesture.** A held pointer button, a drag in flight, a text selection being made, or the composer focused with text arms the reload; the ending event fires it (pointerup/pointercancel, dragend/drop, selectionchange, input, focusout; a window blur releases every hold). No timer. The shell composes gesture state across its same-origin panes and every pane forwards its request to the shell, so one decision reloads the top document. A pane under a foreign parent (an iframe in another app) decides for itself; an iframe can reload.

**Before reloading**: every pane's `window.__rompPersistForReload` runs synchronously (the chat pane records the active tab's scrollTop, follow mode and anchor turn beside the drafts and active tab it already persisted; `pagehide` is the belt), the shell's lifted modals close (settings/picker), and a marker rides sessionStorage. **After**: `landActive`'s first show of that tab lands a follow-mode reader at the bottom and anyone else on their anchor turn, else the raw scrollTop (`ui/webview/reload-restore.ts`), and the fresh page leaves one notification-center line: "Reloaded onto build N — the kernel restarted" or "— a newer romp build was served".

**The banner** stays only as the core's `refused` fallback (a host whose `location.reload` throws). The connection prompt for a same-process socket blip is untouched. **VS Code**: the webview never runs this code; its bundle comes from the installed VSIX and a webview reload cannot fix bundled-code drift, so its reload prompt stays a user click (comment updated in extension.ts to say why, beside the new ruling).

**Federated relay**: a remote kernel's restart cannot reload the page. The core reads only this page's own socket and `/version`; federation drops remote `ka` frames and the relay never forwards a remote boot id. Pinned.

## Tests
- `tests/test_dashboard_auto_reload.py`: node executes the real core (one process per scenario) with fakes: dv drift fires when idle and persists first; a reopen against the same boot is a blip, a new boot fires; `/version` readings feed both signals; held pointer, drag, selection and composer-with-text each arm and fire on their ending event; a window blur releases; a refused reload falls back to the banner hook; the fresh page announces once from the marker; the shell composes pane gesture state and a pane forwards its request; a foreign parent leaves the pane to reload itself. Plus wiring pins (shim raise and reconnect, shell socket, banner fallback and poll, every page embeds the core with its build and boot, remote exclusion, both dates recorded).
- `ui/webview/reload-restore.test.ts`: the pure restore decisions plus render.ts wiring pins (synchronous persist hook, pagehide, load-time take, landActive's one-shot consume).
- `tests/test_dashboard_reload_served.py`: the real page. Holds a pointer over the chat pane, bumps a dist bundle's mtime so the kernel's next keepalive carries a higher `dv`, asserts no reload while held (armed on the pointer), releases, asserts the reload, the restored scroll position and the notice; then kills and relaunches the kernel and asserts the restart reload with its wording. Skips loudly without node playwright (CI installs none); verified locally with the python playwright twin of the same steps:

| step | result (run 2 / run 3, headless Chromium) |
|---|---|
| reader scrolled to mid-history, then pointer held over the chat pane, dist bundle mtime bumped | no reload for 6.5 s (three keepalives); the shell reports the reload armed on the held pointer with reason `build` |
| pointer released | page reloaded; chat scrollTop 6281 → 6282 (anchor turn offset −40 → −41 px); go-to-bottom chip visible; one notice "Reloaded onto build N — a newer romp build was served." |
| fresh page left alone for three more keepalives | no second reload |
| kernel killed and relaunched on the same port | page reloaded once more; second notice "Reloaded onto build N — the kernel restarted." |

- Updated: `test_kernel_ws_heartbeat.py`, `test_kernel_disconnect_banner.py` pins moved to the core routing.
- Python suite: __PY__. `npm test`: __NPM__. `npm run typecheck` clean.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -34795,9 +34795,10 @@ WS_DEAD_S = float(os.environ.get("ROMP_WS_DEAD", "0")) or 3 * KEEPALIVE_S
 def _keepalive_all(now=None):
     # `dv` = the current dist build token (the same value baked into every page's ?v= urls). Riding the
     # keepalive makes build-drift detection EVENT-based on every surface with a live socket: a page whose
-    # LOADEDV is older raises the reload banner within one heartbeat of a rebuild — no per-page /version
-    # polling. The VS Code extension compares it against its bundled build stamp the same way (the user
-    # 2026-07-13: a stale tab sat silent through several rebuilds; drift must always show a banner).
+    # LOADEDV is older reloads itself within one heartbeat of a rebuild (T265, the user 2026-09-08; until then
+    # it raised the reload banner, the user 2026-07-13, whose stale tab sat silent through several rebuilds) —
+    # no per-page /version polling. The VS Code extension compares it against its bundled build stamp the same
+    # way and keeps its banner: a webview reload cannot fix bundled-code drift.
     s = json.dumps({"type": "ka", "dv": _dist_ver()})
     now = _ws_clock() if now is None else now
     with _clients_lock:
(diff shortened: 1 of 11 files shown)


ksarma:
Already merged. The change stands. It owes one prompt follow-up and a relabel. This is a feature, not a fix: it replaces a recorded design with a new mechanism on a new ruling, names no bug, and the body carries a Design section. Under the policy now on main, the label changed no gate for an admin author. The relabel is for the record.

Reviewed inline: the reload core is embedded once per kernel-served page, by the pane shim and by the landing ahead of the stale-banner script. Both signals are exact events: a socket reopen whose `/version` boot id differs from the baked one, and a keepalive dv above the baked `LOADEDV`. The 30 s `/version` poll stays as backstop only, and no timer was added. The gesture guard composes across same-origin panes through the shell. A pane under a foreign parent decides alone, and VS Code keeps its click prompt. The banner survives as the fallback when `location.reload` throws. The injected-voice rule does not apply: the one new line is notification-center copy on the dashboard, not text put into a session. The federated pin holds: remote keepalive frames are dropped in `federation.ts`, and the core reads only its own origin's `/version`. The heartbeat, disconnect-banner and mobile pin edits move those pins to the new routing without weakening them. The old hand-off to the shell banner is asserted gone. The privacy scan of the added lines is clean, and the fixtures are synthetic. After a core-fired reload, a reader mid-history returns to the same turn, a composer draft survives, and the fresh landing posts one notice. The cases that do not hold are listed below: a send queued during the restart outage, a standing selection, a swallowed pointerup, text in a non-composer field, and a plain re-open.

- `kernel/kernel.py:39642`, with `fire()` at 39386 (blocker, now a prompt follow-up): a message sent while the sockets are down queues in the shim. When the shell's socket reopens first, the restart reload fires without flushing the queue. The send is gone, with no draft, bubble or notice. Reproduced in Chromium against a real kernel restart: 5 of 6 sends lost. Before this PR the pane always flushed before the user clicked the banner. Hold the reload while the shim's queue is non-empty, and call `ended()` from the onopen flush.
- `kernel/kernel.py:39378`, 39410 (follow-up): `busyHere` returns 'selection' for any standing selection, and blur resets only `ptr` and `drag`. A word selected in the chat pane therefore holds the reload, with no banner, until the reader clicks back into that pane. The body's statement that blur releases every hold does not match the code.
- `kernel/kernel.py:39401` (follow-up): `ptr` is a counter with no reset. A right-click that opens the native context menu on Linux or macOS, or a native select popup, swallows the pointerup. `ptr` then stays at 1 until a window blur, and every reload is held with no cue. Reset when `e.buttons` is 0.
- `kernel/kernel.py:39379` (follow-up): only `#composer-input` is guarded. A signal that lands while the user types in the picker's directory input, the session search, the remote-kernels host field or a settings field reloads at once. That reload drops the text and closes the lifted modal. Guard any focused text-entry element with a value.
- `ui/webview/render.ts:10382` (follow-up): the pagehide belt writes the scroll record into localStorage on every teardown, and the take never checks that the load is a reload. A tab closed mid-history therefore re-opens mid-history days later with no notice. At main a fresh open landed at the bottom. Gate the take on a reload signal or drop the belt, and state the intended plain-open behaviour in the body.
- `kernel/kernel.py:39387` (follow-up): a standalone pane page that reloads itself leaves `romp:reloaded` in sessionStorage, and nothing on that page consumes it. A later ordinary navigation to `/` then announces a restart from arbitrarily long ago. Consume it in the pane's core when not `inShell()`.
- `tests/test_dashboard_reload_served.py:237` (follow-up): `cfg.json` carries the runner's whole environment, credentials included, for the run's lifetime. The same pattern is in `test_ship_reship.py:301` at main. Write only the overrides and merge them over `process.env` in the driver, keeping `ROMP_STATE_DIR` and `ANTHROPIC_API_KEY` out.
- Body (follow-up): the Tests bullet still reads `__PY__` and `__NPM__`. The Updated line omits `test_kernel_mobile.py` and `pane-shim-stale.test.ts`.
- `tests/test_kernel_ws_heartbeat.py:111` (nit): the class docstring still states the superseded never-automatic rule.

Red-first confirmed at the parent b896d722 with the PR's test files. `test_dashboard_auto_reload.py` failed 18 of 18 (no `_reload_core_js`, missing source strings). `reload-restore.test.ts` failed to build on the unresolved import. `pane-shim-stale` failed 14 of 14, the three moved pins failed, and the served test failed with no reload after release. At 63c7fb23: the five Python files 76 passed, with the served test running in real Chromium rather than skipping. npm test 3446 of 3446, typecheck clean. CI: all nine checks green.
Review run by the maintainer's automated review session; verified against head 63c7fb23.

PR #1101, opened 2026-09-08
The auto converge moves the local main branch, not just a detached head

henrypinkard:
## Problem
`_run_main_update("pull")` checked the advertised commit out **detached** and never touched the local `main` branch, so an auto-converged box sat at a fresh sha while `git checkout main` landed on a pointer months old; the user then pulled an enormous amount by hand.

## Fix
After the existing gates (clean tree, fetch, HEAD an ancestor of the target):
- **main is an ancestor of the target** → `git checkout -B main <target>`: main moves onto the target and is checked out. A fast-forward by construction, so nothing of the user's is rewritten.
- **main has commits the target lacks** (diverged) → the target is checked out detached as before, main is left where it is, and an informational sync notice says so and names the remote. Moving it is the user's call.
- **no main branch** (a bootstrap install detached at a release tag) → detached, as before.

The refusal when HEAD itself has diverged is unchanged and runs first.

## Tests (red before)
`tests/test_converge_main_branch.py` drives the pull step over real temporary repositories (a bare remote, the checkout, a second clone advancing main): main behind and detached is fast-forwarded and checked out; main checked out and behind moves on the branch; main with a local commit is left alone with the notice while HEAD advances detached; no main stays detached; a diverged HEAD still refuses before any branch move. The scripted happy-path pin in `test_main_drift_notice.py` follows the second ancestry check and the branch checkout. Converge and update suites: 114 passed.

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -7462,11 +7462,29 @@ def _run_main_update(kind, immediate=True, manager_port=_PORT_FROM_ENV, target="
                 refuse("the checkout was left alone: the fetch did not bring %s, so it could not be "
                        "verified; the next check re-reads main" % target, anc)
                 return
-            r = subprocess.run(["git", "checkout", "--detach", target], cwd=str(ROOT),
-                               capture_output=True, text=True, timeout=30)
+            # The local `main` BRANCH moves too (the user 2026-09-08): the converge used to check the
+            # target out DETACHED and never touch `main`, so a later `git checkout main` landed on a
+            # months-old pointer and the user pulled "an enormous amount". When main is an ANCESTOR of the
+            # target (it has nothing the target lacks) it is moved onto the target and checked out — a
+            # fast-forward by construction, so nothing of the user's is rewritten. When main has commits
+            # the target does not (diverged), it is the user's to move: the target is checked out
+            # detached as before and the notice says main was left where it is. No main at all (a
+            # bootstrap install detached at a release tag): detached, as before.
+            mb = subprocess.run(["git", "merge-base", "--is-ancestor", "main", target], cwd=str(ROOT),
+                                capture_output=True, text=True, timeout=10)
+            if mb.returncode == 0:
+                r = subprocess.run(["git", "checkout", "-B", "main", target], cwd=str(ROOT),
+                                   capture_output=True, text=True, timeout=30)
+            else:
+                r = subprocess.run(["git", "checkout", "--detach", target], cwd=str(ROOT),
+                                   capture_output=True, text=True, timeout=30)
             if r.returncode != 0:
                 refuse("the checkout did not advance onto %s" % target, r)
                 return
+            if mb.returncode == 1:
+                _sync_notice("main moved at %s: the checkout is at %s, detached. Your local main branch has "
+                             "commits that are not on %s/main, so it was left where it is; merge or rebase it "
+                             "yourself when you want it on the new main." % (remote, target, remote), ok=True)
         except Exception as e:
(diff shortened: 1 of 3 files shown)


ksarma:
No blocker; this review is advisory because it ran after the merge (c896abee). Three follow-ups and one nit are listed below.

Reviewed inline: the pull step keeps its gates in order (named target, clean tree at the checkout, fetch, HEAD an ancestor of the target, the diverged-HEAD refusal first). It adds one ancestry test on main. Exit 0 runs `checkout -B main` onto the target, a fast-forward by construction that keeps main's tracking config and reflog. Exit 1 keeps the detached checkout and posts an informational notice. Exit 128 (no main ref) detaches silently as before. Probes over real temp repos: a stash survives, the clean-tree gate refuses an untracked file in the target's path, a symbolic main writes through to its target branch, and a missing main detaches without a notice. The notice is a kernel row on the dashboard bell, not text injected into a session, so the injected-voice rule does not apply. It fits the badge cut with an eight-letter remote name. The new tests drive bare-remote and clone repos under a private temp root. They patch the checkout root, state dir and converge hooks and poison the manager port, so they never reach a live checkout or manager; conftest keeps git config hermetic. There is no overlap with #1121 (merge-tree clean; it changes only the p2p apply script's restart flag). The p2p apply's `reset --hard` already carries whatever branch the far checkout is on.

- `kernel/kernel.py:7476`, follow-up. `checkout -B main` does not check whether main is checked out in another worktree. With the checkout root detached (the state every earlier converge left) and a linked worktree holding main, git 2.43 moves main under that worktree with exit 0 and no notice. Its HEAD advances while its index and files stay at the old commit. So `status` there shows the update's files as staged deletions, and a `commit -a` there records a revert of the update onto main. On git 2.44 and later the same command refuses. The converge then takes the refuse path, re-arms, and refuses again every cool-down, so the box stops converging. The previous detach never touched main. A `worktree list --porcelain` check for a branch `refs/heads/main` entry outside the checkout root would close both cases: fall to the detached path and post a notice naming the holder. None of the five new tests builds a linked worktree.
- `kernel/kernel.py:7485`, follow-up. The main-left-alone notice lives only in the in-memory sync ring. The same call continues into the dist rebuild and an immediate restart-all whenever kernel code changed. With no dashboard connected during those seconds, the new kernel starts with no record of the notice. The unattended auto-mode box is the one that loses it. Two ways to carry it: the `update-report.json` pattern (the fresh kernel posts at boot), or a boot pass that derives it from a detached HEAD with a non-ancestor main. The rebuild-failed row at the same position has the same pre-existing exposure.
- `kernel/kernel.py:7467`, follow-up. The rationale comment puts a phrase attributed to the user in quotation marks. The privacy rule asks for a paraphrase, and the commit message already has the unquoted wording.
- `kernel/kernel.py:6868`, nit. The `_checkout_branch` docstring still says auto-converged boxes sit detached at main shas. After this change they sit on main unless main diverged.

Test evidence: at origin/main with the PR's two test files, 4 of 54 fail for the stated reasons (main not moved in two cases, no notice in the diverged case, the scripted happy path lacking the second merge-base step). At fded63e4 all 54 pass, and the four converge/update suites give 173 passed (the body's 114 matches no subset). CI: all nine checks pass at fded63e4 after a re-run. The earlier red was Python 3.10 on `tests/test_file_github.py`, a process-kill timing test in a file this PR does not touch. It was green on 3.11 to 3.13 in the same run and three times alone here.

Review run by the maintainer's automated review session; verified against head fded63e4.

PR #1112, opened 2026-09-08
Feed: a session's group folds per column in group-by-session mode (T263c)

henrypinkard:
When cards are grouped by session, folding a session's Blocked (or Completed) cards now leaves its Working cards open (the user 2026-09-08). The fold was keyed by session alone, so the caret on any column's header folded the session in every column.

- **Keying fix, no new control.** The fold key is `(session, column)` (`threadKey` in `feed-view-state.ts`, pure and executed in tests). The header entry carries its column; the caret reads and toggles that key and its tooltip says which column it folds. A card that lands in the column later inherits that column's fold; a card that moves columns shows under the fold state of the column it moves to; a jump into a folded card unfolds the run it sits in. The header count was already per run.
- **Persistence unchanged in shape.** The same `threads` list, prune-exempt as before. A stored bare sid (the earlier spelling, when a fold covered the session everywhere) reads as every column, so an older fold keeps folding until a column is opened.
- **Tests.** `feed-thread-fold.test.ts` pins the per-column key on every site and bans the bare-sid spelling; `feed-view-state.test.ts` executes the key helpers and the round trip; `tests/test_feed_thread_fold_columns.py` drives the served feed page in a browser with a synthetic payload (one session, blocked and working): the Blocked header folds with its count, the Working header stays open with its card, and the fold survives a reload. Red on main: the Working header folds too.

### DIFF
--- a/tests/test_feed_thread_fold_columns.py
+++ b/tests/test_feed_thread_fold_columns.py
@@ -1,0 +1,200 @@
+#!/usr/bin/env python3
+"""T263c (the user 2026-09-08): in the feed's group-by-session mode, a session's run folds INDEPENDENTLY per
+column — collapse its Blocked (or Completed) cards while its Working cards stay open.
+
+The fold used to be keyed by session alone, so the caret on any column's header folded the session across the
+board. It is now keyed by (session, column). The served guard drives the real /feed page from a hermetic
+kernel and hands it a SYNTHETIC payload (one session, one card blocked on the user and one working) through
+the page's own frame path (a MessageEvent, exactly what the socket shim dispatches), then clicks the caret on
+the session's header in the Blocked column and reads the DOM: that header folds and shows its count, the same
+session's header in the Working column stays open with its card visible; a reload keeps the per-column fold.
+Red on main: the Working header folds too (one key per session).
+
+Screenshots with FEED_FOLD_SHOTS=<path-prefix> (dark + light). Skips LOUDLY without the extension deps or a
+Playwright browser (CI installs none); the CI-safe pins ride ui/webview/feed-thread-fold.test.ts and the
+executed key rules ui/webview/feed-view-state.test.ts. All fixtures synthetic (the notes-api demo world).
+"""
+import json
+import os
+import shutil
+import socket
+import subprocess
+import tempfile
+import time
+import unittest
+
+HERE = os.path.dirname(os.path.realpath(__file__))
+ROOT = os.path.dirname(HERE)
+BIN = os.path.join(ROOT, "bin")
+EXT = os.path.join(ROOT, "vscode-extension")
+
+SID = "aaaaaaaa-1111-2222-3333-777777777777"
+BLOCKED = "cccccccc-1111-2222-3333-000000000001"
+WORKING = "cccccccc-1111-2222-3333-000000000002"
+
+
+def _free_port():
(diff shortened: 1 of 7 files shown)


ksarma:
Reviewed after the merge (0004b019). Nothing here would have blocked it; the findings below are follow-ups.

Reviewed inline: the fold key is now (session, column) via `threadKey`. The render filter, the caret toggle and the load-time migration all read the same key. Fold state survives re-renders: the header node is reused per column and session, and the caret handler is re-pointed on each update. It survives reloads: composite keys round-trip through the existing threads list, still prune-exempt. The bare-sid migration is one-way: a stored sid expands to all three column keys on load, and the next state write persists only composite keys. The caret keeps its existing click path, a local toggle with an immediate render as the acknowledgement. Its title now names the column. A card that changes columns changes visibility only when the kernel moves it, so there is no flap. The jump-unfold fires only from the bell, notification and chat-rail gestures. Privacy and vocabulary scans of the added lines are clean, and the fixtures are synthetic.

- `ui/webview/feed.ts:5858` (follow-up): `unfoldThreadsFor` keys on the member's own column, but a turn group renders under the group's column (`buildGroup`, lines 2555-2556). For a completed member of a Working group, a jump unfolds the session's Completed run and leaves the group's run folded. The bug reproduces in the served page at head, and not at main under the single key. No tracked code produces `groupTitle` today, so no user reaches it yet. The pin at `feed-thread-fold.test.ts:70` asserts the wrong key. Key the unfold on the rendered entry's column and move the pin.
- `ui/webview/feed-thread-fold.test.ts:70` (follow-up): the jump-unfold is pinned only by a source regex. Removing the `collapsedThreads.delete(tkey)` statement leaves every PR test green. No test delivers a payload whose card changes column. The served harness can cover both in a few lines.
- `tests/test_feed_thread_fold_columns.py:112` (follow-up): at main, the post-reload wait for the Working card times out inside the driver. The test therefore reports a 10 s Playwright timeout rather than the assertion at line 188. Wait on the working column's session header instead, or emit RESULT before the reload.
- Label (follow-up): the tier rules class this change as a feature. The 2026-07-31 test header and `feed.ts:677-681` on main recorded per-session keying as deliberate. This PR reverses that and adds a persisted key spelling with a migration. The author is an admin, so the gate is the same. Relabel for the record, and update the `feed.ts:677-681` comment, which still describes the old key.
- `ui/webview/feed.ts:3584` (nit): the caret's aria-label reads collapse or expand plus the session name for all three per-column carets. The title names the column; the accessible name does not.

Test evidence: at head, the four changed node test files pass 41/41 and typecheck is clean. The full vscode-extension suite is 3580/3590. All 10 failures are in tab-strip-skip-exec (`tabDotClass` is not defined). Commit a1f0f7d8 in #1117 fixed that on main; the fix is not in this head's history, and tab-strip-skip-exec passes at main. The served Playwright test passes in chromium. Red-first at main: `feed-view-state.test.ts` fails to import `threadKey`, `threadKeys` and `FEED_COLUMNS`. With that file restored, exactly the five new pins fail (3590/3595). The served test fails as described above. CI: the vscode-extension check at 9dd6c7b0 was red for the inherited tab-strip failure and then cancelled by the rebase. At the merged head c055b879, all nine checks pass.

Review run by the maintainer's automated review session; verified against head 9dd6c7b0.

PR #1113, opened 2026-09-08
Kernel: change-gate the pusher's idle path (comment threads, comments stores, the task-store join, state tails, working notes)

henrypinkard:
## What

The kernel at idle burned most of a core and grew past 5 GB resident on the maintainer's devbox. Measured before any change, from the kernel's `/perf` route and `/proc`:

| measure (13 min after boot) | before |
|---|---|
| kernel CPU | 752 s user + 95 s sys over 847 s wall (about 88% of a core) |
| pusher thread busy | 99% of cycles, cycle p50 6.6 s |
| timeline builds | 84 built, 0 served from cache |
| resident set | 4.4 GB, plus 1.4 GB swap |
| after a restart | 5.3 GB resident at 3 min, 5.7 GB at 6 min, CPU near a full core |

The profile put the idle cost in the pusher re-reading and re-deriving inputs that had not changed. This PR is the first of two batches and adds six change gates to the pusher's idle path, each keyed on the stat of the files it reads (mtime, size, inode) or on the chat build signature the kernel already keeps. Every pushed payload is byte-identical to what the ungated code produced whenever the inputs are unchanged, and any input change produces the fresh result in the same cycle it used to. Memos hand out copies, never their own objects.

- **Thread event slices** (`_thread_events`): served while the thread's chat build signature, its clip and the views-dirty mark stand. This was the largest single item: every comments frame rebuilt every thread's session every cycle.
- **Comments stores** (`_load_comments_cached`): the per-session comments file is parsed once per (mtime, size, inode); callers get a deep copy. Writers are unchanged and still publish by atomic rename.
- **Chat-fold eviction**: the thread sids the previous cycle's comments loop touched are kept, so their folds are not evicted and rebuilt every cycle.
- **Task-store join** (`_task_store_resolve`): a failed join is remembered against the fold pairs that produced it, so an unjoinable store is not re-scanned every cycle; a successful join clears the mark.
- **State tails** (`sdk_backend.last_state`): the states file tail is memoized on its stat, bounded.
- **Working notes** (`_working_notes`): memoized on every entry's stat.

The second batch (the timeline's dead-lane memo, which also drops dead sessions' parses from the parse cache, the resident-memory lever) follows as its own PR. After-numbers for CPU and resident set over an hour will be posted as a comment here once this lands and the devbox converges.

## Tier

`fix`: behaviour is unchanged by design; the work is the pusher doing less of it.

## Tests

- `tests/test_pusher_change_gates.py` (new): the comments-store memo (stat-keyed, copies out, missing file), the task-join miss mark, the state-tail memo, the working-notes memo, and the fold eviction keeping touched threads.
- `tests/test_comment_threads.py`: the thread build is served while its inputs stand and rebuilt when they move (counts session builds).
- `tests/test_perf_stats.py`: the `/perf` builds set includes the thread build kind.

Focused suites: 295 passed. Full suite on the two-batch branch: 9729 passed, 52 skipped.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -11299,11 +11299,22 @@ def _working_notes():
     # Read once per directory VERSION (2026-09-08): GET /sessions is polled about once a second by the
     # postal services of every live session, and each call re-read every note file; the key is every
     # entry's (name, mtime_ns, size, ino), so a rewritten or removed note misses exactly.
+    entries = []
     try:
         with os.scandir(WORKING_DIR) as it:
-            entries = sorted((e.name, e.path, e.stat().st_mtime_ns, e.stat().st_size, e.stat().st_ino) for e in it)
+            for e in it:
+                try:
+                    st = e.stat()
+                except OSError:
+                    continue        # unlinked between readdir and stat (a clear, an atomic write's temp renamed
+                    #                 away): that note is gone and the others still stand. One try around the whole
+                    #                 listing returned {} here instead, so for that call every live session read as
+                    #                 owning nothing, which the postal contract takes as free ownership (review
+                    #                 2026-09-08). The idiom is _task_store_fp's.
+                entries.append((e.name, e.path, st.st_mtime_ns, st.st_size, st.st_ino))
     except OSError:
         return {}
+    entries.sort()
     key = tuple((n, m, s, i) for n, _p, m, s, i in entries)
     hit = _working_notes_memo[0]
     if hit is not None and hit[0] == key:
@@ -12766,7 +12777,19 @@ def _thread_events(tsid, cut_uuid, now, tmux):
     try:
         base = _chat_build_sig(sess, tm)
         asig = _active_chat_sig(sess, tm, now, base=base) if base is not None else None
-        key = ("exact", asig) if asig is not None else (("stat", base) if base is not None else None)
+        sig = ("exact", asig) if asig is not None else (("stat", base) if base is not None else None)
+        if sig is not None:
+            # plus the thread's OWN state rows (review 2026-09-08): the backend writes states/<tsid>.jsonl under
+            # the romp sid, while the key above stats states/<fsid>.jsonl for the reg's lastSid (_sdk_sess hands
+            # over no anchor). The two are one file only until a resume mints a new fsid or a /clear moves
+            # lastSid; after that a states-only write (an interrupt settle's idle row, a retry marker, an
+            # orphan-reply salvage) changed the thread's events with no key change. None when absent.
+            try:
+                ss = os.stat(jd.STATESDIR / (tsid + ".jsonl"))
+                states = (ss.st_mtime_ns, ss.st_size, ss.st_ino)
+            except OSError:
+                states = None
+            key = sig + (states,)
     except Exception:
         key = None                              # an input we cannot key → build, never cache
     hit = _built_thread.get(tsid)
@@ -24688,8 +24711,9 @@ def _task_store_dir(fsid):
 
 
 _task_dir_hint = {}   # fsid → content-joined store dir NAME (see _task_store_resolve); reset per kernel run
-_task_join_miss = {}  # fsid → the fold pairs that failed to join — skip re-scanning until the pairs CHANGE
-#                       (event-based retry: new task activity reshapes the fold; a kernel restart clears both)
+_task_join_miss = {}  # fsid → (fold pairs, tasks-root listing) that failed to join: skip re-reading the stores
+#                       until EITHER changes (event-based retry: new task activity reshapes the fold, a store
+#                       appearing or gaining a file reshapes the root listing; a kernel restart clears both)
 
 
 def _task_store_known(fsid):
@@ -24723,7 +24747,10 @@ def _task_store_resolve(fsid, fold):
     the session's OWN record of creating the tasks: the transcript fold's (id, subject) pairs. A
     candidate store that contains them ALL is the session's store; no match or SEVERAL matches → None,
     and the caller stays loud (never guess). The join runs at most once per session per kernel run
-    (_task_dir_hint caches the winner)."""
+    (_task_dir_hint caches the winner). A MISS is remembered too (_task_join_miss), keyed on the pairs and
+    on the tasks root's listing (each store dir's name and mtime_ns): the root scan and a stat per dir run
+    on every call, cheap; the per-file reads are what the memo saves. A miss that a read fault produced (a
+    store listing that failed, a task file mid-rewrite) is never remembered, so the next call retries."""
     d = _task_store_known(fsid)
     if d is not None:
         return d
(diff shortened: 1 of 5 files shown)


henrypinkard:
After-numbers for this PR, read from the kernel now running it (uptime 12 minutes; the box restarted several times this evening for deploys, so no longer window was available yet):

| measure | before (13 min after boot) | after (12 min after boot) |
|---|---|---|
| process CPU | about 88% of a core | about 40% of a core |
| pusher cycle p50 | 6.6 s | 8.0 s (25 cycles, 514 wakes coalesced into them) |
| thread builds | every thread rebuilt every cycle | 44 served from the memo, 506 built |
| timeline builds | 84 built, 0 served | 17 built, 0 served (the dead-lane memo is the second batch) |
| resident set | 4.4 GB + 1.4 GB swap | 4.3 GB + 1.1 GB swap |

The CPU halved; the resident set did not move, as expected: the parse caches are the second batch's target (the dead-lane memo with parse eviction), which follows as its own PR. The thread memo serves fewer than one in ten builds here because most threads on this box change every cycle; the comments-store, task-join, state-tail and working-notes gates are what the CPU drop reflects. A longer window will be posted on the second batch's PR.

ksarma:
Nothing to undo; two follow-ups to file. The PR landed before this review finished.

Reviewed inline: the six change gates on the pusher's idle path. The comments-store, states-tail and working-notes memos key on `mtime_ns`, size and inode. Their writers publish by atomic rename or append, so the key is exact for them. Each memo returns a copy. The thread build reuses the watched tab's exact signature and yields to `_views_dirty` and a changed cut uuid. With live events in the stub-backend harness, the served frame equalled the first build and a cold rebuild. The review-fixes commit fixed three defects at eabe70c1, each with a test. The task-join miss mark latched until the fold changed. It is now also keyed on the tasks root listing, and a miss from a read fault is not remembered. The fold eviction swapped its keep sets after evicting, so it consulted a set two cycles old. `_working_notes` returned an empty map when one note vanished between readdir and stat. The vscode-extension check was red at eabe70c1 because `comments.test.ts` pins the literal slice line in `_thread_events`. The final commit restores that form, and every check is green at the merged head dab2af44. Request threads and the pusher share these memos without locks. Entries are replaced whole and copies go out, so I found no torn read.

Follow-ups:
- `tests/test_comment_threads.py:369`: the served-versus-rebuilt test runs under the None-backend fixture, where the thread build yields zero events. It compares only the unread bit, so a hit path that serves an empty list passes it. Run it under the `_State` stub and assert whole-frame equality against a cold rebuild.
- `bin/romp:1259`: the builds line iterates a hardcoded tuple that lacks the new thread kind, so `romp perf` shows no thread builds. Only `--json` exposes them. Add it there, pin it in `tests/romp-perf.bats`, and list it with `feedJson` in `docs/reference.md:1064`.

Tests: the PR's three test files at eabe70c1, 161 passed. The same files against `origin/main`, 11 failed: the memo globals absent, the join memo never written, the thread build kind missing. Red-first therefore holds. `npm test` at eabe70c1 reproduced the single vscode-extension failure. All nine checks pass at dab2af44.

Review run by the maintainer's automated review session; verified against head eabe70c1.

henrypinkard:
## The kernel CPU series (T267), every round's before and after

Live kernel on the shared box, `/perf` readings at matched uptimes after each deploy restart, py-spy samples for attribution, no synthetic load. Each round's pull request carries its own table; this is the thread's summary.

| round | change | before | after |
|---|---|---|---|
| pusher change gates (#1113) | the pusher's per-cycle rebuilds gated on change | about 88% of a core at 12 min | about 40% of a core at 12 min |
| timeline memos (#1131) | dead-lane memo, parse eviction, entry encode memo | jobs stage the bulk of wall | at 21 min: jobs stage 64% of wall, push stage 14%, CPU about 86% (the push stage cheap, the jobs stage the target from here) |
| placement-gate memo (#1141) | the nudge walk's planner-placement gate derived once per (parse, store); the walk reads the shared view | jobs stage 78% to 79% of wall at 3 to 6 min, 75% at 22 min; about 90% of a core | jobs stage 45%, 44%, 43% at 3, 6, 20 min; the cycle rate doubled, so per cycle the jobs stage fell from 1.38 s to 0.44 s and process CPU from 1.9 to 0.82 core-seconds |
| clear-set memo (#1156) | `cleared.jsonl` parsed once per file state and hoisted per pass; the awaiting-wake sweep reads the view | goal loads per cycle 56, 55, 80; jobs stage 45%, 44%, 43% | loads per cycle 46, 38, 52; jobs stage 39%, 36%, 33%; pusher thread 13%, 13%, 10% of a core; the clear set served 1630 times against one derivation |
| courier change gate and idle counter (#1158) | the courier skips a session whose inputs have not moved; idle cycles counted | 1172 goal loads per triage pass, judge thread 73% of a core | loads per pass unchanged (1103, 1058, 1017), five sessions recorded in 90 passes: the scan's exact placed check disagreed with the placement loop's drift-tolerant one (found by the counters and a line-level sample); idle cycles 84% to 89% at about 300 ms |
| drift-tolerant scan (#1161) | the scan applies `_placed_key` before queueing | as above | loads per pass 285, 267, 242 at 3.8, 6.5, 20.4 min; judge thread 42%, 37%, 40% against 58%, 61%, 71%; process 61%, 54%, 53% against 80%, 78%, 84%; the gate 109 recorded and 4877 skipped over 171 passes; an idle cycle 83 ms against 349 |
| backref memo (#1170) | the sender-board walk built once per state of the sender stores | called per placed unlinked delegate per pass, one writer load per discovered session each | never asked at all on the deployed kernel (served 0, built 0): the courier gate skips the sessions whose repairs would have called it |
| per-file read memos (#1171) | the captions file parsed once per file state for its three readers; the goal archive loaded once for readers | the three caption readers 19% and the archive loader 11% of the judge tier thread | the captions memo served 24154 reads against 32 parses and the archive memo 10609 against 310 loads at 20 min; the judge tier thread fell to 23% of a core and a judge pass from 4.3 to 4.9 s of wall to 1.6 s |
| planner change gate (#1173) | the planner skips a session whose inputs have not moved since a pass that had nothing to do | planner worker pool 37% of the process's samples, judge thread 45% of a core, process 57% to 64% | the gate skipped 8014 sessions against 48 planned over 278 passes; the planner pool fell to 15% to 16% of a core; with the read memos beside it the process's CPU is 38% to 41% of a core against 53% to 64%, the pusher thread 11% |

**What the series says.** The pusher's own cost fell from most of a core to about 8% to 10% of one; what remained of the process's CPU was the judge side, where every gate and memo followed the same rule: key on every input a pass reads, take the key before the read, serve only while every keyed input stands, and count served against derived on `/perf` so the win is checked on the deployed kernel before it is claimed. The one round that did not move its number (the courier gate) was caught by exactly that check. Parked with numbers: the pusher's event-shaped wait (at most 6% of a core against three timing risks).

PR #1115, opened 2026-09-08
Scroll journal audit: the snap is recognizable from the rows, with tests

henrypinkard:
## What

A pure reader for the chat pane's scroll journal (`ui/webview/scroll-journal-audit.ts`) and its tests. It consumes the rows `scroll-write.ts` already files (`scrollwrite`, `scrollgesture`, `tailchange`, `spacer`) and answers: which scroll events moved the transcript without a write to explain them, whether a tail change or spacer re-size since accounts for each one, and which landing values recur.

## Why

Today's snap (T262: one scroll event after the reader stopped, landing a fixed distance above the bottom, no write, no tail change, scrollHeight unchanged) was found and characterized by reading captures by hand, for hours. With this module the fingerprint is a function call over any journal: the laptop's `client-diag.jsonl`, a kernel-side check, a test fixture. Run over the day's captures it reports the 15 pre-fix moves (all landing 63 to 65 px above the bottom across eleven tabs) and the three at 128 px in a local tab, and nothing unexplained after PRs 1109, 1111 and 1108.

## Design points

- `unwrittenMoves(rows, {minPx, idleMs, settleMs, tolPx, sid})`: a move is an isolated gesture row (no further gesture within `settleMs`) at least `idleMs` after the last recorded position and at least `minPx` away from it. A reader's burst is a stream of small steps, so it never qualifies; the snap is one event.
- `explained`: `tail-change` when a `tailchange` row since the last position has `dh == delta` (a shrinking tail clamps a bottom reader), `spacer` when a `spacer` row has `dTop == delta` (anchoring carries the reader with the content), else `none`.
- `repeatedLandings(moves)`: the fixed value per tab was the snap's signature; a reader's flick lands anywhere.
- `journalRowFromDiagLine(line)`: reads a `client-diag.jsonl` line, taking a capture's `ta` arrival stamp when present because the kernel's `t` on those rows is whole seconds.
- No DOM, no wiring into the pane; nothing in `render.ts` changes. Test rows are built with the pane's own row builders so a shape change fails here first. All values synthetic.

## Not in this PR

A live hookup (a kernel-side check that flags a recurrence as it happens) is a natural next step; this PR keeps to the reader and its tests.

### DIFF
--- a/ui/webview/scroll-journal-audit.test.ts
+++ b/ui/webview/scroll-journal-audit.test.ts
@@ -1,0 +1,171 @@
+// The scroll journal's reader recognizes the snap (T262k, the user 2026-09-08): a single scroll event, after the reader
+// had stopped, landing on one fixed value per tab tens of pixels above the bottom, with no write, no tail change and
+// scrollHeight unchanged — and does NOT cry wolf at the reader's own bursts, a write's echo, a tail shrink's clamp or a
+// spacer re-size's anchoring. The rows are built with the pane's own row builders (scroll-write.ts), so a change to a
+// row's shape breaks this file before it breaks the reading. Synthetic values throughout.
+import { test } from "node:test";
+import * as assert from "node:assert/strict";
+import { unwrittenMoves, repeatedLandings, journalRowFromDiagLine, type JournalRow } from "./scroll-journal-audit";
+import { scrollWriteRow, tailChangeRow, spacerRow } from "./scroll-write";
+
+const SID = "11111111-2222-4333-8444-000000000301";
+const SID2 = "devbox:11111111-2222-4333-8444-000000000302";
+const SH = 9000, CH = 900, BOTTOM = SH - CH;           // 8100
+
+const gesture = (t: number, top: number, sh = SH, ch = CH, sid = SID): JournalRow => ({ t, what: "scrollgesture", data: { sid, top, gesture: true, sh, ch } });
+const write = (t: number, writer: string, before: number, after: number, stick = false, sh = SH, ch = CH, sid = SID): JournalRow =>
+  ({ t, what: "scrollwrite", data: scrollWriteRow(sid, writer, before, after, stick, sh, ch) });
+const tail = (t: number, dh: number, last: string, stick: boolean, sh = SH, ch = CH, sid = SID): JournalRow =>
+  ({ t, what: "tailchange", data: tailChangeRow(sid, dh, last, stick, sh, ch) });
+const spacer = (t: number, topBefore: number, topAfter: number, botBefore: number, botAfter: number, sh = SH, ch = CH, sid = SID): JournalRow =>
+  ({ t, what: "spacer", data: spacerRow(sid, topBefore, topAfter, botBefore, botAfter, sh, ch) });
+
+/** A reader's trackpad burst: samples ~16 ms apart, accelerating then easing, from `from` to `to`. */
+function burst(t0: number, from: number, to: number, n = 12): JournalRow[] {
+  const rows: JournalRow[] = [];
+  for (let i = 1; i <= n; i++) {
+    const k = i / n, ease = k < 0.5 ? 2 * k * k : 1 - Math.pow(-2 * k + 2, 2) / 2;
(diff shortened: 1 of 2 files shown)


ksarma:
Recommend merging; nothing to fold. The author merged before this review landed, so everything below is a follow-up list.

Reviewed inline: `unwrittenMoves` groups rows per sid and takes the last recorded position from a write's `after` or a gesture's `top`. It reports a gesture as a move when it is at least `minPx` away, at least `idleMs` later, and no further gesture follows within `settleMs`. The second commit's guard skips a gesture read while `scrollHeight` is at or below `clientHeight`. The skipped read does not seed the next comparison. Its test fails against the first commit alone (11 of 12) and passes at head. `explain` matches a tailchange `dh` or a spacer `dTop` within `tolPx` of the delta. It was probed at exact equality, at 2 px, and just outside. The rows consumed carry sid, pixel geometry, a writer name, and a class list clipped to 60 chars, with no transcript text. The fixtures are synthetic placeholder ids. Nothing but its test imports the module, and it is not an esbuild entry point, so no dist bundle contains it. The feature label fits. The red vscode-extension check at e2dfc009 is ten tab-strip-skip failures. They fail identically at the PR's base 4c58c393 and pass at current main after #1117, so they are not this PR's.

Follow-ups, each about how faithfully the reader matches the fingerprint it documents:

- `scroll-journal-audit.ts:131` and `auditOne`: capped rows (`scrollgesture-capped`, `scrollwrite-capped`) are dropped. Under the default 40 per minute cap, the first gesture after a capped minute is measured against a stale position and reported as an unexplained move. A capped row should clear `pos` and `since` so the next write or gesture re-seeds. Follow-up.
- `scroll-journal-audit.ts:83`, `:87`: a scrollwrite within `settleMs` does not break isolation. The move carries `sh` at the read but not at the previous position. As a result, the scrollHeight-unchanged leg of the header is neither checked nor derivable. A re-show clamp against a stale non-empty DOM reads as a whole-transcript unexplained move. Record `sh` with `pos` and emit `dSh`. Follow-up.
- `scroll-journal-audit.ts:118`: `repeatedLandings` keys on absolute `to`, but the signature is a fixed distance above the bottom. Three snaps at `distAfter` 64 over a growing scrollHeight return no group. Key on `distAfter` with a tolerance, or return both groupings. Follow-up.
- `scroll-journal-audit.ts:91`: `since` resets at every gesture. A tailchange or spacer row filed after the clamp's scroll event therefore never explains it. That order was reproduced in Firefox with a forced layout, not in Chromium. Attribution by interval would fix it, and reverse-order fixtures would pin it. Follow-up.
- `scroll-journal-audit.ts:132`: without a `ta` stamp, the parser returns the kernel's whole-second `t` as a millisecond clock. Nothing in the repo writes `ta`. Whole-second rows report spurious moves across second boundaries and miss same-second snaps. Refuse or flag a coarse journal; `test.ts:179` pins the silent path today. Follow-up.
- `scroll-journal-audit.ts:131`: main now files a `tailmut` row (#1116), and the reader drops it. Whether it should explain a move is a design call. But the row names the cause of a within-frame clamp, so it should surface as a hint rather than vanish. Follow-up.
- `scroll-journal-audit.test.ts:15`: the gesture row is hand-built, because `render.ts` inlines the literal and `scroll-write.ts` has no builder. The row-builder claim therefore holds for three kinds and not for the one every move depends on. `scroll-write.test.ts:58` pins the `render.ts` literal, but nothing ties this reader to it. A shared `gestureRow` builder would close the gap. Follow-up.
- Nit, `scroll-journal-audit.ts:45`: `idleMs` 400 excludes snaps 100 to 400 ms after a land. Those are inside the 0.1 s lower bound the header gives. The right-after-a-land test needs `idleMs` 80 to see its case.

Tests: the new file passes 12 of 12 at e2dfc009. The full vscode-extension suite at head is 3591 of 3601; the ten tab-strip failures reproduce at the base 4c58c393 and are absent at current main. With both PR files on current main the suite is 3607 of 3607. Typecheck and build pass. CI: all nine checks pass at the merged head 382b81c3, where both audit files are byte-identical to e2dfc009.

Review run by the maintainer's automated review session; verified against head e2dfc009.

PR #1119, opened 2026-09-08
Unread comment replies get the needs-you ring, two direction chips and the strip's dots

yatharth:
Tier: feature

## The problem

A reply lands on a comment you scrolled away from, and you forget to come back. The mark's cue and the rail tick only say so where you happen to be looking; nothing tells you a reply is waiting two screens up.

## What changes

**The comment mark.** An unread landed reply now wears the tab strip's needs-you ring — the same dashed `--st-awaiting-bg` outline a tab wears while its session waits on you, scaled to a text run (1.5px, offset outward so it never crosses the glyphs). It replaces the 2026-08-23 yellow corner dot. Read marks wear no ring. The FILL is untouched and still says pending vs landed exactly as before: the green wash is a reply still being written, the 45% yellow tier is a landed one (T237's pins on those lines are byte-intact).

**Two direction chips beside the go-to-bottom chip.** Stacked above it in its own dress and slot (body-level, fixed, bottom-left, the menu-card tokens, 22px pill, dim at rest, accent on hover): "↑ 2 replies unread" / "↓ 1 reply unread", counting the unread landed replies whose marks sit wholly above / below the viewport. A mark on screen counts in neither direction. Each chip exists only while its count is > 0; the tip reads "N replies unread above · nearest: <first line of what you last said in that thread>". Click lands the NEAREST one in that direction through the chat's own scroll-to-uuid route (scrollToAnchor → landOn's one flash) and pulses the mark itself. It does NOT mark the thread read — opening the thread does, as today.

**The scroll strip's dots.** The strip already carried a dot per thread as the comment rail's ticks: same frame as the notches, `--cmt-hl`, the unread one wider and double-ringed, click = jump + open the thread. Pinned rather than duplicated, so one column never carries two marks for one fact.

## Unread semantics: unchanged

One predicate everywhere (`isReplyReady`): the kernel's `unread` bit on an OPEN thread — "a FINISHED reply you have not seen" (T237) — cleared only by opening the thread (`commentSeen`). No kernel change.

## Visibility rule, as implemented

Counts = the kernel's unread set ∩ the active session's threads; direction from each mark's box against the scroller's viewport (a windowed-out anchor takes its direction from event order against the rendered window, in a distance tier past every rendered mark). Recomputed on the events the chat already listens for — the jump chip's own update (passive scroll listener, resize, the content ResizeObserver, tab switch, append), the comments frame and every transcript rebuild (applyCommentMarks' tail) — plus the pane's own size, so a pane measuring 0 drops the chips. No timers, no polling; a signature skips unchanged paints. Hidden in the read-only subagent viewer.

## Verification

- `npm run typecheck`, `npm test` (3610 pass, +28 new: `reply-ready.test.ts`, `comment-mark.test.ts`, the reply-dot block in `scroll-marks.test.ts`; the dot pin in `comments.test.ts` rewritten deliberately with its why), `npm run build`.
- Python (unittest, the suites that read the UI sources plus `test_kernel` and `test_comment_threads`): 650 OK. `kernel/` untouched.
- Screenshots, both themes, from the new synthetic fixture `tools/ui-verify/fixtures/reply-ready-chat.html` (notes-api, placeholder ids): a transcript with one read mark (plain yellow, no ring), an unread mark on screen (yellow + the dashed red ring), one unread scrolled above and one below — so both chips stack over the go-to-bottom chip at bottom-left ("↑ 1 reply unread" / "↓ 1 reply unread"), and the strip shows the blue notches with the yellow ticks, the unread ones double-ringed. Dark and light read the same; the light theme's cream card and clay accents resolve through the same tokens.

## Tier

This account cannot attach labels on the upstream; a maintainer should add **`feature`** — a self-contained capability inside romp's existing comment-thread model (no contract change, no kernel change).


henrypinkard (approved):
Approved on head 9aa1b5d8 on the maintainer's word (feature-review session).

yatharth:
Follow-up, after a live look on 2026-09-10: the chip's deliberate no-commentSeen left the user one click short of the reply, and on a passage commented on twice that click landed on nested marks and opened the wrong thread, so the outer thread's ring never cleared and this chip stayed lit. #1249 makes the ring you click open the thread that owns it (and lifts an unread rail tick above a read one at the same top); the companion PR opens the thread from the chip itself, which is what marks it read.

PR #1120, opened 2026-09-08
Held queued copies: judged against the previous push, released by their own landing, never a phantom after a cancel; the cached pending node is wired once

henrypinkard:
Follow-up to #1111 (the tail-height fix), from its review's reproductions: a held queued copy is now judged against the **previous** push's anchor (a copy that vanishes and lands in one push is released, not held forever); a landing that releases a held copy is not a "later landing" for its siblings; a copy the user cancelled with the ✕, or one that vanished on a settled session, is never held (no phantom "landing" card); copies count per text; the held pass is idempotent; our cached pending node is wired once instead of accumulating hover chrome on every push; a held-only group's header says "landing…".

**Tests:** `ui/webview/queued-held.test.ts` (red on the merged rules for the same-push landing, the sibling drop, the cancel, the settled session and the per-text count), the flipped pins, and both served harnesses green on this tree.

### DIFF
--- a/tests/test_queued_copy_held.py
+++ b/tests/test_queued_copy_held.py
@@ -65,6 +65,12 @@ const page = await browser.newPage({ viewport: { width: 1000, height: 600 } });
 await page.addInitScript(() => {
   window.__rows = []; window.__frames = [];
   window.addEventListener("message", (e) => { const m = e.data; if (m && (m.type === "session" || m.type === "update" || m.type === "chatTail")) window.__frames.push(m); });
+  // while the synthetic sequence runs, the kernel's own frames for this session are held back (a status-only tail
+  // would truncate the injected events): the shim's socket handler is wrapped at the prototype
+  window.__quiet = false;
+  const desc = Object.getOwnPropertyDescriptor(WebSocket.prototype, "onmessage");
+  Object.defineProperty(WebSocket.prototype, "onmessage", { configurable: true, get() { return desc.get.call(this); },
+    set(fn) { desc.set.call(this, (ev) => { if (window.__quiet) { try { const m = JSON.parse(ev.data); if (m && (m.type === "chatTail" || m.type === "update" || m.type === "session" || m.type === "status")) return; } catch (e) {} } return fn.call(this, ev); }); } });
   const orig = WebSocket.prototype.send;
   WebSocket.prototype.send = function (d) {
     try { const m = JSON.parse(d); if (m && m.type === "clientDiag" && m.surface === "chat" && /^scroll/.test(m.what || "")) window.__rows.push({ what: m.what, writer: m.data && m.data.writer, before: m.data && m.data.before, after: m.data && m.data.after, sh: m.data && m.data.sh }); } catch (e) {}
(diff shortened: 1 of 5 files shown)


ksarma:
No objection to the merge. This review lands after it, so everything below is a follow-up. The reviewed head 6936cba2 was superseded before merge by the rebase c780aac5 (merged as ded108dc). The five substantive files carry identical changed lines, so the review applies to what landed.

Reviewed inline: reconcileHeld's per-session memory now judges a vanished copy against the previous push's last kernel uuid. It releases a copy whose record lands in the same push. It excludes a releasing landing from the later-landing drop for its siblings. It counts copies per qid or per trimmed text. It skips uuid-less markers and queued groups when choosing the anchor. Each rule is confirmed by reading and by the new node tests. The identity path stays consistent with #1044, #1067 and #1076. An identified copy is released only by the atom carrying its qid; an id-less copy is held by text for one push. The render.ts pass strips its own held group and landing marks before recomputing, so a full-frame resync or a reconnect replay does not double the group. The wired-once guard stops the accumulation: across eight pushes the cached pending node keeps two hover listeners, where the base grew to 32. Two inputs to the hold still move a card without new information, and the guard that stops the accumulation also loses a child node; all are listed below. The tab-strip hunks were test-only and correct against the base but duplicated #1117 on main. They were the whole DIRTY conflict (`ui/webview/tab-strip-skip.test.ts`, `ui/webview/tab-strip-skip-exec.test.ts`), and the rebase dropped them. Privacy and vocabulary scans of the added lines are clean.

Follow-ups, none blocking:
- `ui/webview/queued-held.ts:110` with `ui/webview/render.ts:476` (follow-up): settled is derived from the chip state. The kernel computes that state from the transcript's open turn, and it still reads ready on the push the feeder pokes right after popping a copy. Reproduced with a real SdkBackend: a copy listed on an idle session vanishes with status ready and no record yet. The hold is refused, and the card disappears for a push. Key it on the kernel's busy bit or its dropped verdict.
- `ui/webview/render.ts:14027` (follow-up): chatTail runs reconcileHeldCopies before applying msg.status at 14037, while update() applies status first. The ordinary delta path therefore judges settled from the previous frame. Executed: a settled-to-working flip that drops a copy is held via update and dropped via chatTail. Hoist the status write above the pass and pin it in `chat-exact-tail-exec.test.ts`.
- `ui/webview/render.ts:14528` (follow-up): a cancelResult ok:false leaves the cancelledQueued note recorded at 15821 in place. The next frame then refuses a hold for the copy the kernel just said will land. Drop the note for (id, md) on ok:false.
- `ui/webview/render.ts:1907` with 4005 (follow-up): the wired flag survives renderPendingGroup's replaceChildren, but the .rail-hit strip was a child. After the first signature change (a second send, a cancel, a minute tick of a held countdown) the strip and its click are gone for the group's life. Turn-level hover survives.
- `ui/webview/render.ts:464-489`, 1907, 3999 (follow-up): the render.ts glue has no test. Removing the wired-once guard, the noteCancelledQueued call, the forget-once-unlisted block, the idempotence pre-pass or the minute term one at a time leaves the node suite at 3591/3591. `tests/test_queued_copy_held.py` skips in CI.

Tests: at origin/main the PR's queued-held.test.ts fails to build (no export lastKernelUuid). Its bodies, run through a signature shim against the base reconcileHeld, fail 4 of 6 on the same-push landing, sibling landing, cancelled/settled and per-text rules. The queued-indicator landing-label pin is red at the base. At head: queued-held 6/6, queued-indicator 24/24, node suite 3591 pass 0 fail, typecheck clean, `tests/test_queued_copy_held.py` 1 passed with a real browser. CI green at 6936cba2.

Review run by the maintainer's automated review session; verified against head 6936cba2.

PR #1121, opened 2026-09-08
T269: every deploy restart bounces at once; romp refresh --quiet is the only quiet-window door

henrypinkard:
## What
Every deploy-triggered kernel restart bounces immediately. The quiet-window deferral survives only as the explicit `romp refresh --quiet`.

The user's decision (2026-09-08, ~2:05 PM PT): the parked quiet window held the box unusable for the full 15-minute backstop on 26 of the morning's 32 restarts, while boot reconcile resumes the cut turns with their history either way, so an immediate bounce costs seconds.

## Which cut, and why
**The caller.** Of the three deploy callers, two were already immediate (the release self-update since T160, and the automatic converge through `_run_main_update`'s default). Only the p2p apply script (a peer's `romp update`) still asked the far manager for `restart-all --quiet`. Changing that one caller leaves the manager's `?when=quiet` route, its 15-minute backstop and `bin/romp refresh --quiet` byte-for-byte as they were, so the explicit door keeps its tests without a special case in the kernel's gate. A gate-side cut (the manager ignoring `when=quiet` for deploy reasons) would have needed the manager to learn the reasons and would have left the audit rows claiming a window nobody honored.

In `kernel/kernel.py`:
- the apply script asks `romp-manager restart-all` (immediate) when the manager owns the polled kernel, writes its `p2p-update` audit row WITHOUT `when=quiet`, and reports `SYNCED:<sha>:MANAGED`; the fallback branch (no owning manager) is unchanged;
- the dialing side reads `MANAGED` as "synced to <sha> + restarting now (through its manager)" and arms `restartExpected` with `quiet: False`. That field is recorded, not read: the tunnel's reinterpretation of the far kernel's gap keys on the sha and the stamp only. The short window is the far kernel's own: its cut attribution (`_recent_restart_audit`) gives a row without `when=quiet` the 90-second window instead of the 20-minute quiet one, so the cut row names this request while it is fresh.

Interplay with T240d: the drift check stood down for a `p2p-update` row carrying `when=quiet`, because the parked restart could take up to 15 minutes and a drift pass inside that window posted a second, immediate restart. A new apply writes an immediate row, so the stand-down does not engage, and it does not need to: the drift check runs once per 300 seconds, and the gap between the far checkout's reset and the manager's bounce is the ownership check plus one POST, about a second. The row still lands before the ownership check exactly as before, so a cut inside that second finds its request on disk. Older peers, a quiet converge and `romp refresh --quiet` still write quiet rows, and the stand-down still reads them; the docstrings and comments now say so. Widening the stand-down to immediate rows was considered and left out: it would also have parked the drift check for 90 seconds behind the no-owning-manager fallback's row, and an existing T240d pin says an immediate request parks nothing.

## Tests
- `tests/test_kernel_remote_update.py`: the three managed-path tests move to the immediate spellings (`restart-all` without `--quiet`, the row has no quiet marker, `MANAGED` tag, `quiet is False`, detail says restarting now, no "quiet" wording; ordering pins kept). New `test_every_deploy_restart_is_immediate_and_only_refresh_quiet_defers` pins the self-update script text, the converge's default, the absence of `restart-all --quiet` anywhere in the kernel, and that `bin/romp refresh` forwards `--quiet` to the manager's `when=quiet` mapping.
- Kept green, unchanged: `tests/romp-refresh-audit.bats` (refresh `--quiet` reaches `restart-all --quiet`), `tests/manager-restart.test.js` (the explicit opt-in), `tests/romp-manager-ensure.bats` (the `?when=quiet` route), `tests/test_main_drift_notice.py` (converge default immediate; the T240d apply-order pin still holds), `tests/test_kernel_update.py` (self-update has no `when=quiet`), `tests/test_deploy_drain.py`.

## Docs
`docs/reference.md`: the `romp update` row says the remotes restart at once, and the `romp refresh --quiet` row names itself the only door to the quiet window.

## Evidence
Targeted modules green in the worktree: 208 tests across the remote-update, drift-notice, deploy-drain, kernel-update, restart-cuts and tunnel-truth modules; 80 manager node tests; 9 bats cases across the manager-ensure and refresh-audit files. Full pytest in the worktree on the first head: 9813 passed, 56 skipped, 0 failed. CI green on that head. The adversarial review (three lenses, two skeptics per finding) found only comment and docstring drift plus the inaccurate expectation comment; all folded in the second commit.

### DIFF
--- a/bin/romp-manager
+++ b/bin/romp-manager
@@ -845,7 +845,7 @@ function startManager() {
       // SIGTERM respawns blinked every dashboard for an hour, and this handler — the one door every
       // programmatic restart walks through — recorded nothing. The immediate path used to be silent.
       log(`restart-all requested (${url.searchParams.get('when') === 'quiet' ? 'deferred to quiet window' : 'IMMEDIATE'}) from ${req.socket.remoteAddress || '?'}`);
-      if (url.searchParams.get('when') === 'quiet') {            // deploy path: land at the next quiet window
+      if (url.searchParams.get('when') === 'quiet') {            // `romp refresh --quiet`, the one quiet door (T269): land at the next quiet window
         return json(202, { ok: true, deferred: true, coalesced: queueQuietRestart('all') });
       }
       clearPendingQuiet();                                       // this bounce delivers any pending deploy too
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -21,7 +21,7 @@ update` starts a session called "update".
 | `romp resume` | Resume a past conversation, chosen from a full-screen picker |
 | `romp status` | Manager and kernel status |
 | `romp refresh` | Restart the postal bus and every kernel immediately, picking up new code (cut turns resume with their history) |
-| `romp update [host…]` | Push this machine's committed Romp to attached remotes and restart them |
+| `romp update [host…]` | Push this machine's committed Romp to attached remotes and restart them at once (every deploy restart is immediate; boot reconcile resumes the cut turns with their history) |
 | `romp up` | Run the kernel manager in the foreground; rare, since the login service runs it |
 | `romp version` | Version report across the moving parts |
 | `romp keyswap [<name>] [--cycle <session,…>\|--cycle-all]` | Switch the API key source without restarting the manager: selects a key command, a 1Password reference or a legacy key from `service.env.<name>`, and `--cycle` reconnects running sessions onto it. Bare, it reports the configured source and candidates without fetching secrets. See [Switching which API key the sessions bill](#switching-which-api-key-the-sessions-bill-romp-keyswap) |
@@ -51,7 +51,7 @@ These are for scripting and for agents rather than daily use:
 | `romp default-dir [PATH]` | The default working directory for new sessions; no argument prints it, `""` clears it |
 | `romp debug [on\|off\|status]` | Judge debug mode, where rejection rows carry the full input and reply |
 | `romp resume <id> [--name <n>] [--detach]` | Resume one exact conversation by UUID |
-| `romp refresh --quiet` | Refresh at the next quiet window instead — waits for sessions to finish their turns (15-min backstop) |
+| `romp refresh --quiet` | Refresh at the next quiet window instead — waits for sessions to finish their turns (15-min backstop). The ONLY door to the quiet window: a deploy (a peer's `romp update`, a release self-update, an automatic converge) restarts immediately, by the user's 2026-09-08 decision |
 
 Raw `POST` callers, anything that talks to the kernel's routes directly rather
 than through `romp`, follow one body contract, and the postal bus's own routes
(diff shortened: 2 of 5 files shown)


ksarma:
Recommendation: accept the merge as shipped, with one follow-up. The PR landed before the review finished (merged at d27a566d).

Reviewed inline: the p2p apply script now asks the far manager for an immediate restart-all when the manager owns the polled kernel. It writes its p2p-update audit row without `when=quiet` and reports `SYNCED:<sha>:MANAGED`. The dialing side reads `MANAGED` as restarting now and records `quiet False` on the expectation. The manager's `?when=quiet` route, `bin/romp refresh --quiet`, the bats and node suites, and the three untouched pytest files are identical to the base at d07207f9. The merged head adds one comment line to the manager route. The T240d stand-down still keys on `when` equal to `quiet`, so a new p2p row no longer parks the far kernel's converge. Older peers, a quiet converge, and `refresh --quiet` still write quiet rows and still park. The first commit's body and kernel comment credited the 90-second window to `restartExpected quiet False`, which nothing reads. b89533c6 corrected both: the short window is the far kernel's row-based cut attribution in `_recent_restart_audit`. That point is closed. The race the stand-down used to cover remains open, listed below. Product note: this reverses the quiet default for every peer's `romp update`, so each deploy cuts every in-flight turn on the far host at once. Boot reconcile resumes them, and the #1097 reload waits for a gesture-free moment, so the thread is kept. A host running many sessions is still interrupted on every deploy, with no quiet opt-in on the pushing side.

Findings:

- `kernel/kernel.py:19764`, follow-up. The immediate row leaves the drift check unguarded between `reset --hard` and the kernel's exit. That window is manager status plus the POST, then the 2 s drain, with no `_TERMINATING` check. A pass in that window on an auto-mode host with the 25-minute cool-down open posts its own restart-all. The manager coalesces it into a trailing restart that SIGTERMs the fresh kernel about a second into boot. Reproduced with a real manager: three spawns for one deploy. When origin/main is ahead of the pushed sha, the checkout can move to origin/main before the respawn. The race is narrow: one 300 s cadence against a window of one to three seconds. It is time-based, though, where the removed cover was event-keyed. Either change closes it: stand down on any unconsumed p2p-update row for the checkout sha inside the 90 s window, or yield when `_TERMINATING` is set. The pin at `tests/test_main_drift_notice.py:351` moves with it.
- `docs/reference.md:54`, nit. The `refresh --quiet` row carries a dated user-decision attribution and the word ONLY in caps, in a user-facing reference. The plain form (the only way to defer a restart; deploys restart at once) matches the neighbouring rows. The update row's term "boot reconcile" is defined nowhere in docs.
- `kernel/kernel.py:17553` and `19700`, nit. The `RESTART_EXPECT_MAX_S` rationale and the pending-restart comment still describe a quiet window the dialing side asked for. No dialing path arms one now.

Locally at d07207f9: `tests/test_kernel_remote_update.py` plus `tests/test_main_drift_notice.py`, 111 passed. With `test_kernel_update.py` and `test_deploy_drain.py`, 116 passed. `node tests/manager-restart.test.js`, 22 of 22. Informational red-first: the PR's test file at origin/main fails on exactly the two moved tests and the new pin (3 failed, 59 passed) and passes at head (62 passed). The merged head differs from d07207f9 only in comments, docstrings, and one assertion message. All nine CI checks are green at d27a566d.

Review run by the maintainer's automated review session; verified against head d07207f9.

PR #1126, opened 2026-09-08
Chat split screen: the dashboard holds N chat columns, each a full chat on its own session

amith-ananthram:
## What

The web dashboard's chat can be split into up to four columns, so several sessions sit side by side instead of behind each other's tabs. Every column is a full chat: its own tab strip, composer, active session, drafts and scroll place. Open one from a tab's right-click menu (**Open in new split**), or from the palette (**Split the chat**, `Mod+\`); close with the `×` in the column's corner or **Close this chat split**. The set of open columns and each column's width persist per browser. Phones never split.

## Design points

- **Shell-level columns, not a change to the chat pane's model.** Column 1 stays the served `#chat-pane` / `#f-chat`. Every later column is a client-made twin (`_LANDING_SPLIT_JS`) around an iframe at `/chat?col=N`, inserted before the outline gutter behind a `chat|chat` gutter of its own. `render.ts` keeps its single active tab; N columns are N chat clients, which the kernel already serves (each column's watched tab is built first, as today).
- **Per-column state without moving anyone's data.** The pane shim reads `?col=` and suffixes its persisted-state key (`romp-vscode-state-chat:N`) and its connect (`&col=`), so each column has its own active tab, drafts and scroll; column 1 keeps the unsuffixed key it always had. Browser-wide preferences (tags, view order, settings, keys) stay shared on purpose.
- **Focus arbitration is client-side and event-exact.** The kernel aims a focus, a revive prompt or a feed click at the dashboard, so every column receives it. The shell answers which column it belongs to (`__rompChatTarget`: the column already showing the session, else the one the user last worked in, else the first) and the others stand down (`focusIsOurs`). The shell's own hand-over to a new column, and the kernel's parked push-tap reveal (consumed by exactly one client), ride `own` and bypass the arbitration.
- **Existing shell services learn about later columns instead of assuming one.** The picker lift marks the asking frame (`.lifted`) rather than `#f-chat`; Alt+Arrow walks every column; the Log tracks each column's socket apart; the outline and feed gutters take the rightmost chat column as their neighbour; the palette's chat-directed commands land in the column last worked in; `activeSession()` reads it too.
- **A refused split says why** (the cap, the phone), and a split of a hidden chat group brings the group forward first.
- Found on the way, older than this change: the gutter grab normalised widths one pane at a time, so a read after a write came back at a mixed scale and a fresh browser's first drag ballooned the first column. The grab now reads every width before it writes.

## Tests

- `tests/test_chat_split.py`: source pins of the shell, shim and kernel, plus the split script executed in node against a DOM stub (open, order, persistence, arbitration, cap and its refusal, close, restore, phone).
- `tests/test_chat_split_served.py`: the served leg, a hermetic kernel with two synthetic sessions driven in headless Chromium (skips loudly without a browser, like the other served tests).
- `tests/test_pane_gutters.py`: the gutter script executed in node (fair grow never NaN, stored widths kept, registered columns, neighbours).
- `ui/webview/chat-split.test.ts`: the pane and palette halves at source.
- Pins that moved (lift selectors, gutter neighbour, visible columns, script count, icon kinds, the addressed reveal) are updated with their reasons. `docs/guide.md` gains a paragraph.

Not covered: the VS Code extension (a single chat panel there is unchanged).

Creating a split:

Screenshot 2026-09-09 at 12 42 22 PM

What a split looks like:

Screenshot 2026-09-08 at 3 50 49 PM

Closing a split:

Screenshot 2026-09-09 at 12 42 37 PM

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -231,6 +231,18 @@ scrolling for the rest; the cap lifts while a row's details are open. Where the 
 it also shrinks the tabs and group headers; on a phone the session picker stands in for the strip, so
 there the setting tightens the panel alone. Like the other chat settings, it is per browser.
 
+**Several sessions at once.** The chat can be split into columns, so two or three sessions
+sit side by side instead of behind each other's tabs. Right-click a tab and pick **Open in
+new split**, and a new chat column opens to the right on that session; **⌘** / **Ctrl**
+with the backslash key, or **Split the chat** in the command palette, opens an empty one. Each column is a full chat, with its own tab strip,
+its own composer and its own place in each transcript; drag the gutter between two columns
+to resize them. A column remembers which session it was on, and its width, across reloads,
+and the set of open columns is remembered per browser. The **×** in a column's top-right
+corner closes it; **Close this chat split** in the palette closes the column you are in, or
+the last one when you are in the first. When a card in the feed or a notification jumps to
+a session, the jump lands in the column already showing it, else in the column you last
+worked in. Four columns at most; the phone shows one pane at a time and never splits.
+
 ### The feed
 
 The feed is Romp's task-management layer: a card for each task. Romp's
--- a/docs/read-side.md
+++ b/docs/read-side.md
@@ -241,7 +241,11 @@ at its send position, above those steps, while the model read it only after them
 so the order on screen contradicted the order the model saw; the read position is
 the one that matches. So the chat's pending bubble sits at the TAIL while pending,
 below every streaming step, and the landed atom appears in that same tail
-position, so nothing moves on landing; the send time rides along as `sentAt` for
+position, so nothing moves on landing. Every other window of the session sees the
+kernel's echo of that send in the same place: the live merge orders an in-flight
+echo after everything the turn holds (a never-delivered one keeps its time), and
+the pane dresses it as the sender's bubble is dressed, so one session in two split
+columns agrees on what is pending (2026-09-11). The send time rides along as `sentAt` for
 the bubble's hover ("sent at HH:MM", shown once landed when it differs from the
 landing by more than a minute). No header, no cue. This supersedes the
 in-place-at-send-position rule of T252/T252b; the kernel's per-copy identities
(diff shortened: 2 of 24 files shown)


amith-ananthram:
Intended tier: **feature** (a self-contained capability inside the existing model; design points in the description). This account has read access only, so it cannot apply the label or arm auto-merge — a maintainer will need to add the `feature` label for the tier checks to run.

amith-ananthram:
@henrypinkard no rush on reviewing but don't seem to have permissions to add labels / reviewers etc so just drawing it to your attention!

amith-ananthram:
Rebased onto main 45b2eb97 (after #1372 and #1388 landed). Still one squashed commit.

- One conflict, in `ui/webview/palette-main.ts`: main's settings iframe (`f-settings`) joins the frames the chords are wired on; the branch's `wireKeys` wires it with the rest.
- One integration fix folded in: main's gear-close focus return (2726bc87) hard-codes the first chat frame; with a split open it pulled the keyboard into column 1 and re-aimed later shell relays there. It now resolves the focused column the way the branch's other shell relays do, falling back to the first. Pinned in `tests/test_chat_split.py`.
- Two pins adjusted for main's newer state: the branch's own palette pin (the six-id list) and `test_pane_state_broadcast`'s `<script>` count (20 + the split's script).

amith-ananthram:
Second commit, 55729cdc: with one session open in two columns, a send from one column showed in the other as a solid bubble at its send time, above the tool steps that streamed after it, while the sender's column kept its dashed "sending…" bubble at the tail. The live merge sorted the turn's atoms by time; an in-flight echo now sorts after everything the turn holds (T252d's rule, applied to every other window), and the pane dresses that echo as the sender's pending bubble. A never-delivered echo keeps its time and its own dress. Not split-specific (another browser sees the same), but the split shows the two views side by side. Pinned in `tests/test_kernel_fed_echo_absorbed.py` and `queued-indicator.test.ts`.

henrypinkard (approved):
Approved on the maintainer's word: rebased onto current main by the maintainer's session, typecheck and the touched webview, Python and served tests green on this head.

PR #1127, opened 2026-09-08
A notification tap on the phone lands after a deploy too: worker self-refresh on push, legacy message accepted, every /reveal logged

yatharth:
Tier: fix

## Why

A tap on a romp push notification on the phone still opened the wrong session after #937/#940 shipped. The running kernel carried the fix (kernel sha contains cf0a55fd), the served `/sw.js` and shell carried it, and nothing anywhere recorded whether the tap had reached the kernel: the `/reveal` route, the shell's landing script and the worker's `notificationclick` log nothing.

Two gaps behind that:

1. **A stale worker on the phone.** A Home Screen app left in the background checks for a new service worker only on a navigation (relaunch or reload), so the first tap after any deploy runs the worker the phone installed last. The worker of builds before 2026-09-06 posts `{romp:'pushReveal'}` into `ws[0]`, a shape the current shell's `_LANDING_REVEAL_JS` did not read, so that tap was a silent no-op on a fresh page.
2. **No evidence.** A "the tap did nothing" report could not be split into never-reached-the-kernel / parked-and-never-consumed / consumed-but-not-shown.

## What

- `_SW_JS` push handler: `self.registration.update()` once `showNotification` has resolved (after the show, never beside it), failures swallowed; `shown` stays in the `waitUntil` list so a failed show still fails the push as before.
- `_LANDING_REVEAL_JS`: reads `pushReveal` beside `notificationClick`; stamps `via: 'sw' | 'link'` on every `/reveal`.
- `_SW_JS` click handler: the window `clients.openWindow()` returns is also handed the routing block (the browser holds a message to a page without a listener until the shell adds one), so a cold start lands even where a browser opens the app on its start URL instead of the link (reported of installed iOS apps; unverified here). With the link arriving too, the kernel sees the same focus asked twice for one window, which is idempotent.
- `/reveal` route + `_reveal_request` / `_consume_pending_reveal` / `_reveal_proven`: one stderr line per tap with the road and the outcome (delivered / parked / copy parked), and the park's end (consumed by the pane's ready, retired by the target's pong). Ids clipped to 8 chars.
- `docs/guide.md`: one paragraph on the phone-side handler and the one-time remedy (close the app from the app switcher and reopen).

## Tests

`tests/test_kernel_webpush.py`, all red before this change: `ServiceWorkerExecutes` (update requested after the show and not before; opened window handed the block; null client tolerated), `LandingRevealExecutes` (legacy shape lands; `via` on every fetch), `RevealRoute` (the log lines), `LandingRevealPins`. Full Python suite: 9816 passed, 56 skipped. UI node tests: 3595 passed.

## Tier

Tier label to be set by a maintainer (`fix` is the intended tier: a bug fix with tests that fail before it). Filed without auto-merge on purpose.


yatharth:
**1e0fdc7b — instruments at every hop of a notification tap, the sid-less test says so, and a missed tap is replayed.**

Live evidence (phone, 2026-09-08 22:50 UTC): after a force-quit + reopen, test notification sent, session and app switched, tap → the app came forward, the session did not change, and the journal held **no** `[reveal]` line at all — so no `/reveal` ever left the phone. Nothing recorded where between the tap and the fetch it stopped. This push adds the recording, fixes the one reproducible gap in hypothesis A, and hardens the cold-start race (B).

**Instruments** — client-diag rows, surface `shell`, through one poster (`window.__rompShellDiag`, defined in the mobile script; queued, capped, until the shell socket opens). The shell socket now carries the shell's `wid`, so its rows match the dashboard's pane rows (and `_reveal_chat_for`'s shell line, aimed by wid, has a target at last). Structure only — never a sid, a name or a text:
- `push-test {sidAttached, host, why, tabs}` at the press (`why`: `no-frame` | `no-doc` | `no-tabs` | `none-active` | `threw`, `''` when attached)
- `deeplink {hasSid, hasCard, controlled}` at every boot
- `sw-message {shape, hasSid, kind, dup, sw:{clients, tops, road, vis}}` — the worker cannot post to the file, so its own block rides the tap message: window clients it saw, top-level among them, road taken (`focus` | `open` | `open-after-refused`), the target's `visibilityState` after `focus()`
- `reveal-post {status, via, boot}` after `/reveal` answers (status 0 = no answer)
- kernel: one `[push] test sid=<8>|none endpoint=<host>: <status>` stderr line per test push

**Hypothesis A on the mobile DOM.** The bell's `activeSession()` and the mobile chip read the same `#tabs .tab.active[data-id]` nodes, and the chat iframe is eager (`src=/chat`) on the phone, so the read comes back empty only in the **boot window**: the strip holds the kernel's placeholder tabs before the first session payload names an active one — the chip then shows the *first* tab (its own fallback) while the test attaches none. Kept that way (no guessing), but never silent: the row names the reason and the result line now reads "The push service accepted it. No session was attached — the tap will only bring romp forward." A node harness presses the button on each DOM shape.

**Hypothesis B.** The worker keeps the last session-addressed tap until a shell acks it; the shell asks `{romp:'tapReplay'}` at boot and on `visibilitychange`→visible / `pageshow`, so a message posted to a suspended page, or to one iOS evicted and relaunched on the start URL, still lands. Per-tap `id` dedupes message/replay/link; `{romp:'tapLanded', id}` retires the kept tap. Last resort: a top-level client that `focus()` brought forward and that **still** reports `hidden` gets `WindowClient.navigate(deep link)` — only for a sid, only when its creation URL has no `push-reveal` yet, never a visible window (navigate is a full page load, so a live desktop window is never reloaded). A refused navigate ends quietly.

Tests: worker harness (id/diag, replay/ack cycle, every navigate gate), landing harness (rows, dedupe, ack, replay asks), new `LandingPushExecutes` (mobile-shaped DOM, all shapes), `[push] test` route line. 153 pass in the three files; full suite 9840 passed with two pre-existing served/Playwright legs failing identically on the shared main checkout without these changes (`test_dashboard_reload_served`, `test_ship_reship::ServedWedge`). vscode-extension typecheck/test (3595)/build green.

yatharth:
**52dd2beb — the warm-app tap, from the 2026-09-09 device round.**

What the instruments showed (structure only): with the app force-quit the tap landed (`deeplink` → `sw-message` → `reveal-post`, kernel `[reveal] link … parked`, `consumed`). With the app alive in the **background**, the kernel logged the `[push] … 201` and then nothing — no `deeplink`, no `sw-message`, no `reveal-post`. The cold-start rows explain it: the worker saw `clients:0, tops:0` even when a window existed. On iOS a backgrounded Home Screen app is **not enumerated by `clients.matchAll`**, so the worker took the `openWindow` road; iOS then foregrounded the *existing* page without navigating it (no load, so no deep link; no client, so no `postMessage`), and ended the worker right after the click — taking the in-memory `pending` tap with it, so the `tapReplay` safety net had nothing to answer.

The fix persists the tap where the **page** can read it without the worker:

- `_SW_JS` `notificationclick`: before any focus/open attempt, write `{id, sid, host, kind, cardId, url, t}` to the Cache API — one entry, `/__romp/tap` in the `romp-tap` cache — awaited under `waitUntil`, so the write is done before iOS moves on. A refusing store is swallowed; the other roads still run. `pending` stays as the fast path. `tapLanded` now also deletes the entry (only when it still holds the acked id — an older tap's ack never deletes a newer one).
- `_LANDING_REVEAL_JS`: on boot, `visibilitychange`→visible, `pageshow` and window `focus`, read the entry and run the existing `land()` path with `via:'store'` (`boot:true` when the page is booting), deduped by id with the message/replay/link roads; then delete the entry and ack the worker. A boot **on the deep link** drops a stored tap instead of landing it (the link is the newer word; the same tap's id is marked seen, so the message handed to the opened window is a dup — the cold start makes one `/reveal` now, not two). Age never retires a tap; only landing does.
- Instrumentation: every check files a `tap-resume` client-diag row `{found, via, ageS (clipped), dup, dropped, sameSid}` — so the next device round shows the resume firing (or not).

Tests (fail-first, Node harness): the worker harness gained a Cache API stub (`ServiceWorkerExecutes`: write before `matchAll`, sid-less no-write, id-matched ack delete, refusing store); the shell harness now boots from env (URL / seeded tap / no Cache API) and `LandingRevealResumesFromStore` plays the warm resume, pageshow and focus, both dedupe orders, two checks in flight, the relaunch on the start URL, a deep-link boot over a stored tap of another and of the same session, and a window with no Cache API. The `ServiceWorkerRoute` pin that banned the word `caches` now bans a *caching* worker (no fetch handler, no `respondWith`, only the tap store opened). `plans/ios-app.md` carries the note.

Suite: webpush + notify_popover + mobile 162 passed; full Python suite 9849 passed with two pre-existing failures on this branch tip that are unrelated and pass on the current integration tree (`test_dashboard_reload_served::ServedAutoReload::…gesture_ends…`, `test_ship_reship::ServedWedge::…reships_heals…` — both fail identically with this commit stashed); webview suite 3595 passed.

yatharth:
**b8d2a90b — the warm-app tap reproduced in a real browser, and a second bug it caught.**

The reproduction (`tests/test_notification_tap_resume_browser.py`, the `test_paste_to_focus_browser` pattern: hermetic kernel + Playwright Chromium, loud skips without either, ~5 s, no network):

1. The real shell and `/sw.js` are served with two synthetic sessions (`web`, `api`); the page loads, the worker is registered the way the bell's opt-in does and takes control of the page (`clients.claim`); `web` is put in front.
2. Inside the worker (Chromium exposes it via `context.serviceWorkers()`), iOS's observed behaviour is installed: `clients.matchAll` resolves `[]`, `clients.openWindow` resolves `null` **without navigating** — exactly the `clients:0 / tops:0 / road "open"` trail with no page load. A `notificationclick` carrying the routing block for `api` is dispatched at the real handler. (Headless Chromium reports `Notification.permission` denied whatever the context grants, so `showNotification` + a real `NotificationEvent` is refused there; the driver then dispatches a plain event carrying the two fields the handler reads, `.notification` and `.waitUntil`, at the same handler and records which road ran.)
3. The kept in-memory tap is cleared, as the worker iOS ended would have lost it.
4. The app "comes forward": the page gets the window `focus` and `pageshow` events a resumed page produces (Chromium will not flip `document.hidden` from a script).
5. Asserted: exactly one `POST /reveal` with `via: "store"` and no boot flag; the chat pane's `#tabs .tab.active` becomes `api`; the store entry is retired; the kernel log carries `[reveal] store sid=… : delivered`; a `tap-resume {found:true, via: focus|pageshow}` client-diag row is on file, structure only.

**Fail before / pass after.** Against the pre-fix sources (1e0fdc7b, in a throwaway worktree) the same test fails on the outcome the user saw: `the chat pane's active tab must become the session that buzzed; it is still 'aaaaaaaa-…' and the page posted 0 /reveal(s)` — with the worker having taken the openWindow road (`opened: '/?push-reveal=…'`), the tap kept only in memory, and nothing in the store. With the store (52dd2beb) it lands.

**The second bug.** With the store fix alone the reproduction passed only one run in three; the diagnostics said why: `[reveal] store … parked`, and the chat pane's socket was `/ws?app=chat&delta=1&iid=…` with **no `wid=`**, while the later panes carried one. The dashboard id was minted by the shell's *body* script, after the iframes, so on a fast origin the chat pane's shim read `sessionStorage` and connected before the id existed; a reveal aimed at a wid no chat socket carries is parked for a ready that never comes on a live page — and the cold path has the same hole (a park aimed at wid W is never consumed by a pane whose `ready` carries an empty wid). A phone's first-ever load runs the same race. The mint now lives in the head's existing script, ahead of `<body>` and so of the first `<iframe>`; the shell's pinned script count is unchanged, and `test_per_viewer_focus::test_the_id_is_minted_before_any_pane_can_connect` fails against the old order. After it: 3/3 runs green.

Suite: first half 4946 passed, second half 4905 passed (the two new tests included); the same two pre-existing, unrelated `Served*` failures as before, which fail identically at the branch tip with these commits stashed and pass on the current integration tree. Webview 3595 passed.

Issue #1118, opened 2026-09-08
Retire romp's own API-key providers: Claude Code's credential resolution becomes the only key path


henrypinkard:
romp should hold no API key. Today it can obtain one itself (a key command through `ROMP_API_KEY_CMD`, a 1Password reference through `ROMP_API_KEY_REF`, a static `ANTHROPIC_API_KEY=` line in `service.env`, remembered by a `service.env.source` marker) and inject it into every session's environment and every judge child. A key romp holds and hands out is a key a session can print, and a recent contributor PR's test did print one from a session's environment. The maintainers' decision (2026-09-08) is to remove that path and make Claude Code's own credential resolution the only one, following the `apiKeyHelper` recipe docs/reference.md already describes as an alternative.

## Target shape

- No key in `service.env`, in the kernel's environment, or in any session's environment. The user configures `apiKeyHelper` in Claude Code's settings; the helper prints the key; `CLAUDE_CODE_API_KEY_HELPER_TTL_MS` sets the refresh; rotation is a change to the vault item.
- `ROMP_EXPECTED_AUTH=key|login` in `service.env` declares the billing. The existing auth check, a session's reported `apiKeySource` against the declaration, stays.
- Judge children (`claude -p`) launch with no credential in their environment and resolve the helper themselves. The "romp holds no key source" notice on that path goes quiet: it is now the normal case.
- The two HTTP calls the kernel makes with a key itself, the model-catalog fetch and the fast-mode org probe, obtain it through the same contract: read `apiKeyHelper` from Claude Code's settings in Claude Code's precedence (managed settings, then the kernel cwd's `.claude/settings.local.json` and `.claude/settings.json`, then `$CLAUDE_CONFIG_DIR/settings.json`), run it in-process, keep the value in memory for the TTL, never in an environment variable or a file.
- A `service.env` that still carries `ROMP_API_KEY_CMD`, `ROMP_API_KEY_REF` or `ANTHROPIC_API_KEY`, or a `service.env.source` marker beside it, stops the kernel at boot with migration text naming the file and the variable names, in the shape the serve-token read already uses. Never a value.

## What goes

- `kernel/keysource.py` and its callers: the provider kinds, the env-file selection, the marker, `startup_api_key`, `work_api_key_source`, `work_api_key`, the `_options` injection, the key argument of `key_fast_org_env`, the judge wire `jd._WORK_KEY_FN` and the per-pass retrieval gate.
[...]

PR #1128, opened 2026-09-08
Retire romp's own API-key providers: Claude Code's credential resolution is the only key path

henrypinkard:
Closes the discussion in #1118.

## What

romp holds no API key. Claude Code's own credential resolution is the only key path: the user configures `apiKeyHelper` in Claude Code's settings, the CLI runs it and holds the key, `ROMP_EXPECTED_AUTH=key|login` in `service.env` declares the billing, and rotation is a change to the vault item behind the helper. The decision follows a contributor PR whose test printed a key from a session's environment: a key romp holds and hands out is a key a session can print.

**Retired.** `kernel/keysource.py` (the `ROMP_API_KEY_CMD` command kind, the `ROMP_API_KEY_REF` 1Password kind, the `ANTHROPIC_API_KEY=` line in `service.env`, the `service.env.source` marker), the kernel's injection of `ANTHROPIC_API_KEY` into session environments and judge children, the 1Password `OP_*` environment claim and the tmux server scrub, `romp keyswap` with `--cycle` and `--cycle-all`, the `/keycycle` route and `cycle_key`, the judges' per-pass retrieval gate, the env-file-vs-declaration check.

**Kept.** `ROMP_EXPECTED_AUTH` and the auth check (a session's reported `apiKeySource` against the declaration; undeclared stays quiet). The per-session Billing pick, which keeps choosing: see the mechanism below. The login tokens the kernel claims out of its own environment at startup (`ANTHROPIC_AUTH_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`) keep their treatment: popped so a key-billed launch never inherits them, handed back to login-billed launches.

**New: `kernel/credentials.py`.** The settings reader (Claude Code's own precedence: managed settings, a project's `.claude/settings.local.json` and `.claude/settings.json`, then `$CLAUDE_CONFIG_DIR/settings.json`; an empty string disables, a null falls through, as in the CLI; the kernel itself acts on the managed and user files only), the in-process helper runner for the kernel's two API calls (the model catalog and the fast-mode org probe) with an in-memory memo for `CLAUDE_CODE_API_KEY_HELPER_TTL_MS`, and the boot check.

**Loud boot failure.** A `service.env` that still carries `ROMP_API_KEY_CMD`, `ROMP_API_KEY_REF` or `ANTHROPIC_API_KEY`, a `service.env.source` marker beside it, or a kernel environment carrying one of those names stops the kernel before anything is spawned, with migration text naming the file and the variable names (the serve-token read's shape), never a value. The manager crash-loops the message into its log until the file is repaired.

## The login pick on a helper box

In Claude Code's precedence the helper outranks every login form, so a login pick has to disable the helper for that one process. romp already hands each session a per-session settings file (the SDK's `settings` option, the CLI's `--settings` layer); a login pick now writes `"apiKeyHelper": ""` into it and a key pick writes nothing. A login-billed judge call passes the same `--settings '{"apiKeyHelper": ""}'`. Verified on this box against Claude Code 2.1.257 with a fixture helper that leaves a marker when run, a hermetic `CLAUDE_CONFIG_DIR` naming it, no credential in the environment and the API base pointed at a closed port:

| launch | helper ran | CLI result |
|---|---|---|
| no override | yes | (request attempted with the helper's output) |
| `--settings '{"apiKeyHelper": ""}'` | no | `Not logged in · Please run /login` |
| `--settings '{"apiKeyHelper": null}'` | yes | (null does not clear the key) |
| `--settings '{"apiKeyHelper": "<other script>"}'` | no; the other script ran | (the layer overrides the configured helper) |

On a box with a helper, every session without a login pick bills the key, so `ROMP_EXPECTED_AUTH=key` describes such a box truthfully.

## Rotation

Change the vault item. Claude Code re-runs the helper after the TTL (five minutes by default) and on a 401 or 403, so running sessions and judges pick the new key up without a reconnect; the kernel's own memo expires on the same TTL. `romp keyswap` prints a note saying so and does nothing else (exit 2).

## Tests

Red first for the new behaviour: `tests/test_credentials.py` (the settings precedence including managed and the disable value, the helper run with a marker-writing fixture and its TTL memo and whitelisted environment, every failure mode in static words, the boot check for each retired line, the marker and each environment name, never a value; the login pick's settings payload and the key pick's absence of one; a launched session's environment carries no `ANTHROPIC_API_KEY`; the judge argv and environment for both billings; the first judge pass after boot runs keyless and latches nothing; the catalog credential from the helper; `main()` runs the boot check before anything spawns; the kernel and backend env validators agree). Existing modules moved onto the new surface; wholly provider-specific modules deleted. Shell: `romp keyswap` prints the retirement note and touches nothing; the help table no longer lists it; no entry point reads the retired names.

## Docs

`docs/reference.md`: per-session billing, the service environment (never a key; the boot failure), the helper recipe promoted to the setup proper with the settings files the kernel reads and its two in-process calls, the login, "Rotating the key". `docs/judges.md`: judges run on Claude Code's own credential resolution. `kernel/README.md`: `credentials.py` replaces `keysource.py`.

## Tier

`major-feature`: it changes what romp does with credentials. Filed without auto-merge; the merge is the maintainers'.

## Devbox note

This box already runs the target shape (no romp key source, `ROMP_EXPECTED_AUTH=key`, `apiKeyHelper` in the user settings) on the current kernel; this PR retires the code paths behind it and adds the boot check that refuses the old ones.

## Review folds (adversarial review of the first commit)

- **The kernel never runs a repository's helper.** The first cut read `apiKeyHelper` from a session's project settings too and would have run a checked-in `.claude/settings.json` command in-process at connect. The kernel now acts on the operator's two files only (managed and user): they decide the box's key side and name the one helper it runs; a project's helper is Claude Code's business (run behind its trust prompt), the auth check reports where such a session lands, and the fast-mode probe stands down for a session whose project resolves a different helper.
- **The manager refuses to start with a retired variable in its environment**, before it starts the tmux server: every pane inherits the server's globals, and the kernel alone refusing to boot left a leftover `ANTHROPIC_API_KEY` reachable from a `romp new -t` terminal session. `romp new -t` also refuses, loudly, a tmux server whose globals carry the variable, instead of scrubbing it.
- Two test modules that pinned the deleted fingerprint helper at import (api-health, auth hardening) and the Node test of the retired scrub moved onto the new contract; the dead launch-fingerprint read and its docstring branch, the stale `--env` comment and the conftest's stale reason are gone.

## Second review round (the maintainers' review)

- The 1Password CLI's names (`OP_SERVICE_ACCOUNT_TOKEN` and the rest, `OP_SESSION_*`) are refused at boot in `service.env` and in the environment like the provider variables, by the kernel and by the manager, with the token line named in the migration text.
- The login tokens the kernel claims at boot ride every launch that bills the login: a login pick and an unpicked session on a helper-less box, matching the judges; a key-billed launch never gets a bearer.
- Every "no helper" test floors the managed settings path (a conftest floor for every loaded copy of the module, pinned), so a box with a managed helper cannot make them lie.
- The helper value is held only within the TTL: cleared by a timer at expiry, on a failed run, and when the helper is removed from the settings.
- A helper set in managed settings outranks the per-session layer, so a login pick is refused with the reason, the picker offers no login choice, and a pick that predates the managed helper is said once at launch instead of billing the key quietly.
- The operator-scope reader is cwd-independent (pinned).
- An unreadable settings file makes the judges bill the login and say so once on stderr; the test that pinned an unreachable credential-error branch moved onto that real branch.
- `docs/judges.md` states the real behaviour of a helper that fails inside a judge call (a reported credential error latches; a hanging or retried call is killed at the alarm and files as a timeout).
- Deferred: a `romp update` preflight that runs the boot check on the remote before restarting it, so the dialing side sees a stale line's failure. Not in this PR; the remote's manager log carries the message today.

## A pre-existing parallel-runner artifact

`tests/test_postal_token.py::PostalLockFailureIsFailClosed::test_no_lock_tolerates_only_a_good_0600_token` creates its lock directory under a parent a sibling test creates, so it fails when the parallel runner schedules it alone in a worker and passes serially (31 passed on this branch and on main at 36b8ba45). Unrelated to this change; noted so a red parallel run is read correctly.


ksarma:
The PR is merged (b7e24a59, merge 212a3b31, 2026-09-09 00:06Z); nothing found would have held the merge, so the items below are follow-ups, plus two points from #1118 that the maintainers still have to settle.

Reviewed inline: `kernel/credentials.py` reads `apiKeyHelper` in Claude Code's documented precedence from the managed and user files only. The kernel's two direct calls run the operator's helper in-process under a whitelisted environment with a 15 s timeout; stdout reaches two request headers and a TTL memo, never a log. `check_boot_environment()` stops the kernel on the retired names, any `OP_` name or the `service.env.source` marker, and `bin/romp-manager` refuses the same names before starting tmux, naming variables and paths only. A login pick writes an empty `apiKeyHelper` into the per-session settings overlay, and the claimed login tokens return only to a login pick or an unpicked session on a helper-less box. Judge children lose the credential and `OP_` names and get the login tokens and `--settings` suppression for login billing only; `_key_available` reads the setting without running it. `mkdocs build --strict` exits 0. Privacy and vocabulary scans of all 2223 added lines are clean, and red-first holds. A read-only check of this machine found 0 retired names in the user manager environment, `service.env` and the live manager, so the next restart is not refused.

Follow-ups, most consequential first:

- `kernel/kernel.py:14078`, `kernel/sdk_backend.py:8297` and `:10147`: `helper_source()` raises `CredentialError` on an unreadable or malformed settings file; no caller catches it, though `key_available` does, so the picker gets no session list and a login-picked launch fails. Route all three through one guarded reader, with a test.
- `bin/romp:2903-2904` and `docs/reference.md:471`: the `romp new -t` remedy names `romp on restart`, which `bin/romp` does not dispatch, and a manager restart cannot clear a scoped tmux server's globals, so the refusal repeats; the guard checks `ANTHROPIC_API_KEY` only where the manager checks eight names. Name `tmux set-environment -gu ANTHROPIC_API_KEY` or ending the server, mirror the manager's list, pin the remedy text.
- `kernel/credentials.py:174`: the managed tier is one file, while Claude Code also merges `managed-settings.d/*.json`; a helper delivered there reads as no helper, so the picker offers Login and `set_auth('login')` is accepted against the module's managed rule.
- `kernel/sdk_backend.py:3022`: a login pick whose per-session settings file cannot be written launches with the helper on and bills the key; the problem row says the opposite. Decide refuse or degrade, fix the rows, add a test.
- `kernel/sdk_backend.py:8362`: the spawn seed guard has no login equivalent, so a remembered login default keeps seeding sessions on a managed-helper box, each filing a per-init row that names the login, not the helper.
- `kernel/credentials.py:54`: the whitelist drops non-secret locators such as `VAULT_ADDR` and `AWS_PROFILE` that the CLI passes to the same helper, so a helper that works in sessions fails both kernel calls; pass a small named set or document the restriction.
- `docs/reference.md:300` and `:562` (also `kernel/README.md:34`, `tests/test_judge_auth_billing.py:15`) say the kernel reads the helper for its working directory; it reads the managed and user files only. `docs/reference.md:554` says a failing helper inside a judge latches judge-auth-down; `docs/judges.md:383-388` and the code say it dies at the 120 s alarm and files a generic call row.
- `tests/tmux-private.bash:31`: the three manager bats suites floor none of the retired names, so a shell exporting `OP_SERVICE_ACCOUNT_TOKEN` turns all 13 red while CI stays green; add the floor `conftest` got.
- Nits: `tests/test_credentials.py:195` sleeps 0.5 s on a real timer instead of joining it; `kernel/sdk_backend.py:3305` has two untested `helper_fast_org_env` branches; stale references at `kernel/judge.py:2348` (key gate), `tests/romp-node-launch.bats:91` (`kernel/keysource.py`) and `tests/test_api_health.py:866` (`romp keyswap`); `docs/reference.md:468` omits that the default start limit (5 in 10 s) fails the unit after five refusals, so the repair also needs `systemctl --user start romp-manager`; `bin/romp:733` tells a login-only box to declare `ROMP_EXPECTED_AUTH=key`; the body omits the tightened `/env` rule.

For the maintainers to settle (from #1118): (1) requirement 3, a one-line boot notice for an undeclared `ROMP_EXPECTED_AUTH`, was declined on the issue; the PR keeps the quiet behaviour (`kernel/sdk_backend.py:3196`). (2) requirement 2's judge half, a helper-specific problem row for the rejected-key stall and an up-front `init.apiKeySource` check, is neither built nor listed as deferred; the judge child still runs `--output-format json` (`kernel/judge.py:787`), so a rejected key reads as a model fault.

Tests: `tests/test_credentials.py` fails to import at the base 10ffb1ce and passes at head; three of the four new `romp.bats` tests fail at the base, all pass at head; the touched Python suites pass at head; `npm test` 3611 passed; `mkdocs build --strict` exits 0; CI nine of nine at b7e24a59.

At review time the only conflict against main was `docs/reference.md`. #1022 must be re-derived: its test staged variables through the deleted keysource module, so #1128 had to land first. #1062 has one trivial hunk in `begin_pass_frame`.

Review run by the maintainer's automated review session; verified against head b7e24a59.

PR #1131, opened 2026-09-08
Kernel: the timeline's dead lanes are derived once and served while their inputs stand; their parses are dropped

henrypinkard:
## What

The second batch of the idle-kernel work; the first (the pusher's change gates, #1113) halved the CPU and left the resident set where it was. This batch is the memory lever and the rest of the timeline's per-cycle cost.

`/perf` on the maintainer's devbox showed the timeline rebuilt on every pusher cycle with zero cache hits: the dashboard signature's 5 s bucket turns over faster than a 6 s cycle, and every rebuild re-parsed every lane's transcript and goals, dead lanes included (the majority within the 12 h window, none of which had changed), then re-encoded every bar of every lane into a 15 MB frame.

- **A dead lane is derived once and served while its inputs stand.** Its parse-derived parts (bars, compactions, the work end, its judging marks) are memoized per lane, keyed on the stat of every file they read: the transcript and its states file, the goals store and its user-override journal (replayed onto the store on every load, so an override moves the goals without moving the store file), the captions file the builder reads, the archive, the session flags, and the branch clip. The clock-dependent parts (awaiting and compacting intervals, `since`) are derived per build as before, so a served lane's frame is byte-identical to a rebuilt one. A goals fault or a marks failure is never cached. Live lanes are untouched.
- **Judging marks are derived once at horizon zero and stamped** with the value the horizon test compares, then filtered per build on exactly that and handed out without the stamp. The stamp matters: the diary and distiller marks are compared on their evidence time but emitted at the segment's work end, so a filter on the emitted time would admit marks a fresh derivation drops.
- **A cached dead lane's parse is dropped from the parse cache.** Nothing else reads a dead session's parse per cycle, and those parses were the bulk of a multi-GB resident set (a 166 MB transcript parses to about 220 MB; two caches held one each). A lane that moves re-parses once; a session that revives is parsed by its chat build as before.
- **The bars encoder reuses the JSON string** of an entry object it encoded in the previous split of the same collection (identity checked, never an id alone; the memo is rebuilt from each split, so it never outgrows one build).

## Before and after

| measure | before (13 min after boot) | after the first batch (12 min after boot) | after this batch |
|---|---|---|---|
| process CPU | about 88% of a core | about 40% of a core | to be posted here after it lands |
| timeline builds | 84 built, 0 served | 17 built, 0 served | |
| resident set | 4.4 GB + 1.4 GB swap | 4.3 GB + 1.1 GB swap | |

## Tier

`fix`: behaviour is unchanged by design (a served lane is byte-identical to a rebuilt one); the work is the timeline doing less of it.

## Tests

`tests/test_timeline_lane_memo.py` (new) builds a real dead lane from a synthetic transcript: the second build serves it with no parse and the parse is gone from the cache, the served lane and bars equal the rebuilt ones, a moved transcript re-derives, a live lane is never memoized, the stamp never reaches the wire; the stamped derivation filtered at six horizons equals a fresh derivation mark for mark, including the straddling closer mark; the entry memo reuses strings by identity and never grows. The wire test's splitter doubles pass the collection key through. Timeline, delta, wire, perf, isolation, change-gate and comment-thread suites: 306 passed on the rebased head.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -34477,7 +34477,7 @@ def _bind_message_execs(messages, turns):
 JUDGE_CAP_LIMIT = 80   # most-recent caption marks kept per session (the view merges adjacent marks anyway)
 
 
-def _derive_judging(sid, caps, goals, t0, out, seg_ends=None):
+def _derive_judging(sid, caps, goals, t0, out, seg_ends=None, stamp=False):
     """Append this session's JUDGE-activity marks to `out` — the second-timeline (data.judging) feed,
     read from the real artifacts the summarizer judges write (docs/judges.md):
       captioner ← captions/<sid>.jsonl   (one mark per segment/turn unit)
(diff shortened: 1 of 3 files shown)


henrypinkard:
After-numbers with both batches deployed, read from the kernel 21 minutes after the deploy restart (every session reconnecting, judges catching up), the same phase the before-numbers were read in:

| measure | before (13 min after boot) | after the first batch (12 min) | after both (21 min) |
|---|---|---|---|
| process CPU | about 88% of a core | about 40% of a core | about 86% of a core |
| pusher cycle p50 / count | 6.6 s, 99% busy | 8.0 s, 25 cycles | 1.2 s, 775 cycles, 79% busy |
| pusher time by stage | (not split) | (not split) | jobs 831 s, push 181 s (chat 158 s, timeline 45 s, feed 34 s) |
| timeline builds | 84 built, 0 served | 17 built, 0 served | 14 built, dead lanes served from the memo |
| resident set | 4.4 GB + 1.4 GB swap | 4.3 GB + 1.1 GB swap | 5.6 GB, no swap |

What the numbers say:

- The two batches did what they set out to do inside the push: a cycle's push work is now about 0.2 s (181 s over 775 cycles) against 6 to 8 s before, and the timeline no longer re-derives dead lanes. Cycles are now short enough that the pusher runs once per event wake (660 wakes in 21 minutes) instead of coalescing them.
- The CPU did not fall in this window because the pusher's time moved to the jobs stage: 831 s of the 1296 s window, with 60132 goal-store loads over it. That is the judge and goal work after a deploy, which neither batch touched. It is the next lever, and it wants its own measurement (which jobs, how many loads per job, what is re-read).
- The resident set is unchanged in total (the swap is gone, the number is the same). The dead-lane eviction had little to evict on this box: most lanes inside the 12 h window are live sessions, whose parses stay cached by design, and the largest of those transcripts run to hundreds of megabytes. Bounding live-session parse memory is a separate change.

So: the idle-path claim holds (a quiet pusher cycle is cheap now), the whole-process claim does not yet, and the reasons are named above rather than left to a later surprise.

PR #1139, opened 2026-09-09
Kernel: a restart-cut turn's whole process tree dies before the resumed CLI launches (T276)

henrypinkard:
When a kernel restart cuts a session's turn, the boot reaper used to SIGTERM the orphaned CLI pid alone, so the turn's Bash-tool processes lived on: a stress harness's 32 busy loops and a benchmark's 11, setsid'd from tool shells and re-parented to the user manager once those shells died, burned cores for over an hour (the user 2026-09-08).

## Design points

- **End the tree, not the pid, and before any cut turn resumes.** The reaper first stops the session's transient scope unit when the CLI runs in one (read from the CLI's own `/proc/<pid>/cgroup`; `bin/romp-cli-scope` names it `romp-session-<sid8>-<pid>-<t>.scope`): systemd ends every process in the cgroup, setsid children and re-parented leftovers included. Then, always, it walks the process tree the `ps` listing still shows under the CLI and signals each descendant's own process group where it leads one (a setsid child), else the process; children before the CLI; SIGTERM, then SIGKILL after a short grace.
- **Leftover scopes.** A scope outlives its CLI when a tool's children keep running, which is exactly the loops the pid-only reap left behind. The reaper lists `romp-session-*.scope`, keeps the units whose sid8 is one of our sessions', skips any whose pid is this kernel's live child, and stops the rest.
- **Nothing outside the session's scope or tree is signaled.** The walk is by ppid from the CLI; this kernel's own process group is never a target; the sweep is filtered to our sids. A descendant whose parent died before the walk is out of the tree's reach; the scope closes that gap, which is why it is tried first.
- **Seams resolve at call time** (`kill`, `run`, `cgroup`), so the existing reaper pins (patched `os.kill` / `subprocess.run`) hold unchanged.

## Tests

`tests/test_cut_turn_tree_kill.py`: pure rules (scope from a cgroup listing, our units from a list-units listing, the ppid walk and its exclusions); the scope path on scripted seams; the no-scope fallback (descendants first, CLI last, bystander untouched); the leftover sweep (our dead session's scope stopped, our live child's and another session's kept, no systemctl means nothing to sweep); the boot reconcile wiring; and real processes on Linux: a fake CLI whose child spawns a detached sleep loop dies with the cut while a bystander survives. The loops sleep (no synthetic load) and every process the tests start is killed by them.

### DIFF
--- a/kernel/sdk_backend.py
+++ b/kernel/sdk_backend.py
@@ -2643,6 +2643,102 @@ def find_orphan_clis(ps_lines: list[str], lastsids: list[str], own_pid: int) -> 
     return out
 
 
+# ENDING A CUT TURN'S WHOLE TREE (T276, the user 2026-09-08). Reaping the orphaned CLI alone left its Bash
+# tool's processes alive: a stress harness's 32 busy loops and a benchmark's 11 (setsid'd from tool shells,
+# re-parented to the user manager once the shells died) burned cores for over an hour after restarts.
+# The CLI runs in a transient scope of its own when bin/romp-cli-scope is in use (`romp-session-<sid8>-
+# <pid>-<t>.scope`), and a scope holds EVERY process the CLI ever spawned — setsid changes the session,
+# not the cgroup — so stopping the unit ends the tree, dead intermediates included. Without a scope the
+# fallback walks the process tree from the CLI in the `ps` listing and signals each descendant's process
+# group (a tool shell's setsid child leads its own group), then the CLI; a descendant whose parent died
+# before the walk is out of reach there — the scope is what closes that gap, which is why it is tried
+# first. Nothing outside the CLI's own scope or tree is ever signaled: the walk is by ppid from the CLI,
+# and the scope list is filtered to OUR sessions' units whose pid is not a live child of this kernel.
+SESSION_SCOPE_PREFIX = "romp-session-"
+_SESSION_SCOPE_RE = re.compile(r"romp-session-([0-9a-fA-F]{1,8})-(\d+)-\d+\.scope\Z")
+SCOPE_LIST_ARGV = ["systemctl", "--user", "list-units", "--all", "--plain", "--no-legend", "--no-pager",
+                   SESSION_SCOPE_PREFIX + "*.scope"]
+SCOPE_STOP_TIMEOUT = 15.0     # systemd's own stop: SIGTERM to the cgroup, SIGKILL at its TimeoutStopSec
+TREE_KILL_GRACE = 1.0         # seconds for SIGTERM to land on the tree before SIGKILL
+
+
+def scope_unit_of(cgroup_text: str) -> str | None:
+    """The romp session scope a /proc/<pid>/cgroup listing places the process in (cgroup v2: one
+    `0::/user.slice/…/romp-session-<sid8>-<pid>-<t>.scope` line; the legacy hierarchy's lines carry the
+    same path per controller), or None when the process runs in no such scope."""
+    for ln in cgroup_text.splitlines():
+        path = ln.rsplit(":", 1)[-1]
+        for comp in path.split("/"):
+            if comp.startswith(SESSION_SCOPE_PREFIX) and comp.endswith(".scope"):
+                return comp
+    return None
+
+
+def scope_pid(unit: str) -> int | None:
+    """The pid a session scope's name carries — the pid its CLI ran as (bin/romp-cli-scope's naming)."""
+    m = _SESSION_SCOPE_RE.match(unit.strip())
+    return int(m.group(2)) if m else None
+
+
+def session_scope_units(list_lines: list[str], lastsids: list[str]) -> list[str]:
+    """The session scopes in a `systemctl --user list-units 'romp-session-*.scope' --plain --no-legend`
+    listing that belong to one of OUR sessions (the name's sid8 is the first 8 characters of a lastSid),
+    in listing order. Another kernel's sessions (other sids) never match. Pure."""
+    sid8 = {s[:8].lower() for s in lastsids if s}
+    out = []
+    for ln in list_lines:
+        head = ln.strip().split(None, 1)
+        if not head:
+            continue
+        m = _SESSION_SCOPE_RE.match(head[0])
+        if m and m.group(1).lower() in sid8:
+            out.append(head[0])
+    return out
+
+
+def descendants(ps_lines: list[str], root: int) -> list[int]:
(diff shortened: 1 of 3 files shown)


ksarma:
Every finding below is a follow-up, because the PR is already merged. It landed at a1a427af (merge f198259b), the owner's auto-merge on green under the feature tier. Three findings would have held the merge; each is marked as such.

Reviewed inline: the four pure helpers (`scope_unit_of`, `scope_pid`, `session_scope_units`, `descendants`) and their pins; `_end_cli_tree`'s scope stop and ppid walk; `_stop_leftover_scopes` and its keep rule; and the boot reconcile wiring, where the sweep runs before any session is respawned. For `_end_cli_tree`, that included which seams it resolves at call time and which calls stay outside them (`os.getpgid`, `os.killpg`, the procfs liveness poll). Also checked: the sid the wrapper names a unit from against the sid the sweep filters on; how the three systemctl calls handle exit status; every test's seams, including which tests leave the real cgroup read and the real systemctl reachable; and `docs/reference.md` at this head against the new behavior. For a wrapped CLI whose scope is its own, the tree kill does what the body describes. The interdiff from 6cb6dbe6 changes only fake pids (all above pid_max now) plus a rebase. The kernel code is byte-identical, so each earlier finding was re-checked at the shifted lines. Privacy and vocabulary scans of the added lines are clean.

Findings:

- `kernel/sdk_backend.py:7138`, blocker (would have held the merge). The scope stop targets whatever romp-session unit the target pid's cgroup names. It never checks that `scope_pid(unit)` equals the pid, or that the unit is not the kernel's own scope. The sweep at 7195 has no self-scope guard either. An unwrapped CLI that inherited a host session's scope makes the reaper stop that host session. That happens with a kernel started from a session's tool shell (scopes are off outside the service). `bin/romp-cli-scope` names the unit with the CLI's own pid, so the check is exact and available.
- `tests/test_cut_turn_tree_kill.py:232`, blocker (would have held the merge). `RealProcessTree` calls `_end_cli_tree` with no cgroup or run seam, so the fake CLI inherits the tester's cgroup. Run from inside a romp session on Linux, the real systemctl stops the session running pytest. This cut two review sessions on 2026-09-09. `ROMP_CLI_SCOPE=0` governs only the wrapper's launches and does not move the tester out of its scope, so the docstring at 196 misdescribes the isolation. CI has no such scope and cannot see this. The full pytest suite at main is not safe to run from inside a romp session without ignoring this file. Fix: pass `cgroup=lambda pid: ""` and a recording `run`, and skip when `/proc/self/cgroup` names a romp-session scope.
- `kernel/sdk_backend.py:7246`, blocker (would have held the merge) and a design question. `_stop_leftover_scopes` runs on every kernel boot, for every alive registration, before any session is respawned. The keep rule at 7195 therefore spares nothing: every still-loaded scope of ours is stopped, cut turn or not. A scope is still loaded exactly when a `tmux -L` server or setsid job started from a tool shell is running in it. `_end_cli_tree`'s scope stop at 7142 does the same to a cut session. `docs/reference.md:635-637` and 673-676 still promise that a kernel restart never touches that work (the #953 design). The PR changes no docs, and the body describes the swept units only as leftover loops. Question for both maintainers: does the restart promise stand? If yes, the sweep needs a narrower trigger or removal, and the docs stay. If no, `docs/reference.md` 622-637 and 669-682 and `bin/README.md` need rewriting, and the change is a contract change (tier 3).
- `kernel/sdk_backend.py:7163`, follow-up. `os.getpgid`, `os.killpg` and the liveness poll sit outside the kill seam. The pid_max fixtures cover only the pids they moved. `tests/test_sdk_lifecycle_hardening.py:433-439` still hands the reaper orphan pid 559. Route group signalling and liveness through a seam; move 559 above pid_max meanwhile.
- `kernel/sdk_backend.py:7112`, follow-up. Without `/proc`, `_pid_alive` falls back to `os.kill(pid, 0)`. The recording mock in the boot pins intercepts that call (`tests/test_cut_turn_tree_kill.py:183`; `tests/test_sdk_lifecycle_hardening.py:413, 437`). Every target then reads alive, and the pins fail on the workflow_dispatch macOS leg that `scripts/release.sh` runs before a tag. A liveness seam fixes this and the previous item together. Land it before the next release.
- `kernel/sdk_backend.py:7231`, follow-up. The sweep filters on `lastSid[:8]`, but the wrapper names the unit from `ROMP_SID`, which is `sess.sid`. A session that ran `/clear` or forked on resume is therefore never swept. No fixture has `lastSid != sid`. Land this with or after the design call above, since widening the filter widens that item's reach.
- `kernel/sdk_backend.py:7142`, 7186, 7198, follow-up. systemctl's exit status is never read. A nonzero stop counts toward the reconcile line's stopped count, and a nonzero listing is a silent zero. `_end_cli_tree`'s failure log fires only on exceptions. The same file reads `returncode` for its scope probe.
- `kernel/sdk_backend.py:7195`, follow-up. The keep rule spares only this kernel's children. A second kernel on the same state root stops the other kernel's live session scopes, for example a hand-run kernel on another port, or the manager's bare `k<port>` kernel with no stateDir. `find_orphan_clis` spares those via `_is_kernel_cmd`. Mirror that check.
- `kernel/sdk_backend.py:7235`, follow-up. One ps listing feeds every orphan, while each reap can block up to 16 s. A pid from the stale listing can therefore be signalled, as a whole group where its new holder leads one. This needs a pid wrap inside the window: unlikely at pid_max 4194304, plausible at 32768. Re-read the listing per orphan, or verify each target before signalling.
- `tests/README.md:33`, follow-up. Neither new convention is recorded beside the systemd-run rule: fake target pids above pid_max, and reaper tests stubbing `cgroup` and `run`.
- `tests/test_cut_turn_tree_kill.py:168`, follow-up, minor. `BootReconcileEndsTheTree` leaves `_ensure` unstubbed, so the cut session reaches the real spawn path. The lifecycle suite stubs it.

Test evidence: review by reading only. Nothing from this PR was executed here. The reaper reads a pid's cgroup and stops the scope it names, and the review session runs inside a romp-session scope; that is the hazard the second finding describes. CI: every required check green at a1a427af (ubuntu, Python 3.10 to 3.13). The Python 3.12 failure at 6cb6dbe6 was fixture pid 4242 being live on the runner, which the pid_max commit removed.

Since this record was drafted, #1148 (open at cd69f5d4) folds the ownership check on the scope stop, the failed-stop return code, a pid identity check at the ps snapshot, and the real-process test's cgroup seam. Its review follows on that PR. The design question about the boot sweep and the remaining follow-ups stand.

Review run by the maintainer automated review session; verified against head a1a427af.

PR #1141, opened 2026-09-09
Kernel: the auto-nudge walk derives its placement gate once per (parse, store) and reads the store through the shared view

henrypinkard:
## What

The third idle-kernel batch. After the two push-stage batches, the pusher's cycle push cost about 0.2 s, but the kernel still burned most of a core: the pusher's **jobs** stage (the ticks that run every cycle beside the push) held the time. Measured on the maintainer's devbox, live kernel, no browser connected:

| measure | before |
|---|---|
| jobs stage share of wall time | 78% to 79% (three and six minutes after boot) |
| goal-store loads per pusher cycle | 55 to 66 |
| process CPU | about 90% of a core |
| py-spy, 90 s, 3905 samples | 77% in the jobs stage; 91% of that in the auto-nudge tick; within it the planner-placement gate's derivation (segment keys, seams, plan units, placement-key normalization) and the per-session store load |

The auto-nudge walk ran, for every idle session on every cycle, with nothing changed since the previous cycle:

- the **planner-placement gate**: re-segment every turn, apply the seams, build the plan units, normalize every recorded placement key;
- a **fresh goal-store load**, most of the cycle's loads.

## The change

- **The gate is derived once per (parse, store) and served while both stand.** Its answer is a pure function of the parse (the parser's own cache key: the transcript's and the states file's stat, the pending cut) and the store's bytes (the store file's, its override journal's and the clears log's stat; the placement-key function scopes its fuzzy match by the episode floor, which the clears log holds). A parse outside the cache is derived every time and never cached; a failed derivation stays loud and is never cached. Bounded by the session count. Counted on `/perf` under `memos.nudgeGate` (`served`, `derived`).
- **The walk's read of the store is the shared read-only view**, one parse per file version, shared with every pusher builder. Every writer downstream reloads at its write moment (the wake's fresh load, the fire list's, the fire path's), and a write through the view raises rather than landing, so a writer that forgets fails loudly.
- **The interrupt-block tick's stand check** (a pure read per marked session per cycle) reads the shared view too; its two writers keep their fresh loads.

No card-move rule changes: the gate answers the same question from the same inputs, and the writers write the same stores at the same moments.

**Not folded here, and why.** The awaiting-lift tick loads fresh per session per cycle because it writes on the same object (2% of the jobs stage in the profile); splitting its read from its write is its own change. The interrupt tick's per-session re-parse is the live-session parse cost the parked memory lever owns.

## Tier

`fix`: performance, behaviour unchanged.

## After-numbers

To be posted here after this lands and the devbox converges, read at the same uptime after a deploy restart: loads per cycle, the jobs stage's share of wall time, process CPU, and the gate's served-to-derived ratio.

## Tests

`tests/test_nudge_gate_memo.py` (new): the gate is derived once and served on the second call; re-derived when the parse, the store, the override journal or the clears log moves; its answer equals the direct derivation before and after a placement lands; a failed derivation is never cached and stays loud; a parse outside the cache is derived every time; the memo is bounded; the counters ride the perf snapshot; source pins for the shared-view reads and the writers' fresh loads. The nudge bundle fixture stubs the shared read beside the fresh loader (it modelled the walk's snapshot through the fresh loader alone, which under the parallel runner read a store another module left at the shared placeholder sid). The nudge suites pass unchanged.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -884,15 +884,19 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   loader (`load_goals`) and `save_goals`; the pusher's read-only loads go
   through the shared store cache and show under `memos.shared`, not here. A
   save that would rewrite identical bytes is a save without a write.
-- `memos`: the three identity memos on the goal-store path. `pass` is the
+- `memos`: the identity memos on the goal-store path. `pass` is the
   judge pass's stat-keyed store memo (`hit`, `miss`, `fail`, `evict`, `punch`,
   and its occupancy `entries`, `bytes`); `shared` is the pusher's shared
   read-only store cache (`hit`, `miss`, `compare_miss`, `refuse`, `dup`,
   `absent`, `corrupt`, `unreadable_journal`, `evict`, `fallback`, `poisoned`,
   with `entries`, `bytes` and `off`); `chain` is the write-moment chain memo
-  (`hit`, `miss`, `populate`, `bypass`). The compaction sweep after each judge
-  pass evicts from `pass` and `shared` the entries of stores no session in the
-  discover window owns, so both stay bounded by the live board.
+  (`hit`, `miss`, `populate`, `bypass`); `nudgeGate` is the auto-nudge walk's
+  planner-placement gate, derived once per (parse, store) and served while
+  both stand (`served`, `derived`; a healthy quiet box serves almost every
+  cycle). The compaction sweep after each judge pass evicts from `pass` and
+  `shared` the entries of stores no session in the discover window owns, so
+  both stay bounded by the live board; the gate memo is bounded by the session
+  count.
 - `judge`: `passes`, `ms_sum`, `ms_last`, `ms_mean` (wall time; a pass waits
   on model calls), `cpu_ms_sum` (CPU time of the judge tier threads and every
   per-session worker they run; the workers' share is `cpu_ms_workers`).
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -381,6 +381,7 @@ class _PerfStats:
                 memos[key] = read()
             except Exception:
                 memos[key] = {}
+        memos["nudgeGate"] = dict(_NUDGE_GATE_STATS)   # the nudge walk's placement gate: served vs re-derived (2026-09-09)
         now = time.time()
         return {"now": now, "since": since, "uptime_s": now - _STARTED, "log": _PERF,
                 "process": _process_stats(), "pusher": pusher, "stages_ms": stages,
@@ -8407,7 +8408,7 @@ def _intr_block_stands(sid, gid):
     once the file read again."""
     if not gid:
         return False
-    store, fault = jd.load_goals_or_fault(sid)
+    store, fault = jd.load_goals_shared_or_fault(sid)   # a pure read: the shared view, one parse per file version (2026-09-09)
     if fault is not None:
         return True
     nd = store.get("nodes", {}).get(gid)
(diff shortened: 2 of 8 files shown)


ksarma:
Recommend holding this PR: two blockers need fixes first; after that it merges on green.

Reviewed inline: the memo key at `kernel/kernel.py:10741` cannot serve a stale answer on a same-second same-size rewrite. Its parse half binds to the parser's cached turns object, and its store half binds to the shared view object's identity. It takes the clears-log stat before the derivation. It caches only when the shared loader still returns that same view afterwards. The derivation helpers (`_segs`, `plan_units`, `_placed_key`) never write. The gate's answer equals the direct derivation before and after a placement lands, so no card moves change. The interrupt-block stand check at 8411 is a pure read. The fire list, the fire path and the wake's due leg reload fresh. The wake's answered leg does not; it hands the frozen view to a writer, which is the first blocker below. The CI red is the PR's own and not caused by main moving. The failing file is byte-identical at base, head and main. The pair `test_clear_wrap.py` plus `test_nudge_injected_turn_arm.py` passes at the base and fails at the head with CI's exact four. The cause: the fixture stubs only `jd.load_goals`, while the walk now reads `load_goals_shared_or_fault`. That loader serves a leftover store at the shared placeholder sid before falling back to the stub. The label fits this series' precedent, though the diff adds a mechanism. Under the owner's role the gate is the same.

- `kernel/kernel.py:10077`, blocker. `_wake_goal`'s answered leg passes the walk's shared view to `_file_wake_answer`. That helper calls `record_verdict`, `rollup_status` and `save_goals` on it with no reload. The `setdefault` on a frozen node raises inside the helper's own `except`. The awaiting row is never filed, and the leg never re-runs because `answeredAt` is already recorded. `_shared_poison` then switches the shared cache off for the process until restart, which also stops this memo caching. Fix: reload at the write moment, with the helper taking its own `jd.load_goals(sid)`; the caller at 10221 already passes a fresh load. Add a test that drives the answered leg on a file-backed store and asserts the row lands and `jd.shared_store_stats()["off"]` stays 0.
- `tests/test_nudge_injected_turn_arm.py:165`, blocker. Stub `load_goals_shared_or_fault` to `(self.store, None)` beside `load_goals`, and add the name to `_orig_jd` at line 144. These are the same two lines the PR applied to the bundle fixture. Verified green in all four contaminating pairs and alone.
- `tests/test_nudge_gate_memo.py:252`, follow-up to land with the fix. The string pin matches the due leg's load and passes with the regression present. The behavioral test above replaces it. The body and the 10858 comment should drop the every-writer-reloads and suites-pass-unchanged claims and describe the identity-bound key the second commit built.
- `kernel/kernel.py:374` and `tests/test_perf_stats.py:117`, follow-up. Both comments still count three identity memos beside the fourth. The memo holds MAX+1 entries, since eviction runs only past 512; the bound test asserts 513.

Red-first at the base 19f2e1f0 held: `tests/test_nudge_gate_memo.py` 13 failed (AttributeError, no `_nudge_gate_memo`) and `tests/test_perf_stats.py` 1 failed (nudgeGate absent from memos). At head the three PR test files pass 81/81. The frozen-view write reproduces on a file-backed store. At head the helper returns False with off=1, poisoned=1 and no row; at the base it files the row with off=0. The serial CI-order run at a local merge with main reproduces CI's four failures (4 failed, 6596 passed). The pair with the two-line stub passes 15/15. CI at head: four Python jobs red on those four tests; Shell, extension and tier checks green.

Review run by the maintainer's automated review session; verified against head 6afb67f1.

henrypinkard:

Live kernel on the shared box, `/perf` readings, no synthetic load. "Before" rows are the readings taken for this PR's opening measurement on the previous code (same box, same reading script); the 15 min row is tonight's reading on the old code minutes before the deploy.

| code | uptime | goal loads / cycle | jobs stage (% of wall) | CPU (% of a core) | RSS | gate served / derived |
|---|---|---|---|---|---|---|
| before | 3 to 6 min | 55 to 66 | 78 to 79% | 80 to 90% | 3.8 GB | n/a |
| before | 22 min | 63 | 75% | 83% | 5.6 GB | n/a |
| before (tonight) | 15 min | 122 | 70% | 96% | 4.7 GB | n/a |
| after | 3.3 min | 56 | 45% | 67% | 3.7 GB | 3668 / 50 |
| after | 6.5 min | 55 | 44% | 69% | 4.0 GB | 7318 / 51 |
| after | 20.6 min | 80 | 43% | 80% | 4.0 GB | 20605 / 55 |

**Cycle rate doubled, so per-cycle cost is the fairer comparison.** The pusher ran 1203 cycles in the 20.6 min after (about one a second) against 472 cycles in tonight's 15.5 min before (one every two seconds): the cheaper stage keeps up with wakes it used to coalesce. Per cycle, the jobs stage went from 1.38 s (70% of 932 s over 472 cycles) to 0.44 s (43% of 1237 s over 1203 cycles), and process CPU from 1.9 to 0.82 core-seconds. The 20 min CPU share of a core (80% against 83% before) reflects that doubled cycle rate during an active hour, not the stage's cost.
The shared view served the walk's reads: 4430 hits against 23 misses at 4 min. The placement gate was derived 51 times over 437 cycles and served 7318 times.

**Goal loads per cycle did not fall.** The walk's read left the counter (it goes through the view now), so the ~55 loads a cycle that remain belong to another reader: a 60 s py-spy sample at 4 to 5 min uptime (1858 samples) puts them in `_awaiting_wake_outcomes`, which does a fresh `jd.load_goals` per wake record per pass (17% of the nudge tick's samples).

**Where the jobs stage goes now.** The stage is 42% of the kernel's samples (77% before). The nudge tick is 73% of it, and 60% of the tick is `_cleared_ids()`, called from `_auto_nudge_session` after the placement gate: it re-reads and re-parses the whole append-only `cleared.jsonl` (2237 lines on this box) for every session on every cycle, with no memo. `_open_top_goal` calls it too. A stat-keyed memo of the parsed clear set (the log is append-only with undo rows, so the file's stat is the change signal) and routing the wake sweep's read through the shared view are the next batch; neither changes a card-move rule.

ksarma:
The PR merged before this record closed. The first of the two blockers filed at 6afb67f1 was not fixed. It is now a regression on main and needs a fix PR.

Reviewed inline: the only commit after the earlier review is 316b65d7. It changes three test fixtures and leaves `kernel/kernel.py` byte-identical to 6afb67f1. It closes the second blocker. `tests/test_nudge_fresh_guard.py`, `tests/test_nudge_injected_turn_arm.py` and `tests/test_nudge_memo_deadlock.py` now stub `load_goals_shared_or_fault` beside `load_goals` and restore it. The pair with `tests/test_clear_wrap.py` that CI failed on passes. The first blocker stands. The walk binds `store` to the shared read-only view at `kernel/kernel.py:10861` and hands it to `_wake_goal` at 10928. The answered leg records `answeredAt`, then calls `_file_wake_answer` on that view. That helper's `record_verdict` write raises `FrozenStoreError` inside its own except. The failure reproduces on a file-backed synthetic store at 316b65d7 and at main 28ecee8c. No awaiting row lands. `shared_store_stats` reads off=1 and poisoned=1. A frozen-store-write judge-errors row is filed. The gate memo this PR added derives on every later call. At the base 19f2e1f0 the row lands with the cache on. `answeredAt` is already in the ledger, so the leg never retries. Only a kernel restart lifts the switch. #1156 moved the sweep's call (main 10278) to a fresh load but left the walk's leg (main 10133) as it was. The 06:40Z after-numbers were read with the cache still on and report neither off nor poisoned. The diff adds a memo mechanism, which the review bar counts as a feature. Under the owner's role the label changed no gate.

- `kernel/kernel.py:10077`, blocker, merged. The first answered wake per kernel process files no nudge/awaiting row. The closer therefore never re-audits with the answer in view. The shared goal-store cache is off for the process from that moment. Fix PR: `_file_wake_answer` takes its own `jd.load_goals(sid)` at the write moment, which covers both callers. Add a test that drives the answered leg on a file-backed store through `load_goals_shared_or_fault` and asserts the row lands and `jd.shared_store_stats()["off"] == 0`.
- `tests/test_nudge_gate_memo.py:252` and `kernel/kernel.py:10858`, follow-up. The pin matches the due leg's `_fresh = jd.load_goals(sid)` and passes with the regression present. The comment and the body state that every writer reloads and that behaviour is unchanged. The body's claim that the nudge suites pass unchanged predates the fixture changes. Replace the pin with the behavioural test and correct the comment in the fix PR.
- `kernel/kernel.py:374` and `tests/test_perf_stats.py:117`, nit. Both still count three identity memos beside the fourth. Eviction at 10764 runs before insertion, so the memo holds 513 entries and the bound test pins that.

Locally at the head: the six PR test files, 118 passed; `tests/test_clear_wrap.py` plus `tests/test_nudge_injected_turn_arm.py`, 15 passed; those two plus `tests/test_nudge_gate_memo.py` and `tests/test_perf_stats.py`, 71 passed. Red-first at the base 19f2e1f0 with the PR's test files checked out: 14 failed, 104 passed. The 14 are 13 in `tests/test_nudge_gate_memo.py` on the missing `_nudge_gate_memo` and 1 in `tests/test_perf_stats.py` on the missing `nudgeGate` key. Added lines carry no private strings, hostnames or home paths. CI is green at 316b65d7. The four Python jobs were red at 6afb67f1 on `tests/test_nudge_injected_turn_arm.py`.

Review run by the maintainer automated review session; verified against head 316b65d7.

PR #1145, opened 2026-09-09
Scroll journal: a unit above the tail, or a box of the scroller outside the thread, names itself when it changes height (T262n)

henrypinkard:
An instrument, not a behaviour change: a unit above the transcript's tail now names itself when it changes height.

**Why.** The laptop capture read with the merged audit reader (T262k) held one unwritten move in eleven minutes of real use: the transcript shrank 24 px with the reader at the bottom and the browser carried them down, still at the bottom. No row named what shrank. The view rail's `tailchange` row fires whenever the view element's height changes and labels the change with the TAIL unit's class, so a unit above the tail changing height in place (a tool head folding, a figure sizing in, a status line going) is a row that names the wrong element, and the next unattributed `scrollHeight` change would have stayed unnamed.

**What.** Every unit in the rendered window is observed by one `ResizeObserver` per view. A non-tail unit's height change files a `unitchange` row: `{sid, dh, cls, fromTail, stick, atBottom, sh, ch}`, with `cls` the unit's class list, `fromTail` how many units above the tail it sits (1 = the unit just above), `stick` the view's recorded follow mode and `atBottom` the measured bottom at the read. Units join and leave the observer through the existing per-view `MutationObserver` (a window slide replaces them all); a unit's first observation is its baseline, never a row. The tail unit's own change stays the rail's `tailchange` row (the fold skips it); spacers keep their `spacer` rows. Same per-kind, per-minute cap as every other row. Nothing is written or decided from the row.

**Pinned** (`ui/webview/unit-change-row.test.ts`, executed): a non-tail unit growing by N files exactly one row with `dh` N naming the unit and its distance above the tail; the tail unit's own change files nothing here while the rail's `tailChangeRow` still carries it; a batch mixing both files the unit alone; spacers, units that left the window and a spacerless window file nothing; and render.ts wires the fold before the active/shown gate (baselines for hidden views too), observes added nodes and drops removed ones, files through the capped path, disconnects at both view-removal sites, and never writes from this observer. The four existing pins naming the row union and the import line are updated.

**The scroller's boxes outside the thread file the same row** (the manager's follow-up, folded in). The rail observes the thread element, so a height change INSIDE the thread always files a `tailchange` row (possibly mislabeled) and now a `unitchange` row too. The 7:38 PM move had NO `tailchange` row, so its 24 px came from outside the thread element, and that tab showed a session on a remote host, where the one box the dashboard adds under the thread is the host-offline foot (`#host-offline-foot`). `#content`'s direct children that are not a thread and not the live-ask host (the foot, `#sub-head`, the build placeholders) are now watched too: a box appearing files its height, a box changing in place files the delta, and a box leaving files the height it had, negative, since its removal under a bottom reader is exactly a clamp to the bottom by that height. These rows carry `fromTail: -1` and name the box by id else class. Baseline and observer both read `offsetHeight`, so the first observation never fakes a row. Pinned: `boxChanges`/`boxLabel` executed; the render wiring (exclusions, appear, leave, no writes) by source.

**Two guards from the adversarial review, so the rows stay about units.** A hidden view's units measure 0x0 on hide and their full height on re-show, and neither is a change: without a guard one switch back would have filed a row per unit and burnt the minute's cap, dropping the very row the instrument exists for. The hide now forgets the reported baselines and files nothing, and the re-show observation records fresh ones, like a first show (the same rule covers the whole pane hidden, for the boxes). A width change reflows every unit at once (a resize, a scrollbar appearing): the new heights become the baselines and nothing is filed. And `fromTail` counts units through the pane's `data-unit`, not children, so a day divider between a unit and the tail is not read as a turn. All three are pinned.

### DIFF
--- a/ui/webview/at-bottom.test.ts
+++ b/ui/webview/at-bottom.test.ts
@@ -85,9 +85,11 @@ test("follow mode and the chip read atBottom at every site", () => {
     /content\.clientHeight > 0 && !atBottom\(content\)\) \{\s*\n\s*writeScroll\(content, content\.scrollTop \+ \(h - lastH\), "box-resize"\);/,   // box-resize compensation
     /stick = !!content && atBottom\(content\);/,                                                          // tab-strip drag
     /const wasAtBottom = !!contentX && contentX\.scrollHeight > contentX\.clientHeight \+ 2 && atBottom\(contentX\);/,   // the ✕ on a pending bubble (T262h)
+    /unitChangeRow\(id, c\.dh, c\.cls, c\.fromTail, view3\.stick, atBottom\(content\), content\.scrollHeight, content\.clientHeight\)/,   // the unit-change row's measured bottom (T262n)
+    /unitChangeRow\(activeId \|\| "", dh, cls, BOX_FROM_TAIL, v\.stick, atBottom\(c\), c\.scrollHeight, c\.clientHeight\)/,   // the scroller's boxes outside the thread (T262n follow-up)
   ];
   for (const re of follow) assert.match(RENDER, re, String(re));
-  assert.equal((RENDER.match(/\batBottom\(/g) || []).length, 11, "ten call sites plus the definition");
+  assert.equal((RENDER.match(/\batBottom\(/g) || []).length, 13, "twelve call sites plus the definition");
 });
 
 test("only the user's own send reveal keeps the 80 px band", () => {
--- a/ui/webview/render.ts
+++ b/ui/webview/render.ts
@@ -59,7 +59,7 @@ import { hostNameNodes, hostPartsNodes, hostPrefix, hostOf, hostIsDown, hostDown
 import { followReader, keepPlaceAcrossShow, followTail, atBottomDist, followBoxBelow, followTailShrink } from "./scroll-keep";
 import { retainLiveOmitted } from "./tab-order";
 import { userTurnShows } from "./user-turn-content";
-import { ScrollDiagBudget, classifyScroll, scrollWriteRow, tailChangeRow, tailLabel, spacerRow, readScrollDiagCap, summarizeTailMutations, tailMutRow } from "./scroll-write";
+import { ScrollDiagBudget, classifyScroll, scrollWriteRow, tailChangeRow, tailLabel, spacerRow, readScrollDiagCap, summarizeTailMutations, tailMutRow, unitChangeRow, unitChanges, boxChanges, boxLabel, BOX_FROM_TAIL } from "./scroll-write";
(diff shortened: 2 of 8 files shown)


ksarma:
Auto-merge landed this PR at 2026-09-09 03:35Z (head a8a1c83a). Nothing found here would have held the merge. One correction: the label should read `feature`, not `fix`.

Reviewed inline: the per-view `ResizeObserver` over the thread's children, which units join and leave through the existing `MutationObserver`. The first observation is recorded as a baseline and files no row. The `unitChanges` fold skips the tail unit and spacers and drops a unit that has left the window. It counts `fromTail` through `data-unit`, so a day divider is not counted as a turn. The hide guard forgets baselines while the view is display none. The width guard rebaselines on a `clientWidth` change. Every unit is a full-width block child of the thread, so a width change reaches the observer in the same frame. `unitW` therefore cannot go stale between callbacks. The `#content` box observer excludes threads and the live-ask host. It files appear, change and leave rows at `fromTail` -1. Both view-removal sites disconnect the new observer. Neither observer writes `scrollTop`. The kernel's `clientDiag` handler is kind-agnostic, so the new row needs no kernel change. The added lines carry no private strings, hostnames or home paths.

- `ui/webview/scroll-write.ts:104` (follow-up, record): relabel the PR `feature`. The body describes an instrument with no behaviour change, but the diff adds a row kind, five exports and three observers. That is `feature` on the repo's tier list. The gate is the same under the owner's role, so the merge stands. Settle the tier for the T262 instrument series as well: #1083, #1106, #1109 and #1114 carry `fix`, #1115 carries `feature`.
- `ui/webview/render.ts:10004` (follow-up): a unit above the tail that changes height by being rebuilt files no `unitchange` row. The incremental render removes and re-appends every node from the first changed unit, so the rebuilt unit gets a fresh baseline. Only `tailmut` and a `tailchange` named by the tail record that case. The body scopes the claim to in-place changes, so this is a limitation to carry into the next capture.
- `ui/webview/render.ts:9830` (follow-up, nit): `View.uh` is written once and never read. The observer closes over `unitHeights` directly.

Tests: at the head, the six PR test files run 29 of 29 and the full `npm test` passes 3623 of 3623. At the base (a8a1c83a^), `unit-change-row.test.ts` fails to build on five missing exports. With it set aside, the five updated pin files (at-bottom, scroll-movers, scroll-write, tail-change-row, tail-mutation) fail 5 of 22 on `render.ts` source pins. The red is missing code rather than a wrong outcome, the pattern of a feature test. CI is green on all nine checks at a8a1c83a.

Review run by the maintainer automated review session; verified against head a8a1c83a.

PR #1146, opened 2026-09-09
T278: the socket budget bounds the backlog behind the frame in flight, so a first frame larger than the budget is delivered

henrypinkard:
The socket queue budget bounds the backlog behind the frame in flight, so a first frame larger than the budget is delivered instead of dropped.

Tier: fix

## Why
Between 7:30 and 8:00 PM PT on 2026-09-08 the devbox kernel dropped the laptop's timeline-bars socket hundreds of times (299 drop lines in the journal for the evening, 6 to 12 a minute, one window id), each with reason "0 bytes behind": `_mk_ws_send` dropped a client when queued bytes plus the frame exceeded `WS_QUEUE_BYTES`, and this machine's bars frame is 17.65 to 17.74 MB against a 16 MiB budget. Every reconnect was dropped on its first frame and reconnected, forever. The laptop's timeline pane never showed this machine's bars, and the kernel rebuilt and serialized a 17 MB frame several times a minute for nothing.

## What changed
`kernel/kernel.py`, `_mk_ws_send` and `_ws_sender`: "behind" is the bytes queued beyond the LARGEST frame in the client's queue: one frame may be arbitrarily big and is not the peer's fault, everything else waiting is the backlog. Frame sizes ride a deque beside the queue (`qsizes`), appended and enqueued as one step under the client's queue lock and popped as each frame completes; one sender thread and one queue make completions FIFO, so the deque mirrors the queue. The drop fires when the backlog exceeds the budget. Consequences:
- An empty queue accepts any frame, so a fresh connection is never dropped on its first frame, whatever its size.
- A healthy peer drains a big head at wire speed while the small deltas that follow queue behind it, well under the budget.
- A wedged peer's head never completes, the backlog behind it grows with every push and keepalive, and the client is dropped when that backlog passes the budget, as before. The liveness pings still catch a peer that answers nothing.
- The drop message names the backlog, not the big frame.
- The largest frame, not the head: a connect push enqueues the small lanes skeleton, then the 17 MB bars, then two small frames within microseconds, and whether the skeleton has left the queue by then is a scheduling accident. Measured against the largest queued frame, the small frames around a big one on a fresh connection can never read as backlog.

Chosen over sending the bars in parts: parts enqueued together add up to the same queued total, so the next frame would still have tripped a total-based budget; the budget itself was measuring the wrong thing. Shrinking the frame is left for a follow-up with a measured field breakdown (see Evidence).

## Tests
`tests/test_ws_send_bounded.py`, real loopback pairs like the rest of the file, with the budget patched to 64 KiB: a synthetic bars frame three times the budget (invented prompts under a placeholder sid) is delivered whole and byte-exact to a draining peer; a fresh connection is never dropped on its first frame even when the peer never reads; with a big frame stuck at the head, exactly nine 8 KiB frames are accepted behind it before the backlog passes the 64 KiB budget, and the drop names a count smaller than the big frame; the connect push's shape (small, big, small, small) is never dropped; a queue whose put asserts the client's lock pins that the size append and the enqueue are one step. `tests/test_ws_drop_loud.py`: the fixtures that sat 10 bytes under the budget now sit past it, and the delta-path case puts the backlog behind its held full past the budget, the rule those tests pin. The wedged-client, keepalive-burst, walk-starvation and byte-exact tests are unchanged and green.

## Evidence
Before, from the devbox journal (read-only): 299 `dropping timeline client` lines between 02:00 and 04:00 UTC on 2026-09-09, every one `0 bytes behind`, `queued=0B`, frame 17,649,511 to 17,740,614 bytes. The frame size is unchanged by this fix; what changes is that it is delivered once per connection and then followed by deltas, so the rebuild-and-drop loop ends. A measuring client cannot be attached to the running kernel, since it drops that client on the same first frame; the after-deploy measurement and a field breakdown for a shrink follow-up come in the report once this lands.

## Review (three lenses, two skeptics per finding)
Confirmed and folded: the queue put ran outside the client's queue lock, so two producers could interleave between the size append and the put and the sender's positional pop would credit the wrong frame (four skeptic confirmations); the accepted-count assertion was a window where the exact value is nine; a comment on the tag-edit acknowledgement still described a lone big reply as a budget drop. Refuted as pre-existing or timing-bound: a peer wedged mid-frame is judged by the ping path only once the ping reaches the socket (unchanged by this fix), the bell row's figure includes the big frame (true as stated), and the connect-time scenario above, which the largest-frame rule now rules out by construction rather than by loopback timing.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -4424,8 +4424,10 @@ def _judge_timeline_views(blob, base=None, seq_floor=0, edited=None, foreign=Non
         # blob missing its own creates. Rows only, never stored — the bound stands.
         # ROWS ARE BOUNDED to `bound` of them (the 2026-09-05 review): the rows, the loud
         # notice and the ack's `error` were O(N) in the posted array, so a 100k-entry post (a client
-        # bug, or any page holding the socket) drew a ~22 MB ack that overran WS_QUEUE_BYTES and
-        # dropped the poster's own socket before the ack was queued. Every unread store tag is still
+        # bug, or any page holding the socket) drew a ~22 MB ack, which under the budget rule of the
+        # time (queued bytes plus the frame against WS_QUEUE_BYTES) dropped the poster's own socket
+        # before the ack was queued; since T278 a lone big frame is delivered, and the bound stands
+        # for the pane's sake, a 22 MB ack being nothing a client should have to parse. Every unread store tag is still
         # kept, however far past the bound its copy sat (at most the store's 32); past the first
         # `bound` unread entries the rest are ONE summary row carrying their count and how many of
         # them `edited` names (`moreEdited`), which the door's ok rule reads — a client whose own
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -35542,6 +35542,15 @@ def _ws_accept(key):
 # stall being fixed here (caught by test_ws_send_bounded, which overran a 32-frame cap with 40 keepalives
 # a healthy client would have drained fine). Bytes track the thing that actually matters: a client behind
 # by 16 MB of view payloads has stopped reading, while any burst of ~40-byte keepalives is noise.
+# "Behind" is the bytes queued BEHIND the frame at the head of the queue, the one in flight or about to be
+# (T278, 2026-09-08): the budget bounds a peer that has stopped draining, not a frame that is simply big.
+# The devbox's timeline-bars frame grew to 17.7 MB against this budget, so every fresh timeline connection
+# was dropped on its FIRST frame ("0 bytes behind"), reconnected, and was dropped again — 299 drops in one
+# evening, a 17 MB frame rebuilt and serialized several times a minute for nothing, and a laptop that never
+# saw this machine's bars. An empty queue accepts any frame; a healthy peer drains the head at wire speed
+# while the small deltas queue behind it; a wedged peer's stuck head lets the backlog behind it grow past
+# the budget, which is the drop. The head is tracked as a size deque beside the queue (qsizes): one sender
+# thread, one queue, so completions are FIFO and the head of the deque is the head of the queue.
 WS_QUEUE_BYTES = int(os.environ.get("ROMP_WS_QUEUE_BYTES", str(16 * 1024 * 1024)))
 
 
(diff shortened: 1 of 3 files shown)


ksarma:
Merged at 04:47Z as the owner's fix on green. Nothing here would have held it. Three follow-ups remain, all wording or test coverage.

Reviewed inline: `_mk_ws_send` and `_ws_sender` in `kernel/kernel.py`. The budget check, byte count, size append and queue put now run as one step under the client's queue lock. That closes the interleaving filed at 49019df7. The frame excluded from the backlog is the largest queued frame. A small skeleton at the head with the bars frame behind it therefore no longer counts as backlog. The connect-shape test pins this. Substituting the head rule fails it. The bound still holds against a wedged peer, because only one frame is ever excluded. A small stuck head plus one very large later frame trips the drop once the small frames pass the budget. At a 64 KiB budget, nine 8 KiB frames are accepted and the tenth is dropped. The byte count and the size deque stay aligned on every path. The one production put is inside send. The sender's finally pops after each completed or failed write. The drop path raises before either write. The None teardown sentinel bypasses both. The drop string names the backlog. The bell row reports total queued bytes, as the body states.

Follow-ups:
- The PR title, the lead sentence, and `tests/test_ws_drop_loud.py:53-55` and `:164-165` still state the head-in-flight rule from 49019df7. `kernel/kernel.py:35629` and the body's What changed paragraph state the largest-queued-frame rule. The merge commit body carries the stale title. What remains is a retitle on GitHub and a reword of the two test comments.
- `kernel/kernel.py:35629`: the admission check leaves out the incoming frame. A wedged peer holding one over-budget full then accepts a second at zero behind and is dropped on the third. Peak queue is the budget plus about two largest frames, where the old check held a hard 16 MiB. One fix counts the incoming frame, `(qbytes + len(s)) - max(max(sizes), len(s))`, which moves the exact-count assertion from nine to eight. The other states the one-frame overshoot in the comment and body. Counting it drops a healthy non-delta client when two fulls overlap in its queue, so this is a design call.
- `tests/test_ws_send_bounded.py:271`: no test observes the sender's popleft. Replacing it with pass, or with pop(), leaves all 19 tests green. A bounded poll after the drain for qbytes 0 and an empty deque catches the first. A mid-drain check catches the second: the skeleton completes while the bars frame is queued, and the deque then holds only the bars size. Open since the earlier pass.

Tests: the PR's two test files against the merge base kernel fail 6 of 19. Four are the first-frame and largest-frame cases at 0 bytes behind. One is the lock case with the put outside the lock. One is the delta-path case on the missing qsizes key. At 4cd68b91 all 19 pass serially, under xdist and on three repeats. CI at 4cd68b91 is green on every required check.

Review run by the maintainer automated review session; verified against head 4cd68b91.

PR #1147, opened 2026-09-09
Billing picker on a one-auth box: both sides listed with the missing one greyed, and no launch ever lands on a missing login

henrypinkard:
## What

Two things the user asked for on a box that offers only one billing side (a headless box with an `apiKeyHelper` and no Claude login, or a laptop with a login and no helper):

1. **Nothing launches onto a side that does not exist.** `_auth_avail`'s default now falls back in both directions (a remembered login pick on a box with no login seeds new sessions on the key, exactly as a remembered key pick on a helper-less box already fell to the login). An explicit pick naming the missing side launches on the other side and says so once, in the log as a problem row and in status (`authPickUnavailable`); `set_auth` refuses the missing side with the reason. One vocabulary for every surface: `credentials.WHY_NO_LOGIN`, `WHY_NO_HELPER`, `WHY_MANAGED_HELPER`.
2. **The picker never disappears.** The tab menu's Billing submenu lists BOTH choices on every box; the one this box cannot bill is greyed (`.disabled`, `aria-disabled`), carries the reason in its hover, and is inert (no `setAuth` posted). The sub-line and the tab hover say `⚠ login unavailable, billing API key` when a pick fell to the other side. Status carries `authAvail: {login, key, loginWhy?, keyWhy?}`; `authBoth` stays for older clients and gates nothing.

Behavior when both sides exist is unchanged (pinned).

## Files
`kernel/credentials.py` (the three reason strings), `kernel/kernel.py` (`_auth_avail` both-way fallback + reasons, `_auth_avail_status`, status payload, setAuth toast names the reason), `kernel/sdk_backend.py` (`auth_unavailable_why`, `pick_unavailable`, `_options` fall-through with one notice per session, `set_auth` reasons, `_note_seed_skipped` per side), `ui/webview/render.ts` + `styles.css` (submenu, sub-line, hover, new-session picker hint), `docs/reference.md` (one-auth boxes), tests.

## Tests
- pytest `tests/test_session_auth.py`: default fallback both ways, stale-pick fall-through with a single notice, refusal reasons, status fields.
- node `ui/webview/billing-one-auth.test.ts` (new) + `auth-selector.test.ts` (source pins updated deliberately: availability gate, reason field per option, disabled class, the fell-through arm in both sub-lines).

Local: pytest full suite 9815 passed / 43 skipped with 14 failures that all reproduce on untouched main on this macOS box (15 there; no branch-only failure); webview typecheck, test, and build green.


ksarma:
Recommend merging as filed. Two follow-ups and two nits, nothing to hold for. This is the owner's feature, so it lands on green under the tier policy. This comment is an advisory record.

Reviewed inline: `_auth_avail` now falls both ways. A remembered login pick on a box with no login defaults to the key. A key pick on a helper-less box already fell to the login the same way. The reasons come from the three shared sentences in `kernel/credentials.py`. In `_options`, a pick that names a missing side launches on the side that exists. It says so once per SdkSession through `_pick_fell_said`. It sets `_launched_keyed` to the side taken, so `_note_auth_source` stays quiet on the expected landing. A pick with nothing to fall to launches plain as before. `spawn` drops a remembered default for either missing side, with one notice per side. `set_auth` refuses a missing side with the reason the `_drive` toast now names. The status payload carries `authAvail` and `authPickUnavailable`. `authBoth` stays for older clients. The Billing submenu renders for every SDK session and lists both options. It greys the missing one, with the reason in `title` and `aria-disabled`, and posts no `setAuth` for it. The flyout attaches to the transient menu popup the existing items already use, so it survives pushes. The new CSS uses the menu tokens: no hex, no new font size, no accent. The quiet boot and the absent helper-specific judge row from #1118 are untouched.

- `kernel/sdk_backend.py:6511`, follow-up: `snapshot()` computes `authPickUnavailable` by probing the box on every push, not from the launch's record. After a logout on the box, a running session launched on the login reads "login unavailable, billing API key". Its process still holds the credentials it launched with. Latch the fall in `_options` beside `_launched_keyed`, and report that for the live row. Keep the probe for the dormant row. `test_the_status_rows_carry_the_unavailable_pick_live_and_dormant` pins the probe and would change.
- `kernel/kernel.py:16117`, follow-up: `Sessions.live` merges `authLive` for the first time. On main every reader fell to the pick. So the cap-switch offer, the limit banner and the "CLI reports" hover now change for a session whose CLI landed opposite its pick. The merge is correct, and the fall-through needs it. The body and commit messages do not say so. The line that both-sides behavior is unchanged does not hold for that case. Name it in the body. Cover the merge with a behavior test through the real `Sessions.live` rather than the source pin.
- `docs/reference.md:319`, nit: the sub-line is on the tab menu's Billing item, not a card. The sentence "nothing ever launches onto a login that is not signed in" overstates it, since `_options` falls only when a helper exists. The tab hover renders a different sentence from the sub-line, not the words the body gives.
- `ui/webview/auth-selector.test.ts:72`, nit: the retained comment states the pre-PR both-or-nothing rule directly above the assertion the PR rewrote.

Locally at 8604149f: `tests/test_session_auth.py` 55 passed; vscode-extension npm test 3620 passed with typecheck clean; the two UI files 12 of 12. Red-first is not required at this tier. Still, the PR's three test files fail as expected at origin/main (16 pytest cases, 4 of 4 and 3 of 8 UI pins) and pass at the head. Privacy and vocabulary scans of the added lines are clean. CI is green at the head. The merge state shows blocked for a reason outside the checks.

Review run by the maintainer automated review session; verified against head 8604149f.

ksarma:
Merged at 0fd3b70f, after the review-fold commit 1cae822b. Nothing in that commit would have held the merge. This comment records the fold against the earlier review at 8604149f.

Reviewed inline: the fold's four claims against the code at 0fd3b70f. First, a settings file that cannot be read answers `key_state` unknown and is reported once per process. A picked launch keeps its pick and says so once per session. Confirmed for picked sessions, not for an unpicked one (below). Second, the account reader carries ok, none and unreadable, and never caches a failed read. The kernel's probe hands the backend None, and `pick_fall` never keys a fall on it. Confirmed for the launch and for the login side of the picker. The key side of the same kernel function still collapses unknown to no helper. Third, `pick_fall` is the one decision `_options` launches from. The backend's snapshot and dormant rows carry `authPickFell`, but `Sessions.live` and `build_session` do not forward it, so both webview surfaces still infer the fall. Fourth, the per-cycle memo is a `threading.local` slot, set inside the pusher cycle's try and cleared in its finally. A handler-thread push or a test therefore computes fresh, and no availability crosses a cycle. Only the status payload and the picker reply read it; no card column keys on it. Of the earlier record, the docs overstatement is fixed and the other three items stand. The PR body was not updated after the fold. The two settled #1118 rulings are untouched.

Follow-ups:
- `kernel/kernel.py:16298` and `:30488`: `Sessions.live` and `build_session` forward `authPickUnavailable` but not `authPickFell`. `render.ts` `authFellTo` therefore always infers the fall from `authAvail`. The two diverge for a key pick on a helper-less box while `~/.claude.json` is unreadable. There the launch goes out as picked, and the hover and sub-line say it bills the login. `docs/reference.md:321` says the status payload carries the field; it does not. The fix is two one-line merges plus a pin in the payload test.
- `kernel/kernel.py:14257`: `_auth_avail` reads the key side through `key_available`, which is False on `key_state` unknown. An unreadable settings file therefore reports the no-helper reason and flips a remembered key default to login. Spawn's seed keeps the key pick. The picker says Login; the session launches on the key. The login side of the same function got the cannot-tell clause. The key side needs the same clause and a kernel-side test.
- `kernel/sdk_backend.py:8654`: an unpicked launch or reconnect during a settings rewrite reads `keyed_box` False and restores the login bearer. A helper box then bills the login for that process with no per-session notice. This is pre-existing at main. The fold's rule (never moves a launch) and the row at 7168 do not cover it.
- `kernel/sdk_backend.py:6704-6705` (earlier follow-up, not addressed): `snapshot()` still probes the box per push for both fields instead of reading the launch record. A logout after a login launch therefore flips a running row to billing API key. `render.ts:273` calls the field the launch's own decision.
- `kernel/kernel.py:16295` (earlier follow-up, not addressed): the `Sessions.live` `authLive` merge changes the cap-switch offer and limit banner for an opposite landing on any box. The PR body still says both-sides behavior is unchanged and names none of the fold's additions.
- `kernel/sdk_backend.py:10919`: `pick_fall` re-runs `pick_unavailable`, so a picked row costs 4 to 6 settings-file reads per push (2 before the fold). One probe per row, feeding both fields, removes the extra reads.

Nits: `docs/reference.md:320` places the sub-line on a card; it is the tab menu's Billing item. `ui/webview/auth-selector.test.ts:72` still describes the both-only gate above the pin the PR changed.

Tests at head: `tests/test_session_auth.py` 71 passed; auth-selector and billing-one-auth 12 passed; typecheck clean. The 16 new Python tests and one test in each UI file fail at 8604149f and pass at head. CI was green at 0fd3b70f.

Review run by the maintainer automated review session; verified against head 0fd3b70f.

PR #1148, opened 2026-09-09
Kernel: the cut-turn reaper stops only the CLI's own scope, never one it inherited (T276 review fold)

henrypinkard:
Follow-up from the adversarial review of #1139, which found a blocker.

`_end_cli_tree` read the CLI's `/proc/<pid>/cgroup` and stopped whatever `romp-session-*` scope it named. A CLI in its own scope always carries its pid in the unit name (`bin/romp-cli-scope` names the unit `romp-session-<sid8>-$$-$t` and execs into it); a CLI whose cgroup names a scope with another pid merely **inherited** it, as every process does that a romp session's tool shell starts, a kernel launched there and its CLIs included. Stopping that unit ended the launching session, not the orphan. The branch's own real-process test did exactly that to the session running the suite.

- **Fix.** The reaper requires the CLI's own pid in the unit name and otherwise treats the CLI as unscoped (tree walk only), logging the inherited unit. Red-first: an inherited scope is never stopped and only the CLI is signaled.
- **Also from the review.** A nonzero `systemctl stop` is no longer reported as a stopped scope (return code read, stderr's first line logged). Each target's start time is captured at the `ps` snapshot and a pid whose identity changed during the grace is skipped. The real-process test pins the cgroup seam empty (its docstring no longer claims `ROMP_CLI_SCOPE=0` removes the inherited cgroup), and the reconcile-level test pins fake pids as gone on every platform.
- **Tests.** `tests/test_cut_turn_tree_kill.py` gains the inherited-scope and failed-stop cases; the existing tree-kill, reaper and scope pins hold (134 passed together).

### DIFF
--- a/kernel/sdk_backend.py
+++ b/kernel/sdk_backend.py
@@ -2729,6 +2729,17 @@ def _read_cgroup(pid: int) -> str:
         return ""
 
 
+def _read_starttime(pid: int) -> int | None:
+    """The process start time (clock ticks since boot, /proc/<pid>/stat field 22) — a pid's identity across a
+    reuse; None where it cannot be read (no procfs, the pid gone)."""
+    try:
+        with open("/proc/%d/stat" % pid) as f:
+            tail = f.read().rsplit(")", 1)[1].split()
+        return int(tail[19])
+    except (OSError, IndexError, ValueError):
+        return None
+
+
 def _read_ppid(pid: int) -> int | None:
     """The parent pid from /proc/<pid>/stat, or None where it cannot be read."""
     try:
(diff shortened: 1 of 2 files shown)


ksarma:
Landed as shipped (43095ab2) before this review finished. Nothing below would have held it. Every finding is a follow-up.

Reviewed inline: the ownership check at `kernel/sdk_backend.py:7156` stops a scope only when the unit name carries the CLI's own pid. `bin/romp-cli-scope` names the unit that way (`romp-session-<sid8>-$$-<t>`) before it execs into it, so an inherited scope now gets the tree walk alone. The systemctl return code is read at 7163 and the first line of stderr logged. A failed stop no longer reports a scope. The start-time identity check at 7177-7180 runs before both the killpg and the kill branch. `RealProcessTree` passes `cgroup=lambda p: ""` and reads no real cgroup. Its `run` seam stays the real `subprocess.run` but is unreachable, because `scope_unit_of("")` is `None`. The full pytest suite at a head containing this commit reads as safe to run from inside a romp session. The maintainer tooling's ignore of `tests/test_cut_turn_tree_kill.py` can be lifted now that it is on main. Run that file alone under a timeout first. The diff does not touch `_stop_leftover_scopes`.

Follow-ups:
- `kernel/sdk_backend.py:7156` and `:7226`: neither stop site refuses the scope the kernel itself runs in. A kernel hand-started from a session's tool shell against the live state dir passes the pid check when that session's CLI is the orphan. The sweep's keep rule spares only this kernel's children. At either site, `systemctl stop` then ends the kernel with the session. The #1139 review asked for the self-scope guard on both paths. Comparing each unit against `scope_unit_of(_read_cgroup(os.getpid()))` at both sites closes it.
- `kernel/sdk_backend.py:7177`: the start-time baseline is captured inside `_end_cli_tree`, after the scope stop and once per orphan. `_boot_reconcile` reads `ps` once at 7265. A pid recycled during an earlier orphan's reap (up to 16 s), or during this orphan's stop, is therefore recorded with the new process's start time and signaled. `same()` also returns `True` on a `None` baseline. On Linux a `None` baseline means the pid was already gone, so a pid reused inside the 1 s grace gets SIGKILL. The PR body's sentence about capture at the `ps` snapshot overstates this. Fix: capture beside the `ps` read, pass the map in, treat `None` as skip where `/proc` exists, and pin it with a test.
- `kernel/sdk_backend.py:7230`: the sweep still counts a nonzero `systemctl stop` as stopped, so the boot summary overcounts.

Nits: `tests/test_cut_turn_tree_kill.py:260` could pass a recording `run=` and assert it stays empty. The no-systemctl property then becomes an assertion instead of a consequence of the empty cgroup. The inherited-scope fixture's sid8 at `:135` does not read as a placeholder next to the file's `11111111` style (it matches no private string).

Still open on main from #1139 and not claimed here:
- `os.getpgid`, `os.killpg` and the procfs liveness check outside the kill seam;
- the boot sweep stopping every alive session's still-loaded scope, a design call for both maintainers;
- the sid8 filter against the full `ROMP_SID`;
- the keep rule sparing only this kernel's children.

Tests: nothing from this PR ran locally. The reaper reads cgroups and stops scopes, and this review session runs inside one. Red-first by reading: at the parent 27868fe0 the stop path sets `stopped = True` with no pid check. `test_an_inherited_scope_is_never_stopped_only_the_clis_own` and `test_a_failed_stop_is_not_reported_as_stopped` fail there and pass at the head. CI at cd69f5d4 is green on every check, 9824 passed with the file included. The PR body's 134 matches the three files' test counts (80 + 43 + 11). No private-string, hostname, home-path or vocabulary hits on the added lines.

Review run by the maintainer automated review session; verified against head cd69f5d4.

PR #1149, opened 2026-09-09
Chat pane: a re-show lands a reader at the true bottom at the bottom, whatever the recorded follow flag says (T262)

amith-ananthram:
## The snap

T262: the chat snapped **up** the moment the reader scrolled to the bottom, in every column, 300 to 1900 px at a time. The scroll journal (`client-diag.jsonl`) named the writer on every occurrence: `land-saved`, with the reader at the true bottom (`scrollTop + clientHeight == scrollHeight` on each row) and the recorded follow flag **off**.

## Why the flag was off at the bottom

`followReader` deliberately records nothing while a deferred build is pending (the reveal's clamp fires a scroll event before the build lands, review fold 2026-09-08). A wheel gesture that ends on the bottom inside that window is therefore not recorded either: `stick` stays false and the saved `scrollTop` stays a screen above. The next full show of the view (`landActive`'s closing rule) lands on that stale spot.

## Fix

At a re-show of the view already on screen, the DOM is the truth: `showActive` re-derives follow mode from the true bottom before the rebuild, through a pure rule (`reshowStick(recorded, atBottomNow) = recorded || atBottomNow`). It only ever turns following **on**; off the bottom the recorded flag stands, so a scrolled-up reader keeps their place through the rebuild exactly as before (the T249 anchor path is untouched).

## Tests

- `scroll-keep-on-show.test.ts`: the journal's case executed (recorded off, at the true bottom, a re-show lands at the bottom, never the stale spot), the rule's four quadrants, and the `render.ts` wiring pinned between the re-show decision and the keep-anchor capture.
- The import-list pins (`follow-tail`, `scroll-keep-on-show`) and the `atBottom` call-site count (`at-bottom.test.ts`) grow by one, with the reason.
- The served scroll tests that drive the real page (scroll marks, pending tail, queued copies, awaiting box) pass unchanged.

Tier: **fix** (labelled).


amith-ananthram:
Reopened, rebased onto current main (17f5251e), and labelled `fix`. In plain terms, what this fixes:

**The symptom.** On a busy session, scroll to the bottom of the chat and it jumps back up a few hundred pixels a moment later, most visibly when you switch back to a tab or when the pane re-renders (a settings change, a split column re-showing).

**Why.** The pane records whether you're "following" the bottom from your scroll gestures, but it deliberately records nothing while a rebuild is pending, because the rebuild's own reveal fires a scroll event that isn't yours. On an active session a rebuild is pending much of the time, so a wheel gesture that ends at the bottom during that window leaves the follow flag off and the saved position a screen above. The next time that view is shown, it lands on the stale saved spot instead of the bottom.

**The fix.** When a view already on screen is re-shown, it reads the DOM: if the reader is at the true bottom right now, follow mode turns on regardless of what was recorded. It only ever turns following on, so a reader who scrolled up keeps their place exactly as before. One pure rule (`reshowStick`), executed in `scroll-keep-on-show.test.ts` with the journal's exact case, plus the wiring pinned at source.

Still present on main as of today: the re-show landing and the pending-build guard are unchanged, and the mechanism reproduces from the scroll journal rows in the PR body.

amith-ananthram:
TLDR: On a busy session, scrolling to the bottom while a transcript repaint was queued didn't get recorded, so the next repaint of the view you were looking at moved you back up to the previously saved spot. This fix makes a repaint of an on-screen view check where the reader actually is: at the true bottom, they stay following the bottom as romp always intended (I believe this is the convention); scrolled up, they keep their place exactly as before.

PR #1150, opened 2026-09-09
Kernel: a safeguards refusal the CLI retried on a fallback model is filed and shown, not logged as unhandled

henrypinkard:
## What

A new Claude Code system event, `model_refusal_fallback` (the model's safeguards declined the request and the CLI retried the turn on the configured fallback model), reached the SDK backend and fell to the once-per-life "unhandled SystemMessage subtype" log line. Nothing was filed for the judges: their only record of the swap was the capacity-fallback card the fallback reply's own model learn minted, with the wrong cause ("an API-side capacity fallback"). The chat notice (from the transcript record) dropped the refusal category and the API's explanation.

Now the frame is handled at every hop:

- **Stream (kernel/sdk_backend.py):** a branch for the subtype, placed before the unhandled memo, files the frame through a kernel-wired class-level hook (the `on_model_fallback` idiom) with pretty model names, the refusal category, the API's explanation and the scope. The unhandled line no longer fires for it. A failing hook is a logged problem, never a raise into the stream loop.
- **Judges (kernel/judge.py):** `mint_refusal_fallback_card` files a completed card naming the refusal: the category in the title, the explanation in the rationale, with its own why key and the swap as data. The fallback model's reply streams before the CLI's end-of-turn notice, and its model learn has already filed the swap as a capacity fallback, so the backend keeps the ids of the capacity cards its learn minted this turn (forgotten at the settle) and the filing folds the ones on the chain from the refused model to the answering one into the fresh refusal card with the store's existing merge (`_merge_nodes`): one swap, one card, and no field of an existing node is rewritten, so a concurrent store writer cannot half-revert it (the refusal node is adopted wholesale by a rebase; each folded card's deletion is a durable tombstone). An earlier hop's refusal card of the same episode (keyed by the refused prompt's uuid) folds the same way; a `provisional` frame (an intermediate hop of a multi-hop chain) files like any other, since on the CLI's chain path no final frame need follow it; the next hop's frame folds its card by episode. A card the user cleared, followed up on, or otherwise stamped is left alone and the refusal files fresh beside it. `mint_fallback_card` stands down when a refusal card for the swap is already on the board. The rollup runs with `session_closed=False`, so the filing never forces the settle of the session's focus card. A frame without model ids is a logged problem, not a card; an unwired hook is said once per kernel life.
- **Transcript → chat event (kernel/event_model.py, kernel/kernel.py):** the atom and the `modelFallback` event carry `category`, `explanation` and `scope`; the kernel wires the hook at boot.
- **Pane (ui/webview/render.ts, styles.css):** the notice is a sourced notice card in the shared notice-card grammar: a "safeguards" chip in the heads-up amber, a head naming whose safeguards flagged the message, the category, and which model answered ("switched to X", or "this reply came from X" for a local scope), and a keyed fold holding the API's explanation and the CLI's own line verbatim. Never the user's bubble. The old slim rail line and its private CSS are gone.

## Field names: the CLI's own schema, not one sample

The public SDK docs do not yet document this subtype. The field list comes from the stream-json schema embedded in the installed CLI (bundled CLI 2.1.259, matching the system CLI 2.1.257): `original_model`, `fallback_model`, `request_id`, `api_refusal_category` (open string; null when neither lane carried one, described as normal), `api_refusal_explanation` (display-only prose; null on server-lane banners), `scope` (`session` | `local`; absent on older CLIs, to be read as `session`), `retracted_message_uuids`, `refused_user_message_uuid`, `direction`, `trigger`, `content`. The transcript record carries the same fields in camelCase (`apiRefusalCategory`, `apiRefusalExplanation`, `originalModel`, `fallbackModel`, `scope`). The sibling `model_refusal_no_fallback` is left as-is: it is the blocked API-error path, already surfaced by the error classification.

## Tests (red first)

- `tests/test_refusal_fallback_notice.py`: the stream branch calls the hook with the normalised fields (null → "", absent scope → session, local rides through and claims no card), the learn captures the capacity card's id and the settle forgets it, every card of the turn rides the hook, a provisional frame files too (the episode fold collapses hops), the unhandled line no longer fires, an unwired hook is said once, a frame without model ids is a logged problem, a failing hook is a logged problem, the branch order is pinned; the card mints completed with the category and explanation, dedupes, folds the same turn's capacity card (and a multi-hop chain of them, and an earlier hop of the same episode), leaves a followed-up or cleared capacity card alone, files fresh on an unknown id, stands the capacity mint down after a refusal card; the kernel wiring is pinned.
- `tests/test_model_fallback_notice.py`: the atom and the chat event carry the new fields; the notice text is never a user event.
- `tests/test_refusal_notice_served.py`: the served chat page against a hermetic kernel renders one refusal notice card with the safeguards chip, a head naming both models and the category (the full head as the hover), the collapsed fold restating the swap and holding the explanation and the CLI line, the notice's rail dot, before the fallback model's reply, and no user bubble carrying the text; a head click opens the fold.
- `ui/webview/model-fallback.test.ts`: the notice-card pins (variant, chip, head fields, fold keys, CSS variant; the slim line gone).

Synthetic fixtures only (placeholder ids, an invented category and explanation, the notes-api demo world).

Tier: fix

### DIFF
--- a/docs/event-model.md
+++ b/docs/event-model.md
@@ -140,11 +140,16 @@ interface Trigger {
 interface Atom {
   // Parity with the streaming API:
   type: "assistant" | "user" | "system" | "result" | "idle";   // "idle" is ours
-  subtype?: "compact_boundary" | "status" | "task_notification"; // when type==="system"
+  subtype?: "compact_boundary" | "status" | "task_notification"
+          | "model_refusal_fallback";                          // when type==="system"
   uuid: string;             // message id (same value in stream and transcript)
   session_id: string;       // = rompUuid
   message?: ApiMessage;     // assistant/user: the Anthropic message object (below)
   compact_metadata?: { trigger: "auto" | "manual"; pre_tokens: number }; // system:compact_boundary
+  content?: string; fallback_from?: string; fallback_to?: string;   // system:model_refusal_fallback — the CLI's
+  refusal_category?: string; refusal_explanation?: string;          //   line, the swap, the refusal's category and
+  scope?: "session" | "local";                                      //   the API's explanation ("" when the record
+                                                                    //   carried none), the scope (absent = session)
   result?: {                // type==="result"
     subtype: "success" | "error_during_execution" | "error_max_turns" | string;
     num_turns: number; stop_reason: string | null;
--- a/kernel/event_model.py
+++ b/kernel/event_model.py
@@ -2148,6 +2148,13 @@ class FileAdapter:
                        "content": r.get("content") or "",          # the CLI's full explanation
                        "fallback_from": r.get("originalModel") or "",
                        "fallback_to": r.get("fallbackModel") or "",
+                       # T279: the refusal category (an open string; null when neither lane carried one),
+                       # the API's explanation (display-only prose; null on server-lane banners) and the
+                       # scope ('session' = the session model is swapped; 'local' = a subagent's or a
+                       # side question's reply only; absent on older CLIs = session)
+                       "refusal_category": r.get("apiRefusalCategory") or "",
+                       "refusal_explanation": r.get("apiRefusalExplanation") or "",
+                       "scope": r.get("scope") or "session",
                        "_seq": seq}
             # other system subtypes (turn_duration, stop_hook_summary, local_command,
             # away_summary) are harness bookkeeping, not conversational messages -> skipped.
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -12301,6 +12301,13 @@ def mint_fallback_card(sid, from_model, to_model, ev_t=None):
                     and prev.get("text") == text and not prev.get("cleared") \
                     and prev.get("id") not in vc:
                 return None
+            # T279: the same swap is already on the board WITH its cause — a safeguards refusal the CLI
+            # retried on the fallback (a CLI that streams that notice ahead of the reply); the capacity
+            # reading of the down-tier transition stands down rather than filing a second card.
+            if prev.get("why") == REFUSAL_FALLBACK_WHY and not prev.get("cleared") \
+                    and prev.get("id") not in vc \
+                    and prev.get("swap") == {"from": from_model or "?", "to": to_model or "?"}:
+                return None
         n = store.get("seq", 0) + 1
         store["seq"] = n
         gid = "%s:g%d" % (sid, n)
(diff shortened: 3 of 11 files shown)


ksarma:
Merged as shipped at ed55a7d4 (the owner's auto-merge, 2026-09-09 05:06Z). Nothing found would have held the merge. Five follow-ups stand.

One record note: the change is a feature, not a fix. It adds a kernel-wired hook, a card kind with its own why key and two node fields, a chain fold with three helpers and per-turn backend state, new atom and event fields, and a notice-card variant that replaces the slim rail line. Under the owner's role a feature merges on green too, so the label changed no gate.

Reviewed inline: The stream branch sits before the unhandled memo. It reads a null category or explanation as empty and an absent scope as session, pinned by test. `_chain_cards` terminates on any input, including a cyclic pair, since each pass grows a finite set or stops. No cycle can arise among capacity cards, because the learn fires only on a down-tier transition. A card off the origin-to-target path is never claimed. `_claimable` leaves a cleared, view-cleared or followed-up card alone. The title parse in `_swap_of` runs only on ids the same turn's learn returned, and only under the capacity why key, so it cannot misfire on another card kind. The fold is keyed on the notice frame. The capacity mint stands down on an uncleared refusal card with the same swap. So for a single hop both stream orders end in one card, and nothing is re-derived per build. The notice uses the shared notice-card grammar with a keyed fold, no new font size and the warn color. Goal-store tests use a private synthetic sid and clean its override journal. The added lines carry no private strings, hostnames or home paths.

- `kernel/judge.py:12458`, follow-up: a multi-hop episode delivered as per-hop frames leaves the second hop's capacity card standing under the capacity why, in either stream order. That per-hop shape is what the PR's provisional handling and test comments assume. The chain is rebuilt from nodes still present, so a card the first frame folded contributes no edge. The capacity mint's stand-down matches an exact swap only. Seed the chain with the folded refusal cards' swaps, make the stand-down chain-aware, and add a test that sequences frame, learn, frame.
- `kernel/judge.py:12309`, follow-up: a local-scope refusal card carries swap too. A later session capacity fallback of the same pair then mints nothing, and the only card on the board says the session model is unchanged. The previous head's title match excluded local cards. Store scope on the node and skip local cards here, with a local-scope variant of the stand-down test.
- `kernel/judge.py:12325`, follow-up, pre-existing: `mint_fallback_card` still runs `rollup_status(store, True)`. That settles a done-but-focus top on a mid-turn swap that carries no information about it. The new sibling passes False for this reason. One token plus a test.
- `kernel/sdk_backend.py:4220`, follow-up: `_reconcile_stranded` resets the other per-turn state but not `_swap_cards`. On the mid-turn reconnect path, an abandoned turn's capacity card can therefore be folded by the next turn's refusal. Add the reset beside the others.
- Commit message, follow-up, record only: it describes the superseded in-place reclassification. The merged code mints a new node and folds the capacity card through `_merge_nodes`, so the capacity card's id does not survive. Say so in the next commit that touches this code.
- `kernel/judge.py:12369`, nit: the title-parse branch of `_swap_of` is unreachable. Every card `capacity_gids` can name was minted this kernel life with swap written. Drop it or reword the docstring.

Red-first holds. With the PR's four test files copied into a scratch worktree at ed55a7d4^, 29 of 34 pytest cases fail: no `mint_refusal_fallback_card` or `REFUSAL_FALLBACK_WHY`, the hook never called, the unhandled memo firing, the new atom and event fields missing. The served test fails with no refusal notice card rendered. The extension suite fails 6 of 3635 tests, all in `model-fallback.test.ts`. At the head, 34 passed, the served test passes against a hermetic kernel, and 3635 of 3635 extension tests pass. CI at ed55a7d4 was green on every required check, Tier policy included.

Review run by the maintainer automated review session; verified against head ed55a7d4.

PR #1151, opened 2026-09-09
T278b: the bars frame carries each prompt's first line and none of the three duplicate ids; the view derives them

henrypinkard:
The timeline bars frame carries each prompt's first line, capped, and none of the three duplicate id fields; the view derives them.

Tier: feature

## Why
After the queue budget fix, the devbox's bars frame is delivered, but it is 20.4 MB (25 lanes, 8,562 bars) and takes 22 s to reach a client over the laptop tunnel. Measured live: the prompt field is 8.2 MB (40%), and seven uuid-shaped fields repeated per bar are 2.9 MB, of which three are exact duplicates in every bar.

## What changed
**Kernel** (`kernel/kernel.py`):
- `_wire_prompt(text)`: the FIRST non-empty line of the prompt after the view's own three passes in the view's order: romp's `<!-- romp-… -->` marks removed, a leading `[romp]` label removed (anchored at the text's start, so a label alone on line one does not become the line the tip then strips to nothing), a repeat storm collapsed over the whole text (`retry ×14`, so the count is never truncated); then cut at 400 characters with an ellipsis. The pane's only reader of the prompt is the prompt-dot tip, which shows at most 90 characters of that first line; the full text is a click away, since the dot opens the chat at the prompt. There is no per-bar detail route, and the view never fetches on demand: the tip is the only reader.
- The bar literal drops `tid` (always the lane key it sits under), `uuid` (always `promptId`) and `workUuid` (always `workId`); all three were duplicates by construction, confirmed live on 8,562 bars. Per-segment index tables were considered and measured: no bar shared a segment id (no sleep-split continuations in the live frame), so an index would have added bytes; the remaining four ids are unique per bar and stay.
- The full prompts stay beside the wire in a builder-local map handed to `_bind_message_execs`, whose postal sender heuristic must see past the first line (a delivery's first line is the postal header). A lane served from the dead-lane memo carries its full prompts in the memo, so a lane that died within an hour of a message it received keeps that binding on later builds.
- The delta path is unchanged: bars are keyed by (lane, id) and `id` stays on the wire, so a delta after a full decodes against the same keys; the shim reads no bar field.

**View** (`ui/romp-timeline-view.js`): `workAnchorOf` reads `replyUuid`, `workId`, `promptId`; the focus anchor object carries bar-shaped keys; the dot click, the bar click, the focus landing and the message navigation open the chat on the lane key and `promptId`; the active-chat lane lookup matches the lane key directly; the two fork-lane `tid` fallbacks were identities by construction and are gone. `reqText` is exported for the round-trip test.

## Field-by-field audit (every consumer of the bars payload)
Readers found: the view (`ui/romp-timeline-view.js`), the kernel's own `_bind_message_execs`, the view-delta shim inlined in the kernel, `ui/webview/federation.ts`, and the node and Python test fixtures. No served test asserts a bar field; the Obsidian host runs this same view file.

| Field | Readers | Verdict |
|---|---|---|
| `id` | view `dotLit`/`barLit`, focus match, hover emit; kernel delta keyer; shim positional keys | keep |
| `promptId` | view `dotLit`; now the source for the dot anchor and `workAnchorOf`'s third rung | keep |
| `workId` | view `barLit`; now `workAnchorOf`'s second rung | keep |
| `replyUuid` | view `workAnchorOf` first rung, focus anchor, message navigation | keep |
| `start`, `end` | view layout, hover, click, focus; kernel binder; federation time rebase | keep |
| `open` | view `barEndT` | keep |
| `cont` | view (suppresses the prompt dot on a continuation piece) | keep |
| `prompt` | view `req`/`barBody` through `reqText` only; kernel binder text fallback | first line, capped; binder gets the full text |
| `summary` | view `barBody` | keep |
| `msgCaption` | view `req` | keep |
| `src` | view focus anchor (prompt vs work landing) | keep |
| `mids` | view message-to-bar join; kernel binder exact join | keep |
| `pending` | view `startAt`, live riders | keep |
| `nudgeAuto`, `romp` | view dot styling and caption | keep |
| `tid` | view lane lookups (all fell back to the lane key), focus anchor, clicks; federation prefix rewrite (guarded, pass-through without it) | dropped; the lane key |
| `uuid` | view `workAnchorOf` third rung, focus anchor, dot click, message navigation | dropped; `promptId` |
| `workUuid` | view `workAnchorOf` second rung, focus anchor | dropped; `workId` |

Same-named reads that are not bars and are untouched: tag edits' `tid`, comment threads' `tid`, messages' `uuid`, judging marks' `open`/`start`/`end`, tag groups' `pending`. `ui/webview/federation.ts` still rewrites a bar's `tid` when present; kernel bars no longer carry one, the lane key it prefixes is what the view derives, and the multi-kernel merge tests keep passing on their own fixtures. Eight node fixtures still set the three dropped fields with values the derivations never equal; nothing asserts on them.

## Tests
- `tests/test_timeline_bars_wire.py`: the shaping (first line, romp marks, the cap and its marker, the storm count over the whole text, empties); the bar literal (the wire prompt call, no dropped keys, every kept key); the binder reads the full text past the first line and documents the memo fallback; a synthetic 25-lane, 8,000-bar payload of the devbox's size class encodes to under 65% of its old size; the committed fixture `tests/fixtures/timeline-bars-wire.json` is the encoder's output.
- `ui/timeline-bars-wire.test.ts`: the view's half of the round trip. For every fixture pair, the tip text of the wire bar equals the old bar's, the work anchor equals the old chain, `promptId` equals the old `uuid`, `workId` the old `workUuid`, the lane key the old `tid`, and hover matching still lights on the fields that ride.
- `ui/timeline-view.test.ts` and `tests/test_kernel.py` follow the new field names; `tests/test_ws_send_bounded.py`'s synthetic bars match the wire shape.
- Extension suite: 3637 passed. Kernel modules touched: 563 passed. Full pytest and the adversarial review run before this arms.

## Measurement
Before, live on the devbox (read-only client): 20,384,374 bytes, 25 lanes, 8,562 bars; prompt 8.24 MB; the three duplicates 1.16 MB; judging 3.97 MB; messages 0.48 MB; delta key list 2.37 MB. After: measured once this deploys, reported to the manager. The synthetic ratio test bounds the new frame under 65% of the old on the same shape; on the live distribution (median prompt 481 characters, first lines far shorter) the prompt field should fall by roughly 80%.

## Review (three lenses, two skeptics per finding)
Folded in the second commit: a label-only first line would have blanked the tip (the view stripped the leading label from the whole text before picking a line; the kernel now does the same, in the view's order); the dead-lane memo had no age gate, so a lane that died within the binder's one-hour lag lost its full prompts on the first memo hit and the sender binding flapped to the wire line (the memo now carries the prompts); `reqText` was already exported, the duplicate key is gone; the literal pin fails with a message instead of an index error.


ksarma:
Already merged: head caafb7b2 auto-merged on green 2026-09-09 05:51Z. Nothing found would have held it. The items below are follow-ups.

Reviewed inline: `_wire_prompt` applies the view's three passes in the view's order. It removes romp marks, removes a leading label anchored at the text start, and collapses the repeat storm over the whole text. It then takes the first non-empty line and cuts it at 400 characters. The view's own 90-character tip cut lies inside that, so the fixture round trip is an identity. The three dropped fields were duplicates by construction. `bars.append` is the only bar constructor, under `turns[sid]` with the same sid that filled `tid`. `uuid` and `workUuid` held the same values as `promptId` and `workId`. No bar reader in `ui/romp-timeline-view.js` or `ui/webview` still reads them. `workAnchorOf`, the focus anchor, dot and bar clicks, message navigation and the active-chat match now read `promptId`, `workId` and the lane key. The two removed fork-lane `tid` fallbacks were unreachable behind the direct `s.id` match. Federation prefixes lane keys, session ids and the active-chat tid with one `prefixId`, so that match holds in a merged view. Its bar tid rewrite is a guarded pass-through. The delta path keys bars by (lane, id), and `id` stays on the wire. Full prompts stay in the builder-local map handed to `_bind_message_execs` and ride in the dead-lane memo. On a memo hit `st_turns` is empty, so the memo is the binder's only source, and it is filled. The fixture is synthetic. The privacy, hostname, home-path and vocabulary scans of the added lines are clean.

Follow-ups:

- `tests/test_timeline_bars_wire.py:136` (follow-up): MemoCarriesPrompts pins the memo-to-binder handoff with two source-string asserts. A mutation that keeps both strings but drops the handoff passes the full pytest suite. A memo-served dead lane's heuristic-joined message then falls back to the sent time on every build after the first. The DeadLaneMemo harness in `tests/test_timeline_lane_memo.py` can carry a behavioral case: build the lane twice with `_postal_messages` stubbed to one message, and assert exec equals the bar start on the memo-served build.
- `kernel/kernel.py:34584` (follow-up, nit): when a prompt's first line alone is a repeated token and its later lines vary, the view re-collapses the single wire line. The tip then shows the collapsed count where it showed the raw line before. Cosmetic; record it in the docstring and add the case to the fixture.
- `ui/webview/federation.ts:174` (follow-up, nit): the comments at 174 and 201 still describe bars carrying tid and the cue matching bars' prefixed tid. Kernel bars no longer carry it, and the cue matches the lane key.

Locally at head: `tests/test_timeline_bars_wire.py` plus `tests/test_ws_send_bounded.py`, 22 passed; `tests/test_kernel.py -k ViewBuilder`, 339 passed; vscode-extension npm test, 3637 pass, 0 fail. At the merge base with the PR's test files checked out, the new pytest file fails 11 of 11, `ViewBuilder::test_timeline_lane_and_segment_bar` fails on tid, and four node tests fail. CI green at caafb7b2.

Review run by the maintainer automated review session; verified against head caafb7b2.

PR #1152, opened 2026-09-09
T276c: the two process tests answer the same on every runner (fake pids above pid_max, seams, exact signal sequence)

henrypinkard:
Two CI-only flakes made deterministic; no behaviour change in the kernel beyond four call-time seams.

**Reaper shield test** (`tests/test_sdk_lifecycle_hardening.py`): the fixture's fake pids (558/559/901) could name a live process on a runner. Then `os.getpgid` succeeded and the tree kill took a real process group through `os.killpg`, which the test did not record (empty list on 3.12), or the liveness poll saw the pid alive and escalated (extra SIGKILL on 3.10). Fake pids now sit above `pid_max`, `os.killpg` is recorded, the liveness poll is pinned dead, and the assertion pins the exact sequence.

**Git cut test** (`tests/test_file_github.py`): it raced a real git against a 0.3 s deadline and a stand-in ssh's pidfile. On a slow runner git had not reached ssh by the cut. It now pins the mechanism on seams: fake `Popen` whose first `communicate` times out; the query must start git in its own session, `killpg(pid, SIGKILL)`, then reap.

**Seams** (`kernel/sdk_backend.py::_end_cli_tree`): `killpg`, `alive`, `sleep`, `now`, resolved at call time like the existing ones. A new test in `tests/test_cut_turn_tree_kill.py` drives the SIGKILL escalation on a fake clock with recorded sleeps.

Before/after loop counts (20 sequential runs each, own transient scope, no load) are in the task report; both tests were already 20/20 on the devbox, which is why the fix targets the mechanism rather than a timing constant.

Tier: fix

### DIFF
--- a/kernel/sdk_backend.py
+++ b/kernel/sdk_backend.py
@@ -7216,7 +7216,8 @@ class SdkBackend:
         except PermissionError:
             return True
 
-    def _end_cli_tree(self, pid: int, ps_lines: list[str], kill=None, run=None, cgroup=None) -> dict:
+    def _end_cli_tree(self, pid: int, ps_lines: list[str], kill=None, run=None, cgroup=None,
+                      killpg=None, alive=None, sleep=None, now=None) -> dict:
         """End an orphaned session CLI and EVERYTHING it left behind (T276): its scope unit when it runs
         in one (systemd ends every process in the cgroup, a tool shell's setsid children and any
         re-parented leftover included), and, always, the process tree the `ps` listing still shows
@@ -7229,6 +7230,12 @@ class SdkBackend:
         kill = kill or os.kill
         run = run or subprocess.run
         cgroup = cgroup or _read_cgroup
+        # the process-group kill, the liveness poll and the grace clock are seams too (T276c): a test then
+        # pins the exact signal sequence with no real process, pid or second behind it
+        killpg = killpg or os.killpg
+        alive = alive or self._pid_alive
+        sleep = sleep or time.sleep
+        now = now or time.time
         unit = scope_unit_of(cgroup(pid) or "")
         # The scope must be the CLI's OWN: bin/romp-cli-scope names the unit with the pid the CLI runs as
         # (`romp-session-<sid8>-$$-$t`, then execs into it), so a CLI in its own scope always carries its pid
@@ -7264,7 +7271,7 @@ class SdkBackend:
         def signal_all(sig, only_alive: bool) -> int:
             n = 0
             for p in targets:
-                if only_alive and not self._pid_alive(p):
+                if only_alive and not alive(p):
                     continue
                 if not same(p):
                     continue          # the pid now names another process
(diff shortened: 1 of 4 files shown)


ksarma:
Merged as shipped by the owner's auto-merge on green (2026-09-09 06:30Z, merge c709caeb). Nothing found here would have blocked it. Reviewed by reading only: the code signals processes and reads cgroups, so no test from the PR was executed on the review machine.

Reviewed inline: the four new seams in `_end_cli_tree` (`killpg`, `alive`, `sleep`, `now`) default to the callables the code called before, so kernel behaviour is unchanged. The #1148 ownership check (kernel/sdk_backend.py:7246) and the RealProcessTree case's empty cgroup seam (tests/test_cut_turn_tree_kill.py:284) are untouched. That file stays safe to run inside a romp session. The rewritten reaper-shield pin holds on a host without /proc. The class-level `_pid_alive` patch keeps the `os.kill(pid, 0)` fallback from reaching the recording mock. `_read_cgroup` and `_read_starttime` fail closed. `os.getpgid` raises for a pid above pid_max. The scope sweep parses the mocked ps text to zero units. The sequence is therefore `[(ORPHAN, SIGTERM)]` on both platforms. The fake-Popen git cut test asserts the three facts `_git_net_out` carries: own session, one `killpg` with SIGKILL on git's pid, a second `communicate`. It fails with any one reverted. `os.getpgid` and `_read_starttime` are not seams. They still read the host pid table and are inert only because every fake pid sits above pid_max. So #1139 finding (3) is closed in effect, not by a seam. The fix label fits: seams plus test determinism, no new mechanism.

Findings:

- tests/test_sdk_lifecycle_hardening.py:424, follow-up: `test_reaps_orphans_but_never_tmux` records `os.kill` without patching `_pid_alive`. On the dispatch-only macOS leg the fallback `os.kill(pid, 0)` reaches the mock. The pid reads as alive, the grace loop waits a real second, and a SIGKILL is recorded in the list. It predates this PR: present at the base and outside the two tests the body names. It will block the next release's macOS gate until the one-line pin from line 455 is applied here too.
- kernel/sdk_backend.py:7279, nit: `os.getpgid` stays outside the seams, so the `killpg` branch has no deterministic positive test. A `getpgid` seam and one fake-clock case where a descendant leads its own group would pin it.
- kernel/sdk_backend.py:7227, nit: the docstring still names three seams; the signature has seven.

Red-first, judged by reading: the new `test_a_survivor_of_the_grace_gets_sigkill_on_a_fake_clock` is red at the merge base by TypeError (no `killpg`/`alive`/`sleep`/`now` in the base signature). The two rewritten tests pass at base and head, the expected pattern for a flake fix, and the body says so. CI at 6b03322f: Python 3.10 to 3.13, Shell, vscode-extension and both tier checks green (3.10: 9954 passed, 62 skipped).

Review run by the maintainer automated review session; verified against head 6b03322f.

PR #1153, opened 2026-09-09
CLAUDE.md: a nudge-walk test stubs the shared read-only view beside load_goals and moves GOALDIR with STATE (T280)

henrypinkard:
Docs only. The "Goal-store fixtures use a PRIVATE synthetic sid" paragraph gains the second face of the same collision, from tonight's four end-to-end tests that were green alone and red in CI's serial order: a test that exercises the nudge walk and stubs `jd.load_goals` as the walk's snapshot must also stub `jd.load_goals_shared_or_fault` (the walk's shared read-only view since the jobs-stage change) and move the goal directory with the state (`jd._rebind_state(tmp)`, since assigning `jd.STATE` alone leaves `GOALDIR` where import bound it). The shared view reads a store file when one exists and delegates to `load_goals` only when none does, so an earlier module's store for the shared placeholder sid at the unrebound `GOALDIR` was what the walk read (no goal due, no fire, a deferral never cleared, a KeyError). The four fixed modules are named as the precedent.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -183,7 +183,18 @@ mid-test. The failure is ordering-dependent: green alone, red only under the ful
 suite. Tests that mint goals therefore use a private synthetic sid of their own
 (any invented uuid; still synthetic, never real) and clean their sid's journal in
 tearDown. Precedent + worked diagnosis: the model-fallback dedupe tests' class
-docstring (`tests/test_model_fallback_card.py`, DedupeBackstop).
+docstring (`tests/test_model_fallback_card.py`, DedupeBackstop). A second face of
+the same collision (2026-09-08, four end-to-end tests green alone and red in CI's
+serial order): a test that exercises the nudge walk and stubs `jd.load_goals` as
+the walk's snapshot must ALSO stub `jd.load_goals_shared_or_fault`, the walk's
+shared read-only view since the jobs-stage change, and move the goal directory
+with the state (`jd._rebind_state(tmp)` repoints GOALDIR and every derived dir;
+assigning `jd.STATE` alone leaves GOALDIR where import bound it), because the
+shared view reads a store FILE when one exists and delegates to `load_goals` only
+when none does, so an earlier module's store for the shared placeholder sid at the
+unrebound GOALDIR was what the walk read (no goal due, no fire, a deferral never
+cleared, a KeyError). Precedent: `tests/test_nudge_injected_turn_arm.py`,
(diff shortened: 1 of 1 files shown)


ksarma:
Recorded post-merge: the PR landed on green before this review posted. One should-fix doc follow-up and one nit follow.

Reviewed inline: the diff is `CLAUDE.md` only (+12 -1), so the `docs` label fits. The paragraph's mechanism claims hold at this head. The auto-nudge walk reads the store through `jd.load_goals_shared_or_fault` at `kernel/kernel.py:10913`, inside `_auto_nudge_session`. `load_goals_shared` (`kernel/judge.py:4305`) opens `GOALDIR/<sid>.json` and falls to `load_goals` only on `FileNotFoundError`. A store file left by an earlier module therefore beats a `load_goals` stub. `GOALDIR` is bound from `STATE` at import (`judge.py:80`). Only `_rebind_state` (`judge.py:144`) repoints it and the other derived dirs. Assigning `jd.STATE` alone leaves it in place, as that function's docstring already says. The kernel loads judge under one fixed module name, so every test module in a process shares that import-time `GOALDIR`, which is why the collision crosses modules. The four named precedent files exist and each stubs the shared view. The added lines carry no private strings, hostname, home path, or quoted user text.

Findings:

- **Should-fix, `CLAUDE.md:196`.** The Precedent sentence covers only the stub half of the rule. None of the four files calls `_rebind_state`. Each assigns `jd.STATE` alone (`test_nudge_injected_turn_arm.py:137`, `test_nudge_fresh_guard.py:56`, `test_nudge_memo_deadlock.py:54`, `test_nudge_bundle.py:255`). The fix commits 316b65d7 and 3fa4a4e4 added only the stub, which keeps the walk off `GOALDIR` entirely. One fix: scope the citation to the shared-view stub, and for the rebind half cite the `_rebind_state` docstring or `tests/test_nudge_gate_memo.py:65`, which does call it. The other: add `_rebind_state` to the four setUps as a test follow-up.
- **Nit, `CLAUDE.md:190`.** "jobs-stage change" has no anchor in the repo beyond one test docstring. Naming commit 3fa4a4e4 would make it findable.

Evidence: the four precedent modules pass at head (62 passed). A driver at the pre-fix commit reproduced the four failures the paragraph lists by seeding a store for the shared sid at the import-time `GOALDIR`. The same seed passes at head. All eleven CI checks green; merged 07:12Z.

Review run by the maintainer automated review session; verified against head 681a7587.

PR #1154, opened 2026-09-09
T278c: compact bars, per-lane compact judging, no delta key list; the view expands once at its boundary

henrypinkard:
The bars frame carries compact bars and per-lane compact judging, and no delta key list; the view expands both once at its boundary.

Tier: feature

## Why
After T278b the devbox's bars frame is 12.18 MB and takes 6 to 11 s to reach a client. Live breakdown: bars 5.45 MB (8,577 bars: 1.1 MB of key names, 0.9 MB of default values), judging 3.91 MB (14,051 flat entries repeating the 36-character lane sid and eleven key names each), the delta key list 2.35 MB (one composite key per bar and per judging entry), messages 0.47 MB. Aim: under 6 MB.

## What changed
**Kernel** (`kernel/kernel.py`):
- A bar is `{id, start, end}` plus one-letter keys with every default omitted (`_BAR_WIRE`: p promptId, w workId, r replyUuid, q prompt, c summary, m msgCaption, s src when not "typed", d mids when non-empty, u open, t cont, a nudgeAuto, o romp when true; `pending` was always false and never rides). `id`, `start` and `end` keep their names: the delta path keys bars by id, the federation merge rebases start and end by name, and the kernel's own readers use them.
- Judging rides per lane as compact entries `{k, t, t1, j, kd, x, ms, in, out, s, r, u}` (`_compact_judging`): `k` is a kernel-minted string key for the delta path, the lane's send time and judge (not the end time: an in-flight run's end is the build clock, so a key carrying it changed every build and the run crossed as a delete plus a set per frame; with a stable key it is one changed entry), the lane sid rides once as the lane key, `text` is cut to the 90 characters the tip shows, zero and empty fields and the default kind are omitted. The delta kind changes from the composite `bykeys:sid,t,judge,t1` to `dictlist:k`, so per-entry deltas stay per entry (a per-lane columnar shape would have re-sent a 2,925-entry lane on every verdict).
- The keyed full carries no key list. The shim derives every key from the frame exactly as `_delta_split` mints it: a dictlist item is lane, separator, id field; a byid item its id; a missing or duplicate key is positional `#n`; an empty or non-list lane is one entry under its bare prefix. Safe because every keyed field is now a string on the wire, which both languages spell alike; the old judging composite spelled floats, which is why a key list existed.
- The message binder reads the wire keys; the skeleton projection empties judging as a map; the dead-lane memo caches the compact bars as before.
- The four remaining id fields stay: 961 of 31,412 distinct values repeat (3%), so a string table would save about 60 KB and cost a decode; `id` embeds the lane sid as a prefix (0.32 MB), a candidate for a later round.

**View** (`ui/romp-timeline-view.js`): `expandBars` and `expandJudging` run once in `_mergeBars` and in `update()` for a one-shot payload, giving every reader the long names it always had. A lane array the shim hands back unchanged across a delta expands once and keeps its identity, so the per-lane caches keyed on array identity keep working. A long-named bar or a flat judging list (an older kernel, a test fixture) passes through with defaults filled. `BAR_WIRE` is the kernel's table, byte for byte, drift-guarded by a test.

**Federation** (`ui/webview/federation.ts`): judging is prefixed, rebased and merged like turns; an older kernel's flat list is converted to the per-lane shape first, so the pane's expander sees one shape whatever build each host runs.

## Field-by-field audit
| Field | Readers | Verdict |
|---|---|---|
| bar `id`, `start`, `end` | delta keyer; federation rebase; binder; view layout, hover, focus | keep by name |
| bar `promptId`, `workId`, `replyUuid` | view anchors and hover | short keys p, w, r |
| bar `prompt`, `summary`, `msgCaption` | view tip through `reqText` / `barBody`; binder (full text via the side map) | q, c, m; empty omitted |
| bar `src`, `mids` | view focus landing; view message join; binder exact join | s omitted when "typed"; d omitted when empty |
| bar `open`, `cont`, `nudgeAuto`, `romp` | view `barEndT`, dot suppression, dot styling | u, t, a, o omitted when false |
| bar `pending` | view `startAt`, live riders | never on the wire (always false at build); expander fills false |
| judging `sid` | view blocks and colour; federation prefix | the lane key |
| judging `judge`, `kind`, `text` | view band filter, rows, tip | j; kd omitted when "run"; x cut to 90 chars, omitted when empty |
| judging `t`, `t1`, `sent`, `recv`, `open` | view placement, merging, tip; federation rebase | t, t1 by name; s, r; u omitted when false |
| judging `ms`, `in`, `out` | view tip sums | omitted when zero |
| `_keys` | the shim's `buildMaps` only | gone; keys derived |
| messages | unchanged | unchanged |

## Tests
- `tests/test_timeline_bars_wire.py`: the encoder twin and `_expand_bar` as its inverse on every fixture bar; the compact literal (each short key assigned, no long key written); the judging round trip; the view's `BAR_WIRE` equals the kernel's `_BAR_WIRE`; a 25-lane, 8,000-bar, 14,000-entry synthetic frame including its old key list encodes to under half its old size; the committed fixture `tests/fixtures/timeline-bars-wire.json` (bars, judging, and the shaped prompts).
- `ui/timeline-bars-wire.test.ts`: the view's half of the round trip: expanding a wire bar gives back the old bar minus the three duplicates with defaults filled, the tip text and anchors match, a long-named bar passes through, an untouched lane keeps its identity, compact judging expands to the flat entries, hover still matches.
- `tests/test_view_deltas.py`: the shim mirror derives keys, fixtures build per-lane judging, the key-list pins say no full carries one; the node harness runs the shim's real functions.
- Adjusted to the shapes: `tests/test_kernel.py` (expands before reading), `tests/test_kernel_timeline_split.py`, `ui/timeline-hidden-hold.test.ts`, `ui/timeline-render.test.ts`, `ui/webview/multi-kernel-merge.test.ts`.
- Kernel modules touched: 594 passed after the folds. Extension suite: 3643 passed. Full pytest on the third commit: 9963 passed, 0 failed (a first run during the node build and mid-run edits flaked 46 tests that all pass alone). The adversarial review's confirmed findings are folded (below).

## Measurement
Before, live on the devbox: 12,181,764 bytes (25 lanes, 8,577 bars, 14,051 judging entries). After: measured once this deploys and reported to the manager with the same read-only client.

## Field audit (four parallel readers over the view, the federation merge, the kernel's post-build readers and the delta tests, then a synthesis)
Everything the audit asked for is in the branch: the judging `open` flag rides as `u`; `kind` is restored to "run" when omitted; `src` is omitted only when "typed", the one value its readers treat like absence; the expanders fill every default so no reader sees `undefined`; federation prefixes, rebases and merges judging per lane in both merge functions and converts an older kernel's flat list; the judging key excludes the end time (second commit). The audit's list of test pins outside the modules first run (the caption test's bar helper, the once-per-build wire test's judging fixture and key-list pins, one delta test's composite key) is folded in the second commit. Pre-existing and unchanged: federation rebases a judging entry's `t1`, `s` and `r` now, where before it shifted `t` alone.

## Review fold (four lenses, two skeptics per finding)
Real on this branch and fixed: the lanes skeleton now carries judging as an empty map and `update()` expanded it only for a one-shot payload, so the judge band's draw would have read a list method on an object on a cold start (every payload's judging is expanded now); the shim's positional-key check used the object's own `hasOwnProperty`, which an id spelling that name could shadow (it goes through the prototype). Honest gaps closed: the ratio test's old frame is the one this change starts from (after T278b), bound at two thirds, measured 0.60 on the synthetic; the binder's reads of the wire's `d` and `q` have tests; the judging table has the same kernel-versus-view drift pin as the bar table, and the federation twin is pinned on the shared fixture, which required the kernel to spell the key's time as JavaScript does (an integral float without ".0"); the shim stream test carries duplicate ids as a sleep-split segment does; the dead composite keyer's note says the shim has no path for it; a rewritten comment line paraphrases the user. Confirmed against the first commit only and already fixed by the later ones: the federation bars merge kept judging as a list, and the test modules outside the first run read the old shapes.


ksarma:
Post-merge record: I would have recommended merging with one follow-up. The PR landed on green at 07:00:46Z as a feature by the repository owner, which the policy allows.

Reviewed inline: the view's `BAR_WIRE` and `JUDGING_WIRE` tables match the kernel's `_BAR_WIRE` and `_JUDGING_WIRE` entry for entry. `test_the_view_expands_with_the_same_table` parses the JS literal, so any drift fails a test. The shim's `buildMaps` derives keys the way `_delta_split` mints them: lane plus separator plus the id field, `byid` ids, a positional `#n` from the order length on a missing or duplicate key, and one entry under the bare prefix for an empty or non-list lane. The node harness in `tests/test_view_deltas.py` runs the real shim over a stream with per-lane judging and duplicate ids. An older kernel's long-named bars and flat judging list pass through `expandBar` and the federation conversion with defaults filled. `expandBars` memoizes on the shim's lane array in a `WeakMap`, so an untouched lane returns the same expanded array across a delta. Federation prefixes lane keys, rebases `t`, `t1`, `s` and `r`, and merges judging per lane in both merge functions. The skeleton's judging is `{}`, and `expandJudging` returns `[]` for it; the new cold-start render test pins this. In the kernel, only `_bind_message_execs` reads bar fields beyond `id`, `start` and `end`. It reads `d` and `q` with the long-name fallback. The fixture is synthetic: placeholder ids, invented prompts, `TESTHOST`, and no private-string, hostname or home-path hits in the added lines.

- Follow-up (`kernel/kernel.py:34778` and `:34791`, `tests/test_timeline_bars_wire.py`). The body's review-fold paragraph says the `d` and `q` reads have tests; they do not. Commit d3846095 deleted the `MessageBinder` and `MemoCarriesPrompts` tests, and nothing replaced them. The three surviving binder tests in `tests/test_kernel.py:6387-6406` use the long names, so removing either the `d` or the `q` read leaves the full suite green (checked by mutation). Restore a binder case on a compact bar with `d` only and `q` only, and re-add the memo prompts pin.
- Follow-up (`tests/test_kernel_timeline_split.py:158`). The `PushSplit` fixture still carries judging as a list, which the `dictlist:k` kind cannot key. The test therefore runs the whole-frame fallback instead of the keyed fill it modeled before. Make judging a map.
- Nit (body Tests bullet). It says the synthetic frame encodes to under half its old size. `tests/test_timeline_bars_wire.py:204-205` pins the ratio between 0.5 and 0.65, and the review-fold paragraph has the right bound.

Locally at 7f7d4be0, the PR's eight changed Python test files gave 618 passed, and the extension suite gave 3644 passed, 0 failed. The new tests exercise the change: at the base with only the PR's test files checked out, pytest gave 45 failed and 4 errors, and the changed UI tests failed on the missing expanders. CI: nine of nine checks green at the head.

Review run by the maintainer automated review session; verified against head 7f7d4be0.

PR #1156, opened 2026-09-09
Jobs stage: the clear set is parsed once per file state, and the wake sweep reads the shared view

henrypinkard:
Tier: fix

## What

The jobs stage, second round (T267d; follows the placement-gate memo and the walk's shared-view read). Two hot spots the after-profile of that round named:

1. **The feed's clear set is parsed once per file state.** `_cleared_ids()` replayed the whole append-only `cleared.jsonl` (two thousand rows on the maintainer's box) for every session on every pusher cycle, from inside the nudge walk; once the placement gate was memoized it was 60% of the nudge tick and 44% of the jobs stage. It now keys a one-slot memo on the file's path and stat (`mtime_ns`, size, inode) taken **before** the read (the chain-memo rule), so a row appended during the read moves the stat the next call takes and a set parsed mid-write is served no further than that call. Every writer appends, so a same-second append moves the size (the kernel's file clock is coarse; mtime alone would miss it). An absent or unreadable file is the empty set, never cached. The slot is replaced whole, so two threads deriving at once can never pair one's key with the other's set. The nudge pass hoists one parsed set and hands it to every session it walks (`cleared=`); a caller without it derives as before.

2. **The awaiting-wake sweep reads through the shared view.** `_awaiting_wake_outcomes` took a fresh writer `jd.load_goals` per wake record per pass; that was the pusher's remaining goal loads after the walk's read moved. It reads the read-only view now (nodes, status, confirming, placements). The answered leg saves, so it alone takes a writer load, at the filing moment.

No card-move rule changes: the same inputs decide the same things; only the reads are cheaper.

## Tests

`tests/test_cleared_set_memo.py` (new): parsed once then served (one read over three calls); an append re-derives even when the file clock does not move (same `mtime_ns`, larger size); an undo row removes; a set read before an interleaved write is served no further than that call (the T267c interleaved-write pattern); an absent file is empty and never cached; the key carries the path so a rebound state root is a new key; a malformed row is skipped as before; the real pass over two alive sessions parses once and hands both the same dict; the counters ride `/perf` (`memos.cleared`).

`tests/test_kernel_awaiting_stamp.py` (three new tests in the sweep class): a silent due wake the walk cannot reach still escalates, decided through the view first with every writer load after it, inside the stamp; a due wake whose card was cleared is inert, one view read, no writer load; an answered wake files through exactly one writer load and the frozen view is never written.

`tests/test_perf_stats.py`: the memos key set and the new counter's shape. `docs/reference.md`: the memos passage. `tests/test_setting_gesture_order.py`: the walk stub accepts the new keyword.

## Measurement (before, on the merged first round; the shared box, no synthetic load)

Two `/perf` readings 120 s apart during an active hour, plus a 60 s py-spy sample (3066 samples):

| | |
|---|---|
| pusher cycles | 102 in 120 s (0.85/s); the loop's wait ended by the 0.5 s backstop 74 times, by an event 28 |
| cycle wall | 785 ms mean, p50 487 ms, p90 1428 ms; cycles cover 67% of wall |
| pusher thread CPU | 198 ms per cycle, 17% of a core |
| jobs stage | 49% of wall |
| process CPU | 94% of a core |
| judge tier | 10 passes in 120 s, 11 s each (93% of wall), 73% of a core |
| goal loads | 1172 per judge pass; 115 per pusher cycle |

**What drives the cycle rate.** The loop is `_pusher_cycle()` then `_pusher_wake.wait(0.5)`: it re-enters after at most half a second whether or not anything changed, so the rate is 1 / (cycle wall + 0.5 s), and a cheaper cycle means more cycles, not less pusher time per second. Three waits in four end on the backstop, not an event. The backstop exists for tmux sessions mid-turn, which have no per-message event (the loop's comment, 2026-06-22).

**What an idle cycle costs.** Not separable from the ring yet (it keeps p50, p90 and max); the p50 of 487 ms wall and 198 ms mean thread CPU bound it from above during this window. An idle-cycle counter (a cycle that sent nothing and changed nothing, with its own wall and CPU sums) is the instrumentation to add before touching the cadence.

**Where the process's CPU is.** The pusher thread is 15% of the sample. The judge tier thread (`_run_tier`) is 61%, and 83% of that is `run_triage > run_courier > load_goals`: the courier loads every session's goal store with the writer's loader on every triage pass, plus one per pending row and per sender. The planner worker is 22%. So the number the user watches, the process's CPU share, is set by the judge tier, not the pusher.

**Proposed, not done here.** (a) The pusher waits on its event, with the 0.5 s backstop only while a tmux session is mid-turn, else the earliest due time among the time-based ticks. (b) The courier's per-session scan reads through the shared view and skips a session whose parse and store stat have not moved since its last pass, the same shape as the placement gate; its writes keep the writer's loader. Both are separate rounds.

## After-numbers

Posted below after the deploy restart at matched uptimes (3, 6, 20 min), with loads per cycle and the jobs-stage share.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -908,7 +908,9 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   (`hit`, `miss`, `populate`, `bypass`); `nudgeGate` is the auto-nudge walk's
   planner-placement gate, derived once per (parse, store) and served while
   both stand (`served`, `derived`; a healthy quiet box serves almost every
-  cycle). The compaction sweep after each judge pass evicts from `pass` and
+  cycle); `cleared` is the feed's clear set, parsed once per state of
+  `cleared.jsonl` (its stat, taken before the read) and served while the file
+  stands (`served`, `derived`). The compaction sweep after each judge pass evicts from `pass` and
   `shared` the entries of stores no session in the discover window owns, so
   both stay bounded by the live board; the gate memo is bounded by the session
   count.
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -9166,7 +9166,9 @@ def _auto_nudge_pass(now, tmux, run_dead_wait):
     alive_ids = {s["sid"] for s in alive}
     waitfor = _wait_for_graph(now, alive_ids)             # {sid:{peerSid,name,inCycle}} — the peer-wait gate
     fired = False
-    cleared = _cleared_ids()                              # one parsed clear set for every session this pass walks (2026-09-09)
+    cleared = _cleared_ids()                              # one parsed clear set for every session this pass walks (2026-09-09):
+    #                                                       a clear landing mid-pass reaches the later sessions next pass; the
+    #                                                       node's own cleared flag, written in the same gesture, covers the gap
     for s in alive:
         # PER-SESSION ISOLATION (2026-07-16): one session's failure — a bad backend snapshot, a
         # malformed store — must not abort the whole tick and silence nudging fleet-wide. A
@@ -30559,7 +30561,8 @@ def build_episode(sid, now):
 
 # ───────────────────────── feed clear / undo (inbox-zero) ─────────────────────────
 _CLEARED_MEMO = {"slot": None}     # (key, parsed set) or None: the clear log's stat taken BEFORE the read, and the set read under it
-_CLEARED_STATS = {"served": 0, "derived": 0}
+_CLEARED_STATS = {"served": 0, "derived": 0}   # bumped from the pusher AND socket threads (undo, connect-time builds) with
+#                                                no lock: a lost count under a race is tolerated, these are diagnostics only
 
 
 def _cleared_ids():
@@ -30599,7 +30602,7 @@ def _cleared_ids():
                 cur[iid] = o.get("t", 0)
     except OSError:
         _CLEARED_STATS["derived"] += 1
-        return cur                                       # vanished between the stat and the read: empty, uncached
+        return cur                                       # absent, vanished after the stat, or unreadable: empty, uncached
     _CLEARED_STATS["derived"] += 1
     if key is not None:
         _CLEARED_MEMO["slot"] = (key, cur)
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -382,6 +382,7 @@ class _PerfStats:
             except Exception:
                 memos[key] = {}
         memos["nudgeGate"] = dict(_NUDGE_GATE_STATS)   # the nudge walk's placement gate: served vs re-derived (2026-09-09)
+        memos["cleared"] = dict(_CLEARED_STATS)       # the clear set: parsed once per file state, served while it stands (2026-09-09)
         now = time.time()
         return {"now": now, "since": since, "uptime_s": now - _STARTED, "log": _PERF,
                 "process": _process_stats(), "pusher": pusher, "stages_ms": stages,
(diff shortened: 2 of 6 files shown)


henrypinkard:

Live kernel on the shared box, `/perf` readings, no synthetic load. The "before" rows are the readings taken after the previous jobs-stage round (the placement-gate memo) landed, at the same uptimes on the same reading script; both rounds ran in the same late-night hour with the same sessions.

| code | uptime | goal loads / cycle | jobs stage (% of wall) | process CPU (% of a core) | pusher thread CPU | clear set served / derived |
|---|---|---|---|---|---|---|
| before | 3.3 min | 56 | 45% | 67% | n/a | n/a |
| before | 6.5 min | 55 | 44% | 69% | n/a | n/a |
| before | 20.6 min | 80 | 43% | 80% | 17% (from the sample) | n/a |
| after | 4.2 min | 46 | 39% | 69% | 13% | 434 / 1 |
| after | 6.4 min | 38 | 36% | 63% | 13% | 655 / 1 |
| after | 19.5 min | 52 | 33% | 71% | 10% | 1630 / 1 |

The clear set was parsed once per kernel and served 1630 times over 1450 cycles. Goal loads per cycle fell this round (the awaiting-wake sweep's fresh load per record moved onto the view): 46, 38 and 52 against 56, 55 and 80 at matched uptimes. The jobs stage's share of wall is down to a third; the pusher thread's own CPU is a tenth of a core at 20 min.

**Process CPU barely moves (71% against 80%) because the pusher is not where it goes.** The judge tier thread is at 58% of a core at this reading, with 1062 goal loads per triage pass over 71 passes: the courier's per-session scan, which the follow-up round (the courier change gate) addresses.

ksarma:
Merged as shipped (12584b63, merge 28ecee8c, 2026-09-09 07:20Z, the owner's auto-merge on green). This is a post-merge record. Nothing found here would have held the merge. Two follow-ups and two nits are listed below.

Reviewed inline: the clear-set memo takes the file's stat (mtime_ns, size, inode) before the read. It leads the key with the path and replaces its one slot whole. An absent or unreadable file returns an empty set and is not cached. Rows that fail to parse or carry no id are skipped. Every writer of `cleared.jsonl` appends, so a same-second append moves the size. The nudge pass parses the set once and hands it down through `cleared=`. The walk derives the set only when the argument is None. So an empty set handed down is honoured, and a caller without it behaves as before. No caller mutates the returned dict. The awaiting-wake sweep reads each store through `load_goals_shared` and uses the view only for nodes, status, confirming and placements. Its answered leg takes a fresh `jd.load_goals(sid)` at the filing moment, so the view is never written there. A probe at this head filed the nudge/awaiting row through one writer load with the shared cache still on. No card-move rule changes: the same fields decide the same outcomes. The one semantic shift is that a clear landing mid-pass reaches later sessions on the next pass. The node's cleared flag, written in the same gesture, covers that.

- `kernel/kernel.py:10133`, follow-up. This is the open #1141 regression; this PR does not touch that line. The walk's answered leg in `_wake_goal` still hands the shared view from line 10919 to `_file_wake_answer`. `record_verdict` raises `FrozenStoreError` inside the helper's own except. So the first answered wake per kernel process files no row, and the shared cache switches off until restart. The new comment at 10278 says it matches the walk's leg, and the one at 10917 names `_fresh`. Both describe a reload that leg does not make. The fix is the one applied to the sweep, plus a walk-class twin of `test_sweep_files_an_answer_through_one_writer_load` driven through the real shared loader.
- `kernel/kernel.py:30563`, follow-up, record only. A one-slot memo with module state, two counters, a `/perf` key and a docs passage is a new mechanism. The review bar counts that as `feature`. The owner's role merges both on green, so the label changed no gate.
- `tests/test_kernel_awaiting_stamp.py:839`, nit. `test_sweep_files_an_answer_through_one_writer_load` is green at the base and never asserts the nudge/awaiting row landed. Adding that assertion completes the pin.
- `kernel/kernel.py:9169`, nit. The hoisted parse now runs outside the per-session try. A non-dict JSON row raises `AttributeError` on `o.get` past the inner except, which would skip the whole pass, sweeps included. An `isinstance` check per row closes it.

Tests: at the merge base c709caeb with the PR's four test files, 13 failed and 166 passed. The failures were all ten memo tests on the missing `_CLEARED_MEMO`, two sweep tests on the loader path, and the `/perf` shape test on the missing key. At head the same four files: 179 passed. The 21 seam-adjacent files under xdist: 1289 passed. CI is green at head, and the Tier policy check passed.
Review run by the maintainer automated review session; verified against head 12584b63.

PR #1155, opened 2026-09-09
Notification titles follow one rule — "Romp needs you: <session>" for a needs-you card, "Romp: <session>" for everything else — and the relay stops grafting the host onto them

yatharth:
Tier: feature

## The rule

- A card moving to **needs-you** (blocked on the user) is titled `Romp needs you: <session>`.
- **Everything else** — a card done/ready, a turn finished, the test notification — is titled `Romp: <session>` (a sid-less test: plain `Romp`).
- The **body** is whatever the event already wrote — unchanged; no extra lines, no host/sid prefixes, no `romp: host:` decoration. This holds for relayed events from other machines too: the title is the session name alone, and the tap still carries the routing in `data`.

## Before / after (synthetic names)

| event | before | after |
|---|---|---|
| card of `web` enters needs-you | title `romp: web` / body `Needs you: pick a migration` | title `Romp needs you: web` / body unchanged |
| card of `api` completes | title `romp: api` / body `Completed: shipped the login flow` | title `Romp: api` / body unchanged |
| `web` finishes a turn | title `web` / body first line it said | title `Romp: web` / body unchanged |
| test, `web` in front | title `romp` / body `Test notification — tap to come back to web.` | title `Romp: web` / body unchanged |
| test, no session | title `romp` / body `Test notification — this device is set up.` | title `Romp` / body unchanged |
| relayed from `boxa`, card of `web` needs-you | title `romp: boxa:web` | title `Romp needs you: web` (as the origin composed it; sid still gains `boxa:` for routing) |

## Design

- One builder, `_notify_title(session_name, needs_you=False)` in `kernel/kernel.py`, next to `_system_notify`. Every producer goes through it: the card leg (`_feed_notifications`, which also feeds the desktop `_system_notify`), the turn leg (`_turn_notify_tick`), and the test push (`_push_test`). The relay (`/push/relay`) stops rewriting the title and touches the sid only.
- `needs_you` is decided from the card's **column** (`needs_input` vs `completed`) — the authoritative state the bell event already carries — never from the body text.
- For a federated test push the title wears the bare remote name (`Romp: api`); the body and the popover's echoed `name` keep the host-prefixed `boxa:api` the merged dashboard shows, so the result line and the lock-screen sentence still agree.
[...]

PR #1157, opened 2026-09-09
The push worker leaves a fingerprint the page can read, and a notification it showed but nobody tapped through is offered when the app comes forward

yatharth:
Tier: fix

Follow-up to #1127 (merged), from the warm-app round on the phone later the same day: three taps, three 201s from the push service, and then nothing — no `[reveal]` line, no `sw-message` row, `tap-resume {found:false}` on every boot and every return to the foreground, and each tap booting a fresh page on the start URL with no deep link. The cold-start round that morning had the worker's own trail (`clients:0`, road `open`) and a link; this round had no sign the worker's `notificationclick` ran at all.

## Two hypotheses, and the fingerprints that settle them

Either iOS delivers a tap on a live Home Screen app to the app itself (relaunching it on the start URL) and bypasses the worker, or the phone still runs an **older worker** — a Home Screen app may not re-check its worker on relaunch — that never wrote the store. The trail could not separate the two, because nothing said which worker ran.

The worker now writes a fingerprint beside the tap: `/__romp/sw` in the `romp-tap` cache, `{version, installedAt, activatedAt, lastPushAt, lastPushSid, lastClickAt, lastClickSid, clicks}`. `version` is baked into `/sw.js` at serve time (`_sw_version()`: the kernel's short sha + the dist token — moves with every deploy and every bundle rebuild) and the shell page carries the same string. `install` starts a fresh record for the new build, `activate` stamps its takeover, every `push` its arrival (before `showNotification`), every `notificationclick` its arrival at the very top of the handler, before anything can await, and bumps `clicks`.

The shell folds the reading into **every** `tap-resume` row — `swVersion`, `swMatchesPage`, `lastPushAgeS`, `lastClickAgeS`, `clicks` (ages clipped; never ids) — files `sw-stale {swVersion, pageVersion}` when the worker's build is not the page's, and asks `registration.update()` at boot and on every `visibilitychange`→visible (`sw-update {ok, reg}`), so a stale worker is replaced at the next opportunity. The next phone round reads as: `swMatchesPage:false` → hypothesis two (an old worker); `swMatchesPage:true, lastPushAgeS` seconds, `lastClickAgeS:-1, clicks:0` → hypothesis one (the click never reached the worker).

## A landing road that needs no `notificationclick`: the offer

Every session-addressed push also writes the notification it displays to `/__romp/shown` (`{id, sid, host, kind, cardId, url, name, t}`, one slot, latest wins) **before** attempting the show, so a show a browser refuses still leaves the record. A page that comes forward with that record and no stored tap does not jump — the user may have opened the app for another reason — it **offers**: a chip at bottom-left in the jump-chip family's dress on the shell's own menu tokens, "Open *name* · from the notification", with a dismiss. Taken → the existing `land()` path, `via:'offer'` (the kernel's `[reveal] offer …` line through the existing route; `offer` joins `_REVEAL_ROADS`), the record retired everywhere it is kept (`tapLanded` by id). Dismissed → retired without landing. Never shown when the active session already is the one named (retired: the notification's purpose is met), never on a deep-link boot, and never beside a stored tap — the tap wins. The name rides the push payload (`data.name`, resolved by the helper the test push always used; the turn leg passes its title). Instrumented: `tap-offer {shown, via, ageS, why}`, `tap-offer-click`, `tap-offer-dismiss`.

## Tests

- `tests/test_kernel_webpush.py`: the worker writes `/__romp/sw` on install/activate/push/click with the baked version; the shown record is written before the show and survives a refused show; every `tap-resume` row carries the fingerprint; `sw-stale` on a mismatch; `registration.update()` at boot and on every visible; the chip: shown-but-untapped → chip with the name; click → `/reveal` via `offer` + retire; dismiss → retire without a reveal; stored tap present → no chip; active session equals the shown sid → no chip; deep-link boot → no chip. The payload builder's `name`, the served `/sw.js` carrying the version.
- `tests/test_notification_tap_resume_browser.py`, second scenario in a real Chromium: the real `push` handler runs on a synthetic `PushEvent` carrying the kernel's payload; headless Chromium refuses `showNotification` (the record is there anyway); `notificationclick` is never dispatched; the page comes forward; the chip appears at bottom-left naming `api`; clicking it switches the chat tab, retires the record, and the kernel logs `[reveal] offer … delivered`. Against b8d2a90b (#1127's tip) it fails on the outcome the phone showed: no chip, tab still `web`, no `/reveal`, no record.

## Verification

- `tests/test_kernel_webpush.py tests/test_kernel_notify_popover.py tests/test_kernel_mobile.py tests/test_notification_tap_resume_browser.py`: 193 passed (after merging `main`, which had moved past #1127 with the `_REVEAL_ROADS` whitelist and the `wsState` boot latch; both kept).
- Full Python suite, run in two chunks under the local time cap: 9862 passed, 42 skipped, 2 failed — `test_dashboard_reload_served` and `test_ship_reship`, both served-browser tests that fail identically on the untouched b8d2a90b tree on this box (pre-existing, unrelated).
- No `ui/` bundle changes.


yatharth:
Heads-up for whichever of #1155 (notification titles) and #1157 lands second — verified on a local merge of the two, notification test files green after:

- `_push_test`: keep #1155's title (`_notify_title(bare_name)`, the bare session name) AND #1157's routing-block `name=` on `_push_payload`; the shell's `label` must still be flattened and clipped (`" ".join(label.split())[:PUSH_LABEL_MAX]`) before it can stand in for a name — #1157 does that inside `_push_session_name`, #1155's inline block does not, so the merged body needs one of the two.
- `_turn_notify_tick`: `sname = str(s.get("name") or _name_of(sid) or sid[:8]); title = _notify_title(sname); _push_notify(title, body, sid, kind="turn", name=sname)`.
- `tests/test_kernel_webpush.py::LandingRevealOffers::test_no_offer_for_the_session_already_in_front` asserts `ageS == 5` against a real clock and came out 6 once on the merged tree — worth making it a range (or freezing the clock) on this branch.

yatharth:
Follow-up from the live round after this merged (2026-09-09 21:45 UTC, iPhone Home Screen app, kernel 4211880a = main + this PR + others). The fingerprints answered the question, and the answer was neither of the two hypotheses:

- `tap-resume` rows on every resume read `swVersion` = this build and `swMatchesPage: true`: the worker the page registered is current.
- Across a test push (201 from the push service) and a tap on it: `lastPushAgeS: -1`, `lastClickAgeS: -1`, `clicks: 0`, `found:false`, no `tap-offer` row (so no `/__romp/shown` either), no `sw-message`, no `[reveal]` line; the app came forward without navigating.

So the worker instance that FIELDS the push and the tap on iOS writes where the app's page cannot read — a storage partition: not the page's Cache Storage — and sees no client of the app in `matchAll` (which is why only the cold-start deep link, routed by the OS, ever landed). Every storage-based hand-off this PR and #1127 added (`/__romp/tap`, `/__romp/shown`, the `/__romp/sw` stamps, the message, the replay) is invisible on the receiving side.

The kernel is the one place both sides reach, so https://github.com/romp-on/romp/pull/1219 makes it the meeting point: every session-addressed push carries a `pid` and files a ledger row; the worker acks `shown` and `clicked` by pid alone (`POST /push/ack`, no cookie needed); the page asks `GET /push/pending` for its own subscription when it comes forward, lands a clicked push via `ack` and offers a shown one from the row. A third real-browser scenario reproduces the partition (the worker's Cache Storage replaced by one the page cannot see, no clients) and fails on d5653088, passes there.

PR #1158, opened 2026-09-09
Courier: skip a session whose inputs have not moved, read the shared view; the pusher counts idle cycles

henrypinkard:
Tier: fix

## What

The judge tier's courier, and the pusher's idle-cycle counter (T267e; follows the jobs-stage rounds).

1. **The courier skips a session whose inputs have not moved since a scan that found nothing to place.** `run_courier` scanned every session's transcript and goal store on every triage pass, loading the store with the writer's loader: on the maintainer's box, 1172 goal loads a pass across 18 sessions, a pass every 12 s, 83% of the judge tier thread's samples, which is where most of the process's CPU went (the judge thread at 73% of a core against the pusher's 17%). The skip key is every input the per-session scan reads, taken **before** the store read (the chain-memo rule): the parse cache's fileset key bound to the session object the scan holds, the store file's key with its journal's and archive's (the shared view's inputs), the episode log's key (the floor), and the transcript path. A session is recorded as seen only when its scan added nothing pending and no placed delegate with a live node lacks its courier link (the link repair's other input is the sender's store, outside the session's key); a delegate filed fyi or retired has no node to attach to and no repair to wait for, so it records like any settled session. One with rows to place or a link to repair is scanned again next pass whatever the key says. A parse the cache does not hold (a stubbed one) is never keyed, so never skipped. The table is pruned to the pass's fleet and cleared with the state root. Counters ride `/perf` as `memos.courierSkip` (`skipped`, `scanned`, `recorded`).

2. **The scan reads the shared read-only view.** Its one writer, the planner-first link repair, asks the view whether the link is missing (`_courier_link_wanted`, the idempotency half split out of `_attach_courier_link`) and takes a writer load only then. The placement loop and the cross-host arm keep the writer's loader: they write.

Follow-up named, not done: `_handoff_backref` walks every discovered session's store with the writer's loader on each call; it runs only when a link is missing, as before, and a per-pass memo of the backref map would remove the walk.

3. **The pusher counts idle cycles.** A cycle that set no wake, sent no client payload and saved no goal store is idle; its wall and CPU go to `pusher.idle_cycles`, `idle_ms_sum`, `idle_cpu_ms_sum` (`pusher.sends` counts payloads that went to a client; a deduped frame the client already holds is not one, or the per-cycle chat frames would mark every cycle busy with a dashboard connected). A conservative undercount: a wake another thread sets during the cycle, or a periodic repost of an unchanged frame past the dedup window, marks that cycle busy. The loop re-enters after a fixed 0.5 s backstop whether or not anything changed (three waits in four end on it), and the idle share is what a cadence change would be judged on. No cadence change here.

No card-move rule changes: a skipped session has no new information by construction, and a moved session decides exactly as before.

## Tests

`tests/test_courier_skip.py` (new; three recipient sessions, one sender, a stubbed courier model): unchanged inputs skip after a scan that placed nothing, and a skipped pass writes nothing; a moved store, an override-journal write, a fresh parse or an episode boundary un-skips that session alone; a write landing during the scan (the interleaved-write pattern) is seen next pass because the recorded key predates it; three sessions with one moved place exactly what the ungated pass places (placements and planted nodes equal, run in two fresh worlds); a delegate filed quiet (an unrooted dispatch) has no repair to wait for and records and skips; a planner-first placement whose link is missing keeps its session scanned every pass until the sender's tracking node appears, then the link attaches through one writer load and the session settles (the review's finding); a parse the cache does not hold is never skipped; a scan with nothing to place asks the writer's loader for nothing; the counters are a copy; a rebound state root forgets the keys.

`tests/test_perf_stats.py`: an idle cycle is counted with its wall and CPU; a cycle that sends a payload, sets the wake, or saves or writes a goal store is not idle; a deduped frame does not break an idle cycle; the memos key set and the courier gate's shape. `tests/test_courier_link.py`: the repair-branch pin follows the split. `tests/test_goal_store_fault_boundary.py`: the three courier fault tests fault the shared loader beside the writer's, or record the view's reads, since the scan's read moved. `docs/reference.md`: the pusher and memos passages.

## Before (the merged jobs-stage rounds; the shared box, no synthetic load)

Two `/perf` readings 120 s apart in an active hour plus a 60 s py-spy sample (3066 samples): judge tier 10 passes in 120 s at 11 s each (93% of wall), judge thread CPU 73% of a core, 1172 goal loads per triage pass; `_run_tier > run_triage > run_courier > load_goals` 83% of the tier thread's samples; pusher thread 17% of a core, 102 cycles in 120 s with 74 waits ended by the backstop.

## After

Posted below after the deploy restart at matched uptimes: judge-thread CPU share and goal loads per triage pass, and the idle-cycle numbers once they have an hour.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -887,9 +887,12 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   `cycle_ms_sum`, `cycle_ms_max` (since start), `cycle_ms_last`,
   `cycle_cpu_ms_sum` (the pusher thread's own CPU time), `cycle_ms_p50`,
   `cycle_ms_p90`, `cycle_ms_ring_max`, `ring_n` from the last 256 cycles,
-  `sends` (every client payload), and `idle_cycles`, `idle_ms_sum`,
-  `idle_cpu_ms_sum` (cycles that set no wake, sent no payload and saved no
-  goal store: what a longer wait between cycles would have skipped).
+  `sends` (every payload that went to a client; a deduped frame the client
+  already holds is not one), and `idle_cycles`, `idle_ms_sum`, `idle_cpu_ms_sum`
+  (cycles that set no wake, sent no payload and saved no goal store: what a
+  longer wait between cycles would have skipped; a conservative undercount,
+  since a wake set by another thread or a periodic repost of an unchanged
+  frame marks a cycle busy).
 - `stages_ms`: `jobs` (the cycle's tick jobs outside the push), `push`, and
   inside it `push.chat`, `push.feed`, `push.timeline`, `push.send`. The
   `push.*` stages count every push, including the one a connecting page gets,
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -885,8 +885,11 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
 - `pusher`: `cycles`, `wakes` (every wake call; a burst of wakes runs one
   cycle), `wakes_event` and `wakes_backstop` (how the loop's wait ended),
   `cycle_ms_sum`, `cycle_ms_max` (since start), `cycle_ms_last`,
-  `cycle_cpu_ms_sum` (the pusher thread's own CPU time), and `cycle_ms_p50`,
-  `cycle_ms_p90`, `cycle_ms_ring_max`, `ring_n` from the last 256 cycles.
+  `cycle_cpu_ms_sum` (the pusher thread's own CPU time), `cycle_ms_p50`,
+  `cycle_ms_p90`, `cycle_ms_ring_max`, `ring_n` from the last 256 cycles,
+  `sends` (every client payload), and `idle_cycles`, `idle_ms_sum`,
+  `idle_cpu_ms_sum` (cycles that set no wake, sent no payload and saved no
+  goal store: what a longer wait between cycles would have skipped).
 - `stages_ms`: `jobs` (the cycle's tick jobs outside the push), `push`, and
   inside it `push.chat`, `push.feed`, `push.timeline`, `push.send`. The
   `push.*` stages count every push, including the one a connecting page gets,
@@ -910,7 +913,10 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   both stand (`served`, `derived`; a healthy quiet box serves almost every
   cycle); `cleared` is the feed's clear set, parsed once per state of
   `cleared.jsonl` (its stat, taken before the read) and served while the file
-  stands (`served`, `derived`). The compaction sweep after each judge pass evicts from `pass` and
+  stands (`served`, `derived`); `courierSkip` is the courier's change gate
+  (`skipped`, `scanned`, `recorded`: a session whose parse, store, journal,
+  archive and episode log have not moved since a scan that found nothing to
+  place is skipped whole). The compaction sweep after each judge pass evicts from `pass` and
   `shared` the entries of stores no session in the discover window owns, so
   both stay bounded by the live board; the gate memo is bounded by the session
   count.
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -164,6 +164,7 @@ def _rebind_state(path):
     _lastsid_memo.clear()   # sdk-registry reads are mtime-memoized per sid — a rebind must not serve the old root's values
     _STORE_FAULTS.clear()   # unreadable-store episodes belong to the old root's files
     _CHAIN_MEMO.clear()     # the write-moment chain memo keys on paths under STATESDIR; a new root is a new world
+    _COURIER_SEEN.clear()   # the courier gate keys on the old root's files
     _episode_memo.clear()   # ...and so are the episode-log reads
     _head_memo.clear()      # transcript heads are immutable per path, but a rebind swaps the whole world of paths
     _namefp_memo.clear()    # names-entry content is memoized per SID against same-second mtimes — across a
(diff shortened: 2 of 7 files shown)


ksarma:
Nothing in this PR needs a revert or a hold. It merged on green at e93162dc while this review ran, so this is a post-merge record.

Reviewed inline: the courier scan now reads each store through the shared read-only view. Given the #1141 regression, I traced every consumer of that view for a write. `_session_settled` and `_awaiting_bg_hold` only read. `_segs` and `apply_seams` read the store and write only `seamOf` on a freshly split segment. `_courier_link_present`, `_courier_link_target` and `_courier_link_wanted` read; `_handoff_backref` loads sender stores through the writer's loader. The one writer, `_attach_courier_link`, takes its own `load_goals(fsid)` at the write moment. The placement loop and the cross-host arm keep the writer's loader throughout. An in-process probe over every skip scenario and the link-attach pass left `shared_store_stats()` at off=0, poisoned=0, with no frozen-store row filed. The skip key is taken before the store read. It covers the parse cache's key bound to the session object, the store file, its journal and archive, the episode log and the transcript path. A session is recorded only when its scan added nothing pending and no placed delegate with a live node lacks its link. A stubbed parse is never keyed. The table is pruned to the pass's discover set and cleared on rebind. `closed[fsid]` is consulted only for sessions with pending rows, which are never skipped, so the rollup's settled input is unchanged. The idle-cycle counters compare wake, send and goal-save marks around a cycle and change no cadence.

Findings:

- `kernel/judge.py:2174` and `kernel/judge.py:15138` (follow-up): two new comments use the word the repo's vocabulary rule bars from new prose. Say "the pass's sessions". The loop variable on 15137 is the pre-existing identifier and is fine. The commit messages are clean; the PR body is not.
- `kernel/judge.py:14320` (follow-up): `_courier_link_wanted` runs `_handoff_backref` on the view, and it runs again inside `_attach_courier_link` on the writer's load. Each attach therefore walks every discovered store twice. A placed delegate whose sender never tracks it is scanned every pass. The per-pass backref memo the PR body names would remove both.
- `docs/reference.md` and the `snapshot` comment in `kernel/kernel.py` (follow-up): the shared cache now also serves the courier's scan reads. The note attributing `memos.shared` to the pusher alone is therefore incomplete.
- `tests/test_courier_skip.py:259` (resolved in the PR): at df1a34f2 the repair test split the transcript on the bare marker text, and `test_postal_marker_form` failed on every Python job. Commit 13061074 reads the id from the fixture. CI at e93162dc is green.
- Label (record only): the diff adds a memo, new `/perf` counters and a helper split, so it is feature-shaped by the review bar. The label matches this series' precedent (#1059, #1141). The owner's gate is the same either way.

Test evidence: the red-first run at the base 28ecee8c with the PR's test files gave 20 failed, 76 passed (13 AttributeError on `_COURIER_SEEN`, 5 KeyError on `idle_cycles`, the memos-set assertion and the `_courier_link_wanted` source pin). At the merged content, 99 passed across `test_courier_skip`, `test_courier_link`, `test_goal_store_fault_boundary`, `test_perf_stats` and `test_postal_marker_form`. Mutation checks (gate disabled, writer's loader in the scan, repair guard dropped, idle never counted) each fail the test written for them. CI at e93162dc: every job green (Python 3.10 to 3.13, bats, vscode-extension, both tier checks); merged 08:08:57Z.

Review run by the maintainer automated review session; verified against head e93162dc (the review began at df1a34f2; the merged head carries that diff plus two fold commits, rebased).

henrypinkard:

Live kernel on the shared box, `/perf` readings, no synthetic load. "Before" rows are readings on the previous code (the clear-set memo round) in the same hour; "after" rows are this change as merged. Loads per judge pass is `goals.loads` over `judge.passes` since the kernel's start.

| code | uptime | goal loads / judge pass | judge passes | judge thread CPU (% of a core) | process CPU | pusher thread CPU | courier gate skipped / scanned / recorded |
| before | 6.4 min | 1152 | 17 | 45% | 63% | 13% | n/a |
| before | 10.6 min | 1067 | 43 | 61% | 76% | 11% | n/a |
| before | 19.5 min | 1062 | 71 | 58% | 71% | 10% | n/a |
| after | 3.5 min | 1103 | 14 | 58% | 80% | 15% | 65 / 341 / 5 |
| after | 6.5 min | 1058 | 27 | 61% | 78% | 11% | 135 / 677 / 5 |
| after | 19.3 min | 1017 | 90 | 71% | 84% | 11% | 450 / 2189 / 5 |

**Goal loads per triage pass did not fall, and the gate recorded five sessions in 90 passes.** A 60 s py-spy sample of the merged kernel put 100% of the courier's remaining writer loads on one line, the placement loop's per-row `store = load_goals(fsid)`, and the judge error log carried no courier rows, so nothing was deferred or failing. Cause: the scan tested placement by exact segment id while the placement loop uses the drift-tolerant `_placed_key`, so a peer segment placed under a time-shifted key was queued every pass, loaded a store every pass, was dropped there unwritten, and kept its session from ever recording. The follow-up fix (the scan applies the same rule before queueing) merged separately; its after-numbers go on that pull request. The scan's own reads did move to the view as intended: `load_goals_shared` and the view-side link check appear in the sample, the scan's writer load does not.

**Idle cycles (the addendum's instrumentation).** Over the 19 minutes: 977 of 1169 cycles idle (84%), 349 ms wall and 70 ms of pusher-thread CPU each; at 3.5 and 6.5 min, 87% and 89%. An idle cycle here is one that set no wake, sent no client payload and saved no goal store, counted conservatively (a wake from another thread or a periodic repost of an unchanged frame marks the cycle busy). Those idle cycles cost the pusher thread about 70 ms of CPU a second at one cycle a second, roughly 7% of a core, the upper bound on what an event-driven wait would save on the pusher thread. The one-hour reading follows.

henrypinkard:
**The one-hour idle reading** (the deployed kernel with the courier gate and the drift fix, 60.6 min uptime): 4402 cycles at 1.21 a second, 3757 idle (85%), 81 ms of wall and 60 ms of pusher-thread CPU each. Idle cycles therefore cost the pusher thread about 6% of a core over the hour; the thread as a whole ran at 8%, the judge thread at 45%, the process at 57%. Goal loads per triage pass 231 over 623 passes; the courier gate skipped 18023 scans against 44 performed. The cadence proposal built on these numbers went to the manager.

PR #1160, opened 2026-09-09
An answered wake files its diary row from its own writer load of the goal store

ksarma:
## What

**Symptom.** Since the nudge walk started reading the goal store through the shared read-only view, the first answered wake in a kernel process filed no diary row and never retried. That row drives the closer's re-audit: a goal whose judged wait passed the dead-man window takes a wake; when the session answers and the judges rule on the answer, the walk records the answer and files a same-why re-assert into the goal's diary, and the closer's filed-since gate reads it and re-audits with the answer in view. Without the row the closer never saw the answer, so the wait stood without a re-audit; a `frozen-store-write` judge-errors row was filed, and the shared store cache switched itself off for the rest of the process (`jd.shared_store_stats()` reads `off=1`, `poisoned=1`).

**Cause.** `kernel/kernel.py`: `_auto_nudge_session` binds `store` to the view from `jd.load_goals_shared_or_fault` and hands it to `_wake_goal`; `_wake_goal`'s answered leg passed that view to `_file_wake_answer`. The helper writes (`jd.record_verdict`, `jd.rollup_status`, `jd.save_goals`), so the first write (`nd.setdefault("log", [])` inside `record_verdict`) raised `FrozenStoreError` into the helper's own `except`, which logs to stderr and returns False; the poison side effect had already run. The other caller, `_awaiting_wake_outcomes`, passed `jd.load_goals(sid)` and was unaffected.

**Fix.** `_file_wake_answer(sid, gid, now)` takes no store from its callers and loads its own writer copy with `jd.load_goals` at the write moment. Every use of the store inside the helper is bound to the object that gets saved (the node it files on must belong to the store `record_verdict` appends to and `save_goals` publishes), so a caller-supplied store had no read-only role left, and dropping the parameter makes a write through a frozen view impossible at this site. Both callers change; `_awaiting_wake_outcomes` still costs exactly one writer load per filing. The walk's comment on which writers reload, and `_wake_goal`'s docstring, name the helper's own load, and the comment now says what a write through the view does (raises, files a `frozen-store-write` row, switches the cache off for the process) rather than that it fails loudly. No card-move rule changes.

## Tests

`tests/test_wake_answer_files_under_shared_view.py` (new) drives the real path on a file-backed store under a private synthetic sid, with the shared loader never stubbed: the walk itself (`_auto_nudge_session`, its store read through the shared cache, asserted by the cache counters), the answered leg handed the `FrozenStore` view directly (the view object stays untouched), and two answers in one process (each pass asserts the shared read, the row, and the cache still on). Each asserts that the row lands at the stamp's anchor with its why, that the anchor never moves, that the closer's filed-since gate opens on it, that `jd.shared_store_stats()['off']` is 0 with no poison count, and that no `frozen-store-write` row was filed. Without the change all three fail: no row, `off=1`, `poisoned=1`, one `frozen-store-write` row, and the helper's stderr carries the `FrozenStoreError` traceback.

`tests/test_nudge_gate_memo.py` drops its source-text pin on `_wake_goal`'s load: the pin matched the due leg's reload and stayed green with this regression present, so the behavioural module above replaces it, and the test's docstring now lists the helper's own load among the writers that reload. `tests/test_kernel_awaiting_stamp.py::AwaitingWakeOutcomeSweep::test_sweep_files_an_answer_through_one_writer_load`, which pins one writer load per filing for `_awaiting_wake_outcomes`, still passes.

Regression from https://github.com/romp-on/romp/pull/1141.

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -10029,7 +10029,7 @@ def _dead_wait_block(sid, gid, at, why, nudged, now, blk_why=None):
     return False
 
 
-def _file_wake_answer(store, sid, gid, now):
+def _file_wake_answer(sid, gid, now):
     """The answered wake's outcome becomes a FILED event (the user 2026-08-25, closing the awaiting
     audit's last live mechanism): the answer IS new information, and new information that files no
     diary row is invisible to every reader that matters — the response segment was often placed
@@ -10039,8 +10039,15 @@ def _file_wake_answer(store, sid, gid, now):
     and patience never churn) — the row's arrival moves _newest_filed past closerLookT, so the
     closer re-audits WITH the answer in view and rules it — done, lift, block, or keep — from real
     evidence. Runs once per answer by construction (the answered leg itself runs once per record).
-    Returns True when the row landed."""
+    Returns True when the row landed.
+    Takes no store from its callers: both hold read-only views (the walk's, _awaiting_wake_outcomes'),
+    and everything here is bound to the one object that gets saved (the node filed on must belong to
+    the store record_verdict appends to and save_goals publishes), so the helper loads its own writer
+    copy (jd.load_goals) at the write moment. A write through the shared view would raise
+    FrozenStoreError into the catch below, landing no row and switching the shared cache off for the
+    process; loading here rules that out by construction."""
     try:
+        store = jd.load_goals(sid)
         nodes = store.get("nodes", {})
         kids = {}
         for x, n in nodes.items():
@@ -10086,7 +10093,8 @@ def _wake_goal(sid, gid, stamp, nudged, turns, store, now, lt, tmux, wake_only=F
                 the stamp still stands → the wait was re-affirmed; keep the record with answeredAt so the
                 NEXT wake measures from the answer — NOT the anchor, which a same-why re-assert
                 deliberately freezes (judge apply_close's coalesce rule; keying episodes on the anchor is
-                what made the one-shot design terminal).
+                what made the one-shot design terminal), and file the answer into the diary
+                (_file_wake_answer, on its own writer load: `store` here is the walk's read-only view).
       silent    no response within the same window (from the FIRE, rec["at"]) → _mark_nudge_failed(wake=True):
                 the block rides the normal ladder to Needs-you. A failed/moot record re-arms only on a
                 genuinely NEW stamp episode (anchor newer than the one that failed).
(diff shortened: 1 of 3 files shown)


ksarma:
Merging as shipped. One wording note is filed as a follow-up.

Reviewed inline: `_file_wake_answer` takes no store and binds one to its own `jd.load_goals(sid)` at the top of the try. The nodes walk, `record_verdict`, `rollup_status` and `save_goals` all use that object, so the node it files on belongs to the store it saves. Its only two call sites, `_wake_goal`'s answered leg and `_awaiting_wake_outcomes`, pass `(sid, gid, now)`. Nothing else in kernel, tests or docs references it, so no path can hand it the frozen view. `_awaiting_wake_outcomes` still reads through `load_goals_shared`, and its filing costs one writer load; the unchanged pin in `tests/test_kernel_awaiting_stamp.py` confirms this. The new module uses a private synthetic sid, unlinks its override journal in tearDown, and never stubs the shared loader. It asserts the row at the stamp's anchor with its why, the anchor unmoved, the filed-since gate open, off 0, no poison delta, no frozen-store-write row and the view untouched. It also covers two answers in one process. The dropped source-text pin in `tests/test_nudge_gate_memo.py` was green at the base with the bug present, so the behavioural replacement is right. The docstring and walk comment now describe the helper's own load and the frozen-write consequence accurately. No card-move rule changes, the fix label fits, and the body and commit message match the diff.

**Follow-up**, `tests/test_wake_answer_files_under_shared_view.py:187`: the body says each test asserts the anchor, the why and the filed-since gate. The third test asserts only the row count, the shared read, off, poison and no frozen-store-write row. The first two carry the rest through `_assert_answer_filed_cleanly`. Tighten the wording or reuse that assertion in the loop.

**Test evidence**: in a scratch worktree at the base 28ecee8c with the PR's test files checked out, `tests/test_wake_answer_files_under_shared_view.py` failed all three tests (0 filed rows, off=1, poisoned=1, one frozen-store-write row, FrozenStoreError traceback on stderr). At head 07ff2b3e it passes. With `tests/test_nudge_gate_memo.py` and `tests/test_kernel_awaiting_stamp.py`, the three files give 70 passed. CI at 07ff2b3e: all eight required checks green (Python 3.10 to 3.13, bats, vscode-extension, Exactly one tier label, Tier policy).

Review run by the maintainer automated review session; verified against head 07ff2b3e.

ksarma:
Addendum: the PR merged at f710d01d (merge 557a2430, 2026-09-09 09:12Z). The only change after the reviewed head 07ff2b3e is a merge of main into the branch; the PR's own diff is unchanged, so the record above stands for the merged head. Review run by the maintainer automated review session; verified against head f710d01d.

PR #1161, opened 2026-09-09
Courier scan: a peer segment placed under a drifted key is not queued

henrypinkard:
Tier: fix

## What

The courier's scan applies the placement loop's drift-tolerant placed check before it queues a peer segment. Completes the courier change gate's criterion that goal loads per triage pass fall.

**Measured on the merged gate's first readings** (3.5 and 6.5 min uptime on the shared box, no synthetic load): goal loads per triage pass did not fall (1103 and 1058 against 1062 to 1217 before), the judge thread stayed at 58% to 61% of a core, and the gate barely recorded (scanned 677, skipped 135, recorded 5 over 27 passes). A 60 s py-spy sample put 100% of the courier's remaining writer loads on one line: the placement loop's per-row `store = load_goals(fsid)`. The judge error log had no courier rows, so nothing was deferred or failing.

**Cause.** The scan tests placement by exact segment id (`seg["id"] in placed_ids`); the placement loop tests it with `_placed_key`, the timestamp-invariant rule (a segment whose parse t drifted after its placement was recorded still counts as placed). A peer segment placed under a drifted key therefore passed the scan as unplaced, was queued, cost a writer load in the placement loop, and was dropped there unwritten by `_placed_key`, every pass. Its session had "rows to place" every pass, so the change gate never recorded it: the win was absent for exactly the sessions with drifted placements, which on this box is most of them.

**Fix.** A peer segment with a known sender that `_placed_key(placed_ids, seg["id"])` reports as placed is skipped in the scan (the same call the placement loop makes, without `live`, on the same store content). No row, no writer load, and the session records like any settled one. The exact check stays first (the common case, a set lookup), so `_placed_key` runs only for peer segments the exact check missed, which the placement loop paid for anyway.

No card-move rule changes: those rows were dropped unwritten every pass already; the outcome for every segment is identical, minus the load.

## Review (by hand; the change is five lines)

Cost: `_placed_key` now runs in the scan only for peer segments with a known sender that the exact check missed, the rows that reached the placement loop's `_placed_key` before, so no new cost. Same rule: the two-argument form the placement loop uses, on the same store content (the view; a write landing between the two reads moves the gate's key for the next pass). Repair: a drifted-placed segment never had a repair path (the repair branch is the exact-hit branch, and `_courier_link_target` looks the placement up by exact id), so none is lost. Twins: without `live`, `_placed_key` treats a byte-identical prompt in another turn as a drifted twin, exactly as the placement loop already did for these rows; the planner's own dedup (which passes `live`) is untouched.

## Tests

`tests/test_courier_skip.py`: a segment placed under a shifted key (the same segment recorded seven seconds later) is not queued, takes no writer load, and its session records on the first pass and skips on the second, with the store untouched. Red without the fix (the session had a row to place and never recorded), green with it.

## After

Posted below after the deploy restart: goal loads per triage pass, judge-thread CPU share and the gate's counters at matched uptime, against the readings above.

### DIFF
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -15118,6 +15118,12 @@ def run_courier(now=None, sessions_cap=PLAN_SESSIONS, concurrency=None, verbose=
                     # fired before). Defense in depth beside the fork's sealed-placements seed.
                     continue
                 pm = _seg_peer(seg)
+                if pm and pm[0] and _placed_key(placed_ids, seg["id"]):
+                    continue                           # placed under a DRIFTED key (the parse's t shifted after the
+                    #                                    placement was recorded): the placement loop's own _placed_key
+                    #                                    check dropped the row unwritten every pass, at a writer load per
+                    #                                    row per pass, and the row kept its session from ever recording
+                    #                                    in the change gate (2026-09-09). The same rule, applied here.
                 if not pm or not pm[0]:                # peer-triggered with a KNOWN sender only. This filter
                     #                                    is one half of a partition contract with plan_units:
                     #                                    the courier places exactly the peer segments it can
--- a/tests/test_courier_skip.py
+++ b/tests/test_courier_skip.py
@@ -303,6 +303,39 @@ class CourierSkip(_World):
         self.assertEqual(self.run_pass(), (3, 0, 3), "second pass: nothing to place, no repair open, all recorded")
         self.assertEqual(self.run_pass(), (0, 3, 0), "third pass: all skipped")
 
+    def test_a_segment_placed_under_a_shifted_key_is_not_queued_and_the_session_records(self):
+        # DRIFT (measured live, 2026-09-09): a peer segment whose parse t shifted after its placement was
+        # recorded is no exact member of placements, so the scan queued it every pass; the placement loop
+        # then loaded the store (a writer load per row per pass) and dropped the row through _placed_key
+        # unwritten. Every such session stayed unrecorded, and the courier's loads per pass did not fall.
+        # The scan applies the same drift-tolerant rule: no row, no load, the session records and skips.
+        path = self.proj_dir / (A + ".jsonl")
+        session = jd.parsed_session(A, [str(path)], T0 + 200)
+        fresh = jd.load_goals(A)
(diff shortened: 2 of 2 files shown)


henrypinkard:

Live kernel on the shared box, `/perf` readings, no synthetic load. "Before" rows are the courier change gate as merged (the readings on its own pull request, same hour, same sessions); "after" rows are this fix on top of it. Loads per judge pass is `goals.loads` over `judge.passes` since the kernel's start.

| code | uptime | goal loads / judge pass | judge thread CPU (% of a core) | process CPU | pusher thread CPU | courier gate skipped / scanned / recorded | idle cycles (share, wall each, pusher CPU each) |
| before | 3.5 min | 1103 | 58% | 80% | 15% | 65 / 341 / 5 | 87%, 289 ms, 75 ms |
| before | 6.5 min | 1058 | 61% | 78% | 11% | 135 / 677 / 5 | 89%, 302 ms, 69 ms |
| before | 19.3 min | 1017 | 71% | 84% | 11% | 450 / 2189 / 5 | 84%, 349 ms, 70 ms |
| after | 3.8 min | 285 | 42% | 61% | 13% | 784 / 57 / 56 | 82%, 90 ms, 60 ms |
| after | 6.5 min | 267 | 37% | 54% | 11% | 1295 / 68 / 67 | 85%, 79 ms, 56 ms |
| after | 20.4 min | 242 | 40% | 53% | 9% | 4877 / 111 / 109 | 85%, 83 ms, 57 ms |

**The gate records now.** Sessions recorded went from 5 in 90 passes to 67 in 47 passes, and skips outnumber scans twenty to one: a triage pass scans one or two sessions instead of every one. Goal loads per triage pass fell by three quarters at every mark (242 against 1017 at 20 min), the judge thread's CPU from 71% to 40% of a core, the process's from 84% to 53%, and the pusher's jobs stage sits at a fifth of wall.

**An idle pusher cycle got cheaper too** (79 ms of wall against 302), though the pusher was not this change's target: the judge thread's loads held the process's file and lock time, and the pusher's cycle waited on them. The idle share itself is unchanged (82% to 85% of cycles), so the cadence question stands as measured: at about one cycle a second, idle cycles cost the pusher thread about 56 to 60 ms of CPU each, 5% to 6% of a core, the upper bound on what an event-driven wait would save there. The one-hour reading follows.

PR #1162, opened 2026-09-09
Codex backend: prune_live takes the kernel's four arguments, and its echo atoms carry the input-echo marker

ksarma:
**Symptom.** Any live merge of a Codex session holding an input echo (a send queued behind a busy turn, a steer in flight, a send the app-server never recorded) raises `TypeError: CodexBackend.prune_live() takes from 3 to 4 positional arguments but 5 were given`. The chat build of that session fails (`build_session`), the feed build fails for every session while the echo exists and that session's parse is cached (`build_feed` merges inside its loop), and the timeline bars log a live-merge failure and fall back to disk-only bars. The failure lasts as long as the echo does.

**Cause.** `_merge_live_atoms` calls `be.prune_live(sid, tx_uuids, tx_text_t, human_floor)`. `SdkBackend.prune_live` and the tmux backend's take the fourth argument; `CodexBackend.prune_live` and the `SessionBackend` ABC declare three. Nothing checked the shape: the backends duck-type the ABC, and the conformance test checks that methods exist, not what they accept.

Once the call binds, a second problem shows. `CodexBackend.live_atoms` emitted its echo atoms without `_echo_text`, the key the merge uses to recognise an input echo: it hides an echo behind its queued bubble or its landed record only when the key is present, and counts an atom without it as live assistant work, forcing the last turn open. So a queued Codex send would paint twice, and a session whose only live item is a pending send would show a working chip.

**Fix.**
- `CodexBackend.prune_live(sid, tx_uuids, tx_user_texts=(), human_floor=0)`, and the ABC declares the same four with the kernel's call shape in its docstring. An echo retires by uuid, or by text under `echo_text_key` on both sides; when the caller passes the kernel's text-to-newest-record-time mapping, only through a record written at or after the echo's send (the rule the SDK backend applies: a repeated "ok" must not retire the newer echo before its record); a plain set keeps the unfloored match. `human_floor` is accepted and retires nothing here: no Codex echo carries the `command` flag the SDK's floor is for, and no floor retires a plain echo.
- `send()` stamps the echo in whole seconds like the SDK and tmux echoes: record times are whole seconds, and a float stamp would keep an echo whose record was written later in the same second. The text is stored under `echo_text_key`, imported the way `sdk_backend.py` imports it (the kernel's copy of `session_backend` when it is loaded, else the file under its own module name).
- `live_atoms` marks each echo with `_echo_text`.

**Visible change.** A mid-turn steer now reads as a pending bubble until the app-server records it, because `build_session`'s queued-bubble pass takes every unlanded echo of a backend without `unqueue` while the session is busy. A queued send is already shown from the queue and is not doubled.

**Tests, each failing before the change.**
- `tests/test_backend_call_parity.py` (new, static: an AST walk over the sources, no kernel import): every call the kernel makes on a `Sessions.backend_for` backend binds on every backend that defines the method; the `prune_live` call is four positional arguments and binds on all three backends; every backend accepts at least the ABC's shape; the ABC declares the floor. On main three of four fail: `kernel.py:28854 be.prune_live(4 positional) does not bind on CodexBackend.prune_live`, and `SessionBackend.prune_live` declares 3, not 4.
- `tests/test_codex_backend.py`: `PruneLive` (the four-argument call with no floor retiring a plain echo; a text landing only through a record at or after the send; a record later in the send's own second; the key rule on both sides, mapping and plain set; retire by uuid; unknown sid) and `EchoAtoms` (the marker). On main every four-argument call raises `TypeError`. With only the two `def` lines changed, four stay red: the mapping-floor test (a 999.0 record retires a t=1000 echo), the same-second test (the float stamp), the both-sides key test (the raw echo text compared against a stripped key), and the marker test (`KeyError: '_echo_text'`).
- `tests/test_codex_echo_merge.py` (new): the real `_merge_live_atoms` over the real `CodexBackend` with a scripted app-server client. An echo behind its queued bubble adds no atom and leaves the turn ended; an echo alone paints once and never forces the turn open; a record in the send's second retires the echo and the text paints once. On main all three fail (two on the `TypeError`, the third on the float stamp before it reaches the merge). Without the marker all three fail; without the whole-second stamp the third does.

**Not in this PR.** Several sends queued before a turn starts go out as one turn, and the normalizer joins them into one text block, so none of their echoes retires; that is a separate fix with its own test, filed next. The Codex model picker opening blank (an empty catalog cached for the process, no reason on `/models`, no models frame when the first Codex session opens the list, no reason row in the picker) is separate work in the kernel and the picker. This PR touches no `kernel.py`, no UI and no docs.

Tier: fix

### DIFF
--- a/kernel/codex_backend.py
+++ b/kernel/codex_backend.py
@@ -40,6 +40,14 @@ from pathlib import Path
 HERE = Path(os.path.dirname(os.path.realpath(__file__)))
 _events = SourceFileLoader("romp_codex_events", str(HERE / "codex_events.py")).load_module()
 _runtime = SourceFileLoader("romp_codex_runtime", str(HERE / "codex_runtime.py")).load_module()
+# The by-text KEY RULE (session_backend.echo_text_key): the one normalization under which an input echo's
+# text is compared with a transcript record's, shared with the kernel's _atom_user_texts and
+# SdkBackend.prune_live, so an echo whose text carries a trailing newline still lands. The kernel's own
+# copy of that module when it is loaded (kernel.py loads it as romp_session_backend, and TmuxBackend
+# subclasses that copy's ABC); otherwise the file is loaded under its OWN module name, as sdk_backend
+# does, so re-executing the source never rebinds the ABC out from under a subclass.
+echo_text_key = (sys.modules.get("romp_session_backend") or SourceFileLoader(
+    "romp_session_backend_keys", str(HERE / "session_backend.py")).load_module()).echo_text_key
 
 SDK_PIN = "openai-codex==0.144.4"     # bin/romp-codex-setup installs exactly this into codexvenv
 SETUP_HINT = ("Session not created: the Codex backend isn't installed. "
@@ -769,7 +777,12 @@ class CodexBackend:
         with s.lock:
             if s.dead:
                 return False
-            s.echoes.append({"text": text.strip(), "t": time.time(),
+            # WHOLE seconds, as the SDK and tmux echoes stamp theirs: record times are parse_z's int
+            # seconds and prune_live lands an echo by text only through a record at or after its send,
+            # so a float stamp would keep an echo whose record was written later in the same second. The
+            # text is stored under the shared key rule (echo_text_key), the key prune_live compares
+            # against; for a str that is the stripped text _append matches.
+            s.echoes.append({"text": echo_text_key(text), "t": int(time.time()),
                              "uuid": "echo-%s" % uuidlib.uuid4().hex[:8]})
             turn_id = s.turn_id
             tid = s.tid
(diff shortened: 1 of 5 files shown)


ksarma:
Merging as shipped. One follow-up below; nothing to fold in.

Reviewed inline: the kernel's one `prune_live` call site passes four positional arguments. The SDK, tmux and Codex backends and the `SessionBackend` ABC now declare the same four. `TmuxBackend`, the only ABC subclass, already matched. The Codex retire rule follows `SdkBackend.prune_live`. It retires by uuid, or by text under `echo_text_key` on both sides. When the caller passes the text-to-record-time mapping, it retires through a record stamped at or after the send. A plain set is unfloored. `human_floor` retires nothing, since no Codex echo carries the command flag. The whole-second stamp in `send` matches `parse_z`'s integer record times. The `_append` retire still matches, because it compares stripped record text against the stored key. The `echo_text_key` import is the same `sys.modules` probe `sdk_backend` uses over a pure strip, so two module copies cannot disagree. The `_echo_text` marker makes the merge hide the echo behind its queued bubble and stop counting it as live work. The declared visible change comes from `build_session`'s queued-bubble pass, which is chat rendering. An echo-only merge keeps the turn's ended state, so no card-move rule changes. The parity test is a static AST walk with no kernel import. The Codex tests stub the client factory and start no app-server. The added lines carry no private strings, hostnames, home paths or verbatim quotes. The body and commit message match the diff.

Follow-up, `kernel/codex_backend.py:785`: a whitespace-only send passes the `not text` check and is stored as the empty key. Its atom then carries an empty `_echo_text`. The merge reads that as live work, and `prune_live` can never retire it by text. The defect is pre-existing in effect: the merge raised before this PR. Refusing blank text in `send()` closes it.

Red-first, by my own run: the PR's three test files against origin/main gave 13 failed and 65 passed. The failures were parity 3 of 4, all six PruneLive tests and EchoAtoms, and all three CodexEchoMerge tests. At head 8cb2904f the same files gave 78 passed. CI at head has all 11 checks green, including the tier checks, and the PR is mergeable and clean.

Review run by the maintainer automated review session; verified against head 8cb2904f.

PR #1166, opened 2026-09-09
Codex backend: several queued sends land one text block each, and each retire takes back one echo per send

ksarma:
**Symptom.** Two sends queued before a Codex turn starts (an idle session sent twice quickly, sends while the app-server client is down or backing off, a queue resumed after a kernel restart) go out as one turn, and their input echoes never retire. They stay live for as long as the session does: pending sends while the session is busy, user atoms beside the landed record when it is idle, in every later build. Two related over-retires: a send that finds its session dead during a steer takes back every echo carrying its text, so an earlier same-text send the app-server never recorded, which must stay visible, vanishes with it; and the backend's own retire on a landed record drops every echo carrying the record's text, so a text sent twice loses both echoes when the first record lands.

**Cause.** `CodexBackend._run_turn_in_mode` (kernel/codex_backend.py) starts one turn from the whole queue, one input per queued send, and the app-server answers with one userMessage item carrying them all. `_user_input_text` (kernel/codex_events.py) newline-joined those inputs and `ThreadNormalizer._item` wrote them as one text block, `"first send\nsecond send"`, which matches neither send's echo: `CodexBackend._append` keyed the record by its space-joined text, and the kernel's `prune_live` receives the keys `_atom_user_texts` yields (the joined text and each block), where the only block was the joined one. `send()`'s dead path retired by text.

**Fix.**
- `codex_events._user_input_texts` returns one stripped entry per input (blanks dropped; an image or skill placeholder is an entry of its own), and `_item` writes one text block per entry. A user record with several text blocks is a shape the kernel already reads per block (`_atom_user_texts`, behind the echo prune and the pending-bubble pass; `sdk_backend._landed_texts`): romp bundles its own injected messages as several blocks in one record, so no kernel change is needed.
- `CodexBackend._rec_texts` keys each text block of a user record under `echo_text_key` (a string content is one entry, blanks dropped), and `_append` retires one echo per landed block, the oldest carrying the text, so a second identical steer keeps its echo when the first one's record lands and a second send dropped after that stays visible. The kernel's `prune_live`, floored by record time, is the other retire; it also matches the record's joined text, which `_append` does not repeat.
- `send()` keeps the uuid it mints and, when the session died during the steer RPC, takes back that echo alone.
- `plans/codex-backend.md`'s event table says one text block per input. The test fake client's default turn answers a userMessage carrying the input list, the app-server's shape.

**Tests, each failing before the change.**
- `tests/test_codex_events_golden.py` `test_several_inputs_land_as_one_text_block_each`: two inputs land as two stripped blocks, a blank input is dropped, placeholders are blocks of their own, one input writes one block as before, all-blank inputs write no record. Before: one block, `"first send\n  second send"`.
- `tests/test_codex_backend.py` `Lifecycle.test_two_sends_queued_before_the_turn_land_as_two_blocks_and_retire_both_echoes`, through the real worker: one `turn_start` with two inputs, one user record with two blocks, both echoes retired. Before: one joined block, both echoes live.
- `tests/test_codex_backend.py` `EchoAtoms`: a landed record retires one echo, the oldest carrying its text (before: both `"ok"` echoes went on one record); a two-block record retires one echo per block, the oldest of each, and a later repeat of the first text stays (before: the space-joined key matched no echo and every echo stayed); a joined-text echo does not match a two-block record (before: the space-joined key retired the echo spelling both texts in one message); a send that finds the session dead takes back only its own echo (before: the earlier same-text echo went with it).
- `tests/test_codex_echo_merge.py` `test_a_two_block_record_in_the_sends_second_retires_both_echoes_and_paints_each_text_once`: the real normalizer and the real `_merge_live_atoms` over the real backend. Before: both echoes live after the merge.

Follows https://github.com/romp-on/romp/pull/1162 (merged), which this change builds on: the key rule `send()` stores on an echo and `prune_live` compares, and the `_echo_text` marker the tests read, come from that PR.

Tier: fix

### DIFF
--- a/kernel/codex_backend.py
+++ b/kernel/codex_backend.py
@@ -705,19 +705,43 @@ class CodexBackend:
         with open(path, "a", encoding="utf-8") as f:
             for r in recs:
                 f.write(json.dumps(r, ensure_ascii=False) + "\n")
-        # a landed user record replaces its optimistic echo (uuid-independent: match by text)
-        landed = {self._rec_text(r) for r in recs if r.get("type") == "user"}
+        # A landed user record replaces its optimistic echoes (uuid-independent: match by text, under
+        # echo_text_key on both sides). ONE echo per landed text BLOCK, the OLDEST carrying the text. A
+        # block is one send: the worker starts a turn from its whole queue, one input per queued send, and
+        # the normalizer writes the app-server's one userMessage item as one record with a block per input
+        # (codex_events._user_input_texts), so a turn started from two queued sends lands both echoes here;
+        # a one-send record is one block. One per key, not every echo carrying it: a second identical
+        # STEER (delivered mid-turn, never queued) keeps its echo when the first one's record lands, and a
+        # second send dropped after that (the client dying mid-queue) stays visible. The kernel's
+        # prune_live is the other retire, floored by record time; this one sees only the records it just
+        # wrote.
+        landed = [t for r in recs if r.get("type") == "user" for t in self._rec_texts(r)]
         if landed:
             with s.lock:
-                s.echoes = [e for e in s.echoes if e["text"] not in landed]
+                kept = list(s.echoes)
+                for text in landed:
+                    for i, e in enumerate(kept):
+                        if echo_text_key(e.get("text")) == text:
+                            del kept[i]
+                            break
+                if len(kept) != len(s.echoes):
+                    s.echoes = kept
 
     @staticmethod
-    def _rec_text(rec):
+    def _rec_texts(rec):
+        """A user record's texts under the shared key rule (echo_text_key: outer whitespace stripped,
+        nothing else), the key send() stores on the echo and _append and prune_live compare: one entry per
+        text BLOCK (a string content is one entry), empty ones dropped. Per block and never the blocks
+        joined: each block of a Codex user record is one send (see _append), and this retire takes exactly
+        one echo per send. The kernel's prune_live also matches the record's space-joined text
+        (_atom_user_texts yields the joined text and each block), floored by record time; that match is
+        not repeated here."""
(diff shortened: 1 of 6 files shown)


ksarma:

Reviewed inline: the worker starts one turn from its whole queue, one input per queued send. The normalizer now writes the app-server's single `userMessage` as one record with a stripped text block per input. Blanks are dropped; placeholders get their own block. `_rec_texts` keys each block under `echo_text_key`. `_append` removes the first echo carrying each landed block, in order. So a two-block record retires exactly two echoes, oldest first, and a later repeat of a text keeps its echo. The dead-session path in `send()` retires by the uuid it minted, not by text. No kernel change is needed. `_atom_user_texts` already yields the joined text and each block. `_merge_live_atoms` builds `prune_live`'s record-time map from it. The chat ships per-block blocks for such records. An echo `_append` removes is gone before `prune_live` runs. `prune_live`'s floor and joined-text match are unchanged from #1162, so nothing retires twice. Fixtures are synthetic; the privacy scan of the added lines is clean. The tests stub the client factory and start no app-server.

Follow-up, not a blocker: the chat bubble's text is the blocks space-joined. A two-send Codex turn now reads on one line where the joined block read on two. The SDK backend's bundled records already render that way. A break between blocks for both backends would be its own change.

Tests: a scratch worktree at base 557a2430 with the PR's three test files gave 7 failed / 101 passed. The seven are exactly the named tests: golden several-inputs, Lifecycle two-sends, the four EchoAtoms cases, echo-merge two-block. Each failed for the reason the body gives. At head bda83b1e, 108 passed. CI at head is green on every check; merge state clean.

Review run by the maintainer automated review session; verified against head bda83b1e.

PR #1167, opened 2026-09-09
GET /models names why the Codex model list is empty; the backend caches only a non-empty catalog

ksarma:
**Symptom.** A Codex session's model picker opens on a blank menu with no word of why, and stays blank for the life of the kernel. `GET /models` carries `codex: {models: []}` in every empty state, and the picker has nothing else to read.

**Cause.** `CodexBackend.model_catalog()` caches whatever its first fetch produces. An empty page `is not None`, so a first answer of `[]` holds the empty catalog for the process, and later picker opens show the blank menu after the app-server has models to list. Two other paths to `[]` are silent to the caller: a client in its retry backoff (nothing recorded, nothing logged) and a `model_list` that raised (logged on every read, unreadable by the caller). The `/models` handler swallows a raise out of `model_catalog()` into the same empty list, and it serves the same `[]` where the Codex consult is gated off (no live Codex session, Codex neither the default backend nor the judge engine) and where the backend module failed to load.

**Fix.** `model_catalog()` caches only a non-empty list. It records why it answered `[]` (the client's failure text, `model_list`'s error, or an empty page) and logs each reason once per distinct reason, since the route is re-read on every picker open and every models frame. The read runs under a lock: `/models` is served from handler threads, and without it two readers racing the same first read could both log the reason, or one that found no client could record its reason after the other had stored the list and cleared it. `model_catalog_error()` returns the recorded reason, `None` while a catalog is held or before one was asked for.

The handler's `codex` section gains `error`, always present: `null` beside a non-empty list, else one sentence (a string, never an object or a code) naming the reason:
- after an empty answer, the backend's recorded reason, or `the Codex app-server sent no model list` when none was recorded;
- when the catalog read raised (`model_catalog` or `model_catalog_error`), `model catalog: <error>`, logged to the kernel log once per distinct reason under the backend's `codex-backend:` prefix, and again when the same fault recurs after the read answered;
- the closed gate: `no live Codex session; the list is read once one runs`;
- an absent backend: `the Codex backend is unavailable (see the kernel log)`.

The last two name themselves so an empty list is never read as the app-server's answer. The gate itself is unchanged.

**Tests** (each red before the change).
- `tests/test_codex_backend.py`: an empty first page is not cached and is named (before: the second read serves the cached `[]`); a `model_list` fault is readable through `model_catalog_error` and logged once per distinct reason (before: the method does not exist, and the fault logs on every read); a failed client factory names the client failure (before: nothing recorded or logged by `model_catalog`); the held-catalog test also asserts a null reason.
- `tests/test_codex_models_route.py` (new): the handler in-process on a loopback server over a fake backend, and once over the real clientless `CodexBackend`: a live session gets the list and a null `error`; the backend's reason, the fallback sentence, a raise out of the catalog, a raise out of the reason read, the closed gate, the default-backend gate and an absent backend each carry their sentence; a raising catalog logs once per distinct reason and the same fault again after a recovery (before: the section has no `error` field, and every test fails on it). The module resets the kernel's log latch around each test, so no test depends on the order the others ran in.

Full suite: 10034 passed, 61 skipped.

**Not in this PR.** The models frame when the first Codex session opens the gate (so an already-open dashboard re-reads the list) and the picker's reason row in the UI are separate work; this PR's `error` field is the value that row will read.

Tier: fix

### DIFF
--- a/kernel/codex_backend.py
+++ b/kernel/codex_backend.py
@@ -929,8 +929,12 @@ class CodexBackend:
     def model_catalog_error(self):
         """Why the last model_catalog() answered [] (one sentence for a picker to show), or None when
         a catalog is held or none has been asked for yet. The kernel's /models reads it after an empty
-        answer; the failure is the app-server's or the client's, so the sentence names that side."""
-        return self._catalog_err
+        answer; the failure is the app-server's or the client's, so the sentence names that side.
+        Read under _catalog_lock like every writer of the reason, so a concurrent read that is mid-way
+        through storing a list or recording its own reason cannot hand this caller a half-updated value
+        (review find, 2026-09-09)."""
+        with self._catalog_lock:
+            return self._catalog_err
 
     def set_mode(self, sid, mode):
         s = self._session(sid)
--- a/kernel/codex_backend.py
+++ b/kernel/codex_backend.py
@@ -285,7 +285,9 @@ class CodexBackend:
         self._client_retry_at = 0.0
         self._client_failures = 0
         self._client_generation = 0   # successful app-server client installations
-        self._catalog = None          # model_catalog() cache — fetched once per process
+        self._catalog = None          # model_catalog() cache: a NON-EMPTY list, fetched once per process
+        self._catalog_err = None      # why the last model_catalog() answered [] (str), or None
+        self._catalog_lock = threading.Lock()   # one model_catalog() read at a time (see its docstring)
         self._client_lock = threading.Lock()
         self._sessions = {}           # sid → _Session
         self._sessions_lock = threading.RLock()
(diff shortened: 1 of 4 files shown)


ksarma:

Reviewed inline: `model_catalog()` returns the cache only when it is a non-empty list, so a first empty page is re-asked on the next read. Each of the three empty paths (no client, `model_list` raised, empty page) records a sentence readable through `model_catalog_error()` and logs it once per distinct consecutive reason. The read sits under the new `_catalog_lock`. `model_catalog()` is its only taker and the `/models` handler its only caller, so the order catalog, then client, then sessions has no reverse edge and cannot deadlock. The handler's codex section carries `error` on every branch as a string or null, never an object. The gate expression is unchanged. Both UI readers (`render.ts`, `romp-timeline-view.js`) read only `models` and `efforts` behind `Array.isArray`, so older clients ignore the key. The kernel latch clears on any non-raising read, so the same raise after a recovery logs again. The route module resets the latch on its own privately named kernel module, so nothing leaks across modules under xdist. The route test binds 127.0.0.1 port 0, sets the backend to False in setUp, and its one real `CodexBackend` uses a raising `client_factory`, so no app-server starts. No `ui/` file reads `codex.error` yet. After this merge a picker blanked by a first empty page fills on a later open. A list that is still empty shows a blank menu, with the reason on `GET /models` and once in the kernel log, as the body's "Not in this PR" section says.

Two notes for the record, neither a hold:
- `tests/test_codex_models_route.py:79`: at the base every route test errors in setUp on the missing `_codex_catalog_fault` attribute, before reaching the error-field assertion the body names. The test is red on a PR-added symbol either way. Follow-up: record only.
- `kernel/kernel.py:14182`: `_note_codex_catalog_fault` compares and sets outside any lock, so two handler threads raising the same fault at once can each log the line. Once per reason holds for sequential reads. Follow-up, cosmetic.

Tests: at the base 3e4398a3, in a scratch worktree with the PR's two test files checked out, 14 failed and 75 passed. The 14 are the four `model_catalog` cases in `tests/test_codex_backend.py` and all 10 in `tests/test_codex_models_route.py`. At ab07bd64 the same files pass, 89 passed. CI at ab07bd64 is green on every check, and the Tier policy check reports the fix gate. Privacy and vocabulary scans of the added lines are clean.

Head note: after this review the author pushed 1eb63ca8, a four-line change that reads the catalog reason under `_catalog_lock`, the unlocked read noted above. Read at that head: the lock is a plain `threading.Lock`, `model_catalog()` never calls `model_catalog_error()` while holding it, and the handler's two calls stay separate, so no nested acquisition. The PR's two test files pass at 1eb63ca8 (89 passed).

Review run by the maintainer automated review session; verified against head 1eb63ca8.

PR #1168, opened 2026-09-09
Tab strip: one tag group per row becomes a gear setting, on by default

ksarma:
With the tabs grouped by tag, every tag group still starts on its own row (#1087). A new gear checkbox, on by default, lets a user turn that off so the groups follow one another across the strip and wrap as they need.

## Why: the many-groups case, in the strip's own numbers

- One group per row means at least one row per tag plus one for the untagged trail. The strip's cap is 150px, about four rows (`TABBAR_H_DEFAULT` in tabbar-resize.ts; a row measured 31.67px in #1108). Four rows fill it; whatever comes fifth, another tag or the untagged trail, sits below the cap, reachable only by scrolling the strip or dragging the grip, whose ceiling is 60% of the window.
- A dozen tags is thirteen rows, roughly 410px: nearly three times the cap, and a large share of a laptop window even at the grip's ceiling. Folding does not shorten it, because a folded group is still its header row alone (a #1087 design point): twelve folded tags are twelve rows of one chip each, where inline they share a row or two at desktop width. A session under several tags (#1091) has a copy in every group and adds wrap rows to each.
- The strip is the switcher the user glances at, and groups below the cap are groups they cannot see. #1087 noted that many groups scroll inside the cap; this PR is for the strips where scrolling the switcher is the problem. The per-row layout stays the default because it reads best at a handful of tags.

## Design points

- Default kept: `stripGroupRows` defaults to true. A settings store written before the key reads true (loadSettings spreads the defaults); no migration, no kernel involvement.
- Where it lives: RompSettings in settings.ts and the browser-side `romp:settings` store, like every gear row (per browser profile; synced to same-origin tabs by the storage event and to sibling VS Code panes by the existing settingsSync post). The row sits in the gear's Chat section, after Show git branch and before Context gauge in tabs, so the two strip rows sit together. It is not in the tag menu's foot: the Group tabs by tag row there switches sectioning itself, is stored in the romp:tabgroups blob, and rides a menu shared with mounts that have no strip.
- Off is the strip as it was before #1087. The row breaks are emitted only under the setting, and the untagged trail stands behind the 13px divider again (`makeTrailSep`, the pre-#1087 element and its rule), scoped `.tab-group-sep:not(.tab-group-break)` so the trail's break keeps no box. The divider returns only in this layout, because there the trail has no row of its own to say "in no tag". Its gutters are padding, so the drag's virtual layout measures its footprint (the dragover already selects `.tab-group-sep` and measures a non-break box live); it is a group edge for the drop's in-group neighbour walk (#1091) and the boundary `sectionHeadOf` reads, as it was.
- The row hairlines (T134) are unchanged: `#tabs` is `align-items: stretch` and the trail's boundary exists only when loose tabs do, so the divider always shares a row with a tab and never forms a row alone; the painter's tabs-and-headers rule and its pin stand. If you would rather count the divider as the pre-#1087 painter did, it is one clause in the painter's filter plus the pin.
- A flip repaints at once: the setting rides the strip's rebuild signature (tab-strip-skip pins it; the executed harness drives it), and the existing settings listener already calls renderTabs.
- Unchanged in both states: keyboard order and focus, ARIA (the break stays aria-hidden; the divider carries a title only; neither is focusable), the phone layout (flat strip), the mobile scrape (`.tab` only), the copies of a multi-tag session, the resize grip and cap, and the Group tabs by tag switch.
- Deliberately left out: no automatic switch by group count (a threshold would decide for the user; the checkbox is their call). The opt-out layout keeps its pre-#1087 wrap, in which a header can end a row while its tabs start the next; a keep-with-next break is a natural follow-up if that bothers anyone. docs/reference.md is unchanged (it documents kernel-side configuration). The guide's tab-groups paragraph now describes both layouts: its opening still said the untagged sessions sit after a divider, which #1087 had retired, so it now says they take a row of their own, and the sentence about the setting names the divider for the inline layout. The kernel is untouched.

## Tier

Tier: feature

## Tests

- `ui/webview/tab-groups.test.ts` runs the head loop's break gate at both values (on: a break ahead of every header but the strip's first item and never the trail's; off: none), runs the drop's in-group neighbour walk, lifted from the handler, over a fake strip (a tab dropped at the trail's head anchors on its trail neighbour and never on the tagged group's last tab, with the divider as the edge and with the break), and pins the trail branch, the divider builder, the scoped divider rule, the signature entry, the settings listener, the gear row and its wiring, and the guide's two sentences.
- `ui/webview/settings.test.ts` pins the default and the round trip of an explicit false and of a store from before the key.
- `ui/webview/tab-strip-skip.test.ts` lists the new signature input; `ui/webview/tab-strip-skip-exec.test.ts` executes the flip as exactly one repaint.
- `ui/webview/tab-drag-live.test.ts` pins the divider's footprint (width and padding, no margin) and the absence of an unscoped rule.
- `tests/test_kernel_settings_sections.py` places the row under Chat and pins the gear's default in both of load()'s literals, its listener and its fill.
- `tests/test_tab_groups_rows.py` drives the served /chat page in both layouts: the default with nothing written to the settings store, and the inline layout with the setting written off before load (no break, one visible 13px divider, each group's tabs contiguous after its header, no zero-height item besides the hairlines). It skips loudly without a browser, so the local run is recorded here: both drives green, `2 passed in 6.50s`. Before the change the inline drive found three breaks.
- Ran locally: `cd vscode-extension && npm run typecheck && npm test` (typecheck clean; 3651 tests, 3651 pass); `python3 -m pytest -q tests/test_kernel_settings_sections.py tests/test_tab_groups_rows.py` (9 passed; 2 passed); the rest of the Python suite with `-n 8`, the served guard run separately: 10054 passed, 40 skipped.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -85,8 +85,8 @@ instead; a PDF too large to show offers a download in its place.
 several. Right-click a tab and open **Tags** to add or remove them. Tags filter every
 surface (the tag button in the strip narrows the tabs to the tags you pick), and they group
 the tabs: as soon as any session carries a tag, the strip shows one section per tag, in your
-tag order, each with a header in the tag's color, and the untagged sessions after a divider
-at the end. A session with several tags appears under each of them; every copy is the same
+tag order, each with a header in the tag's color, and the untagged sessions on a row of their
+own at the end. A session with several tags appears under each of them; every copy is the same
 session (click either to open it, and closing either ends it). Each header shows a chevron, the tag's color, its name, and a
 member count. Click a header, or press Enter on it, to fold its section down to the header
 alone; the count then says how many tabs are folded away, and a small dot after it says when
@@ -102,7 +102,10 @@ reorders the tags on every surface (the timeline's tag table shows the same orde
 a tab into another group, right-click it and pick **Move to <tag>** under **Tags**: one click
 adds that tag and drops the tag of the group you right-clicked it in, leaving its other tags alone. The row's
 **+** adds the tag without moving the tab. **Group tabs by tag**, at the foot of the tag
-button's menu, turns the sections off for this browser.
+button's menu, turns the sections off for this browser. Every group starts on its own row; turning off
+the gear's **One tag group per row in the tab strip** lets the groups follow one another across the
+strip and wrap as they need, with the untagged sessions behind a thin divider, so a strip with many
+tags stays short.
 
 **Coming back after a dropped connection.** When the dashboard's link to the kernel
 drops and comes back (a laptop lid closed and opened, a network change, a phone that
--- a/tests/test_kernel_settings_sections.py
+++ b/tests/test_kernel_settings_sections.py
@@ -46,7 +46,7 @@ class SettingsSectionsTest(unittest.TestCase):
                     "id=rs-conserve", "id=rs-thinksum", "id=rs-fileedit"):
             self.assertTrue(h.index(">Sessions<") < h.index(rid) < h.index(">Chat<"), rid)
         # Chat: transcript prefs AND the comment defaults (comments are part of the chat)
-        for rid in ("id=rs-compact", "id=rs-branch", "id=rs-cmtmodel", "id=rs-cmtfast"):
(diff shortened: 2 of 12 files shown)


henrypinkard (approved):
Approved on head fd1b4ff5 on the maintainer's word (feature-review session).

PR #1169, opened 2026-09-09
Codex: a session going live sends the models frame, so an open dashboard learns the model list without a reload

ksarma:
**Symptom.** `GET /models` consults the Codex catalog only where this machine opted in: Codex as the default backend or the judge engine, or a live Codex session. A dashboard loaded before the first Codex session read `codex: {models: []}` (with #1167, the reason `no live Codex session; the list is read once one runs`) and was never told the list existed: the models frame that makes every picker re-read `/models` fires from the pick memory and the catalog fetch, not from the session that opens the consult. The chat's Codex model menu and the timeline lane's stayed empty until a reload; the same after a kernel restart that found every Codex session dead and a revive that made one live again.

**Cause.** `_create_codex_session_inner` and `_revive_session_inner`'s Codex arm return without `_models_changed()`. Those are the only kernel call sites of `CodexBackend.spawn` and `resume`.

**Fix.** Both doors call `_models_changed()` once the backend has landed the row: the spawn returned its sid; the resume answered True. Every landing sends the frame, not only the one that opens the consult: the flip is the exact event, but a door cannot see it without the backend counting closings at every live-to-dead write (two first creates can both land before either checks; a kill between a door's check and its landing hides a close-and-reopen). Sending on every landing never misses. The cost is one `GET /models` per open picker (chat, timeline, the feed's gear) per Codex create or successful revive, and the payload's `rev` already makes a repeated or late read harmless; the catalog is cached after its first non-empty answer. A spawn that returns a launch-error row (client unavailable) opens the consult too and sends the frame, so the re-read carries the client's reason in `error`. A revive of an already-live row also sends one (the backend's resume answers True for any known session). Nothing fires on a spawn that raised, a resume that answered False, or a revive the readiness check refused. `_models_changed`'s docstring and the `_models_rev` comment name the third cause. The calls are bare, like the two pick-memory callers: `_models_changed` cannot raise (the WS send path catches every exception and marks the client dead). A flip-only rule (the backend counting closings under its lock, the doors snapshotting before landing) is available if preferred; this PR sends one cheap re-read instead.

**Tests** (`tests/test_codex_models_frame.py`, new; the real doors over a fake backend and fake WS clients for the three apps; each send records whether the fake's live set was non-empty at that moment, and each frame assertion fails before the change because neither door reached `_models_changed`). The first Codex spawn sends one frame to chat, timeline and feed, each rev above the last, after the row is live. A second spawn beside a live one sends one: the rule is every landing. A revive of a dead Codex session sends one and focuses the asker's chat; a revive of a row already live sends one too; a resume the backend refuses sends none and files `reviveFailed` to the asker's chat and feed as before; a revive refused by the readiness check sends none. A spawn that raises sends none and the raise propagates. Two first creates racing, both landing before either continues, still reach every app. Removing the create door's call fails the two spawn tests and the race; moving it before the spawn fails the first on a frame sent while nothing was live; removing the revive arm's call fails both revive-landing tests, gating it on the row having been dead before the resume fails the already-live one, and dropping its `ok` guard fails the refused-resume test. The existing door tests (create ack, names, revive, command cache, discovery's source pin) run unchanged; their fakes need nothing new.

**Not in this PR.** The exact flip (a closings counter in the backend) if wanted. A frame when the last Codex session dies (a picker on a dead session cannot open). The consult opened by a settings change (default backend or judge engine set to Codex) with no live session. The chat picker's own row for an empty Codex list (an empty-state row that shows the route's reason) is a separate PR; the `/models` `error` field is #1167.

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -1567,8 +1567,8 @@ _learned_announced = set()   # ids already announced on stderr as outside the ca
 # file both parse it and store equal tuples, and holding _catalog_lock around file I/O would stall the
 # catalog refresh thread.
 _learned_reg_cache = {}   # str(path) -> ((mtime_ns, size, ino), liveModelId or "") — see _reported_model_ids
-# Bumps on every pick-memory change and every catalog growth; rides the models frame (_models_changed) AND
-# the /models payload, so a picker can drop a response older than one it has applied. Seeded from the clock
+# Bumps on every pick-memory change, catalog growth and Codex landing; rides the models frame (_models_changed)
+# AND the /models payload, so a picker can drop a response older than one it has applied. Seeded from the clock
 # rather than 0: the counter is per process, and a kernel restart must never hand a page that kept its
 # high-water mark a LOWER rev, or that page would ignore every re-read until the count caught up — a silent
 # stale list. Milliseconds leave room for one bump per ms across a restart.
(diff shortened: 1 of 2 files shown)


ksarma:
No findings to fold in. The PR merged on green before this review landed, so this comment is the record.

Reviewed inline: the two calls sit where the body says. The first follows the sid `_codex().spawn` returns in `_create_codex_session_inner`. The second sits inside the `if ok:` after `cx.resume` in the Codex arm of `_revive_session_inner`. Those are the only kernel call sites of the Codex spawn and resume. The frame goes through `_send_to_app` into `_client_send`, which catches every send exception and marks the client dead. A bare call therefore cannot turn a landed resume into a `reviveFailed`. A spawn that raises, a resume that answers False, and the readiness refusal reach neither call. Sending on every landing over-approximates the closed-to-open flip of the `/models` consult. The trigger is still a discrete event (a Codex row landing), not a timer or a per-build recomputation. Each consumer (chat picker, timeline lane, gear) applies a payload only when its rev is not below the one applied, and refills its list in place. Nothing that moves a card reads the frame. The race test holds both doors on a Barrier with bounded waits and no sleeps. The fake backend and fake clients are in-memory, so no test reaches a live port or starts an app server. The added lines carry only placeholder sids, the web/api names and `/TESTDIR`.

Test evidence:
- Merge base 1566f7ae with the new test file checked out: 5 of 8 fail (first spawn, second spawn, dead-row revive, live-row revive, two-creates race, each on no frame reaching any app); the 3 controls pass.
- Head 09ef6d3c: all 8 pass, also under `-n 4`; the existing door and models suites pass (248).
- CI at the merged head 1d84bb51, whose PR diff matches 09ef6d3c apart from hunk offsets: green on all nine checks.

Head note: the PR merged at 1d84bb51 (merge b0e5103a, 2026-09-09 12:30Z, by the repository owner). 1d84bb51 is 09ef6d3c plus a further merge of main; the PR's own diff is identical apart from hunk offsets, so this record covers the merged head.

Review run by the maintainer automated review session; verified against head 1d84bb51.

PR #1170, opened 2026-09-09
Courier link repair: the sender-board walk is built once per state of its inputs

henrypinkard:
Tier: fix

## What

The sender-board walk behind the courier's link repair (`_handoff_backref`) is built once per state of its inputs and served while they stand. Every call walked every discovered session's goal store with the writer's loader, and the courier asked it for each placed delegate whose link was missing, on every triage pass. The map now answers every message id from one walk over the read-only view, keyed the same way as the courier gate: the discover order and, per session, the store file's key with its journal's and archive's (the shared view's inputs; a journal replay can complete a handoff node), taken **before** the reads, so a store written during the walk moves the key the next call takes and a map read mid-write is served no further than that call. The first sender in discover order wins, as the walk answered; a completed handoff is no backref; a store that raises leaves nothing cached and the caller's own boundary logs it, as before. Counters ride `/perf` as `memos.backref` (`served`, `built`); a rebound state root clears the map. No card-move rule changes: the repair attaches exactly what the walk would have found.

## Review (by hand)

Key completeness: the walk read `nodes[*].handoff.msgId` and `nodeComplete` of every discovered store; both come from the store file with the journal replayed (the journal's resolve can set `nodeComplete`) and the archive is the view's third input, so the three keys plus the discover order cover every input; discover itself is cached by its own fingerprint. Faults: `_file_key` returns a fresh sentinel on a stat error, so the key never matches and the map rebuilds; a raising store propagates before the slot is written. Thread safety: one slot replaced whole; the only caller is the courier on the judge tier thread. Semantics: `setdefault` in discover order reproduces the walk's first hit; `mp.get(mid)` keeps the walk's equality on the raw msgId. Cost: one view read per discovered session per input state instead of one writer load per session per placed unlinked delegate per pass.

## Tests

`tests/test_backref_memo.py` (new): built once then served (one view read per sender); a sender store write re-derives even when the file clock does not move (same mtime, larger size); a journal row re-derives; a fleet change re-derives; a write landing during the build is seen next call (the interleaved-write pattern); the first sender in discover order wins; a completed handoff is no backref; a store that raises leaves nothing cached and a healed read builds; a rebound root forgets; the counters are a copy. `tests/test_perf_stats.py`: the memos key set and the counter's shape. `docs/reference.md`: the memos passage.

## Measurement for the judge thread (the dispatch's second half; nothing changed for it here)

A 60 s py-spy sample of the deployed kernel at 63 min uptime (1931 samples), the courier gate live (18585 scans skipped against 62 performed), 231 goal loads per triage pass, judge thread 45% of a core, planner workers 35%, process 57%.

**The judge tier thread is 18% of the process's samples; the planner worker pool is 37%.** Within the tier thread, `run_index` is 54%, the rewind reconcile 19%, propagate 14%, the courier 9%.

Top three romp lines by CPU, with what they read:

| where | line | share of its thread | reads, per session per pass | in the courier's skip key? |
|---|---|---|---|---|
| index tier | `tasks_for` (judge.py:2527) | 14% | `pcache/<sid>.json`, read and JSON-decoded every pass although its content is keyed on the transcript's fileset | no (the pcache file); its key input, the transcript, yes |
| triage tier | `load_goal_archive` (judge.py:4503) | 11% | `goals-archive/<sid>.json`, read and decoded per call (the live re-plan's recently-cleared context) | yes (`_archive_key`), but it is read fresh every call |
| index tier | `captioned_ids` + `_live_natoms` + `session_turn_captions` (judge.py:2563 to 2736) | 19% together | `captions/<sid>.jsonl`, the whole file JSON-decoded line by line, three times per session per pass | no |

Planner worker: `_seg_key` (judge.py:8585 to 8586, 12%), `_segment_id` (event_model.py:2434, 7%), `_placed_key` (judge.py:8640, 5%), `task_store_plan` (event_model.py:3028, 5%), `_mint_quote` (judge.py:7473, 5%). `_plan_session` calls `plan_units(session, store)` twice per session per pass and re-derives every session's segmentation and placements every pass; its inputs are the parse, the store with journal and archive, the episode log (all in the courier's skip key), plus Claude's task store for the session and the session's reg (not in it).

**Proposed next cuts, with the numbers above.** (a) The captions file parsed once per stat and the three readers derived from that one parse: about 19% of the tier thread. (b) `load_goal_archive` through a stat-keyed memo (or the shared-view idiom): 11% of the tier thread. (c) The planner's per-session change gate, the courier gate's construction with the task-store and reg keys added to the key, and `plan_units` computed once per `_plan_session`: the planner pool is 35% of a core and the courier gate's skip ratio (300 to 1) says most sessions would skip most passes. (c) is the largest; it touches the planner's mint path, so it wants the same red-first fixture discipline as the courier gate (three sessions, one moved, identical placements).

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -940,7 +940,9 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   stands (`served`, `derived`); `courierSkip` is the courier's change gate
   (`skipped`, `scanned`, `recorded`: a session whose parse, store, journal,
   archive and episode log have not moved since a scan that found nothing to
-  place is skipped whole). The compaction sweep after each judge pass evicts from `pass` and
+  place is skipped whole); `backref` is the sender-board walk behind the
+  courier's link repair, built once per state of the sender stores and served
+  while they stand (`served`, `built`). The compaction sweep after each judge pass evicts from `pass` and
   `shared` the entries of stores no session in the discover window owns, so
   both stay bounded by the live board; the gate memo is bounded by the session
   count.
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -165,6 +165,7 @@ def _rebind_state(path):
     _STORE_FAULTS.clear()   # unreadable-store episodes belong to the old root's files
     _CHAIN_MEMO.clear()     # the write-moment chain memo keys on paths under STATESDIR; a new root is a new world
     _COURIER_SEEN.clear()   # the courier gate keys on the old root's files
+    _BACKREF_MEMO["slot"] = None   # ...and so does the sender-board walk's map
     _episode_memo.clear()   # ...and so are the episode-log reads
     _head_memo.clear()      # transcript heads are immutable per path, but a rebind swaps the whole world of paths
     _namefp_memo.clear()    # names-entry content is memoized per SID against same-second mtimes — across a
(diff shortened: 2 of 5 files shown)


henrypinkard:

Live kernel on the shared box, `/perf` readings, no synthetic load; the "before" row is the reading taken on the previous kernel minutes before this change deployed, with the courier gate and the drift fix already live.

| code | uptime | goal loads / judge pass | judge thread CPU (% of a core) | planner pool | process CPU | backref memo served / built | courier gate skipped / scanned / recorded |
| before | 69 min | 231 | 45% | 35% | 56% | n/a (the walk ran per call) | 20091 / 93 / 87 |
| after | 6.7 min | 255 | 45% | 33% | 64% | 0 / 0 | 1684 / 56 / 54 |
| after | 20.3 min | 241 | 40% | 30% | 53% | 0 / 0 | 4739 / 104 / 97 |

**The walk no longer runs at all.** On the deployed kernel the memo was never asked: the courier gate skips the sessions whose placed delegates would have needed a link repair (they had nothing pending and no repair open), so `_handoff_backref` has no caller left in steady state. The memo's win therefore landed upstream of it, in the gate; the memo stands as the guard for the pass that does reach a repair (a new planner-first placement whose sender's tracking node appears), where one walk over the view now serves every message id instead of a writer load per discovered session per call. The judge thread's remaining share at this reading, 40% of a core, is the index tier's per-file reads and the planner's per-session work, the two rounds that follow this one.

ksarma:
Nothing found would have held the merge. The owner's auto-merge landed the PR at 12:08Z (merge 1566f7ae), before this review. One follow-up and two nits.

Reviewed inline: the memoized walk in `_handoff_backref` returns string pairs from a map that no caller mutates. So no shared-view object escapes, and nothing downstream writes into the view. `_attach_courier_link` still loads its own writer copy with `load_goals`. The walk raises no `FrozenStoreError` and leaves the shared cache's disable flag unset. The key covers every input the shared view reads: discover order, then each store file's stat with its journal's and archive's. It is taken before the reads, and an absent store keys as a stable `None`. The sole caller chain is `run_triage` to `run_courier` to `_courier_link_wanted` on the triage thread, so the one-slot-replaced-whole claim holds. A raising store leaves the slot untouched. The counters appear on `/perf` as `memos.backref`, and the docs passage matches. The privacy scan of the 236 added lines is clean. On the label: a new memo slot, key function and `/perf` counter looks like a feature under the brief's rule. The owner's role merges both tiers on green, so this is recorded only.

- `kernel/judge.py:14448`, follow-up: the walk is all-or-nothing. One sender store that raises on read (EACCES, EIO, a directory at the path) now fails the lookup for every message and every recipient. It files a link-attach pass-crash row per placed unlinked delegate per pass until the store reads again. The old early return answered for senders that sorted before the faulty store without reaching it. Fix: read each sender through `load_goals_shared_or_fault`, skip a faulted store, and withhold the slot for that call. The body says the boundary logs it as before; that understates the widened effect.
- `tests/test_backref_memo.py:104`, nit: the resolve row is keyed `id`, but the replay reads `node` (`judge.py:3711`). So the test pins the journal's stat, not the drop the docstring claims.
- `kernel/judge.py:14420, 14440` and `tests/test_backref_memo.py:9, 48, 109`, nit: eleven new occurrences in new code of the word the vocabulary rule forbids (identifiers and one docstring line). The commit message and docs hunk are clean; the PR body carries it once.

Tests: red-first verified at the merge base dc1b4554 with the PR's test files checked out. All 10 in `tests/test_backref_memo.py` fail with AttributeError on the absent memo, and `test_perf_stats.py` fails on the missing backref key. At head, `tests/test_backref_memo.py`, `test_perf_stats.py`, `test_courier_link.py` and `test_courier_skip.py` pass (80 passed). A mutation battery caught 8 of 9 single-seam breaks; the suite misses dropping the archive key from `_backref_key`. CI green at head.

Review run by the maintainer automated review session; verified against head c1cf74f2.

PR #1171, opened 2026-09-09
Judge readers: the captions file parsed once per file state, the goal archive loaded once for readers

henrypinkard:
Tier: fix

## What

Two per-file read memos on the judge side, keyed the same way (the file's stat, inode, mtime_ns and size, taken **before** the read; served only while the key stands):

1. **The captions file is parsed once per file state.** `captioned_ids`, `_live_natoms` and `session_turn_captions` each read and JSON-decoded the whole `captions/<sid>.jsonl` for every session on every index pass, 19% of the judge tier thread's samples. One parse (`_captions_rows`) now serves the three readers, each derived from the rows exactly as before. The file is append-only (`append_caption`), so a same-second append moves the size. An absent or unreadable file is empty and never cached; a malformed line is skipped as before, and a well-formed line that is not an object is skipped too (the old readers would have raised on it).

2. **The goal archive is loaded once per file state for readers.** `load_goal_archive` decoded `goals-archive/<sid>.json` per call, 11% of the tier thread; most calls were the re-plan's cleared context, the override replay's restore membership, two peer-node joins and the propagate pass's per-pass map, none of which write. Those take `load_goal_archive_shared`, the memo, whose object is shared and read-only by contract; every archiver (the compaction, the rewind sweep, the undo restore, the ledger merge in the kernel) keeps `load_goal_archive`, a fresh private object it mutates and saves under the archive lock.

Both memos are bounded (256 entries, oldest-inserted out) and cleared with the state root; counters ride `/perf` as `memos.captions` (`served`, `parsed`) and `memos.goalArchive` (`served`, `loaded`). No card-move rule changes: the readers' answers are the same functions of the same file content.

## Review (by hand)

Key rule: each key is the file's stat taken before the read, so a write during the read leaves the cache under a key the file no longer has and the next call re-derives (the interleaved-write tests pin it for both files); a stat error yields a fresh sentinel that never matches. Writers: every `save_goal_archive` site loads through `load_goal_archive` (the fresh loader), pinned by source; `load_goal_archive_shared` is only reached from readers, and the shared object is never handed to a saver. The captions memo's rows are shared across the three derivations, none of which mutate a row. Bounds: the eviction pops the oldest-inserted entry, so a box with more sessions than the cap re-parses the oldest rather than growing. Threads: the index tier and the planner workers both read captions; a race between two fills writes the same content under the same key, and a slot is replaced whole.

## Tests

`tests/test_file_read_memos.py` (new): parsed once and the three readers equal the direct three-pass derivations; an append re-derives even when the file clock does not move; a live row superseded by its final through `append_caption` is seen at once; a write landing during the read is served no further than that call; an absent file is empty and never cached; a malformed or non-object line is skipped; the memo is bounded; the archive is loaded once then served while every writer's load is fresh; an archive write re-derives; the interleaved write; an absent archive is the empty shape and never cached; the switched readers and the kept writers are pinned by source; a rebound root forgets both; the counters are copies. `tests/test_perf_stats.py`: the memos key set and the two shapes. `docs/reference.md`: the memos passage.

## After

Posted below after the deploy restart at matched uptime: judge tier thread, planner pool and process CPU shares, with the two memos' counters.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -942,7 +942,10 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   archive and episode log have not moved since a scan that found nothing to
   place is skipped whole); `backref` is the sender-board walk behind the
   courier's link repair, built once per state of the sender stores and served
-  while they stand (`served`, `built`). The compaction sweep after each judge pass evicts from `pass` and
+  while they stand (`served`, `built`); `captions` and `goalArchive` are the
+  per-file read memos behind the index tier's caption readers and the re-plan's
+  cleared context, each parsed once per file state (`served`, `parsed` or
+  `loaded`). The compaction sweep after each judge pass evicts from `pass` and
   `shared` the entries of stores no session in the discover window owns, so
   both stay bounded by the live board; the gate memo is bounded by the session
   count.
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -166,6 +166,7 @@ def _rebind_state(path):
     _CHAIN_MEMO.clear()     # the write-moment chain memo keys on paths under STATESDIR; a new root is a new world
     _COURIER_SEEN.clear()   # the courier gate keys on the old root's files
     _BACKREF_MEMO["slot"] = None   # ...and so does the sender-board walk's map
+    _CAPTIONS_MEMO.clear(); _GOALARCH_MEMO.clear()   # ...and the per-file read memos
     _episode_memo.clear()   # ...and so are the episode-log reads
     _head_memo.clear()      # transcript heads are immutable per path, but a rebind swaps the whole world of paths
     _namefp_memo.clear()    # names-entry content is memoized per SID against same-second mtimes — across a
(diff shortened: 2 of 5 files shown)


henrypinkard:

Live kernel on the shared box, `/perf` readings, no synthetic load. The "before" rows are readings on the kernel carrying every earlier round of the series (the courier gate, the drift fix, the backref memo) minutes before this change and the planner gate deployed together; the "after" rows carry both, so the judge tier thread's fall is this change's and the planner pool's is the planner gate's.

| code | uptime | judge pass wall | goal loads / judge pass | judge tier thread CPU (% of a core) | planner pool | process CPU | captions memo served / parsed | archive memo served / loaded |
|---|---|---|---|---|---|---|---|---|
| before | 6.7 min | 4.3 s | 255 | 45% | 33% | 64% | n/a | n/a |
| before | 20.3 min | 4.9 s | 241 | 40% | 30% | 53% | n/a | n/a |
| after | 6.7 min | 2.3 s | 219 | 23% | 15% | 41% | 6930 / 30 | 3071 / 112 |
| after | 20.8 min | 1.6 s | 206 | 23% | 16% | 38% | 24154 / 32 | 10609 / 310 |

**The captions file is parsed once per state and read 750 times per parse; the archive once per state and read 34 times per load.** The judge tier thread fell from 40% to 45% of a core to 23%, and a judge pass from 4.3 to 4.9 s of wall to 1.6 s at 20 min. The process's CPU fell from 53% to 64% of a core to 38% to 41%, the two rounds together.

ksarma:
Merged already; nothing found would have held it, so this is a post-merge record. The owner's auto-merge landed head a8c5c592 as merge 709b0480 at 2026-09-09 12:17Z.

Reviewed inline: the two per-file read memos and the two new /perf keys. Each memo is keyed on the file's (inode, mtime_ns, size), taken before the read, and served only while the key matches. An absent file is never cached. Entries cap at 256, oldest out, and a state-root rebind clears them. I traced every caller of `load_goal_archive_shared` by reading the code. `_replay_overrides` tests membership only; `_cleared_context` reads fields. `_deleg_report_lines`, `_deleg_frame` and the propagate pass copy the nodes map with `dict()` before updating it, and `_ref_goal` copies each node. None mutates the object. Every `save_goal_archive` site (judge.py 4685 and 4728, kernel.py 31114 and 31185) loads through the fresh `load_goal_archive`. The captions memo's three derivations each build a fresh container and write no row. The declared behaviour change is an improvement: a decodable non-object captions line is skipped where the old readers raised. That AttributeError escaped `captioned_ids` in `_run_index`'s serial loop, which has no per-session guard, and ended the index pass for every session. Tier note for the record: the added mechanisms (two memos, a public loader, /perf keys) are feature-shaped. The author is the owner, so the label changed no gate. Privacy and vocabulary scans of the 363 added lines are clean.

Follow-ups, none of which would have held the merge:

- `kernel/judge.py:4557`, follow-up. The memoized archive store is a plain dict of GuardedNode, and only PROTECTED keys are guarded. A reader's write would land silently and be served to every later reader until the file's stat key changes, with no error row. `load_goals_shared` freezes its shared view and files a frozen-store-write row on such a write. Suggested fix: freeze before `_memo_put`, with the archive's own off switch and message, since `_shared_poison` flips the live-store cache's. Replace the source-substring pin at `tests/test_file_read_memos.py:229` with a behavioural test that a write raises. The full suite passes with the memoized store frozen (10092 passed, 61 skipped).
- `kernel/judge.py:2586`, follow-up. `_memo_put` evicts with `memo.pop(next(iter(memo)))` under no lock. `load_goal_archive_shared` is reached from every thread that calls `load_goals`. At the 256 cap, two concurrent fills for new sids can raise KeyError or RuntimeError, and `_or_fault` (OSError only) lets them through. I reproduced it with a stress script at the default switch interval. The failure is narrow and self-healing. The chain memo does the same eviction under `_CHAIN_LOCK`. Suggested fix: one module lock around the served check, `_memo_put` and the counters.
- `kernel/judge.py:2565`, nit. The block comment says an unreadable file is never cached. For the archive memo, an existing but unreadable or corrupt file is cached as the empty shape, because `load_goal_archive` swallows the exception and the stat key is a tuple. That is the same answer the fresh loader gives, so the fix is comment only. The `load_goal_archive` docstring at 4538 also now disagrees with the new function below it about who mutates the archive.

Test evidence: red-first verified by my own run. At merge base 1566f7ae with the PR's test files checked out, 15 failed and 47 passed. The 15 were 14 in `tests/test_file_read_memos.py` on the missing `_CAPTIONS_MEMO` attribute and 1 in `tests/test_perf_stats.py` on the memos key set. At head a8c5c592, 62 passed, serially and under -n 4. CI was green at head.

Review run by the maintainer automated review session; verified against head a8c5c592.

PR #1172, opened 2026-09-09
Chat: a Codex model or effort menu with no list says why and asks for the list again instead of opening blank

ksarma:
A Codex session's model or effort menu opened blank when the kernel held no Codex model list and said nothing about why; it now shows the reason `GET /models` carries (`codex.error`, #1167) and asks for the list again.

## Tier

Tier: feature

Labelled feature because it adds a UI state with its own copy, a CSS class, a re-read policy and two placement rules. The behaviour it replaces, a menu that opens on nothing and says nothing, is a bug, so `fix` is defensible if you prefer that label.

## What

A Codex session's model or effort menu opened blank when the kernel's `codex.models` was empty, with the session's default badge showing above it and no word of why, and nothing re-read the list for that tab. `GET /models` now carries `codex.error` (#1167); this PR shows it. A Codex model or effort menu with no rows shows one row: a head (`No model list from Codex` / `No effort list from Codex`) and a sub-line with the kernel's sentence. The row is a statement, not a choice: no focus, no click, no hover wash. Every open on an empty list re-reads `/models` once: a held reason may be stale, and a dashboard loaded before the first Codex session holds none. While no reason is held the sub-line says `asking for the list now` with the loader dots; otherwise it shows the held reason. When the read lands the menu rebuilds with the list, or the sub-line shows the fresh reason in place.

## Design points

1. Where: only a Codex session's model or effort menu with zero rows; SDK menus, the mode menu and the timeline lane's picker are untouched. The effort list is a constant of four on the kernel, so its row is in practice the wait row (one condition, one string).

2. The rebuild re-resolves its badge instead of reusing the button captured at open. The statusline is rebuilt on every kernel push, and the spawn that makes the list readable also pushes, so the captured button is often detached by the time the list lands, and a menu positioned from a detached element's rect (all zeros) sits off-screen. `metaButton` stamps `data-sid` on the comment popover's badges (the chat's carry none) and `metaAnchor` finds the connected badge of the same kind and session; with none the menu stays closed and the next open shows the list. Re-resolving rather than rebuilding in place because the row builder closes over the menu element and its locals.

3. A chat menu whose tab was dismissed under it closes on the landing. `dismissSession` moves the active tab to the most recent survivor without `setActive`'s close, and the chat's badges carry no session, so a re-anchor could not tell the dismissed tab's badge from the survivor's. A popover's menu names its own session and is unchanged.

4. The hook belongs to the menu that set it and is cleared in `closeMetaMenu`, so a late response never reopens a menu the user closed, and a read that fails after the close is not recorded as if a menu were waiting.

5. A non-2xx `/models` answer is recorded as `HTTP <status>`: an empty or text body made `.json()` report a parse message. A failed re-read tells the waiting menu (`could not read the model list: ...`); with no menu waiting a failed read stays silent as today, so the `r.ok` check changes nothing user-visible for the loader's other reads (page load, the models frame).

6. No new font sizes or colours: the sub-line is `.meta-item-sub`, the dots are `.meta-dots`. The row wraps at 280px because a reason is a sentence and `.meta-item` is nowrap.

## Not in this PR

The timeline lane's Codex picker has no reason row; its common cause (no models frame when a Codex session goes live) is a separate kernel PR. The `/models` `error` field itself is #1167.

## Tests

`ui/webview/codex-meta-choices.test.ts`: the existing loop pin follows the new `rows` variable; one source-pin test over render.ts and styles.css (the row, the hook's set and clear, the dismissed-tab guard, the re-anchor, `data-sid`, the `r.ok` check, the CSS); the loader lifted and transpiled as models-rev.test.ts does: the error reaches the picker and the hook fires once per applied read while a stale-rev response fires nothing, a rejected re-read tells a waiting menu and records nothing with none waiting, a 403 with an empty body is recorded as its status and not as a parse message; the menu executed over a small DOM stand-in that records every rect read: the row shows the kernel's reason, held at open with no dots and one fetch all the same, or landing on the open menu in place (the same menu element, dots gone, no rebuild, no rect read), and a read with neither list nor reason leaves the fallback that names the kind; an SDK session's menu with no rows takes no reason row and issues no fetch; the landing rebuilds against the badge in the document now with no detached read and the row takes no focus and no listener; no badge leaves the menu closed and the next open shows the list without another fetch; a badge still in the document stays the anchor; the anchor resolves per session; a dismissed tab's menu closes on the landing with no reason row and no rect read; a popover's menu rebuilds by its own session whatever tab is active; a menu closed before the landing stays closed and a failure after the close is not recorded. The stand-in's statusline carries a mode, a model and an effort badge with their own rects, in the order the real one appends them, so a rebuild anchored on a badge of another kind is positioned from the wrong rect. `ui/webview/models-rev.test.ts`: the fetch stub answers a 200, since the lifted loader now checks `r.ok`. `docs/codex.md`: two sentences in the Auto paragraph, after the model-picker advice.

Each new assertion fails on the current sources: the source pins on missing text, the loader tests with a ReferenceError at construction, the menu tests on the missing row or the missing fetch (13 of the 16 tests red before, 16 green after; the SDK-session test is green before by design, pinning what the branch must not change). Removing the row branch, the backend gate, the reason written at open or on the landing, the re-anchor or its kind match, the dismissed-tab guard, the hook's clear in `closeMetaMenu`, the `r.ok` check, the `data-sid` stamp or the catch's waiting-menu guard, or moving the hook before the rev check, each turns an executed test red. Ran `npm run typecheck` and the full `npm test` (3661 pass, 0 fail).


henrypinkard (approved):
Approved on head c7c6f3f0 on the maintainer's word (feature-review session).

PR #1173, opened 2026-09-09
Planner: a session whose inputs have not moved since a pass that had nothing to do is skipped

henrypinkard:
Tier: fix

## What

The planner skips a session whose inputs have not moved since a pass that had nothing to do. `_plan_session` re-derived every session's segmentation, plan units and placement normalization on every planner pass; on the maintainer's box the planner worker pool was 37% of the process's samples (`_seg_key` 12%, `_segment_id` 7%, `_placed_key` 5%, `task_store_plan` 5%, `_mint_quote` 5%).

**The key** is every input the pass reads, taken **before** the store read (the chain-memo rule): the parse cache's fileset key bound to the session object the pass holds; the store file's key with its journal's and archive's; the episode log's key; the session's Claude task-store files, each name with its key (the declared-plan sync reads them); the reg file's key; and the transcript path. A parse the cache does not hold is never keyed, so never skipped; a task-store listing error is a fresh sentinel, so an unreadable store is never skipped over.

**The record rule.** A session is recorded only when the pass placed nothing, collected no unit and retired nothing, and the store's key after the pass equals the one before it: a heal, a mint (including the declared-plan mirror's), a retirement or a rollup change all write the store and move the key, so the next pass plans the session again whatever the table says. A moved session runs the whole pass exactly as before. The table is pruned to the pass's fleet and cleared with the state root; counters ride `/perf` as `memos.plannerSkip` (`skipped`, `planned`, `recorded`).

**No card-move rule is touched.** A skipped session has no new information by construction: every input the planner reads is in the key (parse, store, journal, archive, episode log, task store, reg, path), and a pass whose inputs stand would compute what the recorded pass computed, which was nothing.

## Review (by hand)

Inputs outside the key: `_hidden_from_feed` (session-flags) is evaluated by `run_plan` before the gate, so a muted session never reaches it; `now` enters `plan_units` only through the parse (states-aware, cached by fileset) and the settled gate, whose other input, the awaiting hold, reads the task store and the reg, both keyed; discover is the fleet. Threads: `_plan_session` runs on the planner pool, several sessions at once, each touching only its own table entry; the table is a dict with per-key writes, and a lost prune races only with a pop. Writers: the pass writes the store through the same `save_goals` as before; the record compares the store key, so any write un-records. The pass counts `pass_done` for a skipped session too: the pass over that fsid did complete, with nothing to do.

## Tests

`tests/test_planner_skip.py` (new; three sessions, one human prompt each, the planner's model calls deterministic stubs): unchanged inputs skip after a pass that had nothing to do and a skipped pass writes nothing; a moved store, a fresh parse (the new prompt placed), a task-store change alone (a file appearing, then a status moving) and a reg change alone each un-skip that session only; a write landing during the pass is seen next pass (the interleaved-write pattern); three sessions with one moved leave nodes, placements and status byte-identical to the ungated passes, run in two fresh worlds; a parse the cache does not hold is never skipped; a rebound root forgets; the counters are a copy. `tests/test_perf_stats.py`: the memos key set and the shape. `docs/reference.md`: the memos passage.

## After

Posted below after the deploy restart at matched uptime: judge tier thread, planner pool and process CPU shares, with the gate's counters.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -945,7 +945,10 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   while they stand (`served`, `built`); `captions` and `goalArchive` are the
   per-file read memos behind the index tier's caption readers and the re-plan's
   cleared context, each parsed once per file state (`served`, `parsed` or
-  `loaded`). The compaction sweep after each judge pass evicts from `pass` and
+  `loaded`); `plannerSkip` is the planner's change gate (`skipped`, `planned`,
+  `recorded`: a session whose parse, store, journal, archive, episode log,
+  task store and reg have not moved since a pass that had nothing to do is
+  not planned again). The compaction sweep after each judge pass evicts from `pass` and
   `shared` the entries of stores no session in the discover window owns, so
   both stay bounded by the live board; the gate memo is bounded by the session
   count.
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -165,6 +165,7 @@ def _rebind_state(path):
     _STORE_FAULTS.clear()   # unreadable-store episodes belong to the old root's files
     _CHAIN_MEMO.clear()     # the write-moment chain memo keys on paths under STATESDIR; a new root is a new world
     _COURIER_SEEN.clear()   # the courier gate keys on the old root's files
+    _PLANNER_SEEN.clear()   # ...and the planner gate
     _BACKREF_MEMO["slot"] = None   # ...and so does the sender-board walk's map
     _CAPTIONS_MEMO.clear(); _GOALARCH_MEMO.clear()   # ...and the per-file read memos
     _episode_memo.clear()   # ...and so are the episode-log reads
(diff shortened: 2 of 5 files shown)


henrypinkard:

Live kernel on the shared box, `/perf` readings, no synthetic load. The "before" rows are readings on the kernel carrying every earlier round (the courier gate, the drift fix, the backref memo) minutes before this change and the per-file read memos deployed together; the "after" rows carry both, so the planner pool's fall is this change's and the judge tier thread's is the read memos'.

| code | uptime | planner pool CPU (% of a core) | judge tier thread | process CPU | judge pass wall | planner gate skipped / planned / recorded |
|---|---|---|---|---|---|---|
| before | 6.7 min | 33% | 45% | 64% | 4.3 s | n/a |
| before | 20.3 min | 30% | 40% | 53% | 4.9 s | n/a |
| after | 6.7 min | 15% | 23% | 41% | 2.3 s | 2280 / 40 / 39 |
| after | 20.8 min | 16% | 23% | 38% | 1.6 s | 8014 / 48 / 45 |

**The planner plans one session in 170.** Over 278 passes the gate skipped 8014 sessions and planned 48, of which 45 recorded (nothing to do, store unchanged); the planner pool's CPU fell from 30% to 33% of a core to 15% to 16%. With the read memos beside it, the process's CPU is at 38% to 41% of a core against 53% to 64%, and the pusher thread at 11%. The one-hour reading follows as a line below.

ksarma:
Merged before this review landed (the owner's auto-merge at 12:26Z). Nothing found would have held it as shipped. Three gaps in the key need a follow-up PR, plus one wording fix.

Reviewed inline: I read `_plan_session` and every input it reads against the new key. The key holds the parse: the transcript, the anchor and the states file, so a recorded death's paired states row too. It also holds the store with its journal and archive, the episode log, the reg and the path. `_hidden_from_feed` runs before the gate. The nudge ledger, session flags and peer stores are reached only on a placement path, which un-records. The record rule holds: every heal, mint, retirement or rollup change writes the store and moves the after-key. The prune runs on the calling thread before the pool, so no prune is lost. An uncached parse and a task-store listing error are never recorded. Three inputs are outside the key: the task store, the captions file, and the settled gate's deadline expiry (details in the list below). The label is feature-shaped under the brief's rule (a new gate, counters, a /perf key). Under the owner's role that changed no gate.

Follow-ups:

- `kernel/judge.py:2222`: the task store is keyed under the romp sid, while `_sync_declared_plan` reads it under leafFsid. The two differ for every SDK session after a /clear. In live use, the backend's TaskCreate/TaskUpdate hook rewrites the reg (keyed) on the same tool call, so the card still moves on the next pass. The gate's stated contract does not hold there, and the test does not cover that shape. Key the task store under `session.get("leafFsid") or fsid`, the id line 6280 reads. Add a test whose SDK reg lastSid differs from the sid.
- `kernel/judge.py:2221`: the floor-title heal reads the captions file, which is not keyed. A coerce-floor title on an idle recorded session does not heal until an unrelated input moves. Add the captions file to the key.
- `kernel/judge.py:7247`: the settled gate's deadline expiry is a clock input no file key carries. The closer's ungated settle at the end of every pass covers it at the default. Either do not record a session while `_bg_unresolved(path)` is non-empty, or fold the expiry booleans into the key as `kernel.py`'s lift gate does. The body's sentence that the awaiting hold reads the task store is wrong: it reads the transcript's bg pairing, the reg, the death marker and the clock.
- `kernel/judge.py:2191` and 9475, the commit message and the body: the word the vocabulary rule bans appears in new prose. Rephrase the two comments.

Red-first by my own run: `tests/test_planner_skip.py` and `tests/test_perf_stats.py` at the merge base 709b0480 give 11 failed, 47 passed (10 AttributeError on `_PLANNER_SEEN`, 1 missing plannerSkip key); at c1e90a53, 58 passed. CI was green at c1e90a53. Privacy scan of the added lines is clean.

Review run by the maintainer automated review session; verified against head c1e90a53.

henrypinkard:
**The one-hour reading** (60.8 min uptime): 837 judge passes at 1.4 s of wall each, 202 goal loads per pass; the planner gate skipped 24220 sessions against 82 planned, 74 of them recorded; the judge tier thread 23% of a core, the planner pool 16%, the process 37%, the pusher thread 11%. The captions memo served 72867 reads against 39 parses, the archive memo 31852 against 869 loads, the courier gate 24201 skips against 72 scans.

PR #1174, opened 2026-09-09
Gear: the automatic-update help says the converge restarts at once or converges in place, not at the next quiet moment

ksarma:
## What

The gear's help line for *Install automatically* (`ui/webview/gear.js`, the Automatic updates row) said the mode converges "restarting at the next quiet moment". That has been false since T269 (#1121): every deploy restart is immediate, and `romp refresh --quiet` is the only door to the quiet window, as the reference's row for that command already says (`docs/reference.md`).

The automatic converge takes one of two routes, decided by what the pulled commits touch. `_main_drift_check` (`kernel/kernel.py`) calls `_run_main_update(kind, target=target)` with no `immediate` override, so the restart it asks the manager for is the `immediate=True` default. Inside `_run_main_update`, `if not _kernel_code_changed(_kernel_sha(), pulled) and _in_place_converge(pulled): return` picks the route: a change to code the running kernel executes restarts it at once, and the cut turns resume with their history at boot; anything else (the UI, the docs, the postal bus) converges in place with the kernel left up. The copy says "with the kernel left up" because `_in_place_converge` bounces the postal bus when its files changed, so a bus-only change does restart a process, just not the kernel. The `romp` command itself is deliberately not in the in-place list: `_restart_class` puts every `bin/` path in the kernel class (and `_ast_equal_blobs` cannot prove a bash script unchanged), so a change to `bin/romp` takes the restart route; only the `cli/` helper modules skip.

The line now reads:

> Install automatically converges by itself: a change to kernel code restarts it at once (turns in flight are cut and resume with their history); anything else (the UI, the docs, the postal bus) converges in place with the kernel left up;

Nothing else in the row changes: its label, its three options and the rest of the sentence stay as they were.

## Tests

`tests/test_kernel_update.py`, `Wiring.test_the_copy_says_an_automatic_update_restarts_at_once_or_converges_in_place`: pins the route line in `_run_main_update` that the copy describes, refuses "quiet moment" in `gear.js`, and requires the new clause. Red on main at the "quiet moment" check (1 failed, 58 passed); 59 passed with the change. `node esbuild.js` and `npm run typecheck` pass.

Tier: fix


ksarma:
Merged as shipped at the reviewed head. The owner landed it while this review ran, and no fixes were needed.

Reviewed inline: the auto caller in `_main_drift_check` (`kernel/kernel.py:7580`) passes no immediate override to `_run_main_update`. That function's default at :7598 is `immediate=True`, so the request at :7713 is `/restart-all` with no quiet parameter. `docs/reference.md:53` already names `romp refresh --quiet` as the only way into the quiet window. The pinned route line is at :7689 as written. `_restart_class` puts `kernel/` and `bin/` in the kernel class, `postal/` and `bin/romp-postal-service` in the bus class, and `.md` plus everything else (ui, docs, tests, cli) in skip. `_in_place_converge` restarts the bus when bus files changed, rebuilds dist, and returns with the kernel up. If the build fails it falls back to the restart. Cut turns resume with their history through the boot reconcile and BOOT_RESUME_NUDGE. The change is help copy inside the existing hover popover (`gear.css:57-58`). It touches no font size, menu token, or click handler, and the row keeps its label and three options. The `fix` label is right. The body and commit message match the diff and quote no one. The added lines carry no private strings, hostname, or home path.

- `ui/webview/gear.js:161`: the semicolon after the history parenthetical sits at the same level as the three option separators. A reader scanning on semicolons sees four segments. A comma or "and" there keeps the options distinct. Follow-up.
- `ui/webview/gear.js:161`: "at once" is right about the mode. But `_main_drift_check` holds an automatic converge until 25 minutes after the last deploy restart (`kernel.py:7576-7578`, :7024), so on a busy main the restart can come later. The old copy omitted this too. Follow-up.

The test is red before the fix and green after, in my own run. At base b0e5103a with the PR's test file checked out, `tests/test_kernel_update.py` gave 1 failed, 58 passed, failing at line 917 on the quiet-moment check. At head 886979b6, 59 passed. CI at 886979b6: all eleven checks green, with the Tier policy check applying the fix gate and requiring no approval.

Review run by the maintainer automated review session; verified against head 886979b6.

PR #1175, opened 2026-09-09
Timeline: the dead-lane memo's key carries each file's ctime and the host's recorded suspensions

ksarma:
## Symptom
A dead lane whose transcript cannot be read stays empty until a restart or an eviction, even after a chmod or chown repairs the read. An unreadable transcript parses as an empty session with no complaint (the read layer returns no records on an OSError); a parser fault draws the same empty lane and says so once on stderr; either way the empty lane is cached and served. Separately, a host suspension recorded after a dead lane was cached is never cut out of that lane's bars.

## Cause
The dead-lane memo (#1131) serves a lane while every file its derivation reads stands, keyed on each file's (mtime_ns, size, inode) plus the branch clip. Two inputs the derivation reads are not in that key.

- **The transcript's ctime.** An empty lane is cached like any other (the populate runs; a parser fault's complaint dedups). That is right for a dead transcript with no writer, but the lane is derived again only when a keyed file moves, and a chmod or chown that repairs the read moves neither mtime, size nor inode.
- **The host's recorded suspensions.** `_awake_spans` excises every interval in `_downtime` from each segment's span. It reads the in-memory list (`_record_suspend` appends the jsonl mirror best-effort, so the list, not the file, is the input); the list grows when the producer's tick detects a sleep, and the timeline builds on the pusher thread. A lane derived after a resume but before that tick (a session that died across the sleep) draws one bar across the sleep and caches it; the interval recorded seconds later moves no file.

## Fix
`_stat_key` now carries `st_ctime_ns` as its fourth member: the repair moves the key and the parse is attempted once more. `_stat_key` is shared: it also keys the cleared-set memo and the nudge-gate memo's episodes file. The cleared-set memo re-derives once on a chmod of cleared.jsonl, one extra read (a new test pins it, and the docstring names the fourth member); the nudge-gate memo compares the episodes file's key by equality, so the same follows there. If a dead-lane-only stat is preferred: a second stat helper carrying ctime, called by `_dead_lane_key` alone, with `_stat_key`, the other two memos and their tests unchanged.

`_dead_lane_key` now ends in `tuple(_downtime)`, so the next build re-cuts the bars. The cost: a recorded suspension re-derives every cached dead lane once on the next build (a parse per lane, from the assembly cache while it holds the leaf and from disk once it does not, then re-segmentation and the judging marks), about 18 times a night under macOS dark-wake by the figure in `_awake_spans`. Clearing the memo from `_record_suspend` would cost the same; the key form keeps every input in one place and is what the tests pin. A later change could store each lane's cached span and re-derive only the lanes a new nap overlaps. The tuple is a copy of the append-only list, built per dead lane per build, and each cached lane holds one; a hoist per build would share one object if that matters.

No change to what is cached or when; a served lane stays byte-identical to a rebuilt one; the memo's header comment names both inputs.

## Tests
`tests/test_timeline_lane_memo.py`, three cases in `DeadLaneMemo`, each red on main:
- a chmod alone moves the transcript's stat key while mtime, size and inode stand (red: the two keys are equal);
- a parse that raises is cached as the empty lane and served once, with one stderr line across the two builds, then attempted again after the transcript's ctime moves and draws its bar (red: the parse count stays at 1 and the lane stays empty);
- a suspension set inside a cached segment splits the bar on the very next build and moves the key, a different nap of the same count re-derives again, one outside every segment re-derives to equal bars, and the same list rebound is served (red: the single bar is served).

The ctime is moved by chmod in a bounded loop, checked by stat after each attempt, that fails loudly if the filesystem never ticks it. The vehicle for the empty lane is a raising stub, the path with an observable complaint; an unreadable file parses as empty silently and is cached the same way. `tests/test_cleared_set_memo.py`: a new case pins that a chmod of cleared.jsonl re-derives the set once (red on main: the set is served), and the key-length pin reads 5. Module counts: 10 to 13 and 10 to 11.

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -30701,12 +30701,12 @@ def _cleared_ids():
     Parsed once per file state and served while the file stands (2026-09-09): the nudge walk read it for
     every session on every pusher cycle, and once the placement gate was memoized this replay of the whole
     log (two thousand rows on the maintainer's box) was 60% of the nudge tick. The key is the file's path
-    and stat (mtime_ns, size, inode), taken BEFORE the read (the chain-memo rule): a row appended during the
-    read moves the stat the next call takes, so a set parsed mid-write is served no further than that call;
-    every writer appends, so a same-second append moves the size (the kernel's file clock is coarse, so
-    mtime alone would not see it); a rebound state root is a different path. An absent or unreadable file
-    is the empty set, never cached. One slot, replaced whole, so two threads deriving at once can never pair
-    one's key with the other's set. Callers read the returned dict and never mutate it."""
+    and stat (mtime_ns, size, inode, ctime_ns), taken BEFORE the read (the chain-memo rule): a row appended
+    during the read moves the stat the next call takes, so a set parsed mid-write is served no further than
+    that call; every writer appends, so a same-second append moves the size (the kernel's file clock is
+    coarse, so mtime alone would not see it); a rebound state root is a different path. An absent or
+    unreadable file is the empty set, never cached. One slot, replaced whole, so two threads deriving at
+    once can never pair one's key with the other's set. Callers read the returned dict and never mutate it."""
     path = jd.STATE / "cleared.jsonl"
     st = _stat_key(path)
     key = (str(path),) + st if st is not None else None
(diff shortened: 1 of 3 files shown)


ksarma:

Reviewed inline: `_stat_key` has three consumers at this head (the nudge-gate episodes key, the cleared-set memo and `_dead_lane_key`). All compare by equality and none unpacks or indexes, so the fourth member changes no shape they read. Since ctime moves with every content write, the member can only add re-derivations, one per chmod or chown of a keyed file. The `tuple(_downtime)` term is taken before `_suspended_after` and `_awake_spans` read the list. A mid-build append from the producer thread therefore caches fresher bars under the older key. That costs one extra derivation next build, never a stale serve. The copy is one C call under the GIL. The 512 bound and oldest-inserted eviction are untouched. The failed-parse path still stores the empty lane under its key, and the ctime lets a repair through. An unreadable transcript still parses silently as an empty lane. That is the read layer's existing contract in a file this PR does not touch. The PR turns a stuck lane into one that recovers, so it is recorded, not held. No card mechanism is involved, and body and commit match the diff. The added lines are clean on the privacy and vocabulary scans.

Follow-ups, neither blocking:
- `kernel/kernel.py:35166`: each cached dead lane holds its own tuple copy of `_downtime`. One copy per build, shared by every lane's key, would do the same work once, as the body already suggests.
- `tests/test_timeline_lane_memo.py:204`: the unreadable-transcript case is pinned through a raising stub. A chmod(0) variant, skipped under root, would pin the silent path directly.

Red-first at the base ab144390 with the PR's two test files checked out: 5 failed, 19 passed. The five failures are the three new DeadLaneMemo cases, test_a_chmod_of_the_log_re_derives_the_set_once and the key-length pin, each for the reason the body gives. At the head the same files pass 24/24, the related memo suites 171/171 and tests/test_kernel.py 485/485. CI at e1ed5a7f: every required check is green (Python 3.10 to 3.13, bats, vscode-extension, Exactly one tier label, Tier policy).

Review run by the maintainer automated review session; verified against head e1ed5a7f.

ksarma:
Head note: the branch gained one empty commit (5d7b1324, no file changes) so GitHub re-evaluates the merge; the PR's diff against main is byte-identical to the reviewed e1ed5a7f, so the record above covers this head. Review run by the maintainer automated review session; verified against head 5d7b1324.

PR #1176, opened 2026-09-09
Tests: restore the shared judge's roots and the sessions-file seam after every test, and fail the test that does not

ksarma:
## What

**Symptom.** Under xdist with `--dist loadfile` or `--dist loadscope`, unrelated test modules went red together on one worker in about half the runs, and every failing test passed when its module ran alone. Three default-scheduler runs (`-n 8`) were clean; the default scheduler rarely puts the cause and the victim on the same worker. CI's serial run passes only because a test that loads a kernel between them (`tests/test_kernel_redistill.py`) re-executes `judge.py` and rebinds the roots from the environment. The victims:

- `tests/test_restart_cuts.py::CutRow` (3 tests): FileNotFoundError writing `restart-audit.jsonl`
- `tests/test_tag_edit_ack.py::LegacyStoreStampedOnce` (2): FileNotFoundError on `timeline-views.json`
- `tests/test_kernel_tunnel_truth.py::ExpectedRestart` (1): an empty restart reason, because `_audit_restart_request` swallows the write error
- `tests/test_dead_wait_block.py::DeadWaitBlock` (4): `_dead_wait_sweep` globs `STATE / "goals"` under `except OSError` and finds nothing
- `tests/test_postal_read_receipts.py::RecipientQueuesReceipt` (5): relays refused with no live session named `web`
- `tests/test_judge_auth_billing.py::KeylessKeyBilledCalls` (1): a stderr line the test did not expect

**Mechanism.** Two process-wide objects carry state from one test module to the next: the judge module and the process environment.

The judge module is one object per process. `kernel.py` loads it as `SourceFileLoader("romp_judge", ...).load_module()`, and `load_module` re-executes into the module already in `sys.modules` under that name, so every kernel-loading test module's `km.jd` is the same object. A fixture that rebinds `jd.STATE` to a per-test temp dir and removes the dir in tearDown without putting the prior root back leaves every later STATE reader on that worker pointing at a removed directory. Four fixtures did that (`test_episode_boundary`, `test_kernel_goal_compaction`, `test_session_retry_suppress::SessionRetrySuppress`, `test_subagent_transcripts`, the last also leaving `jd.PROJECTS` in the removed dir); `test_session_retry_suppress` rebinds STATE alone, so `GOALDIR` stayed valid while `_dead_wait_sweep` read `STATE / "goals"`, which is why it alone breaks `test_dead_wait_block`. Two more left a root changed without breaking anything yet: `test_chat_delta_resync` rebinds STATE to `tmp_path` in three tests and leaves it there; `test_clear_chat_view` restores STATE but leaves PROJECTS in its removed dir.

The sessions-file seam is read per call: the postal service reads `ROMP_SESSIONS_FILE` from `os.environ` each time it lists the live sessions. Three modules popped it in tearDown or a cleanup instead of restoring what they found (`test_postal_undelivered`, `test_postal_live_only`, `test_postal_relay_honesty`), so `test_postal_read_receipts`, which set the seam once at import, resolved no local sessions after any of them. Two modules pinned their own file in setUp without restoring it (`test_postal_quarantine`, four classes, and `test_postal_self_identity`); in a full run the value they set differs from the one the last-collected module left, so the guard names their first test. The new pin in `test_postal_read_receipts` restores for the same reason.

The billing case is a class-order dependency inside one module: the judge says once per process that a bare-alias envelope carries no `modelUsage` map, the test's fake envelope had none, and the test passed only after a sibling class had spent the notice.

**Fixes (tests only).**

- `tests/test_episode_boundary.py`: setUp saves `jd.STATE`; tearDown rebinds it before the rmtree.
- `tests/test_kernel_goal_compaction.py`: the same.
- `tests/test_session_retry_suppress.py` (`SessionRetrySuppress`): saves `km.jd.STATE` and restores it before `td.cleanup()`, as the module's second class already did.
- `tests/test_subagent_transcripts.py`: saves and restores `jd.STATE` and `jd.PROJECTS`, and clears `jd._discover_cache` so no discover result built under its roots is served afterwards.
- `tests/test_chat_delta_resync.py`: the three rebinding tests use a `state_root` fixture that rebinds and rebinds back.
- `tests/test_clear_chat_view.py`: both classes restore `jd.PROJECTS` alongside STATE.
- `tests/test_postal_undelivered.py`: the three tearDown pops restore the prior value.
- `tests/test_postal_live_only.py`: the four tearDown pops restore the prior value; the setUp pop in `KernelSilenceIsNotDeadness`, which tests the no-seam path, stays and now records what it popped.
- `tests/test_postal_relay_honesty.py`: `_RelayBase`'s cleanup restores the prior value; its setUp pop and the three in-test pops that exercise the no-seam fallback are unchanged.
- `tests/test_postal_read_receipts.py`: `_Base.setUp` pins the module's own file per test and tearDown restores the prior value, so the module no longer depends on what the last-collected module left in the environment.
- `tests/test_postal_quarantine.py`, `tests/test_postal_self_identity.py`: the existing setUp pins gain a tearDown restore (`TokenProvenDialerGate.tearDown` now calls `super().tearDown()`).
- `tests/test_judge_rate_limit_gate.py`: the seg-key test restores `ROMP_SERVE_TOKEN` and `ROMP_KERNEL_NO_OPEN` after the kernel import that needed them set.
- `tests/test_judge_auth_billing.py` (`KeylessKeyBilledCalls`): the fake envelope carries the `modelUsage` map a current CLI reports, naming the version `_ALIAS_HEAD` assumes for `sonnet`, so the test's own fixture satisfies the premise it asserts (nothing missing, nothing announced); the served record is put back to what it was afterwards.

The STATE restores follow `tests/test_judge.py::ClearedSeal` and `tests/test_kernel_resolve_node.py`. The environment restores call one helper, `restore_env(name, prior)` in `tests/conftest.py`, imported the way `thread_census` is; `prior` is the `os.environ.get` taken before the change, `None` meaning unset.

**Guard.** `tests/conftest.py` gains an autouse, function-scoped fixture, `_shared_state_restored`. Before each test it snapshots the shared judge's `STATE` and `PROJECTS` (when `romp_judge` is loaded) and the environment names `ROMP_SESSIONS_FILE` and `ROMP_SERVE_TOKEN`; after the test it fails the test that changed one of them and did not restore it, or left a path that was a directory pointing at nothing. The failure names the test and what it left behind (for the environment names only the kind of change, never the value; one of them is a credential). It reports as an ERROR at teardown, which is still red.

It is transition-based: a module-level preamble runs at collection, before any snapshot, and is not seen; a test that changes and restores is quiet; a test that runs under another test's leftover is not blamed for it. A test that loads a kernel re-executes `judge.py` into the shared module, which rebinds every root from the environment as it stands at that moment (other modules' import-time `XDG_STATE_HOME` and `CLAUDE_CONFIG_DIR` writes decide those values, and the per-test `CLAUDE_CONFIG_DIR` floor differs from the collection-time one). That reset is the loader's, so for such a test (told apart by the function objects `judge.py` redefines) the guard compares no values. Two things are still checked for it: `judge.py` creates `STATE` at import, so a `STATE` that is not a directory afterwards is the test's own doing and is named; and the environment names.

With the guard in place, a green run no longer depends on the kernel-loading test running between a cause and its victim.

**Not here.** `tests/test_kernel_webpush.py::LandingRevealOffers::test_no_offer_for_the_session_already_in_front` asserts a clock-derived `ageS` of 5 and saw 6 twice under load across many runs; it passes alone (9/9, module 86/86) and is timing, not shared state. Goal-store tests that mint goals under the shared placeholder sid can be re-flagged by another module's journaled override on the same worker; that is a separate class with a separate fix (a private sid per module), and nothing here touches it.

## Tests

Commands run from the repo root as `pytest -q <targets>`. Red counts were taken with the guard in place and the named module at its previous version; green counts at this commit.

STATE and PROJECTS polluters, alone under the guard (the guard errors at every test of the polluting class; nothing else in the process):

- `tests/test_episode_boundary.py`: 17 errors before, 17 passed after
- `tests/test_kernel_goal_compaction.py`: 8 errors before, 8 passed after
- `tests/test_session_retry_suppress.py`: 6 errors before (the second class was quiet), 24 passed after
- `tests/test_subagent_transcripts.py`: 27 errors before (STATE and PROJECTS both named), 27 passed after
- `tests/test_chat_delta_resync.py`: 3 errors before, 12 passed after
- `tests/test_clear_chat_view.py`: 14 errors before (PROJECTS), 14 passed after
- `tests/test_judge_rate_limit_gate.py`: 1 error before (ROMP_SERVE_TOKEN was unset and is now set), 20 passed after

The victims paired with their polluter, whole modules in this order (before: the base tree; after: this commit):

- `tests/test_episode_boundary.py tests/test_restart_cuts.py`: 3 CutRow failures before, 25 passed after
- `tests/test_kernel_goal_compaction.py tests/test_restart_cuts.py`: 3 before, 16 passed after
- `tests/test_session_retry_suppress.py tests/test_dead_wait_block.py`: 4 DeadWaitBlock failures before, 47 passed after
- `tests/test_subagent_transcripts.py tests/test_restart_cuts.py`: 3 before, 35 passed after
- `tests/test_episode_boundary.py tests/test_kernel_tunnel_truth.py`: 1 before, 39 passed after

Seam poppers with the victim, whole modules. Before the fix, popper first: 5 RecipientQueuesReceipt failures plus 1 guard error on the popper's first test, since the victim's import had set the seam. After: green in both orders.

- `tests/test_postal_undelivered.py tests/test_postal_read_receipts.py`: 30 passed, and 30 reversed
- `tests/test_postal_live_only.py tests/test_postal_read_receipts.py`: 48 and 48
- `tests/test_postal_relay_honesty.py tests/test_postal_read_receipts.py`: 59 and 59
- `tests/test_postal_live_only.py tests/test_postal_self_identity.py`: 2 guard errors before (the pop, then the pin), 30 passed after
- `tests/test_postal_live_only.py tests/test_postal_quarantine.py`: 58 passed. The quarantine pins have no pair line: before their restore, a full run's guard named the pins' first tests, since the file they set differs from the one the last-collected module left.

Billing: `tests/test_judge_auth_billing.py::KeylessKeyBilledCalls` alone: 1 failed before, 1 passed after; the module: 40 passed.

Reload exemption: `tests/test_session_auth.py tests/test_kernel_redistill.py`: 78 passed (without the exemption the guard named the redistill test for a change the loader made). A scratch test that reloaded the judge and then rebound `STATE` to a temp dir it removed errors under the guard; a reload that leaves `STATE` standing is quiet.

The pairs are not a subprocess meta-test: the guard already fails a popper's first test in any run that collected a seam-setting module in the same process, which every full run does (several postal modules set the seam at import), so CI catches the class in-process without a nested pytest run. The pair commands are for a reader who wants to see the victim fail without the guard.

Full suite, `-n 8`, on this commit:

- default scheduling: 10104 passed, 61 skipped, 1671 subtests passed in 78 s
- `--dist loadfile`: 10104 passed, 61 skipped, 1671 subtests passed in 97 s
- `--dist loadscope`: 10104 passed, 61 skipped, 1671 subtests passed in 66 s

Tier: docs

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -318,6 +318,95 @@ def _stub_place_llm(monkeypatch):
     yield
 
 
+# No test may leave the shared judge or a call-time environment seam changed (2026-09-09). kernel.py
+# loads the judge as SourceFileLoader("romp_judge", ...).load_module(), and load_module re-executes
+# into the module object already in sys.modules under that name, so every kernel-loading test
+# module's km.jd is ONE process-wide object. A test that rebinds jd.STATE to a temp dir and removes
+# that dir in tearDown without restoring the prior value leaves every later STATE reader in the
+# process pointing at a removed directory: a FileNotFoundError on restart-audit.jsonl or
+# timeline-views.json, or a silent empty read where the writer swallows OSError. The postal
+# sessions-file seam has the same shape: postal_service reads ROMP_SESSIONS_FILE from os.environ at
+# call time, so a tearDown that pops it instead of restoring the prior value leaves a later module,
+# which set the seam once at import, resolving no local sessions. Neither shows when the victim runs
+# alone, and the serial order of the whole suite passes only because a test that loads a kernel
+# between the cause and the victim re-executes judge.py and rebinds the roots; any other order (a
+# subset, another scheduler) fails a module that did nothing wrong. This fixture names the cause
+# instead: it snapshots the shared judge's STATE and PROJECTS and the watched environment names
+# before each test and fails the test that changed one and did not restore it, or left a path that
+# was a directory pointing at nothing. Transition-based on purpose: a module-level preamble runs at
+# collection, before any snapshot, and is not seen; a test that changes and restores is quiet; and a
+# test that merely runs under another test's leftover is not blamed for it. A test that loads a
+# kernel (or the judge itself) re-executes judge.py into the shared module, which rebinds every root
+# from the environment as it stands at that moment: that is the loader's reset, made from values
+# other modules' import-time writes decide, not a directory the test made and removed, so the path
+# check compares no values for that test (the re-execution recreates every function object, which is
+# how it is told apart from an assignment). Only the values: judge.py creates STATE at import, so a
+# STATE that is not a directory after a reload is the test's own doing and is still named, and the
+# environment names are still checked. Values of the environment names are never printed (one of
+# them is a credential), only the kind of change.
+_SHARED_JUDGE_PATHS = ("STATE", "PROJECTS")
+_SEAM_ENV_NAMES = ("ROMP_SESSIONS_FILE", "ROMP_SERVE_TOKEN")
+
+
+def _shared_judge_paths():
+    """({name: (path text or None, is a directory)}, marker) for the shared judge's watched globals,
+    the marker being a function object judge.py defines (a re-execution replaces it); ({}, None) when
+    no module has loaded the judge under its shared name yet."""
+    jd = sys.modules.get("romp_judge")
+    if jd is None:
+        return {}, None
+    out = {}
+    for name in _SHARED_JUDGE_PATHS:
+        p = getattr(jd, name, None)
+        text = None if p is None else str(p)
+        out[name] = (text, text is not None and os.path.isdir(text))
+    return out, vars(jd).get("_rebind_state")
+
+
+@pytest.fixture(autouse=True)
+def _shared_state_restored(request):
+    paths_before, marker_before = _shared_judge_paths()
+    env_before = {name: os.environ.get(name) for name in _SEAM_ENV_NAMES}
+    yield
+    paths_after, marker_after = _shared_judge_paths()
+    left = []
+    if marker_after is marker_before:      # not re-executed: whatever differs, this test assigned
+        for name, (text0, isdir0) in paths_before.items():
+            text1, isdir1 = paths_after.get(name, (None, False))
+            if text1 != text0:
+                left.append("romp_judge.%s changed from %s to %s" % (name, text0, text1))
+            elif isdir0 and not isdir1:
+                left.append("romp_judge.%s %s was a directory and is gone" % (name, text0))
+    else:                                  # re-executed: the loader bound the roots, and created STATE
+        text1, isdir1 = paths_after.get("STATE", (None, False))
+        if text1 is not None and not isdir1:
+            left.append("romp_judge.STATE %s is not a directory after the test reloaded the judge" % text1)
+    for name in _SEAM_ENV_NAMES:
+        v0, v1 = env_before[name], os.environ.get(name)
+        if v0 == v1:
+            continue
+        if v1 is None:
+            left.append("%s was set and is now unset" % name)
+        elif v0 is None:
+            left.append("%s was unset and is now set" % name)
+        else:
+            left.append("%s was changed" % name)
+    if left:
+        pytest.fail("%s left shared state changed after its teardown: %s. Save the prior value before "
+                    "changing it and put it back at the end of the test; for a path, before removing the "
+                    "directory it named." % (request.node.nodeid, "; ".join(left)), pytrace=False)
+
+
+def restore_env(name, prior):
+    """Put the environment name back the way a test found it: `prior` is the os.environ.get(name) taken before
+    the test changed it, None meaning unset. A tearDown that pops the name instead leaves a later module in the
+    process without the value its own import set; this is the restore the fixture above expects."""
+    if prior is None:
+        os.environ.pop(name, None)
+    else:
+        os.environ[name] = prior
+
+
 # No test report may carry a process-environment VALUE, or a credential-shaped token (2026-09-05). A
 # test that renders an env mapping in an assertion (assertNotIn on os.environ, on a _judge_env() copy
 # of it, on a launch env) prints the whole mapping when it fails, and on a developer's box that
--- a/tests/test_chat_delta_resync.py
+++ b/tests/test_chat_delta_resync.py
@@ -22,6 +22,8 @@ import re
 from importlib.machinery import SourceFileLoader
 import tempfile
 
+import pytest
+
 BIN = os.path.join(os.path.dirname(os.path.dirname(os.path.realpath(__file__))), "bin")
 # Hermetic state BEFORE the loads — they resolve their state root at import time, and only
 # pytest runs conftest's floor (a bare unittest or script run otherwise writes REAL state).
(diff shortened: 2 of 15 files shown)


ksarma:
Merging as shipped. Two follow-ups and a nit below; nothing to fold into this PR.

Reviewed inline: All 15 files are under `tests/`, so nothing shipped changes, and the docs tier (tests-only changes) fits. The guard in `tests/conftest.py` snapshots `romp_judge.STATE`, `PROJECTS` and the two environment names before each test and compares them afterwards. A test running under another test's leftover is therefore not blamed. A first load of the judge compares nothing. The replaced `_rebind_state` function object tells a judge re-execution apart from an assignment; STATE is still checked for existence on that path. The environment messages carry only the kind of change, never a value (confirmed in the actual error text). Every per-module save-restore matches the body. The deliberate pops in `test_postal_live_only` and `test_postal_relay_honesty` stay. `restore_env` gets a prior taken before the change at each call site. The dropped `km.jd = jd` lines in `test_chat_delta_resync` were no-ops: the kernel loads the judge into the same module object. The billing fixture's served id parses to the version `_ALIAS_HEAD` assumes for sonnet, so class order no longer decides the test. The judge's notice for a CLI lacking the field keeps its own pin in `tests/test_judge.py`. The body and commit message match the diff and quote no one. The added lines carry no private strings, hostname or home path.

- `tests/test_sdk_rate_limit_usage.py:135` and `tests/test_postal_bus_lifetime.py:94` (KernelUp): follow-up. Two untouched modules leak a watched name when run alone. One does a setdefault of `ROMP_SERVE_TOKEN` with no restore. The other has a tearDown that puts `ROMP_SESSIONS_FILE` back only when the prior was set. Each is 1 error alone at this head and clean at main. Full runs show neither error because module-level writers set the names first. The shape this PR applied at `tests/test_judge_rate_limit_gate.py:199-206` closes both.
- `tests/README.md:101`: follow-up. The suite conventions do not yet name the guard, the four watched names, or `restore_env`.
- `tests/conftest.py:367`: nit. The guard compares after monkeypatch's undo only because its name sorts before `_stub_place_llm`. A comment naming that, or a teardown hookwrapper, removes the dependence.

At the base with only the PR's conftest, the seven polluter modules error with the body's counts (17/8/6/27/3/14/1), and the seam pair with 2. At the head, the touched modules, their victims and the reload pair pass under `-n 4 --dist loadfile` and `loadscope` (516 each). Every touched module passes alone. The full suite under loadfile is 10132 passed, 0 failed. CI at ba878a98: every required check green.

Review run by the maintainer automated review session; verified against head ba878a98.

ksarma:
Head note: the branch gained one empty commit (327b5ee4, no file changes) so GitHub re-evaluates the merge; the PR's diff against main is byte-identical to the reviewed ba878a98, so the record above covers this head. Review run by the maintainer automated review session; verified against head 327b5ee4.

PR #1177, opened 2026-09-09
Courier link repair: an unreadable sender store is skipped for that sender, not for every recipient

ksarma:
## What

#1170 built the sender-board walk behind the courier's link repair (`_handoff_backref`) once per state of its inputs, reading every discovered store through the read-only view. That loader raises for a store that exists and cannot be read (EACCES, EIO, a directory at the path), and the walk had no per-sender boundary and wrote its memo slot only after the loop, so one such store failed the lookup for every message id and every recipient: a link-attach pass-crash row per placed unlinked delegate per pass for as long as the fault lasted, and no repair landed anywhere. The walk it replaced returned at the first hit in discover order, so a fault reached only the lookups past it. The review record on #1170 names this as the follow-up.

Each sender is now read through the per-session boundary (`load_goals_shared_or_fault`): a faulted store is skipped for this call and files its one store-unreadable row per fault episode, as every other reader of a goal store files it; the other senders answer, and the first in discover order still wins. The map is not published when a sender was skipped, because a permission fix moves no file key (neither inode, mtime nor size) and a cached map missing a sender would be served after the store reads again. While the fault lasts, a call whose key no longer matches the published map walks the shared views again (a map published before the fault keeps serving while its key stands, since the content it describes has not changed); a recipient whose sender is faulty gets the empty pair, attaches nothing, stays unrecorded by the courier gate and is scanned again next pass, the wait the gate already models for a sender that does not yet track the message.

## Review (by hand)

The boundary catches OSError only, as every reader boundary does: a raise that is not a read fault (a journal row that parses as JSON but is not an event, a bug in a loader) leaves the walk as before and reaches the courier's own catch; a row `json.loads` rejects is skipped by the replay and raises nothing. The counters keep their shape (`{served, built}`); the slot is written only for a complete walk; the first sender in discover order still wins. `_handoff_backref` is the only code changed. One docstring outside it moves: `_log_judge_error`'s row contract said a store-unreadable fault inside a judge pass reaches the pass-crash row, which held only for a strict load (run_propagate's recipient read already filed the row through the same boundary); it now names the readers that file it from inside a pass and keeps the pass-crash clause for a strict load.

## Tests

`tests/test_backref_memo.py`. The raising-store test becomes the per-sender rule. It had pinned the all-or-nothing raise, which was the memo's first shape; the walk the memo replaced answered every sender before a fault, and the per-session boundary is how every other reader of a goal store contains one, so the test's docstring records the change of contract. The fixture gains a third sender after the second in discover order, so a sender past the fault is covered. Three tests red at 1566f7ae with the OSError out of their first lookup, and one that pins a contract that already held:

- a patched read fault on the second sender with its file untouched: the first and third senders answer, the second's message is unanswered, the slot is withheld, one store-unreadable row stands across three lookups; the loader is restored without touching the file, so the key does not move, and the sender answers and the slot is published (this heal is what tells a withheld slot from a cached partial map);
- a directory at the second sender's store path, a real fault with no patching: the same skip, row and withheld slot, and the sender answers once the file is back;
- the courier gate itself (`_courier_link_wanted`, `_attach_courier_link`) with two recipients holding planner-placed unlinked delegates, one from each sender: the readable sender's link lands on its recipient, the faulted sender's recipient is answered None where the walk raised before, and one store-unreadable row covers both recipients;
- a ValueError out of a sender's read propagates, caches nothing and files no row: the boundary is OSError-only, and a wider catch around it fails this test.

Publishing the map despite a skip fails the withheld-slot assertion and the key-stable heal; `break` for `continue` leaves the third sender unanswered; a bare catch around the loader files no row. Full suite: 10126 passed, 61 skipped.

Tier: fix

### DIFF
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -947,8 +947,10 @@ def _log_judge_error(judge, fsid, err, note=None, goal=None, seg=None):
              behind and the shape of the menu that died), "store-quarantined" (a goals store that
              could not be parsed was moved aside to <file>.corrupt-<stamp> and the session started fresh),
              "store-unreadable" (a goals store that cannot be READ — EACCES, EIO — filed once per fault
-             episode by load_goals_or_fault; inside a judge pass the raise instead reaches the pass
-             wrapper's "pass-crash" row), "store-unwritable" (a goals store that read but whose publish
+             episode by load_goals_or_fault and the readers that share its boundary, run_propagate's
+             recipient read and the courier's sender-board walk among them; a reader inside a judge pass
+             that loads a store strictly instead raises into the pass wrapper's "pass-crash" row),
+             "store-unwritable" (a goals store that read but whose publish
              then failed under a user gesture, the save path's own strict read or the write itself; filed
              once per fault episode by save_goals_or_fault, and the gesture is answered on its socket)
       note   the evidence — reply tail, error message, exception name, or the give-up scope + re-arm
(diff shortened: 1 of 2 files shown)


ksarma:
Merging as shipped. One doc nit to fold in later.

Reviewed inline: each sender is read through `load_goals_shared_or_fault`. It catches `OSError` only and returns `(store, None)` or `(None, exc)`. A faulted sender is skipped. The loop reads the view with `.get` alone into a private dict of string pairs, so no view escapes, and `_attach_courier_link` keeps its own writer load. A `ValueError` out of a loader leaves before the counter and slot write, and reaches `run_courier`'s link-attach catch as before. Withholding the slot after a skip is right: `_file_key` is `(inode, mtime_ns, size)`, which a permission fix leaves unmoved, so a published partial map would outlive the fault. A map published before the fault stays correct while its key stands: no save lands without moving `mtime_ns` or `size`, and the save path's strict read faults first. A directory at the path keys as `None` and re-walks. The re-walk during a fault costs one shared-view read per sender per lookup, the pre-#1170 cost, for the episode only. The store-unreadable row files once per sid and fault text, and ends on the next successful boundary read. The counters keep their shape. The first readable sender in discover order wins. The `_log_judge_error` docstring matches `run_propagate`'s boundary read.

- `docs/reference.md:943`, follow-up nit: the backref counter line could add that the map is withheld while a sender store is unreadable. Then built climbing with served flat during a store-unreadable episode reads as that symptom.

Tests: red-first verified at base 50e824f3 with the PR's test file checked out: 3 failed, 10 passed. The two directory-at-path tests failed with `IsADirectoryError` out of `_disk_read`, and the patched-loader test with `PermissionError`, each at its first lookup. The loud-raise test already passed there. At head: `tests/test_backref_memo.py` 13 passed; `test_courier_link`, `test_courier_skip`, `test_goal_store_fault_boundary`, `test_judge_store_cache` and `test_wake_answer_files_under_shared_view` 80 passed. CI: all eleven checks green at head.

Review run by the maintainer automated review session; verified against head 6bcde04c.

PR #1178, opened 2026-09-09
Judge read memos: the eviction at the cap runs under a lock

ksarma:
Symptom. #1171 bounded the captions and goal-archive read memos at 256 entries with an oldest-out eviction in `_memo_put`, which iterates and pops the dict with no lock. The archive memo is filled from every thread that loads a goal store (a journaled restore row reaches it inside `load_goals`), the captions memo from the index tier and the planner workers. At the cap, two concurrent fills for new sids can both pick the same oldest key (the second pop raises KeyError) or one can change the dict's size during the other's iteration (RuntimeError). The per-session boundary (`_or_fault`) catches OSError only, so the raise reaches the feed builders, the tick jobs and a planner pass-crash row, and the index tier's serial caption loop, which has no per-session guard, ends its pass for every session. Narrow (it needs the cap) and self-healing, but a raise on a read path; reproduced by the review record on #1171 with a stress script.

Cause. `_memo_put` evicts with `memo.pop(next(iter(memo)))` and inserts under no lock.

Fix. One leaf lock, `_FILE_MEMO_LOCK`, around the eviction and the insert. The served path is unchanged (one dict read of an immutable tuple; a stale or missing read re-derives, the chain memo's rule), and the counters stay unlocked: a lost increment under-counts a diagnostic and raises nothing. The chain memo already evicts under its own lock. Nothing inside `_memo_put` takes another lock, so the archiver path that reaches it under `_GOAL_ARCH_LOCK` (an undo restore whose `load_goals` replays a restore row) stays safe; the note beside the lock says so. Comment corrections from the same record: the archive memo does cache the fresh loader's empty answer for an existing unreadable or corrupt file (an absent file is never cached), and `load_goal_archive`'s docstring now names the archivers as they are (the rewind take and the dead-branch reconciliation in the judge, the kernel's store compaction and undo restore, all saving under `_GOAL_ARCH_LOCK`) and `load_goal_archive_shared` as the readers' twin. Not here: freezing the memoized archive store (the record's first item), a separate change.

Tests. Two in `tests/test_file_read_memos.py` (class `MemoEviction`), both through `load_goal_archive_shared` with real archive files and the cap patched to 4, made deterministic by hooks on the memo's `pop` and `__setitem__`: two fills that chose the same victim, and a fill landing between another's eviction and insert. Red on main before the change (ab144390): KeyError out of the fill, 20 of 20 runs; 5 entries against a cap of 4, 20 of 20 runs. Green with the lock, 20 of 20 runs. The second test also fails a lock around the pop alone. Each hold is bounded by a one-second timeout, so the class adds about two seconds; the threads are daemons joined under a wall-clock cap that fails loudly. The unlocked `loaded` counter is pinned only where the two increments are ordered; both tests pin `served == 0`, which never races. Full suite: 10125 passed, 61 skipped.

Tier: fix

### DIFF
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -2611,16 +2611,27 @@ def tasks_for(fsid, leaf, files, now):
 # every index pass (19% of the thread), and load_goal_archive decoded goals-archive/<sid>.json per call
 # (11%). Each file is now parsed once per file state: the key is _file_key (inode, mtime_ns, size) taken
 # BEFORE the read (the chain-memo rule), so a row appended during the read moves the key the next call
-# takes and content read mid-write is served no further than that call; an absent or unreadable file is
-# never cached; a stat error (a fresh sentinel key) never matches. Bounded at _FILE_MEMO_MAX entries,
-# oldest-inserted out; a rebound state root clears both. The captions memo serves the three readers from
-# one parse; the archive memo serves READERS only (load_goal_archive_shared): every archiver keeps
-# load_goal_archive, a fresh private object it mutates and saves.
+# takes and content read mid-write is served no further than that call; an absent file is never cached
+# (its key is None); a stat error (a fresh sentinel key) never matches. An existing file that cannot be
+# read is [] and never cached by the captions memo; the archive memo caches load_goal_archive's answer
+# for it, the empty shape, the same answer the fresh loader gives every archiver for that file. Bounded
+# at _FILE_MEMO_MAX entries, oldest-inserted out, the eviction and the insert under _FILE_MEMO_LOCK; a
+# rebound state root clears both. The captions memo serves the three readers from one parse; the archive
+# memo serves READERS only (load_goal_archive_shared): every archiver keeps load_goal_archive, a fresh
+# private object it mutates and saves.
 _CAPTIONS_MEMO = {}        # fsid -> (file key taken before the read, the parsed rows)
 _CAPTIONS_STATS = {"served": 0, "parsed": 0}
 _GOALARCH_MEMO = {}        # fsid -> (file key taken before the read, the guarded archive store: read-only)
 _GOALARCH_STATS = {"served": 0, "loaded": 0}
 _FILE_MEMO_MAX = 256
+_FILE_MEMO_LOCK = threading.Lock()   # the two memos are filled from the index tier, the planner workers and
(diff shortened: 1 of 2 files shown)


ksarma:
Merging as shipped; nothing to change before it lands.

Reviewed inline: `_memo_put` now holds `_FILE_MEMO_LOCK` across the eviction and the insert. Its body is dict operations only, so the lock is a leaf by construction. I traced every path into it. `_captions_rows` reaches it from the index pass. `load_goal_archive_shared` reaches it from the override replay inside `load_goals`, `_cleared_context`, the two peer-node joins, and `run_propagate`. One caller arrives holding another lock: the kernel's undo restore, which calls `load_goals_or_fault` under `_GOAL_ARCH_LOCK`. That order is archive lock then memo lock. Nothing takes them the other way; the archivers' `save_goals` rebase reads raw bytes and never replays the journal. The served path is unchanged: one dict read of a whole tuple and a tuple compare. A miss or stale key re-derives, so it can neither raise nor return a torn entry. The corrected block comment is accurate. An existing unreadable or corrupt archive is memoized as the empty shape under its stat key; an absent file never is. That caching predates this PR. The archiver roster in `load_goal_archive`'s docstring matches all four `save_goal_archive` callers. The tests wait on a Barrier and Events with timeouts as the failure backstop. They join daemon threads under a cap, keep the hooks on a per-test dict, and restore the memo and cap in tearDown. They load the judge under a private module name so the patches stay local under xdist.

Two follow-ups; neither holds the merge.
- `kernel/judge.py:4609`: the docstring reads as if every judge reader takes the shared twin. `_delegate_user_rooted` (`kernel/judge.py:14838`) still reads through the fresh loader. Name it there or switch it.
- `tests/test_file_read_memos.py:356`: the same-victim test never asserts its barrier was reached. An eviction rewritten as `popitem()` or `del` would leave it green and vacuous. Assert the hook fired, as the sibling test does.

Test evidence: at base 50e824f3 with the PR's test file, MemoEviction fails 2 of 2 in 5 of 5 runs (KeyError out of the fill; 5 entries against a cap of 4). At head the file passes 16 of 16, the two tests pass 5 of 5 repeats, and the file passes under xdist. CI: all 11 checks green at head.

Review run by the maintainer automated review session; verified against head f33cf8b7.

PR #1179, opened 2026-09-09
Planner gate: the key carries the leaf's task store, the captions file and each running launch's deadline crossing

ksarma:
## What

Three inputs the planner reads are outside its change gate's key (#1173: a pass that had nothing to do records the session, and it is skipped while the key stands), so a recorded session is not planned again when only they move:

- After a `/clear`, an SDK session's to-do change alone does not re-plan it. Today the backend's TaskCreate/TaskUpdate hook rewrites the reg on the same call and the reg is keyed, so the card still moves; the gate's own contract does not hold there, and when the hook finds the reg unreadable it writes nothing, so no keyed input moves.
- A coerce-floor title on an idle session never heals when its prompt caption lands. Nothing masks this one; it needs a judge outage at mint time to arise.
- A done focus top held at `working` by a launch whose CLI died mid-watch completes only when another input moves. At the default the closer's own settle at the end of each triage pass covers it; `run_plan` alone does not.

The review record on #1173 lists the three.

Cause: `_plan_key` keys the task store under the romp sid while `_sync_declared_plan` reads it under the leaf fsid (the two differ for every SDK session after a `/clear`); `_heal_floor_titles` reads `captions/<sid>.jsonl`, which no term carries; `_awaiting_bg_hold` applies `em._bg_expired` to every running launch with the clock, and no file records the crossing.

Fix: the key stats `tasks/<leaf fsid>/` (the directory the sync reads, replacing the sid's) and the captions file, and carries each running launch's expiry boolean under the pass clock, appended after the existing terms so the record rule's compare of the store key (`pkey[2]`) is unchanged. `_bg_unresolved`, `_awaiting_bg_hold` and `_session_settled` take the pass clock from the planner (a default keeps the wall clock for every other caller), so the key and the verdict read one clock. The source pin on the settled gate's expiry moves with it and still asserts a fresh clock outside the scan cache.

Why a key term and not "never record while a launch runs": a launch with no recorded ceiling (a backgrounded Bash, a dev server) never expires, so that alternative would plan the idle sessions the gate is meant to skip, on every pass. The boolean is the value the derivation reads, keyed the way the awaiting lift's gate in `kernel.py` (`_lift_spent_awaiting`) keys the same crossing, so the key moves once, at the crossing, and not before. Threading the clock moves the planner's expiry view from the wall clock to the pass's start second; a launch whose expiry falls during the pass is judged one pass later, the tolerance the kernel's lift gate accepts.

Overlap with #1062, which is open: it threads the same clock through the same three helpers and rewrites the same source pin. Its code lines (the three signatures, the `now = time.time()` default, the rollup call in `_plan_session`) and its `tests/test_bg_scan_monitor.py` hunk are identical to this PR's, so its rebase meets the same change there; the `_bg_unresolved` docstring differs by a sentence, and #1062 also documents `now` on `_session_settled`, which this PR leaves as it is. Its `_plan_session` tail and `run_plan` loop need a manual resolve. Its evidence signature (`_sig_inputs`) already carries the captions file and the leaf's task store, with a not-before clock in place of the boolean, so on that rebase its gate supersedes the terms added here.

## Review (by hand)

The store key stays at index 2. A key that cannot be computed is a fresh sentinel and the session is planned (the rule `_task_store_key` already applies to a listing error). The death marker needs no term: the kernel writes a states row with it and the states file is in the parse key. The awaiting hold reads the transcript's pairing, the goal store's awaiting stamps and audited turns, the parsed turns, the reg, the death marker and the clock, not the task store (a correction to #1173's body). One extra `em.scan_bg_tasks_cached` call per session per pass in the key: a cache hit on an unchanged transcript.

## Tests

`tests/test_planner_skip.py`, in the module's own fixture. Three are red at 42db9cc5 with the counts stated: a forked leaf's task-store change alone un-skips and mints the mirror (red: `(0, 3)`, no mirror node); a prompt caption landing heals a floor title on a recorded session (red: `(0, 3)`, the verbatim head stays); a launch crossing its deadline re-plans the session and its done focus top completes (red for two reasons: no expiry term, and the wall-clock settle completes the top before the crossing, so the `working` precondition fails first). The same test checks that a pass 100 s later, the launch still inside its ceiling, skips every session, which a term keyed on the raw clock instead of the boolean fails. Three more pin the design: a backgrounded Bash on a recorded session leaves it skipped 200 s later and a day later (the term is a stable False); the settle reads the wall clock when no pass clock is handed in (a frozen clock, no sleeping: past the ceiling the launch is expired, inside it running, and a clock handed in outranks the wall clock); an expiry view that cannot be computed never equals a recorded one. `tests/test_bg_scan_monitor.py`: the source pin now looks inside `_bg_unresolved` for `em._bg_expired(t, now)` and `now = time.time()`. Docs: the `plannerSkip` sentence in `reference.md` names the two files and the crossing. Two comments rephrased for the vocabulary rule.

Tier: fix

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -947,11 +947,12 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   cleared context, each parsed once per file state (`served`, `parsed` or
   `loaded`); `plannerSkip` is the planner's change gate (`skipped`, `planned`,
   `recorded`: a session whose parse, store, journal, archive, episode log,
-  task store and reg have not moved since a pass that had nothing to do is
-  not planned again). The compaction sweep after each judge pass evicts from `pass` and
-  `shared` the entries of stores no session in the discover window owns, so
-  both stay bounded by the live board; the gate memo is bounded by the session
-  count.
+  its leaf's task store, captions file and reg have not moved since a pass
+  that had nothing to do, and none of whose running background launches has
+  crossed its deadline, is not planned again). The compaction sweep after
+  each judge pass evicts from `pass` and `shared` the entries of stores no
+  session in the discover window owns, so both stay bounded by the live
+  board; the gate memo is bounded by the session count.
 - `judge`: `passes`, `ms_sum`, `ms_last`, `ms_mean` (wall time; a pass waits
   on model calls), `cpu_ms_sum` (CPU time of the judge tier threads and every
   per-session worker they run; the workers' share is `cpu_ms_workers`).
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -2184,11 +2184,15 @@ _COURIER_SEEN = {}         # fsid -> the scan key of its last pass that found no
 # store where it was has no new information for the planner by construction, so the pass returns at once.
 # The key is every input the pass reads, taken BEFORE the store read (the chain-memo rule): the parse
 # cache's fileset key bound to the session object, the store file's key with its journal's and archive's,
-# the episode log's key, the session's task-store files (each name with its key: the declared-plan sync
-# reads them), the reg file's key, and the transcript path. Recorded only when the pass did nothing and the
-# store's key after the pass equals the one before it (a heal, a mint, a retirement or a rollup change
-# moves it); a pass with units, placements or a moved store is planned again next pass whatever the key
-# says. A parse the cache does not hold is never keyed. Pruned to the pass's fleet; a rebound root clears.
+# the episode log's key, the LEAF's task-store files (each name with its key: the declared-plan sync reads
+# the leaf fsid's directory, which a /clear forks away from an SDK session's sid), the reg file's key, the
+# captions file's key (the floor-title heal reads it), each running background launch with whether it has
+# crossed its deadline under the pass clock (the settle's one input no file records; see _bg_expiry_key),
+# and the transcript path. Recorded only when the pass did nothing and the store's key after the pass
+# equals the one before it (a heal, a mint, a retirement or a rollup change moves it); a pass with units,
+# placements or a moved store is planned again next pass whatever the key says. A parse the cache does not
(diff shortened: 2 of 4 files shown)


ksarma:
The owner merged this on the reviewed head before this review posted, so there is nothing to fold in.

Reviewed inline: the planner's change gate at b98d7a7c. The key now stats `tasks/<leaf fsid>/`, the directory `_sync_declared_plan` reads, and the captions file `_heal_floor_titles` reads under the same sid. The new `_bg_expiry_key` term pairs each running launch with `bool(em._bg_expired(t, now))` under the pass clock. The launches come from the scan cache `_bg_unresolved` shares, and each pair is the tuple the kernel's `_lift_spent_awaiting` keys. A launch with no recorded deadline is a stable False, so a backgrounded Bash keeps a recorded session skipped. The key and the settle read one clock, so the term moves at the crossing and not before. Both new terms follow index 2, so the record rule's store compare is unchanged. A scan that raises returns a fresh sentinel. On the nothing-to-do path of `_plan_session` I found no input outside the key. The death marker is written with a states row, and the states file is in the parse key. Courier verdicts land in the store. The declared-plan sync's peer reads run only on a mint. The three helpers keep a wall-clock default. The clock-threading lines and the `tests/test_bg_scan_monitor.py` hunk are byte-identical to #1062's head, so its rebase will not conflict there.

Nit, `docs/reference.md:950`: in the plannerSkip sentence, `its leaf's task store` sits inside a list already governed by `whose`. Drop `its`.

Tests: at base 50e824f3 with the PR's test files checked out, 7 failed and 22 passed:
- the forked-leaf and caption tests, at `(0, 3) != (1, 2)`;
- the deadline-crossing test, at its precondition that the top is still working;
- the three design pins and the source pin, on the missing symbol, the old signature and the old body text.

At b98d7a7c, 29 passed. CI: all eleven checks green; privacy scan of the added lines clean.

Review run by the maintainer automated review session; verified against head b98d7a7c.

PR #1180, opened 2026-09-09
T273c: the tier workflow's job is green whenever it posted a verdict; its row says what it does

henrypinkard:
The tier workflow's job is green whenever it posted a verdict, and red only when the evaluation itself failed; its row says what it does.

Tier: fix

## Why
The job "Tier policy evaluation" exited nonzero whenever the verdict it posted was a failure. A contributor's feature awaiting the owner's approval therefore showed two red rows, the required "Tier policy" check run and the job itself, and after several events (opened, edited, synchronize, labeled, the hourly sweep) several rows of each. The user read a PR with seven of nine rows red as broken when one gate was waiting on them.

## What changed
- `scripts/ci/tier_policy_check.py`: the single-PR path returns 0 once `run_one` has posted a verdict, whatever its conclusion. The verdict is the check run's conclusion, not the job's. The job goes red only when the evaluation itself failed: `run_one` re-raises after posting the "evaluation failed" verdict, so a fetch or post error still reaches the job's status and log. The hourly `--all-open` sweep already exited nonzero only on an error; unchanged.
- `.github/workflows/tier-policy.yml`: the job is named "Tier policy: post verdict"; the header explains that the job's status is not the verdict. It still checks out only the base branch; permissions unchanged.
- `docs/pr-tiers.md`: names the two required checks ("Exactly one tier label" and the API-posted "Tier policy"), says what the job row means, and why a PR shows several rows per name.

## Duplicate rows
They are GitHub showing every run on the head, from three sources. (1) `pr-tier.yml` runs one "Exactly one tier label" job per event under `concurrency: queue: max`, which keeps every pending run so the last one to execute reads the final label set; that is by design (a cancelled required check blocks the merge, and a replaced run judged a stale label set on pull 1039). (2) `tier-policy.yml` runs its job once per event and once per hourly sweep, each leaving a job row on the head. (3) `post_check` creates a new "Tier policy" check run per evaluation, so the same head accumulates one per evaluation; the merge box follows the newest same-named run, which the hourly approval propagation relies on. Nothing in the concurrency settings should change. The one thing that would collapse rows is `post_check` updating the head's existing "Tier policy" run in place instead of creating a new one; a small follow-up if the rows still read as noise once the job stops going red.

## Tests
`tests/test_tier_policy.py`: `JobExitCode` pins the exit for a failing verdict (0), a passing one (0), an evaluation error (the exception reaches the job), and the sweep's two cases (0 with failing verdicts, 1 when one PR's evaluation raised, with the error on stderr). The job-name pin follows the rename; the check-name pin still holds (no job named exactly "Tier policy"). Tier modules: 137 passed.

Touches `.github/` and `scripts/ci/`; merges by hand on the user's word, auto-merge not armed.


ksarma:
Recommend merging as shipped. One clause in `docs/pr-tiers.md` needs correcting in a follow-up.

Reviewed inline: The single-PR path of `scripts/ci/tier_policy_check.py` returns 0 once `run_one` has posted a verdict. `run_one` still re-raises after posting the "evaluation failed" verdict, and the final `post_check` sits outside the try. So a fetch or post error fails the job, and no path exits 0 without a posted verdict. The merge gate never read the job. The live rules on main require the six CI contexts, "Exactly one tier label" and "Tier policy" pinned to integration 15368. No `workflow_run` trigger, `needs:` or script in the repo keys on the job's name or conclusion. A green job beside a failing verdict therefore cannot admit a gated PR. The renamed job cannot equal the check name, the quoted line is pinned, and the YAML parses. Triggers, permissions and the base-only checkout are unchanged. The rows on this PR still show the old name because `pull_request_target` runs main's copy, so the rename first runs after merge. `docs/pr-tiers.md` and the CLAUDE.md tier section agree on the two required checks and the role gates. The `fix` label fits: a wrong exit code with a red-first test, a rename and a docs paragraph beside it. Body and commit match the diff and quote no one. The added lines carry no private strings, hostnames, home paths or banned vocabulary.

- Follow-up, `docs/pr-tiers.md:3`: the clause that a PR shows several rows per name holds for the job and label-counter rows, not for "Tier policy". Those rows come one per workflow run, each in its own check suite. Every API-posted run on a head lands in one suite, and the merge box shows only the newest, so one row (8 runs, 1 row on #1030). The body's duplicate-rows account also has the hourly sweep leaving a job row on the head. Schedule runs attach to main's sha, so the sweep leaves only the verdict. The proposed in-place update of the check run would remove no visible row. It would also need the fetcher to read `/commits/<sha>/check-runs`, which a WorkflowPins test forbids.
- Nit, `tests/test_tier_policy.py:675`: the anti-collision pin does not match the quoted form `name: "Tier policy"`, the quoting the file now uses.
- Nit, `tests/test_tier_policy.py:1317`: the post-error half of the loud-failure claim has no test. `JobExitCode` replaces `run_one` wholesale, so one `FetcherShapes` case raising from the check-runs POST would pin it.

Red-first at main (67773fb2) with the PR's test file: `WorkflowPins::test_the_job_name_is_NOT_the_check_name` and `JobExitCode::test_a_failing_verdict_is_a_posted_verdict_and_the_job_exits_zero` fail (the exit is 1, not 0); 2 failed, 125 passed. At head, `test_tier_policy.py` and `test_pr_tier_workflow.py`: 137 passed. CI at c6a20aa5: all 11 rollup rows green.

Review run by the maintainer automated review session; verified against head c6a20aa5.

henrypinkard:
Thanks for the review. All three items are on the PR now:

- `docs/pr-tiers.md`: the paragraph says which rows repeat (the label-counter row and the "Tier policy: post verdict" job row, one per workflow run, each in its own check suite) and that the API-posted **Tier policy** check shows one row because every posted run on a head lands in one suite and the merge box shows the newest; it also says the hourly sweep's runs attach to `main`'s sha, so on a PR's head the sweep leaves only the verdict.
- `tests/test_tier_policy.py`: the anti-collision pin now covers the bare, double-quoted and single-quoted forms of the name (and a trailing YAML comment).
- `tests/test_tier_policy.py`: a `FetcherShapes` case pins that a failing check-runs POST fails the job, for the plain post failure and for the evaluation-failed post.

The PR body's "Duplicate rows" account still carries the two points you corrected; the docs are the reference. Merging as a fix on green.

ksarma:
Addendum: the PR merged at e2b78cbe (merge b5e187ec, 2026-09-09 20:18Z). The one commit after the reviewed head c6a20aa5 folds the three items of the record above: the docs paragraph now says which rows repeat and that the hourly sweep leaves only the verdict on a PR's head, the anti-collision pin covers the quoted name form, and a new case pins that a refused verdict post (a 502 from the check-runs API) fails the job instead of exiting 0, including when the evaluation itself also failed. Read against the diff; nothing else changed. Review run by the maintainer automated review session; verified against head e2b78cbe.

PR #1181, opened 2026-09-09
Meta commands: the route no longer evaluates the drive-op gate itself; each setter already decides

ksarma:
Symptom. `/effort` and `/fast` typed into the composer, picked from the timeline lane menu or sent through POST /send paid the drive-op gate twice on the handler thread, and `/model` paid it once, for a value nothing read. `_ops_gate` is the expensive gate: a tmux fork, a discover pass, a read of the usage file and the backend's `busy()`. `_route_meta_command` evaluated it before dispatching, and the effort and fast setters evaluated it again.

Cause. `_route_meta_command` assigned `gate = _ops_gate(sid)` and never read it. Every branch takes `queued` from its setter's return: `_set_effort_or_park` and `_set_fast_or_park` decide under `_gate_or_park`, which runs `_ops_gate` itself, and `_set_model_or_park` has its own rule and never calls `_ops_gate`. The review of #954 (#986) removed this read on 2026-09-07; the #923 branch, rebased past that removal, re-added it in place of the comment that explained it, and the #923 merge the same day carried it to main. The line's comment said `state` reads it, which was not so; ruff reports the line as F841.

Fix. In `kernel/kernel.py`, `_route_meta_command`: the assignment is removed. The docstring's `state` clause and the `/model` branch comment now say that `queued` is each setter's own return and that the route evaluates nothing itself. The parked-or-fired outcome and the `queued` answer are unchanged.

Tests. `tests/test_kernel_meta_command_gate.py` patches `_ops_gate` with a counting stub and drives the three commands through `_route_meta_command`. `/effort` and `/fast` evaluate the gate once and `/model` never, both when the setters fire (each backend call recorded once, no queue created) and when they park (nothing fired, the three ops queued in press order), with `queued` read correctly in both. Before the change both cases fail at their first assertion, two evaluations for `/effort` instead of one. The full pytest suite passes.

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -27868,10 +27868,13 @@ def _route_meta_command(be, sid, text, client=None, floating=False, state=None):
     command, a bare "/model" (the CLI's own picker), plain text that merely contains one — is the
     caller's to send verbatim: the CLI owns what executes. A refused fast toggle is told to
     the client (fail loudly): a dormant SDK session has no live CLI to apply it, and the typed text
-    used to at least draw the CLI's own refusal. `state`, when given, receives {"queued": bool} — whether
-    the change PARKED: the effort/fast setters park under _ops_gate (read here), the model setter returns
-    its own verdict (_set_model_or_park) — so POST /send answers `queued` for a meta command exactly as for
-    a text send (2026-09-03: a parked /model read as plain 'ok')."""
+    used to at least draw the CLI's own refusal. `state`, when given, receives {"queued": bool}: whether
+    the change PARKED, taken from each setter's own return, so POST /send answers `queued` for a meta
+    command exactly as for a text send (2026-09-03: a parked /model read as plain 'ok'). The effort/fast
+    setters report whether they parked under _gate_or_park (the one _ops_gate evaluation those two pay),
+    the model setter has its own rule (_set_model_or_park), and this route never evaluates _ops_gate
+    itself: a read here cost every command a tmux fork, discover pass and usage read for a value nothing
+    used (#986, the review of #954, removed it; the #923 merge restored it)."""
     head, _, rest = (text or "").strip().partition(" ")
     value = rest.strip()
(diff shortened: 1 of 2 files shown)


ksarma:

Reviewed inline: the removed `gate = _ops_gate(sid)` in `_route_meta_command` had no reader. The name appears nowhere else in the function, and ruff flags it F841 at the base and not at the head. `_ops_gate` only reads state (compacting, working, pending ops, a move in progress, the account limit hold). It warms no cache and writes no park state, so dropping the route's own evaluation loses no side effect. The parks happen in `_gate_or_park`, which the effort and fast setters still run. `_set_model_or_park` keeps its own rule and never calls `_ops_gate`. Every branch still takes `parked` from its setter's return, and `state["queued"]` is untouched. The parked-or-fired outcome and the queued answer are therefore unchanged. Only the extra evaluation is gone: a tmux fork, a discover pass, a usage read and the backend's `busy()` on the handler thread. The history claim holds: d9f9c113 in #986 removed the line on 2026-09-07 with a comment saying why. 1648035d in #923 was rebased past that removal the same day and put the line back; the #923 merge carried it to main. The test patches `_ops_gate` with a counting stub and stubs the model setter's own gates, so nothing forks tmux. It covers the fire and park cases and pins three ops queued in press order. It uses a private synthetic sid and writes only under its own state root. Docstring and comment edits match the code. The added lines carry no private strings, hostnames or home paths. The `fix` label fits.

Test evidence: at the base 67773fb2, with the PR's test file checked out, `tests/test_kernel_meta_command_gate.py` fails both cases at their first assertion. Line 78: `(2, False) != (1, False)`; line 87: `(2, True) != (1, True)`. At 39762df7 both pass. CI at 39762df7: every check green (Python 3.10 to 3.13, Shell, vscode-extension, both tier checks).

Review run by the maintainer automated review session; verified against head 39762df7.

PR #1182, opened 2026-09-09
Reference: the Billing row's four readings, as render.ts renders them, pinned

ksarma:
docs/reference.md described the tab hover's Billing row, when the CLI's report disagrees with the pick, as `Login (CLI reports API key)`. ui/webview/render.ts renders no such text.

The hover row (the `Billing` entry among the tab tooltip's rows, `ui/webview/render.ts` line 5145) has four readings:

- The default, when none of the three below applies: `API key`, or `Login (name@example.com)` (`Login` alone when the account name is unknown). The code needs no report from the CLI for this reading; an empty `authLive` (no init landed yet) renders it too.
- A switch is still reconnecting the session (`authPending`): `Login (applying — not confirmed yet)`, or the same after `API key`.
- The pick names a side this machine cannot bill (`authPickUnavailable`): `⚠ Login picked, but no Claude login signed in on this machine — this session bills the API key`, the reason being the kernel's (`kernel/credentials.py`, `WHY_NO_LOGIN`); with neither side, the tail reads `— nothing to fall to, so the launch went out as picked`.
- The CLI's own report contradicts the pick (`authLive`): `⚠ Login picked, but the CLI reports the API key — this session bills that`, and the same with the sides swapped for a key pick.

The tab menu's Billing sub-line (`ui/webview/render.ts` line 6105) mirrors them in fewer words: `API key` or `Login (name@example.com)`, `applying…`, `⚠ login unavailable, billing API key`, and `⚠ CLI reports API key`.

The row is an SDK session's. A tmux session's billing romp cannot know, and a Codex session bills no Claude account; `Sessions.live()` forwards no `auth` for either, and the row renders only under `if (s.status.auth)`, so neither tab carries one.

Change:

- `docs/reference.md`: the hover-row paragraph now names each reading with the text the code renders, and the sub-line's mirror of each. The paragraph reads:

  > An SDK session's chat tab carries the same fact as a `Billing` row in its hover tooltip, one-auth machines included; tmux sessions, whose billing romp cannot know, and Codex sessions, which bill no Claude account, show no row. The row has four readings. Unless one of the three cases below applies, it reads `API key` or `Login (name@example.com)` (`Login` alone when the account name is unknown). While a switch is still reconnecting the session, the row appends `(applying — not confirmed yet)` to the side: `Login (applying — not confirmed yet)`. A pick naming a side this machine cannot bill leads with the warning, the reason, and the side the launch fell to: `⚠ Login picked, but no Claude login signed in on this machine — this session bills the API key`; with nothing to fall to, the tail says the launch went out as picked. A pick the CLI's own report contradicts (a login pick whose CLI reports a key, a key pick whose CLI landed on the login) leads with the warning too: `⚠ Login picked, but the CLI reports the API key — this session bills that`, and, for a key pick, the same with the sides swapped. The tab menu's Billing sub-line says the same in fewer words: `API key` or `Login (name@example.com)`, `applying…`, `⚠ login unavailable, billing API key`, and `⚠ CLI reports API key`.

  One word in the paragraph above it changes too: the sentence on an explicit pick of the missing side said the fall shows "on the card's Billing sub-line"; the sub-line lives in the tab's right-click menu (`showTabMenu`), so it now says "the tab menu's Billing sub-line", the name the new paragraph uses.

- `tests/test_reference_billing_label.py` (new): pins the copy `render.ts` renders each reading from (` (applying — not confirmed yet)`, ` picked, but the CLI reports `, ` — this session bills that`, `⚠ CLI reports `, `, billing `, the side words, and the rest) and the kernel's reason string the unavailable reading interpolates, then asserts that `docs/reference.md` quotes the readings assembled from those same fragments and no longer carries `(CLI reports`. The pins are the rendered text, not the TypeScript around it: a rename or a reshuffle of the row's code that leaves every rendered word as it was stays green (checked by renaming `s.status.` to `st.` throughout the hover row: 8 passed), and a wording change turns the module red before the reference goes stale (checked twice, rewording `applying — not confirmed yet` and the sub-line's `unavailable`: one code pin red each, every doc pin green). Against the previous paragraph the five doc pins fail and the three code pins pass. The fix for a red run after a wording change is to change the paragraph and the pinned fragment together.

No kernel, backend, judge or webview code changes.

Tier: docs

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -317,7 +317,7 @@ a remembered key pick on a helper-less machine already fell to the login, and
 the fall is said once per process as a problem row. An explicit pick that
 names the missing side (a session picked "login" on a box that later lost its
 login) launches on the other side when one exists and says so once per session
-start, on the card's Billing sub-line as `⚠ login unavailable, billing API key`
+start, on the tab menu's Billing sub-line as `⚠ login unavailable, billing API key`
 and in the log; the fall itself rides the status as `authPickFell`, so the hover
 and the sub-line never infer one. A pick with nothing to fall to (a box with
 neither side) launches as picked and the CLI decides; the sub-line then says
@@ -350,13 +350,23 @@ covered by the picker: their CLI lives in the tmux server's environment, which
 the kernel does not control, and resolves its credential the way any `claude`
 in a terminal does.
 
-Each chat tab's hover tooltip carries the same fact as a `Billing` row,
-`API key`, or `Login (name@example.com)`, whenever the session's backend
-reports it, one-auth machines included; only tmux sessions, whose billing romp
-cannot know, show no row. When the CLI's own report disagrees with what the
-session was launched for (a login pick whose CLI reports a key, a key pick
-whose CLI landed on the login), the row carries both: `Login (CLI reports API
-key)`.
+An SDK session's chat tab carries the same fact as a `Billing` row in its hover
+tooltip, one-auth machines included; tmux sessions, whose billing romp cannot
+know, and Codex sessions, which bill no Claude account, show no row. The row
+has four readings. Unless one of the three cases below applies, it reads
+`API key` or `Login (name@example.com)` (`Login` alone when the account name is
+unknown). While a switch is still reconnecting the session, the row appends
+`(applying — not confirmed yet)` to the side: `Login (applying — not confirmed
+yet)`. A pick naming a side this machine cannot bill leads with the warning,
+the reason, and the side the launch fell to: `⚠ Login picked, but no Claude
+login signed in on this machine — this session bills the API key`; with
+nothing to fall to, the tail says the launch went out as picked. A pick the
+CLI's own report contradicts (a login pick whose CLI reports a key, a key pick
+whose CLI landed on the login) leads with the warning too: `⚠ Login picked, but
+the CLI reports the API key — this session bills that`, and, for a key pick,
+the same with the sides swapped. The tab menu's Billing sub-line says the same
+in fewer words: `API key` or `Login (name@example.com)`, `applying…`, `⚠ login
+unavailable, billing API key`, and `⚠ CLI reports API key`.
 
 Failures are loud rather than silent: a session that lands on the other auth
 than it was launched for is flagged in the Log panel, and a dead credential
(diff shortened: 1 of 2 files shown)


ksarma:
Merging as shipped. Two follow-ups below; neither needs a change here.

Reviewed inline: I checked every string the new paragraph quotes against `ui/webview/render.ts` at this head. The hover row (render.ts 5145 to 5160) and the tab menu's Billing sub-line (6105 to 6111) render the default, applying, unavailable-pick and contradicted-pick readings with the text the doc quotes. Warning glyph and dash spelling match. The reason the unavailable reading interpolates is `kernel/credentials.py` WHY_NO_LOGIN. The row renders only under the auth status. Sessions.live forwards no auth for tmux rows or Codex rows, so the claim that neither tab carries one holds. The new paragraph's account of the fall holds for the webview's inference from authAvail, which is what runs at this head. The diff touches `docs/reference.md` and a new text-only test. Nothing under kernel, ui, bin, cli or postal changes, so the docs label fits. The added lines carry only the name@example.com placeholder. Body and commit message match the diff.

Follow-ups:
- `docs/reference.md:321` (follow-up). The sentence this PR edits still says the fall "rides the status as authPickFell", so the hover and sub-line never infer one. `kernel/kernel.py` forwards no such field at this head; Sessions.live and build_session forward authPickUnavailable only. So render.ts authFellTo always infers from authAvail. The two diverge for a key pick on a helper-less box with an unreadable account file. Already recorded on #1147. Either land the two one-line kernel merges with a payload pin, or reword the clause.
- `tests/test_reference_billing_label.py:44` (follow-up). UNAVAILABLE_HEAD and FELL_TAIL are substrings of the contradicted reading's fragments at render.ts 5157 and 5158. Rewording the unavailable reading alone (5151 or 5154) therefore leaves the module green while the doc's quote goes stale. A mutation check confirmed this. Pin each with the interpolation brace that ends it; the brace occurs once per fragment.

Tests: `tests/test_reference_billing_label.py` at base 67773fb2 with the PR's test checked out: 5 failed (the five doc pins), 3 passed (the render.ts pins). At head: 8 passed. CI at 2bec80fb: every check green (Python 3.10 to 3.13, bats, vscode-extension, both tier checks).

Review run by the maintainer automated review session; verified against head 2bec80fb.

PR #1183, opened 2026-09-09
File viewer: the edit-mode notice sits above the body, not inside it

ksarma:
## Symptom

In the file viewer's edit mode, a notice (the plain-textarea fallback when the editor bundle cannot load, or a refused save such as "changed on disk since you opened it") cuts off the bottom of the editor by its own height, and scrolling the editor carries the notice out of view.

## Cause

`.fileview-body` is `flex: 1 1 auto; min-height: 0; overflow: auto`, and `.fileview-editor` / `.fileview-cm` are `height: 100%` inside it. Both notices were mounted with `body.prepend(bar2)`, so the body's content was the bar plus 100% of the body: the editor overflowed by the bar's height, and the body's scroll took the bar with it. The comment beside the save-failure mount says the bar is meant to sit above the textarea; the DOM put it inside the same scroller.

## Fix

- One `noteBar(msg)` helper in `file-view.ts` builds the notice and mounts it as a child of the card between the title bar and the body (`box.insertBefore(bar2, body)`), replacing any notice already up. The fallback-editor path and the refused-save path both call it; the conflict notice still appends its Reload button to the returned bar.
- `.fileview > .fileview-err { flex: 0 0 auto; }` in `styles.css` and `feed.css` (byte-equal): the card is a column flex container, so the bar keeps its height and the body row shrinks under it, and the editor's 100% resolves against the smaller body.
- The card holds its notice by reference (a closure variable), and `exitEdit` removes that notice explicitly: the body swap that used to take it no longer reaches it, and a save that landed, Cancel and Escape all leave through `exitEdit`, while Reload file re-opens a fresh card. The reference rather than a lookup by id matters because every open registers a document keydown handler that only an Escape-close removes, so after Reload file the replaced viewer's handler still runs its `exitEdit` on Escape; a lookup by id from there would strip the live card's notice, Reload button included, while that card is still in edit mode over a dirty buffer. (The handler leak itself is untouched here.)

No new font size, no new control, and the notices keep their copy. The read-mode failure notices (a refusal, a decode failure) are unchanged: those replace the body's content and have no editor beneath them, and the new rule does not select them.

## Tests

- `ui/webview/file-view-notice.test.ts` (new) runs `openFileView` against a copy of the DOM stand-in `fileview-chip.test.ts` and `github-link.test.ts` use, extended with a parent/child tree (`parentNode`, `insertBefore` that throws on a reference that is not a child as the DOM does, `prepend`, `remove`, `getElementById` over it). The text lands, Edit is clicked, the editor chunk has no bundle `<script src>` to derive its URL from, and the fallback notice must be the card's child directly above `.fileview-body` with the textarea alone in the body. Then Cancel, Escape, a save that lands and Reload file each leave no notice standing; a refused save's notice replaces the last with the edited buffer intact beneath it and its Reload button inside; and Escape run through a replaced viewer's handler and the live one's, oldest first, leaves the live card's notice up and its dirty buffer in edit mode. Before the change every case fails at the mount (the parent is `.fileview-body`); with `exitEdit`'s removal deleted, the Cancel, Escape and saved-exit cases fail; with the removal done by id instead of by reference, only the last case fails.
- `file-edit.test.ts`: the mount pin moves from `body.prepend(bar2)` to `box.insertBefore(bar2, body)`, asserts the old mount is gone, and pins the refused-save path to `noteBar(err)`.
- `fileview-parity.test.ts`: `.fileview > .fileview-err {` joins the byte-equal rules.

`npm run typecheck` and `npm test` (3654 passing) in `vscode-extension`.

Tier: fix

### DIFF
--- a/ui/webview/feed.css
+++ b/ui/webview/feed.css
@@ -1458,6 +1458,9 @@ a.fileview-gh-note { border-style: dashed; }
 /* a refusal explains itself in place, in the kernel's own words (too large — with the size and the
    cap; not a text file; not found) over the path it was asked for. Never a blank pane. */
 .fileview-err { padding: 18px 14px; font-size: 0.86em; color: var(--warn); }
+/* the edit-mode notice (file-view.ts noteBar) is a row of the card, between the title bar and the body:
+   it keeps its height and the body row shrinks under it, so the editor's 100% is the body it can see */
+.fileview > .fileview-err { flex: 0 0 auto; }
 .fileview-err-hint { margin-top: 6px; color: var(--dim); font-size: 0.92em;
   font-family: var(--mono, ui-monospace, monospace); word-break: break-all; }
 /* the way out of a refusal: when the file exists but cannot render (413/415), the kernel's words are
--- a/ui/webview/file-edit.test.ts
+++ b/ui/webview/file-edit.test.ts
@@ -49,7 +49,12 @@ test("no exit path can silently eat an edited buffer", () => {
 });
 
 test("a conflict keeps the buffer, says why, and Reload asks before discarding", () => {
-  assert.match(VIEW, /body\.prepend\(bar2\);/, "the error bar sits ABOVE the textarea — the buffer survives");
+  // a child of the CARD between the title bar and the body (noteBar), never the body's first child: inside
+  // the body it sat above an editor that is 100% of that same body, so the editor's bottom was cut off by
+  // the bar's height and the body's scroll carried the bar away (file-view-notice.test.ts runs the mount)
+  assert.match(VIEW, /box\.insertBefore\(bar2, body\);/, "the error bar sits ABOVE the body that holds the textarea, so the buffer survives");
+  assert.doesNotMatch(VIEW, /body\.prepend\(bar2\);/, "never inside the body, where the editor's 100% height leaves it no room");
+  assert.match(VIEW, /const bar2 = noteBar\(err\);/, "the refused-save notice goes through the one helper that mounts there");
   assert.match(VIEW, /if \(\/changed on disk\/\.test\(err\)\) \{/);
   assert.match(VIEW, /dirty = false;\s*\/\/ confirmed once — the replace guard must not ask twice/);
 });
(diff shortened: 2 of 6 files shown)


ksarma:

Reviewed inline: `noteBar` mounts the notice with `box.insertBefore(bar2, body)`. That is a card row between the title bar and the body, not the first child of the scrolling body, whose editor is height 100%. It removes any notice already up before mounting the next. Both `body.prepend` mounts are gone. `exitEdit` removes the held reference, so Cancel, Escape and a landed save leave no notice standing, and Reload file re-opens a fresh card. Holding the notice by reference is right. Every open leaves a document keydown handler that only an Escape-close removes, so a replaced viewer's `exitEdit` would otherwise strip the live card's notice by id. The last new test pins that. The `.fileview > .fileview-err` rule is byte-equal in both sheets and adds only flex, so the notice keeps its class, its size and `var(--warn)`. The child selector does not reach the read-mode notices, which mount inside the body. The Reload button's listener sits on a node that only a later refused save rebuilds. Nothing outside `file-view.ts` re-renders the viewer, and the confirm dialog acknowledges the click. The DOM stand-in models `parentNode`, `insertBefore` throwing on a non-child reference, `prepend`, `remove` and `getElementById` over the connected tree, the operations the pins rely on. The added lines carry no private strings, hostnames or home paths.

Follow-ups, neither a hold:
- `ui/webview/file-view.ts:834`: the keydown handler leak the body names is unchanged here. A stale handler's `exitEdit` also nulls the module-level `editHooks`, so a live save in flight at that Escape loses its ack or conflict notice. Removing `onKey` in `closeFileView` and on the replace path would close it.
- `ui/webview/file-view.ts:793`: the saved hook's in-flight-typing branch returns without `exitEdit`. An earlier refused save's notice therefore stays up after a later save lands while the user keeps typing. Pre-existing; removing the notice before that branch would meet the invariant the PR states.

Red-first at base 67773fb2 with the PR's test files: all six file-view-notice cases fail at the mount with parent `.fileview-body`, file-edit case 5 fails on the moved pin, and fileview-parity fails on the missing rule. At head all pass, npm test 3654/3654, tsc clean. CI: all eleven checks green at head.

Review run by the maintainer automated review session; verified against head 910ef34d.

PR #1184, opened 2026-09-09
Courier: a raise in one session's scan is that session's pass-crash row, not the tier's

ksarma:
## What

A raise inside one session's courier scan files that session's pass-crash row and the pass goes on. Before, it left the tier: `run_courier` exited before its write loop (no session's delegate placed that pass), and `run_triage`, one try/finally around the tier sequence, skipped the propagate, group, consolidate and distill passes too. `_run_tier` wrote a stderr traceback and no judge-errors row, so the fault surfaced nowhere but the kernel's stderr, and the pass repeated the same way while the fault lasted.

**Cause.** `run_courier`'s per-session try covered the parse (note `parse:`), the store read (`store:`) and the link attach (`link-attach:`) and nothing else. `_session_settled` (through `_awaiting_bg_hold` and `_bg_unresolved`), `episode_floor`, `_segs`, `_seg_peer`, `_placed_key`, `_unit_text` and `_seg_anchor` ran outside it. `_bg_expiry_key`'s docstring notes that a settle that raises still raises; in the planner that raise lands in a per-session future and `run_plan` files it as that session's row, in the courier it left the tier.

**Fix.** `kernel/judge.py`, `run_courier`: the scan from the settle to the end of the turn walk runs under the same per-session boundary as the parse and the store read. On a raise the session files one `pass-crash` row with note `scan: <repr>`; the rows its unfinished walk queued this pass are dropped, and only those (a scan that raised places nothing, as one whose parse or store read raised places nothing, and the next completed scan queues them again); a record from an earlier scan is dropped, as the rows-to-place arm drops it, so the session is scanned again next pass and placed once the fault lifts; the loop moves to the next session. Unchanged: the write loop, the cross-host plant, the scan key, the record rule and the link-attach catch. The gate's header comment names the new case. The judge.py hunk is mostly a re-indent of the walk under the new try; `git diff -w` shows the lines that changed.

No card-move rule changes: a session whose scan raised placed nothing before (the tier left) and places nothing now; the other sessions place what the pass places when no scan raises.

## Tests

`tests/test_courier_skip.py`, `CourierScanCrash` on the module's fixture (three sessions, a declared delegate delivered to each, `courier_llm` stubbed). The poisoned session is C, which `discover` scans after A and B (the names directory, sorted), so A's and B's rows are queued when C's scan raises; each poison records the order it saw and the assertions check that premise. All five red at 67773fb2: the `RuntimeError` leaves `run_courier`, and in the last case `run_triage`.

- The segment walk raises on the second turn of C, which holds two delegates: A and B place in that pass; one row (`courier`, C, note starting `scan: RuntimeError(`) is filed; the row C's first turn queued is not placed; C is absent from `_COURIER_SEEN`. With the poison lifted, the next pass plants both of C's delegates and C records and skips like the others, with no further row.
- The same with `episode_floor` raising for C, and with `_session_settled` raising for C.
- C recorded, a delivery moves its inputs, the rescan raises: C alone is scanned, its record is dropped rather than kept under the stale key, and it places on the next pass.
- `run_triage` with the other tiers stubbed and the walk raising for C: every tier after the courier runs, in order, and A and B place.

Each line of the boundary has a case that fails without it: the settle moved above the try (the settle case), `del pending[:]` in place of `del pending[n_pending0:]` (the four cases with A's and B's rows queued), the `del` removed (the segment-walk case), the `pop` removed (the stale-record case). With the whole try reverted, all five fail.

Full suite locally: 10142 passed, 61 skipped, and one browser-clock case in `tests/test_kernel_webpush.py::LandingRevealOffers` (`ageS` 6 for 5) that passes alone and reads nothing this change touches.

The courier's change gate (`_COURIER_SEEN`, `_courier_scan_key`, `repair_open`, the drifted-key rule) is untouched; the change widens the crash boundary around the scan the gate keys.

Tier: fix

### DIFF
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -2175,7 +2175,8 @@ _PARSE_CACHE = {}          # fsid -> (fileset_key, parsed_session)
 # file's key and its journal's and archive's (the shared view's inputs), the episode log's key (the
 # floor) and the transcript path. A session is recorded only when its scan added nothing pending; one
 # with rows to place is scanned again next pass however its inputs stand (its placements move the store
-# anyway). A parse the cache does not hold (a stubbed one) is never keyed, so never skipped. Pruned to
+# anyway), and so is one whose scan raised (its pass-crash row is filed, the rows it queued are dropped).
+# A parse the cache does not hold (a stubbed one) is never keyed, so never skipped. Pruned to
 # the pass's fleet, so bounded by it; a rebound state root clears it.
 _COURIER_SEEN = {}         # fsid -> the scan key of its last pass that found nothing to place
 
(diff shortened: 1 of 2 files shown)


ksarma:
Merged as shipped by the owner before this record posted (0ce4da04). Nothing would have held it.

Reviewed inline: With whitespace ignored, the `kernel/judge.py` change is two header-comment lines on the `_COURIER_SEEN` gate, `n_pending0` hoisted above a new `try`, `repair_open` kept inside it, and one `except` block. The rest is a re-indent. The write loop, cross-host plant, scan key, record rule and link-attach catch are unchanged. A raise before `closed[fsid]` is assigned is safe. Every downstream read is `closed.get(fsid, False)`. The only `pending.append` is inside the walk. The truncation drops only rows appended after `n_pending0`, which is taken after the earlier sessions' rows landed. The `continue` skips the record step, and the `pop` drops any earlier record. The loop-head skip check reads `_COURIER_SEEN.get(fsid)`, so the session is scanned again next pass. The note carries `repr(e)` only, into a log under local state. That is the same exposure as the existing parse, store and link-attach rows. A session whose scan raised placed nothing before and places nothing now. The other sessions and the later tiers now run in that pass, which is the fix. The tests stub `courier_llm`, use private synthetic sids, and assert the A, B, C scan order rather than assume it. The `fix` label fits: an existing boundary widened, no new mechanism.

Red-first verified: At 67773fb2 with the PR's test file, `tests/test_courier_skip.py` gave 5 failed, 14 passed. Each failure was the poison's `RuntimeError` leaving `run_courier` (and `run_triage` in the tiers case). At 71f62083, 19 passed. Mutations of each boundary line (`del pending[:]`, the `del` removed, the `pop` removed, the settle above the `try`) redden exactly the cases the body names. CI at head: every required check green, both tier checks included.

Review run by the maintainer automated review session; verified against head 71f62083.

PR #1193, opened 2026-09-09
Staged messages release as one message, and the strip shows four and scrolls the rest

ksarma:
A release of the staged run goes out as one message, and the strip above the composer shows about four items and scrolls the rest.

## Tier

Tier: feature

## Design

**One message per release.** `stagedPosts` and `stagedRunBody` (ui/webview/staged-messages.ts) compose the staged items in stage order, each as its quote block(s) and then its comment, byte for byte what the item sent on its own, separated by a blank line, with the typed message last; `flushStaged` routes the result through the existing `routeUserMessage`, one call per post. One post, one bubble, one turn. Send now and the next-message path share the composition, and with nothing staged the typed message routes exactly as before.

**What e35381e3 set, and what changes.** That commit released a run as a series of messages, each keeping its quote, so the run read quote, comment, quote, comment. The property survives: the one body keeps every quote with its comment in stage order, so the agent reads the same sequence. What changes is how the run arrives. A series lands in pieces. The first note starts a turn on its own, and the rest reach the agent only at its next tool boundary, or after the turn when it has none; the agent begins its answer before it has read the whole run, and the chat shows a bubble per note. One message is one turn over the whole run, read in full before it is answered. If the series is preferred as a choice, a setting (one message or a series) is the fallback: the composition is a pure function, so the setting would select between `stagedPosts` and a per-item loop.

**Two kinds of item still go alone, at their place in stage order.** A goal follow-up: the kernel wraps one goal per message through askFollowUp, so a staged item citing a goal posts on its own, and a typed goal citation wraps the run before it as that follow-up's text. A slash command, staged or typed: the kernel fires a command only at the head of its own text, so after another section a typed `/clear` would be prose the agent read, and ahead of one a staged `/compact` would take the comments after it as its argument. `isSlashCommand` mirrors the kernel's `_SLASH_CMD_RE` shape (a slash, a name, then whitespace or the end, at the head of the trimmed text), with a Unicode letter and digit class for the name's tail because Python's `\w` is Unicode where JavaScript's is ASCII. A command staged with a quote chip goes with its chip, and `routeUserMessage` wraps it as it wraps any quoted message, so the CLI reads it as prose; that is what it did when every item was its own message, and going alone costs it nothing. Several goals in one message would need a multi-goal askFollowUp and is out of scope; the kernel is unchanged.

**`quoteReplyBody` moves into the module.** The live send, the release and the chip's audit preview compose from one function, and the test executes it. Its no-quote case returns the text alone; render.ts only ever called it with at least one quote, so existing sends are byte-identical.

**The strip.** The items sit in a `.staged-list` under the head with `max-height: 209px`, derived from the existing chip geometry (a one-quote item is 50px: the 22.8px quote pill, the 19.2px comment row, the 2px gap between them, 2px padding and 1px border top and bottom; four items plus three 3px gaps), its own vertical scroll, `overflow-x: hidden` (the pane never scrolls sideways; an expanded label holding an unbroken token wraps with `overflow-wrap: anywhere`) and `overscroll-behavior: contain`. The head with the count and Send now stays outside the scroll; an expanded item grows inside it. The scroll offset is kept per tab under the sid that built the list (`strip.dataset.sid`), read once before the rebuild, so an expand or a discard keeps the place and a tab switch never carries one tab's offset onto another; an emptied list forgets it. Staging is the one caller that reveals the new item, since the new item is the event that justifies the move.

**The caret.** A button on the head (aria-expanded, aria-label, keyboard-reachable, focus kept across the rebuild) and the label collapse the strip to the head line; the count and Send now stay on it. The caret takes the chip ✕'s padding, so its hit target is that of the strip's other small control (about 13 by 15px) rather than the glyph's own 9px. The state is a page-lifetime Set written by the click and cleared when the stack empties, so the collapse belongs to the stack the user collapsed (the next stack starts shown, its first item in view) and a reload shows the list again. No new font sizes or colors: the caret wears `--dim` like the label. The tooltip on the count names the one-message release and its two exceptions; docs/guide.md gains a sentence on the list and says "as one message".

**Left out.** No kernel change. Persisted staged items keep their `{text, cites}` shape, so nothing staged before this lands is lost or migrated. The cap is in px at the pinned 12px chip text; a zoomed pane shows more or fewer than four, and the input pins hold the geometry so a change there fails a test rather than quietly showing three or five. The client's slash-command test mirrors the kernel regex by hand; a Python-side pin of the regex text would tie the two together.

## Tests

- `ui/webview/staged-messages.test.ts`: ten new executed cases over `quoteReplyBody`, `stagedRunBody`, `isSlashCommand` and `stagedPosts` (plain shapes, a typed command last and alone, a staged command alone at its place, a goal follow-up alone with the runs around it, an empty quote adding no section, and an invariant check over every shape: stage order across posts, each item exactly as often as it was staged, no shared body carries a command section). Before the change the module has none of the four exports, so the test bundle is refused with four "no matching export" errors.
- `ui/webview/staged-list-cap.test.ts` (new): `routeUserMessage`, `flushStaged` and `renderStagedStrip` lifted out of render.ts, transpiled at run time and executed over a fake DOM with the real `StagedStack` and the module's composition under them (the tab-strip-skip-exec.test.ts idiom). Six executed cases: the head first and every chip in the list, with the tooltip and no drag; the scroll offset kept across an expand, a discard and a switch, per tab, the reveal on stage and the forgetting on empty; the caret and the label collapsing the strip and back, per tab, the mark ending with the stack; focus kept on the rebuilt caret; the frames the kernel receives from a release (one `sendMessage` for the shared body, `askFollowUp` and a bare command at their place, a quoted command wrapped as prose); Send now alone, and held with a toast while the session is unreachable. All six fail against render.ts before the change (no `.staged-list`, no caret, a series of sends). Three pin cases cover what a fake DOM cannot see: the CSS cap with its geometry inputs and never on the strip or the head, the wrap rules and the caret's dress, and the composer closure's call sites (deliver's single `flushStaged`, the stage path as the one reveal, exactly three writers of the collapse Set, the one routing loop).
- `ui/webview/staged-list-layout.test.ts` (new): the cap measured in a real browser against the stylesheet, in the queued-romp-layout.test.ts shape, skipping loudly where no Playwright browser is installed (CI installs none): a one-quote item is 50px and a bare note about 25px, the list's client height is 209px with exactly four whole items in view of six, four items fill the cap with no scroll and three are 156px, the head sits above the list, a 400-character token in an expanded label wraps without sideways overflow, the caret's hit target is at least the chip ✕'s. Against the stylesheet before the change four of the five measured cases fail; one CI-safe pin checks that the page mirrors the renderer's markup.
- Re-aimed pins: `composer-citation.test.ts` (deliver's single `flushStaged` call, the wrap and the lead-in read from the module), `send-scroll-preserve.test.ts` (flushStaged's signature and loop), `rewind-edit.test.ts` (the edit branch precedes `flushStaged(sid, {`), `optimistic-send.test.ts` (image paths ride each post), `composer-send.test.ts` (the two open-label rules gain `overflow-wrap: anywhere`).
- Run: `npm run typecheck` clean; `npm test` 3687 passed, 0 failed, 0 skipped (the layout suite ran against a local Chromium; on a runner without one its five measured cases skip).

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -24,15 +24,16 @@ complementary views of what the agents are doing:
 **Reviewing a document.** Select any passage in the file viewer and it lands in the
 composer as a quote chip, labeled with the file and the line the passage lives on. Type
 what should change and press **⌘⏎** to set the note aside; keep reading, select the next
-passage, and repeat — each staged note remembers its quote and its place. **⏎** sends
-everything you staged along with whatever is in the box, so the session applies the lot
-in one pass, and you never copy a line out of the document by hand. The line in each
-label is checked against the file at the moment you select, so numbers that moved under
-you are caught rather than quietly carried. When several sessions work in the same
-repository, or in worktrees of it, the viewer's title bar says which one you opened the
-file from: a chip with the session's name, in the same color as its tab. The title bar's
-**GitHub ↗** button opens the file on GitHub. While the check runs, the button waits dimmed
-with pulsing dots beside it.
+passage, and repeat — each staged note remembers its quote and its place. The list above
+the composer shows about four staged notes and scrolls for the rest; its caret collapses
+it to the count. **⏎** sends everything you staged along with whatever is in the box as
+one message, so the session applies the lot in one pass, and you never copy a line out of
+the document by hand. The line in each label is checked against the file at the moment
+you select, so numbers that moved under you are caught rather than quietly carried. When
+several sessions work in the same repository, or in worktrees of it, the viewer's title
+bar says which one you opened the file from: a chip with the session's name, in the same
+color as its tab. The title bar's **GitHub ↗** button opens the file on GitHub. While the
+check runs, the button waits dimmed with pulsing dots beside it.
 When there is nothing to open, the button stays in place, dimmed, and a caption beside it
 says why (the file is not in a git repository or not committed — untracked, staged but in no
 commit, or on a branch with no commits yet — the repository has no origin remote, its origin
--- a/ui/webview/composer-citation.test.ts
+++ b/ui/webview/composer-citation.test.ts
@@ -10,6 +10,7 @@ import * as path from "node:path";
 
 const RENDER = fs.readFileSync(path.resolve(process.cwd(), "..", "ui", "webview", "render.ts"), "utf8");
 const FEED = fs.readFileSync(path.resolve(process.cwd(), "..", "ui", "webview", "feed.ts"), "utf8");
+const STAGED = fs.readFileSync(path.resolve(process.cwd(), "..", "ui", "webview", "staged-messages.ts"), "utf8");   // quoteReplyBody lives here
 const CSS = fs.readFileSync(path.resolve(process.cwd(), "..", "ui", "webview", "styles.css"), "utf8");
 const SKELETON = fs.readFileSync(path.resolve(process.cwd(), "src", "page-skeleton.ts"), "utf8");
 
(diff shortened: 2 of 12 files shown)


henrypinkard (approved):
Approved on head 629aa062 on the maintainer's word (feature-review session).

ksarma:
Note for the record: main's required vscode-extension check has been red since this merge (02d96006; first failing run 34415660201, still failing at 4a31744c in run 34416913654). Two node tests fail on the combined tree: the #1017 pin that every active-tab display path reads through `liveSession`, and this PR's own "Send now releases the stack alone" case, whose reachability refusal #1217 (merged five minutes earlier) routes through `ephemeralWarnToast`. Each PR was green on its own merge-ref. Every open PR's merge-ref now inherits the red check, so nothing auto-merges until main is fixed; a fix-tier PR is in preparation on the author's side.

Review run by the maintainer automated review session; verified against main 4a31744c.

ksarma:
Correction to the note above, with thanks to the author's session for the check: at 02d96006 only the staged-list-cap case ("Send now releases the stack alone") was red; the liveSession pin from #1017 turned red at 24680741, the #1119 merge (its updateReplyChips added the fifth `sessions.get(activeId)` read), not here. Both are fixed on main by #1226 (5b4623d2, 23:50Z).

Review run by the maintainer automated review session; verified against main 5b4623d2.

PR #1194, opened 2026-09-09
File viewer: a text-size control (A− / A+, Ctrl/Cmd + wheel) and a rendered-markdown layout that follows the viewer's width

ksarma:
The file viewer gets a text-size control (A−, A+, Ctrl/Cmd + wheel, and the percentage as the reset) that persists per browser and applies to every text view, and its rendered markdown follows the viewer's width instead of sitting in a fixed 860px column.

## Tier

Tier: feature

## What changes for a user

Every text file opened in the viewer (a markdown file's Rendered and Raw views, every other text file's code view, the SVG Source view, and a markdown document opened from a link on the dashboard's own address) can be read smaller or larger in fixed steps from 70% to 200%. A− and A+ in the title bar step it; Ctrl (or Cmd) and the wheel over the text do the same; once the size is off 100%, the percentage appears between the buttons and a click on it goes back. The choice is kept in the browser and applies to every file opened there. Prose keeps a readable line length that grows with the size, code blocks keep that width and wrap, a table is as wide as its columns need up to the viewer's width and scrolls horizontally on its own beyond that, a table inside a quote or a list item keeps the prose width, and pictures shrink to fit, so the page never widens. The title bar wraps instead of pushing the close button off the card on a narrow window or a phone.

## Design

- One property, a fixed table. The step rides the viewer root as `data-fv-text` (nine values, 70 to 200 percent) and both sheets map it to `--fv-scale`. Every text size in the viewer is an existing size multiplied by it (prose 1em, headings and inline code in em of the prose, fenced code and the Raw view's rows and gutter at 12px), so no new size enters the scale; the title bar, its buttons and the editor keep the page's size. A bundle without the rules reads every fallback as 1: today's sizes exactly.
- Progressive disclosure in the bar. A− and A+ only, until the size is off 100, when the percentage appears between them as the reset. Its slot is reserved from the start (visibility, not display), so neither button moves under the pointer when it fills, and the empty slot leaves the tab order.
- An end of the table is `aria-disabled` rather than `disabled`: dimmed and hover-inert through the shared bar-button dress, keyboard focus kept so the ring does not vanish on the press that reached the end, and its press is the same no-op as a step to the size in force. The dress that only the GitHub unit's no-link button wore is now every bar button's, so the editor's Save while a save is in flight (disabled, reading Saving…) is dimmed with a default cursor too.
- Click-safe and same-tick. The three buttons are built once per open and never rebuilt by a paint; each press writes the attribute, the readout and the end states synchronously. Nothing time-based: a step changes one attribute and the sheets reflow.
- Shown over text only. The control appears once the kernel's Content-Type says the body is text, never beside the loader, over a picture or a PDF, or in edit mode; each viewer's `renderBody` syncs it on every paint, the first before the fetch, so a picture opened over a slow link never shows the control and then takes it away.
- Ctrl/Cmd + wheel over the body steps the same table. Deltas are normalized across `deltaMode` and summed (`foldWheel`), so a trackpad pinch steps about once per notch and a reversal starts over; a plain wheel scrolls; the keyboard's Ctrl+plus/minus stays the browser's. The listener is non-passive so the page zoom is preempted only with the modifier held over a text body.
- Persistence like the Rendered/Raw choice: `romp:fileviewTextSize` in localStorage, per browser, one size for every file opened there; a foreign or malformed value reads as 100.
- The measure moves from the root to its direct children, written at zero specificity (`:where`) so a picture's own cap wins whatever the rule order (a plain `:not()` chain would tie `.fileview-md img` on specificity and leave the winner to the order of the two rules), and scales with the size (the same characters per line at every step). A table takes the viewer's width when its columns need it and scrolls horizontally on its own beyond that, whole words kept (`overflow-wrap: normal` against the root's `anywhere`); code blocks keep the measure and wrap; a bare top-level image, svg, canvas or video is capped at `min(100%, the measure)` (the class-specificity measure would otherwise outrank the sheet's element-level media cap for a direct child); a table nested in a quote or a list item keeps the prose width and scrolls inside itself when its columns need more.
- The title bar wraps: the path keeps 12em and the action row drops to a right-aligned second line. Three more buttons pushed the close button off an `overflow: hidden` card below about 600px, and a phone has no Escape. The wrap is scoped to `.fileview-bar`, so the file browser's action row, which wears `.fileview-acts` outside a bar, keeps one line.
- A press on a bar control settles no selection. The mouseup seed returns early when the lift lands on a button or anchor inside the bar, so a passage still selected in the body does not re-fetch the file and re-seed the quote chip on every step; a drag released over the bar's path or padding settles as before.
- The URL-mode viewer (`openUrlView`) mounts the same control through one builder (`textSizeControl`) and carries the same root attribute, so a document opened from a link on the dashboard's own address honours the same size as a local file. This is deliberate: the two viewers already share the Rendered/Raw preference, and a size that applied to one and not the other would read as a bug.
- Shared chrome stays byte-equal across `styles.css` and `feed.css`; `fileview-parity.test.ts` gains the new heads.

Considered and left out:

- A free zoom multiplier. A fixed table keeps the buttons, the wheel and the stored value on the same few sizes and gives the wheel ends to clamp to.
- A keyboard chord (Ctrl+plus/minus). It is the browser's page zoom and stays so.
- Any observer or repaint pass after a step or a resize. Nothing here measures the body; the sheets are fluid on their own.
- A per-file size. One size per browser, like the format preference.

## Tests

- `ui/webview/file-view-text-size.test.ts` (new), against a self-contained DOM stand-in (a tree, attributes and `dataset`, bubbling events with their registration options, a small selector matcher, localStorage, a fetch answering the kernel's Content-Type and mtime headers, a real `Response` for the URL viewer):
  - The pure table: `TEXT_SIZES` is bounded and ascending with 100 present; `stepTextSize` clamps at both ends and steps from the default off the table; `parseTextSize` returns a table step and reads absent, garbage, off-table and multiplier values as 100; `foldWheel` steps once per notch, sums a pinch's small deltas, starts over on a reversal, normalizes lines and pages, and reads wheel-up as larger. Before the change: the exports do not exist.
  - The control over the real `openFileView`: a markdown file opens at 100 with the readout slot empty and the control after Rendered/Raw and before Edit; each press steps, stores under `romp:fileviewTextSize`, and sets `data-fv-text` and the readout in the same tick; the ends set `aria-disabled` (never `disabled`) and a press there stores nothing; reset returns to 100 and empties the slot; the size survives a close and a reopen and is on the root before the bytes land; a foreign value opens at 100; the control is hidden beside the loader, shown when text lands, hidden for a picture and a PDF, present for the SVG Source view, hidden in edit mode and back on exit; Ctrl/Cmd + wheel over the body steps and prevents the default, a plain wheel does not, a wheel over the bar does not, a picture leaves the browser its zoom; the body's one wheel listener is registered non-passive; the three buttons are the same nodes across a Rendered/Raw flip and a step; with a passage selected, a mouseup and click on A+ steps the size without re-reading the file or posting `editorSelection`, while a lift on the bar's path or padding still settles. Before the change: every case fails at "the readout between A− and A+" (no control in the bar).
  - The URL viewer: `openUrlView` carries the stored step on its root before the document lands, mounts the control after Rendered/Raw and before Open, steps and stores under the same key with the same non-passive wheel listener, and a local file opened next honours that size. Before the change: fails the same way.
  - Both sheets: the nine `data-fv-text` rules map every `TEXT_SIZES` entry to `--fv-scale` and nothing else sets it; exactly `.fileview-md`, the `:where` measure, `.fileview-md > :is(img, svg, canvas, video)`, `.fileview-md pre code`, `.fileview-pre` and `.fileview-gutter` read it; the root has no max-width; the table rule; the top-level image cap follows the general one; the bar wraps and the two scoped rules exist while the bare `.fileview-acts` stays one row and `.fb-bar` has no wrap; the shared disabled dress; the readout slot; all byte-equal across the sheets. Before the change: fail on the missing rules and the 860px root.
  - A browser leg (headless Chromium through a child driver that exits 3 without playwright or a browser, so CI skips it): the page never widens at 1000 and 420px at 100% and 150% in both sheets; prose and code keep the 860px measure and the top-level table takes the viewer at 1000px while the same table inside a quote and inside a list item stays within the prose width, scrolling inside itself at 150% and at 420px; at 150% the prose, heading and code sizes and the measure grow together; from 380 to 1400px in both modals the close button stays on the card, the path keeps at least 12em, the action row ends at the bar's right edge, up to 600px the row is the line below the path and at 1400 the two share a line, and A− and A+ hold still when the readout fills; the file browser's bar stays one line from 320 to 1000px. Before the change: the root is 860px wide against a 948px body, and at 380px the close button sits at x 787 on a card that ends at 370.
- `fileview-parity.test.ts`: the two scoped bar rules, the renamed disabled heads and the two readout heads join the byte-equal list.
- `github-link.test.ts`: the disabled-dress regexes follow the shared selectors, and the `aria-disabled` exclusion is narrowed to the GitHub unit and anchors.
- `fileview-chip.test.ts`: the stand-in's element gains `dataset`.
- `file-view.test.ts`: the mouseup pin follows the handler's `(ev)` parameter.

`npm run typecheck` and `npm test` (3762 passing, 0 skipped, the browser leg included) in `vscode-extension`.

Rebased onto main at 3f06288e (2026-09-09). Four hunks met the sanitizer change that merged in between: the guide paragraph (both kept, the sanitizer's first), the byte-equal head list (the new heads appended), the root rule (`contain: layout` kept beside the scale and the child measure) and the media caps (the top-level cap widened to `:is(img, svg, canvas, video)`, since a multi-line svg block is a direct child of the root and the element-level cap loses to the measure's class specificity; a browser case with a 1600px svg pins it). The table rule matches the sanitizer's, plus `overflow-wrap: normal`. Full `npm test` on the rebased head: 3762 passing, 0 failed.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -73,6 +73,19 @@ lands on it. An image map (`<map>`, `usemap`) is dropped. The same rules apply t
 a chat message, where a link to an element's own `id` or `<a name>` lands on it under the
 prefix.
 
+**Text size and width.** The **A−** and **A+** buttons in the viewer's title bar make the
+text of any text file smaller or larger in fixed steps from 70% to 200%: a markdown file's
+Rendered and Raw views, the code view of every other text file, and a document opened from
+a link on the dashboard's own address. They appear wherever the viewer opens (over the chat
+or the feed), and not for a picture or a PDF, which have no text to size. Ctrl (or Cmd) and
+the mouse wheel over the text do the same. Once the size is off 100%, the percentage appears
+between the buttons; click it to go back. The choice is kept in this browser and applies to
+every file you open here. Prose keeps a readable line length that grows with the text size,
+and code blocks keep that width and wrap long lines. A table is as wide as its columns need,
+up to the width of the viewer, and scrolls sideways on its own beyond that; a table inside a
+quote or a list item stays within the prose width. Pictures shrink to fit, so the page is
+never wider than the viewer.
+
 **Opening a PDF.** A PDF the session mentions, or one you click in the file browser, opens
 inside the dashboard like an image: the chat's PDF card opens it full-view, a path or a
 file-browser row opens it in the file viewer. Cmd-click it instead (Ctrl on Windows and
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -249,6 +249,18 @@ ghostty --working-directory={dir}   # Linux: Ghostty
 code {dir}                          # VS Code instead
 ```
 
+### The file viewer's per-browser choices
+
+The file viewer keeps two choices in the browser's own storage, not on the
+kernel, so they survive a kernel restart and apply wherever the viewer opens
+(over the chat or the feed, and for a document opened from a link on the
+dashboard's own address): the Rendered or Raw view of a markdown file
+(`romp:fileviewFmt`) and the text size (`romp:fileviewTextSize`, one of 70,
+80, 90, 100, 115, 130, 150, 175 or 200 percent, set by the **A−** / **A+**
+buttons or Ctrl/Cmd + wheel over the text). The size scales the prose, its
+headings, the code and the Raw view together, and the prose measure with
+them; a value outside the table reads as 100.
+
 ### Model and effort, from the statusline or a typed command
 
 Typing `/model X` or `/effort X` into the chat composer, or sending one with
--- a/ui/webview/feed.css
+++ b/ui/webview/feed.css
@@ -1393,10 +1393,14 @@ body { display: flex; flex-direction: column; position: relative; }   /* column 
   background: var(--bg, #1e1e1e); border: 1px solid var(--box-border); border-radius: var(--radius-modal);
   box-shadow: var(--shadow-modal); }
 body.fileview-open { overflow: hidden; }
-.fileview-bar { flex: 0 0 auto; display: flex; align-items: center; gap: 10px; min-width: 0;
+.fileview-bar { flex: 0 0 auto; display: flex; flex-wrap: wrap; align-items: center; gap: 6px 10px; min-width: 0;
   padding: 7px 10px; border-bottom: 1px solid var(--box-border); }
 /* Only the DIRECTORY may be truncated: the filename identifies the file, so it holds its width however
-   deep the path is, and the dimmed directory in front of it gives up room first. */
+   deep the path is, and the dimmed directory in front of it gives up room first. The bar WRAPS: the path
+   keeps at least 12em and the action row drops to a second line, right-aligned, when the two do not fit
+   side by side (the text-size control's three buttons pushed Download, Copy path and the close button off
+   the card below about 600px; .fileview is overflow: hidden, and a phone has no Escape). The wrap is the
+   bar's own: the two .fileview-bar rules after .fileview-acts below. */
 .fileview-name { flex: 1 1 auto; min-width: 0; font-size: 0.86em; color: var(--fg);
   display: flex; align-items: baseline; white-space: nowrap; }
 .fileview-dir { flex: 0 1 auto; min-width: 0; color: var(--dim);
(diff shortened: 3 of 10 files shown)


henrypinkard (approved):
Approved on head 62b5f6e7 on the maintainer's word (feature-review session).

PR #1197, opened 2026-09-09
Bottom bar: an API health cell beside the usage readout, from the sessions' API errors, retries and the retry pause

ksarma:
An API health cell in the dashboard's bottom bar: one dot and one word beside the usage readout that say whether the API is serving your sessions (ok, how many are waiting and on what, or paused and why), with the reading on hover and a click detail that lists the waiting sessions and carries the Stop / Resume auto-retry button.

## Tier

Tier: feature

## Design

- **What it reads.** The cell reads what the kernel already owns: each alive session's latched newest API-error record, the live retrying state from the api_retry frames, and the retry-pause file. It does not read the `/api-health` ratio signal: that signal's states are threshold-and-hold transitions evaluated at read time, while the cell answers "is a session waiting on the API right now".
- **Moves on events, never on a clock.** Every field of the frame moves on one named event: a retry frame, an `isApiErrorMessage` record, fresh assistant output, a pause set or lifted, a session leaving the roster. Two computes over an unchanged world serialize identically and send nothing; rows sort by (since, sid) so roster order cannot cause a send. Per-attempt counters (attempt, retryAt) are not inputs.
- **The latch.** `_api_last_failed` is `_api_error`'s latched sibling: a user prompt does not clear it. romp's own auto-retry IS a user prompt (`RETRY_MSG`), so a count keyed on `_api_error` flapped "1 waiting / ok / 1 waiting" on every backoff rung. One tail pass (`_api_error_pass`) serves both readers, so the pusher's second question about a transcript in the same cycle is a cache hit. The pass keeps the original scan's loop and comments as written, changed only at the sites that assign a verdict. `_api_error_scan` stays as a shim for the existing differential tests, whose verdicts are unchanged.
- **What counts.** On-you failures (too long, model limit, dead credential, refusal) do not count: they already wear red on the session's own tab and card, and the cell answers for the API. A spend cap counts, and engages the spend pause in the same cycle.
- **Each pause lifts on the event that engaged it.** A limit pause lifts when the usage report stops naming an account-wide window (`_account_limited`, the engage's own filter). The mtime rule could flip the file every cycle on a key-billed session's output, and a rule keyed on a login session's output could never fire after the reset, since the blocked login sessions are the ones the pause gates. A spend pause records the capped session's billing and lifts on fresh assistant output from that billing only. A manual pause keeps the mtime rule.
- **The spend stand-down.** A spend cap is on-you, so the capped session's record outlives the lift. The un-pause records `liftedAt` (the lifting record's own time, not the cycle's clock) and the spend engage skips older records, so the pause no longer re-engages the cycle after every lift. `_auto_pause_on_limit` latches `reason="limit"` at the event so the cell names the pause without a clock comparison.
- **The billing of a row with no auth on record.** `_bills_login` reads the CLI's `authLive`, then the registry's `auth`; a row with neither (a tmux session, an SDK session before its init landed) bills the login only when no apiKeyHelper is configured on the machine (`_auth_key_present`), and is taken as key-billed otherwise. A spend pause engaged from such a row on a login-plus-helper machine therefore records `bills=key`.
- **`seq`.** It counts pause-file writes, so the frame that answers a press on the detail's pause button differs from every earlier frame whatever state it brings; the button's acknowledgment clears on the truth (paused again included), never on a matching-state guess.
- **Delivery.** Unchanged: `_set_retry_paused` marks the views dirty and the pusher's next cycle carries the flip. The frame is built after the cycle's pause decisions and sent only when it changed; a shell that sends `ready` gets the last frame verbatim, and the client diffs state and text before touching the DOM so a re-send cannot pulse the cell.
- **UI.** The word is `.ru-pct`'s declaration byte for byte (no new font size). The dot wears status hexes (the label gray, the tab-retrying amber, the API-error red), never the accent, which marks only the footer's links. The cell ships hidden and shows on its first frame, so a kernel that sends none shows nothing rather than a false ok. Progressive disclosure: one word on the rail, the reading on hover, actions only in the pinned detail.
- **Click-safe.** A frame that lands while a primary pointer is down over the detail is painted on release. A press is acknowledged before the round trip and stays disabled until the answering frame; a failed send restores the button and says why; a socket close clears a press that can no longer be answered and says so. The detail is an aria-modal dialog with a focus trap; focus moves to the card before the pressed button is disabled, since a disabled element cannot hold focus.
- **One modal at a time.** The detail shares `#ru-back` with the usage modal: opening either closes the other first, and Escape closes the API detail before the usage modal since both ride that backdrop.
- **Deliberately left out.** The VS Code strip twin (`strip.ts` apiCell's sibling) and the phone's usage-modal section (no rail on the phone) are follow-ups. A hover history of the cell's readings is a separate change. No palette command on the cell yet.
- **Where the pins live.** The shell page has no jsdom harness, so its source pins sit in the Python modules beside the behavior they approximate (`tests/test_api_health_rail.py`, `tests/test_kernel_pane_rail.py`); the one TypeScript addition pins `render.ts` only.

## Tests

Executing tests, and what each fails on before the change:

- `tests/test_api_health_rail.py` (new, 58): the frame's states, plurality and tie order, the on-you exclusion, the pause outranking, the no-flap rule (a byte-identical recompute sends once; attempt/retryAt changes move nothing; seq moves only on pause writes; a ready re-send reaches shells only, with recording sinks on every client since the send is guarded), the frame's exact keys and its agreement with `sdk_backend.api_health_status_class`, the wiring (the jobs block run with its other jobs stubbed builds the frame after the pause decisions, plus the source pin; the ready handler; `reason="limit"`), the docs subsection, and the detail's source. Before: `_api_health_frame`, `_api_health_push`, `_apih_resend`, `_LANDING_APIH_JS` and the reference subsection do not exist, and `_set_retry_paused` takes no `bills` or `lifted_at`.
- `tests/test_api_health_browser.py` (new, 21): a scratch copy of the shell page driven by playwright's chromium over a WebSocket shim: hidden before the first frame, keyboard button, hover re-anchor, one modal on the shared backdrop, a held press across a frame, the acknowledgment held until the answering seq, a failed send, a row and a pause press through the page's own send while no socket is open, a socket loss and redial, Tab cycling, primary-button-only defer, the light theme's dot colors. Before: the served page has no `window.__rompApiHealth`, so the driver times out. Skips without the extension's node deps or a browser (CI installs none); it ran here, and it removes its scratch directory.
- `tests/test_api_error_tail.py` `ApiLastFailedLatch` (new, 16): the latch holds through a user prompt and through `RETRY_MSG`, clears on assistant output, carries `t`, takes a refusal record's mark, shares one read with `_api_error`, widens past a tail of prompts, keeps its own cache. Before: `_api_last_failed`, `_api_last_output_t` and `_api_error_pass` do not exist, and the error dict carries no `t`. The existing differential tests keep passing through the `_api_error_scan` shim (their recorder now wraps `_api_error_pass`).
- `tests/test_retry_pause_autoresume.py` `LimitPauseLift`, `SpendPauseLift`, `SpendPauseStandDown`, `PauseWriteSeq` and two additions to the existing class (new, 25; the two spend classes share a fixture base, so each test is collected once): a limit pause lifts only when the report stops naming a window; a spend pause lifts on fresh assistant output from the recorded billing only, never the mtime, never the other account; `liftedAt` is the record's time; a stale spend record does not re-engage; seq per write. Before: a limit pause lifts on any transcript's mtime, a spend pause lifts on the other billing's output and re-engages the next cycle, the file records no `bills` or `liftedAt`. The existing `RetryPauseAutoResume` tests (the writer's dirty mark, no inline push) still pass.
- `tests/test_kernel_pane_rail.py` `ApiHealthCell` (new, 11): the cell's placement after `#rail-usage` inside `.rail-scroll`, the hidden markup and its author `[hidden]` rule, `.ah-text` equal to `.ru-pct` byte for byte, status hexes never the accent, the light theme's ok scoping, `#rail-usage:empty`, the shared `#ru-tip` skin, the shell-socket routing and the Escape order. Ten are red before: none of the markup, rules or routing exists. The eleventh, the shell socket's wid dial minted before it connects, passes before and after: it pins the existing dial the detail's row relies on. `ShellSocketIdentity` (3) passes before and after too: it pins the wid-scoped reveal.
- `tests/test_kernel_usage_limit.py` (2): the pause latches `reason=limit` at the event with the countdown unchanged (red before: the pause records no reason); an already-paused file keeps its reason (passes before and after: it pins the engage's existing idempotence).
- Pin adjustments that follow the change, red before it because they name what it adds: `tests/test_kernel_apierror_working.py` and `tests/test_session_auth.py` read the tail pass (`_api_error_pass`), `tests/test_kernel_model_limit.py` reads the shared filter (`_account_limited`), `tests/test_kernel_mobile.py` counts 20 scripts, and `tests/test_perf_stats.py`'s pusher job list gains `_api_health_frame` and `_api_health_push`. `ui/webview/apierror-spend-limit.test.ts` passes before and after by construction: it pins `render.ts` only (`retryPausedText` checks spend first and renders a limit reason as the countdown).

Full runs on the final tree: `pytest -n 8 tests/` 10311 passed, 40 skipped, 0 failed; `npm test` 3655 pass, 0 fail; `npm run typecheck` clean.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -624,3 +624,14 @@ themselves. The analytics modal in settings shows what you actually spent,
 separating your sessions from the judge pipeline. You can also reconfigure the
 judges from the gear: the high-volume indexing tier defaults to Haiku, and the
 judgment tier defaults to Sonnet.
+
+The bottom bar's **API** cell, a dot and a word, shows how the API is treating
+your sessions. Gray **ok** means no session is waiting on the API. Amber shows
+how many sessions are waiting and names the problem: **rate limited**,
+**overloaded**, **offline** (this machine cannot reach the API), or **errors**.
+Red **paused** means auto-retry and the judges are stopped, and says why: a
+usage limit, the monthly spend cap, or that you stopped them. Hover for the same
+reading with the waiting sessions listed. Click the cell, or press Enter on it,
+for the detail: each waiting session (click one to open that session), a button
+that stops auto-retry for every session while sessions are waiting and resumes
+it while paused, and links to the usage figures and the Log.
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1374,6 +1374,62 @@ after it. The pre-restart state is not carried over: an empty ring is no
 evidence. A state file, or an entry in it, that cannot be read is skipped and
 logged, and never keeps the SDK backend from starting.
 
+### The bottom bar's indicator
+
+The dashboard's bottom bar carries an API cell (a dot and a word beside the
+usage readout) that is computed independently of this signal, from two things
+the kernel owns directly:
+
+- Each alive session's newest transcript API-error record, latched until the
+  session produces assistant output again (a user prompt does not clear it,
+  romp's own retry included), plus the live retrying state of SDK sessions.
+- The retry-pause file (`retry-paused.json` under the state directory). A
+  pause writes `paused`, `t` (when it began, the auto-resume floor) and its
+  `reason`: `limit`, `spend`, or none for a manual stop. A spend pause adds
+  `bills`, the billing the capped session was on (`login` or `key`); only
+  fresh assistant output from a session on that billing lifts it. Un-pausing
+  a spend pause, by that lift or by the Resume button, records `liftedAt`
+  (the time of the output record that lifted it, or of the Resume click) and
+  `supersedes` (the floor of the pause it cleared; informational, nothing
+  reads it); both ride every later write until a newer spend un-pause
+  replaces them, and a spend-limit record older than `liftedAt` engages
+  nothing, since the lift already ruled on it. A limit or manual un-pause
+  records neither. A limit pause lifts when the usage report stops naming an
+  account-wide window at 100%, a manual pause when any live session not
+  blocked on an API error writes to its transcript after the pause began.
+  When a limit pause lifts while a spend-limit record is standing, the file
+  reads unpaused for one cycle before the spend pause engages: each writer
+  rules on one signal per cycle, and the spend engage runs before the lift in
+  the pusher's order, so it sees a paused file and rules on the record the
+  next cycle.
+
+The kernel pushes the cell's frame to shell clients only when it changed, and
+again to a shell that sends `ready`:
+
+```json
+{"type": "apiHealth", "state": "ok | degraded | paused",
+ "cls": "429 | 529 | offline | errors | ''", "reason": "'' | limit | spend | manual",
+ "text": "<the rail's words>", "waiting": 0, "retrying": 0, "blocked": 0,
+ "since": 0, "tmux": 0, "seq": 0,
+ "sessions": [{"sid": "", "name": "", "color": null, "kind": "retrying | blocked",
+               "cls": "", "status": null, "since": 0, "suppressed": false}]}
+```
+
+`seq` counts the retry-pause file's writes since the kernel started. A press
+on the detail's pause button writes that file, so the frame that answers the
+press carries a moved `seq` whatever state it brings, and the shell clears
+the button's acknowledgment on it; a frame from before the press carries the
+old one. It is an event counter, not a clock, and restarts at 0 with the
+kernel. `waiting` is `retrying` plus `blocked`. `cls` is the plurality class
+over the affected sessions, ties resolved 429, then 529, then offline, then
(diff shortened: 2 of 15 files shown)


henrypinkard (approved):
Approved on head b3f8e363 on the maintainer's word (feature-review session).

PR #1198, opened 2026-09-09
Sessions & tags: one colour dot per tag row, folded pane filters, and a height budget that keeps the sessions on screen

ksarma:
The Sessions & tags dialog draws one colour dot per tag row (the palette opens as a popover), folds the pane filters to one summary line per pane, and keeps the sessions table on screen with a height budget, so a dialog with dozens of tags still shows the sessions the user came to act on.

## Tier

Tier: feature

## Why

With ten tags the dialog filled the page. Every tag row carried the whole identity palette inline (twelve swatches in two rows, about 38px for one bit of information), and the pane filters were a five-pane matrix of every tag, so the sessions table showed two rows below the visible area. The sessions are the working area; the sections above them should take one line per item and open their detail on demand.

## What changes

The change is two commits: the code and its tests in the first (`ui/romp-timeline-view.js`, inside the dialog and the two menu helpers it uses; nothing else in the file), the `CLAUDE.md` rule alone in the second.

- **A tag row is one line**: the pill (still the reorder handle), delete, rename, one colour dot wearing the tag's colour. Click, Enter or Space on the dot opens the identity palette as a popover (the same swatches in the same balanced split, the current one ringed and focused). A pick performs the same recolor write the inline swatches performed (`_editTagUnion` with `color`), re-resolving the tag by union key at pick time so a repaint between open and pick (an ack, a refusal, a rename, a remote half landing) cannot act on a stale object; the optimistic copy, the ack and the refusal paths are unchanged. A held dot (a remote half with no bridge) wears the reason and takes no click, as the held actions do.
- **Popover placement and lifetime**: it lives in the dot's own document, one z-index above the dialog's backdrop, placed by the dot's rect (below, or above when below cannot hold it). It closes on the pick, on Escape (the dialog closes on the next press), on a press anywhere else on the dialog (the anchor dot exempt, so its click toggles rather than close-and-reopen), when Tab or focus leaves it, when its dot moves (the table's scroll, the card's scroll, a window resize), and with the dialog. One popover at a time. A repaint re-places it under the rebuilt dot, or closes it when the row is gone, scrolled out of the table's box, or held now (a remote half landed with no bridge to reach it: a pick could only be refused, and left open the popover hung on the detached dot of the last build, whose rect is all zeros).
- **Two menu-helper changes** the popover needs: `_menuHost(anchorRect, anchorEl)` returns the anchor's own document untranslated when the anchor already lives outside `document` (the dialog is adopted into the host document when it opens; translating an adopted node's rect a second time put the popover an iframe offset from its dot, below the viewport in a 200px band). The optional parameter changes nothing for the existing menus, which pass no element. `menuTop` flips above only when the menu also ends on screen, else clamps; before, an anchor below the viewport passed the `above >= 6` check and the menu landed off the bottom.
- **Pane filters fold** to one summary line per pane (`lensSummary`): All, no tags, or the tag names in the user's tag order, cut with a count past six, naming only tags that still exist (a lens naming only gone tags reads "none of the tags here"; the hover lists everything plus the gone names). The caption's caret opens the full matrix in its own cell, bounded to 25vh and scrolling within itself, with the chips wrapping inside their own cell so a second line starts under the first chip. The open or folded choice holds for the page lifetime as a module variable (the same idiom as `modelChoicesRev`), deliberately not a persisted setting.
- **The height budget**: the tag table caps at 30vh and scrolls within itself. The sessions box carries `flex:1 1000 auto` so it gives way first when height runs out (flex distributes shrink by base size, so a plain 1:1 squeezed a tag table that fit), down to a floor of min(live, 4) rows; at that floor the table gives way to three rows and the open matrix to 52px; past those the card scrolls as a whole (`overflow-y:auto`) instead of clipping, so a phone held sideways can still reach the search box and "tag all". Both floors are measured off the rendered rows at the end of each build rather than set as constants: the row height is the font's, and a 22px-a-row constant left 6px of blank under one tag and cut the third of three rows. The sessions floor is min(live, 4) times the smallest rendered row (a first row whose chips wrap would otherwise leave blank), by the live count so typing in the search box never moves it; both heights persist across builds so a build under a query, or in a document without layout, keeps the last floor; the card is appended to its final document before the first build so the first measurement happens where it is laid out.
- **[+ New tag]** is a card child under the table, outside its scroll (with ten tags at 800px the cap held exactly the ten rows and hid it as the table's last row). The table's scrollTop is read before a repaint and restored once the card is complete.
- **The reorder drag over a table that scrolls**: only rows whose whole box plus the 2px cue lies inside the table's box take the cue and the drop (the cue is a border on the row's edge; an edge under the clip means an invisible cue); the held row counts while any of it shows; the table's scroll re-ranks with the last pointer y for the drag's duration (a wheel mid-drag fires no pointermove) and the drop re-ranks if the table scrolled since the grab; the cue comes off for each measurement and goes back in the same task (it grows the pill cell 2px and shifts the rows under it); the table carries `overflow-anchor:none` so scroll anchoring does not follow that 2px and oscillate.

The second commit adds a short rule to `CLAUDE.md` under the UI section: tag and session surfaces are laid out for dozens of items, per-item controls take one line, per-item detail opens on demand, the working area keeps the space, and a change to such a surface is checked against the thirty-tag fixture. It is its own commit so it can be reworded or dropped without touching the code.

## Measured

At every height measured the search box, "tag all" and [+ New tag] stay on screen and the card itself never scrolls. Chromium and Firefox at 1600 wide with thirty tags and forty sessions (tag rows shown / session rows shown without scrolling the dialog): 1300px tall 14 / 18 folded and 14 / 12 with the matrix open; 800px 8 / 7 folded (Firefox 6) and 8 / 4 open; 700px 7 / 4 folded and open; ten tags at 800px 8 / 7 (Firefox 6). 30vh was chosen over 35vh from the same measurement: it trades two tag rows for two session rows at 1300px and 800px, at the cost that ten tags scroll at 800px (8 of 10 shown).

## Not included

No change to `render.ts`, `styles.css`, `settings.ts` or `gear.js`. No persisted setting for the filters' fold. No docs change: nothing under `docs/` describes the dialog's layout. No test measures WebKit.

## Tests

- `ui/timeline-tags-scale.test.ts` (new): twenty-two cases executed over the fake-DOM shim with the real `TimelinePanel`, no source pins. `lensSummary`'s ordering, cut and gone-name handling; one dot per row and no inline swatch; the popover's open state, current swatch, the pick posting the same recolor write (op, tid, color, writeId) and its ack and refusal paths; the pick re-resolving after a rename and a remote half; a held dot taking no click; keyboard open, arrows, Tab, focusout, Escape order, presses on the card, the dot and the backdrop; one popover at a time; the scroll and resize closers and the repaint re-place; the table's scroll memory; the drag's candidate rule over a cue-aware rect model with a wheel mid-drag; folded and open filters and the page memory; thirty tags and forty sessions rendering. Two cases execute what only the browser legs reached before: a repaint that turns the popover's dot held closes the popover; and in a framed shell the popover lands in the dialog's own document, placed by the dot's rect untranslated, with its resize closer on that window (the shim's nodes carry an `ownerDocument`, adopted with the subtree on append as a browser adopts them, and the frame stand-ins are set so the frame walk would translate). The floors are measured over index-placed rows, so a floor read off the whole table fails, and two live sessions beside four dead ones give a two-row floor, so a floor by every session fails. The shim is reset before every case (the body emptied; the window's listeners, the feed row's echo, the geometry hooks and the recorded posts cleared; the filters' page-lifetime fold put back), so one failure cannot take the cases after it down. Before the change all twenty-two fail (`lensSummary is not a function`, no dot, no popover, the old styles).
- `ui/timeline-tags-scale-browser.test.ts` (new): three legs per browser in Chromium and Firefox, measuring the layout at 1300, 800 and 700px tall with thirty and ten tags, real pointer drags past the table's edge with a wheel mid-drag and a cued row scrolled under the clip, the floors at 420 and 500px (none, one, two of six, three, four and forty live) and with wrapped chips at a 390px page, a phone held sideways, the popover's ring and focus, Tab, scroll, resize, scroll memory and repaint, and two framed shells (an offset iframe and a 200px band at the page's foot). The card's and the matrix cell's overflow are read as rendered (`getComputedStyle`), not inferred from a scrollTop write, which a clipping box also takes; with the matrix open at 800px the cell keeps to 25vh and `elementFromPoint` finds every chip inside the cell's box and none past it, so a cell without its bound and scroll, whose chips print over the sessions table, fails here (63 of 64 chips past the box painted under that mutation). The sessions box and the matrix cell are found by structure (the sessions grid's parent, the "All surfaces" row's parent), and the flex declarations are asserted on the found boxes, so a change to a declaration is measured rather than crashing the leg. Waits are on conditions (the popover open with its current swatch focused; the popover closed after a resize), never fixed delays. Skips loudly per leg without a playwright browser; these ran locally (`npx playwright install chromium firefox` under `vscode-extension`, then `npm test`): 6 of 6 pass. Before the change all six fail (no sessions box floor to read; the ringed pill sits at 0px inside the table).
- Re-aimed pins: `ui/timeline-views-panel.test.ts` (the card's `overflow-y:auto`, the chips' wrapping cell, the sessions box's `gridStyle` floor and `min(liveN, 4)`, the recolor through `now`, the held dot, New tag under the table), `ui/timeline-menu-place.test.ts` (a new executed case: an anchor below the viewport clamps on screen, 1232 for the 1300px case where the old code returns 1280; the `_menuHost(anchorRect, anchorEl)` pin and its own-document branch), `ui/webview/menu-echo.test.ts` (counts the popover among the `MENU_STYLE` menus and pins its `data-romp-menu` mark), `ui/webview/tab-color-picker.test.ts` (two source pins on the inline swatch split replaced by an executed half: the dialog opened, the dot clicked, the popover grid's columns read for 12, 9, 13, 7, 6 and 5 swatches), `ui/timeline-kernel-post.test.ts` (opens the folded filters through the caption before clicking a chip; red before because the caption had no click listener). `ui/timeline-tagorder-drag.test.ts` seats its cells inside a table rect so the whole-row rule admits them; it passes before and after.
- Mutations checked red: the sessions floor counting every session (the fake-DOM floors case), the `_menuHost` own-document branch removed and the popover left in the pane's body (the fake-DOM framed case), the held-dot close reverted (the fake-DOM held case), the card `overflow:hidden`, the matrix cell without its bound and scroll, and the sessions box at `flex:1 1 auto` (the browser legs, by measurement).
- Full run under `vscode-extension`: `npm run typecheck` clean; `npm test` 3699 tests, 3699 pass, 0 fail, 0 skipped (browser legs included).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -407,3 +407,17 @@ times" bug. Every interactive control MUST therefore:
 Reuse `ui/webview/actions.ts` for any new dashboard control. (`.romp-acted` is
 defined in both `styles.css` and `feed.css` since the feed page loads only the
 latter.)
+
+### Designs must accommodate many tags and many sessions (user rule, 2026-09-09)
+Assume dozens of tags and dozens of sessions on every surface that lists them, and lay
+the surface out for that case, not for three. Per-item controls take ONE line: a tag row
+is its pill, its actions and one colour dot; a session row is its name, its [+] and its
+chips. Palettes, pane-by-tag matrices and any other per-item detail open on demand (a
+popover, a fold), never inline for every item at once. The working area keeps the space:
+the sections above it fold or cap their height and scroll within themselves, so the part
+the user came to act on (the sessions table in Sessions & tags) is never pushed under
+the fold. Triggered by that dialog with ten tags, where twelve inline swatches per row
+and a five-pane matrix of every tag filled the page and left two session rows showing.
+A change to a tag or session surface is checked against a fixture of thirty tags (the
+dialog's is `ui/timeline-tags-scale.test.ts`, with the measured layout in
+`ui/timeline-tags-scale-browser.test.ts`).
--- a/ui/romp-timeline-view.js
+++ b/ui/romp-timeline-view.js
@@ -283,6 +283,38 @@ function lensLabel(l) {
   if (l.none) parts.push('no tags');
   return parts.join(' + ') || 'All';
 }
+// The folded "pane filters" line for one pane (the user 2026-09-09, whose ten tags made the open
+// matrix fill the dialog): All, no tags, or the tag names the lens shows in the USER'S tag order
+// (`order`, the union's, not the order the chips were clicked in), 'no tags' last as lensLabel puts
+// it, cut with a count past `max` names (six by default; Infinity for the full list a hover shows).
+// Only names the union lists (`order`, when given) count: a rename or a delete leaves the old name in
+// the lens (the kernel keeps unknown names by design), and a name the open matrix draws no chip for is
+// not one the folded line should claim either (review find, 2026-09-09). A lens left with none says so.
+function lensSummary(l, order, max) {
+  if (lensAll(l)) return 'All';
+  const idx = new Map((order || []).map((n, i) => [n, i]));
+  const at = (n) => (idx.has(n) ? idx.get(n) : idx.size);
+  const parts = (l.tags || []).filter((n) => !order || idx.has(n)).sort((a, b) => at(a) - at(b));
+  if (l.none) parts.push('no tags');
+  if (!parts.length) return 'none of the tags here';
+  const cap = max || 6;
+  if (parts.length <= cap) return parts.join(', ');
+  return parts.slice(0, cap).join(', ') + ' +' + (parts.length - cap) + ' more';
+}
+// The Sessions & tags dialog's "pane filters" fold: folded by default, and the choice holds for the
+// page lifetime (a module variable, the way modelChoicesRev does), not per open, so a user who opened
+// the matrix finds it open on the next look and one who left it folded is not shown it again.
+let paneFiltersOpen = false;
+// The tag table's share of the viewport in the Sessions & tags dialog: its natural height up to this,
+// scrolling within itself past it (the height budget is described where the table is built). 30vh over
+// 35vh (measured in Chromium at 1600 wide, thirty tags, forty sessions, filters folded; tag rows shown /
+// session rows shown): 1300px 14/18 against 16/16, 800px 8/7 against 10/5, 700px 7/4 against 8/4. The
+// sessions are the working area, so the two tag rows go; the one cost is that ten tags scroll at 800px
+// (8 of 10 shown) where 35vh showed all ten.
+const TAG_TABLE_CAP = '30vh';
+// The open pane-filter matrix's share: bounded and scrolling within itself, so many tags cannot push
+// the sessions off the card.
+const MATRIX_CAP = '25vh';
 // THIS surface's lens: the timeline keys on actives.timeline (per-surface selections, the user
 // 2026-08-25). A pre-lens blob (an older kernel, a client-held legacy shape) derives its lens from
 // the legacy scalar EXACTLY as the kernel normalizer seeds it — behavior migrates, not just shape:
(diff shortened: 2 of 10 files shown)


henrypinkard (approved):
Approved on head dda35628 on the maintainer's word (feature-review session).

PR #1200, opened 2026-09-09
Spend: a connect seeds the watermarks from the transcript's last cost-state record

ksarma:
Follows #1192 (merged), which edited the same settle lines; this change is one commit on top of it.

## Symptom

No wrong figure has been seen: on the CLI as probed (print mode, 2026-09-06) a resume restores no cost-state record, so the connect-time seed this change adds reads zero, the CLI starts its counters at zero, and the spend record is unchanged. The change guards a restore path the CLI still carries. Its transcript loader files the `cost-state` record last-wins, its writer emits `totalCostUSD` and `modelUsage` (the two counters the settle diffs), and a resume that restores the record sets both. Should a print-mode resume restore them, the first result after a reconnect would carry the whole session's history in `total_cost_usd`, and the day's spend would gain that history as one turn.

## Cause

Every connect resets `_last_cost_total` and `_last_usage_totals` to zero, whatever the resumed transcript says a restoring CLI would hold. The init handler's cwd adoption changes the path the transcript lives under without re-reading it.

## Fix

- `last_cost_state(path)` returns the resumed transcript's last `cost-state` record as watermarks, `{"total": totalCostUSD, "tokens": modelUsage summed per field}` in the snake_case keys `_turn_usage` diffs against, or None. It scans backwards in 64 KB chunks, reassembles a line split by a chunk edge, and stops at the first valid record; `cost_state_watermarks(o)` validates one record (a negative or non-finite total is unusable) and does the sum. `_turn_usage` and the settle's token line are unchanged.
- `SdkSession._seed_spend_watermarks(resume_sid=None)` zeroes both watermarks, arms the first-result flag and, given a resume target whose transcript carries a record, seeds the totals from it and logs one info line. It replaces the two zero resets at connect.
- The init handler captures the sid the CLI loaded before the fsid flip and, inside the existing cwd adoption, re-seeds from that sid's file under the CLI's cwd while no result has settled. A fresh session has no loaded transcript and re-seeds nothing; a flip alone re-seeds nothing; an init that ends a /clear leaves the watermarks at the zero the event set.
- The settle consumes the first-result flag once the cost watermark moves. A first delta above `SANE_TURN_USD` (200) is recorded as is and traced as an info line after the spend is recorded, with `problem=False`: on the CLI as probed it is the turn's own cost, and a problem row would send the user to check a figure that is correct.

Residuals: a record the CLI wrote but did not restore (a lastSid left on the conversation a /clear abandoned) seeds watermarks the process never holds; the shrunken-counter rule records the first turn whole while its total sits below the seed, and a first turn costlier than the whole recorded history is under-counted by the seed. The same holds per token field: a first turn whose count in one field exceeds the recorded history's is under-counted in that field. A transcript with no record costs one sequential backwards read of the whole file at connect, on the event loop.

## Tests

All in `tests/test_sdk_backend.py`, class `SpendRecord`, red before the change: each fails on `AttributeError` because `_seed_spend_watermarks` and `last_cost_state` do not exist. With the watermarks at zero, the first case's result would record 13.0 USD and 1500 input tokens for a 0.5 USD turn.

- `test_connect_seeds_the_watermarks_from_the_transcripts_last_cost_state`: a transcript with a superseded record, a last record (12.5 USD) and a uuid-less trailer seeds 12.5 and its four token totals, said once as an info line with the problem ring empty; a result at 13.0 records 0.5 USD and 500/60/0 tokens and adds no line; a re-seed says so again, and a result at 0.7 (a CLI that did not restore) is recorded whole.
- `test_no_cost_state_record_and_no_resume_target_leave_the_watermarks_at_zero`: stale watermarks go to zero with a record-less transcript, the first result is recorded whole, and no resume target seeds nothing.
- `test_last_cost_state_reads_backwards_across_chunk_edges_and_takes_the_last_valid_record`: a record 200 KB from the tail behind an unusable one; a record straddling the chunk edge (the test asserts the edge falls inside it); last-wins; two models summed per field with a non-map entry skipped; negative, NaN and infinite totals; missing, empty and record-less files return None.
- `test_init_correcting_the_cwd_re_seeds_the_watermarks_before_the_first_result`: the seed under the registry's cwd variant reads zero; an init adopting the CLI's cwd re-seeds 12.5 and the first result records 0.5 USD; a cwd correction after a settle leaves the watermarks alone.
- `test_an_init_that_lands_a_new_fsid_and_corrects_the_cwd_re_seeds_from_the_file_the_cli_loaded`: the re-seed reads the old fsid's file under the CLI's cwd while the reg carries the new fsid; a flip without a cwd change keeps the seed while the record on disk changes to 99.0 under it (a re-seed would have read it).
- `test_an_init_that_ends_a_clear_and_corrects_the_cwd_keeps_the_watermarks_at_zero`: with `_clearing` set and a 7.0 record planted in the new fsid's file, the init's flip and cwd adoption leave zero (no file is read) and the first post-clear result is recorded whole.
- `test_a_first_result_after_connect_above_the_single_turn_mark_is_recorded_and_traced_as_info`: 5.0 says nothing; 250.0 after a reconnect is recorded and traced once with the problem ring empty; a 500.0 result next (a 250 USD delta) adds no line.
- `test_result_message_records_and_the_kernel_serves_it` (existing source pin): follows the seed call at the connect site. The connect path needs the SDK, which the suite runs without, so the call at that site is pinned by source text, as the two lines it replaces were.

Full suite: `pytest -q -n 8 tests/` on this head, twice: 10167 passed, 61 skipped, 0 failed (76 s and 78 s). Two runs of an earlier revision of this change each had nine failures in modules the change does not touch (`tests/test_sdk_live_rows.py`, `tests/test_api_health.py`, `tests/test_sdk_kernel.py`), a module-load ordering effect under xdist; each module passes alone.

Tier: fix

### DIFF
--- a/kernel/sdk_backend.py
+++ b/kernel/sdk_backend.py
@@ -821,7 +821,7 @@ def cost_state_watermarks(o):
     return {"total": float(total), "tokens": tokens}
 
 
-def last_cost_state(path):
+def last_cost_state(path, scan_bytes: int = 8 << 20):
     """The resumed transcript's LAST `cost-state` record, as the watermarks a CLI that restores it would
     hold (cost_state_watermarks), or None when the file has no such record or cannot be read.
 
@@ -851,16 +851,19 @@ def last_cost_state(path):
 
     Scans BACKWARDS in 64 KB chunks (a line split by a chunk edge is carried into the earlier chunk and
     reassembled) and stops at the first hit, so a transcript that carries the record costs a chunk or
-    two, and one that never carried it costs one sequential read of the whole file, on the event loop
-    at connect time."""
+    two. The scan is BOUNDED to the last `scan_bytes` (8 MB by default; last_record_uuid's tail read has
+    the same reason: a transcript can be tens of MB and this runs on the event loop at every connect):
+    a record older than that is treated as absent, which is the answer the print-mode CLI gives for its
+    own counters today (it restores nothing), so the two sides still agree (review find, 2026-09-09)."""
     marker = b'"cost-state"'
     try:
         with open(path, "rb") as f:
             f.seek(0, os.SEEK_END)
             pos = f.tell()
+            floor = max(0, pos - int(scan_bytes))   # the oldest byte the bounded scan may reach
             carry = b""
-            while pos > 0:
-                step = min(pos, 1 << 16)
+            while pos > floor:
+                step = min(pos - floor, 1 << 16)
                 pos -= step
                 f.seek(pos)
                 chunk = f.read(step) + carry    # carry: the later chunk's first-line fragment, which this
--- a/kernel/sdk_backend.py
+++ b/kernel/sdk_backend.py
@@ -786,6 +786,109 @@ def last_record_uuid(path, tail_bytes: int = 262144) -> str:
     return ""
 
 
+# The first result after a connect whose cost delta exceeds this is traced as an INFO line in the settle
+# (SdkSession._on_message), never a problem. On the CLI as probed (print mode, 2026-09-06) a resumed
+# process starts its counters at zero: a resume neither restores a cost-state record nor arms its writer.
+# So a first delta this large is the turn's own cost, recorded right, with nothing for the user to act
+# on; a problem row here would send them to check a figure that is correct. The trace is for the day a
+# CLI restores cost history in print mode: then a first result carries the whole history in
+# total_cost_usd, and the figure is wrong only if that restore and the connect-time seed read different
+# files (last_cost_state).
+SANE_TURN_USD = 200.0
+
+
+def cost_state_watermarks(o):
+    """The watermarks a CLI would hold after restoring the `cost-state` transcript record `o`, or None
+    when `o` is not one or its total is unusable: {"total": float, "tokens": {snake_case: int}}. The
+    tokens are the record's modelUsage summed per field across models into the four keys
+    SdkSession._turn_usage diffs against, the same sum it makes of a result's map; an absent or empty
+    map seeds no token watermarks."""
+    if not isinstance(o, dict) or o.get("type") != "cost-state":
+        return None
+    total = o.get("totalCostUSD")
+    if not isinstance(total, (int, float)) or not (0 <= total < float("inf")):
+        return None
+    tokens = {}
+    mu = o.get("modelUsage")
+    if isinstance(mu, dict) and mu:
+        tokens = {k: 0 for k, _ in SdkSession._USAGE_KEYS}
+        for m in mu.values():
+            if not isinstance(m, dict):
+                continue
+            for k, mk in SdkSession._USAGE_KEYS:
+                v = m.get(mk)
+                tokens[k] += int(v) if isinstance(v, (int, float)) else 0
+    return {"total": float(total), "tokens": tokens}
+
+
+def last_cost_state(path):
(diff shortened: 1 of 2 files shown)


ksarma:
The owner merged this PR at ac901fdb (3dc6cf5f, 2026-09-09 19:49Z) while this review closed, which settles the tier question here; the code review below stands as the record. The merged head adds one commit on the reviewed 34cb7f92: `last_cost_state` now scans only the transcript's last 8 MB (a `scan_bytes` parameter, default 8 MB), which bounds the connect-time read the review flagged; the rest of the diff is unchanged.

Reviewed inline: the backwards reader, the `modelUsage` sum through `_USAGE_KEYS` into the keys `_turn_usage` diffs, the seed at connect, the re-seed inside the init handler's cwd adoption from the pre-flip sid with its stand-down after a `/clear`, and the first-result flag consumed once after the watermark moves. The reader uses 64 KB chunks, carries a fragment across a chunk edge and reassembles it, lets the last valid record win, skips a truncated trailing line, and rejects a negative or non-finite total. The trace files as info after `_record_spend`. The #1192 lines keep one `_record_spend` per result and their once-per-cause flags. In the installed CLI 2.1.266 the cost-state writer is gated on a claim, and a print-mode resume restores nothing. So every seed reads zero today, and the recorded spend is unchanged, as the body says. That makes the change preventive: no wrong figure exists on the CLI romp runs. The diff adds a transcript scan at every resumed connect, plus a seed, a flag, a constant and a trace. It accepts a rare under-count today (a record the print-mode CLI does not restore) to guard a path print mode does not exercise. Under `docs/pr-tiers.md` and the review bar, that is a feature. A member's feature waits for the owner's approval on the head, so this session does not merge it.

Findings:

- Tier label (record only): the review read this as a feature (a preventive guard delivered as new mechanism, no observed wrong figure); the owner merged it as a fix, which is his call.
- `kernel/sdk_backend.py:866` (follow-up): `chunk = f.read(step) + carry` copies the carried fragment again on every 64 KB step, so the scan is quadratic in line length. Ordinary lines totalling 60 MB scan in 0.04 s; one 20 MB line takes 0.43 s; one 40 MB line takes about 8 s. Collect fragments in a list and join at the newline, or drop a carry past a few MB; a record is under 1 KB.
- `kernel/sdk_backend.py:844` (follow-up): the docstring names the abandoned lastSid as the one way a record reaches a resumed file. Its previous sentence says the interactive CLI claims the ledger at startup. A transcript opened there and later resumed by romp is a second way, with the same under-count. Name it.
- `kernel/sdk_backend.py:3901` (follow-up): the watermark comment still says the connect resets to zero. Add the seed.

Red-first at main (c7b175c9) with the PR's test file: SpendRecord 8 failed, 18 passed. The failures are the seven new cases, on AttributeError for `_seed_spend_watermarks` and `last_cost_state`, plus the source pin. At head, `tests/test_sdk_backend.py` 241 passed, 35 skipped. With the symbols present but the seed never applied, three cases fail on 0.0 != 12.5. CI at 34cb7f92: all eleven checks green.

The 8 MB bound leaves the carry concatenation quadratic within the bound only (a single 8 MB line costs one pass of about a hundred 64 KB steps), so the follow-up at 866 is smaller than stated above but still worth the list-and-join.

Review run by the maintainer automated review session; verified against head ac901fdb (the reviewed 34cb7f92 plus the 8 MB bound commit, read in full).

PR #1201, opened 2026-09-09
Chat page: a Compact tabs and agents setting shrinks the tab strip and the background-work panel

ksarma:
Builds on #1168 (Tab strip: one tag group per row becomes a gear setting, on by default): both changes add a row to the gear's Chat section and edit the same one-line settings literals (DEFAULT_SETTINGS in settings.ts, load() and the open-time fill in gear.js) and the same closing paragraph of the guide's chat section, and one dense rule here scales the untagged trail's divider that #1168 introduces. This PR is one commit on top of #1168's head; the diff to review is that commit.

A "Compact tabs and agents" setting in the gear's Chat section, off by default, shrinks the tab strip's tabs and group headers and tightens the background-work panel under the transcript (its rows, and a cap of about four rows while none is open), so a small screen keeps more of the transcript in view. Density only: one body class from a pure applier and one scoped block at the end of styles.css; every default is byte-identical.

## Tier

Tier: feature

## Design

- Why a new key. `compact` is the transcript's own tidy-up (tool runs collapsed, thinking hidden) and defaults on; the density of the strip and the panel is a different choice with a different default (off), so it is its own boolean, `denseChrome`, stored per browser with the other chat settings in romp:settings. The kernel never sees it.
- One body class from a pure applier. dense-chrome.ts applyDenseChrome(doc, s) toggles body.dense-chrome on `=== true`; render.ts calls it from applyChatScheme after applyTheme, so it runs at startup and on every settings change, and a flip in another tab or pane repaints both surfaces through the cascade with no rebuild of the strip or the panel. Pure over the document it is handed, like theme.ts, so a test runs it against a fake body.
- Density only, defaults untouched. Every rule is scoped to body.dense-chrome in one block at the end of styles.css, each shadowing a default written above and recording the shadowed value in a trailing comment; no render path changes. Off, the page is exactly what it was before the setting existed. The block stays at the tail because tag-mounts.test.ts and bg-tasks-layout.test.ts read the FIRST match of .tab-add, .tab-tagbox, .bg-sum and .bg-head.
- Sizes reuse the sheet's rungs, no new font size. Tabs go 0.92em to 0.86em (both on the six-rung em ladder css-vocab.test.ts pins); the panel's rows take its own 11px and 10px (.bg-since, .bg-status). The group header keeps 0.82em: the next rung, 0.72em, is 9.4px at the 13px base, under the 10px floor the sheet's smallest labels sit at, and the sections wear one size (tab-groups.test.ts). A federated tab's host prefix is restated at 0.92em because 0.86em inside a 0.86em tab would compound to 9.6px; at 0.92em it keeps its 10.3px render. .tab-add restates its 1.1em because the dense .tab rule (0,2,1) outranks .tab-add's own (0,1,0).
- Click targets after, measured in a browser (dense-chrome-layout.test.ts). A tab 25px tall (32px by default); a group header the same, since #tabs stretches every item on a row to the tallest (on its own it is about 25px, from 31px); a panel row about 20px, or 24px with the open-transcript arrow, which keeps its 18px. The tag control's floor follows the + tab (18px line + 2 x 3px + 1px border = 25px), since #tabs stretches every item on a row to the tallest. Under the 24px touch floor on a coarse-pointer tablet wider than the phone layout is the opt-in trade the setting exists for; the row's arrow and the Stop button keep their sizes.
- The status word hides only where the dot already says it. A row's caption is its status (running, armed, completed, failed) and the dot wears that status's tint, so the word repeats the dot; a timer's caption stays because its dot says only waiting. Hidden by the sheet, so the word is back the moment the setting goes off.
- The list cap. About four rows (100px: four 24px rows plus the list's padding) while no row's details are open, using the inner scroll .bg-list already has; the rule carries :not(:has(.bg-task.open)) so an open row lifts the cap to the panel's own min(50vh, 340px) and a command's output tail is not read through two nested scrolls.
- #tabbar is untouched on purpose. Under the phone layout's media rule (pointer:coarse and max-width 1024px, kernel.py _CHAT_MOBILE_CSS) the kernel hides the strip and shows its session picker, so the strip rules apply wherever the strip shows and on a phone the setting tightens the panel alone. The gear's sub-line and the guide say so, leading with the panel.
- Deliberately left out: no change to the phone layout's own header or picker, no change to `compact`, no kernel change, and no render-path change; a denser transcript would be its own setting.

## Tests

- ui/webview/settings.test.ts: DEFAULT_SETTINGS.denseChrome is false, a fresh store reads off, saveSettings({denseChrome: true}) round-trips through localStorage, and a store from before the key reads off. Before the change the key is absent and the first assertion fails (undefined is not false).
- ui/webview/dense-chrome.test.ts: executing, the applier against a fake body adds the class for true alone and removes it for false, the default, junk, and a store from before the key; before the change the module does not exist and the test bundle fails to resolve it. Source pins in the same file (the theme.test.ts / chat-scheme.test.ts convention, each whole-file pin asserting a boolean so a red pin prints its message, not the file): render.ts imports the applier and calls it inside applyChatScheme after applyTheme; gear.js has the row between Compact transcript and Show git branch, the handle, the save path, the open-time fill, the default in both load() literals and never `denseChrome: true`; every dense rule with its exact value; every line naming the class is body-scoped; every shadowed default unchanged and before the block; the status-word rule covers exactly the four statuses and leaves .bg-sum and the arrow alone; the cap's guard and 100px; the tag control's 25px recomputed from the + tab's line, padding and border; every size a rung and no compounded em under 10px at the 13px base; feed.css has no dense rule; the guide names the setting by its gear label and leads with the panel.
- ui/webview/dense-chrome-layout.test.ts: measured, where a Playwright browser exists (CI installs none, so there it skips loudly, the queued-romp-layout.test.ts pattern). The strip and panel markup render.ts builds is laid out against the real stylesheet with the class off, on, on with a row's details open, and off again: the tab, the + tab, the tag control and the group header at 25px (32px off); the divider's width unchanged and its line 13px; the host prefix at the same 10.3px in both states; the list at 100px with every row closed and its cap gone with a row open, inside the panel's 340px; an agent row at 24px with its 18px arrow and a plain row at about 20px; the four status words hidden and the timer's shown; the label and elapsed sizes at 11px and 10px; and every measurement identical before the class went on and after it came off. It sees what the text pins cannot: a later default rule that outranks the guarded cap leaves every pin green and turns this test red.
- tests/test_kernel.py test_gear_has_compact_tabs_and_agents_toggle (the test_gear_has_show_git_branch_toggle convention) and tests/test_kernel_settings_sections.py (the row listed under Chat); both fail before the change on the missing id=rs-dense.
- Existing pins that stay green with the block at the tail: css-vocab (em ladder), tab-groups (one size across the .tab-group- rules), tag-mounts and bg-tasks-layout (first-match reads), theme, chat-scheme, tab-drag-live.
- Ran: npm run typecheck; the full npm test (3670 tests, 0 failures, 0 skipped); pytest tests/test_kernel_settings_sections.py plus the two gear pins in tests/test_kernel.py (11 passed).

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -141,6 +141,13 @@ not loaded yet.
 ![Clicking a skeleton tab puts up the loader until its transcript arrives](assets/guide/reconnect-skeleton-click.png){ width="32%" }
 ![The clicked tab, loaded](assets/guide/reconnect-skeleton-loaded.png){ width="32%" }
 
+**On a small screen.** To keep more of the transcript in view, turn on the gear's
+**Compact tabs and agents** setting. It tightens the rows in the background-work panel above the
+composer (the one headed **Awaiting** or **In the background**) and shows about four of its rows,
+scrolling for the rest; the cap lifts while a row's details are open. Where the tab strip is showing,
+it also shrinks the tabs and group headers; on a phone the session picker stands in for the strip, so
+there the setting tightens the panel alone. Like the other chat settings, it is per browser.
+
 ### The feed
 
 The feed is Romp's task-management layer: a card for each task. Romp's
--- a/tests/test_kernel.py
+++ b/tests/test_kernel.py
@@ -4734,6 +4734,18 @@ class ViewBuilder(unittest.TestCase):
         self.assertIn("showBranch: false", _gear_src())               # load() default OFF, both branches
         self.assertNotIn("showBranch: true", _gear_src())             # the old default must not linger
 
+    def test_gear_has_compact_tabs_and_agents_toggle(self):
+        # the user 2026-09-08: a "Compact tabs and agents" checkbox in the Chat section shrinks the tab strip's
+        # tabs and group headers and tightens the rows of the background-work panel (one body class, styles.css
+        # body.dense-chrome, applied by dense-chrome.ts from render.ts). OFF by default: an explicit stored true
+        # opts in. It mirrors render.ts' loadSettings().denseChrome read, persisted in romp:settings.
+        self.assertIn("id=rs-dense", _gear_src())
+        self.assertIn("Compact tabs and agents", _gear_src())
+        self.assertIn("s.denseChrome = dn.checked", _gear_src())          # change → persist
+        self.assertIn("dn.checked = s.denseChrome === true", _gear_src())  # open → reflect (default OFF)
+        self.assertIn("denseChrome: false", _gear_src())              # load() default OFF (dense-chrome.test.ts counts both branches)
+        self.assertNotIn("denseChrome: true", _gear_src())
+
     def test_chat_body_has_an_explicit_send_button(self):
         # The web-dashboard composer (kernel _chat_body, a SECOND copy of vscode-extension/src/page-skeleton.chatBody)
         # carries an explicit send button beside 📎, so ⏎ isn't the only way to send (the user 2026-06-17).
--- a/tests/test_kernel_settings_sections.py
+++ b/tests/test_kernel_settings_sections.py
@@ -46,7 +46,7 @@ class SettingsSectionsTest(unittest.TestCase):
                     "id=rs-conserve", "id=rs-thinksum", "id=rs-fileedit"):
             self.assertTrue(h.index(">Sessions<") < h.index(rid) < h.index(">Chat<"), rid)
         # Chat: transcript prefs AND the comment defaults (comments are part of the chat)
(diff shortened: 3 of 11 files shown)


henrypinkard (approved):
Approved on head 5ca1f323 on the maintainer's word (feature-review session).

henrypinkard (approved):
Approved on head f481c84f on the maintainer's word (feature-review session), after the branch update onto main.

PR #1202, opened 2026-09-09
File viewer: rendered markdown in a document type scale, a centred 80-character measure, numbered fences with Copy, six more grammars, readable code comments, and a print sheet

ksarma:
A markdown file opened in the file viewer is set for reading: a document type scale on a centred 80-character measure, every code block numbered and given a Copy button, six more grammars, code comments that clear 4.5:1 on a code block, and a print sheet that prints the file alone. This change is stacked on the text-size control (#1194) and its diff is against that head: the type scale multiplies `--fv-scale`, the property that change introduced, so A+ and A- scale the prose, its headings, its fences and its column together.

## Tier

Tier: feature

## Builds on #1194

Every text size in the rendered document is a multiple of `--fv-scale`, and the measure is 80 of the root's own `ch`, so the two changes are one story for a reader. That change's per-child 860px measure and its top-level picture cap are replaced here by the root's padding (`.fileview-md img { max-width: 100% }` already caps a bare picture at the column once the measure is the root's padding), and its layout test is re-pinned and extended below. Until #1194 merges, this PR shows both commits; the second is this change.

## What changes for a reader

- The rendered document wears a document's type scale: a sans face (`--font-doc`, sans in both themes) at 1.15 times the page's size (14.95px at the default 13px), line-height 1.5, GitHub's heading ladder (2 / 1.5 / 1.25 / 1em, h5 and h6 dimmed, rules under h1 and h2), a 2em list gutter.
- The column is 80 characters of the document's own type, centred, never inset less than 18px; a pane too narrow for it lays the prose at the pane's width.
- A table no wider than the column sits with the prose; a wider one grows evenly into both gutters up to the body's width and scrolls inside itself beyond that; a table inside a quote or a list item keeps the prose width. Headers have a fill, even rows a tint, and a column aligned with `:---:` or `---:` is honoured.
- Task items lose the bullet beside the checkbox, and the checkbox is drawn for the page's theme. A checkbox an author's raw HTML puts after an item's text is left where the file put it, bullet and all. `<kbd>` is styled. Fences take tab-size 4.
- Every code block, labelled or not, has the chat's numbered rows and a Copy button. Copy copies the block as the file holds it: tabs come back as tabs (marked expands a leading tab to four spaces before the rendered text exists).
- Fences labelled rust, go, c, java, sql or toml are highlighted; the code view opens .rs, .go, .c, .h, .java, .sql, .toml and .ini highlighted too.
- Code comments are readable on a code block: `--hl-cmt` rises from 2.85:1 to 4.79:1 (dark) and from 4.39:1 to 4.86:1 (light).
- Ctrl/Cmd + P while a rendered file is open prints the file alone, black on white, across as many pages as it needs, tokens and gutters in full black, tables laid out as tables.

Three chat-visible changes ride along:

- The chat's gutter rule in styles.css changes. Its basis is `max(2.5em, digits in ch)` where it was a fixed 2.3em (the shared module writes `--ln-digits` on the code element, so a 10000-line block's rows share one text column; every chat code block's gutter is about 2px wider at 12px), its line box is `line-height: 1` with `overflow-wrap: normal`, and a blank row keeps its full height (a word joiner in the empty cell, `.ct::before`).
- The chat's code comments recolour with the raised `--hl-cmt`.
- A fence in a reply labelled rust, go, c, java, sql or toml now highlights by name, since the six grammars register on the core the chat bundle carries too; an unlabelled fence is still guessed among the chat's ten.

## Design

- **One shared module.** `wrapCodeLines` and `addCopyBtn` were module-private in render.ts, and file-view.ts cannot import render.ts (render.ts imports it, and the feed bundle must not carry the chat). code-block.ts is their meeting point; both callers use the same two functions and the chat's behaviour is unchanged.
- **`--font-doc` rather than `--font-prose`.** The mono ruling for the light theme was about assistant output; a README in mono reads as source. `--font-prose` is untouched.
- **1.15 and 1.5.** The prose is 1.15 times `--fs`, with line-height 1.5 set on the root, so the same document is the same height over the chat (whose body inherits 1.6) and over the feed (1.5). Inline code stays 0.92em, fenced code 12px, `<kbd>` takes the 0.86em both sheets already use for it elsewhere, and everything else is in em of the prose; the 1.15 factor is the one new constant, in the `.fileview-md` rule.
- **The measure is padding, rounded down.** 80ch as inline padding replaces the 860px cap, so the column is the same characters wide at every text size. The padding is rounded down to whole pixels: a fractional left edge put text on sub-pixel positions where a drag selection stayed collapsed. A plain `calc()` fallback precedes the `round()` form for an engine without it. `contain: layout` on the root turns anything else a document carries that is wider than the column (a fixed-width svg, an iframe) into ink overflow the body cannot scroll to.
- **The body's width is written on the tables.** A top-level table's cap is the body's content width less the root's inset, which the table's own percentages cannot reach (its 100% is the column). A ResizeObserver on the body writes it as `--fv-body-w` on each top-level table after every paint; the property is registered non-inherited with `@property`, so a write restyles the tables alone rather than every node under the body. Before the first report, or without ResizeObserver, the sheet's fallback holds and the cap is the column. The shifted table never scrolls the body: scrollable overflow is computed from the transformed box.
- **The task stamp reads the item's first node.** marked emits a task item's checkbox as the li's first node (inside its first paragraph in a loose list), with no hook on the li, so mdBlock stamps GitHub's `task-list-item` after the sanitize on a disabled checkbox that opens its item. `:first-child` counts elements alone, so the stamp also checks the checkbox has no previous sibling: a checkbox after an item's text keeps its place and its item keeps its bullet.
- **Copy reads the file rather than the DOM.** marked's lexer turns a line's leading tabs into four spaces each before it tokenizes, so the rendered text has spaces where the file has tabs. fence-source.ts maps marked's view of the source back to the file, character by character, and matches each code token's lines in document order; a fence it does not find copies the rendered text as before. Nothing about the parse changes.
- **Space acts on the keydown, and the acknowledgement finds its button by source.** A button's Space activation is native to the keyup, so a re-render between the keydown and the keyup lost the press; Space now acts on the keydown like Enter. The Clipboard API answers asynchronously, and the chat re-renders cards on kernel pushes, so a Copied label written to the pressed node could land on a detached one; the label is written to the button on screen of the fence with the copied source, and a swap inside the window moves it on (a MutationObserver on the ancestors, alive for the 1.2s window). The viewer's fences are replaced only by the reader's own clicks (the Rendered/Raw toggle, Reload), which can still fall between a press and the clipboard's answer.
- **The chat's auto-detection is held to ten.** The six new grammars register on the shared hljs core, which the chat bundle also carries, so `highlightAuto` would otherwise guess among sixteen: slower tails and different guesses. `highlightHtml` passes the chat's ten as a language subset. The six cost about 18 KB minified, 6.5 KB gzipped per bundle.
- **Where the print block sits.** After the viewer's image and PDF rules, in the viewer's own section of each sheet, not at the sheet's end.

Deliberately left out: any change to the chat's `.md` prose rules, to `--font-prose`, or to the Raw view's layout; a task checkbox accent (Chromium paints a disabled checkbox grey whatever accent is declared); a toml grammar of its own (highlight.js ships none; its ini grammar declares the toml alias).

## Tests

Written first; each fails on the sources before this change as noted. The harness has no HTML parser, so mdBlock is not executed under node: its fence pass and task stamp are pinned by source, and the stamp is executed in the browser leg below.

- `ui/webview/code-block.test.ts` (new). Executed over a DOM stand-in: `wrapLinesHtml` rows and the straddling-span rebalance; `wrapCodeLines` writes `--ln-digits`; `addCopyBtn` is idempotent, copies the caller's text on click, falls back to execCommand, acts on Space's keydown once per press, and lands the Copied label on the on-screen button of the fence with the copied source after a swap, by ordinal; the six grammars register with their aliases (against the real hljs); `highlightHtml` passes the ten-name subset. Pinned by source: render.ts and file-view.ts import the module and render.ts keeps no copy; the bundle boundary; mdBlock's fence pass (the language branch is read whole, head to closing brace, and holds neither call; the two calls close the forEach) and its task stamp (the selector, then the first-node check); both sheets' scoped row and Copy rules byte-equal. Before: the test bundle fails to resolve `./code-block` and `./viewer-grammars`.
- `ui/webview/file-view-fence-source.test.ts` (new; over the real marked): `markedView` maps every character; tabs come back at top level, in a list item (the partial tab), in a quote, with CRLF, in an indented block; the renderer's trailing-newline shape; document-order matching; `fenceCopyQueue`; the empty fence. Before: the test bundle fails to resolve `./fence-source`.
- `ui/webview/codeblock-copy.test.ts` and `ui/webview/codeblock-wrap.test.ts` (re-pinned onto code-block.ts and the gutter basis). Before: code-block.ts is absent.
- `ui/webview/file-view.test.ts` (re-pinned, plus two tests). The LANG map row for row, including rs, go, c, h, java, sql, toml and ini; `langFor("src/main.rs")` is rust; the `--hl-cmt` literal; the fence pass shape; the viewer-grammars import. `watchBodyWidth`, executed: lifted from the source (the suite's composerWindow lift) and run over a fake ResizeObserver and stand-in nodes: nothing before the first report; after one, the width on each top-level table and on neither a nested table nor the prose; the returned stamp writes it on a fresh root's tables; a repeated width writes nothing; the last of a callback's entries wins and an empty callback reads the body; the next watch and the drop disconnect; without ResizeObserver nothing is written. The wiring: both viewers watch the body they build, both rendered paints stamp the fresh root, the close drops the watch. Before: LANG lacks the eight rows, `--hl-cmt` is `#6f6a5f`, the fence pass returns on an unnamed fence, and `watchBodyWidth` does not exist.
- `ui/webview/theme-parity.test.ts`: the pair `--hl-cmt` on `--box-bg` at 4.5. Before: 2.85 dark, 4.39 light.
- `ui/webview/fileview-parity.test.ts`: the rule reader pins every occurrence of a head at a line start; the type-scale heads and the whole `@media print` block are added. Before: the heads are absent, and `.fileview-md th {` differs between the two sheets.
- `ui/webview/file-view-text-size.test.ts` (re-pinned): the root's font-size, its padding-inline with the fallback first, the reader list, the heading ladder, `.fileview-md > table`, no 860px anywhere. Its browser leg (headless Chromium through Playwright, skipped where none is installed) measures the 80ch measure and its whole-pixel centring at 1000 and 420px and at 100% and 150%, the even table break-out up to the body, the nested tables' width, the heading ratios, the fence rows (a blank row at full height; a five-digit gutter wider than a one-digit one), the task item's rules, and the print sheet (the bar and the Copy buttons gone, black ink, a white page, a table as a table, the body's overflow visible). A second browser case runs mdBlock's task stamp over marked's own output for a synthetic list (tight, numbered, loose, a raw checkbox opening an item, an author's own class, a checkbox after the item's text, one after bold text, an enabled one), sanitized in the page by DOMPurify with mdBlock's own options and stamped by the statement lifted from mdBlock: the opening checkboxes make task items with no bullet and the box in the gutter; the others leave the item and its bullet as the file wrote them. Before: the layout leg fails on the measure and the ladder; the stamp statement is absent.

Mutations checked red against these tests: the first-node check removed; each of the three width-watch hooks removed; the two fence calls moved inside the language branch; the observer stamping nested tables, dropping its same-width return, not dropping the last watch, and writing before the first report.

Runs on the committed head: `npm run typecheck` clean; `npm test` 3708 passed, 0 failed, with the browser legs running under a local Chromium; `node esbuild.js` builds. The browser leg is the only geometry measurement; CI installs no browser, so it skips there and the sheet pins carry the layout claims.

Restacked onto #1194's rebased head 62b5f6e7 (2026-09-09), which sits on main at 3f06288e. The shared markdown sanitizer that merged in between meets this change in `mdBlock` (the fence collection now feeds `sanitizeMd`, and the body is adopted the way the sanitizer does it) and in both sheets: the sanitizer's `contain: layout`, element-level media caps and table rule are kept, and the parent's top-level media-cap rule is dropped here because the measure is the root's padding, so the element-level caps alone hold a direct-child medium to the column. One sanitizer browser assertion was adapted: a top-level table may grow evenly into both gutters up to the body's 18px inset, as this change's own tests pin; its content still scrolls inside it and the body never scrolls sideways. Full `npm test` on the restacked head: 3790 passing, 0 failed.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -73,6 +73,19 @@ lands on it. An image map (`<map>`, `usemap`) is dropped. The same rules apply t
 a chat message, where a link to an element's own `id` or `<a name>` lands on it under the
 prefix.
 
+**Text size and width.** The **A−** and **A+** buttons in the viewer's title bar make the
+text of any text file smaller or larger in fixed steps from 70% to 200%: a markdown file's
+Rendered and Raw views, the code view of every other text file, and a document opened from
+a link on the dashboard's own address. They appear wherever the viewer opens (over the chat
+or the feed), and not for a picture or a PDF, which have no text to size. Ctrl (or Cmd) and
+the mouse wheel over the text do the same. Once the size is off 100%, the percentage appears
+between the buttons; click it to go back. The choice is kept in this browser and applies to
+every file you open here. Prose keeps a readable line length that grows with the text size,
+and code blocks keep that width and wrap long lines. A table is as wide as its columns need,
+up to the width of the viewer, and scrolls sideways on its own beyond that; a table inside a
+quote or a list item stays within the prose width. Pictures shrink to fit, so the page is
+never wider than the viewer.
+
 **Opening a PDF.** A PDF the session mentions, or one you click in the file browser, opens
 inside the dashboard like an image: the chat's PDF card opens it full-view, a path or a
 file-browser row opens it in the file viewer. Cmd-click it instead (Ctrl on Windows and
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -56,7 +56,13 @@ file served from the dashboard's own address (a published report, an evidence do
 and links inside the document resolve relative to the document, so a `![fig](fig.png)`
 beside it shows, and a link to a sibling document opens in the same viewer. Links to files
 on other sites open in a new tab, as before — and a ctrl- or ⌘-click still opens the file in
-a tab.
+a tab. The document is set for reading: a sans face at a slightly larger size, headings in
+proportion, a centred column about 80 characters wide, and task lists, keyboard keys and
+aligned table columns as GitHub shows them. Every code block is numbered by line and carries a
+**Copy** button that copies the block as the file holds it, tabs included; fences labelled
+`rust`, `go`, `c`, `java`, `sql` or `toml` are highlighted, in addition to the languages the
+chat already knows. Printing the page while a rendered file is open prints the file alone,
+black on white, across as many pages as it needs.
 
 **A file's own HTML.** The Rendered view keeps the HTML a markdown file carries, under rules
 modelled on those GitHub applies to a README, so nothing in a file can move, hide or cover the
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -249,6 +249,18 @@ ghostty --working-directory={dir}   # Linux: Ghostty
 code {dir}                          # VS Code instead
 ```
 
+### The file viewer's per-browser choices
+
+The file viewer keeps two choices in the browser's own storage, not on the
+kernel, so they survive a kernel restart and apply wherever the viewer opens
+(over the chat or the feed, and for a document opened from a link on the
+dashboard's own address): the Rendered or Raw view of a markdown file
+(`romp:fileviewFmt`) and the text size (`romp:fileviewTextSize`, one of 70,
+80, 90, 100, 115, 130, 150, 175 or 200 percent, set by the **A−** / **A+**
+buttons or Ctrl/Cmd + wheel over the text). The size scales the prose, its
+headings, the code and the Raw view together, and the prose measure with
+them; a value outside the table reads as 100.
+
 ### Model and effort, from the statusline or a typed command
 
 Typing `/model X` or `/effort X` into the chat composer, or sending one with
--- a/ui/webview/code-block.test.ts
+++ b/ui/webview/code-block.test.ts
@@ -1,0 +1,411 @@
+// The shared code-block module (code-block.ts): the per-line rows and the Copy button every rendered fence wears, in the
+// chat and in the file viewer. The chat's highlight() (render.ts) and the viewer's mdBlock (file-view.ts) import the same
+// two functions; file-view.ts cannot import render.ts (render.ts imports it, and the feed bundle must not carry the chat),
+// so the module is their meeting point. Executed: the pure walk (wrapLinesHtml) over plain and highlighted lines, a span
+// straddling a newline, a trailing newline; wrapCodeLines over a code element (the rows, the digit count it writes for the
+// gutter); addCopyBtn over a DOM stand-in (one button, idempotent; the click and Space copy the text the caller gave it,
+// through the Clipboard API or the execCommand fallback; the acknowledgement lands on the button on screen of the fence
+// with the copied source, after a rebuild swapped the pressed one out). Pinned: the two callers' shapes, the bundle
+// boundary (file-view.ts and feed.ts import nothing from render.ts), the sheets' scoped copies byte-equal in styles.css
+// and feed.css with the counter reset on `.fileview-md pre code`, the six viewer grammars registered on the core by
+// viewer-grammars.ts (executed) and reached through file-view.ts, and the chat's auto-detection kept to its ten
+// (highlight-cache.ts AUTO_LANGUAGES, executed).
+import { test } from "node:test";
+import * as assert from "node:assert/strict";
+import * as fs from "node:fs";
+import * as path from "node:path";
+import hljs from "highlight.js/lib/core";
+
+const UI = path.resolve(process.cwd(), "..", "ui", "webview");
+const read = (f: string) => fs.readFileSync(path.join(UI, f), "utf8");
+const BLOCK = read("code-block.ts");
+const RENDER = read("render.ts");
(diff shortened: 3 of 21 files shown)


henrypinkard (approved):
Approved on head bb22750a on the maintainer's word (feature-review session).

PR #1203, opened 2026-09-09
API health indicator: the hover and detail carry a History section read from GET /api-health, and the aggregator owns the boot stamp

ksarma:
Builds on #1197 (the bottom bar's API health cell), whose body names a hover history of the cell as a separate change. The API cell's hover and click detail gain a History section read from `GET /api-health` when they open, and the backend's aggregator now owns the boot stamp it serves as `bootAt`, so the head, the divider and the restart row name one time. Every JS and CSS hunk here edits the cell's script and skin that PR introduces, the guide sentence extends its paragraph, and the hover tests import the cell's frame builder and push, so the diff to review is this branch against that PR's head; GitHub retargets it to `main` when #1197 merges.

The section shows the signal's state with its since-time and reason, one row per window (attempts, 429 and 5xx shares, give-ups, sessions that retried), the newest six state changes with how long each held, and a kernel restart as its own row or a divider. It reads the signal when it opens, with no poll and no timer, and a failed read shows one line in place of the rows.

## Tier

Tier: feature

## Design

- **Data source.** The hover reads `GET /api-health` at show time on the dashboard's own cookie (a same-origin GET sends no Origin, which `_origin_ok` accepts; no token in the call), the same shape as the shell's `/usage/fleet` read. It does not ride the pushed frame: the signal derives its state at read time by design, a frame would need a periodic snapshot, and its numbers would defeat the frame's byte-equality dedupe. The frame is unchanged, with no new keys; `romp api-health` and the hover show the same document.
- **No timers.** The show (mouseenter, focus, click, Enter/Space) is the event that reads; a frame landing on an open card re-reads; a read sequence number keys the paint to the newest answer; a fresh show drops the last answer so a hover never paints an earlier hover's numbers; a pin reads only when the tip was hidden, so hover-then-click costs one read, and a pin from hidden keeps the last hover's answer behind its read (its as-of says when it was read; the card the user opened is not blanked for the read's duration, and the dots stand in only before the first answer); an answer landing under a held primary pointer waits for the release, the frame's own gate.
- **One boot clock.** The aggregator owns `boot_stamp`: the kernel's `_STARTED` floored to the millisecond (the payload's stamp precision; floor, not round, so its whole seconds are `/version`'s `started`), clamped one millisecond past a restored transition that is not before it (the previous kernel filed after this one started, or the clock stepped; logged once), and served as `bootAt`. Every seeded `stateSince`, every restart row and `bootAt` are one number in every case, so the head, the divider and the restart row agree. `SdkBackend` takes `boot_at` and the kernel passes `_STARTED`; the route no longer stamps its own `int(_STARTED)`. **Contract change to flag:** `bootAt` becomes a millisecond float on the wire; `int(bootAt) == started` holds unless the clamp moved it. Nothing in `bin/romp` reads `bootAt` (`romp api-health` prints the payload verbatim); the reference documents the stamp.
- **Restart rendering.** A row the boot filed reads `unknown · kernel restarted`; where the tail crosses `bootAt` without one (the bucket was already unknown at the previous stop, so the boot filed nothing) a `kernel restarted` divider is inserted and takes none of the six slots; any restart row above suppresses it, so one restart is one mark and a restart is never hidden. A hold from before the boot ends at the boot, since every bucket comes back unknown at a restart; `so far` is a flag, never a stamp comparison, so the transition the hover's own read files (its `t` is `asOf`) still closes the state before it.
- **Window rows.** `requests` excludes no-status attempts in the aggregator, so a row counts every attempt once and names how many had no status with the shares' base beside them (`15 attempts, 7 of them without a status · 25% 429 · 0% 5xx of the other 8`); a window reads `no attempts` only when every count is zero; a window longer than the uptime carries `kernel up <duration>`. Figures are in the window's tense (`retried`); the live set is the Sessions waiting list above.
- **Failure is loud.** A non-2xx, no answer or an answer without the signal's shape shows one line in place of the rows, never the previous numbers, never silence; the loader's three dots stand in before the first answer.
- **Accessibility.** The shown tip is `role=tooltip` on every show, the hover after a pinned dialog closed included (the close leaves the hidden tip's role alone; the show puts the tooltip role back and drops `aria-modal`), and the cell is described by a short visually-hidden summary (`#ah-summary`: the state word, its since, how to open the rest), refreshed when the read lands and never the tip's whole text; before the answer it carries the state word the frame already put on the cell. The pinned card is the `aria-modal` dialog with focus inside. Focus shows, blur or Escape hides in place. A focus the browser re-dispatches on window refocus is skipped only when the cell already held focus at the window's event (recorded there, cleared on the next animation frame, no timer), so a Tab out of a pane iframe onto the cell still shows it.
- **Hardening found on the way.** `_row_ok` requires a finite numeric `t`: a null `t` used to pass, reach the seed's max and drop every restored row and bucket through the outer guard; an int past float range raised `OverflowError` inside `math.isfinite`. Both are now one skipped, counted row.
- **UI rules.** Progressive disclosure (state word and since in the head, windows and tail under, actions only on the pinned card); the section wears the spend hover's grammar and skin, no new font size (the reason and as-of reuse `.ru-tip-reset`); status hexes never the accent; the failure red meets 4.5:1 on both themes (`#ef6b6f` on the dark tip, the light theme's `--err #B02A1C`); the light-theme quiet-dot rule extends to `healthy` and `unknown`; the hover is measured after a reset (a fixed element's shrink-to-fit width is taken against where it last sat) and capped to the room above the rail, so a wide row keeps its margin and a short window clips instead of spilling over the rail.
- **Test lane.** The kernel-inlined script is tested from the Python side: its source pins sit in `tests/test_api_health_hover.py` (beside `tests/test_api_health_rail.py`'s existing pins) and its behaviour runs in Chromium in `tests/test_api_health_hover_browser.py`; no TypeScript test reads `kernel.py`.
- **Deliberately left out.** The VS Code strip twin (`strip.ts` has no API cell; the strip pin records the follow-up) and the phone's Usage-modal section.

## Tests

Executing tests, and what each fails on before the change:

- `tests/test_api_health_hover.py` (new, 34). Payload: the fields the section reads (worst bucket with `stateSince`, `why`, windows keyed by the config), transition rows newest last, the 50-row bound, the restart row's reason text, and a bucket already unknown filing nothing so the boundary rides `bootAt`. One clock: the aggregator seeds `stateSince` and the restart row from `boot_at`; `SdkBackend` hands it through; a backend built without it seeds from its own clock; the kernel's own construction of the backend, run with the backend module swapped for a recorder, passes `_STARTED` itself as `boot_at`; the real route serves `bootAt`, `stateSince` and the restart row as one number, plain and in a same-second race with the previous kernel's last row, with one log line; floor semantics across the second boundary; the clamp judged at the millisecond, a bucket's own tail included. State-file rows: null, missing, string, bool, NaN and an int past float range each skipped as one row with the rest kept. Route: the cookie alone with no Origin, a cross-site cookie refused, a 503 without a backend. Frame: no history rides it and an unchanged world sends once across a read. Docs: the section, the stamp, the three failed reads. Script pins: the read's call sites, no timers, the paint gate, the row grammar, the tail rules, the roles and the description, the CSS with no new font size, `strip.ts` has no API cell. Before the change: no `boot_stamp`, no `bootAt` in the snapshot, `_row_ok` passes a null `t`, no History block in the script, no docs; 31 red, 3 pass before and after (the transition rows' fields; the frame carries no history and dedupes across a read; the strip pin, which holds on #1197's head by construction).
- `tests/test_api_health_hover_browser.py` (new, 29). A scratch copy of the shell page is served with switchable synthetic `/api-health` payloads and driven by Playwright's chromium: one read per show with the loader first; race and fresh-show reset; the storm's head, windows and tail with its restart row; the divider and the pre-boot hold; a bucket the boot seeded and a boot at :58; same-second ordering and the clamped row; the `so far` flag at `asOf`; bucket naming by family and auth; offline and mixed windows; the six-row cap with and without a divider; the three failure lines; section order; focus, blur and Escape; the pointer on a focus-shown tip; the window-refocus re-dispatch; a real Tab out of an iframe; tooltip and dialog roles, the hover after the dialog closed a tooltip again; the short description at focus time and after the read; pin, frame re-read and Escape; click read counts; the held pointer, waited for by the page's own count of parsed answers, never a pause; the light theme's head dot and failure line read from the computed style; geometry at 830x600 (the pointer enters near the right edge, so the first anchor sits flush at the margin and the re-render after the rows land is measured against it) and at 1200x280. Any page error fails. Before the change the served page has no History section, so every wait times out: 29 red. Skips loudly without the extension's node deps or a browser; CI installs none, so this leg is not CI-verified: it ran locally at this head, 29 passed.
- `tests/test_api_health_rail.py` (3 pins rewritten): the cell script has exactly one fetch (the history's) and still no `setInterval`/`setTimeout`; `anchor()` measures after a reset; `keydown` handles Escape before Enter/Space. Green on #1197's head before, red until the hover's JS lands.
- `tests/test_kernel_auth_hardening.py`: `_ApiHealthBackend` seeds its aggregator from `_STARTED` the way `_sdk_locked` does, and `test_boot_identity_is_versions_own` reads `bootAt` as the aggregator's stamp with `int(bootAt) == started` (red before: the aggregator has no `boot_stamp` for the rewritten equality to read; the old `bootAt == started` equality would go red after the change, so the rewrite is part of it).
- `tests/test_kernel_pane_rail.py`: the light-theme ok-dot pin names `.ah-dot[data-state=healthy]` and `[data-state=unknown]` (red before).
- `tests/test_api_health.py` (existing, unchanged) stays green: the millisecond floor is exact on its seeds and no case seeds at or before a restored row.

Mutation checks on the final tree, each turning an executing test red (a source pin alongside): the anchor's left/top reset removed (the 830 px geometry case: the rows wrap), the show leaving the tip a dialog (the tooltip case after the dialog), the light-theme quiet-dot rule dropped and the light failure red dropped (the computed-style case), and `_sdk_locked` no longer passing `boot_at` (the recorder case).

Full runs on the final tree: `pytest -n 8 tests/` 10373 passed, 40 skipped, 0 failed; `npm test` 3655 pass, 0 fail; `npm run typecheck` clean.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -631,7 +631,11 @@ how many sessions are waiting and names the problem: **rate limited**,
 **overloaded**, **offline** (this machine cannot reach the API), or **errors**.
 Red **paused** means auto-retry and the judges are stopped, and says why: a
 usage limit, the monthly spend cap, or that you stopped them. Hover for the same
-reading with the waiting sessions listed. Click the cell, or press Enter on it,
+reading with the waiting sessions listed, and the history under it: the API's
+state over the last 1, 5 and 15 minutes (attempts, the 429 and 5xx shares,
+give-ups, sessions that retried) and the most recent state changes with how long
+each held. A kernel restart shows as its own line there, because the counts
+start over with the kernel. Click the cell, or press Enter on it,
 for the detail: each waiting session (click one to open that session), a button
 that stops auto-retry for every session while sessions are waiting and resumes
 it while paused, and links to the usage figures and the Log.
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -624,3 +624,14 @@ themselves. The analytics modal in settings shows what you actually spent,
 separating your sessions from the judge pipeline. You can also reconfigure the
 judges from the gear: the high-volume indexing tier defaults to Haiku, and the
 judgment tier defaults to Sonnet.
+
+The bottom bar's **API** cell, a dot and a word, shows how the API is treating
+your sessions. Gray **ok** means no session is waiting on the API. Amber shows
+how many sessions are waiting and names the problem: **rate limited**,
+**overloaded**, **offline** (this machine cannot reach the API), or **errors**.
+Red **paused** means auto-retry and the judges are stopped, and says why: a
+usage limit, the monthly spend cap, or that you stopped them. Hover for the same
+reading with the waiting sessions listed. Click the cell, or press Enter on it,
+for the detail: each waiting session (click one to open that session), a button
+that stops auto-retry for every session while sessions are waiting and resumes
+it while paused, and links to the usage figures and the Log.
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1141,8 +1141,13 @@ label the account digest itself, so a bucket can be matched to the log.
   is the response time. A clock step moves it; a reader that wants a freshness
   check a clock step cannot fake uses `seq`.
 - `bootId`, `bootAt`, `uptimeS`: the kernel process identity, the same id
-  `/version` and `X-Romp-Boot` carry. A changed `bootId` means a restart, and
-  the windows restarted with it.
+  `/version` and `X-Romp-Boot` carry. `bootAt` is the boot's stamp in this
+  signal: the kernel's start truncated to the millisecond, the precision of
+  every other stamp in the payload, or, when the previous kernel's last
+  transition overlaps the start, one millisecond past that row; every bucket
+  the boot seeded carries this same number as its `stateSince`, and so does
+  every row the boot filed. `/version`'s `started` is the whole-second boot
+  time. A changed `bootId` means a restart, and the windows restarted with it.
 - `complete`: true once the longest window (900 s) fits inside the uptime.
 - `seq`: count of ring events (attempts, successful responses and give-ups)
   ingested since boot. Monotonic within a boot: two reads with the same `seq`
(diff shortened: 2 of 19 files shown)


henrypinkard (approved):
Approved on head 57a6f2d8 on the maintainer's word (feature-review session).

PR #1204, opened 2026-09-09
File viewer: URLs and file paths in a shown file are links

ksarma:
Links inside a file the viewer shows: an http(s) URL in the text opens in a new tab, a file path opens that file in the viewer (a `:12` or `#L12` after it scrolls to the line), and a Markdown link's target follows the same two rules. The chat's path matcher moves into a shared module first, with no behaviour change, so both surfaces link paths from one grammar.

## Tier

Tier: feature

## Two commits

1. `Chat: the path matcher moves out of render.ts into path-links.ts, a module any surface can import`. A pure lift, no behaviour change: the regex, its shape gates, the trailing-punctuation trim and the span it emits move as they are; `linkifyPathTokens` walks a body's text nodes as the chat's walk did and hands back the hits in document order; every span carries `data-path` (and `data-rel` for a bare path) and binds no action; render.ts binds the click and the middle button per span off that data (`bindPathLink`) exactly as before and reads the hits for its figure pass. The source pins that read the matcher from render.ts now read it from path-links.ts; the tokenizer parity fixture is unchanged, the regex's text being unchanged. Typecheck and the full test run are green at this commit on its own.
2. `File viewer: URLs and file paths in a shown file are links`. The feature, described below.

## What the viewer does now

- An http(s) URL in a code view or in rendered prose is an anchor (`target=_blank`, `rel=noopener noreferrer`, the URL as its title).
- A file path opens that file in the viewer, in place: a relative path is joined onto the shown file's directory, an absolute or `~/` path passes as written for the kernel's `/file` route to resolve (`~` and the file's session's machine), and the result is normalized once (`docs/../src/app.py` opens and is titled `src/app.py`; a local `file:///a/b/../c.md` goes through the same resolve and is titled the same way). No kernel change: `_resolve_open_path` already does this.
- A `:12` or `#L12` after a path (or on a Markdown target) rides in the link; the viewer scrolls the code view to that row once the text lands, once, with the row centred. A Markdown file takes its Raw view for that open only; the saved Rendered/Raw preference is untouched. A line past the end lands on the last row and the viewer's notice says so rather than landing silently on the wrong row.
- In Markdown, `[text](target)` follows the same rules. A per-parse `walkTokens` hook (this parse only, never the marked singleton the chat shares) puts a same-directory `name.ext:12` and a local `file:///` URI in a form the sanitizer keeps; marks are set after `sanitizeMd` ran (md-sanitize.ts, `ALLOW_DATA_ATTR: false`), so a document's own `data-*` never reaches the page, and an element a document dresses in the link's class names no file. A `#section` link scrolls to an element with that id or an `<a name>` of that name, read under the sanitizer's `user-content-` prefix or bare (`userContentTarget`, the lookup the chat's `#` click reads too), or to a heading by its slug (the viewer's minted `md-` ids, set after the sanitize and never prefixed), in the rendered document only (never the viewer's chrome or the page), under a middle click too; when there is none the hover says so. A target the sanitizer removed, one that resolves to no path (`[up](../)` from a file named without a directory), or a host with a port (`127.0.0.1:3000`, `localhost:8080`, `api.example.com:8443`) renders as a dead link whose title says why: never a silent click, never a file named after a host. A Markdown link with a scheme keeps the `target=_blank` and `rel=noopener` it had; a query-only link (`?x=1`) opens a tab with `rel=noopener noreferrer`.
- The sibling links #958 taught the viewer keep what they had (they open in the viewer and land on their `#fragment`) and gain the line, the gesture and the dead-link verdicts. Their `fv-open` stamp and the body delegate are replaced by the shared path-link shape on the anchor and one body listener; md-url-view.test.ts's pins on that code are re-aimed, not deleted. The URL viewer's `fv-anchor` delegate is unchanged.

## Design

- **One grammar.** The viewer runs the chat's matcher with options (`inPre`, `accept`, `resolve`, `lineSuffix`, `unit`) under one extra gate, rather than a second matcher. The chat's behaviour is unchanged except for the three things stated at the end of this section.
- **The viewer's gate**, since this surface has no kernel stat verdict to narrow matches: a token links when it has a slash and a letter-led extension of one to eight letters or digits on its last segment (a dotfile only under `/`, `./`, `../`, `~/`); it is not glued to the character before it (line start, whitespace, or an opener: a quote, a bracket, `=`, `,`, `;`, `|`, or Markdown's `*` closed by a `*` after the path), so `$HOME/docs/a.md`, `${dir}/out.json`, `@scope/pkg/index.js`, `C:/Users/x.txt` and `git@host:user/repo.git` stay text; it is not inside a URL on its line; an unanchored first segment that reads as a hostname does not link; an unanchored token named by an import statement or a `require()`/`import()` call does not link, while the English `from` or `export` before a quoted path in prose does; a `~` start is `~/`. A local `file://` URI (empty authority or `localhost`) is a path; `file://host/path` is prose. The grammar is about 200 lines because every rule has a case that would otherwise link a path never written; the executable grammar tests are the record of those cases.
- **The pass changes no character.** It runs over the DOM the highlight and the sanitizer built, a line at a time (a code view's row, a rendered block; a `<br>` ends a unit), and only wraps runs of text, so selections and the quote chip over a line with a link keep working. A token a highlight span cuts into two nodes is left as text: a span around `$HOME` must never expose `/docs/a.md` as an absolute path. Text inside an inline SVG is read but never marked (an HTML element inside SVG text does not render).
- **Linear in the text.** The regex's g-flag loop is quadratic on a long run of word characters (one slash plus a 40K run of hex or dashes costs seconds), and a viewer shows whole files; `PathTokenScanner` drives the same regex in linear time, and a differential test pins that it finds exactly what the regex finds from every start position. The import look-behind is bounded to the line, the line suffix is read in place, and the span holding a token is found by binary search.
- **Gesture: the project's PDF rule** (`wantsOwnTab`). A plain click acts in the dashboard (a path opens in the viewer, a URL in a tab, a section scrolls); a Cmd/Ctrl-click or a middle-click opens the link in a browser tab of its own, a path off the kernel's `/file` route through a new `openFileTab` that `openPdfTab` now wraps (one `window.open` in the gesture, the opener severed by hand, a blocked popup falls through to the viewer). A click that arrives with text selected inside a link opens nothing, in the viewer's listener and in the chat's capture-phase anchor opener alike (`selectionOpenIn`, read for a non-draggable anchor only, so a triple-click on a chat paragraph does not deaden the link inside it); the next click, its press having collapsed the selection, opens. The viewer's anchors are not draggable and are selectable: a press-drag inside a path link selects its text as on plain text, a press held on a URL anchor and then dragged selects too (Chromium's rule for a link: an immediate drag on one selects nothing), and neither starts the browser's link drag.
- **Click-safe.** One listener on the viewer body, installed once per open and kept across Rendered/Raw swaps. A plain path click is not stopped, so the document's own listeners (the feed's focus return, menu closers) still run. Enter or Space on a focused path link is its click, with a held Cmd/Ctrl carried. A middle-click on a section link scrolls and opens no tab (the browser's own would open a second copy of the page).
- **Dress, progressive disclosure.** A token keeps the colour the highlight gave it under a faint dotted underline that turns solid on hover; a Markdown link that names a file keeps the ordinary link ink (`--link`); a dead link is dimmed with `cursor: help` and its reason in the title. No new font sizes, no new colour tokens, no loading state (the pass is synchronous on text that has already landed). Both sheets, byte-equal (fileview-parity.test.ts).
- **Security.** The `walkTokens` hook rewrites link targets before the sanitizer runs, and a local `file:///` target becomes a path; marks are set after `sanitizeMd` with its `ALLOW_DATA_ATTR: false`; the URL anchors the pass makes carry `rel=noopener noreferrer`, and a Markdown link with a scheme keeps its `rel=noopener`; `openFileTab` severs the opener; a dead anchor loses its href; a path link's href comes off, so the browser never follows it and the chat's opener never sees it.
- **The three chat-facing changes** the shared module brings: `file://host/path` no longer links as a relative path named after its host (it is prose); a path link is a tab stop that Enter or Space clicks, with a mouse press leaving focus where a click on plain text leaves it; and a path inside an inline SVG's text in a chat message stays text (an element inside SVG text does not render, so the span the old walk put there made the token vanish).

## Considered and not done

- A second matcher for the viewer: two grammars drift, and the kernel's parity fixture pins one regex.
- `joinDocPath` for Markdown targets: it keeps `..` above a relative start and decodes `%`; the viewer's `resolveViewerPath` normalizes once and is the one resolver for text tokens, URIs and Markdown targets alike, so a file is titled the same whichever way it was reached. `joinDocPath` still serves the figure sources.
- A lookup of its own for section links: the sanitizer's `userContentTarget` is the lookup the chat's `#` click already reads, so both surfaces resolve an author's id the same way.
- Left out on purpose: no back stack after following a link; no kernel existence check for a path (a `:line` into a file that shrank says so; a path that no longer exists gets the viewer's own not-found pane); the marked-failure fallback stays bare text.
- Known limits of the grammar: a `*`-wrapped span holding two paths links only the second; a letter-led `host:port` Markdown target is stripped by the sanitizer before the module runs, so its dead link carries the generic title; a two-letter country TLD without a registry label reads as a file; an extension on the TLD list (`Makefile.local:5`) reads as a host; a line holding only `from` and a quoted path has a hand-split import's shape and stays text.

## Tests

Each executing test below fails on the tree before the change as described; the pins re-aimed to the new shapes fail before it too.

- `ui/webview/file-view-links.test.ts` (new, 25 tests over a DOM stand-in): the URL segments and ranges; `viewerPathGate` alone (extensions, anchored dotfiles, hostnames, `~user`) and with the line (glue, openers, emphasis stars vs globs, imports and requires, hand-split imports, the English `from`); the bounded look-behind over an 8000-line fence; `normalizePath` and `resolveViewerPath`; the local `file://` arms, a URI's path normalized through the resolve; the pass over code rows (URLs, quoted paths, `:line`, prose left alone), the split-span fixture (no fabricated `/docs/a.md`), text that must not link, escaping; `viewerLinkTarget` and `isHostName`; `linkMarkdownAnchors` (URL, file, query-only, fragment, `<a name>`, an id and a name under the sanitizer's `user-content-` prefix, dead, host:port, empty `../`); soft breaks, `<br>`, a blank row, a thousand spans; SVG text; `wantsOwnTab` and `openFileTab` against a fake window; source pins on the viewer's wiring and both sheets. Before the change the module does not exist, so the test bundle fails to resolve it.
- `ui/webview/file-view-links-browser.test.ts` (new, 5 tests; headless Chromium over the real viewer module, plus a page running the chat's own anchor opener lifted from render.ts with a check that every name it uses is defined; skips loudly without a browser, as md-sanitize-browser.test.ts does): hljs's real rows and split spans; a plain click opening the sibling; the `:30` open landing Raw with row 30 itself centred (a neighbour is a row's height off) and the preference unwritten; the shared sanitizer's verdicts on Markdown targets, an author's id and name reaching the DOM prefixed and the section links that land on them, a document's own element dressed as a path link losing its `data-*` and opening nothing, a section link scrolling the document and not the page under a plain click and a middle click (no tab), the line-past-the-end notice; Cmd-click and middle-click tabs off `/file`, a URL anchor's tab; a press-drag inside a path link selecting its text and opening nothing, a drag from plain text into a link opening nothing; under the chat's opener a triple-click then a link click opening one tab, a click on the URL anchor with its text selected opening nothing and the next plain click one tab, and the viewer's own section link landing under the prefix. Before the change: no anchors and no path links in the viewer, a Markdown `[same](notes.md:7)` carries no path, a Cmd-click opens no tab. Removing either selection yield, landing the `:30` one row late, dropping the section arm's preventDefault, or reading section ids without the prefix each turns a browser test red.
- `ui/webview/path-links.test.ts` (new in the first commit, extended in the second): the span's shape, the shape gates on the real functions, the walk over a chat body, inline code, skipped text, the pathLinks verdict, the resume rule; the keyboard and press handlers; `markPathLink` on an anchor; the local-URI arm; the scanner's arms, the differential against the regex from every start position over adversarial and fuzzed texts, the parity fixture on the walk's own loop, the trailing-punctuation source, pathological inputs; `textUnits`, `spanHolding`, `rewriteSpan`; the walk under the viewer's options.
- `ui/webview/fileview-parity.test.ts`: the five link rules exist in styles.css and feed.css byte-equal (absent before).
- Re-aimed pins: `md-url-view.test.ts` (the file kind's anchors are the module's, one body listener, `scrollToFragment`'s lookup, the opts shape, the marks after `sanitizeMd`), `pdf-new-tab.test.ts` (`openPdfTab` wraps `openFileTab`; the viewer's own open reads no gesture), `file-view.test.ts`, and the chat pins in `chat-path-links`, `chat-relpath-link`, `chat-inline-preview`, `file-uri-link`, `chat-space-paths` (render.ts to path-links.ts in the first commit, the walk's options and its open target in the second); `tests/test_path_links.py` names the new home.

Runs: `npm run typecheck` clean at both commits; `npm test` 3714/3714 at the first commit and 3757/3757 at the second, 0 skipped (a Chromium is installed here, so the browser leg ran); `pytest tests/test_path_links.py` 20 passed.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -101,6 +101,37 @@ way a paper opens from OpenReview: full size, and it stays open beside the dashb
 you keep working. If the browser blocks that new tab, the PDF opens inside the dashboard
 instead; a PDF too large to show offers a download in its place.
 
+**Links in a file.** Wherever the viewer shows a file's text, the links in that text work. A
+web address opens in a new browser tab. A file path opens that file in the viewer, in place of
+the one you were reading: a relative path such as `docs/guide.md` is taken from the folder of
+the file you are reading, an absolute or `~/` path as written, on the machine of the session the
+file belongs to, and a line written after the path (`src/app.py:12`, or `src/app.py#L12`)
+scrolls the code view to that line. A Markdown file opens in its Raw view for that one open, since
+the Rendered view has no lines; your Raw/Rendered choice is unchanged. A line past the end of the
+file lands on the last line, with a notice saying so. In a Markdown file, a `[link](target)`
+follows the same two rules: a web target opens a tab, a file target opens the file (a host with a
+port, `127.0.0.1:3000` or `api.example.com:8443`, is neither, and says so). A link to a section
+of another file (`report.md#results`) opens that file at the section. A link to a section of the
+same document scrolls to it when the document has a heading or an anchor by that name
+(`<a name="install">` included), and otherwise says so when you hover it; it scrolls under every
+click, since a section of the shown file has no tab of its own. One click does one thing: a plain
+click acts in the dashboard, and a Cmd-click (Ctrl on Windows and Linux) or a middle-click opens
+the link in a browser tab of its own. Inside a file the test for a path is stricter than the one
+a chat message gets: a path links only when it has a slash and a file extension, starts on its
+own, at the start of a line or after a space, a quote, a bracket, a comma, a semicolon, an
+equals sign, a pipe or Markdown's `*` (so `$HOME/docs/a.md`, `@scope/pkg/index.js` and
+`C:/Users/x.txt` stay text), is not part of a web address, does not start with a site name
+(`www.example.org/docs/index.html`), and is not the package an `import` statement or a
+`require()` call names, whether the statement fits one line or its `from` starts the next (a
+relative import such as `./app.css` still links, and so does a path after the English word
+"from" in prose, unless that line holds nothing but `from` and the
+quoted path). After a `*` the path must be the whole emphasised text, closed by a `*` of its
+own: `*docs/a.md*` and `**./scripts/setup.sh**` link; a glob's `**/docs/a.md`, an operand's
+`w*h/img.size` and the first path in `**docs/a.md and docs/b.md**` stay text. Web addresses and
+paths found in the text wear a dotted underline that turns solid under the pointer; a Markdown
+link that names a file keeps the ordinary link look. Selecting text across a link works as
+before, and a click that lands while text is selected inside a link opens nothing.
+
 **Tags and groups.** A tag is a named, colored set of sessions; a session can be in
 several. Right-click a tab and open **Tags** to add or remove them. Tags filter every
 surface (the tag button in the strip narrows the tabs to the tags you pick), and they group
--- a/tests/test_path_links.py
+++ b/tests/test_path_links.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 """Chat file links are filesystem-VERIFIED, and shortened mentions are FIXED.
 
-The client linkifies path-shaped tokens by shape alone (render.ts CLICKABLE_PATH_RE), so a bare
+The client linkifies path-shaped tokens by shape alone (path-links.ts CLICKABLE_PATH_RE), so a bare
 `render.js` in a reply became a blue link that 404'd on click — the token resolved against the
 session's cwd, where no such file lives. The kernel is the machine with the filesystem, so at
 message-build time it resolves every shape-matched token in three tiers (exact stat; unique
@@ -192,7 +192,7 @@ class RepoListEdges(_Repo):
 
 
 class TokenizerParity(unittest.TestCase):
-    """The Python port and render.ts CLICKABLE_PATH_RE must agree on what a token IS — the map's keys
+    """The Python port and path-links.ts CLICKABLE_PATH_RE must agree on what a token IS — the map's keys
     are what the client looks up, so a tokenizer drift silently unlinks. The client side of the same
     fixture runs in chat-path-links.test.ts."""
 
--- a/ui/webview/chat-inline-preview.test.ts
+++ b/ui/webview/chat-inline-preview.test.ts
@@ -156,7 +156,7 @@ test("a verified relative path is previewable exactly like an absolute one — t
   // AND a bare filename both ride), previewKind is extension-only (relativity-blind), and both the
   // eager and the expanded render hand previewFull the SAME entry previewable carries — pin lookup
   // included, so a relative embed rides its own pin key ((pathPins || {})[p]).
-  assert.match(LINKS, /const open = isUri \? fileUriToPath\(tok\) : \(fixed \?\? tok\);/);   // the walk's open target (path-links.ts)…
+  assert.match(LINKS, /const target = isUri \? fileUriToPath\(tok\) : \(fixed \?\? tok\);\n\s*const open = opts && opts\.resolve \? opts\.resolve\(target\) : target;/);   // the walk's open target (path-links.ts; the chat passes no resolve)…
   assert.match(RENDER, /for \(const \{ el: link, open, verified \} of linkifyPathTokens\(root, pathLinks\)\) \{/);   // …is what the chat reads per hit
   assert.match(RENDER, /previewable\.push\(open\);/);
   assert.match(PREVIEW, /const ext = path\.slice\(path\.lastIndexOf\("\."\) \+ 1\)\.toLowerCase\(\);/);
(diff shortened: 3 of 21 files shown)


henrypinkard (approved):
Approved on head 6ddea73c on the maintainer's word (feature-review session).

henrypinkard (approved):
Approved on head 485c5013 on the maintainer's word (feature-review session).

henrypinkard (approved):
Approved on head 9abd445a on the maintainer's word (feature-review session), after the branch update onto main.

PR #1205, opened 2026-09-09
T290: the Log opens from the settings panel's Updates & debug section; the bottom bar keeps its few important controls

henrypinkard:
The dashboard's bottom-right bar keeps only its few important controls; the Log's opener (the red triangle with the unread count) leaves it, and the Log opens from an "Open log" button at the bottom of the settings panel's "Updates & debug" section instead. Started by romp_ui, finished by romp_feed on the same branch.

**Design points**
- The bar (`.rail-acts` in the landing HTML) no longer carries `#rail-errs`. The Log script mounts without the icon; the panel, its entries, filters and the mobile bar's `#merr` cue are untouched.
- `gear.js`: "Open log" is the last row of Updates & debug, beside "Token usage analytics". It closes the settings modal first, then posts `{romp:'openLog'}` to the shell, which routes it to the one Log opener (`window.__rompOpenErrs`), so the Log opens as it always did: a centered modal over the dimmed, unchanged dashboard (the panels rule). Shown in the web shell only; VS Code's cross-origin parent has no Log panel.
- The command palette's `log.open` and the mobile bar's `#merr` are unchanged by design.
- The bar-level error signal is gone by the user's choice: the triangle's red cue and count no longer sit on the bottom bar (the mobile bar still paints its own). So that new errors stay discoverable one click away, the unread count rides the "Open log" button instead: the shell posts it into the feed pane on every repaint of its Log and on the panel's query, and the button reads "Open log · N" with the count in the same red the triangle used.

**Tests.** Pins that named the bar's controls and the gear section's rows follow (test_error_center, test_hotkey_hints, test_kernel_disconnect_banner, test_kernel_settings_sections). New `tests/test_log_opener_moved.py`: source pins plus a served Playwright guard that the bar has no opener, the gear button opens `#rerr-back`, and the settings modal closes first. Screenshots of the bar before and after and of the gear section are in the task report.

Tier: feature

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -451,7 +451,7 @@ own a direct path; the dashboard is then a plain URL on your tailnet.
     Tunnels the extension still relays each whole view payload to the local
     window as it changes. It does not yet take the deltas the browser panes do.
     A pane that falls 16 MB behind is dropped and reconnects on its own; the
-    drop is logged in the kernel log and shows in the dashboard's bell, so a
+    drop is logged in the kernel log and shows in the Log (the settings panel's "Open log" button carries the unread count on the desktop; the phone's bottom bar reddens its bell), so a
     link that cannot keep up reads as what it is rather than as a flaky network.
 
 ### From your phone
--- a/docs/read-side.md
+++ b/docs/read-side.md
@@ -74,7 +74,7 @@ completed); the feed just paints columns. (Reflected in `docs/judges.md`.)
   by the kernel, loudly: one `ws: dropping` line in the kernel log, written where
   the drop is decided so every send path is covered, naming the pane, the
   dashboard, the backlog and the push slot whose frame tipped the budget when a
-  push did; and a row in the dashboard's bell (at most five drop rows, none older
+  push did; and a row in the dashboard's Log (at most five drop rows, none older
   than an hour, so they never crowd out a backend problem). Every close the
   browser reports for a socket that opened leaves a `wsclose` breadcrumb (code,
   reason, socket age) in `client-diag.jsonl` (rotated to `.1` at 8 MB); a socket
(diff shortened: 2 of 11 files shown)


ksarma:
Nothing found would have held this PR. It merged before this review landed, by the owner's auto-merge at dd42cc90, so everything below is a follow-up.

Reviewed inline: the Open log button is part of `GEAR_HTML`, inserted once behind `initGear`'s guard, and its onclick is bound once, so it survives re-renders. `closeSettings()` hides the modal synchronously, which acknowledges the click. The shell routes the `openLog` message to `window.__rompOpenErrs`, so the Log still opens as a fixed centered modal over the unchanged dashboard. The close and the open are two message tasks, so one undimmed frame between them is possible, but no pane shifts. `paint()` calls `tell(n)` on every Log repaint, and the shell answers `logUnseenQuery` when the panel opens, so the button's count is fresh. The count's red is the literal `#ff6b6b` that `#merr.has` uses, a status color rather than the accent. `.rs-log-n` adds no font size. When a Log entry lands on the desktop with settings closed, nothing on the bar changes; the cue is Open log · N after opening the gear. Connection loss still shows in-pane, and kinds with a feed card keep their chip, but Log-only notices have no passive desktop cue. The triangle's live-down red did not travel to the button, so a socket still down has no desktop cue once its entry is read. The body records the removal as the user's choice. Attributions paraphrase, the commit trail matches the two-session body, and the privacy scan of the 241 added lines is clean.

Follow-ups:

- `tests/test_error_center.py:64`: the executed harness still stubs a `rail-errs` element, so CI covers the desktop shape of `_LANDING_ERRS_JS` (icon null) only by source pins. The served guard in `test_log_opener_moved.py` skips in CI because the Python job never runs `npm ci`. A second harness run with `rail-errs` absent, asserting `__rompOpenErrs` is set and `logUnseen` posts 1 then 0, closes the gap in about ten lines.
- `docs/guide.md:454`: the new sentence tells phone readers the bottom bar reddens its bell. The phone's Log cue is the warning triangle (`#merr`), and the phone bar's bell is the push-notification toggle. The line is also 185 characters in a paragraph wrapped at 78.
- `kernel/kernel.py:45657` and `tests/test_error_center.py:5`: comments still describe the opener as a bell in the bottom bar's action cluster.

Tests: pytest on the four re-pinned files plus `test_log_opener_moved.py`, 44 passed. The served Playwright leg ran here against a hermetic kernel on a free port and passed (5 passed, no skips). The same test files at the merge base: 7 fail, 37 pass. `npm test` from `vscode-extension`: 3744 passed, css-census re-pinned at 13 for `gear.css`. CI green at dd42cc90.

Review run by the maintainer automated review session; verified against head dd42cc90.

PR #1206, opened 2026-09-09
Figure previews: a link failure heals on the reconnect event only, and an attempt never changes the box's layout (T291)

henrypinkard:
The chat transcript flickered on a remote session viewed through a failing relay (the user, 2026-09-09, on the laptop: the whole visible transcript shifted by about a hundred pixels and back, many times a second, for the length of a streaming turn). The scroll journal's `unitchange` rows named it: two assistant turns changing height by five and three text lines in step, up to ten times a second, with the thread's height flapping by their sum and Chrome's anchoring shifting the reader's text by the difference. Both turns hold a figure preview whose fetch fails through the relay.

**Root cause** (`ui/webview/preview.ts`, `previewFull`). A fetch failure whose text names the LINK ("tunnel to … is not answering", "no attached host", "re-dialing") was classed transient and retried for free on every kernel message (render.ts calls `retryFailedPreviews()` on each incoming message; a streaming session pushes per SDK event). Each attempt swapped the box's content: the managed attempt painted a fresh one-line "fetching…" wait, the instant failure painted a fresh multi-line "<reason> — retrying · tap to retry now" note, and the next message started over. Two layouts, alternating at the push rate, under a reader scrolled up.

**Fix, two parts, no new heuristic.**
- A link failure heals on the reconnect-class event only: `romp:wsup` when this page's kernel socket comes back, `hostUp` when the federated tunnel does (the tunnel poll already dispatches it on the down-to-up transition). It registers in `settledPreviews`, never in the per-message `failedPreviews`; the budget stays untouched as before, the tap still retries at once. A real verdict (a true not-found, a transfer that died with no progress) keeps the bounded per-message budget it always had.
- An attempt never changes the box's layout: one wait element, one swirl, one note, reused from attempt to attempt (`waitBox`); only the note's words change, and only when they do (`setNote`). A retry from a failure note no longer shrinks the box to one line and grows it back; a settled chip keeps its shape through its one new-evidence heal (the chip carries the words while the attempt runs).

**Pinned** (`ui/webview/preview-retry-pace.test.ts`, executed over a minimal fake DOM and a stubbed fetch, driving the entry points the way render.ts calls them): after a link failure, sixty kernel messages make no attempt and leave the box byte-identical (no DOM change, so no height change and no `unitchange` row); one reconnect event makes exactly one attempt, and the box's elements and words are unchanged while it runs and after the same failure; a later reconnect with the link up lands the picture; a non-link failure spends its bounded budget on the per-message heal with the wait box's elements unchanged across attempts, then settles on the tap chip, which pushes do not move either. The inline-preview pins follow the new branch and the reused note.

**From the adversarial review, folded in before arming.** The settled chip carries the words only for its one new-evidence heal; a tap or a reconnect heal, which re-arm the budget, bring the loading persona back (swirl and note, inside the same wait element), so the chip stays the give-up state alone. The link note says what will happen ("retries when the link is back · tap to retry now") instead of promising a retry it does not schedule. A remote kernel's restart reopens the relay socket and fires neither `romp:wsup` nor `hostUp`, so the relay-reopen event (`romp:hostRelayUp`) now drains the settled previews too. The PDF card's probe treats a 502 as the link's verdict and a markdown image on a relay URL waits for the reconnect-class heal likewise, so no path re-fetches through a dead relay on every push. Tests are text-sensitive per attempt, isolated per test, and pin the tap's persona and the relay-reopen heal.

**Residual, stated.** A relay failure while the tunnel row never leaves "up" (a request that timed out mid-transfer and the supervisor's next poll answered) has no reconnect-class event; the box waits for a tap, the next reconnect, or the turn's re-render. A kernel-side recovery counter on the tunnel row would close that gap; it is a separate change.

### DIFF
--- a/ui/webview/chat-inline-preview.test.ts
+++ b/ui/webview/chat-inline-preview.test.ts
@@ -61,8 +61,10 @@ test("the failure chip narrates what happens next, escalates on repeat, and a re
   // while auto-retries remain the box KEEPS its loading persona (swirl + note, whole box tappable);
   // the ⚠ chip is the GIVE-UP state only (the user 2026-08-16, third report: state bouncing between
   // "trying" and "unavailable" on every retry cycle read as impatient even when it eventually loaded)
-  assert.match(pf, /if \(autoRetries > 0 \|\| transient\) \{\s*\n\s*if \(!transient\) autoRetries--;\s*\n\s*failedPreviews\.set\(box, \(\) => build\(true\)\);/);
-  assert.match(pf, /\+ " — retrying · tap to retry now";/);
+  // a LINK failure registers for the reconnect-class heal only, a real verdict spends the budget on the per-message heal (T291)
+  assert.match(pf, /if \(autoRetries > 0 \|\| transient\) \{[\s\S]*?if \(transient\) settledPreviews\.set\(box, \(\) => build\(true\)\);\s*\n\s*else \{ autoRetries--; failedPreviews\.set\(box, \(\) => build\(true\)\); \}/);
+  assert.match(pf, /\+ \(transient \? " — retries when the link is back · tap to retry now" : " — retrying · tap to retry now"\)\);/,   // set on the reused note; the link case says what will happen (T291)
+    "the note says what happens next: a retry on the link's return, or the bounded retrying");
   assert.match(pf, /wait\.onclick = \(ev\) => \{ ev\.stopPropagation\(\); autoRetries = 3; ackTap\(ev\); build\(true\); \};/,
     "the whole retrying box is the tap target — a tap re-arms persistence and acknowledges");
   assert.match(pf, /"⚠ preview unavailable"\)\)\s*\n\s*\+ " — tap to retry";/,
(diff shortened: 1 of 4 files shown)


ksarma:
Merged on green as shipped. This is the post-merge record: one narrow follow-up and two nits.

Reviewed inline: the link-failure heal is keyed on exactly the three reconnect-class events and nothing else. Only `refreshSettledPreviews` drains `settledPreviews`. It runs on `romp:wsup` (`render.ts:1724`, fired by the page shim on a reconnect only). It runs on the `hostUp` frame (`render.ts:14918`, dispatched by the tunnel poll on a host's down-to-up transition). New here, it runs on `romp:hostRelayUp` (`render.ts:13440`, the relay socket's open, which covers a remote kernel restart and a first attach that `hostUp` misses). The per-message drain never touches it. No timer or counter was added. A parked link failure cannot get stuck. It stays tappable, and the turn's re-render mints a fresh box. Any of the three events drains every parked box, so a `hostUp` for another host costs one extra attempt rather than a miss. The stated residual is accurate. Layout stability holds on a reading of the code. `waitBox` reuses the wait, swirl and note and creates each once. `setNote` writes only when the text differs. The settled chip keeps its shape through its one new-evidence heal. The transient regex matches the kernel's two exact link texts. A false positive parks a figure until reconnect or tap. A false negative falls back to the bounded budget. The swirl is the romp loader reused. No CSS or font size changed. The attribution paraphrases.

- **Follow-up**, `ui/webview/preview.ts:545`: the transient branch's `settledPreviews.set` replaces the chip's re-arming entry from line 587. Sequence: a chip settles, its new-evidence heal fails with a link error, and the reconnect heal then runs at `autoRetries=1` with the chip persona. The result is no swirl, a note still promising a retry while the fetch is in flight, and one attempt instead of three. Confirmed by execution at head and base. It is a regression, contrary to the body and the comment at 498-500. A `settledPreviews.has(box)` guard plus a pin closes it.
- **Nit**, `ui/webview/preview.ts:508`: `waitBox` keeps the literal swirl path where line 97 uses `mediaSrc()`, so the retry swirl does not resolve in the VS Code webview. This is pre-existing and was consolidated here.
- **Nit**: the commit message omits the `hostRelayUp` wiring, the PDF 502 route and the markdown image change that the body describes.

Red-first verified. At the base with the PR's test files, preview-retry-pace fails 3/3. It still fails 3/3 with the wording assertion removed, on the second fetch per push and the fetching swap. chat-inline-preview fails 2 of 14 at the base. At head both pass, npm test 3747/3747, typecheck clean. CI: all nine checks green at head.

Review run by the maintainer automated review session; verified against head b892e570.

PR #1207, opened 2026-09-09
Spend: the connect-time scan for the last cost-state record reads a long transcript line in linear time

ksarma:
Follows #1200 (merged), which seeds the spend watermarks at connect from the transcript's last cost-state record. The review record on that PR found the reader's cost quadratic in the length of a transcript line and named two docstring corrections; this PR makes those changes.

## Symptom

A resumed session's connect gets slower with the length of the longest line in the scanned part of its transcript, beyond the cost of the bytes scanned. Within the 8 MB bound the reader walks one long line in about a hundred 64 KB steps and copies the line so far at each step, on the event loop: a 7 MB line costs about a hundred growing copies and holds three copies of the line at its peak (21 MB measured). A transcript line is one JSON record, and a tool result can make one several MB long.

## Cause

`last_cost_state` reads backwards in 64 KB chunks and carries the fragment a chunk begins with into the earlier chunk by concatenation (`f.read(step) + carry`). The concatenation copies the whole carried fragment on every step, so a line spanning n chunks costs n copies of growing length, and the `partition` on the result copies the line once more.

## Fix

The reader keeps the pieces of the carried line in a list and joins them once, when it reaches the newline before the line, so each byte read is copied a fixed number of times. A line longer than 4 MB (`max_line`, a new keyword with that default; a record is under 1 KB) is not a record: the pieces of it past the cap are dropped as they arrive and the line is skipped without a join, so on such a line the scan holds at most the cap plus a chunk (4.07 MB on the same 7 MB line; 4.01 MB on a line one byte past the cap, whose pieces all fit under the cap and are still not joined). A record-shaped line just under the cap is joined and parsed as before and held about three times over at that moment, bounded by the cap. The state is a list of pieces and one running length: past the cap the length keeps counting while the pieces are dropped, and every join checks the length, so no separate flag is needed. The range of bytes read, the 8 MB `scan_bytes` bound and its floor are unchanged, and so are the answers: the last valid record wins, an unusable one is skipped, the line the bound cuts through is dropped as before, and a line longer than the cap is skipped wherever it sits. `_turn_usage` and the settle are untouched. The module's other backwards reader, `_lines_from_end`, keeps its per-block concatenation: its caller reads the last line and stops, so it is left alone here.

The two docstring corrections: `last_cost_state` names the second way a record reaches a resumed file (a transcript the interactive CLI wrote, whose writer runs from startup, that romp later resumes, with the same under-count), and the `_last_cost_total` comment says the connect seeds the watermark from that record rather than resetting it to zero.

## Tests

All in `tests/test_sdk_backend.py`, class `SpendRecord`, over the real `last_cost_state`. Three of them route the module's `open` through a stand-in file object that records each read and hands the bytes back; the patch is undone at tearDown.

- `test_last_cost_state_reassembles_a_long_line_and_finds_the_record_before_it`: a record, then a 3 MB line with no newline, then a trailer. The record is found; the reads sum to the file size, one read per 64 KB and none longer. Passes before the change too: it pins that the range read does not change.
- `test_the_bound_reads_the_same_tail_and_drops_the_line_it_cuts_through`: with `scan_bytes` at three chunks, a record followed by about 290 KB reads as absent and the reads sum to exactly `scan_bytes`; a file whose floor lands on a record's first byte reads as absent (the cut line is a fragment) and, with the bound one byte larger, yields that record. Passes before the change too: a guard on the bound, since the leftover pieces at the floor are what this change drops where the old reader dropped its carry.
- `test_a_line_longer_than_the_cap_is_not_a_record_and_the_earlier_record_wins`: a record-shaped line of 5 MB is skipped and the record before it wins; alone, or unterminated at the file's head, it yields None with no exception; a record after it is found. A record-shaped line of exactly 4 MB between two other lines (about 65 chunks) is found with its total, which pins the order the pieces are joined in (out of order it would not parse), and one byte longer it is skipped for the record before it. With `max_line` set to the length of a record-shaped line the scan carries in three pieces, that line is found, and at one byte less the earlier record wins; the same pair for a line whole inside a chunk and for the head piece alone in the file. Fails before the change on `7.0 != 1.0`: the old reader joins the whole line and takes it.
- `test_the_scan_copies_a_line_once_and_holds_at_most_the_cap_however_long_the_line`: over a record and a 7 MB line, the stand-in hands back a bytes subclass that records any concatenation onto it, and `tracemalloc` measures the process's growth over the scan, from what it held as the scan began (so a tracer already running, `PYTHONTRACEMALLOC`, adds nothing). The new reader records no concatenation and peaks under 6 MB (4.07 MB measured); a line one byte past the cap peaks the same (4.01 MB), which pins that the join is refused rather than made and discarded. The old reader fails on 113 recorded concatenations growing from 65536 bytes, and peaks at 21 MB. No timing assertion.

The existing chunk-edge test keeps passing. `tests/test_sdk_backend.py`: 245 passed, 35 skipped. Full suite (`pytest -q -n 8 tests/`): 10202 passed, 61 skipped, 1671 subtests passed, no failures, in 85 s.

Tier: fix

### DIFF
--- a/kernel/sdk_backend.py
+++ b/kernel/sdk_backend.py
@@ -821,7 +821,7 @@ def cost_state_watermarks(o):
     return {"total": float(total), "tokens": tokens}
 
 
-def last_cost_state(path, scan_bytes: int = 8 << 20):
+def last_cost_state(path, scan_bytes: int = 8 << 20, max_line: int = 4 << 20):
     """The resumed transcript's LAST `cost-state` record, as the watermarks a CLI that restores it would
     hold (cost_state_watermarks), or None when the file has no such record or cannot be read.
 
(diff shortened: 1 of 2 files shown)


ksarma:
Post-merge record: no defect found on main, and nothing to fold. The PR landed as 8d3cd1a0 by the owner before this review ran.

Reviewed inline: the diff touches only `last_cost_state` and two comment passages; no process-signalling code moved or ran. The new loop splits each 64 KB chunk on newlines. It keeps the carried line's pieces in a list in file order and joins them once at the newline that opens the line. Copies are therefore linear in the line. The `max_line` cap drops pieces past 4 MB as they arrive and skips the line unjoined. A record is a few hundred bytes, so no real record is affected, and 4 MB is a sound default. The range read, the 8 MB `scan_bytes` bound and its floor are unchanged. A differential probe of the base and head readers found no answer change except the declared over-cap skip. It covered the boundary cases (empty file, head piece alone, unterminated last line, CRLF, blank lines, newlines on either side of a chunk edge, a record on the scan floor, multi-MB lines) plus several thousand randomized files. An exactly 4 MB record-shaped line is found by both; one byte longer is skipped by the new reader. Measured peak on a 7 MB line: 21 MB at base, 4.2 MB at head. The docstring corrections match the #1200 record's asks. The fix label fits, body and commit match the diff, and the privacy scan of the added lines is clean.

Follow-up, record only, no code change: the body's reason for leaving `_lines_from_end` alone (its caller reads one line and stops) does not hold for `last_state_value` (`kernel/sdk_backend.py:1585`). That function iterates through overlay records. The reader stays fine because `states/<sid>.jsonl` lines are short, so its concatenation is quadratic only in a line that never spans blocks.

Tests:
- Red-first at base 3f06288e with the head test file: `test_a_line_longer_than_the_cap_is_not_a_record_and_the_earlier_record_wins` fails on 7.0 != 1.0. `test_the_scan_copies_a_line_once_and_holds_at_most_the_cap_however_long_the_line` fails on 113 recorded concatenations. The two guard tests pass at base as declared.
- At head: SpendRecord 30 passed; `tests/test_sdk_backend.py` 245 passed, 35 skipped, stable across serial and xdist repeats.
- CI at head: all eleven checks green.

Review run by the maintainer automated review session; verified against head 39f2cf1d.

PR #1208, opened 2026-09-09
Comments: a remote session's comment is named by the owning kernel, never by the viewer's decorated prefill

henrypinkard:
## What

A comment made on a REMOTE session's transcript (a session of another attached kernel, viewed through federation) could be refused with the create door's toast: `"<host>-<session>-comment-N" is a comment thread of <session> — open it from that session's comments, or pick another name`. The user saw it after typing what they took to be a follow-up in a thread, on a laptop viewing a devbox session.

Root cause (convicted in code and in the owning kernel's comments store): the comment dialog prefilled its name box from the viewer's DECORATED session name (`host:name`, which federation prefixes on every session-bearing frame) sanitized to `host-name-comment-N`, with N = the threads the VIEWER knew + 1, and sent that prefill as if the user had chosen it. The kernel treats any non-empty name as the user's explicit choice: it claims it verbatim (no host-label stripping, no numbering past a taken name) and refuses when taken. So the first comment stored a thread literally named with the viewer's host label, and the next dialog (the viewer's count regresses when a thread is deleted, and lags when a create's ack is lost, or after the dashboard reloads) prefilled the same name and was refused. The routing was never the defect: both `commentCreate` and `commentReply` carry the parent's host-prefixed sid and reach the owning kernel by it.

## The fix

- **Client (`ui/webview/comment-name.ts`, new; `ui/webview/render.ts`):** the prefill is built from the BARE session name (`host-prefix.ts`, the rename dialog's rule since 2026-08-02) for the comment, break-out and fork dialogs (the fork's name is required, so its labelled prefill was always sent: the same defect one dialog over, found in review); the comment dialog remembers what it prefilled, and an UNTOUCHED prefill is sent as `""`, so the owning kernel picks its own default (bare name, next free number). A typed name is still the user's choice; its draft now dies with the create's ack even when the popover closed first (it resurfaced as a taken name on the next comment on that message). The box is worded as a suggestion.
- **Kernel (`kernel/kernel.py`):** a comment create is idempotent on its identity (parent sid, anchor uuid, passage, text). Both doors that can run one create (the WebSocket handler on a server thread, the pusher's parked retry) reserve the identity under one lock before creating: a repeat of a completed create is answered with the SAME open thread's ack (and a kernel log line says so); a copy arriving while the other door holds the identity gets the typed transient nack, so the popover keeps its mark and adopts the thread from the pusher's ack; a lag-parked create is parked once. A repeat whose thread was resolved, merged or promoted is a NEW comment, never a swallowed one. This closes the second route into the same refusal the trace found: a create parked for transcript lag is retried by both the pusher and the client, and before this the second copy either collided on its explicit name or, with the name now left to the kernel, would have minted a duplicate thread. Every thread-name refusal at the create, break-out and rename doors now also writes a kernel log line (before, only a toast the viewer could miss, and the refusing kernel's log held no trace).

## Tests (red first)

- `ui/webview/comment-name.test.ts`: the pure rules (bare name, comment/break-out/fork prefills, untouched → empty) and pins that all three dialogs use them, that no decorated-name stem remains anywhere in render.ts, and that a typed name's draft dies with the ack.
- `ui/webview/comment-remote-routing.test.ts`: `commentReply` and `commentCreate` for a remote session route to the owning host by sid with tid and name untouched (the suspected routing, pinned sound).
- `tests/test_comment_name_refusal_log.py`: a create under a taken name is refused aloud AND logged; an empty name takes the kernel's default past the taken one; the break-out and rename doors log their name refusals.
- `tests/test_comment_create_idempotent.py`: the same create twice yields one thread and two acks naming it; a different comment on the same passage is a second thread; a lag-parked create is parked once; the client's re-post after a parked create landed is answered with the same thread; a re-post arriving on another thread WHILE the pusher's retry is creating does not mint a second thread (the race, red before the reservation); a repeat after the thread was resolved is a new comment; an answered repeat is logged.
- `tests/test_remote_comment_name_served.py` (both kernels run with the shared postal bus off and clean up on every exit path): a hub kernel with a second hermetic kernel checked in as host TESTHOST (the mobile handshake, no ssh); the real /chat page opens the remote session from its host-prefixed tab, selects a passage, opens Comment, and the send is captured at the socket: the prefill reads `web-comment-1` (main: `TESTHOST-web-comment-1`), and the frame carries an empty name on the `/remote/TESTHOST/ws` relay with the parent's bare sid.

Existing pins on the dialog's prefill line were updated to the new shape. Synthetic fixtures only.

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -13468,6 +13468,67 @@ def _comment_markers(sid):
 ANCHOR_LAG_ERR = "that message isn't in the transcript yet; try again in a moment"
 _parked_creates = []                       # [{sid,uuid,exact,text,name,model,effort,fast,color,tries}]
 _PARK_MAX_TRIES = 30                       # pusher cycles (~15-90s) — past this the record isn't coming
+# A create's IDENTITY (T289): (parent sid, anchor uuid, passage, text) -> the thread it made. A lag-parked
+# create is retried by BOTH the pusher (above) and the client (its frame-keyed re-post), and a popover
+# whose ack was lost sends its create again; the second copy used to collide on its explicit name and
+# come back as the create door's refusal toast (the user 2026-09-09, on a remote session), or — with the
+# name now left to the kernel's default — would mint a SECOND thread for one comment. The memo answers a
+# repeat with the SAME thread's ack, and a parked copy is parked once. Bounded (oldest out), in memory:
+# a kernel restart forgets it, and the client's re-post after one creates exactly once.
+_recent_creates = {}                       # (sid, uuid, exact, text) -> tid
+_RECENT_CREATES_MAX = 256
+# The two doors that can run one create — the WS handler on a server thread and _retry_parked_creates on
+# the pusher — reserve the identity under ONE lock before creating (review, 2026-09-09): the pusher sends
+# the chat frame before it retries parked creates, the client re-posts on that very frame, and with no
+# reservation both copies ran _comment_create before either was noted, minting twins (or, under an
+# explicit name, the create door's refusal the branch set out to remove). RLock: the helpers below take
+# it and so do their callers' compound steps.
+_create_lock = threading.RLock()
+_inflight_creates = set()                  # keys whose _comment_create is running right now, either door
+
+
+def _create_key(sid, uuid, exact, text):
+    return (str(sid), str(uuid), str(exact), str(text))
+
+
+def _note_create(key, tid):
+    with _create_lock:
+        _recent_creates[key] = tid
+        while len(_recent_creates) > _RECENT_CREATES_MAX:
+            _recent_creates.pop(next(iter(_recent_creates)), None)
+
+
+def _repeat_create_tid(key):
+    """The OPEN thread an identical create already made, else None. A resolved, merged or promoted thread
+    cannot take a message, so a same-worded comment after that is a new comment, never a swallowed repeat
+    (review, 2026-09-09); a deleted one is gone from the store and the memo forgets it."""
+    with _create_lock:
+        tid = _recent_creates.get(key)
+        row = _comment_thread(key[0], tid) if tid else None
+        if row and (row.get("status") or "open") == "open":
+            return tid
+        _recent_creates.pop(key, None)
+        return None
+
+
+def _reserve_create(key):
+    """Claim one create's identity for the caller, atomically: ("repeat", tid) when the same comment was
+    already made and its thread is open; ("busy", None) when another door is creating or holding it parked
+    right now; ("free", None) after marking it in flight — the caller must _release_create in a finally."""
+    with _create_lock:
+        again = _repeat_create_tid(key)
+        if again:
+            return "repeat", again
+        if key in _inflight_creates or any(_create_key(pk["sid"], pk["uuid"], pk["exact"], pk["text"]) == key
+                                           for pk in _parked_creates):
+            return "busy", None
+        _inflight_creates.add(key)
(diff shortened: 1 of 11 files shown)


ksarma:
Post-merge record: #1208 landed as shipped under the owner's policy, since an admin's fix merges on green. Nothing in this review would have held it. Two follow-ups stand on main.

Reviewed inline: the comment, break-out and fork dialogs all build their prefill from the bare session name through `comment-name.ts`. An untouched prefill goes out as an empty name, so the owning kernel numbers its own default. No sanitizer stem of the decorated name remains in `render.ts`; the new pin and a grep agree. The create-identity reservation is released on every exit path. Both doors wrap `_comment_create` in try/finally. The pusher's two `continue` statements before the reservation never mark the key in flight. The repeat check reads the thread's current status from the store under the lock. A repeat after resolve, merge, promote or delete is therefore a new thread. The busy nack is typed transient. The client keeps its mark and re-posts on the next frame under a bounded cap. The pusher's ack reaches every chat client, so no path leaves the popover stuck. The body matches the diff, and the fixtures are synthetic. The 1006 added lines have no private-string, hostname or home-path hits. The fix label fits the client half outright. The kernel half (memo, lock, in-flight set) is new mechanism in service of the same refusal, which changes no gate for the owner. The two-kernel served test was reviewed by reading only: it boots real kernels, which the review machine does not run.

Follow-ups on main, for the fixing session:

- `kernel/kernel.py:13490` and `:13501` (follow-up). A deliberate second comment with the same words on the same passage collapses into the first thread. The identity omits the typed name, model and effort. The memo answers any identical create on an open thread until 256 later creates or a restart. The viewer sees the old popover and no toast; only the kernel log records it. Reproduction: post "fix this" on a passage and receive the ack. Post "fix this" on the same passage under a different name. Result: one thread, two acks, the second name nowhere on disk. The body states the identity rule but not this consequence. A per-gesture create id, or a memo bounded to the in-flight and parked lifecycle, keys on the event instead of the text.
- `tests/test_remote_comment_name_served.py:132-138` (follow-up). Each hermetic kernel's env is the caller's whole environment minus three keys. CI skips this test for lack of a browser. On a developer box the kernels inherit `ROMP_MANAGER_PID` and exit if the live manager restarts mid-test. The boot-time postal ensure pings the default port 25302 in any peers mode. Where that port is silent, it spawns a detached bus that `tearDownClass` never reaps. The comment at line 137 and the body's claim that the kernels run with the postal bus off overstate what `ROMP_POSTAL_PEERS=0` does. The sibling served fixtures share the pattern. A whitelisted env plus `ROMP_POSTAL_PORT` with `ROMP_POSTAL_CLIENT_ONLY=1` covers all of them. Under pytest, conftest already scrubs `ROMP_SUPERVISED`, so no systemd probe runs.

Nit: the commit message names the comment and break-out dialogs, two updated pins and a repeat after resolve or promote. The diff also rewires the fork dialog, re-pins three test files and treats a merged thread as closed. The body has all three right.

Tests: at head 91605c75, pytest 11/11, npm test 3758/3758, typecheck clean. CI: all nine checks green. At the base bce52234 with the PR's test files checked out, pytest ran 8 failed / 3 passed. The eight fail for the body's reasons. The three passing are negative controls the body claims as pins. The vitest bundle failed to resolve `./comment-name`. With that file set aside, the other four gave 94/97, with one pin red each in comments, fork-session and comment-reply-landed. comment-remote-routing passes at base, as the body says.

Review run by the maintainer automated review session; verified against head 91605c75.

PR #1209, opened 2026-09-09
Tunnel rows carry a recovery counter, and the dashboard treats a bump as the host coming back (T291b)

henrypinkard:
Closes the residual the preview fix left stated: a figure preview whose fetch failed through a link that never left "up" (a request that timed out mid-transfer while the supervisor's next poll answered, or a re-dial that came back between two dashboard polls) had no reconnect-class event to heal on once link failures stopped riding the per-push retry.

**Kernel.** The supervisor pass notes a recovery on each tunnel row: `_note_recovery(r, st, had_miss)` bumps the row's `upSeq` once when a row that had missed polls (read before the pass's own bookkeeping), or was not up, is up at the end of the pass. A steady healthy row never bumps; the counter is per process and not saved (like the poll run counters). `/tunnels` serves `upSeq` beside `lastOk`.

**Dashboard.** Federation's tunnel poll remembers each host's counter; a change while the row reads up is treated as that host coming back and joins the same `hostUp` dispatch the down-to-up transition uses, so render.ts drains the settled previews once (they make their one attempt) and un-parks the path-image chips. One event per bump, none on a first observation or a steady poll; no per-push retry returns.

**Tests.** The kernel bump rule (`tests/test_kernel_tunnel_truth.py`: a steady row never bumps, a miss-then-answer bumps once and only once, a not-up row coming up bumps, a row that is not up never bumps, the row serves it and the store never saves it, the pass reads the misses before its bookkeeping and notes against the previous status); the dashboard poll (`federation-reconnect.test.ts`, executed: a first observation fires nothing, a steady poll nothing, a bump one `hostUp`, the same counter again nothing); the preview (`preview-retry-pace.test.ts`, executed: a link failure with the row staying up makes no attempt under thirty pushes and exactly one when the bump's `hostUp` drains the settled previews, with the box unmoved).

**From the adversarial review, folded in before arming.** The bump requires a pass whose poll ANSWERED (after the pass's bookkeeping, misses is zero): a second silent poll under the stale bound keeps the row's "up" without anyone answering and used to read as a recovery, which would have fired `hostUp` into a link the supervisor already knew was dying. A miss a data path preloaded (`_demand_redial`'s timeout) mints one bump until a real silent poll or a status change re-arms it, so a request that keeps stalling on a healthy link cannot mint its own recovery, retry, stall and mint again forever; a real silent poll leaves its own mark. A checked-in peer's timeout preloads the miss and wakes the supervisor like an ssh row's, so a mobile's link that stayed "up" heals its parked previews too. The kernel tests cover each: the second silent pass, the demand bound and its two re-arms, the checked-in peer.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -16628,7 +16628,40 @@ def _note_poll(r, answered):
         return False
     r["ok_polls"] = 0
     r["misses"] = int(r.get("misses") or 0) + 1
+    r["_poll_miss"] = True                           # a REAL silent poll: the recovery counter's evidence (T291b)
     return r["misses"] >= STALE_MISSES
+
+
+def _note_recovery(r, st):
+    """The row's RECOVERY counter (T291b, the user 2026-09-09): bumped once when a pass that ANSWERED finds
+    the row had missed polls, or was not up — the event a dashboard needs when a data-path request failed
+    through a link whose status never left "up" (a request that timed out mid-transfer, a re-dial the
+    supervisor's next poll answered): /tunnels serves it, and the dashboard's poll treats a change as the
+    host coming back (hostUp), so the previews parked on that link make their one attempt. Called once per
+    pass after `st` is final and before the pass overwrites the row's status.
+    Two guards from the review: (1) only a pass whose poll answered counts — after this pass's bookkeeping
+    that is exactly misses == 0 (a second silent poll under the stale bound keeps the row's "up" without
+    anyone answering, and must not read as a recovery); (2) a miss a DATA PATH preloaded (_demand_redial's
+    timeout) mints one bump until a real silent poll or a status change re-arms it, so a request that keeps
+    stalling on a healthy link cannot mint its own recovery, retry, stall, and mint again forever. A steady
+    healthy row never bumps; a fresh boot starts at 0 (the counter describes THIS process, like the poll run
+    counters, and is not saved). Returns whether it bumped."""
+    answered = st == "up" and int(r.get("misses") or 0) == 0
+    if not answered:
+        if st != "up":
+            r.pop("_demand_bumped", None)            # a status change re-arms the demand path
+        return False
+    real_miss = bool(r.pop("_poll_miss", False))
+    demand_miss = bool(r.pop("_demand_miss", False))
+    had_miss = real_miss or (demand_miss and not r.get("_demand_bumped"))
+    bumped = had_miss or r.get("status") != "up"
+    if bumped:
+        r["upSeq"] = int(r.get("upSeq") or 0) + 1
+    if bumped and demand_miss and not real_miss:
+        r["_demand_bumped"] = True                   # one demand-sourced bump until the link shows a real miss
+    if real_miss:
+        r.pop("_demand_bumped", None)
+    return bumped
 
 
 def _tunnel_established(r):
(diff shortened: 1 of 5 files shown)


ksarma:
Post-merge record. The owner merged on green under his own policy, and the review confirms the fix as shipped. Nothing here calls for action on the PR. Two follow-ups and one nit go on the record for the fixing session.

Reviewed inline: the kernel bump keys on the answered-poll event and on state, not on elapsed time. `_note_recovery(r, st)` runs after `_note_poll` and before the status write. It bumps only when this pass answered (misses is zero after the bookkeeping) and one of three conditions holds: a silent pass left the `_poll_miss` mark, `_demand_redial` preloaded a `_demand_miss` not yet spent, or the previous status was not up. The demand latch cannot loop. A preloaded miss mints one bump, and a second stall on the same healthy link mints none. A real silent poll or any not-up status re-arms it, so a genuine recovery after a demand bump still bumps. Preload, one real silent poll, stale teardown, redial traced to exactly one bump. On the dashboard, a first observation records without firing, a repeated counter fires nothing, and a counter change while the row reads down is absorbed. After a kernel restart the rows load as down with the counter at zero, so the reset is seen while down. The later down-to-up fires one hostUp through the existing status path, deduped against the counter path. A restart plus a redial completing between two polls fires one hostUp, a real recovery. hostUp runs refreshSettledPreviews and healPathImgs once per event, and neither touches a loaded preview, so nothing ping-pongs. The privacy scan of the 207 added lines is clean; the one synthetic placeholder path follows the file's existing fixture convention. The fix label fits: the PR closes a stated residual of T291 with tests red at the base, and the counter and the /tunnels field are the vehicle, not a new capability.

Defects on main and in the record, for the fixing session:

- `kernel/kernel.py:21331` (follow-up). During a Start hold (`booting` set by `_start_remote`), the pass calls `_note_recovery` but skips the status write. Every answered pass inside the hold therefore bumps `upSeq` again. Reproduced with the pass's order: three answered passes with booting=True gave 1, 2, 3, then 4 when the hold cleared. The dashboard is immune: the row is served as `starting`, and the counter path requires the row to read up, so the user sees one hostUp per Start. The docstring's one-bump-per-recovery contract does not hold. Fix shape: no bump while `r.get("booting")`, plus a kernel test that sets booting across two answered passes.
- PR body (follow-up, record only). The body names `_note_recovery(r, st, had_miss)` and twice says the misses are read before the pass's bookkeeping. The merged function takes two arguments, runs after `_note_poll`, and reads the `_poll_miss` mark a silent pass leaves. The cited ordering test pins the call before the status write, not a pre-bookkeeping read. Docstring, commit message and behavior are accurate.
- `ui/webview/federation-reconnect.test.ts:451` (nit). The new test drives only status up. Untested: the clause that absorbs a counter change while the row reads down (the kernel-restart path), and the map cleanup for a host leaving the poll. Both behave as intended by reading.

Test evidence: red-first verified at base bce52234 with the PR's three test files. `tests/test_kernel_tunnel_truth.py`: 7 failed, 22 passed (six RecoveryCounter tests on a missing `_note_recovery`, one on `upSeq` absent from `_NOT_SAVED`). The two node test files: 2 failed, 25 passed. Federation-reconnect test 20 fails behaviorally with no hostUp on a bump. Preview-retry-pace test 26 fails only on its source-text pin; its behavioral assertions already pass at base. At head 38d1c581: 29 passed and 27 passed. CI at head: all nine checks SUCCESS.

Review run by the maintainer automated review session; verified against head 38d1c581.

PR #1210, opened 2026-09-09
A refused clear, drop or undo names goals/<sid>.json in its dialog, not the absolute state path

ksarma:
## Symptom

The err dialog that answers a clear, a sub-goal drop or an undo refused by an unreadable or unwritable goals file shows that file's absolute path under the home directory. The quoted fault reads `[Errno 5] Input/output error: '<state root>/goals/<sid>.json'`, with the state root spelled out in full, on whatever screen the dashboard is on; a federated dashboard shows it on another machine's screen.

## Cause

The four `skipped[sid] = str(fault)` sites (the load fault and the save fault in `_mark_nodes_cleared` and in `_restore_goal_archive`) hand `str(fault)` to `_gesture_store_refusal`, which interpolates it into the dialog's text. The fault is the OSError the loaders return; its filename is the path the store was read from, built from `jd.GOALDIR`, so it is absolute. `_oserror_text` already states the rule for the frames it serves (an absolute state path has no business in a pane), but nothing applied it to the gesture refusals.

## Fix

A new helper in kernel/kernel.py, `_store_fault_copy(fault)`, returns `str(fault)` (the type name when the fault has no text) with `str(jd.STATE) + os.sep` removed wherever it occurs, and the four sites use it. The dialog now reads `[Errno 5] Input/output error: 'goals/<sid>.json'`: the errno text and the file name stay, so the user still learns what refused and which session's store. A fault that names two paths, a failed publish's rename of the temp file onto its destination, loses the root from both. The strip is textual; every loader builds its path from `jd.GOALDIR`, so that is the prefix a fault carries. The judge-errors row `_file_store_fault` writes is unchanged and keeps the whole path, which is diagnostic there.

Alternative considered: calling the existing `_oserror_text(fault)` at the four sites also removes the path, but drops the file name with it (errno and strerror only). That suits a frame that dedupes on its text; a dialog about one session's file needs the name, or a user with several sessions cannot tell which store refused. The helper keeps it.

## Tests

All in tests/test_goal_store_fault_boundary.py, class GestureRefusal. Six of them take a gesture through the real dispatcher with the store faulting and check the dialog's text; the seventh calls the helper.

- `_assert_clear_refusal`, the check shared by `test_a_card_clear_the_fault_skipped_answers_the_socket_that_asked`, `test_a_clear_all_answers_once_per_session_and_reads_true_for_several_cards` and `test_a_fault_at_the_save_step_of_a_clear_answers_the_socket_instead_of_dropping_it`, now asserts that the err text does not contain the temporary state root and still names `goals/<sid>.json`. That covers the load-fault and the save-fault site of `_mark_nodes_cleared` for both clear gestures. Before the change all three fail on the state root appearing in the text.
- `test_a_sub_goal_drop_the_fault_skipped_answers_with_its_own_account` gains the absence check for the drop's dialog (the load-fault site of `_mark_nodes_cleared`, reached through nodeOverride). Fails the same way before the change.
- `test_the_refusal_names_the_goals_file_relative_to_the_state_root` (new): an undo of a compacted card whose store faults at the load (the load-fault site of `_restore_goal_archive`). Asserts that the err text does not contain the state root, still names `goals/<sid>.json` and `Input/output error`, and that the `store-unreadable` judge-errors row keeps the whole path. Fails on the state root before the change.
- `test_a_publish_that_fails_at_the_undo_answers_and_leaves_the_ids_owed` gains the absence check for the failed publish's temporary file path (the save-fault site of `_restore_goal_archive`). Fails the same way before the change.
- `test_the_fault_copy_keeps_the_errno_text_and_the_file_name` (new): the helper's exact text for an EIO on the goals file, for an ENOENT naming both the publish's temp file and its destination (both lose the root), and the type-name fallback for a fault with no text. Fails with AttributeError before the change.

Reverting any one of the four sites to `str(fault)` fails at least one of these tests; a helper that strips only the first occurrence fails the two-path case.

Before the change: 7 failed, 23 passed in the module. After: 30 passed. Full suite (`pytest -q -n 8 tests/`): 10200 passed, 61 skipped, 1671 subtests passed in 73.47s, 0 failed.

## Tier

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -30820,7 +30820,7 @@ def _mark_nodes_cleared(item_ids, value, src="user", why=None):
     for sid, ids in by_sid.items():
         store, fault = jd.load_goals_or_fault(sid)
         if fault is not None:
-            skipped[sid] = str(fault)                  # its row is filed; the flag cannot be written on a store we
+            skipped[sid] = _store_fault_copy(fault)    # its row is filed; the flag cannot be written on a store we
             continue                                   # cannot read (the view-level clear still holds), the other
         nodes = store.get("nodes", {})                 # sessions' clears proceed, and the CALLER answers the user
         touched = False                                # (a WS gesture reports the refusal to its own socket)
@@ -30869,7 +30869,7 @@ def _mark_nodes_cleared(item_ids, value, src="user", why=None):
         jd.rollup_status(store, closed)
         fault = jd.save_goals_or_fault(sid, store)
         if fault is not None:
-            skipped[sid] = str(fault)                  # the store read a moment ago and faults at its SAVE (the
+            skipped[sid] = _store_fault_copy(fault)    # the store read a moment ago and faults at its SAVE (the
             continue                                   # save path's own strict reads, or the publish itself): the
         #                                                flag did not land, and the caller answers the user exactly
         #                                                as for a load fault. Left to raise, an OSError out of a WS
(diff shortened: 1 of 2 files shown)


ksarma:
Merging as shipped; no review fix to add.

Reviewed inline: the four `skipped[sid]` sites (the load and save faults in `_mark_nodes_cleared` and `_restore_goal_archive`) now hand `_store_fault_copy(fault)` to `_gesture_store_refusal`. That helper is the whole behavior change: `str(fault)`, or the type name when the fault has no text, with `str(jd.STATE) + os.sep` removed wherever it occurs. That is the prefix the faults carry: every store path in `kernel/judge.py` is built from `GOALDIR`, which `_rebind_state` derives from `STATE`. pathlib normalizes a trailing or doubled separator. An `OSError` keeps the filename as the loader spelled it, so a symlinked or relative root strips the same way, and a rename fault loses the root from both paths. `_gesture_store_refusal` is the only reader of the map's values; `_undo_clear` reads keys, and the other two callers discard the return. `_file_store_fault` still writes the whole path to the judge-errors row, which the new undo test pins. `_oserror_text` states the same rule for its frames. Keeping `goals/<sid>.json` here suits a dialog about one session's file. Body and commit message match the diff and quote no one. The added lines carry no private strings, hostname or home path. The fixtures use private synthetic sids, unlink their override journals in tearDown, and dispatch in-process against a bare Handler with no socket.

Follow-up, outside this PR: `kernel/kernel.py:48960`, the nodeOverride resolve path, still hands `str(_e)` to the toast (`ui/webview/feed.ts:5631`). On an unreadable store that string carries the same absolute path. The same helper fits there.

Tests: the PR's test file, `tests/test_goal_store_fault_boundary.py`, ran red-first at main bce52234 with 7 failed, 23 passed. The seven were six GestureRefusal dialog tests on the state root found in the err text, plus the helper test on the missing function. At head, 30 passed. Reverting any one of the four sites fails at least one test. A first-occurrence-only replace fails the two-path case. CI at 733a9178: all eleven checks green (Python 3.10 to 3.13, bats, vscode-extension, both tier checks).

Review run by the maintainer automated review session; verified against head 733a9178.

PR #1211, opened 2026-09-09
State root: an empty XDG_STATE_HOME reads as unset in every Python reader, as the shell surfaces already do

ksarma:
## Symptom

A kernel started with a blank `XDG_STATE_HOME=` line (a service.env entry with no value) keeps its state under the relative path `romp` in its working directory, while romp-sdk-setup, which builds the SDK venv, romp-service, the hooks and the manager use `~/.local/state/romp`. As the user sees it: the SDK venv that romp-sdk-setup installed is reported as not installed, because the kernel looks for it under its own state root. The same split applies to postal mail and session names, to the idle-dots pidfile, and to the Codex runtime the live smoke helper looks up.

## Cause

Six state-root lines read `os.environ.get("XDG_STATE_HOME", default)`, which keeps an empty value; `Path("") / "romp"` is the relative path `romp`, so the root follows the process cwd (and judge.py creates and chmods that relative directory at import). The XDG base-directory spec reads an empty variable as unset. Every shell surface's `${XDG_STATE_HOME:-...}` and the manager's `||` already do, and cli/update.py and cli/spend_rebuild.py already use the `or` form, so these six lines were the only readers that disagreed: five runtime readers (postal/postal_service.py STATE and NAMES_DIR, kernel/event_model.py, kernel/judge.py, cli/idle_dots.py) and the hand-run tests/smoke_codex_live.py's RUNTIME_STATE.

## Fix

- The five runtime lines fall through with `or`: `Path(os.environ.get("XDG_STATE_HOME") or str(HOME / ".local/state")) / "romp"`. One note at event_model.py states the rule; the other sites point to it.
- tests/smoke_codex_live.py's RUNTIME_STATE takes the same one-line change. It was the last instance of the idiom in the tree, and it reads the root to find the installed Codex runtime (`ensure_codex_sdk(RUNTIME_STATE)`) before rebinding XDG_STATE_HOME to a scratch dir, so it had the same failure mode. Included, and covered by the test below.
- Behaviour change, deliberate: anyone with XDG_STATE_HOME set to the empty string had their Python-side state under `./romp` and their shell-side state under `~/.local/state/romp`. The Python readers now join the shell surfaces on the home root.

## Tests

- `tests/test_state_dir_override.py::PythonSurfaces::test_empty_xdg_state_home_is_unset` (new): imports each PY_SURFACES module in a subprocess with exactly `{PATH, HOME=<temp>, TMPDIR=<temp>, XDG_STATE_HOME=""}` and asserts `<temp>/.local/state/romp` plus the module's suffix, one subTest per surface so a regression names every affected reader. Before the change it fails on all six: event_model STATE, judge STATE, idle_dots STATE and the smoke helper's RUNTIME_STATE print `romp`, postal STATE prints `romp/postal`, postal NAMES_DIR prints `romp/names` (6 failed, 8 passed, exit 1). After the change all six print the absolute path under the temp home (8 passed, 6 subtests passed). Reverting any one of the six lines alone turns exactly that surface's subTest red.
- PY_SURFACES gains two rows, `("cli/idle_dots.py", "STATE", "")` and `("tests/smoke_codex_live.py", "RUNTIME_STATE", "")`, so the existing override and default cases cover those readers too. Both cases pass before and after the change; the empty case is the only red.
- `_derive` takes an optional `cwd` and the empty case passes the temp dir: judge.py mkdirs its root at import, so a regression creates `romp/` in the temp dir rather than in the checkout. `_derive` also sets TMPDIR to the case's temp home, since the smoke helper mkdtemps at import; the checkout and /tmp stay clean either way.
- Full suite (`pytest -n 8 tests/`) on this head: 10203 passed, 62 skipped, 1677 subtests passed in 70.97 s, no failures.

Tier: fix

### DIFF
--- a/cli/idle_dots.py
+++ b/cli/idle_dots.py
@@ -42,7 +42,8 @@ from pathlib import Path
 
 HOME    = Path.home()
 STATE   = Path(os.environ.get("ROMP_STATE_DIR")   # per-kernel state root override (plans/multi-kernel.md)
-               or Path(os.environ.get("XDG_STATE_HOME", str(HOME / ".local/state"))) / "romp")
+               or Path(os.environ.get("XDG_STATE_HOME") or str(HOME / ".local/state")) / "romp")
+# `or`, not a .get default: an empty XDG_STATE_HOME is unset (the note at kernel/event_model.py's STATE).
 PIDFILE = STATE / "idle-dots.pid"
 
 STALE_AFTER_SECS = 3600          # 1h — matches dashboard STALE_AFTER_SECS + timeline STALE
--- a/kernel/event_model.py
+++ b/kernel/event_model.py
@@ -32,7 +32,10 @@ from pathlib import Path
 
 HOME     = Path.home()
 STATE    = Path(os.environ.get("ROMP_STATE_DIR")   # per-kernel state root override (plans/multi-kernel.md)
-                or Path(os.environ.get("XDG_STATE_HOME", str(HOME / ".local/state"))) / "romp")
+                or Path(os.environ.get("XDG_STATE_HOME") or str(HOME / ".local/state")) / "romp")
+# `or`, not a .get default: an EMPTY XDG_STATE_HOME is unset (the XDG spec, and every shell reader's
+# ${XDG_STATE_HOME:-...}). A .get default kept the empty string and made the root the RELATIVE path
+# romp under the process cwd, so the kernel and the shell surfaces used different roots.
 PROJECTS = Path(os.environ.get("CLAUDE_CONFIG_DIR") or str(HOME / ".claude")) / "projects"   # per-kernel Claude root (plans/multi-kernel.md phase 2)
 NAMES    = STATE / "names"
 STATES_DIR   = STATE / "states"
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -63,7 +63,8 @@ _cred = sys.modules.get("romp_credentials") or SourceFileLoader(
 
 HOME     = Path.home()
 STATE    = Path(os.environ.get("ROMP_STATE_DIR")   # per-kernel state root override (plans/multi-kernel.md)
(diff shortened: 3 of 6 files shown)


ksarma:

Reviewed inline: the six state-root lines now fall through with `or`, so an empty XDG_STATE_HOME reads as unset. They are `postal_service.py` STATE and NAMES_DIR, `event_model.py`, `judge.py`, `idle_dots.py`, and the hand-run `smoke_codex_live.py`. The XDG spec and every shell reader's `${XDG_STATE_HOME:-...}` already read it that way. No other reader keeps an empty value: `cli/update.py` and `cli/spend_rebuild.py` already used `or`, the three JS readers use `||`, and every shell surface uses `:-`. The new test gives each subprocess a fresh environment with HOME, TMPDIR and cwd on a temp directory. That puts `judge.py`'s import-time mkdir under the temp home, and neither the checkout nor the live state root gains a romp directory. A probe under that environment confirmed it. The smoke helper is outside pytest's default `test_*.py` pattern, and no ini file widens it, so CI does not collect it. `docs/reference.md` and `docs/architecture.md` already give the root as `${XDG_STATE_HOME:-~/.local/state}/romp/`. That form reads empty as unset, so no doc change is needed. No template, installer or service script writes a blank `XDG_STATE_HOME=` line. The declared behaviour change therefore affects only a user who set the empty string by hand. Body and commit message match the diff. The 38 added lines pass the privacy scan.

Red-first confirmed: with the PR's test file at main, `test_state_dir_override.py` fails 6 and passes 8. The six failures are every surface of `test_empty_xdg_state_home_is_unset` reporting a relative path. At the head it passes 8 with 6 subtests. Reverting any one fixed line alone fails exactly that surface's subTest. CI at the head is green on all eleven checks.

Review run by the maintainer automated review session; verified against head 1c7a8144.

PR #1212, opened 2026-09-09
PR links: the boundary before a reference is judged over the rendered text, not per text node

ksarma:
## Symptom

A `#12` written right after inline code, emphasis or a link in a chat message links to pull request 12 of the session's repository. `docs/a.md`#12 (the path in backticks), **docs/a.md**#12 and [text](https://x/y)#12 all do it: the path's fragment, or a number the writer glued to a word, becomes a link to an unrelated PR. A path link followed by its fragment has the same shape wherever a path pass runs before the PR pass.

## Cause

pr-links.ts requires a word boundary before a reference (the header's rule, from https://github.com/romp-on/romp/pull/1004), but the walk judged it per text node: REF_SRC accepts the start of the text (`^`) as a boundary, prRefSegments had no notion of what precedes its text, and visit() called linkifyTextNode on each text child with no carry across the element before it. marked renders each shape above as an element followed by a text node that begins `#12`, so the number sat at the start of its own node and passed. A bare URL is not a case: GFM's autolink swallows the fragment into one anchor.

## Fix

The general rule: a reference's boundary is judged over the rendered text, not per text node. Where the renderer or an earlier pass has cut one run of text into several nodes, the start of a node is no boundary; the character before it in the rendered text decides.

The instance:

- `prRefSegments(text, repo, before = "")` takes the one character that precedes `text` in its run ("" for the start of the run; the DOM walk passes `text.slice(-1)`, and a longer string is never a boundary). A match at the start of `text` counts only when `before` is a boundary character. A refused match is scanned again from its second character, so a boundary inside it still counts (`d` then `PR #12` links the bare `#12`, as the one-node reading of `dPR #12` does), while a glued run (`d` then `#12/#13`) links nothing, `/` being no boundary. The boundary class is one constant (`BOUNDARY`) shared by the regex and the single-character test, so the two cannot drift.
- `linkifyPrRefs`'s `visit(node, before)` carries the last character of the text before each node into it. A skipped element's text counts (a `<code>`, an `<a>`, a path link). The carry crosses inline elements only: an ``, an empty `<span>`, an empty `<code>` or any other element with no text passes the run through, so the text before it decides. A comment node is ignored.
- `BLOCK_TAGS` (47 block-level and line-breaking tags) resets the carry at a line or block edge, where the rendered text has a boundary whether or not a whitespace node sits there. marked renders a GFM hard break as `<p>done<br>#12</p>` with no text between the `<br>` and the `#`; a carry without this reset would refuse that `#12`. The list follows how an element renders, not what the sanitizer admits (md-sanitize.ts strips form, fieldset, legend and dialog today and keeps their text). An element the list does not name is treated as inline, the safe default: a missed edge costs a link, an invented one makes a wrong link. Tag names are read case-insensitively, as the skip list has always read them.
- The public API is additive (an optional third parameter); no call site changes (the chat's `md()` and `userMd()`, the feed's `linkifyPrRefs` and `setLinkedText` callers, the outline's goal rows).

The rule has two limits, both pinned in the tests. `before` is one UTF-16 code unit, so an emoji directly before a split `#12` yields a lone surrogate, which is no boundary; the one-node regex refuses that shape too, so the two readings agree (no link). What follows a reference is still judged per node: a `#12` at the end of an emphasis with letters glued to it in the next node links as it did before; the trailing lookahead is outside this change.

Checked beyond the tests: 4000 random token strings, every cut point, four split shapes (adjacent text nodes, one or both halves inside a wrapper), 193k cases. The split reading equals the one-node reading except at the trailing side named above and at slash runs cut between their members (a shape marked never emits), which link fewer references, never more.

## Tests

`ui/webview/pr-links.test.ts`, three tests after "the chat's path links are skipped", all executing through the file's plain-object DOM stand-in; each fails on the unchanged source:

- "the boundary is judged over the rendered text: a reference at the head of a text node follows the character before that node": `prRefSegments` with the preceding character (a word character, `/` and `-` glue; every boundary character and the empty string do not; a lone surrogate is no boundary and the one-node reading of the same text agrees; a string longer than one character never is; the two-argument call is unchanged; a glued `PR #12` links its `#12`; a glued cross-repo form and a glued run are refused and the scan resumes), then a path-link span followed by `#12 and #13` through the walk, which links `#13` alone. Before: `prRefSegments("#12 next", REPO, "d")` links `#12`, the span case links both, and the third argument is a TypeScript error, so `npm run typecheck` is red as well.
- "an element's text ends the run too: #12 glued to <code>x</code> is no reference, #12 after a space is; an inline element with no text carries the run": `use <code>x</code>#12` links nothing and `use <code>x</code> #12` links once; the same after an `<a>` and a `<strong>`; an empty `` or `<span>`, and an empty skipped `<code>` or `<input>`, carry the run; a walked wrapper's text carries out of it and the run before it carries in; nested wrappers; a comment node; the trailing side's limit (a `#12` ending an emphasis links although letters follow in the next node). Before: `use <code>x</code>#12` returns 1 and wraps `#12` in a pr-link anchor.
- "a line or block edge is a boundary with no whitespace node: marked's hard-break <br>, abutting paragraphs, list items, table cells; an inline element with no text is not": marked's HTML (a `Marked` instance with the chat's options, read into the stand-in by a tags-and-text reader) for the hard break, its shape pinned as `["done", "BR", "#12 next"]`, the backslash break, `PR #12` and `example-org/other#4` after a break, a run at the head of a line, the block shapes marked emits (paragraph, list item, table cell, heading, quote, fence, rule), the inline shapes it cuts before a number (inline code, strong, em, an explicit link, a raw span); then hand-built abutting blocks (p/p, li/li, td/td, h2/p, p/blockquote, pre then text, hr, br, div/div, center, legend, dialog, hgroup, details/summary), every one of the 47 block tags (the list copied into the test) as an edge on both sides (a `#12` at its head after glued text links, and text right after it starts a run; a `<pre>` is code-like, so its far side only), a lower-case tag name, and the glue protections. Before: the inline shapes link. Against a carry with no block-edge reset, the hard break stops linking; a tag dropped from the block list fails its own row.

`pr-links.test.ts`: 55 pass (52 before). `npm run typecheck`: clean. `npm test` in vscode-extension: 3750 tests pass, 0 fail (every bundled test under node --test).

Tier: fix

### DIFF
--- a/ui/webview/pr-links.test.ts
+++ b/ui/webview/pr-links.test.ts
@@ -4,10 +4,12 @@
 // rule and the DOM applier are EXECUTED here (pr-links.ts is pure but for createElement/createTextNode,
 // which a small plain-object DOM stand-in provides); the wiring into render.ts's md(), feed.ts and
 // fleet.ts is pinned at the source, the way the other webview tests pin the chat renderer.
+// The chat's own markdown shapes come from marked, read into the same stand-in.
 import { test } from "node:test";
 import * as assert from "node:assert/strict";
 import * as fs from "node:fs";
 import * as path from "node:path";
+import { Marked } from "marked";
 
 // ── a DOM stand-in: text and element nodes with the handful of members the applier touches ─────────
 class T {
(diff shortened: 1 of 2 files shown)


ksarma:
Post-merge record: the repository owner merged this PR as 26a1ab06 at the reviewed head before this record posted; the review passed, and it would have merged as shipped. Nothing to fold in.

Reviewed inline: `prRefSegments` takes the one character before its text. It refuses a match at index 0 when that character is present and not in `BOUNDARY`. It then rescans from index 1, where the start-of-text alternative cannot match again, so a boundary later in the text still counts. Examples: `d` then `PR #12` links `#12`; `d` then `#12/#13` links nothing. `BOUNDARY` feeds both `REF_SRC` and `BOUNDARY_RE`, and the class is the one the old regex carried. `visit(node, before)` carries the last character of the original text. It takes a skipped element's `textContent` as its tail. It passes the carry through an element with no text and through comment nodes. It resets the carry at a `BLOCK_TAGS` element; with that reset removed, marked's hard break stops linking. Probes against marked 12 output confirm the shapes the PR body names: inline code, emphasis, an explicit link, a path span before its fragment, an empty span, nested wrappers, block edges, and a reference at the start of a message. `prRefSegments` has one production caller, `linkifyTextNode`, which passes the new argument. Every other caller goes through `linkifyPrRefs`, unchanged. The added lines carry synthetic fixtures only, with no private strings, hostnames or home paths. The PR body matches the diff.

I verified red-first with my own run. With the PR's test file at `origin/main`, `pr-links.test.ts` fails exactly the three new tests and passes the other 52. The three: the pure rule links `#12` after `d`; `use <code>x</code>#12` returns 1; marked's inline shapes link. At head, 55 of 55 pass, `npm test` 3750 of 3750, typecheck clean. CI at head: all eleven checks green, and the Tier policy check reports the fix gate.

Review run by the maintainer automated review session; verified against head 3d09b760.

PR #1213, opened 2026-09-09
esbuild.js: a failed rebuild leaves dist/ as it was, each bundle lands by rename, and the last stderr line names the cause

ksarma:
## Symptom

The kernel rebuilds the served bundles in place by running `node esbuild.js` (from `_rebuild_dist` for the converge and pre-restart paths, and from `_ensure_bundles` at boot). When that build fails on the webview bundle after the extension bundle has built, `dist/` is left half new: `dist/extension.js` rewritten, every webview bundle old. The kernel's cache token is the newest mtime under `dist/*.js`, so it jumps: every open dashboard is told a newer build exists and reloads into the same stale bundles, and the drift pass reads dist as newer than the sources it just failed to build. The notice that says the served UI is stale carries the last 300 characters of the script's stderr, which for an esbuild BuildFailure printed whole is its stack through esbuild's transport (`at Socket.emit`, `errors: [Getter/Setter]`); it names neither the failing import nor what to do about it. The common trigger is a merge that imports a package the checkout's `node_modules` predates, the shape of the katex failure of 2026-08-10 that `_ensure_bundles`' docstring records.

The same write has a second, smaller defect: each output was written into its served path in place, so a `/dist/` request served on another thread during the write got an empty or partial bundle, and the page that loaded it ran a bundle cut mid-statement until a reload.

## Cause

`main()` ran `esbuild.build(extension)` and then `esbuild.build(webview)`, each writing `dist/` as it finished, and `main().catch` printed the error object whole.

## Fix

`vscode-extension/esbuild.js` only; the kernel is untouched.

- `buildAll(configs)` builds every config with `write: false`, then writes each output whole to a hidden sibling in its own directory (`.<name>.tmp-<pid>-<n>`: the same filesystem, so the rename is atomic; a name the kernel's `dist/*.js` glob never matches and its `/dist` route never types) and renames it over the served name. An fs error while staging removes the staged files and rethrows, so `dist/` stays byte-for-byte as it was with its mtimes standing. A staging file left by an exited pid (a build killed between its write and its rename) is removed first; one owned by a running pid is kept, so two builds on one `dist/` each rename their own.
- The one-shot build (`node esbuild.js`, with or without `--production`) goes through `buildAll`. `--watch` and `--tests` write directly as before: a dev loop wants each rebuild on disk at once.
- `failureSummary(e, untouched)` composes the last stderr line: `esbuild.js: build failed with N error(s); dist/ is unchanged.` plus the unresolved specifiers (two named, the rest counted, each cut at 60) and, when one is a bare package or a `node_modules/` path, `not in this checkout's node_modules; run npm install in vscode-extension/ and rebuild.` A relative import, a path under a directory of the checkout (`src/nowhere`) or a syntax error gets its text (cut at 120) and location; a non-esbuild error gets its message. Every line is cut to 300 characters from the end (297 plus `...`), so the head is what the kernel's `[-300:]` shows. esbuild's own report with its code frames still precedes it (`logLevel: "info"` is unchanged); the BuildFailure object is no longer printed, a non-esbuild error still is, since esbuild printed nothing for it and its stack is the diagnosis.
- `main()` runs only under `require.main === module`; `buildAll`, `failureSummary`, `extension`, `webview` and `testBuild` are exported for the tests.

`write: false` holds every bundle in memory until all have built (a few MB, plus sourcemaps in a dev build); nothing changes on success. The 300 pairs with `_rebuild_dist`'s `.strip()[-300:]`; the served-page test classes under `tests/` read the same stderr at `[-200:]` into a SkipTest message, so a line near 300 loses its head there (the npm install cure at the end survives). Those classes each run `node esbuild.js` into the shared `dist/` and `copytree` it in `setUpClass`. CI runs pytest serially, so they are unaffected; two local `pytest -n` workers building at once now fail loudly with a `shutil.Error` on a staging file that vanished mid-copy, where before a truncated in-place bundle could be copied silently. A one-build-then-copy guard for those classes, or `ignore_patterns('.*.tmp-*')` at the copytree calls, is a follow-up if local xdist runs matter.

## Tests

Three new files under `vscode-extension/src`, run by `npm test` through the existing `src` entry of `testBuild()`. Every case fails on the previous script, which exports nothing and builds on require (`buildAll is not a function`; the child-process case exits 1 on a build started in an empty directory). Against a variant that adds the exports and the gate but keeps the two per-build writes, the write and summary cases fail on the defect itself: the first bundle's old bytes are replaced, the dist snapshot changes, the interposed write is never reached, no staging file is removed, and there is no summary line. Each stated cut and branch has a case that goes red when it is removed: the 120 text cut (removed, or moved to 1200), the cwd-directory clause of the npm install rule, the printing of a non-esbuild error, and the `--production` path through `buildAll`.

- `esbuild-build.test.ts` (8 cases) drives `buildAll` and `failureSummary` against synthetic entries in scratch dirs: a failed second bundle leaves the first bundle's old bytes and its 2026-01-01 mtime; a success writes code, sourcemaps and file-loader assets in their subdir and returns the written paths; the npm install line appears for a bare specifier and a `node_modules/` path and not for `./nowhere`, `src/nowhere` or a syntax error; long specifiers are cut at 60 and a long list is counted. A child process shows that requiring the script in an empty directory exits 0, reports no build and writes nothing. The last case runs `node esbuild.js` itself over a stub tree of the real configs' entry points: a webview import of a missing package after a good build, run without `--production` and again with it, leaves every file under `dist/` with its size and mtime and ends the stderr's last 300 characters in the summary line; then, with the sources good again and a regular file in `dist/`'s place, the write step fails with EEXIST, the error's stack is printed, the last line is `esbuild.js: build failed: EEXIST...`, and nothing is staged beside it.
- `esbuild-failure-cap.test.ts` (4 cases) pins the cuts: a composed line of 300 goes out whole, one of 301 comes back as 300 ending in `...` with its first 297 characters intact; a 200-character error text over a short path is cut to 117 and `...` with its location still closing the line, and a text of exactly 120 goes out whole; a real syntax error under a 13-deep directory whose location path alone exceeds 250 characters, where the uncut line through `slice(-300)` starts mid-word; a non-esbuild message over 300.
- `esbuild-write-rename.test.ts` (3 cases) replaces `fs.writeFileSync` for output writes with a two-half write that reads the served names between the halves: each read sees the old file whole or the new one whole; the staging names are hidden, in the destination's directory, end in none of the suffixes the kernel's `/dist` route types (`.js`, `.css`, `.map`, `.svg`, `.ttf`, `.woff`, `.woff2`, `.png`), carry the pid, and none is left behind. An ENOSPC on the second staging write leaves the old bytes and mtime and no staging file. A staging file whose pid has exited is removed, beside one whose pid runs and a `.keep` that is not a staging file, both kept.

Verification: `npm run typecheck` clean; the three files 15 of 15 (five consecutive runs); full `npm test` 3762 of 3762 passed in 15 s; `tests/test_bundle_build_mode.py` 7 passed (its pins on the minify flags are untouched).

Tier: fix

### DIFF
--- a/vscode-extension/esbuild.js
+++ b/vscode-extension/esbuild.js
@@ -113,6 +113,118 @@ function testBuild() {
   };
 }
 
+// Every bundle of a one-shot build (`node esbuild.js`, with or without --production) is built into memory and
+// dist/ is written only once ALL of them have built. esbuild writes each build()'s outputs as that build
+// finishes, so with the extension bundle first and the webview bundle second, a webview failure left
+// dist/extension.js rewritten and every webview bundle old.
+// The common failure has exactly that shape: a merge imports a package this checkout's node_modules predates
+// (the 2026-08-10 katex import broke every restart's rebuild for a day). The kernel's cache token is the
+// newest mtime under dist/, so the half-written dist bumped it: every open dashboard was told a newer build
+// existed and reloaded into the same stale bundles, and the kernel's drift pass then read dist as newer than
+// the sources it had just failed to build. A failed build now leaves dist/ byte-for-byte as it was, so "the
+// rebuild failed" and "dist is older than the sources" stay true together. Watch mode and the test build
+// write directly as before: a dev loop wants each rebuild on disk at once.
+//
+// Each output then reaches its served name by RENAME, not by a write into it. The kernel serves /dist/ from
+// this directory on other threads while it rebuilds, and a write into the served path truncates the file
+// first and fills it afterwards, so a request in that window got an empty or partial bundle (longest for
+// render.js, the largest output) and the page that loaded it ran a bundle cut mid-statement until a reload.
+// So each output is written whole to a hidden sibling in its own directory (`.<name>.tmp-<pid>-<n>`: the same
+// filesystem, so the rename is atomic; a name ending in neither .js nor .css nor .map, so the kernel's
+// newest-mtime token (dist/*.js) never counts it and its /dist route never types it) and then renamed over
+// the served name, and a reader gets the old bytes or the new, complete either way. Every output is staged
+// before any is renamed, so an fs error while staging (no space, a permission) removes the staged files and
+// leaves dist/ unchanged. A staging file an earlier run left behind (killed between its write and its rename)
+// is removed once that run's pid is gone, and kept while the pid runs: two builds on one dist/ each rename
+// their own.
+const STAGING = /^\..+\.tmp-(\d+)-\d+$/;
+
+function stagingPath(final, n) {
+  return path.join(path.dirname(final), "." + path.basename(final) + ".tmp-" + process.pid + "-" + n);
+}
+
+function pidRunning(pid) {
+  try { process.kill(pid, 0); return true; } catch (e) { return !(e && e.code === "ESRCH"); }
+}
+
+function removeStaleStaging(dir) {
+  for (const name of fs.readdirSync(dir)) {
+    const m = STAGING.exec(name);
+    if (m && !pidRunning(Number(m[1]))) fs.rmSync(path.join(dir, name), { force: true });
+  }
+}
+
+async function buildAll(configs) {
+  const results = [];
+  for (const cfg of configs) results.push(await esbuild.build({ ...cfg, write: false }));
+  const outputs = results.flatMap((r) => r.outputFiles);
+  for (const dir of new Set(outputs.map((f) => path.dirname(f.path)))) {
+    fs.mkdirSync(dir, { recursive: true });
+    removeStaleStaging(dir);
+  }
+  const staged = [];
+  try {
+    for (const f of outputs) {
+      const tmp = stagingPath(f.path, staged.length);
+      fs.writeFileSync(tmp, f.contents);
+      staged.push(tmp);
+    }
+    outputs.forEach((f, i) => fs.renameSync(staged[i], f.path));
+  } catch (e) {
+    for (const tmp of staged) fs.rmSync(tmp, { force: true });
+    throw e;
+  }
+  return outputs.map((f) => f.path);
+}
+
+// The LAST line on stderr is the one the kernel shows: its in-place rebuild puts the tail of this process's
+// stderr into the notice that tells the person the served UI is stale, and the tail of an esbuild
+// BuildFailure printed whole is its stack through esbuild's own transport (`at Socket.emit`,
+// `errors: [Getter/Setter]`), which names neither the failing import nor what to do about it. esbuild has
+// already printed each error with its code frame (logLevel "info"), so this names the cause in one line of
+// at most 300 characters (the kernel's tail: _rebuild_dist in kernel/kernel.py) and, for the common cause,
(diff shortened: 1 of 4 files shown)


ksarma:
Post-merge record: the repository owner merged this PR as 459e50d6 at the reviewed head before this record posted; the review passed, and it would have merged as shipped. One follow-up to file; nothing to fold into this PR.

Reviewed inline: buildAll builds every config with write false. It stages each output to `.<name>.tmp-<pid>-<n>` in the output's own directory, on the same filesystem, and renames only after every output is staged. The catch removes the staged files and rethrows. A BuildFailure or an fs error while staging therefore leaves dist/ with its bytes and mtimes as they were. The staging name ends in digits, so the kernel's `DIST.glob("*.js")` token never counts it. The /dist route serves by exact requested path, so it never serves a staging file. Leftover cleanup decides by `process.kill(pid, 0)`, which delivers no signal: ESRCH removes the file, anything else keeps it. Two builds on one dist/ each rename their own. The one-shot build goes through buildAll. The --watch and --tests branches are unchanged, and main() runs only under `require.main === module`. failureSummary cuts the last line to 300 characters from the end, matching `_rebuild_dist`'s `.strip()[-300:]`, so the converge and pre-restart notices include the cause. `_ensure_bundles` prints the CalledProcessError only, as before; the body does not claim to change it. The kernel is untouched. The body and commit message match the diff. The added lines carry no private strings, hostnames or home paths. The tests spawn only their own children, send no signal, write under os.tmpdir() and restore the interposed fs.writeFileSync.

Findings:

- tests/test_awaiting_box_sync.py:135 and the 15 sibling served-page classes, follow-up. Each rebuilds the shared dist/ and copytrees it in setUpClass. Under local `pytest -n`, a sibling worker's staging file can vanish between copytree's listing and its copy, and the class errors with shutil.Error. Before this PR it copied a truncated bundle silently. The body discloses this; CI runs pytest serially and is unaffected. `ignore=shutil.ignore_patterns(".*.tmp-*")` at the 16 copytree calls, or one build-then-copy guard, fixes it. Until then, re-run a served-page setUpClass error under -n alone before attributing it to anything.
- vscode-extension/esbuild.js:220, nit. With `errors: []` the else branch reads `errors[0].location` and throws inside main().catch. This is unreachable through esbuild's API; one length check fixes it.
- vscode-extension/esbuild.js:198, nit. No test on either side checks that the 300 matches kernel.py's `[-300:]`.

Verification, red-first by my own run: at the base, the three test files fail 15 of 15 (six `buildAll is not a function`, five `failureSummary is not a function`, the child-process case exiting 1 on an unresolved src/extension.ts, and the base's require-time build writing 31 files into the scratch dist/). At head, 15 of 15 pass, npm test 3762 of 3762, typecheck clean, and tests/test_bundle_build_mode.py 7 passed. Mutation runs (in-place writes, no 300 cut, no stale-staging removal, no cleanup on error) each fail the case the body names. CI at 09a8e6cd: every check SUCCESS.

Review run by the maintainer automated review session; verified against head 09a8e6cd.

PR #1214, opened 2026-09-09
Feed: a second user gesture in one judge pass replays onto a fresh snapshot copy, never in place on the copy an earlier read served

ksarma:
## Symptom

Nothing renders wrong from this today; the PR closes an aliasing hazard. While a judge pass is in flight, `_feed_goals` serves the feed from the pass snapshot and replays the user's journaled gestures onto it. The first gesture on a session copies the snapshot entry before the replay (copy-on-punch); a second gesture in the same pass, or a retry after a failed replay, used to replay onto that first copy in place and hand back the same object. So an object `_feed_goals` had already served changed under its holder.

`build_feed` is the only caller and nothing keys on the served object's identity, which is why no card reads wrong now. The hazard becomes wrong rows the moment a reader keeps the served store by identity. The consumer this prepares for is a per-session feed memo: a session's rows kept under its served store's identity and served while that object stands. Without this change such a memo hits on the first copy and serves the pre-gesture rows for the rest of the pass, 30 to 80 s on a busy kernel (a resolve on an open sub after a follow-up would read open on the board while a fresh build read done).

## Cause

In `_feed_goals` the json round-trip copy sat inside the `if sid not in _goals_snap_owned:` guard, whose purpose is to keep the `punch` counter at one per session per pass. The guard therefore also gated the copy: only the first moved mark in a pass copied; every later one replayed in place on the copy an earlier read had served. The punch test pinned that behaviour: `assertIs(km._feed_goals(SID), served, "a second gesture punches the copy already made")`.

## Fix

`kernel/kernel.py`, `_feed_goals`: the copy statement moves out of the guard, so every moved mark (and every retry after a failed replay) lands on a fresh copy of the snapshot entry. The guard stays around `_goals_snap_owned.add(sid)` and the `punch` increment, so `punch` still counts sessions copied once per pass each. The counters test's assertion is unchanged and green; its message now says one punch per pass per sid where it said one copy. The COPY-ON-PUNCH comment, the docstring and the `_goals_snap_owned` comment now state the contract: an object `_feed_goals` has handed out is a fixed value. The memo's object is never mutated, as before, and `_feed_goals` keeps its one-value return.

Cost: one json round trip per gesture instead of one per session per pass. Gestures are user-rate events (one click each).

## Tests

- `tests/test_kernel.py::ViewBuilder::test_a_second_gesture_in_the_same_pass_never_changes_the_copy_an_earlier_read_served` (new): a follow-up and then a resolve on one settled goal inside one pass. Asserts the second read is a different object, that it carries the resolve (`nodeComplete` set), that the first served copy still carries only the reopen (`nodeComplete` cleared), that the memoized object still equals the raw parse of the file, and that `punch` moved by exactly one. Before the change it fails at the identity assertion: both reads are the same dict, mutated in place, and the first copy shows the resolve. It mints its store under its own sid through the class's `_settled_store` helper (which now takes a `sid` keyword; the default is unchanged), removes that sid's journal and mark afterwards, and moves the mark by hand for the second gesture so the test never depends on the clock advancing between two gestures.
- `tests/test_kernel.py::ViewBuilder::test_a_user_punch_copies_the_entry_and_never_mutates_the_memoized_store`: the standing pin on the second gesture flips from `assertIs` to `assertIsNot`, with a status assertion that the new copy carries the second reopen; the test's comment follows. Before the change it fails on the flipped assertion.
- `tests/test_kernel.py::ViewBuilder::test_the_memo_counters_and_report_follow_the_passes`: the `punch == 1` assertion is unchanged; only its message is reworded from "one copy per pass per sid" to "one punch per pass per sid". With the guard removed (punch counted per gesture) this test stays green, since it reads one gesture twice; the new test's punch assertion goes red (2 != 1).
- `tests/test_kernel_restore_flicker.py::RestoreReplyFlicker::test_a_failed_replay_does_not_burn_the_punch`: adds that the retry's read is a different object and that the object the failed read served still reads completed. Before the change it fails on the identity assertion.
- Full suite on the final tree: `pytest -q -n 8 tests/`, 10199 passed, 61 skipped, 1671 subtests passed, 0 failed (74 s).

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -25490,7 +25490,8 @@ _judge_gen = [0]                                 # bumped when a producer pass C
 _goals_snap = [None]                             # {sid: store} while a judge pass is mid-flight, else None
 _goals_snap_at = [0.0]                           # when that snapshot's file reads STARTED (see _feed_goals)
 _goals_snap_done = {}                            # sid → the user-write mark already punched onto THIS snapshot
-_goals_snap_owned = set()                        # sids whose snapshot entry is THIS pass's private copy (copy-on-punch)
+_goals_snap_owned = set()                        # sids whose snapshot entry is THIS pass's private copy (copy-on-punch;
+#                                                  the punch counter's once-per-pass guard, not the copy's: see _feed_goals)
 _goals_snap_lock = threading.Lock()
 # STAT-KEYED STORE MEMO (2026-09-06). The snapshot used to json.loads EVERY goals/<fsid>.json at the
 # start of every pass — on one busy kernel 72 files of up to 1.3 MB, about 3% of its interpreter time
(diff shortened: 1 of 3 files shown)


ksarma:

Reviewed inline: the kernel change moves one statement in `_feed_goals`. The json round-trip copy of the snapshot entry now runs on every moved user-write mark. The `_goals_snap_owned` guard keeps only the set insert and the `punch` increment, so the counter still counts each copied session once per pass. Copy, replay, rollup, the done stamp and the `snap[sid]` replacement stay under `_goals_snap_lock`. Served content is unchanged: a second gesture replays the same idempotent journal onto a copy of the already-punched entry and runs the same rollup, so no card moves differently. Only object identity changes, and `build_feed` (kernel/kernel.py:32179), the only caller, does not key on identity. On a failed replay the entry is the copy the raise left, the done mark is not stamped, and the retry copies again. The object a failed read served therefore never changes under its holder; a probe at head confirmed this and showed the base mutating it. The new test mints under a private synthetic sid and removes its mark and journal in finally. The added lines pass the private-strings, hostname, home-path and vocabulary scans. The body matches the diff. The fix label fits: one statement moved and no new mechanism.

Optional, no action needed: the flipped pin at tests/test_kernel.py:5281 relies on `time.time()` advancing between two `_note_user_goal_write` calls. The new test beside it sets the mark by hand for that reason; the same line would do here.

Red-first, run here with the PR's test files at base 3f06288e: the new ViewBuilder test, the flipped punch pin and the flicker retry test fail at their identity assertions (tests/test_kernel.py:5324 and :5281, tests/test_kernel_restore_flicker.py:151); the counters test passes, as the body says. At head all four pass, and tests/test_kernel.py plus tests/test_kernel_restore_flicker.py give 491 passed. CI at 66d85ff2 is green on all 11 check runs; the Tier policy check reports the fix gate.

Review run by the maintainer automated review session; verified against head 66d85ff2.

PR #1215, opened 2026-09-09
Names records: a record that reads with no name is left alone by the kernel's writers, and bin/romp publishes the record atomically

ksarma:
## Symptom
A rename, a recolor or a palette switch erases a session's cwd and colors from the names registry and reports success. Afterwards the registry holds no directory for the session and its identity color is gone; after a rename it holds the new name and nothing else. The case that reaches it most directly is a tmux-backed session renamed from the dashboard, where the kernel and the tmux after-rename hook rewrite the same record at the same moment. Any session whose record reads as 0 bytes reaches it through a recolor or a palette switch.

## Cause
The kernel's three writers of `names/<sid>` (`_set_name`, `_set_session_color`, `_set_palette`) read the record, edit their field and publish the whole line. A record that reads as 0 bytes splits to `['']`; padded to four empty fields, it is published as `<name>\t\t\t\n` by the rename and `\t\t<bg>\t<fg>\n` by the recolor, and the palette writer rewrites any no-name record that carries a palette color. Two things produce the empty read:

- bin/romp's `_romp_record` wrote with `printf > file`: truncate, then write. It is the one shell writer of a record, behind the tmux `after-rename-session` hook (`romp _renamed`), every tmux launch and `_romp_free_title`'s rewrite of other sessions' records. Since https://github.com/romp-on/romp/pull/1066 the kernel publishes the name synchronously right after a live tmux rename, so `_set_name` and the hook rewrite the same file at once: the hook truncates, the kernel reads 0 bytes and its `os.replace` publishes `name\t\t\t`, and the hook's bytes land on the unlinked inode. Tmux-backed sessions reach this path (the tmux backend is a setting, off by default); the SDK and Codex backends rename through their own writers with in-memory fields and have no tmux session for the hook to fire on.
- A damaged 0-byte record, whoever wrote it. The recolor (`POST /color`, the tab's color picker) and the palette switch publish over it for any session, and a dead-tab rename of a sid no module backend owns reaches `_set_name` the same way.

https://github.com/romp-on/romp/pull/1138 made a names write that did not land raise, so the doors no longer say "renamed" over a file the kernel never rewrote. This is the remaining case: the kernel did rewrite the file, with empty fields.

## Fix
- `_names_fields_for_edit(sid, what)` reads the record for the three writers. Every writer puts the name first, so an empty first field is never a real record: it is another writer's window or a damaged file. One re-read after a short pause (`_NAMES_REREAD_S`, 50 ms, paid only on an empty read) tells the two apart: a record whole on the second read is edited as usual; one still without a name is left byte for byte as it was and reported through `_names_problem` to the kernel log and the dashboard's error center (`_sdk_problem`, the bell). A record that cannot be read raises through the helper, so each writer keeps its existing policy for that: the recolor answers False, the palette switch skips, the rename raises the real errno.
- Over a no-name record `_set_session_color` answers False, `_set_palette` skips the record and recolors the rest, and `_set_name` raises `RuntimeError` naming the sid. `_rename_claimed` and `_rename_session` are unchanged: a dead-tab rename says "the rename did not take" with the cause on both doors, and a live tmux rename says the session was renamed but its name on file could not be updated and keeps its old name until that write lands (true there: the hook's rewrite lands). The "is that session known" question stays reserved for an absent record.
- `POST /color` names which cause it hit (`_names_refusal`): the existing "no names record" text for an absent record, and "reads with no name (being rewritten, or damaged); nothing written" otherwise. The WS `setSessionColor` arm sends nothing on False, as before; the error-center row is what the dashboard shows.
- bin/romp `_romp_record` writes the line to `.$sid.tmp.$$` in the names directory and moves it into place with `mv -f`. A failed write or move removes the temp; that arm is cleanup for a full or read-only filesystem, and no test drives it. `_romp_rename_record` and `_romp_free_title` come through it unchanged. The dot prefix keeps the temp out of `_romp_free_title`'s glob and the picker's listing. The kernel's `NAMES.iterdir()` scanners apply no dot filter: they see the temp for the microseconds it exists, and a copy left by a kill between the write and the move reads to them as a phantom entry, the same class as the kernel's own `_atomic_write` temps and the module writers' `<sid>.tmp`.

## Tests
Every test executes the real code path and fails before the change:
- `tests/test_kernel_session_color.py` `NoNameRecordIsLeftAlone` (3 tests, 10 subtests; it extends the module's `SessionColor` fixture, so the class also re-runs that fixture's 11 tests, as `HighSlotSwitches` does). The recolor and the rename over four no-name shapes leave the file byte-identical, re-read exactly once, and report the sid to stderr and the error center (before: the recolor answered True and the file read `\t\t#54B204\tblack\n`; the rename answered True and it read `api\t\t\t\n`). The palette switch skips the no-name record and recolors the rest (before: recolored it, still nameless, unreported). The re-read catches a writer mid-write and the edit lands on the whole record (before: `api\t\t\t\n`); a name-only record `web\n` is still padded and published, so the guard keys on the name field, not on the field count. The kernel's pause is recorded through a stand-in for the module's `time` name, so a sleep by other code in the process is not counted.
- `tests/test_kernel_names.py` `RenameFaultsThroughTheDoors` (3 tests). A dead-tab rename over a 0-byte record answers `warn` with "the rename did not take" and the record's path on the WS door, and `ok: false` on `POST /rename` (before: `renamed`, file `web2\t\t\t\n`). A live rename answers "was renamed, but its name on file could not be updated ... until that write lands", with tmux renamed once and the file untouched (before: `renamed`).
- `tests/test_color_route.py` (2 tests). A 0-byte record gets the no-name refusal and is left alone (before: `ok: true`, file rewritten). An absent record still gets "no names record" (a control: passes before and after).
- `tests/romp.bats` "names map: the record writer publishes atomically, so a reader racing the rename hook never sees an empty record". The record is seeded under a stale name, and `romp _renamed` rewrites it forty times through a mock tmux while a reader polls it and walks the directory with the `*` glob `_romp_free_title` uses. The reader must never find the record empty, and the glob must never show anything but the record. Every hook run must reach the writer (forty `send-keys` in the mock's log), the record must end as the renamed line with the dir and colors intact, and no temp may be left. On the old writer it fails at the empty count in three of three runs; the truncate window is a printf's worth of time, so that leg is probabilistic by nature, and the kernel tests are the deterministic half. It also fails on a writer that copies through the temp instead of moving it (the empty count), on a temp without the dot prefix (the glob count, four of four runs), on a hook that never writes (the send-keys count) and on a writer pointed at another directory (the final content).

Full suite (`pytest -q -n 8 tests/` on the committed tree): 10221 passed, 62 skipped, 1681 subtests passed, 0 failed. Targeted: the three pytest modules 108 passed (10 subtests); `bats tests/romp.bats` 119 of 119.

Tier: fix

### DIFF
--- a/bin/romp
+++ b/bin/romp
@@ -107,7 +107,17 @@ _romp_record() {   # <sid> <name> <dir> [bg] [fg] — write/replace the map entr
     # users from reading anything beneath it (the traverse bit is enough).
     # Idempotent; best-effort.
     chmod 700 "${ROMP_STATE_DIR:-${XDG_STATE_HOME:-$HOME/.local/state}/romp}" 2>/dev/null || true
-    printf '%s\t%s\t%s\t%s\n' "$nm" "$dir" "$bg" "$fg" > "$ROMP_NAMES_DIR/$sid"
+    # Atomic: the line is written to a temp in the same directory and moved into place, one rename(2),
+    # so no reader ever sees the record empty or half-written. The kernel's names writers read the
+    # record while this hook rewrites it (its rename writer runs right after a live tmux rename; its
+    # color and palette writers touch any session's record) and publish their edit over what they
+    # read, so a truncate-then-write here cost the session its name and cwd. This is the one shell
+    # writer of a record (_romp_rename_record and _romp_free_title come through here). The temp is
+    # dot-prefixed, so _romp_free_title's glob and the resume picker's listing skip it; a failed write
+    # or move removes it.
+    local tmp="$ROMP_NAMES_DIR/.$sid.tmp.$$"
+    printf '%s\t%s\t%s\t%s\n' "$nm" "$dir" "$bg" "$fg" > "$tmp" && mv -f "$tmp" "$ROMP_NAMES_DIR/$sid" \
+        || { rm -f "$tmp"; return 1; }
 }
 
 _romp_rename_record() {   # <sid> <name> — update name, keep dir + identity color
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -2258,19 +2258,67 @@ _MODEL_VALUES = {m["value"] for m in MODEL_CHOICES}
 _EFFORT_VALUES = {e["value"] for e in EFFORT_CHOICES}
 
 
+_NAMES_REREAD_S = 0.05   # the pause before the ONE re-read of a names record that read with no name (below)
+
+
+def _names_problem(text):
+    """A names-registry problem, reported where someone sees it: the kernel log AND the dashboard's error
+    center (the bell), the path SDK problems take (_sdk_problem). Never a silent degrade."""
+    sys.stderr.write("[names] %s\n" % text)
+    _sdk_problem("names registry: %s" % text)
+
+
+def _names_fields_for_edit(sid, what):
+    """The tab fields of names/<sid> for a read-edit-publish by one of the kernel's three writers of the
+    record (_set_name, _set_session_color, _set_palette), or None for a record that reads with NO NAME,
+    which is left exactly as it is. Every writer puts the name first, so an empty first field is never a
+    real record: it is another writer's window (a writer that truncates before it writes, as bin/romp's
+    record writer did, behind the tmux after-rename hook that races _set_name's synchronous publish
+    after a live rename) or a damaged file. Padded to four fields and published over, it cost the
+    session its name, cwd and colors while the edit reported success. One re-read after a short pause
+    (_NAMES_REREAD_S, paid only on an empty read; the window is a printf's worth of time) tells a window
+    from damage: a record whole on the second read is edited as usual; one still without a name is
+    reported, naming the sid and `what` went unwritten. A record that cannot be read RAISES
+    (FileNotFoundError when there is none): each caller keeps its own policy for that."""
+    text = (NAMES / str(sid)).read_text()
+    parts = text.rstrip("\n").split("\t")
+    if not parts[0]:
+        time.sleep(_NAMES_REREAD_S)
+        text = (NAMES / str(sid)).read_text()
+        parts = text.rstrip("\n").split("\t")
+        if not parts[0]:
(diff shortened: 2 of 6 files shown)


ksarma:
Merging as shipped. Nothing to fold in.

Reviewed inline: the three kernel writers of `names/<sid>` (`_set_name`, `_set_session_color`, `_set_palette`) now read through `_names_fields_for_edit`. It treats an empty first field as not a record and re-reads once after 50 ms. A second empty read leaves the file byte for byte. The sid goes to the kernel log and the error center, once per user gesture. Each writer keeps its policy for an unreadable record: recolor False, palette skip, rename the real errno. A rename over a nameless record raises, so both entry points report the cause. `POST /color` names which cause it hit. A named record with fewer than four fields is still padded and published, so no path writes a partial line. The 50 ms re-read is a time window, but neither branch gives a wrong outcome. It either edits over a whole record or refuses with the file untouched. `bin/romp`'s `_romp_record` writes a dot temp in the names dir and moves it into place with one rename(2). The failure arm removes the temp and returns 1, which under `set -euo pipefail` aborts the callers as the old failed redirect did. The dot prefix keeps the temp out of `_romp_free_title`'s glob and the picker's listing. The kernel's directory scanners tolerate a leftover as they tolerate `_atomic_write`'s temps. Fixtures are synthetic, and the added lines carry no private strings.

Optional follow-up: the four no-name tests in `test_kernel_names.py` and `test_color_route.py` each sleep the real 50 ms, whereas the color module stubs the module's time name.

The new tests are red at the base and green at head. At the base with the PR's four test files, the three pytest modules gave 16 failed, 102 passed. The seven new tests were red for the stated reasons, and the absent-record control was green. The bats tests failed at the empty count in every run. At head: 108 passed, 10 subtests, and the bats tests green in every run through the file's mock tmux under a private state dir. CI is green on all eleven checks at head.

Review run by the maintainer automated review session; verified against head 3d95fa72.

PR #1216, opened 2026-09-09
Pane shim: a pane hidden after the user has looked at it no longer raises the stale banner

ksarma:
## Symptom

On the phone, a pane the user had looked at and then switched away from could put the "connection lost, reload" banner over a dashboard that was working. The phone shell shows one pane at a time and hides the rest with `display:none`; the hidden pane's throttled watchdog found its own socket quiet and raised. The 2026-08-15 fix (a hidden pane never raises the banner) covered a pane hidden since load and not a pane hidden after the user had looked at it. The same happens on the desktop when a pane is toggled off the rail after it has been shown.

## Cause

The shim's `paneHidden()` had one witness, the zero-viewport probe (`window.innerWidth === 0 || window.innerHeight === 0` inside a frame). In Chromium a `display:none` iframe keeps the size of its last show, so once a pane has been shown the probe reads it as visible for good; only a never-shown pane reads 0. The same probe is mirrored in perf telemetry's `hidden_pane` row and in the skeleton-prefetch gate in render.ts (#1017), so all three consumers gave the same wrong answer for a pane shown once and then hidden.

Measured in the browser legs below: in Chromium the desktop shell's hidden frame reads 600 x 400 after the hide (the phone shell's frame keeps its own size the same way), so the probe says shown; in Firefox the same frame reads 0 x 0, so the probe says hidden, but Firefox does not run an IntersectionObserver inside a `display:none` frame, so any word the pane published before the hide stays at its last value.

## Fix

The rule, stated once in paint-gate.ts: whether the user can see a pane has two witnesses, and every consumer of the question reads their union.

- The witnesses. The shim's zero-viewport probe (right for a pane hidden since load in any browser, and in Firefox always) and the pane's own two measures, `document.hidden` and its IntersectionObserver over its list (right for a pane hidden after a first show in Chromium).
- The publishers. Every frame that measures its own visibility publishes `document.hidden || !intersecting` as `window.__rompPaneHidden`, on its own events (the observer's callback, `visibilitychange` on both arms, the paint release), never on a timer, and nothing until the observer has spoken: a page loaded in a background tab gets no observer callback until its first rendering step after the return, so the return's `visibilitychange` runs first, and a word published then would be `document.hidden` alone, which says visible for a `display:none` pane; the probe is right at that moment (a never-shown frame reads 0 everywhere). paint-gate.ts gains `publishPaneHidden` and `PaneHiddenHost`; feed.ts and the Outline pane call it as the first line of `releasePaint` and from the hidden arm of `visibilitychange`; the timeline panel gains `_publishPaneHidden` on its hold's two events; the chat page, which gates no paint, gets chat-visibility.ts (one IntersectionObserver over the body plus `visibilitychange`), installed once at top level in render.ts.
- The consumers. The shim's `paneHidden()`, perf telemetry's `hidden_pane` and render.ts's prefetch probe say hidden when the probe OR a published word of `true` says so, never the word first: a stale `false` must not override a zero viewport.
- The paint gate's observer word widens to `boolean | null`: `intersecting === false` holds the paint, null holds nothing (the same as the old `true` default), so a pane can tell "not measured yet" from "on screen". The `hidden_pane` clause in docs/reference.md says what the row now means.

This builds on the paint gate from #1016 (its two measures are the second witness) and touches the prefetch gate from #1017 (the third consumer of the probe). Left unchanged: the Outline pane's 15 s live-clock skip keeps its own probe (a pane hidden after a first show gets a tick nobody sees; the comment there says so). On a standalone page (a bare /chat or /feed, no shell) the shim reads the same word: a page in a background tab holds its own reload bar until the tab is visible again, and if the connection is still stale on the return, the foreground redial's arm raises it. The VS Code webview publishes a word no shim reads.

Open PRs on the same files, none overlapping textually: #1126 edits the shim's JavaScript in three other places (the nearest eight lines below `paneHidden`); #1119 and #1202 add render.ts imports within a few lines of this one; #1057 inserts a docs/reference.md paragraph nine lines below the `hidden_pane` clause; #1198 edits romp-timeline-view.js elsewhere.

## Tests

Every leg below is executing and was red on the unchanged sources, except the pins listed last:

- `tests/test_kernel_disconnect_banner.py::test_a_hidden_pane_never_raises_the_stale_banner` runs the served `paneHidden()` under node over window stand-ins, eight verdicts: hidden since load, shown, a viewport with a word of `true`, a viewport with a word of `false`, a stale word of `false` with a zero viewport, a standalone page with and without a word, a non-`boolean` word. Before: a viewport with a word of `true` read shown (`AssertionError: False is not true`).
- `tests/test_pane_hidden_word_browser.py` (new; node driver `tests/pane_hidden_word_browser.js`) drives real browsers through the shell's two hiding mechanisms, with the rules lifted from the served landing page and the served shim's `paneHidden()` in the frame: shown, hidden after a first show, shown again, and a second frame hidden since load, under the desktop rail's wrapper rule and the phone shell's `m-on` switch. Readings here: Chromium, the desktop shell's hidden frame 600 x 400 and the phone shell's 984 x 150, probe false, word true; Firefox, hidden frame 0 x 0, probe true, word stale at false; the never-shown frame 0 x 0 in both, word true (Chromium) or unset (Firefox). Before: the driver cannot bundle the chat page's publisher (module absent). Skips with a reason without playwright (CI installs no browsers). WebKit was not measured (its playwright build needs system libraries this machine lacks); in any browser the union reads hidden whenever either witness does, so the change cannot raise a banner today's probe suppresses.
- `ui/webview/paint-gate.test.ts`: `publishPaneHidden` over a host (each of the four pairs equals `paintHeld(d, i, true)`), and the null word (the paint proceeds; nothing is published on either arm until the observer speaks). Before: the import does not bundle, and `paintHeld(false, null, true)` was true.
- `ui/webview/outline-visibility.test.ts`: the Outline's own wiring lifted from fleet.ts (the state, the observer, `releasePaint`, both `visibilitychange` listeners and `render()`'s head through the gate), transpiled at run time and run over the pure module with the page stood in: held frames and the newest painting once on show; the word unset until the observer's first entry, hidden after a first show, the tab's blink over a `display:none` pane, both arms after the first entry. Before: the import does not bundle. Removing the publish line, the hidden arm or the null start from fleet.ts turns the run tests red.
- `ui/webview/feed-hidden-paint.test.ts`: the feed's own wiring lifted from feed.ts the same way and run: nothing before the observer's first word, the union on every event, the hidden arm publishing without a release, the paint held meanwhile. Before: red on the anchor (feed.ts started the word at `true` and published nothing).
- `ui/webview/skeleton-tabs-wiring.test.ts`: render.ts's `paneHidden()` lifted and run over the same eight stand-ins as the shim's. Before: a viewport with a word of `true` read shown.
- `ui/timeline-hidden-hold.test.ts`: the panel publishes on its two events and not before the observer speaks; without an observer it never publishes. Before: `_publishPaneHidden is not a function`.
- `ui/webview/perf-telemetry.test.ts`: `installPerfTelemetry("chat")` under a framed window: `hidden_pane` follows the word, a zero viewport wins over a stale word, a standalone page's word counts. Before: `hidden_pane` stayed false with the word true.
- `ui/webview/chat-visibility.test.ts` (new): the chat publisher over stand-ins (nothing before the observer's first word, then the union on every event; no observer or no body installs nothing). Before: module absent.
- Pins that follow the code, for the lines no harness runs: the render.ts install of the chat publisher (`chat-visibility.test.ts`), the prefetch call's union text (`skeleton-tabs-wiring.test.ts`), and the feed's and Outline's wiring text (`feed-hidden-paint.test.ts`, `outline-visibility.test.ts`), which the lifted runs above hold in behaviour as well.

Runs on the final tree: `npm run typecheck` clean; `npm test` 3760 passed, 0 failed; `pytest -n 8 tests/` 10226 passed, 41 skipped, 1671 subtests passed (2 min 19 s), the skips including this PR's webkit leg, which is not installed here.

Tier: fix


ksarma:

Reviewed inline: all three consumers of the hidden-pane question read the zero-viewport probe OR a published word of true, never the word first. The consumers are the served shim's `paneHidden()`, perf telemetry's `hidden_pane`, and `render.ts`'s prefetch probe. So the probe covers Firefox's stale false under a zeroed frame, and a non-boolean word is ignored. The publishers are `publishPaneHidden` in `feed.ts` and the Outline pane, the timeline's `_publishPaneHidden`, and the new `chat-visibility.ts` installed once in `render.ts`. They fire only on the observer callback, visibilitychange on both arms, and the paint release. No timer was added. The paint gate's null word holds nothing, the same as the old true default, so panes that never publish behave as before. Real Chromium and Firefox runs here reproduce the body's readings: a hidden frame keeps 600 x 400 and 984 x 150 in Chromium with the word true, and reads 0 x 0 in Firefox with a stale false the probe covers. A visible pane with a stale connection still raises. The fix label fits: the change corrects the detection behind the 2026-08-15 rule, adds no user-facing capability, and carries a before-failing test on every leg. Added lines carry no private strings, hostnames or home paths. The open PRs the body lists touch disjoint hunks.

Record only, no change needed:
- `ui/timeline-hidden-hold.test.ts:266` and `ui/webview/feed-hidden-paint.test.ts:18`: the body's stated base failures differ from the observed ones (an assertion on the observer's first word; a bundle error on the missing export). Both legs are red at the base.
- `tests/test_pane_hidden_word_browser.py:122`: WebKit is unmeasured. The original report was the iOS phone shell, so a measurement there is a worthwhile follow-up.

Red-first at origin/main with the PR's test files: `test_kernel_disconnect_banner.py` 1 failed 13 passed (False is not true on the word-of-true case); skeleton-tabs-wiring 15/17, perf-telemetry 33/34, timeline-hidden-hold 10/12. The paint-gate, outline-visibility, feed-hidden-paint and chat-visibility files fail to bundle on the missing export or module, and the browser leg fails on the absent module. At head: 90/90 across the seven node files, 14/14 shim tests, browser leg 2 passed with webkit a clean skip, typecheck clean. CI green on every required check at head.

Review run by the maintainer automated review session; verified against head ed363d8e.

ksarma:
Head note: the head moved from ed363d8e to 7d96c21d by one empty commit (no tree change; the PR diff against the merge-base is byte-identical), pushed to re-trigger the merge evaluation. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 7d96c21d.

ksarma:
Head note: the head moved from 7d96c21d to 53421664 by a merge of main into the branch (this session updated the branch so the required check runs against main re-greened by #1226). The PR's added and removed lines against the new merge-base are identical to the reviewed ones; only the context around the render.ts import hunk changed with main. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 53421664.

PR #1217, opened 2026-09-09
Auto-reload: the notice raised as the last pending ship retires is shown again by the page that follows the reload

ksarma:
## Symptom

A kernel restart during an upload can lose the notice that the upload failed. The dashboard reloads itself on a restart, and the chat pane holds that reload while a file is still shipping. When the relaunched kernel nacks the re-shipped file (it could not save it), the page shows the nack toast (the file couldn't be saved on the kernel, so it was not attached; your message was NOT sent) for a fraction of a second and then reloads. The fresh page shows no chip, no attachment and the draft back in the composer, and nothing says that the file was not saved or that the held message never went. The two other notices raised at the same moment go the same way: dismissing the last pending chip with a send held on it (the held message was NOT sent), and the last ship landing on another tab (the held message was not sent; review it there).

## Cause

The last pending ship's retirement is the event that ends the reload hold, and each of the three notices is raised in the same task as it. The dropSaveFailed handler in render.ts retires the ship, ends the hold (endReloadHoldIfIdle, which calls __rompReload.ended()) and raises the toast after that; the dismiss handler for the last pending chip and the other-tab ack raise theirs just before their own endReloadHoldIfIdle. ended() re-tries on a setTimeout(0), finds the pane idle and fires: persist(), then location.reload(). persist() runs window.__rompPersistForReload, which the page installs as persistScrollForReload: the scroll record alone. A warnToast is a DOM element that lives 12 s and dies with the page. The fresh page's loss toast reads shipsInFlight, which persistDrafts (called from retirePendingShip) had already emptied, so it has nothing to say.

## Fix

- New pure module ui/webview/reload-notices.ts. liveNotices(box) reads the texts of `#warn-toasts .warn-toast:not([data-ephemeral]) .warn-toast-msg` in DOM order, blanks dropped, none without the container. keepReloadNotices(store, texts) writes them under the sessionStorage key romp:reloadNotices, or removes the key when there is nothing to say (a reload the browser refused leaves its record behind; the next core reload with no toast on screen clears it rather than replaying it). takeReloadNotices(store) returns the non-empty strings and removes the key whenever it is present. A store that throws is left alone.
- render.ts: persistNoticesForReload() keeps the live toasts, and persistForReload() (the scroll record, then the notices) is installed as window.__rompPersistForReload. pagehide keeps persistScrollForReload alone, so a navigation of the user's own replays nothing; the loss toast fires once for the same reason. The load-time replay follows the lostShips block directly: the loss toast first, then what the last page was saying, with the record taken out in the same call.
- warnToast returns its element, and ephemeralWarnToast(msg) sets data-ephemeral on it. It replaces the one refusal that reports a state rather than an event, at both of its sites (the staged strip's Send now and the empty send that releases a staged stack): the refusal that says the session isn't reachable and the staged sends stay staged. Its predicate is hostIsDown (a remote host's tunnel) or isProvisionalId (a tab still being created), and the fresh page shows that state for itself (the host mark and the staged strip; a provisional tab does not survive a reload), so a replay would be redundant at best and stale at worst. Every toast that reports what happened to a send or a file stays unmarked, since what it says is as true after the reload as before: the nack, the dismissal and the other-tab ack (the three the reload wipes), the refusal at send time on a disconnected host (this message was not sent and is still in the composer), and the re-established-connection toast about a tag edit (a past event).
- Storage: the record rides sessionStorage like the scroll record, for the reason #1104 gave: the persisted webview state is localStorage on the served page, shared by every dashboard tab of the origin, so one tab's nack would replay in another. The record is text only: a toast has no action beyond its dismissal today, and a future toast with one would replay as its words alone (the module comment says so).
- No kernel change. No docs change: the docs carry no auto-reload prose to amend.

## Tests

- ui/webview/reload-notices.test.ts, 8 cases under node --test. Six execute the module: texts in order, blanks dropped, none without the container; the reading asks for the toasts without the ephemeral mark; a record is kept only when there is something to say, and a page with no toast clears one left behind; taken once, strings only, the key removed whatever it held, junk reads as none; kept then taken round trip; a refusing store throws nothing. Two pin the render.ts wiring the way reload-restore.test.ts pins the scroll record's: the hook, pagehide, the replay's place after the loss toast, the ephemeral mark at both of its sites, and the nack, the dismissal and the other-tab ack unmarked. Before the change the test bundle does not build: Could not resolve "./reload-notices".
- tests/test_ship_reship.py::NackNoticeSurvivesReload, served: a real Chromium against a real kernel through the module's existing _ShipLab. A file ships to a SIGSTOPped kernel and a send is held on the gate; a toast wearing the ephemeral mark is put on screen (a stand-in with warnToast's shape, since the lab has no unreachable host for the page to raise its own); the drops path is made a regular file and the kernel relaunched; the reconnect re-ships, the relaunched kernel nacks (_save_dropped_file fails), the held reload fires, and the fresh page's #warn-toasts is read at once. It asserts that the old page saw the nack with the marked toast on screen, the reload landed on the new boot id, the fresh page shows the nack exactly once and neither a loss toast nor the marked toast, the chip and the attachment are gone, the record left sessionStorage, a navigation of the user's own with the replayed toast still on screen replays nothing, and the draft is still in the composer. Before the change the old page shows the nack and the fresh page's toasts read as the empty string, where the assertion fails. With the hook set back to the scroll record alone, the selector without its ephemeral exclusion, or the replay raising each text twice, it fails at the corresponding assertion (each checked). It skips loudly without the extension deps or a browser, as ServedWedge and ReloadLossToast do. The dismissal and the other-tab ack are not executed: they ride the same reading and are pinned unmarked; the nack path is the one run.
- tests/test_ship_reship.py::SourcePins::test_the_notices_on_screen_ride_the_cores_reload_and_are_shown_once pins the hook, pagehide, the writer, the reader and the replay's place after the loss toast; red before the change.
- ui/webview/reload-restore.test.ts: the hook pin follows the code (persistForReload installed; pagehide still persistScrollForReload).
- At a62bdcb9: npm run typecheck clean; npm test 3755 pass, 0 fail; pytest tests/ -n 8: 10226 passed, 40 skipped, 0 failed. tests/test_ship_reship.py alone: 9 passed, three consecutive runs.

Tier: fix

### DIFF
--- a/tests/test_ship_reship.py
+++ b/tests/test_ship_reship.py
@@ -16,7 +16,7 @@ The fix, both faces:
     the page, so the ship NAMES persist beside the drafts and the next load says LOUDLY what
     was lost — never a silent vanish.
 
-Two guards here:
+The guards here:
   * SourcePins — runs everywhere, CI included: the wiring above, pinned in the sources (the
     webview-side twins live in ui/webview/pending-attach.test.ts).
   * ServedWedge — the executed guard: boots the hermetic kernel, opens the real /chat page,
@@ -25,6 +25,10 @@ Two guards here:
     reconnect re-ships, the thumbnail lands, and the held send fires. Plus the regression leg:
     a normal ship+send against the restarted kernel behaves exactly as before. Skips LOUDLY
     when the extension deps or a playwright browser are absent (CI installs no browsers).
+  * NackNoticeSurvivesReload: the same wedge with the relaunched kernel unable to save the
+    re-shipped file. Its nack retires the last pending ship, which lets the restart's held
+    reload fire on the next task, and the notice the nack raised (the file was not saved, the
+    held message not sent) must be shown again by the fresh page, once.
 
 All fixtures synthetic.
 """
@@ -96,6 +100,21 @@ class SourcePins(unittest.TestCase):
     def test_a_reload_loss_is_loud_never_a_silent_vanish(self):
         self.assertIn("shipsInFlight: [...pendingShips.values()].flat().map((p) => p.name)", RENDER)
         self.assertIn("still uploading when this page reloaded, so it was NOT attached", RENDER)
+
+    def test_the_notices_on_screen_ride_the_cores_reload_and_are_shown_once(self):
+        # the nack is raised in the same task as the reload hold's ending event, just after endReloadHoldIfIdle (the
+        # dismissal of the last pending chip and the ack landing on another tab, just before it), one task before the
+        # owed reload fires, so the toast was never read and the loss toast above has nothing to say (shipsInFlight is
+        # already empty). The core's synchronous hook keeps the toasts on screen in this tab's sessionStorage and the
+        # fresh page shows them once, after the loss toast (reload-notices.ts); pagehide keeps the scroll record alone,
+        # so a load the user asks for replays nothing. NackNoticeSurvivesReload executes the nack's.
+        self.assertIn("(window as any).__rompPersistForReload = persistForReload;", RENDER)
+        self.assertIn("function persistForReload(): void { persistScrollForReload(); persistNoticesForReload(); }", RENDER)
+        self.assertIn('window.addEventListener("pagehide", persistScrollForReload);', RENDER)
+        self.assertIn('keepReloadNotices(sessionStorage, liveNotices(document.getElementById("warn-toasts")))', RENDER)
+        self.assertIn("for (const text of takeReloadNotices(sessionStorage)) warnToast(text);", RENDER)
+        self.assertLess(RENDER.index("shipsInFlight: [] });"), RENDER.index("takeReloadNotices(sessionStorage)"),
+                        "the loss toast first, then what the last page was saying")
 
 
 def _free_port():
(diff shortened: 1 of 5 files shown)


ksarma:
Merging as shipped. One follow-up to file.

Reviewed inline: the reload core runs `persist()` synchronously before `location.reload()`, and `ended()` defers one task. So the nack, the dismissal and the other-tab ack are in the DOM when `persistForReload` reads them; the nack is raised after `endReloadHoldIfIdle` in the same handler. `liveNotices` reads only the message spans of toasts without `data-ephemeral`. `keepReloadNotices` removes the key when the page has nothing to say, so the next core reload clears a refused reload's record instead of replaying it. `takeReloadNotices` removes the key as it reads, and the replay runs once, directly after the loss toast. `pagehide` keeps only the scroll record, so a navigation of the user's own replays nothing. Every store access is wrapped, so a store that throws breaks nothing. I classified every `warnToast` site in `render.ts`. The two marked sites are the reachability refusal under `hostIsDown` or `isProvisionalId`. The nack, dismissal, other-tab ack, disconnected-host refusal and tag-edit toast stay unmarked and replay truthfully. The record is text only, as stated. I read the served leg `NackNoticeSurvivesReload` but did not run it. It signals only pids the lab recorded. Its assertions cover the nack shown once on the fresh page, no loss toast, the record gone, and a user navigation replaying nothing. The `fix` label fits: a lost user-facing notice with red-first tests, and the new module applies the scroll record's idiom to a second record, matching #1200.

Follow-up:

- `ui/webview/render.ts:15451`: the edit-mode staging refusal is unmarked and `composerEdits` is never persisted. A refusal raised within 12 s of a core reload therefore replays on a page with no edit in progress. The siblings at 15450, 15452 and 16478 also report states; their replay is redundant rather than false. The PR body's count of one state refusal undercounts. Route them through `ephemeralWarnToast` and update the count pins at `reload-notices.test.ts:109-112`.

Red-first at main with the PR's test files: the node bundle fails to build because the module is unresolved. With the module dropped in, the three `render.ts` pins fail (`reload-notices.test.ts:172` and `:187`, `reload-restore.test.ts:70`). `SourcePins::test_the_notices_on_screen_ride_the_cores_reload_and_are_shown_once` fails at its first `assertIn`. At head: typecheck clean, `npm test` 3755 pass, `pytest tests/test_ship_reship.py` 6 passed with the three served legs deselected. CI 11 of 11 green at a62bdcb9.

Review run by the maintainer automated review session; verified against head a62bdcb9.

ksarma:
Merging as shipped. One follow-up to file.

Reviewed inline: the reload core runs `persist()` synchronously before `location.reload()`, and `ended()` defers one task. So the nack, the dismissal and the other-tab ack are in the DOM when `persistForReload` reads them; the nack is raised after `endReloadHoldIfIdle` in the same handler. `liveNotices` reads only the message spans of toasts without `data-ephemeral`. `keepReloadNotices` removes the key when the page has nothing to say, so the next core reload clears a refused reload's record instead of replaying it. `takeReloadNotices` removes the key as it reads, and the replay runs once, directly after the loss toast. `pagehide` keeps only the scroll record, so a navigation of the user's own replays nothing. Every store access is wrapped, so a store that throws breaks nothing. I classified every `warnToast` site in `render.ts`. The two marked sites are the reachability refusal under `hostIsDown` or `isProvisionalId`. The nack, dismissal, other-tab ack, disconnected-host refusal and tag-edit toast stay unmarked and replay truthfully. The record is text only, as stated. I read the served leg `NackNoticeSurvivesReload` but did not run it. It signals only pids the lab recorded. Its assertions cover the nack shown once on the fresh page, no loss toast, the record gone, and a user navigation replaying nothing. The `fix` label fits: a lost user-facing notice with red-first tests, and the new module applies the scroll record's idiom to a second record, matching #1200.

Follow-up:

- `ui/webview/render.ts:15451`: the edit-mode staging refusal is unmarked and `composerEdits` is never persisted. A refusal raised within 12 s of a core reload therefore replays on a page with no edit in progress. The siblings at 15450, 15452 and 16478 also report states; their replay is redundant rather than false. The PR body's count of one state refusal undercounts. Route them through `ephemeralWarnToast` and update the count pins at `reload-notices.test.ts:109-112`.

Red-first at main with the PR's test files: the node bundle fails to build because the module is unresolved. With the module dropped in, the three `render.ts` pins fail (`reload-notices.test.ts:172` and `:187`, `reload-restore.test.ts:70`). `SourcePins::test_the_notices_on_screen_ride_the_cores_reload_and_are_shown_once` fails at its first `assertIn`. At head: typecheck clean, `npm test` 3755 pass, `pytest tests/test_ship_reship.py` 6 passed with the three served legs deselected. CI 11 of 11 green at a62bdcb9.

Review run by the maintainer automated review session; verified against head a62bdcb9.

PR #1218, opened 2026-09-09
Chat tabs: `ready` sends no tab strip of its own; the connect push's strip is the one source and `ready` clears its slot

ksarma:
## Symptom

A dead session the user chose to view read-only (View read-only in the revive prompt) disappears from the tab strip on every page load, including the reload a kernel restart prompts, and comes back on the pusher's next cycle. Its unsent draft is kept and returns with it, so what the user sees is a tab that blinks out and returns a moment later. On a browser dashboard it returns at the end of the strip rather than in its place (see Cause; read from the code, not run in a browser).

Since #1017 there is a second symptom on the same path: a chat page whose socket was sent the tab strip before its bundle had a message listener (the pusher fires from the moment the socket opens, while the bundle may still be downloading) gets no strip at all at `ready` while the strip is unchanged, and paints no tabs until the strip changes or the dedup repost window (60 s) passes.

## Cause

The WS `ready` handler sent a second tabOrder frame after its connect push, and the two frames list different sets:

- the connect push (inside `_push`, chat clients only) lists `_chat_tab_sessions`: every living session plus the dead sessions the user kept open read-only;
- the handler's own frame listed `_ordered_alive`: living sessions only, from a second off-cycle `_tmux_sessions()` read.

The client treats a listed id that a later frame omits as a close unless that frame affirms it live (`render.ts` `applyTabOrder`, the omitted path; `tab-order.ts` `retainLiveOmitted` spares live sids only). A kept-open dead session is never in `live`, so the handler's frame closed every kept-open tab the connect push had just listed, at every `ready`.

The chat, feed, sessions-pane and timeline pages load `federation.ts`, which absorbs a local tabOrder frame into the viewer's stored arrangement (`absorbHostReport`: `pruneViewOrder` drops an id whose host is reporting and no longer lists it, `writeViewOrder` persists the result to localStorage). The living-only frame therefore pruned every kept-open id from the arrangement on whichever page received it, and the pusher's next strip re-added them at the end (`adoptArrivals`). This paragraph is read from the code, not observed in a browser.

#1017 (lczh) moved that second frame from a raw `client["send"]` onto `_send_tab_order` and the `("taborder",)` dedup slot, which fixed the slot bypass and the hand-recorded views seq; this PR builds on that routing. It left the list mismatch and the second liveness read in place, and it exposed a third fault: `_client_reset_chat_base` (the `ready` reset) pops the `("chat", sid)` and `("status", sid)` slots but not `("taborder",)`, so on a socket served a strip before the bundle listened, the connect push's frame at `ready` deduped against the pre-listener frame, and, once routed through the same slot, so did the handler's own. The raw send it replaced always delivered, which is why the loss did not show before.

## Fix

- `kernel/kernel.py`, the `ready` arm: the block that read `_tmux_sessions()`, built `_ordered_alive`, resolved the reconnect flag (a no-op there by its own comment) and called `_send_tab_order(client, _o, _tabs, _tm)` is gone. The connect push's frame is the one strip a `ready` produces, and `_send_client` already captures its views seq for the caps frame.
- `_client_reset_chat_base` also pops the `("taborder",)` slot: a renderer that just evaluated holds no strip either, so the connect push's frame goes out whatever this socket was sent before.
- Docstrings follow the code: `_tab_order_frame` names its three senders (the pusher's tabs-first send, which is also the connect push a `ready` triggers; `_push_session_now`; `_confirm_close_now`) and says why the `ready` handler sends none of its own; `_resolve_reconnect` no longer lists `ready` among the senders that consume the reconnect flag (the connect push inside `_push` does that at `ready`); `_ordered_alive` no longer calls itself the source of the connect-time strip (nothing in the kernel calls it now; tests still do).

Nothing changes for the caps frame: for a chat client `viewsSeq` comes from the connect push's frame through `_send_client`'s capture; a push that served no views blob falls to the store's current seq as before. `_resolve_reconnect`, `_send_tab_order` and `_send_chat_or_status` stay as #1017 wrote them; the connect push resolves the reconnect flag itself and the `ready` reset pops it.

Non-chat pages (feed, the sessions pane, timeline, shell) received the handler's frame too, since it went to every app while `_push` sends tabOrder to chat clients only. None consumes it: feed.ts learns `selfHost` from the feed frame (`build_feed` stamps it), and the sessions pane, timeline and shell have no tabOrder branch; federation.ts's absorption of a local tabOrder feeds the merged tabOrder it re-emits, which only the chat page handles, and the arrangement write described under Cause.

If a `ready`-time strip is wanted for skeleton delivery after all, the alternative is to build it from `_chat_tab_sessions` through `_send_tab_order` and still pop the slot; that keeps the second read and the dedup work, so this PR removes the frame instead. One residual of the removal: when `_chat_tab_sessions` raises inside `_push` (it also runs `_sessions(now)` and `_gc_session_order`), `_push` logs the traceback and returns, so that `ready` yields no strip and no session frames until the pusher's next cycle, where the removed frame read only `_alive_sessions` and could have produced a living-only strip. A strip with no session frames behind it is of little use, and the pusher retries every cycle.

## Tests

Three new tests in `tests/test_tag_edit_ack.py` (`ReadyStripSource`), on the real `Handler._dispatch_ws` with a chat client, the connect push stubbed to send the frame `_push` builds (living plus kept-open), and the liveness reads the ready arm used to make (`_tmux_sessions`, `_ordered_alive`) pinned, so the failure before the change is the same with or without tmux. Each fails on the unchanged kernel:

- `test_a_kept_open_tab_the_connect_push_lists_is_never_omitted_by_a_frame_at_ready`: every strip at `ready` lists the kept-open tab and, by `applyTabOrder`'s rule, no frame closes a tab an earlier one listed. Before: the second strip lists the living session only.
- `test_ready_clears_the_tab_order_slot_so_the_connect_pushs_frame_goes_out`: with the slot armed with the connect push's signature (the pre-listener send), exactly one strip arrives and it is the connect push's. Before: one strip arrives, the handler's own, without the kept-open tab.
- `test_an_unchanged_strip_still_reaches_a_renderer_served_before_its_listener_existed`: no kept-open tab, slot armed the same way; exactly one strip arrives. Before: none, both frames deduped.

`tests/test_kernel_tabs_first.py`: `test_connect_ready_handler_also_sends_tabs` (a source pin on the ready arm's call) becomes `test_connect_ready_handler_sends_no_tab_order_of_its_own`, behavioural: with the connect push recorded as a marker, a `ready` on a chat client produces the marker and the caps frame and nothing else. Before: `['_pushed', 'tabOrder', 'caps']`. The count of the ready arm's call literal is dropped; the counts of the three senders' spelling stay.

Each part of the change is pinned separately: with only the slot pop reverted, the two slot tests fail and the other two pass; with only the ready arm's old strip block restored, the kept-open test, the slot test and the tabs-first test fail while the unchanged-strip test passes.

Adjusted to the one-strip shape: `Capability` (the `_tmux_sessions` pin in setUp/tearDown dropped; five tests expect one tabOrder per `ready`, the caps frame's `viewsSeq` from the connect push's frame, and the store's seq for a push that served no blob, with the `_ordered_alive`-raises arm gone) and `BlobLessConnectPushCaps` (the class-level `_ordered_alive` raise and `_tmux_sessions` pin dropped, along with the stderr redirects; the test that runs the real `_push` pins `_tmux_sessions` itself, as `Capability`'s end-to-end test does; assertions unchanged) in `tests/test_tag_edit_ack.py`; the ready-arm literals in `tests/test_chat_skeleton_reconnect.py` (now asserts the ready arm sends no strip of its own; the reset's slot-tuple pin names the third slot) and `tests/test_timeline_views.py` (the ready-arm assertion dropped).

Runs: the five touched or reading modules whole (`test_tag_edit_ack.py`, `test_kernel_tabs_first.py`, `test_chat_skeleton_reconnect.py`, `test_timeline_views.py`, `test_chat_delta_resync.py`), 174 passed. Full suite (`pytest -n 8 tests/`) on this head: 10205 passed, 62 skipped, 1671 subtests passed, 0 failed.

Tier: fix


ksarma:
Merging as shipped. One follow-up below; nothing to fold in.

Reviewed inline: at the head, the ready arm calls `_client_reset_chat_base` and then `_push_one`. `_client_reset_chat_base` now pops the `("taborder",)` slot along with the chat and status slots. The connect push's strip is therefore the one strip a ready produces. It goes out even when the pusher served the socket before the bundle listened. That strip lists `_chat_tab_sessions` (living plus kept-open read-only). It goes through `_resolve_reconnect` and `_send_tab_order`, the builder and slot that #1017 routed every sender through. `_send_client` still records each frame's views seq before the dedup decision, so the caps frame's viewsSeq path is unchanged. #1017's body says ready clears the skeleton set, so skeleton delivery never rode the removed frame. That frame came from `_ordered_alive`, a second living-only read. `applyTabOrder` in `render.ts` treats an omitted listed id as a close unless live affirms it. Every kept-open tab therefore closed at each ready on no new information: an inference flap, which this PR removes. One path still leaves a chat page without a strip at ready: `_chat_tab_sessions` raising inside `_push`, which logs the traceback and returns. The pusher's next cycle retries, which is acceptable. Non-chat pages received the removed frame too. The feed, the sessions pane and the timeline have no tabOrder handler, and `federation.ts` only wrote it to the arrangement store, so nothing else loses a frame.

Follow-up: `tests/test_kernel_tabs_first.py:75` pins the no-strip ready for an app "chat" client only. A feed or timeline client has no pin, so a later ready-arm strip for every app would reach the arrangement write again unnoticed.

Tests: at origin/main 26a1ab06 with the PR's test files, the three ReadyStripSource tests and `test_connect_ready_handler_sends_no_tab_order_of_its_own` fail. They fail for the body's reasons: a living-only second strip; no strip under an unchanged signature; the frame list `['_pushed', 'tabOrder', 'caps']`. At the head, the five touched or reading modules pass 174/174. The same 174 pass on current main plus the PR patch, a clean merge. CI is green on every check at 4e548730. The Tier policy check reads fix, so it merges on green.

Review run by the maintainer automated review session; verified against head 4e548730.

ksarma:
Head note: the head moved from 4e548730 to 59911fa8 by one empty commit (no tree change; the PR diff against the merge-base is byte-identical), pushed to re-trigger the merge evaluation. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 59911fa8.

ksarma:
Head note: the head moved from 59911fa8 to 248a6016 by a merge of main into the branch (the owner updated the branch after main was re-greened by #1226); the PR diff against the new merge-base is byte-identical to the reviewed one, so the review above stands for this head.

Review run by the maintainer automated review session; verified against head 248a6016.

PR #1219, opened 2026-09-09
The tap on Apple is the OS's own callback: Declarative Web Push carries the deep link as navigate, and every road that inferred a tap is gone

yatharth:
Continues #1157 (merged). Branch cut from `main` at ded5e987, merged with `main` at 0658d2aa. Reworked 2026-09-10 in two steps: the tap for a **killed** app is the OS's own callback (Declarative Web Push), and the tap for a **live** app is the notification gone from the screen (the vanish road), a trade-off the user accepted the same day.

## The theory, as the device established it (2026-09-09/10, a real iPhone)
- A Home Screen web app that is **alive** (background or foreground) gets **no `notificationclick`** from iOS; a killed app gets the click and the deep link (that always worked). The worker's `push` handler runs and its acks reach the kernel.
- iOS fires **no `notificationclose`** either: zero `closed` acks ever, so a notification gone from the screen cannot tell a tap from a swipe-dismiss.
- Declarative Web Push (this branch) **works for a killed app**, verified on the device 2026-09-10: iOS navigates to the `navigate` link, the page boots on it, the reveal lands (trail: `deeplink hasSid/hasPid`, `[reveal] link … boot`, `[push] landed`). For a **live** app iOS only foregrounds it: no navigation, no worker click, no event anywhere.
- So the one thing a live app leaves for the page is the **screen**: `registration.getNotifications()` lists what is still displayed, and a shown push whose notification is gone is read as tapped: the **vanish road**, first shipped as a1a9d4b5 and taken out over the next point.
- **The accepted trade-off (the user, 2026-09-10).** Because iOS fires neither the click nor the close for a live app, a swiped-away notification leaves exactly the evidence a tapped one leaves, and lands on the next foregrounding as if tapped. The user weighed this and decided that a working background tap is worth an occasional wrong landing after a swipe. This is their explicit call, recorded as such in the code (the ledger block above `_push_ledger`, the `fromLedger` comment in `_LANDING_REVEAL_JS`, `plans/ios-app.md`) and here; it is not an oversight. The foreground case stays non-switching (no wake event reaches the page, so nothing moves until it next comes forward); a killed app uses the native `navigate`.

## What the kernel sends (the declarative round, unchanged)
To an Apple endpoint (`web.push.apple.com`, decided by the endpoint host, never a user-agent sniff), `_push_declarative`:
```json
{"web_push": 8030,
 "notification": {"title": "api", "body": "Two migrations could go first; which one?",
                  "navigate": "https://TESTHOST.example/?push-reveal=bbbbbbbb-1111-2222-3333-444444444444&push-pid=PID-a-0000000000000",
                  "tag": "romp:bbbbbbbb-1111-2222-3333-444444444444",
                  "data": {"sid": "…", "host": "", "kind": "turn", "cardId": "", "url": "/?push-reveal=…&push-pid=…", "name": "api", "pid": "PID-a-0000000000000"}},
 "mutable": true,
 "app_badge": 2}
```
Every member verified against the W3C Push API editor's draft and WebKit's `NotificationJSONParser.cpp` (`app_badge` top-level; `navigate` parsed with no base URL, so absolute). `silent` rides for the quiet card push; `app_badge` only when the leg had a count. The deep link carries `push-pid` so the page settles the row it lands. Non-Apple endpoints keep the imperative payload byte for byte. WebKit dispatches the mutable declarative push as a **`push`** event carrying `event.notification` (verified in source; `pushnotification` never shipped), so the worker feature-detects on `e.notification`.

The `navigate` URL needs the page's origin: the bell's subscribe posts `location.origin` and the kernel keeps it. A subscription made before that **gains it from the requests the device already makes** (`_push_backfill_origin`: the page's `GET /push/pending`, the worker's `POST /push/ack`, the popover's `POST /push/test`, via `_request_page_origin`); only a missing origin is filled, a conflict logs and the first stands. Until then the endpoint gets the imperative shape and a stderr line, never a declarative message WebKit would refuse whole.

## What the worker does, per event (unchanged)
- `push` with `e.notification` (Safari 18.4+ / iOS): ack `shown` by pid, keepalive, under `waitUntil`; **no `showNotification`**.
- `push` with `e.data` (Chrome, Mozilla; or the declarative JSON reaching an older Safari): read the notification block if `web_push === 8030`, else the imperative top level; ack `shown` first; `showNotification`.
- `notificationclick` (browsers that dispatch it): ack `clicked` first, close, focus the shell and post `{romp:'notificationClick', …, pid}`, or `openWindow`. No kept tap, no store, no fingerprint, no `registration.update()`, no `notificationclose` handler (iOS never fires it).

## What the page does
`fromLink` at boot **and** on `pageshow`/`popstate`: validates `push-card`/`push-pid`, lands via `link`, settles `/push/landed`, strips the params, and returns the session it landed. `fromLedger` on boot/visible/pageshow/focus, after the link: `GET /push/pending` lists **every unsettled row** for the device, newest first, each with its stage (`clicked` | `shown` | `sent`; no `closed`), and the page reads the screen once (`getNotifications`, by `data.pid`). The decision table, per row:
- `clicked` → land via `ack`, once by pid; settle `landed`.
- `shown` and still displayed → untouched.
- `shown` and gone → tapped as far as the page can tell: **exactly one** such row lands via `vanish`, silently; settle `landed`.
- two or more gone → nothing lands, nothing shows; each settled `/push/dropped`, so none inflates a later count.
- a displayed notification for the same session as an older gone row → that row was replaced by the per-session tag, not tapped: `/push/superseded` (the kernel does the same at the `shown` ack, one line each).
- `sent` (never acked shown), all displayed, or a screen it cannot read (no `getNotifications`, or it throws) → nothing.
- a link this page landed at the same time (a killed app, navigated by iOS) or a clicked row landed here is the newer word: whatever else vanished is dropped.

No chip, no prompt, no `closed` stage, no timers. The worker's message lands via `sw`. One pid lands once whichever road carries it. `_REVEAL_ROADS = {sw, link, ack, vanish}`.

## Removed (and pinned gone by source tests)
The stored tap `/__romp/tap` and the `tapReplay`/`tapLanded` replay; the worker fingerprint `/__romp/sw` with its `sw-stale`/`sw-update` rows; `registration.update()` on every push; the `closed` stage and the `notificationclose` ack; the offer chip and `/push/dismissed`; the `store`/`offer` roads of `/reveal`; the `pushReveal` message shape. Kept: the push ledger (`{pid, endpoint, sid, host, kind, cardId, name, sentAt, shownAt, tappedAt, landedAt, supersededAt, droppedAt, swVersion}`), the `shown`/`clicked` acks by pid, `/push/landed` | `/push/superseded` | `/push/dropped`, the `[push]`/`[reveal]` lines, the `deeplink`/`sw-message`/`tap-pending`/`tap-pending-land`/`tap-vanish-land`/`reveal-post` diag rows, the wid-mint-in-head fix.

## Tests (fail first)
- **Declarative round** (41 failed on a1a9d4b5): the exact Apple payload and its members, FCM/Mozilla unchanged, per-device pids, the endpoint-host decision, no origin → imperative + a line, subscribe records the origin; the origin backfill by each of the three requests, never overwriting; the worker's declarative event acks and shows nothing, the declarative JSON at `e.data` shows from the notification block, click acks first; the link at boot and gained later on `pageshow`/`popstate`; the browser's click and link roads.
- **Vanish round** (17 failures + 5 errors on fe460d0f): kernel: `/push/pending` lists every unsettled row newest first with its stage, the three settles each stamp their own field, the supersede at the shown ack and its `[push] superseded` line, `/push/superseded|dropped` ride the token, `/push/dismissed` stays gone, `/reveal` admits `vanish` and refuses `store`/`offer`. Node harness (the page's script executed): one vanished lands via `vanish` and settles; two are dropped silently; all displayed, sent-only, no `getNotifications`, or it throws → nothing; superseded; a link boot outranks the vanish; clicked beside vanished; a boot lands a vanished push as a boot; rows without a stage are nobody's; a source pin that the landing site's comment names the conflation, its cause and the decision. Browser (real Chromium, hermetic kernel, four synthetic sessions): three shown pushes, two of three displayed → the missing one lands via `vanish` with no chip element anywhere and `[reveal] vanish` once across the two coming-forward checks; all three displayed → nothing; one of three → nothing lands, both rows dropped, `/push/pending` empty after.
- Full Python suite green locally (5 chunks, ~10.9k tests, 0 failures); the targeted files re-run green after merging `main` at 0658d2aa. One over-strict log-order pin in the click scenario (`ack clicked` before `[reveal] sw`: two connections on two kernel threads) relaxed to what the design guarantees.

## Device check plan (read from the trail, not from memory)
After deploy, press the test button on `api` and:
1. **Killed** app, tap: `[push] test … 201` → `[push] ack stage=shown` → `[reveal] link sid=… boot: parked` + consumed on the pane's ready → `[push] landed`; a `deeplink` row with `hasPid:true` via `boot`. (Verified 2026-09-10.)
2. **Background** app, tap: the app comes forward on `api` with no link: `[push] ack stage=shown`, then on the foregrounding `[reveal] vanish sid=…: delivered` and `[push] landed`; a `tap-pending` row `{rows:1, getNotifications:true, displayed:0, vanished:1}` and a `tap-vanish-land` row. No `[reveal] sw`, no `ack stage=clicked`.
3. **Foreground** app, tap the banner: nothing moves (no wake event reaches the page); the next time the app comes forward the row lands via `vanish` if the notification is gone by then.
4. **The trade-off, seen**: swipe one notification away, then open the app: `[reveal] vanish` for it, the accepted wrong landing. Swipe two away, then open: `[push] dropped` twice and nothing moves.


yatharth:
Heads-up for whichever of #1155 (notification titles) and #1219 lands second — verified on a local merge of the two, notification + browser test files green after: in `_push_test` the payload line combines both sides, `_push_payload(_notify_title(bare_name), body, sid=sid, kind="test", host=host, name=name, pid=pid)` — #1155's title builder with #1219's ledger `pid` — and #1155's inline name block must flatten/clip the shell's `label` (as `_push_session_name` does) before it can stand in for a name.

yatharth:
Correction to the theory this PR opened with, and a second commit (a1a9d4b5) that replaces the offer-chip design.

**No partition.** On the real phone (2026-09-09, 23:30 UTC) the worker's `push` handler ran and its ack reached the kernel one second after the send. The fingerprint gaps were the worker's install wiping its own record on every `registration.update()` the push handler asks for. **iOS does not dispatch `notificationclick` to the worker for a Home Screen web app that is already alive** — the tap just foregrounds the app. A killed app gets the click and the deep link.

**The vanished-notification road.** The one thing the page can read is the screen: `registration.getNotifications()`. `/push/pending` now lists EVERY unsettled row for the device (it named only the newest — another session's push, sent 40 s later, which is what the chip showed). A `shown` row whose notification is gone, with no `closed` ack on record, was tapped: exactly one such row lands silently (`via:'vanish'`); anything else — two or more gone, all displayed, sent-only, an unreadable screen — does nothing visible. A newer displayed notification for the same session supersedes an older row (at the kernel on the shown ack, and on the page for a newer push whose ack was lost); spent rows are `/push/dropped` so they never inflate a later count.

**The chip is removed** at the user's request: the `#tap-offer` element and CSS, the `offer()` code and diag rows, the `/__romp/shown` record, the `offer` road (now refused by `/reveal`), and `/push/dismissed`.

Browser scenario 4 fails on 3b8b60f8 on the outcome (tab still `web`, 0 reveals, the kernel knew only the newest pid, chip element present) and passes on a1a9d4b5; the offer scenario went with the chip. Full python suite green under pytest by chunks (10260 passed, 42 skipped).

yatharth:
Reworked (tip 09c44ecf): the guesswork is out and the tap is the OS's own callback.

- **Why:** the device showed a live Home Screen app gets no `notificationclick` and no `notificationclose`, so a gone notification could not tell a tap from a dismiss. Every road that inferred a tap (stored tap + replay, fingerprint + `sw-stale`/`sw-update`, the vanish comparison, the `closed` stage, `/push/dropped|superseded`) is removed and pinned gone.
- **What replaces it:** an Apple endpoint gets a Declarative Web Push message — `{"web_push":8030,"notification":{title,body,navigate,tag,data[,silent]},"mutable":true[,"app_badge":n]}` — with `navigate` the absolute deep link (`/?push-reveal=<sid>[&push-card=<id>]&push-pid=<pid>`; the bell now records the page origin at subscribe). iOS displays it and navigates the app on a tap; the page lands the link at boot and on `pageshow`/`popstate` and settles the row by pid. Verified against the W3C Push API draft and WebKit's `NotificationJSONParser.cpp`; one deviation from the brief, from WebKit's source: the mutable message dispatches a **`push`** event with `event.notification` (no `pushnotification` ever shipped), so the worker feature-detects on that, acks `shown`, and shows nothing. Non-Apple endpoints keep the imperative payload and the click road.
- **Size:** −1010 lines against the previous tip (`kernel/kernel.py` −177).
- **Tests:** 41 of the new webpush tests fail on the previous tip; unit modules 268/268; browser scenario 2/2 (click road and link road, incl. the open page gaining the link); full Python suite green in three chunks (2898 + 3603 + 3745, 42 skipped).
- **Device check plan** (in the body): three test taps — app killed / background / foreground — read from the kernel log (`ack stage=shown` → `[reveal] link` → `[push] landed`) and the `deeplink` diag row (`hasPid:true`). First toggle This device off and on so the subscription carries the origin.

yatharth:
76d6a0f3: the origin an Apple endpoint's declarative `navigate` needs is now backfilled for subscriptions made before 09c44ecf, from requests the device already makes and with nothing asked of the user. `_push_backfill_origin` runs on the page's `GET /push/pending` (its own endpoint), the worker's `POST /push/ack` (the row's endpoint; the pid is the credential, as for the ack itself) and the popover's `POST /push/test` (before the send, so that test push itself goes out declarative). `_request_page_origin` reads the origin as the request states it: the `Origin` header (a POST carries one; a same-origin GET does not, per Fetch), else the `Referer`'s origin, else a proxy's `X-Forwarded-Proto` + `X-Forwarded-Host`/`Host`, else `Host` alone with the Push API's secure-context scheme (https unless loopback). Only a missing origin is filled; a recorded one is never overwritten, a different one logs `[push] origin conflict … kept=… saw=…` and the first stands; a fill logs `[push] origin recorded …`. The stderr fallback line remains only for an endpoint no request has yet revealed and no longer asks for a toggle; the device plan in the body drops that step. No kernel helper built absolute dashboard links before this (nothing read `X-Forwarded-*`), so the reader is new. Fail-first tests in `OriginBackfill` (7) plus the reworded no-origin test; webpush 100, popover + browser 60, full suite 2898 + 3603 + 3754 passed, 42 skipped.

yatharth:
Tip 01e894e4: the vanish road is back on top of the declarative work, by the user's decision of 2026-09-10. On the device the declarative navigate lands a tap only for a killed app; a live app (background or foreground) is merely foregrounded with no event of any kind, so the page again holds the kernel's unsettled shown rows against registration.getNotifications() on boot/visible/pageshow/focus, after the link and clicked roads: exactly one shown row gone from the screen lands via 'vanish' and settles /push/landed; two or more gone are settled /push/dropped without landing; a newer same-session notification on the screen supersedes an older row (/push/superseded, and the kernel does the same at the shown ack); all displayed, sent-only or an unreadable screen decide nothing. No chip, no prompt, no 'closed' stage, no timers; _REVEAL_ROADS = {sw, link, ack, vanish}. The known conflation is recorded at the landing site, in the ledger block and in the body: iOS fires neither notificationclick nor notificationclose for a live web app, so a swipe-dismiss reads as a tap on the next foregrounding, and the user accepted that in exchange for background taps working. Fail-first (17 failures + 5 errors on fe460d0f), then kernel, node-harness and real-browser tests for the decision table; the full Python suite ran green in five chunks; main (0658d2aa) merged in with two conflicts resolved keeping both sides, and one over-strict log-order pin in the click scenario relaxed to what the design guarantees.

henrypinkard (approved):

PR #1221, opened 2026-09-09
Tests: every tag-edit ack test runs under a state root of its own

ksarma:
Tests only, one file: `tests/test_tag_edit_ack.py`. Every test in the module now runs under a state root it makes, binds, restores and removes itself. No assertion changes.

## Symptom

Under xdist, `tests/test_tag_edit_ack.py` failed with `FileNotFoundError` on `timeline-views.json` while every test in it passed alone (#1176 names `LegacyStoreStampedOnce` among the tests that failed together on one worker). The exposure is per test. A test fails when its first touch of the views file is a direct `write_text` (a legacy, torn, over-cap or foreign fixture written before any `seed()`) and it runs first on a worker after a fixture elsewhere left the shared judge's `STATE` bound to a removed directory.

Measured on this tree with such a fixture run first: `LegacyStoreStampedOnce` (2 of 3), `LegacyTagsStampedOnFirstRead` (1 of 1), `MigrationStampsTheArchivedTag` (1 of 1), `ReaderRestampKeepsTheDiskCap` (4 of 8) and `ReaderRestampUnwritableNamesTheDrop` (5 of 5) fail when each runs as a class; `ReaderRestampUnwritable` passes as a class (its first test seeds, which recreates the directory) and its read-only test fails when run alone. Every failure is the same `FileNotFoundError`; the other classes pass.

## Cause

The judge is one module object per process: `SourceFileLoader(...).load_module()` re-executes into the module already in `sys.modules`, so `km.jd.STATE` is shared with every module an xdist worker collected, and `km._views_path()` is `jd.STATE / "timeline-views.json"` on that shared object. `_atomic_write` recreates a missing parent, so a test whose first touch went through it (a `seed()`, a setter call) recreates the directory; a direct `write_text` does not.

The polluting fixtures are fixed at the cause by #1163 and #1176: they restore the root they found, and `tests/conftest.py`'s `_shared_state_restored` fails any test that leaves the shared roots changed. This module still runs on the shared root, unlinking the views file before and after each test, so its order-independence rests on that check holding across the whole suite. Among the modules that use the shared-root views file in-process, it is the only one whose first touch can be a direct `write_text` (`test_kernel_timeline_split` mkdirs first, `test_tag_federation_v2` writes through `_atomic_write`, `test_views_stale_writer` writes through `_set_timeline_views`); the other in-process users bind a private root per test, the shape used here.

## Fix

- `_own_state(test)`: `tempfile.mkdtemp()` (under the run's private root; the conftest redirects `tempfile.tempdir`), `jd._rebind_state(root)`, and a cleanup on the same test that drops the views path's entries from `_flags_cache` and `_VIEWS_SEQ_FLOOR` (both keyed by `str(path)`, so a root per test would otherwise leave an entry per test), rebinds the prior root, sets the dir back to 0o755 (the unwritable-store tests leave it 0o555 if they fail before their `finally`) and removes it with `ignore_errors=True` (a dir that still resists removal lies under the run's private temp root and is removed with it). It is registered as a cleanup, so it runs after `tearDown`.
- `_Wire.setUp` and `SetterReturnsRefusals.setUp` call it in place of the shared-path unlink; their `tearDown`s no longer unlink the views path either, since the cleanup removes the whole root. The four in-test store resets (`km._views_path().unlink()` inside a test body) are unchanged.
- `WebBootWiring` derives from `_Wire` (it reads the kernel's source only), so no class in the module runs on the shared root.
- Three setUps that reset a kernel global put the prior value back on cleanup: `Capability` (`_UNKNOWN_OPS_SEEN`), `ReaderRestampUnwritable` and `ReaderRestampUnwritableNamesTheDrop` (`_VIEWS_RESTAMP_ERR[0]`; the latter's tearDown reset is dropped).
- The three in-test `mkdtemp` dirs (the names dirs in `Capability` and `BlobLessConnectPushCaps`, the second store in `SeqFloorOutlivesTheCacheEntry`) gain `addCleanup(shutil.rmtree, ...)`.
- The module-level `XDG_STATE_HOME = mkdtemp()` stays: it is the import-time floor the three loads resolve against, and under the conftest it lies inside the run's private temp root.

## Tests

- `OwnStateRoot.test_a_dangling_binding_left_by_an_earlier_test_is_not_inherited_by_a_direct_write` (new). Its setUp binds `jd.STATE` to a removed temp dir before `_Wire.setUp` runs, the shape the polluting fixtures left, with the root it found registered as the first cleanup, so it runs last and the class leaves the binding as it found it (the conftest check compares before and after the test, and passes). The test writes the views file directly and asserts the write landed under a live root of the test's own and that the reader reads it back. On the unchanged module plus this class: `FileNotFoundError: [Errno 2] No such file or directory: '.../timeline-views.json'`, `1 failed`. With the change: `1 passed`.
- The new test covers the `_own_state` call in `_Wire.setUp` only: disabling that call turns it red again (`1 failed`, the same `FileNotFoundError`). Disabling the call in `SetterReturnsRefusals.setUp` leaves the regression test green, and the module with it, because that class's tests first touch the store through `_set_timeline_views`, which recreates the parent. Nothing in the module enforces the helper class by class. The conftest check forbids the dangling binding suite-wide and pins the cleanup's rebind: with `km.jd._rebind_state(saved)` removed from the cleanup, 94 of the module's 95 tests error on that check (`romp_judge.STATE changed from ... to ...`). The cache pops, the chmod, the `rmtree` cleanups and the three global restores have no test of their own.
- The module alone: `95 passed, 4 subtests passed` (94 tests before). With a scratch polluter run first (a test outside `tests/` that rebinds `jd.STATE` to a temp dir and removes it), each of the six classes named above passes, and the read-only test alone passes.
- With `tests/test_episode_boundary.py` under `-n 4`, both orders, twice each: `112 passed, 4 subtests passed`; the pair in one process, both orders: `112 passed`.
- Full suite (`pytest -q -n 8 tests/`) on the committed head: `10248 passed, 63 skipped, 1687 subtests passed` in 71 s.

Tier: docs


ksarma:

Reviewed inline: `_own_state` makes its root with `mkdtemp` under the run's private temp root (conftest sets `tempfile.tempdir`). It binds the root through `jd._rebind_state` and registers the restore with `addCleanup`, so the restore runs after `tearDown`. The restore pops the views path from `_flags_cache` and `_VIEWS_SEQ_FLOOR` by `str(path)`, the key `kernel.py` uses for both. A root per test therefore leaves no entry behind. A probe after the module finds both caches empty at the head. The chmod restore targets the root itself, which the two unwritable-store classes set to `0o555`. `ignore_errors` on the `rmtree` is safe: conftest removes the whole temp root at session end. `OwnStateRoot` registers the found root as its first cleanup, so it runs last and the class passes the conftest `_shared_state_restored` check. The check is live: deleting the cleanup's rebind makes 94 of 95 tests error on it. The three global restores, the three in-test `rmtree` cleanups and `WebBootWiring` as a `_Wire` match the body. The four in-test store resets are untouched. The `docs` label fits a tests-only change. The credit to #1163 and #1176 matches those merges. The heads of this PR and #1218 merge cleanly.

Locally:
- Base module plus the new `OwnStateRoot` class: fails 1 with `FileNotFoundError` on the removed root's `timeline-views.json`; passes at the head.
- Head module: 95 passed, 4 subtests (94 at base).
- Module at the three-way merged tree with #1218: 98 passed.
- No privacy hits in the added lines.

CI at bb86f2c6: all eleven check runs green.

Review run by the maintainer automated review session; verified against head bb86f2c6.

ksarma:
Head note: the head moved from bb86f2c6 to 6b144eb5 by one empty commit (no tree change; the PR diff against the merge-base is byte-identical), pushed to re-trigger the merge evaluation. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 6b144eb5.

ksarma:
Head note: the head moved from 6b144eb5 to 1ae6bb78 by a merge of main into the branch (the owner updated the branch after main was re-greened by #1226); the PR diff against the new merge-base is byte-identical to the reviewed one, so the review above stands for this head.

Review run by the maintainer automated review session; verified against head 1ae6bb78.

PR #1222, opened 2026-09-09
Markdown images that fail are parked, heal on the reconnect event off the DOM, and the page files its bundle build (T291c)

henrypinkard:
The snapback persisted on the fixed bundle (the user restarted the laptop kernel and reloaded; the 3:46 PM recording shows three image captions appearing at the ends of three paragraphs and vanishing again, two to four lines each, matching the capture's ±41.6/62.4/83.2 px rows on three assistant turns). Those are markdown-inline images.

**Root cause.** marked leaves `fig` as `` (the chat rewrites no image source), the browser resolves the path against the page's origin, the kernel answers 404, and the markdown image heal (2026-08-24) re-set that img's `src` on every kernel message: while the reload ran the browser hid the alt text, on the 404 it showed it again, so a streaming turn flipped every such caption on and off at the push rate. The preview fix (1206) moved only relay URLs (`/remote/<host>/file`) off the per-message path; a path URL like these stayed on it, unbounded. Every re-render of the same markdown also minted a fresh img that fetched, failed and flipped once more.

**Fix.** A failed markdown image is PARKED, not retried: the capture-phase listener strips its `src` into `data-md-src`, marks it `md-img-failed`, remembers the URL for the page life, and the browser shows the alt text as a stable inline caption (no src, no fetch, nothing to flip). No per-message path exists for markdown images. The reconnect-class heal (`refreshSettledPreviews`, the same three events: `romp:wsup`, `hostUp`, `romp:hostRelayUp`) probes each parked URL once OFF the DOM with a detached `Image`: a failure changes nothing on screen; a success lands the picture on every parked img with that URL. A sanitizer post-pass (`mdImgPostPass`, registered by render.ts beside the listener) parks a re-rendered img whose URL is known to fail BEFORE the browser fetches it, so a re-render reuses the state.

**The page-load row.** The chat pane files one client-diag row per page load (`pageload`, `distVer`: the `?v=` the kernel stamped on the render.js script the page loaded, 0 in a VS Code webview) so the next report can tell the page's build from the kernel's.

**Tests.** `ui/webview/md-img-park.test.ts`, executed over a fake DOM: a failed image is parked with its caption intact and sixty pushes leave it byte-identical with no probe and no fetch; the reconnect heal probes once per URL off the DOM, a failed probe changes nothing, a successful one lands the picture and un-parks only that URL; the post-pass parks a remembered URL at render and leaves a healed one alone; render.ts installs the listener once, registers the post-pass and files the page-load row. The preview pins follow (no `failedPreviews` registration for markdown images; the heal rides `refreshSettledPreviews`).

**From the adversarial review, folded in before arming.** The park runs on the chat's own markdown only (`md()` and `userMd()` call the post-pass after the sanitize), never through the shared sanitizer: the file viewer rewrites its images' paths to `/file` URLs after its own sanitize, and a park inside the sanitizer removed the `src` that rewrite keys on. The heal re-queries the page on load instead of restoring a snapshot, so a turn re-rendered while the probe ran (its fresh img parked by the post-pass) gets the picture too. A URL the kernel actually serves (`/file`, a remote relay's `/file`, `/media`) gets a bounded probe on the per-message path, three attempts off the DOM so the caption never moves for them, then the reconnect-class heal alone; an origin-resolved path no route serves stays park-only. The tests count `src` writes (zero across sixty pushes; the old code wrote two per push), each builds its own state, and they cover the served-URL budget and the re-render during a probe.

### DIFF
--- a/ui/webview/md-img-park.test.ts
+++ b/ui/webview/md-img-park.test.ts
@@ -1,0 +1,134 @@
+// A markdown-inline image whose fetch fails is PARKED, never re-set per message (T291c, the user 2026-09-09: three
+// captions in a remote session's transcript flipped on and off at the push rate; the 2026-08-24 heal re-set each failed
+// img's src on every kernel message, hiding the alt text while the reload ran and showing it again on the 404). The
+// heal is EXECUTED here over a minimal fake DOM: the capture-phase listener parks the img (no src, the alt text as a
+// stable caption); sixty pushes write the src zero times; the reconnect-class heal probes each parked URL once OFF the
+// DOM, a failed probe changes nothing, a successful one lands the picture on every parked img with that URL on the
+// page at that moment (a re-render during the probe included); a URL the kernel serves gets a bounded off-DOM probe on
+// the per-message path; and the chat's post-pass parks a re-rendered img with a remembered URL before the browser
+// fetches it. Every test builds its own state.
+import { test } from "node:test";
+import * as assert from "node:assert/strict";
+import * as fs from "node:fs";
+import * as path from "node:path";
+
+const RENDER = fs.readFileSync(path.resolve(process.cwd(), "..", "ui", "webview", "render.ts"), "utf8");
+const CSS = fs.readFileSync(path.resolve(process.cwd(), "..", "ui", "webview", "styles.css"), "utf8");
+
+class FakeEl {
+  tagName: string; attrs: Record<string, string> = {}; _cls = new Set<string>(); dataset: Record<string, string> = {};
+  isConnected = true; onerror: any = null; children: FakeEl[] = []; srcWrites = 0;
+  constructor(tag: string) { this.tagName = tag.toUpperCase(); }
+  get src(): string { return this.attrs.src || ""; }
+  set src(v: string) { this.srcWrites++; this.attrs.src = v; }   // a src write is a fetch, and hides the alt text until the response
+  get alt(): string { return this.attrs.alt || ""; }
+  set alt(v: string) { this.attrs.alt = v; }
+  hasAttribute(k: string) { return k in this.attrs; }
+  getAttribute(k: string) { return this.attrs[k] ?? null; }
+  setAttribute(k: string, v: string) { if (k === "src") this.srcWrites++; this.attrs[k] = v; }
+  removeAttribute(k: string) { if (k === "src") this.srcWrites++; delete this.attrs[k]; }
+  get classList() { const s = this._cls; return { add: (...c: string[]) => c.forEach((x) => s.add(x)), remove: (...c: string[]) => c.forEach((x) => s.delete(x)), contains: (c: string) => s.has(c) }; }
+  closest() { return null; }
+  querySelectorAll(sel: string): FakeEl[] { return sel === "img" ? this.children.filter((c) => c.tagName === "IMG") : []; }
+  shape(): string { return `${this.tagName}[${Object.entries(this.attrs).sort().map(([k, v]) => k + "=" + v).join(" ")}].${[...this._cls].sort().join(".")}{${JSON.stringify(this.dataset)}}`; }
+}
+const live: FakeEl[] = [];                            // what document.querySelectorAll sees
+let errorListener: ((e: any) => void) | null = null;
+(globalThis as any).document = {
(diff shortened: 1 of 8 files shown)


ksarma:
Post-merge record: the repository owner merged this PR as 09d40b82 at 359d0322 while this review ran; the review found nothing that would hold a tier-1 fix, and one should-fix below now stands on main. One follow-up to fold in; nothing here holds a tier-1 fix.

Reviewed inline: the park is keyed on the image error event (the capture-phase listener, `preview.ts:808-822`). The heal is keyed on the three reconnect-class events (`romp:wsup` at `render.ts:1808`, `hostUp` at 15097, `romp:hostRelayUp` at 13572). No timer and no src write remain on the per-message path. The only per-push work for a markdown image is the attempt-bounded off-DOM probe for kernel-served URLs. The post-pass runs from `md()` and `userMd()` alone (`render.ts:1178, 1193`). The shared sanitizer and the file viewer's post-sanitize `/file` rewrite are untouched. The detached Image is created in the page document with no crossOrigin, so a same-origin or relay probe carries the kernel cookie, as an in-DOM img does. `probeMdImgUrl.onload` re-queries the live DOM and un-parks only imgs whose `data-md-src` matches the probed URL. Every parked copy on the page therefore lands, including a turn re-rendered during the probe. The pageload row is one clientDiag post per bundle evaluation, carrying `distVer` and `location.pathname` (`/chat` on the served page). The kernel's client-diag handler accepts any `what` and appends the row with no allowlist and no log line. The label fits: a fix with tests that fail at the base for the stated reason. The privacy scan of the added lines is clean. The body paraphrases the recording without quoting anyone.

- **Follow-up**, `ui/webview/preview.ts:818-820`: the served-URL budget is closed over the img element and is lost with it. A streaming turn is rebuilt on every push (`render.ts:10231`, then `bubble.innerHTML = md(...)` at 3195), so the parked img is a fresh element each push. `mdImgPostPass` parks it without a budget (764-770). `retryFailedPreviews` drops the detached entry unrun (718-719; it runs at `render.ts:15088`, before upsert at 15099). Reproduced over the PR's own fake DOM with a `/file` URL that 404s during its streaming turn: one probe fires instead of three. Once the file lands, nothing re-probes until a reconnect-class event. `healMdImgs` (799-805) reads only connected imgs, so a turn outside the tail window is skipped even then and re-parks when it scrolls back. The body's claim of three attempts does not hold for the case the code comment at 771-774 names. Suggested fix: key the budget on the URL, as `mdImgFailed` already does, so a post-passed img inherits the remaining attempts. Then have `healMdImgs` probe the union of `mdImgFailed` and the DOM's parked URLs. The third md-img-park case keeps `s` connected between rounds, so it cannot see this.
- **Follow-up**, `ui/webview/preview.ts:816` and `styles.css:3004`: a parked served-URL caption has no manual path once its probes are spent. A reload is the only recovery. The preview chips in the same file keep a tap as the backstop after their budget (711-714). A delegated click on a parked img that runs `probeMdImgUrl(data-md-src)` would close that gap. For an origin-resolved path the park-only behavior is right.
- **Nit**, `ui/webview/render.ts:1817`: the row also files `path: location.pathname`. The body and commit message name `distVer` only.
- **Nit**, `ui/webview/md-img-park.test.ts:9`: the header says every test builds its own state, but `fresh()` does not reset `preview.ts`'s module-level `mdImgFailed` set. Harmless in the file's order today.

Red-first held. The five PR test files checked out over the base 4a31744c fail 8 of 74, for the reasons the body gives. The eight are the four md-img-park cases plus the four edited pins in pr-links, preview-core, preview-heal and preview-retry-pace. The md-img-park cases fail because the base re-sets src on failure and probes nothing on reconnect. At dca35a20 and at the rebased head 359d0322 they pass 74/74, and `tsc --noEmit` is clean. 359d0322 is dca35a20 rebased onto a7fd4ac3 with a byte-identical diff. At dca35a20 the vscode-extension check was red from two tests the PR does not touch (skeleton-tabs-wiring, staged-list-cap). Those were fixed on main by 5b4623d2, which the rebase carries. CI at 359d0322 was queued when this was written.

Review run by the maintainer automated review session; reviewed at dca35a20 and verified against head 359d0322 (the rebase onto a7fd4ac3 left the PR diff byte-identical; the five test files and the typecheck were re-run at 359d0322).

PR #1223, opened 2026-09-09
Tests: a bare `!` negation in a bats test is refused unless it is the test's last command; three inert sites armed

ksarma:
This is the tests-only part of the closed #1010, filed on its own. #1010 was closed so the tab emoji can return as part of a configurable tab-status design, and the test below was not reviewed there.

## Symptom

Three bats tests pass with the behaviour they assert against. `tests/romp-headless.bats` says `romp end self --now` sends no `"when"`; `tests/romp.bats` says a `new -t` refused over a leftover `ANTHROPIC_API_KEY` runs no `new-session` or `set-environment`; `tests/tmux-status-hook.bats` says `PostCompact` with `trigger=auto` sets neither `waiting` nor `compacting`. Each of those assertions is inert: with `--now` changed to keep sending `"when": "idle"`, the headless test still passes.

## Cause

bats runs a test body under `set -e` with an ERR trap, and bash exempts an inverted command from both. A bare `! grep -q x "$LOG"` followed by another command therefore has no effect on the test's outcome. Only as the test's last command is it checked, through the function's return status, which is why last-line sites work and stop working the moment an assertion is appended after them.

#383 armed nineteen such sites and #403 one more, both by hand. The three above arrived afterwards, each asserting nothing: `tests/romp-headless.bats` in c2a5f844 (the day after #383 merged), `tests/tmux-status-hook.bats` in e0d76b85, `tests/romp.bats` in 989854c6. Nothing in CI refused the form.

## Fix

`tests/test_bats_bare_negation.py` (stdlib only, collected by the existing pytest job) scans every `tests/*.bats` for a bare `!` inside a `@test` block whose next command is not the closing brace, and fails naming each `file:line`. The module docstring records the checked forms and the scan's limits: it is a line scan, not a bash parser, so it does not see `!cmd` without a space, a `! cmd` sharing a line with another command, a bare `!` in `setup()`, `teardown()` or a helper, or a `@test` whose opening line does not end in `{`; and it does report `! cmd || <fallback>`, which errexit checks, so that form is written as `run` + status too.

Five lines are rewritten to `run <cmd>` followed by `[ "$status" -ne 0 ]`, the form #383 and #403 used and the one `tests/romp-manager-tmux-scope.bats` already wraps as `log_lacks`: the three inert sites, and in `tests/romp.bats` and `tests/tmux-status-hook.bats` the last-command negation that follows each, which checks today only by its position and would go inert the moment a line is appended after it. In `romp.bats` the armed lines stay after the three `[[ "$output" ... ]]` checks, since `run` overwrites `$output`; in `romp-headless.bats` the next line is another `run`, so nothing downstream reads the clobbered values. The other 40 bare `!` lines in the suites are each their test's last command; they are unchanged, and the scan names any that stops being last.

## Tests

- `tests/test_bats_bare_negation.py::BatsSuites::test_no_bats_test_negates_a_command_with_a_bare_bang_before_another_command`: on the unchanged bats sources it fails with exactly `romp-headless.bats:78`, `romp.bats:996`, `tmux-status-hook.bats:144`; green after the rewrites.
- `Scanner`, five tests on synthetic snippets: flags a bare `!` before another command; allows it as the last command past comments and blanks; ignores `run !`, `run` + status, `[ ! -s ]` and `if ! ...`; ignores a bare `!` in a helper before the block and in one after it, so a scanner that never closes the block at its `}` fails here; reports from a written file the line the suite test would. Green throughout; each of five hand-made scanner mutations (block never closes, lookahead inverted, no comment skip, `@test` never opens, loosened `!` match) turns at least one of them red.
- Inertness, executed: with `[[ "$_a" == "--now" ]] && _when=""` in `bin/romp` changed to keep `_when="idle"`, so that `end self --now` still sends `"when": "idle"` and every other assertion in the test stays well-formed, the bare form of the headless test passes; the armed form fails at its `[ "$status" -ne 0 ]`; with `bin/romp` restored the armed form passes.
- `bats tests/romp-headless.bats tests/romp.bats tests/tmux-status-hook.bats`: 183 tests, green before and after.
- Full pytest suite on this head (`-n 8`): 10253 passed, 63 skipped, 1687 subtests passed.

Tier: docs


ksarma:
Merging when CI is green; no changes requested. The one red check, vscode-extension, fails the same two tests on main's own run at 4a31744c. This PR touches no extension code. The merge therefore waits on main being fixed and a re-run, not on the author.

Reviewed inline: the bash claim, the scanner, the five rewrites and the scan's blind spots. Under `set -e` with an ERR trap, a bare `! true` followed by another command fires neither, and execution continues. As a function's last command, its nonzero status is returned. That is why last-line sites checked and mid-body sites did not. The scanner opens a block on a `@test` line ending in `{` and closes it on a lone `}`. It reports a `! ` line whose next non-blank, non-comment line is not that brace. The never-closes and inverted-lookahead mutations each turn the Scanner tests red. Each rewrite asserts the same nonzero grep exit the bare form intended. Nothing after a rewritten line reads the clobbered `$output` or `$status`. The romp.bats and hook pairs end their tests, and the headless line is followed by another run. Every documented blind spot is empty across `tests/*.bats`. The other 40 bare `!` lines are each a test's last command. The docs label fits: tests only. The history claims match `git log`. The added lines carry no private strings, hostnames or home paths.

Follow-up at `tests/install-sh.bats:217`: the settings.json heredoc's lone `}` lines end the scanned block early. Lines 220 to 229 are therefore invisible to the scanner. No bare `!` sits there today; this is the documented heredoc limit.

Red-first confirmed by my own run. At origin/main plus the new module, `tests/test_bats_bare_negation.py` fails and names exactly `romp-headless.bats:78`, `romp.bats:996` and `tmux-status-hook.bats:144` (1 failed, 5 passed). At head, 6 passed. The three rewritten bats tests pass locally under a private HOME and tmux socket directory. CI at head: Python 3.10 to 3.13, Shell (bats), tier label and Tier policy green; vscode-extension red on two main-side tests.

Review run by the maintainer automated review session; verified against head 323094f2.

ksarma:
Head note: the head moved from 323094f2 to 0b5baab9 by a merge of main into the branch (the owner updated the branch after main was re-greened by #1226); the PR diff against the new merge-base is byte-identical to the reviewed one, so the review above stands for this head.

Review run by the maintainer automated review session; verified against head 0b5baab9.

PR #1224, opened 2026-09-09
Chat: a send's queued copy carries the id the press minted, so its ✕ cancels exactly that copy

ksarma:
## Symptom

A message sent while the session compacts, sits behind a queue or waits out a usage limit shows as a queued bubble. Until the kernel's drain hands it to the backend, the chat can tell it apart from another copy of the same words only by text and position, and its ✕ posts the bubble's index and body. With two same-text copies queued (two presses, or one from each of two dashboards), a drifted index removed the first copy by body while the client dropped its own entry by id: the surviving copy and the surviving entry wore different ids, and the chat showed a phantom "sending..." bubble, or a chip for the send that had been cancelled.

## Cause

The copy's id is minted where the copy enters the backend's queue (`SdkBackend.send`). A parked send carried no id until the drain (`tests/test_queued_copy_identity.py` stated this as a gap), the client latched an id from the first same-text copy it saw, and no cancel named an id, so text and position decided what the id was meant to decide.

## Fix

The rule: the kernel mints an id for every copy it queues itself; a client that pressed the send supplies its own, so the copy, the bubble and the ✕ agree from the press.

- **Client** (`ui/webview/send-pending.ts`, `render.ts`): `mintQid` mints the copy's id at the press in the kernel's own `echo:` form (32 hex digits, the shape `SdkBackend.send` mints, so `isKernelEchoUuid`, the landed-stamp echo skip and the re-queue prefix test all read it as the kernel's). The composer's three posts (`askFollowUp`, the quote-chip `sendMessage`, the plain `sendMessage`) and the provisional tab's adoption carry it as `qid`; the post still goes first and the entry is registered under the same id right after (`registerOptimistic` gained a trailing `qid`; `newPending` takes it, or mints one for a caller that posts nothing). Our bubble's ✕ carries `qid`, and the `cancelQueued` post names it beside `md`/`idx`/`park`. The entry never latches an id from a kernel copy: a push that shows the send's id nowhere (the kernel has not received the send yet, an older kernel, tmux) reads by text as before ids existed, and the entry records which kernel copy stood in for it on that push (`cover`, re-read every push), so a ✕ on a copy the kernel identified with an id of its own still drops the entry it covered. After an echo cover the held copy wearing the echo's own uuid is hidden with the echo (one copy, two forms). `dropPending` by an id no entry owns and no copy covered removes nothing of ours.
- **Kernel handlers** (`kernel/kernel.py`): `sendMessage` and `askFollowUp` pass `qid=_client_qid(msg, sid, be)`. The id is admitted only in the echo form (`_wire_qid`, `^echo:[0-9a-f]{16,64}$`) and only when the session does not already hold it (not on a parked op, not on a queued copy per `pending_queued_meta`, not on a live echo); anything else is dropped, logged (sid only) and the kernel mints as before. `_send_or_park(be, sid, text, echo=None, qid=None)` parks `("send", text, echo, qid)` (and `("command", text, echo, qid)` for a typed slash command) when an id rides, else the three-slot op as today; `_op_qid` reads the slot length-guarded, so an older mirror and every existing reader are unchanged. The parked chip carries `qid`. The drain (`_deliver_send_batch`, the command arm) and the immediate handover go through `_send_with_id`, which passes the id only to a backend whose `send` takes a `qid` parameter (`_takes_qid`, read from the signature: `SdkBackend`) and the text alone to every other (`TmuxBackend`, `CodexBackend`, a stand-in). The gate is the signature and not a neighbouring capability: `TmuxBackend` exposes `pending_queued_meta` for the chat's stamps and its `send` takes no id.
- **Cancel exact by id**: `_cancel_parked(..., qid=None)` removes the op wearing the id (never the in-flight head) and answers the miss when none does; `_cancel_backend_queued(..., qid=None)` hands the id to `be.unqueue(sid, -1, None, qid=qid)` on a backend whose `unqueue` takes it and keeps the index/body path otherwise; `SdkBackend.unqueue` and `SdkSession.unqueue` take `qid`, locate the copy under the session lock, and drop its own echo by the shared key and no other (an echo already gone leaves the text match unrun, so a same-text neighbour's echo stays). The three `cancelQueued` arms pass `_wire_qid(msg)`. An id-less client keeps every current path.
- **Slot layout**: the id sits at parked-op index 3. PR #994 (user todos) parks a four-slot op only when a todo rides; whichever of #994 and this lands second pads to the five-slot layout (`user_todo or None` at 3, `qid` at 4) in a one-hunk rebase.
- **Follow-up**: PR #1030 (editable queued message) adds `editQueued` arms beside the `cancelQueued` arms and will conflict textually whichever lands second; an edit by id is the natural continuation of this change.
- **Docs**: one paragraph in `docs/read-side.md` states where the identity comes from and that a copy without an id (the tmux route's queue, which the CLI holds) is still cancelled by index and body.

Three residuals remain. (1) A kernel that refused the client's id (a collision or a forgery, since the ids are 32 random hex digits) mints its own: the client's bubble then stands for a copy the kernel identifies otherwise, its ✕ names an id neither queue holds and gets the miss (never a relocation by body), the client drops its entry, and the copy shows under the kernel's id with its own ✕, which cancels it exactly. (2) An older kernel ignores `qid` on the wire and cancels by index and body as today; on the client, a ✕ on such a kernel's copy drops the entry that copy covered on the last push. (3) On the tmux route the parked send carries the id (its chip and its ✕ are exact) and the CLI's queue does not; that queue's copies are cancelled by index and body as today.

## Tests

Executing; each is red on main's sources unless marked as a guard.

- `tests/test_queued_copy_press_id.py` (new, 22 tests). Red on main: two same-text parks carry their ids in press order and a park without one stays three-slot; the slot survives the disk mirror; the handover passes the id to a backend whose send takes it and the text alone to one whose send does not; the drain hands each parked send and a parked command its own id (TypeError on `qid=`, or the command fired without its id); `_cancel_parked` by id removes exactly the op wearing it, misses on an id nothing holds, and never targets the in-flight head (TypeError; the index would have removed the other op); `_cancel_backend_queued` by id pops the named entry, misses on an unknown id without popping, and takes the index/body path on a backend whose `unqueue` takes no id (TypeError on `qid=`); through `_drive`: a send parks under the posted id, a follow-up parks its wrapped body under it, a malformed or already-held id is not taken, and a cancel naming the id removes that copy in either queue whatever index the click carried (in the idx arm the index names the first copy and the id the second, and the id wins), with the honest miss and untouched neighbours in both queues otherwise; against the real `TmuxBackend`, a parked send keeps its id for its chip and its exact cancel until the drain hands the pane the text. Guards (green on main; red on an earlier cut of this change that keyed the gate on `pending_queued_meta`): a composer send, a follow-up and a parked slash command carrying an id reach the pane as the text and the queue behind the command is not dropped, on the real `TmuxBackend` and on a stand-in with its shape; the id-less cancels keep the index/body fallback in both queues.
- `tests/test_queued_copy_identity.py`: the parked-copy pin is inverted (a parked copy carries the id it was pressed with, shows it on the chip, and enters the backend's queue and echo under it; a kernel-parked copy still carries none until the backend mints one); new class `TheSdkQueueTakesTheClientsId`: `unqueue(qid=)` pops the named copy and its echo, leaving the same-text other and its echo, and misses on an id the queue no longer holds (before: `unqueue()` had no `qid`); with the named copy's echo already gone, the same-text neighbour's echo stays (before: the text match popped it); `_client_qid` takes a fresh id in the echo form and refuses one held by the queue or a live fed echo, and every other shape. Guard: `send(qid=)` keys the copy and its echo by the pressed id.
- `tests/test_queued_copy_held.py` (served page, Chromium): the send posts an id in the echo form and our bubble's ✕ carries the same one (red: nothing posted); the fed gap under a kernel-minted id stays one bubble.
- `ui/webview/send-pending.test.ts`: `newPending` mints an `echo:` id, unique per press, and `dropPending` by that id removes the right one of two same-text entries (red: `qid` undefined); the kernel's copies wearing the press id cover, land and hide by identity from their first appearance; a ✕ on a kernel copy the kernel identified with an id of its own drops the entry it covered by text on the last push, nothing when it covered none of ours, and a stale cover is forgotten once the frame shows the send's own id (red on main, which latched the kernel's id); the T252c identity tests re-expressed on the press-time id; the fed gap under a kernel-minted id (the echo and its held copy hidden as one; a copy under another id left alone).
- Source pins in `tests/test_kernel.py`, `tests/test_kernel_send_park.py` and seven client test files follow the changed lines.

Full suites on this head (`5000d3f0`), each run alone: `pytest -n 8 tests/`: 10457 passed, 40 skipped, 0 failed. `npm run typecheck`: clean. `npm test`: 3789 tests, 3789 pass, 0 fail.

Tier: fix


ksarma:
Post-merge record: the owner merged this PR as a6760364 at e3b8575e before this review ran. The merge stands as shipped. One defect goes to a follow-up on main; nothing needs folding into this PR.

Reviewed inline: the client mints the copy's id at the press, in the kernel's echo form. It posts that id as `qid` on the three composer sends and on the provisional adoption. It never takes an id from a kernel copy afterwards. `_client_qid` admits the id only in that form, and only when the session holds no parked op, queued copy or live echo under it. That check is not atomic across two handler threads. A duplicate would need two clients to mint the same 128-bit id, so I record it as an observation. Every reader of a parked op tolerates the fourth slot: `_parked_md`, `_edit_parked` (which rebuilds from `op[2:]` and keeps the id), `_apply_pending_ops`, the disk mirror's load and save, and the diagnostic. kernel.py has no exact-three unpack. `_takes_qid` reads the send signature. SdkBackend therefore receives the id; TmuxBackend and CodexBackend receive the text alone. A `send(*a, **kw)` stand-in also receives the text alone, which is the conservative answer. `_cancel_parked` by id skips the in-flight head and returns a miss. `SdkSession.unqueue` locates the copy under the session lock. `SdkBackend.unqueue` drops that copy's echo and no other. On a miss the client toasts and repaints from the kernel's events. The queued indicator counts rendered bubbles and is unchanged. The added lines carry no private strings, hostnames or home paths. The PR body predates the rebase. It names seven client test files where the merged diff changes eight; staged-list-cap.test.ts was adapted at the conflict merge. Its suite counts are from the pre-rebase head.

- `ui/webview/send-pending.ts:510`, follow-up: `dropPending` resolves our own bubble by `ts` and text before `qid`. With two same-text entries registered in one millisecond, a cancel drops the wrong client entry while the kernel cancels the right copy. Examples: two identical staged slash commands, two same-text follow-ups to two goals, or two identical held sends adopted together. The survivor then covers the remaining copy by text, and its ✕ posts an id no queue holds. The user sees a false too-late toast, and the copy stays queued until a third ✕ on the kernel's chip. Fix: resolve `qid` first when the ✕ carries one, and fall back to ts and text for an id-less bubble. Add a send-pending.test.ts case with two same-text entries sharing `ts`. `applyQueuedEditLocally` (render.ts:13914) retargets the ✎ edit by ts and text and has the same ambiguity.
- `docs/read-side.md:261`, nit: the paragraph and the PR body's residual 3 say the tmux route's CLI-held queue copies are still cancelled by index and body. TmuxBackend defines no `unqueue`, so `build_session` renders those copies with no ✕. The copies cancelled by index and body are id-less parked ops and id-less SDK queue entries.
- `kernel/kernel.py:28687`, nit: `_CLIENT_QID_RE` ends in `$` under `re.match`, which admits an id with a trailing newline. `fullmatch` enforces the stated bound.

Red-first at the merge base aa902cb4 with the PR's test files: tests/test_queued_copy_press_id.py plus the three source pins 19 failed, 6 passed; the six green are the guards the body names. tests/test_queued_copy_identity.py 4 failed, 20 passed. The changed UI test files 23 failed, with staged-list-cap.test.ts failing to import `mintQid`. At head e3b8575e: press_id and pins 25 passed, identity 24 passed, test_kernel.py 487 passed, the nine UI files 143 passed, typecheck clean. tests/test_queued_copy_held.py is a served browser leg and was reviewed by reading only. CI at e3b8575e is green on all nine checks.

Review run by the maintainer automated review session; verified against head e3b8575e.

PR #1226, opened 2026-09-09
Re-green two node tests after tonight's merges (reply chips via liveSession; staged-list prelude stubs ephemeralWarnToast)

henrypinkard:
Two node tests fail on main after tonight's merges. The skeleton-tabs wiring test counts the sessions.get(activeId) reads in render.ts that may bypass liveSession (name reads and the fork-prompt gates) and expects four; the reply-ready change added a fifth in updateReplyChips, a display path that over a skeleton tab would paint chips from the stale pre-outage session, so that read now goes through liveSession(activeId). The staged-list cap test evals a slice of render.ts whose unreachable-session word became an ephemeralWarnToast call with the auto-reload notice change, which the test's prelude did not define; the prelude now stubs it into the same toast record as warnToast. Typecheck and the full node suite pass (3939 pass, 0 fail, 3 skipped).

Tier: docs

### DIFF
--- a/ui/webview/render.ts
+++ b/ui/webview/render.ts
@@ -11198,7 +11198,7 @@ function dressReplyChip(b: HTMLButtonElement, dir: Dir, chip: ReadyChip | null):
 }
 function updateReplyChips(): void {
   const c = document.getElementById("content");
-  const s = activeId ? sessions.get(activeId) : null;
+  const s = activeId ? liveSession(activeId) : null;   // a skeleton tab's stale session shows no chips (skeleton-tabs-wiring.test.ts)
   const v = activeId ? views.get(activeId) : null;
   const H = c ? c.clientHeight : 0;
   const ready = activeId ? (commentThreads.get(activeId) || []).filter(isReplyReady) : [];
--- a/ui/webview/staged-list-cap.test.ts
+++ b/ui/webview/staged-list-cap.test.ts
@@ -95,6 +95,7 @@ function lift(): (hooks: Hooks) => Api {
     const persistDrafts = () => { H.persists++; };
     const hostIsDown = (id) => H.down.has(id); const isProvisionalId = (id) => H.provisional.has(id);
     const warnToast = (msg) => { H.toasts.push(msg); };
+    const ephemeralWarnToast = (msg) => { H.toasts.push(msg); };   // the slice's unreachable-session word is the ephemeral one
   `;
   const epilogue = `return { routeUserMessage, flushStaged, renderStagedStrip, stagedMsgs, stagedOpen, stagedCollapsed, stagedScroll };`;
   return new Function("HOOKS", prelude + js + epilogue) as (hooks: Hooks) => Api;


ksarma:
Post-merge record: this landed as shipped (a7fd4ac3), and main is correct. Nothing needs folding. Two items for the record: one pre-existing defect on main and one label correction.

Reviewed inline: the `render.ts` line changes behaviour. `updateReplyChips` paints the unread-reply chips over the active transcript. Over a skeleton tab (after a redial), the old `sessions.get` read returned the stale pre-outage session. `showActive` only hides that session's view, and its comment threads are never cleared, so a comments frame or a resize could paint chips from that copy. `liveSession` returns undefined for a skeleton id, and the early return hides the chips. `updateStatusline`, `paintScrollMarks`, `updateCommentRail` and `renderBgTasks` use the same read. The four `sessions.get(activeId)` sites left in `render.ts` (10931, 14942, 15065, 16621) are two name reads and the two fork-prompt gates, all on the wiring test's allowlist. This was the only `sessions.get` line any of tonight's merges added to `render.ts`. The staged-list-cap prelude stub records `ephemeralWarnToast` into the same toast list as `warnToast`. The Send-now test therefore still asserts one toast per refused click, the stack kept and nothing posted. `reload-notices.test.ts` pins the ephemeral form separately. A missing stub still fails loudly as a ReferenceError, which is how this surfaced. The body matches the diff. Its suite count is a local run with the same total as CI. Privacy and vocabulary scans of the added lines are clean. Tier, for the record: the PR is labelled docs, but a behaviour change with a test that fails before it is a fix. docs and fix apply the same gate for every author, so no gate changed. Relabelling #1226 to fix would make the record accurate.

- `ui/webview/render.ts:10734`, follow-up, pre-existing on main and narrowed by this PR. `showActive`'s no-session branch hides every view and shows the skeleton loader without re-measuring the chips. When the tab you leave sits at the top of a long transcript with an unread reply below, no scroll clamp and no resize fires. Its reply chip and the go-to-bottom chip then stay over the skeleton's loader until the full lands. A click on the stale chip resolves the old tab's uuid against the new activeId. A scrolled-down leaving tab recovers on its own through the scroll clamp event. Reproduced in headless Chromium and Firefox with the lifted source. Fix shape: call `updateJumpBtn()` before the return at 10734 and extend `skeleton-tabs-wiring.test.ts`'s `showActive` pin.
- `ui/webview/skeleton-tabs-wiring.test.ts:177`, follow-up (nit). Add `updateReplyChips` to the named sweep, so the new read is pinned by name like its siblings and not only by the count. The `render.ts` comment names this test as the pin.
- `ui/webview/staged-list-cap.test.ts:97`, follow-up (nit). The `warnToast` stub is dead now that the slice calls only `ephemeralWarnToast`. Both stubs share one record, so a regression back to plain `warnToast` passes here and is caught only by `reload-notices.test.ts`.

Tests: at the base 4a31744c the unchanged tests fail as the body says. skeleton-tabs-wiring: expected 4, actual 5, the fifth being the `updateReplyChips` read. staged-list-cap Send-now: ReferenceError, `ephemeralWarnToast` is not defined. With only the PR's test file at the base, staged-list-cap passes and the wiring test stays red, so the `render.ts` change is what that test pins. At head 5b4623d2, skeleton-tabs-wiring, staged-list-cap, reply-ready, comment-mark and reload-notices pass 58 of 58, and `tsc --noEmit` is clean. CI at head is green on all 11 checks (vscode-extension job: 3942 tests, 3910 pass, 32 skipped, 0 fail). main's vscode-extension check is failure at 4a31744c and success at a7fd4ac3.

Review run by the maintainer automated review session; verified against head 5b4623d2.

PR #1227, opened 2026-09-09
Auto-reload: the refusals that report a state do not come back on the page that follows the reload

ksarma:
## Symptom

Staging the composer (Cmd/Ctrl+Enter) while a picker is waiting on it, while an edit is in progress, or with attachments on it is refused with a toast, and so is a branch jump to a session that is not on this dashboard. When a kernel restart's reload takes the page within 12 s of one of those refusals, the fresh page shows the refusal again. For the edit refusal that is a false report: the fresh page has no edit in progress, yet a toast says an edit replaces a past message. For the other three the replay is redundant: the fresh page shows the picker, the attachments and the roster for itself, and the toast repeats what is already on screen.

## Cause

#1217 added the notices record: the reload core's synchronous hook keeps the texts of every warning toast on screen (liveNotices), and the page that follows shows them once. A toast that reports a state rather than an event is meant to stay out of the record through ephemeralWarnToast, whose data-ephemeral mark the reading skips. #1217 marked one such refusal, the staged sends' "Can't send yet", and its body counted that as the one state refusal. That undercounted: the three staging refusals in stageComposer and the branch jump's refusal in the branchjump handler also report states, and all four were plain warnToast calls. composerEdits is an in-memory Map that persistDrafts never writes, so the edit refusal in particular replays on a page where the state it describes does not exist.

## Fix

- ui/webview/render.ts: the four refusals are raised through ephemeralWarnToast (stageComposer: "A picker is waiting on this box", "An edit replaces a past message", "Attachments can't be staged"; branchjump: "That session isn't on this dashboard right now"). The comment above ephemeralWarnToast now lists every state refusal and says which states the fresh page shows for itself (the picker, the attachments and the roster come back from the kernel and the persisted drafts) and which it no longer has (a provisional tab; an edit in progress, which lives in memory alone). A comment at each of the two places (stageComposer, the branchjump handler) says why.
- The toasts that report an event stay unmarked, as #1217 left them: the nack (the attachment was not saved, the held message not sent), the dismissal and the other-tab ack, the refusal on a disconnected host, the tag-edit toast. No change to reload-notices.ts, no kernel change, no docs change.

## Tests

- ui/webview/reload-notices.test.ts, 14 cases under node --test (8 before). Six are new and execute render.ts code: warnToast, ephemeralWarnToast, stageComposer and the branchjump handler are lifted out of render.ts and transpiled at run time (the chat-exact-tail-exec.test.ts idiom) and run over a fake DOM whose querySelectorAll evaluates the reading's selector (.warn-toast:not([data-ephemeral]) .warn-toast-msg) as the DOM would, whether the mark was set through dataset or setAttribute.
  - One test per state refusal (a picker waiting on the composer; an edit in progress; attachments on the composer; a branch jump to a session not on this dashboard): the gesture is refused (nothing staged, no switch, the draft untouched), the toast is on screen, and liveNotices skips it. Before the change each fails at the reading, which returns the refusal's text.
  - A toast about what happened (the nack) raised beside two refusals is read, and the refusals are not: the mark is on the state refusals alone. Before the change liveNotices returns all three.
  - Staging with nothing owning the composer stages and raises no toast: the lifted closure is the real one (passes before and after).
  - The source pins now match the four sites and count seven ephemeralWarnToast occurrences (the definition, the two reachability sites, the four state refusals); before the change the count is 3.
- At 252e71af: npm run typecheck clean; npm test 3792 pass, 0 fail.

Tier: fix


ksarma:
Merging once the required check is green. One follow-up below; nothing to fold into this PR.

Reviewed inline: the diff is the four call-site swaps, the census comment above `ephemeralWarnToast`, and two site comments. The swaps are the three staging refusals in `stageComposer` (render.ts:15467 to 15469) and the branch jump's refusal at 16496. Nothing else in `render.ts` changes; `reload-notices.ts` is untouched. I listed every `warnToast` site at the head. The nack, the dismissal, the other-tab ack, the disconnected-host refusal and the tag-edit toasts stay unmarked. No event toast gained the mark. The lifted closures are the real `render.ts` text: each anchor occurs once at head and at main, and the source pins match the same string. The fake DOM evaluates `:not([data-ephemeral])` as a browser does: `setAttribute` writes the dataset key, and the check is presence, not value. The PR leaves the staged strip's Send now path and `renderStagedStrip` untouched. The `fix` label fits. The body and commit message match the diff: 14 cases (8 before), 7 marked sites (3 before).

Follow-up:
- ui/webview/render.ts:15571: the never-started send refusal on a failed provisional tab is still a plain `warnToast` and missing from the census comment. It reports a state in the same `isProvisionalId` family as the marked reachability refusal. After a reload the tab is gone, so a replay names a session the page does not show. Fix: one call swap, a census line, the pin count 7 to 8, one lifted case.

Tests: at the base with the PR's test file, reload-notices cases 8 to 12 fail at the reading. Each returns its refusal's text; 12 returns both refusals beside the nack. Case 13 fails at the first new pin. At the head all 14 pass and tsc is clean. Applied onto current main a7fd4ac3, the PR passes reload-notices, skeleton-tabs-wiring and staged-list-cap, 39 of 39.

CI at the head: every check is green except vscode-extension. Its two failures (3141, 3215; 3948 run, 3914 pass) are main's own at the recorded base 4a31744c, fixed on main by #1226, not this PR's.

Review run by the maintainer automated review session; verified against head 252e71af.

ksarma:
Delta record; nothing here is an action. The review passed at 252e71af, and the owner merged the rebased head a9cbf827 (aa902cb4). This covers only the delta.

Reviewed inline: the delta is three commits. 9c1826cf merges main, 8ae51e1c raises the queued-edit staging refusal through `ephemeralWarnToast`, and a9cbf827 adds the queued-edit test pinning both staging refusals in their ephemeral form. The merge resolution is exact. Relative to main, `render.ts` carries only the PR's own hunks: the census comment, the three `stageComposer` swaps with their comment, and the branchjump swap with its comment. The two `mdImgPostPass` calls from #1222 and its pageload row are present once each. `reload-notices.test.ts` is byte-identical to the branch. No line is dropped or duplicated. Marking the sixth refusal is right. `queuedEdits` is an in-memory Map that `persistDrafts` never writes and no load path restores, so that refusal reports a state the fresh page cannot have. The pin count 8 matches the eight occurrences at the head. What aa902cb4 landed on main equals the PR head's change line for line.

Findings:

- `ui/webview/render.ts:15836`, follow-up (a small defect on main). The never-started send refusal on a failed provisional tab, the `isProvisionalId` branch in `sendComposer`, is still a plain `warnToast`. It is unchanged since the base review named it. `provisionalId` and `failedProvisionals` live in memory only. A kernel-restart reload inside the toast's 12 s therefore replays a toast naming a tab and a box the fresh page does not have. The census comment omits this site. Fix: the same one-call swap, a census line, the pin 8 to 9, and a lifted case or source pin.
- PR body, nit. The Tests section still says 14 cases, seven occurrences and four refusals. The merged head has 15, eight and five, and also touches `ui/webview/queued-edit.test.ts`.

Tests: at a9cbf827, `reload-notices.test.ts` and `queued-edit.test.ts` pass 25 of 25, and typecheck is clean. With `render.ts` from 9c1826cf (before the review fix), the head's tests fail 3 of 25: the lifted queued-edit case, the census pin, and the queued-edit adjacency pin. At the merged base 924e56a7 with the head's test files, they fail 8 of 25. CI at a9cbf827: all nine checks green. Privacy scan of the delta's added lines and commit messages: clean.

Review run by the maintainer automated review session; verified against head a9cbf827.

PR #1228, opened 2026-09-09
Awaiting overlay: the states log is read append-incrementally, not re-walked per call, and the interrupt tick's two memos report under GET /perf

ksarma:
## Symptom

A kernel with idle sessions spends every pusher cycle re-walking each one's unchanged states log. `_states_awaiting_overlay` walked every parsed row of `states/<sid>.jsonl` on every call, and its one caller, `_session_awaiting`, runs from the nudge tick, the timeline lanes, `build_session`, the feed, the background-work read and `GET /sessions`: per idle session, per site, per cycle. The shared reader already served the rows from its cache, so no I/O was spent, but the walk ran in full each time. Measured side by side on one file, 2,000 calls each in one process (the walk over `_read_jsonl_incremental` against `fold_records` on the unchanged file; a states log of six state rows per awaiting row):

| rows | walk, per call | `fold_records`, unchanged file, per call |
|---|---|---|
| 500 | 36 us | 4.3 us |
| 5,000 | 337 us | 4.3 us |
| 20,000 | 1,364 us | 4.5 us |

A long-lived session's log is about 5,000 rows. The figures are one run on one machine; the ratio is the point.

<details>
<summary>The measurement, to rerun from the repository root with `python3 bench.py`</summary>

```python
# The awaiting overlay's read on an UNCHANGED states log: the row walk the reader used against fold_records.
# Reproducible: python3 bench.py, run from the repository root; prints microseconds per call. Every directory
# it makes is removed at the end.
import json, os, shutil, sys, tempfile, time
from importlib.machinery import SourceFileLoader
dirs = [tempfile.mkdtemp() for _ in range(3)]
os.environ["XDG_STATE_HOME"], os.environ["CLAUDE_CONFIG_DIR"], td = dirs
em = SourceFileLoader("romp_event_model_bench", os.path.join("bin", "romp-event-model")).load_module()

def walk(path):                      # the reader before the change: every parsed row, every call
    last, working_after = None, False
    for o in em._read_jsonl_incremental(path):
        if not isinstance(o, dict):
            continue
        if "awaiting" in o:
            last, working_after = o, False
        elif o.get("state") == "working":
            working_after = True
    return last, working_after

def step(state, o):                  # the fold's step, the walk's if/elif one row at a time
    if "awaiting" in o:
        return (o, False)
    if o.get("state") == "working":
        return (state[0], True)
    return state

def fold(path, cache, on):
    return em.fold_records(cache, path, lambda: (None, False), step, on=on)

try:
    for rows in (500, 5000, 20000):
        p = os.path.join(td, "states-%d.jsonl" % rows)
        with open(p, "w") as f:
            for i in range(rows):
                f.write(json.dumps({"t": 1781100000 + i, "state": ("working", "idle", "waiting")[i % 3]}
                                   if i % 7 else {"t": 1781100000 + i, "awaiting": i % 2 == 0, "why": "a build"}) + "\n")
        walk(p)
        cache, kinds = {}, []
        fold(p, cache, kinds.append)
        assert walk(p) == fold(p, cache, kinds.append)
        n = 2000
        t0 = time.perf_counter()
        for _ in range(n):
            walk(p)
        walk_us = (time.perf_counter() - t0) / n * 1e6
        t0 = time.perf_counter()
        for _ in range(n):
            fold(p, cache, kinds.append)
        fold_us = (time.perf_counter() - t0) / n * 1e6
        print("%6d rows  walk %8.1f us/call   fold(hit) %6.1f us/call   x%.0f   kinds=%s" % (
            rows, walk_us, fold_us, walk_us / fold_us, sorted(set(kinds))))
finally:
    for d in dirs:
        shutil.rmtree(d, ignore_errors=True)
```

</details>

The interrupt-marks memo the interrupt tick already trims to its alive set (`_intr_marks_memo_stats`) keeps hit, miss and evict counters that nothing reports: `GET /perf` has no key for it, the same gap the review of the goal-store memos closed by adding the `memos` block.

## Cause

The overlay reader predates `fold_records`, which the other readers of this file (`_state_intervals`, `_last_machine_cut`) go through, so it kept its whole-file loop when the rows moved to the shared incremental reader. `_intr_marks_memo_stats` landed with its bump under `_INTR_MARKS_STATS_LOCK` and its trim in the tick, but without a report wired into `_PerfStats.snapshot`.

## Fix

- `_states_awaiting_overlay` reads the log through `_fold_records` with a step (`_states_overlay_step`) that is the walk's if/elif applied one row at a time, so an unchanged file steps nothing, an append steps only the new rows, and a rewrite, a shrink or a same-size new mtime steps every row again from record 0 (the shared reader's identity gate; every states writer appends). The stale-true rule is unchanged. One row is read differently from the walk: a complete final record still waiting for its newline, which the incremental reader leaves unconsumed and the walk therefore skipped until the newline landed, is stepped provisionally onto a copy of the carried state (`_fold_eof_fragment`, as `_state_intervals` and `_last_machine_cut` already read it) and for good once the newline lands. Every states writer puts the row and its newline in one write, so that is a write caught between the two.
- The interrupt tick's tail drops the entries for sessions outside its alive set (`_states_overlay_forget`) next to `_intr_marks_forget`: a session leaving the alive set is the event that retires both. The cache is bounded the way the other `fold_records` caches are (cleared whole above 256 entries; that drop is not counted under `evict` and shows as `entries` falling); a dormant session read by `GET /sessions` re-enters and leaves again on the next tick, one pass over cached records with no I/O.
- `fold_records(cache, path, init, step, on=None)`: `on` is called once per call with the path taken, `hit`, `append`, `refold` or `fail`. `_read_jsonl_incremental(path, on_fail=None)`: `on_fail` receives the exception for a stat, open or read that raised on a file that exists (any `OSError` but `FileNotFoundError`); an absent file stays a state and answers `[]` quietly. On `fail` the cache entry is dropped, the answer is `init()` and nothing is memoized, so the next call reads again; when a lost pin forces the re-read, the re-read's verdict is the one that counts. `fold_records` keeps no counters of its own.
- `_states_overlay_on` counts each call under `_STATES_OVERLAY_LOCK` (the counters are bumped from the pusher, connect-time builds on socket threads and `GET /sessions` at once) and on `fail` writes one stderr line per episode naming the file; a later good read ends the episode.
- `GET /perf` `memos` gains `intrMarks` (`hit`, `miss`, `evict`, `entries`; `_intr_marks_memo_report`) and `statesOverlay` (`hit`, `append`, `refold`, `fail`, `evict`, `entries`; `_states_overlay_report`), copied under their locks. The `_PerfStats` docstring and `docs/reference.md` describe both.
- The three test stubs over `_read_jsonl_incremental` in `tests/test_bg_scan_fold.py` take the new keyword.

Not in this PR: no CLI or `romp perf` row (the CLI prints the goals line only, as for the other memos). Apart from the newline-less final row above, what the overlay answers is unchanged.

## Tests

Red first, on the unchanged sources: 27 failed, 1 passed, no errors (the one pass is the no-`on` regression guard, `test_without_on_the_fold_is_unchanged`). Every test below is an executing test over the real code paths; the two `chmod 0` tests skip for a user who reads through mode 000.

`tests/test_states_overlay_fold.py` (new; a private synthetic sid per the goal-store fixtures rule, a fresh state root per test through `jd._rebind_state`, the memos cleared before the root is rebound so a failed setUp leaks nothing):
- `OverlayFold`, equality against a private copy of the walk it replaced on every newline-terminated shape: awaiting then idle; a later work turn superseding a stale true; a work turn before the overlay row; a row carrying both keys (awaiting first); state rows with extra keys; the eight marker rows the other writers append; a trailing false row; no rows and no file (no `fail` counted); unparseable lines. Before the change: `AttributeError`, `_states_overlay_cache` does not exist.
- `test_a_final_row_without_its_newline_is_read_provisionally`: the one row the two read differently, in both shapes (a final overlay row: the walk answers None, the new reader answers the row; a final work turn: the walk answers the stale true, the new reader the superseded verdict); the carried state does not hold the row; the newline lands, the answer does not change, the row is counted as one `append`, and the walk agrees from there on. Red under a mutation that drops the provisional read.
- `test_an_appended_row_folds_without_re_parsing_the_earlier_records`: a counting step sees 3 rows on the first read, 0 on an unchanged read, exactly 1 after one appended row; the report reads (refold, hit, append, entries) = (1, 2, 1, 1).
- `test_a_truncated_file_refolds_from_the_start` (refold = 2, one row stepped) and `test_a_same_size_rewrite_with_a_new_mtime_refolds` (same byte length, mtime moved 2 s).
- `test_a_read_failure_on_a_file_that_exists_is_counted_logged_once_and_not_memoized`: the file grows, then `chmod 0`: two calls answer None, fail = 2, entries = 0, exactly one `unreadable` stderr line naming the file; readable again, every row is stepped again from record 0 and the episode clears.
- `test_forget_releases_entries_for_sessions_outside_the_alive_set`: two sids read, `_states_overlay_forget({SID})` leaves (entries, evict) = (1, 1); a second call drops nothing and counts nothing.
- `test_the_report_has_the_documented_shape`: the key set, integer values, and that the report is a copy (writing to it leaves the counters alone; red under a mutation that returns the live dict).
- `InterruptTickRetires.test_a_sid_leaving_the_alive_set_loses_its_entry`: a names entry, a transcript and a rebound state root, so `_interrupt_block_tick(NOW, tmux)` resolves its alive set through discover; the departed sid's entry goes, the live one stays, evict counts 1; an empty alive set (not headless) retires the last entry.
- `IntrMarksReport`: the report's key set, the copy semantics, and `entries`/`evict` after a forget. Before the change: `AttributeError`, `_intr_marks_memo_report` does not exist.

`tests/test_event_model_incremental.py::FoldRecordsReports`: `refold` on a first read, `hit` unchanged, `append` after one appended record, `refold` after a rewrite; an absent path answers `init()` with `on_fail` never called; an unreadable file that exists (`chmod 0` on the file: the open raises) reports `fail`, answers `init()`, pops the cache entry and steps every record again once readable; a stat that raises on a file that exists (`chmod 0` on its directory: the reader's first `except` branch) reports `fail` the same way and hands `on_fail` a `PermissionError`; a failed first read followed by a clean re-read on a lost pin (a stub that fails once and then delegates to the real reader, whose entry an earlier read left) answers the re-read's records, reports `refold` and not `fail`, and memoizes; no `on` leaves `fold_records` unchanged. Before the change: `TypeError` on the `on=` and `on_fail=` keywords. Each of the two failure-branch tests is red under a mutation that removes the line it covers (the stat branch's `on_fail` call; the reset of the failure list before the re-read).

`tests/test_perf_stats.py`: the exact `memos` key set gains `intrMarks` and `statesOverlay`, each equal to its report, with the documented key sets and integer values; the reference doc names both. Before the change: the set assertion and the doc assertion fail.

Neighbouring modules run green on the change: `test_interrupt_marks_memo`, `test_awaiting_count`, `test_awaiting_rows`, `test_kernel_awaiting_stamp`, `test_kernel_timeline_awaiting`, `test_stage0_log_readers`, `test_scanner_folds`, `test_kernel_jsonl_cache`, `test_bg_scan_fold`, `test_judge_parse_cache`, `test_injected_voice`, and `test_kernel.py -k "awaiting or overlay or states or interrupt"`.

Full suite (`pytest -q -n 8 tests/` on the final tree): 10433 passed, 64 skipped, 0 failed, 1687 subtests passed.

Tier: fix


ksarma:
Merging once the vscode-extension check reruns green. Nothing in the PR needs to change. That check is red for main's own reason: it fails exactly the two node tests main was failing, the liveSession pin and the Send-now test. Both were fixed on main at a7fd4ac3, after this PR's run. The PR touches nothing under `vscode-extension` and merges clean onto that main.

Reviewed inline: the fold step is the old if/elif, one row at a time, so on newline-terminated rows the answer is unchanged. Every states writer appends row and newline in one write, and nothing truncates the file. The writers are nine `sdk_backend` helpers, the kernel's picker and idle rows, `idle_dots`, and three tmux-status printfs. So the provisional read of a newline-less final row touches only a write caught mid-syscall. A torn fragment fails `json.loads` and steps nothing. The step lands on a deepcopy. The memoized verdict equals the answer once the newline lands, so a card can read one write earlier but cannot flap. Counters take `_STATES_OVERLAY_LOCK`. The cache is a bare dict of immutable tuples, like the sibling fold caches, so concurrent folds end last-writer-wins with a consistent entry. A failed read pops the entry and answers `init()` unmemoized. An absent file stays a state. The lost-pin re-read's verdict counts. The 256 whole-clear is the shared idiom, stated in body, docstring and docs. Past it, a refold steps cached records with no I/O, no worse than the walk it replaced. The `fix` label fits: the `/perf` keys are counters inside the existing memos block, as in #1200.

Two wording nits, follow-up only:
- Three tests carry the chmod-0 skip guard, not two: `tests/test_event_model_incremental.py:202`, `:231`; `tests/test_states_overlay_fold.py:232`.
- The `_states_awaiting_overlay` docstring's writer list omits `cli/idle_dots.py`, which also appends in one write.

Tests: the PR's four test files over current main's sources gave 28 failed, 86 passed. The body's 27 are AttributeError on `_states_overlay_cache` and `_intr_marks_memo_report`, TypeError on `on` and `on_fail`, and the memos-set and doc assertions. The 28th is one untouched timing flake. At head, 114 passed, with the chmod-0 tests running. Neighboring modules: 214 passed, 190 subtests. `test_kernel.py -k 'awaiting or overlay or states or interrupt'`: 26 passed. CI: Python 3.10 to 3.13, bats and both tier checks are green. vscode-extension is red with main's two tests only.

Review run by the maintainer automated review session; reviewed at 335a583f and verified against head 469938c8 (a merge of main into the branch after main was re-greened; the PR diff against the new merge-base is byte-identical to the reviewed one).

ksarma:
The delta would not have changed the verdict; one follow-up goes to main. The review passed at 335a583f and the owner merged the rebased head 142ddc5c. The delta is the merge of main, the review-fix commit 21de3fc1 and the conflict resolution with #1230.

Reviewed inline: the conflict resolution is the union of both sides in all three regions. The `_PerfStats` report list, the key set in `tests/test_perf_stats.py` and the docs memos paragraph each name liftGate, bgTops, intrMarks and statesOverlay once. The docs carry #1230's compaction-sweep sentence and this PR's interrupt-tick sentence once each. The PR's own blocks are byte-identical between 21de3fc1 and the head, so no line of either PR was dropped. `GET /perf`'s snapshot, the key-set test and `docs/reference.md` agree on the 14 memo keys. The review fix discards a departed sid's path from `_states_overlay_failed` under `_STATES_OVERLAY_LOCK` in a block of its own. The later evict bump therefore does not re-enter the lock. A fail always pops the fold-cache entry, so the original loop could not reach such a path. No stderr line marks the end of an episode by either route, which matches the once-per-episode contract. Privacy scan of the 801 added lines: clean.

- `kernel/kernel.py:24523`, follow-up. The discard resets the once-per-episode stderr latch for a departed sid whose states file is still read. `build_session`'s `_session_awaiting` call at 31219 has no liveness gate. A dead session kept open as a read-only tab, and `GET /sessions/<sid>`, therefore re-reads the file on every rebuild. An unreadable file then logs one line per rebuild instead of one per episode. Timeline dead lanes never read the overlay, so the rate is that tab's rebuild cadence, not the pusher cycle. Log noise only; the overlay answer is unchanged. Suggested fix: drop the per-path discard and clear `_states_overlay_failed` whole above a cap, the fold cache's idiom.

Tests: red-first at the base 924e56a7 with the head's test files, 28 failed and 87 passed. The 28 are the 19 overlay tests, the new one included, the 7 FoldRecordsReports tests and the 2 perf-stats assertions. At the head, the six named modules report 224 passed. With the discard block removed, exactly the new test fails. CI at 142ddc5c: all nine checks green, with the Tier policy check reporting the fix gate.

Review run by the maintainer automated review session; verified against head 142ddc5c.

PR #1229, opened 2026-09-09
The feed's foreign cleared-id list skips ids that name no session: parked, quarantine and placeholder rows

ksarma:
## Symptom

On a merged board, the local feed payload's list of clears taken for other kernels' cards (`clearedForeign`, added by #1190 so a viewer's clear applies over a remote row whose kernel's own projection cannot see it) carries every parked-handoff, quarantine and placeholder row the clear log holds, this kernel's own included, on every frame. One footer Clear-all on a board with many placeholders appends a fresh set of such rows, and since the list is newest first under a cap of 500, those rows crowd genuine clears of remote cards out of it. #1190's own text promises the list never carries a local id; this is that promise's residual.

## Cause

`_cleared_foreign` classifies an id by the text before its last colon against the stems of `goals/` and `goals-archive/`. Five id families the feed builds have a word there, never a session id: `parked:<msgId>` and `quarantine:<mid>` key a message; `provisional:<sid>`, `awaiting:<sid>` and `blocked:<sid>` are the placeholders `build_feed` re-lists every build whatever the clear log holds. So each such row read as foreign whether or not the card was this kernel's. The `":" in i` filter from #1195 is a proxy these ids pass. The rows are written live: the footer's Clear-all fans out to every attached kernel, and each kernel's `clearAll` handler clears every ask its feed lists, placeholders and quarantine holds included (cards the client's own card-level and session Clear leave alone), so each click appends another set and the append-only log keeps them.

## Fix

A tuple `_CLEARED_NO_SESSION` beside the clear-log memo names the five prefixes, and the comprehension skips ids that start with one. None of the five names a session whose clear rows a kernel reads: a parked or quarantine id keys a message, so this kernel cannot tell its own from a remote one (after delivery or a decision the local record is gone, the row is not); no kernel honors a clear row for a placeholder, so no overlay could apply one. Only `sid:gN` ids ride. The docstring's first sentence now states the rule the function applies: the cleared ids that name a session with no goal store and no archive here. The payload key, the sort and the cap are unchanged, as are the write path and the client helper (`applyViewerClears` is a no-op for ids it never receives). An explicit list rather than a shape test on the stem: the `goals/` stems are the ground truth for a session id and any uuid text is a valid one, so a new family that keys no session is added to the tuple.

Declined here, named so it is a known gap: a viewer's clear of a REMOTE parked card whose route to its kernel failed no longer overlays (the state before #1190 for that card class; the kernel that lists the card honors the clear whenever it arrives, and Clear-all reaches every attached kernel directly). The exact alternative, the client sending the row's host with the clear, is a separate change. The write side (a `clearAll` handler that clears placeholders the client would refuse) is also a separate change and would not clean an existing log, so the read-side skip stays needed either way. The same last-colon reading in `_mark_nodes_cleared`, `_subtree_item_ids` and `_delegation_linked_ids` loads a store for a sid named `parked` and finds nothing (`load_goals` returns a fresh store without writing, and the save is guarded), harmless and left alone.

## Tests

- `tests/test_kernel_goal_compaction.py::ClearedLedgerIsAuthoritativeAcrossTheCompaction::test_the_foreign_id_list_skips_cleared_ids_that_name_no_session` (new): appends one clear row per family (`parked:`, `quarantine:`, `provisional:`, `awaiting:` and `blocked:` for the local session, plus `blocked:` for a remote session) and a remote session's `sid:g1` and a local node's id, resets the clear-log memo, and asserts `_cleared_foreign(_cleared_ids())` is the remote node clear alone; then the same over a direct dict; then a direct dict of six hundred `blocked:` rows all newer than one remote clear, asserting the remote clear alone rides. Before the change the first assertion gets the remote clear followed by the six prefixed ids newest first, the second gets the remote clear followed by `blocked:<sid>`, `quarantine:m2` and `parked:m1`, and the third gets the five hundred newest placeholder rows with the remote clear cut off by the cap. A variant that applies the skip after the cap instead of before it fails the third assertion (the list comes back empty), so the skip's position is pinned too.
- The module's existing foreign-id test (local id off, undone id off, cap of 500 newest first, bare node id off, the payload key's source pin) is unchanged and green; the module passes whole (15 tests).
- Full suite: `pytest -q -n 8 tests/` on the committed tree: 10523 passed, 85 skipped, 1688 subtests passed, 0 failed (72 s).

Tier: fix


ksarma:
Merging once the extension check re-runs green against current main. One wording nit, no blockers.

Reviewed inline: `_cleared_foreign` gains `_CLEARED_NO_SESSION` and skips ids with those five prefixes inside the comprehension, before the newest-first sort and the 500 cap. The payload key, sort, cap, ledger writers and `applyViewerClears` are untouched. The five prefixes match every non-node `itemId` build_feed emits (provisional, awaiting, blocked, parked, quarantine). Each clear-log writer (Clear-all, askClear and askClearMany, mute, the episode boundary, undo) writes only node ids or those prefixes, so the tuple is complete. A scratch mutation that applies the skip after the cap fails only the third assertion, with an empty list, so the position is pinned. `applyViewerClears` checks `foreign.has(id)` per remote row and returns on an empty set, so it ignores ids it never receives. The #1190 bare-id comparison defect is closed by 90815a93, an ancestor of both main and this head. The declined case reads as the body says: a clear of a remote parked card, taken while its owner was unreachable, no longer hides that card on the viewer's board. The current client writes no such row locally, so the case needs a copied ledger or an older client. Body and commit match the diff; fixtures are synthetic.

Follow-up (wording), kernel/kernel.py:31253 and the body: the sentence that no overlay could apply a placeholder clear names the wrong mechanism. The kernel half holds. The client overlay drops any remote row whose id is in the set, with no shape test. So a `blocked:<remote-sid>` row in the local ledger did hide that session's placeholder. The skip removes that latch, one more reason for it.

Tests: at origin/main a7fd4ac3 with the PR's test file, the new test fails at its first assertion (the remote clear followed by six prefixed ids): 1 failed, 14 passed. At 9d89c1a8 the module passes, 15. CI at head: Python 3.10 to 3.13, bats and both tier checks green. vscode-extension is red with exactly main's two node tests. 5b4623d2 fixed those on main; this PR's merge ref predates it. main is green at a7fd4ac3.

Review run by the maintainer automated review session; reviewed at 9d89c1a8 and verified against head 35dedd7e (a merge of main into the branch after main was re-greened; the PR diff against the new merge-base is byte-identical to the reviewed one).

PR #1230, opened 2026-09-09
The awaiting lift decides on the shared read-only goal store and loads the writer's copy only when a lift is due

ksarma:
Two commits, in dependency order: the placed-launch memo first (`_bg_placed_tops`), then the two-phase awaiting lift that hands it the store it decides on.

## Symptom

On a board with several live sessions carrying awaiting stamps, the kernel reads each of those sessions' goal store from disk and replays its override journal on every pusher cycle whose inputs moved, and nearly every one of those reads files nothing: the dispatch the stamp waits on is still out. Beside it, the placed-launch resolver behind the lift and the feed's background-task chips re-loads the same store privately whenever the lift and the feed ask about different launch ids, which they do every cycle. Nothing on the board looks wrong; the pusher thread spends its time deciding nothing. Both sites are listed as `UNWIRED` by `tests/test_kernel_goal_cache_wiring.py`, left on the writer's loader when the shared read-only store cache landed.

## Cause

`_lift_spent_awaiting` had one inline body that loaded through `jd.load_goals_or_fault` (the writer's loader) and filed verdicts in four arms of the same loop. The only store it could decide on was a private mutable copy, so every session whose fingerprint moved paid the full load before the rules could say "nothing due".

`_bg_placed_tops` kept one cache slot per session, keyed on the transcript, store and journal stats plus the exact tuple of launch ids asked. The lift asks with every task id the transcript records and the feed with the live ids (often none), so each ask evicted the other's fill, and every miss called `jd.load_goals`. The stat key was also a weak version: a rewrite that keeps the mtime and the byte count did not move it (the old answer was served until the file next moved), and since it was taken before the load, a publish landing between the two paired the old identity with the new placements for one cycle.

## Fix

Commit 1, `_bg_placed_tops(sid, path, tids, store=None)`. It reads through `jd.load_goals_shared`, or the store the caller hands in, and keeps per session one map of every launch id asked so far under one (parse object, store object) pair. A hit requires both objects in hand to be the entry's own. `_parse` returns one object per transcript version and `load_goals_shared` one `FrozenStore` per store version, so identity is the version and no stat is taken. Placements are resolved through a per-store index of `jd._seg_key`-normalized keys built with `setdefault`, which answers what `jd._placement_of`'s scan answered for each of the four suffixes, None-valued retired entries included. An absent store file answers nothing after one presence check (no parse, no fallback load). A store that is not the shared cache's `FrozenStore` (a writer's private copy, no file, the cache off) is computed on and never published. A fill copies the map it extends and publishes a new tuple, never writing into an entry, since the pusher and the handler threads both run this; the counters sit behind a lock. An ask with no live ids releases the entry only when its pinned parse is no longer the transcript's current one. Callers are unchanged. `/perf` gains `memos.bgTops` (`hit`, `miss`, `resolve`, `walk`, `walk_neg`, `idx_build`, `entries`). A negative walk cache was priced and not built: phase 2 re-walks the transcript once per due lift for launch ids no segment resolves (positives are cached in `_task_seg_cache`, negatives are not), and `walk_neg` counts the upper bound on what such a cache would save.

Commit 2, the two-phase lift. The `_sid_inputs_fp` / `_lift_seen` gate is unchanged. Behind it, phase 1 probes the shared view through `jd.load_goals_shared_or_fault` and runs the whole decision body, moved into `_lift_decisions`, which consults the verdict gate read-only through `jd.may_apply` and writes nothing, returning `(node key, top, drop_rec, end_ev)` per lift due. The probe sits behind the same per-session boundary as before: a fault forgets the gate and the next cycle retries, and a store whose journal did not read forgets it too. An empty list ends the session's tick, as a save-less tick did before. Phase 2 runs only when a lift is due: the writer's copy is loaded through `jd.load_goals_or_fault`, `_lift_decisions` runs again on that copy, and only its own decisions are filed with `record_verdict` on the node the key names, the spent nudge record dropped where flagged, then `rollup_status` and `save_goals` as before. The frozen view is never written to and never reaches `save_goals`. A writer publishing between the probe and the load (the closer re-placing the deciding launch under another card) costs one writer load that files nothing, counted as `noop`, never a lift from the stale decision. Placement inside a decision is read from the store being decided on (`store=store`), so the stamps ruled on and the launches placed come from one version. Phase 1 rests on `jd.may_apply` being the gate `record_verdict` consults before it appends, which it is today: `test_the_lifts_decision_body_loads_nothing_and_writes_nothing` pins the call, and the four floor cases below run `record_verdict` on the writer's copy beside phase 1's decision. A later change there that bypassed `may_apply` would make phase 1 decide a lift that phase 2 refuses, a `noop` and one wasted load, never a wrong lift.

The rule body moved verbatim. A whitespace-normalized diff of the removed inline body against `_lift_decisions` shows only the extraction's substitutions: the two candidate lists in `_lift_candidates` and one early return when the store has no candidate (so the candidate scan runs once per probed session), `record_verdict(...)` to `jd.may_apply(...)` plus an append in the four arms, the three `changed`/save blocks to `return out`, the node key where the id field was read (`_top_of(nid)`, `stamped.remove((nid, nd))`), `store=store` on the placement ask, the nudge-record drop moved to phase 2 with its comment, and one comment losing the word "above". The `or now` agents horizon, the `_return_ev` shape and the `end_ev` journaling are as they were.

Also in commit 2: after the loop, `_lift_seen` entries for sids that left the alive set are dropped (nothing pruned them before; bounded by the session count, so a leak rather than a bug), and so are the placed-launch memo's entries for those sids. `/perf` gains `memos.liftGate` (`skip`, `load`, `shared`, `writer`, `noop`, `entries`); `docs/reference.md` describes both memos.

The gain was measured once and no test asserts a timing. On a copy of a live state directory with 31 sessions, three of them stamped with their dispatches still out, the lift tick's steady cost per pusher cycle went from 20.8 ms with three writer loads to 6.5 ms with none, timed around the `_lift_spent_awaiting` call in `_pusher_cycle_jobs`. The check anyone can run on their own board is `GET /perf`: `memos.liftGate.writer` stays at zero while no lift is due, and `memos.bgTops.hit` grows while `miss` does not once the lift and the feed have each asked once.

Not included: the inputs fingerprint does not stat the goals archive, so a restore row whose node sits in the archive is re-examined only when another input moves; unchanged here. The 3-argument `_bg_placed_tops` stubs in `tests/test_kernel_bg_tasks.py` and `tests/test_kernel_owned_yield_awaiting.py` are reached through `_bg_pending` and `_bg_split`, never with `store=`, and are left as they are.

## Tests

Every test below runs the real code paths and fails on the tree without its commit, with one exception named where it appears.

Commit 1, `tests/test_kernel_bg_placed_tops.py` (new, `PlacedTops`, 15 cases here and 16 with commit 2), over a synthetic two-segment transcript and a store whose placements are recorded under drifted segment ids. On the tree before the change the module errors at setUp, since the memo names do not exist; the cases then pin:
- the lift's all-ids ask and the feed's live-ids ask share one shared read and one walk, with the writer-loader spy left empty (the old one-slot cache missed on the second ask and called `jd.load_goals`); in the other order, the lift's ask after the feed's resolves only the two ids the map lacked, with one walk for them, and the feed's next ask is a hit;
- a fill publishes a new tuple holding a new map, and the old tuple's map is as its holder read it (red when a fill writes into the map it found);
- an unresolvable launch is absent and its walk counted negative, and the None answer is memoized under the pair;
- the indexed lookup equals the four-suffix scan for plain, live-twin, retired (None-valued), drifted and non-conforming keys, the exact key that is not first in seg-key order among them (red when the exact-key branch is dropped);
- a `save_goals` publish, a journal append and a new parse object each move the version, with one index per store object;
- a publish between a read and the ask is served for the object in hand and then fresh; a same-identity rewrite (same inode, byte count and mtime) is served fresh;
- a writer's copy is computed on and never published (a TypeError before: no `store` parameter); an absent store file costs no load of either kind and no parse;
- an empty live set keeps the fill while the parse is current and releases it once the parse is stale; `/perf` reports the memo.

`tests/test_kernel_goal_cache_wiring.py`: `_bg_placed_tops` moves to the wired set (0 shared loads before), and a pin that no stat is taken in it. `tests/test_perf_stats.py`: the `bgTops` key set (missing before). `tests/test_kernel.py`: a comment on the cache clears.

Commit 2, `tests/test_kernel_awaiting_lift.py`, new class `LiftGate` (26 cases), counting `jd.load_goals` (writer) and `jd.load_goals_shared` (probe) apart, with the shared cache's poison canary checked in every tearDown. On the tree with commit 1 alone, 25 of the 26 fail; the exception is the dormant-session case, which pins behaviour the change keeps (a dormant session is skipped before the gate and nothing is remembered for it):
- an unchanged unstamped store is probed once then gated; a stamped store with its dispatch still out costs one probe and no writer load on the tick that reads it, then is gated until an input moves, and lifts on the return (before: one writer load and zero probes, `(1, 0)` against `(0, 1)`);
- a due lift takes one probe and one writer load and files one lift row, dropping the spent nudge record once;
- the rolled-up, peer-superseded and empty-registry arms each take one writer load; a node keyed apart from its id field still lifts (found by key on the writer's copy);
- a lift the verdict gate refuses, in each of the four arms (a user follow-up after the stamp, `followupAt` past the stamp's anchor, is the evidence floor `record_verdict` applies to a lift), is decided on the shared view and costs no writer load: one probe, the stamp stands, no row, `writer` and `noop` unchanged, the session gated on the next tick, and `record_verdict` on the writer's copy refuses the same lift (red when any arm decides without `jd.may_apply`: a wasted writer load and a `noop` per tick);
- a writer racing the probe costs one reload and never a wrong skip; a writer racing between the probe and the load costs one reload counted `noop` and never a wrong lift;
- a save publish, a journal restore row, a block row, an override row, a same-size in-place rewrite and a same-size rename with the old mtime each move the key;
- a missing store is gated until it appears, its first read counted under `load` and not under `shared` (red when `shared` counts every probe); a faulting store read and a swallowed journal read are not cached as nothing to lift; a failed probe records no skip; a dormant session is neither gated nor recorded; a sid that left the alive set loses its gate entry (kept before); `/perf` reports the gate.

`tests/test_kernel_goal_cache_wiring.py`: the two-phase pin (one `jd.load_goals_shared_or_fault(`, one `jd.load_goals_or_fault(`, no bare load in `_lift_spent_awaiting`; 0 and 1 before), and the pin that `_lift_decisions` loads nothing, writes nothing and asks `jd.may_apply` exactly four times (one per arm; the name did not exist before) and that `record_verdict` asks `may_apply`. `tests/test_kernel_bg_placed_tops.py`: the end-of-tick prune evicts a sid that left the alive set (kept before). `tests/test_perf_stats.py`: the `liftGate` key set. `tests/test_stage0_log_readers.py`: the counted loader becomes `jd.load_goals_shared`, which the tick's store read now is (counting `jd.load_goals` there reads 0 after the change). The five 3-argument `_bg_placed_tops` stub lambdas on the lift's path accept `store=None`; `LiftGateFallback` keeps counting `jd.load_goals` unchanged, since the unreadable-journal case still reaches it through the shared loader's hand-off.

Targeted modules on the final tree (`pytest -q -n 6`): `tests/test_kernel_awaiting_lift.py tests/test_kernel_bg_placed_tops.py tests/test_kernel_goal_cache_wiring.py tests/test_perf_stats.py tests/test_stage0_log_readers.py tests/test_kernel_bg_tasks.py tests/test_kernel_owned_yield_awaiting.py tests/test_wake_answer_files_under_shared_view.py tests/test_goal_store_fault_boundary.py tests/test_kernel_bg_tasks_stale.py tests/test_wake_deadman_toggle.py tests/test_paused_matrix.py tests/test_kernel_tick_jobs_under_frame.py tests/test_kernel.py`: 833 passed, 1 failed. The failure, `tests/test_kernel.py::ServeSecurity::test_restart_error_echo_is_bounded_and_keeps_its_explanation`, reads the restart door's body parser and nothing this change touches; it passed alone on the same tree, and it passed in the full run below.

Full suite on the final tree (`pytest -q -n 8 tests/`): 10570 passed, 85 skipped, 0 failed, 1688 subtests passed.

Tier: fix

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1021,12 +1021,20 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   `plannerSkip` counts the sessions the outer gate let through, not every
   planner skip: an idle session stops at the outer gate and appears in neither
   `skipped` nor `planned`. Outside a pass frame (`romp-judge --plan`) the
-  outer gate stamps nothing, and the inner gate does the skipping. The
-  compaction sweep after each judge pass evicts from `pass` and `shared` the
-  entries of stores no session in the discover window owns, so both stay
-  bounded by the live board; the courier's and the planner's change-gate
-  tables are pruned to the sessions each pass discovers, and the evidence
-  gate's stamps are cleared at a fixed cap.
+  outer gate stamps nothing, and the inner gate does the skipping. `bgTops`
+  is the placed-launch memo behind the awaiting lift and the feed's
+  background-task classification, keyed on the parse object and the store
+  object: `hit` and `miss` (calls answered from the per-version map against
+  looked up), `resolve` (launch ids looked up on a miss, placed or not),
+  `walk` and `walk_neg` (transcript walks, and the walks that left a launch
+  unresolved: an upper bound on what a negative walk cache would save),
+  `idx_build` (placement indexes built, one per store object asked, a
+  writer's private copy included) and the gauge `entries` (sessions holding
+  a map). The compaction sweep after each judge pass evicts from `pass` and
+  `shared` the entries of stores no session in the discover window owns, so
+  both stay bounded by the live board; the courier's and the planner's
+  change-gate tables are pruned to the sessions each pass discovers, and the
+  evidence gate's stamps are cleared at a fixed cap.
 - `judge`: `passes`, `ms_sum`, `ms_last`, `ms_mean` (wall time; a pass waits
   on model calls), `cpu_ms_sum` (CPU time of the judge tier threads and every
   per-session worker they run; the workers' share is `cpu_ms_workers`).
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1021,20 +1021,28 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   `plannerSkip` counts the sessions the outer gate let through, not every
   planner skip: an idle session stops at the outer gate and appears in neither
   `skipped` nor `planned`. Outside a pass frame (`romp-judge --plan`) the
-  outer gate stamps nothing, and the inner gate does the skipping. `bgTops`
-  is the placed-launch memo behind the awaiting lift and the feed's
-  background-task classification, keyed on the parse object and the store
-  object: `hit` and `miss` (calls answered from the per-version map against
-  looked up), `resolve` (launch ids looked up on a miss, placed or not),
-  `walk` and `walk_neg` (transcript walks, and the walks that left a launch
-  unresolved: an upper bound on what a negative walk cache would save),
-  `idx_build` (placement indexes built, one per store object asked, a
-  writer's private copy included) and the gauge `entries` (sessions holding
-  a map). The compaction sweep after each judge pass evicts from `pass` and
-  `shared` the entries of stores no session in the discover window owns, so
-  both stay bounded by the live board; the courier's and the planner's
-  change-gate tables are pruned to the sessions each pass discovers, and the
-  evidence gate's stamps are cleared at a fixed cap.
+  outer gate stamps nothing, and the inner gate does the skipping. `liftGate`
+  is the awaiting lift's per-session inputs gate and two-phase read: `skip`
+  and `load` (session-cycles that took no store read against the ones that
+  read it, a probe on the shared read-only view), `shared` (probes the shared
+  cache answered), `writer` (session-ticks that loaded the writer's copy
+  because a lift was due) and `noop` (writer loads whose fresh decision filed
+  nothing, the store having moved between the probe and the load), and the
+  gauge `entries` (sessions remembered). `bgTops` is the placed-launch memo
+  behind the awaiting lift and the feed's background-task classification,
+  keyed on the parse object and the store object: `hit` and `miss` (calls
+  answered from the per-version map against looked up), `resolve` (launch ids
+  looked up on a miss, placed or not), `walk` and `walk_neg` (transcript
+  walks, and the walks that left a launch unresolved: an upper bound on what a
+  negative walk cache would save), `idx_build` (placement indexes built, one
+  per store object asked, a writer's private copy included) and the gauge
+  `entries` (sessions holding a map). The compaction sweep after each judge
+  pass evicts from `pass` and `shared` the entries of stores no session in the
+  discover window owns, so both stay bounded by the live board; the courier's
+  and the planner's change-gate tables are pruned to the sessions each pass
+  discovers, the evidence gate's stamps are cleared at a fixed cap, and the
+  awaiting lift's tick drops the gate's and the placed-launch memo's entries
+  of sessions that left the alive set.
 - `judge`: `passes`, `ms_sum`, `ms_last`, `ms_mean` (wall time; a pass waits
   on model calls), `cpu_ms_sum` (CPU time of the judge tier threads and every
   per-session worker they run; the workers' share is `cpu_ms_workers`).
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -399,7 +399,8 @@ class _PerfStats:
         for key, read in (("pass", _goals_memo_report), ("shared", jd.shared_store_stats),
                           ("chain", jd.chain_memo_stats), ("courierSkip", jd.courier_skip_stats),
                           ("backref", jd.backref_memo_stats), ("captions", jd.captions_memo_stats),
-                          ("goalArchive", jd.goal_archive_memo_stats), ("plannerSkip", jd.planner_skip_stats)):
+                          ("goalArchive", jd.goal_archive_memo_stats), ("plannerSkip", jd.planner_skip_stats),
+                          ("bgTops", _bg_tops_report)):
             try:
                 memos[key] = read()
             except Exception:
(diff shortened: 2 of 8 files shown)


ksarma:
Merging when the four Python checks finish green. Nothing to fold.

Reviewed inline: the write path first, since #1141's `_wake_goal` wrote the frozen shared view. Phase 1 reads through `load_goals_shared_or_fault` and runs only `_lift_decisions` on it. Phase 2 loads the writer's copy and re-runs the decision body on that copy. `record_verdict`, `rollup_status` and `save_goals` all take that copy (`kernel.py` 9813 to 9859), so the frozen view is never written or saved. A scan of `_lift_decisions`, `_bg_placed_tops` and `_placement_index` finds no assignment into the probed store. The index builder fills a local dict. The LiftGate tests run each arm on a real FrozenStore, with the poison canary checked in tearDown. `record_verdict` opens with the same `may_apply` call the four arms consult. A writer publishing between the probe and the load therefore costs one no-op load and cannot file a stale lift. A whitespace-normalized diff of the removed inline body against `_lift_decisions` shows only the substitutions the PR body lists. `_parse` and `load_goals_shared` each return one object per version, so the identity key cannot serve another version's placements. The tier fits `fix` at its large end. Each defect has a test that is red without it, the rules moved verbatim, and the `/perf` blocks follow #1200's shape. The head moved during review. 4527f099 is b562bab1 merged with main, and git's own merge of the two parents produces the identical tree.

Follow-ups, neither blocking:
- `kernel/kernel.py:9827`: the liftGate writer counter increments before the fault check at 9829, so a faulted writer load is counted as a writer load.
- `docs/reference.md:1034`: the phrase describing `bgTops miss` is clipped. The docstring's wording, "calls that looked up at least one tid", is complete.

Tests: red-first is met by my own runs. `tests/test_kernel_bg_placed_tops.py` at base 4a31744c with the head's file: 16 failed, 16 errors at setUp on the missing `_PLACEMENT_IDX`. `tests/test_kernel_awaiting_lift.py::LiftGate` at commit 1 alone (8063f4a5): 25 failed, 1 passed; the pass is the dormant-session case the PR body names. At 4527f099 the PR body's 14 targeted modules: 834 passed. CI at 4527f099: Shell, both tier checks and vscode-extension are green, and main's two node failures are fixed in the merged main. The four Python jobs are in progress.

Review run by the maintainer automated review session; verified against head 4527f099.

PR #1232, opened 2026-09-09
Feed: the hover-freeze tells the hovered session header apart by column, not by session alone

ksarma:
## Symptom

In grouped mode, rest the pointer on a session's header row in one column while that session also has cards in another column, and let a push arrive while the row is held. The deferred-churn hints paint wrong: the session's header in the other column shows no +N/-N badge at all, and the floating hint (the +N count with its mouse-away note) is drawn under that other header instead of under the row the pointer is on. Only sessions with cards in two or more columns are affected; single-column sessions behave as #1188 intended.

## Cause

`reconcileCol` keys a grouped session header per (column, sid): one session heads a run in every column it has cards in, so such a session has two header elements, and `updateSessHead` stamps both with the same `data-fsid`. The header's hover-freeze key (`sessFreezeKey`, from #1188) named the session only (`"h:<sid>"`), and `paintFreezeBadges` compared each header's sid to the hovered sid. Every header of the hovered session therefore took the hovered branch: its in-row badge was removed (`put(h, undefined)`) and the single `#freeze-headnote` was re-positioned under it, so the last twin in document order won the float and the twin lost its badge.

## Fix

Three edits in `ui/webview/feed.ts`, on the shapes #1188 introduced:

- `updateSessHead` stamps the column beside the sid, compare-first like the sid: `if (h.getAttribute("data-fcol") !== e.col) h.setAttribute("data-fcol", e.col);` (`e.col` is already in the header Entry).
- `sessFreezeKey` names the (column, session) row: `"h:" + data-fcol + ":" + data-fsid`, still read from the stamps at event time. The stamps survive the re-key an exiting header gets (`startSessHeadExit` sets `data-key` to `x:N`), so a ghost's mouseleave computes the key its mouseenter stored and releases the gate; a key derived from `data-key` would not, and the second test below holds that.
- `paintFreezeBadges` keeps the whole key (`hoveredHead = freezeKey` when it starts with `"h:"`) and compares `sessFreezeKey(h) !== hoveredHead`, so exactly one live header is the hovered row; every other header, the same session's twins included, keeps its in-row badge. The float's creation, placement and removal are unchanged.

Other readers of an `"h:"` key (`cardElByKey` via `selfKey`, and the keyboard-scope Tab entry path) query `[data-key="a:<key>"]` and return null for a header key before and after. No CSS or kernel change.

## Tests

- `ui/webview/feed-render-incremental.test.ts`, new: "a hovered session header is one (column, session) row: the same session's header in another column keeps its in-row badge and the floating hint sits under the hovered row". It boots `feed.ts` under the file's DOM stand-in, renders one session with a card in Working and one in Blocked (two headers with the same `data-fsid`), dispatches `mouseenter` on the Working header, dispatches a feed frame adding a Completed card of that session, and asserts: the payload is queued; the Blocked header carries a `.freeze-badge` child; the hovered header carries none; `#freeze-headnote` exists and its `style.right` equals the hovered row's right offset (the stand-in gives the two rows distinct rects). It then releases on `mouseleave` and asserts the queued frame applied and the note came off. Before the change it fails on the Blocked header's missing badge (expected true, actual false); with that assertion skipped it fails on the note's position (`710px`, the Blocked row, where `770px`, the Working row, is expected).
- `ui/webview/feed-render-incremental.test.ts`, new: "a hovered header's gate releases through its ghost: a Clear that takes the row out from under the pointer re-keys it, and the ghost's mouseleave still applies the queued frame". With the pointer on a session's header and a frame queued behind it, it clicks Clear on that session's one card: `dressHeaderIfLast` starts the header's exit in the same click, with no render between, so the element under the pointer is re-keyed to `x:N` and gains `.sess-exit` while its `data-fcol` and `data-fsid` stamps are unchanged. It then dispatches `mouseleave` on the ghost and asserts the queued frame applied. On the unchanged code it fails on the column stamp (no `data-fcol` is written there); with this change's `sessFreezeKey` derived from `data-key` instead of the stamps, it fails on the release (the queued frame stays queued).
- Both new tests release the gate in a `finally`, so a failure never leaks a held gate into the next test of this shared-world file; the second restores its world there too, so a failure leaves no cleared card behind either.
- `ui/webview/feed-render-incremental.test.ts`, extended: "a header's data-fsid and data-fcol are written only when they differ" now checks that each header's `data-fcol` is the column of the list it sits in and that three renders of one frame re-set neither stamp. Fails before the change on the missing column stamp.
- `ui/webview/feed-freeze.test.ts`: the pins on `sessFreezeKey`'s body, the hovered compare and the `hoveredSid` line are re-aimed at the new shapes, and a pin for the compare-first `data-fcol` write sits beside the `data-fsid` one.
- Targeted run after the change: `node --test` over the two files, 51 tests, 51 pass, twice more for determinism. `npm run typecheck` clean.
- Full `npm test` on this head (2d5bfeb8): 3913 tests, 3911 pass. The two failures are the two tests that were red on main from 02d96006 until #1226 (the liveSession read count in `skeleton-tabs-wiring.test.ts` and the Send-now test's lifted prelude in `staged-list-cap.test.ts`); neither file is touched here, and the two files this change touches pass in full.

Tier: fix

### DIFF
--- a/ui/webview/feed-freeze.test.ts
+++ b/ui/webview/feed-freeze.test.ts
@@ -166,8 +166,10 @@ test("badges wear the header conventions: accent adds, block-red removes, the co
   // painted on the build-once column heads and the data-fsid-stamped session headers; cleared when quiet
   assert.match(FEED, /put\(document\.querySelector\("\.feed-col\.col-" \+ key \+ " \.feed-col-head"\), d\.cols\[key\]\);/);
   assert.match(FEED, /h\.setAttribute\("data-fsid", e\.sid\);/);
-  assert.match(FEED, /const c = groupedNow \? d\.sess\[sid\] : undefined;\s*if \(sid !== hoveredSid\) \{ put\(h, c\); return; \}/,
-    "every header but the hovered one carries its in-row badge (the hovered one's floats — T285)");
+  assert.match(FEED, /if \(h\.getAttribute\("data-fcol"\) !== e\.col\) h\.setAttribute\("data-fcol", e\.col\);/,
+    "the column stamp beside the sid, compare-first like it: the two together are the row the hover key names");
+  assert.match(FEED, /const c = groupedNow \? d\.sess\[sid\] : undefined;\s*if \(sessFreezeKey\(h\) !== hoveredHead\) \{ put\(h, c\); return; \}/,
+    "every header but the hovered ROW carries its in-row badge (the hovered one's floats, T285); the same session's header in another column is not the hovered row");
   assert.match(FEED, /document\.querySelectorAll\("\.freeze-badge"\)\.forEach\(\(n\) => n\.remove\(\)\);/,
     "nothing pending → every badge comes off");
   // local renders while frozen re-sync the hints, so a rebuilt board never strands a stale count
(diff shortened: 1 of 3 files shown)


ksarma:

Reviewed inline: `reconcileCol` keys a grouped header per (column, sid). A session with cards in two columns therefore has two headers with the same `data-fsid`. #1188's hover-freeze named the hovered header by sid alone, so the painter took the hovered branch on both. The fix stamps `data-fcol` beside `data-fsid` in `updateSessHead`, compare-first, so repeated renders of one frame re-set neither. `sessFreezeKey` reads both stamps at event time and names the (column, session) row. `paintFreezeBadges` compares each header's whole key to the held key. Exactly one live header is then the hovered row, and the twin keeps its in-row badge. A header's column stamp is constant for the element's life, because `reconcileCol` hands each list only its own column's entries. `startSessHeadExit` rewrites only `data-key`, so a ghost's mouseleave computes the key its mouseenter stored and releases. The existing release paths (mouseleave, the stale-freeze heal on a local render, window blur) are unchanged; no new held-gate path is added. The other readers of an `h:` key (`cardElByKey` through `selfKey`, the Tab entry) query `a:` keys and return null for a header key before and after. The label fits: a defect in #1188's mechanism, no new mechanism, no CSS or kernel change. Added lines are synthetic; body and commit match the diff.

Follow-up, not blocking: `ui/webview/feed-render-incremental.test.ts:906` dispatches the queued frame before the try opens. A throw inside that dispatch would skip the finally and leave the gate held for the file's later tests. The second new test dispatches inside the try.

Test evidence:
- The PR's two test files at base 35362afc: 5 of 51 fail, for the reasons the body gives (the missing `data-fcol` stamp in the extended stamp test and the ghost test, the twin's missing in-row badge, the two feed-freeze pins).
- At head, 51 of 51 pass and tsc is clean.
- CI at 2d5bfeb8: all 11 checks green; the vscode-extension run started after #1226 merged.
- The merge onto current main 09d40b82 is clean, and main's moves since the base touch none of the PR's files.

Review run by the maintainer automated review session; verified against head 2d5bfeb8.

ksarma:
Head note: the PR merged at 24304650, a merge of main into the reviewed head 2d5bfeb8 made by the owner's branch update after auto-merge was armed; verified after the fact: the parents' own merge-tree (2d5bfeb8 with 924e56a7) equals the merged head's tree, so the merge carried no hand edits and the review above stands for what landed.

Review run by the maintainer automated review session; verified against head 24304650.

PR #1233, opened 2026-09-10
Override replay: a same-second re-clear or re-undo replays over a snapshot that already holds the earlier gesture

ksarma:
## Symptom

A card cleared, undone and cleared again within one second can come back uncleared in its goal store: a triage pass that loaded the store after the undo saves over the re-clear, the store then reads the card uncleared while cleared.jsonl hides it, and the flag and the verdict stay wrong until the next compaction re-seals the root. The mirror shape (clear, then undo, clear and undo within one second, with the pass's snapshot taken after the re-clear) keeps the flag on a card cleared.jsonl has released: the feed hides it, the compaction archives it flagged, the archive projection keeps the flag, and no Undo can reach it.

## Cause

#1187 journals the clear and #1190 settles same-second clear and unclear rows by journal order (`last_seal`). Both arms still asked whether ANY same-kind user write at the row's second survived (`_twin`), which reads an earlier same-second sibling's write as the row's own. Over a snapshot taken after the undo (log: clear T, undo-reopen T; cleared false), the re-clear's row found the first clear at T and skipped, so nothing re-sealed the card. Over a snapshot taken after the re-clear (log: clear T0, undo-reopen T, clear T; cleared true), the last undo's row found the first undo at T and skipped. #1190's same-second test takes its snapshot before the first clear, where the first row's replay supplies the clear and the third row skips on the flag, so the after-undo half had no red test there.

The loss reaches the store through `save_goals` as well as through a raw overwrite: `_rebase_onto_disk` keys verdict identity on (ev_t, src, kind), so a pass that loaded the store after the undo and publishes after the re-clear rebases the re-clear away as a duplicate of the first clear (the third test below).

## Fix

`kernel/judge.py`, `_replay_overrides`, the clear and unclear arms:

- Of a node's clear and unclear rows in one second only the last acts. The pre-scan that fills `last_seal` also fills `last_seal_at`, (node, second) to the last such row's op at that second; a row whose op is not that entry was superseded within its second and skips, whatever the log holds.
- The acting row asks `_newest_seal()` in place of `_twin`: the node's newest user clear or undo-reopen at that second, in `_fold_node`'s order. When the second's last word is in the log already (survived, or re-recorded by an earlier row of the same second in this load) the row skips; when not, it re-records it, and the next load finds the write and skips. One row added per lost gesture, none after that.

`nd.get("cleared")`, `last_seal`, the strictly-later guards, `record_verdict`, and the resolve, followup, move and block arms are unchanged.

The last-row rule is what keeps the newest-seal check from regressing two shapes that stand today. An undo and a re-clear in one second followed by an undo a second later, over a snapshot taken after the re-clear: read on its own, the superseded undo row finds the second's newest seal is a clear and re-records a reopen the second did not end on; the later undo's reopen then follows a reopen instead of a clear, finds nothing to restore, and a completed top comes back Working. A clear and an undo in one second followed by a re-clear a second later, over a snapshot taken after the undo: the superseded clear row re-records a clear stamped with the first second, and the re-clear's own row skips on the flag. A count of surviving same-kind writes against the row's ordinal was considered and rejected: two undo rows of one second around a re-clear that never replays would record two reopens (the seventh test below fails under it).

Not changed here: a clear whose undo follows it in the journal never replays over a snapshot that predates both (`last_seal` reads the node's last word as the undo), so the undo's reopen finds no clear to restore from and a completed top comes back Working, with the flag correctly off. That is the `last_seal` skip's own shape, the same before and after this change.

## Tests

`tests/test_kernel_goal_compaction.py`, class `ClearedLedgerIsAuthoritativeAcrossTheCompaction`, under a sid of its own. `time.time` is pinned so the gestures share a second where the shape needs it: the journal `t` and the verdict `ev_t` and `at` all read it, so the premise does not depend on the wall clock (#1190's test assumes its three gestures land inside one second).

Red before this change:

- `test_a_same_second_re_clear_replays_over_a_snapshot_taken_after_the_undo`: clear, undo, snapshot, clear; overwrite, load. Fails with `AssertionError: False is not true : the last row, a clear, wins over a snapshot that holds the first clear and its undo`. Also asserts the log gains exactly one clear row and a further save and load add none.
- `test_a_same_second_re_undo_replays_over_a_snapshot_taken_after_the_re_clear`: clear at an earlier second; then undo, clear, snapshot, undo; overwrite, load. Fails with `AssertionError: True is not false : the last row, an undo, wins over a snapshot that holds the re-clear`. Also asserts exactly one reopen row added, none on a further load, and the top's status back to completed.
- `test_a_stale_pass_save_rebases_the_re_clear_away_and_the_next_load_re_seals`: clear, undo, a pass loads its copy, clear; the pass publishes through `save_goals` (premise asserted: the published log holds the first clear and its undo and the card reads uncleared); load. Fails with `AssertionError: False is not true : the next load re-seals the card`.
- `test_an_undo_a_second_after_a_same_second_undo_and_re_clear_over_a_snapshot_taken_before_them`: clear at an earlier second, snapshot; undo and clear in one second; undo a second later; overwrite, load. Fails with `Lists differ: ['clear', 'reopen', 'reopen'] != ['clear', 'reopen']`, the top Working.

Green before this change, red under the newest-seal check without the last-row rule (checked by dropping the `last_seal_at` guards):

- `test_an_undo_a_second_after_a_same_second_undo_and_re_clear_replays_once`: the same gestures with the snapshot taken after the re-clear; asserts one reopen added, at the last undo's second, and the top back to completed.
- `test_a_re_clear_a_second_after_a_same_second_clear_and_undo_carries_its_own_second`: clear and undo in one second, snapshot, clear a second later; asserts the re-recorded clear carries the re-clear's second.

Green before and after:

- `test_two_same_second_undos_around_a_skipped_re_clear_record_one_reopen`: clear at an earlier second, snapshot, then undo, clear, undo in one second; asserts one reopen and the top completed. Under the rejected count it fails with `['clear', 'reopen', 'reopen'] != ['clear', 'reopen']`.

Replacing `_newest_seal()` with `_twin` while keeping the last-row rule turns the first three red again. The helpers `_completed_top` and `_clobber_with` take a `sid` keyword with the old default; the existing tests in the class are untouched.

Targeted run: the seventeen test modules that exercise the replay, the clear path or the undo path, 778 passed. Full suite: `pytest -q -n 8 tests/`, 10572 passed, 115 skipped, 1688 subtests passed.

Tier: fix

### DIFF
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -4267,7 +4267,12 @@ def _replay_overrides(fsid, store, lines=None):
     a card reply in the same second are two gestures, and the old >= guard silently dropped the
     reply's replay, bouncing the card back mid-pass). Judge events do not cancel a replay: the user
     event is appended anyway and the fold's authority rules arbitrate. `block` keeps at-or-after: a
-    user reply in the same second as a nudge stamp genuinely answers it.
+    user reply in the same second as a nudge stamp genuinely answers it. The two seal ops (clear,
+    unclear) do not read a same-kind twin as their own write: a clear, an undo and a clear can share
+    one second, and the first clear's survival says nothing about the last one. Of a node's clear and
+    unclear rows in one second only the LAST acts (`last_seal_at`; an earlier one was superseded within
+    that second), and it asks whether the node's NEWEST seal at that second already is its op
+    (`_newest_seal` below).
 
     `lines`: the journal's lines when the caller has already read them (load_goals_shared reads the
     journal from a descriptor it also takes the file's identity from, so the replayed rows and the key
@@ -4291,13 +4296,15 @@ def _replay_overrides(fsid, store, lines=None):
     applied = False                                    # any write → load_goals re-runs rollup (one truth)
     arch_nodes = None                                  # the archive is read once, only if a restore entry needs it
     last_seal = {}                                     # node -> the LAST clear/unclear row's op, in journal order: a
-    for ln in lines:                                   # clear, undo and clear inside one second are settled by the rows'
-        try:                                           # order, not by their integer seconds (review find, 2026-09-09)
-            ev0 = json.loads(ln)
+    last_seal_at = {}                                  # clear, undo and clear inside one second are settled by the rows'
+    for ln in lines:                                   # order, not by their integer seconds (review find, 2026-09-09).
+        try:                                           # last_seal_at: (node, t) -> the last such row's op AT THAT SECOND;
+            ev0 = json.loads(ln)                       # an earlier row of the second stands down (see the arms)
         except ValueError:
             continue
         if ev0.get("op") in ("clear", "unclear") and ev0.get("node"):
             last_seal[ev0["node"]] = ev0["op"]
+            last_seal_at[(ev0["node"], int(ev0.get("t") or 0))] = ev0["op"]
     for ln in lines:
         try:
             ev = json.loads(ln)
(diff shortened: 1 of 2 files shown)


ksarma:

Reviewed inline: the clear and unclear arms of `_replay_overrides`. The pre-scan fills `last_seal_at` per (node, second) in journal order, so only a row carrying its second's last op reaches the guard. There `_newest_seal` replaces `_twin`: it reads the node's user clear and undo-reopen rows at that second and stable-sorts them by `at`. That is `_fold_node`'s `(ev_t, at)` order restricted to one second, so equal timestamps keep journal position. I walked the two-row and three-row same-second shapes with the snapshot before, between and after the rows. In every case the final flag matches the last journal row, and the log gains at most one row per load. Two independent enumerations through the real clear and undo paths (130 and 66 cases) found no disagreement at the head, against 13 at the base. `_rebase_onto_disk` keeps its `(ev_t, src, kind)` identity and needs no change. The re-recorded row shares the collapsed row's identity, so a stale rebase can collapse it again and the next load re-records it. The third test pins that. Widening the identity would change dedup for every verdict kind. The tests run under a private sid, with state rebound to a per-test temp dir removed in tearDown. The time pin is a `with` block. The fix label fits. The body and commit message match the diff and quote no one. The privacy scan of the added lines is clean.

Follow-up, pre-existing (identical at the base):
- `kernel/judge.py:4405`: a completed top cleared, undone, re-cleared and undone comes back Working with the flag correctly off. That happens when a pass whose copy was taken after the first clear publishes through `save_goals` after the last undo. The rebase collapses the re-clear into the writer's clear and keeps both reopens. The fold then reads the second undo-reopen, which has nothing to restore, as open. The next load has nothing to re-record. The body's residual paragraph names the raw-snapshot form; the rebase form reaches the same state. Candidates: the fold treating an undo-reopen with nothing to restore as a state no-op, or the rebase deduping disk rows among themselves.

Test evidence: at the merge-base with the PR's test file, the four tests the body names fail with the stated messages, and 17 pass. The four are the re-clear over a snapshot after the undo, the re-undo over a snapshot after the re-clear, the stale pass save rebasing the re-clear away, and the undo a second after a same-second undo and re-clear over a snapshot before them. At the head 21 pass. On the merged tree with today's main, 22 pass. With the `last_seal_at` guards dropped, the two guard tests fail. With `_twin` in place of `_newest_seal`, the first three fail. With the rejected count, the seventh fails, as the body states. 28 replay, clear and undo modules: 1496 passed. CI at the head is green on every required check, and the Tier policy check reports the fix gate.

Review run by the maintainer automated review session; verified against head 75bfae17.

ksarma:
Head note: the head moved from 75bfae17 to 095241c2 by a merge of main into the branch (this session updated the branch to re-run the checks against current main after a long hold); the merge is mechanical (the parents' merge-tree equals the head's tree) and the PR's added and removed lines are identical to the reviewed ones, so the review above stands for this head.

Review run by the maintainer automated review session; verified against head 095241c2.

PR #1236, opened 2026-09-10
Comments: a second comment in the same words on the same passage makes a second thread, not a second ack on the first

ksarma:
## Symptom

Post a comment on a passage and take the ack. Post the same words on the same passage again, under another name. No second thread appears: the popover adopts the first thread again, and the second name is written nowhere. The viewer shows no toast; only the kernel log says `comment create repeated`. The post-merge record on #1208 reports this as a follow-up on main.

## Cause

The create memo from #1208 (T289) keys a create on its words, `(parent sid, anchor uuid, passage, text)`, and answers any identical create on an open thread until 256 later creates or a kernel restart. The words cannot tell a re-post of one create (a lost ack; a lag-parked copy that landed before the client's re-post reached the kernel) from a fresh comment that happens to repeat them, so the memo swallows the deliberate second comment.

## Fix

The popover stamps each send gesture with a `createId`, and every re-post of that gesture carries the same one: a repeat arrives with an id the kernel has seen, a fresh gesture with a new one.

- `ui/webview/comments.ts`: `CommentCreate`, `mintCreateId` (the moment and a random tail, the provisional tab id's shape; random rather than a counter because a reloaded viewer restarts its counters while the kernel's memo stands), `newCommentCreate` (the held create, stamped, picks defaulted to `""`), `commentCreateFrame` (the frame the send and every re-post build).
- `ui/webview/render.ts`: the send mints the held create, posts its frame and holds it; the frame-keyed retry re-posts the held frame; `cmtCreateInFlight` holds `CommentCreate` rows.
- `kernel/kernel.py`: `_create_key` takes the id and keys on `(parent sid, createId)` when the frame carries one, else on the words as before; `_parked_key` reads a parked copy's key; the WS door, the reservation, the park-once check and the pusher's retry all key on it, and a parked copy carries the id. A frame without an id (a client that sends none) still keys on its words, and the two shapes never meet. The repeat rule is otherwise unchanged: the memo answers only an open thread, and the busy nack and the park-once check stand.
- The federation route passes the field through untouched, as it does every field, so a comment on a remote session reaches the owning kernel stamped.

One case moves the other way: a comment the user posts again by hand while the first is still parked (after a viewer reload, which loses the in-memory hold, or after the viewer gave up at `CMT_CREATE_MAX_TRIES` while the kernel's park at `_PARK_MAX_TRIES` is still alive) is a new gesture and lands as a second thread once the anchor arrives. The words key answered that re-post busy and collapsed both into one thread. The duplicate is visible and the user can delete it; the collapsed second comment was lost silently. The parked-copies test below covers the kernel side of this case.

The alternative considered was a memo bounded to the create's in-flight and parked lifecycle, cleared once the ack has gone out. It was not taken: the memo exists for the re-post that arrives after the ack (the client re-posts on the frame the pusher sent before it retried the parked copy, and a lost ack is re-sent on the next frame), so an ack-bounded memo would mint twins in the cases the memo was built for, and no lifecycle bound tells a late re-post from a fresh gesture. The gesture's own id does, and the client already keys its hold per gesture (one create per anchor at a time), so the stamp extends how it identifies the create today.

## Tests

`tests/test_comment_create_idempotent.py`, seven added (14 in the module). Four are red before the change:

- `test_a_fresh_gesture_in_the_same_words_on_the_same_passage_is_a_second_thread`: two gestures, two ids, same words, two names; before, `1 != 2` (one thread, two acks, the second name nowhere on disk).
- `test_a_fresh_gesture_after_a_parked_create_landed_is_a_second_thread_and_its_repost_is_not`: the parked copy carries the id through the pusher's retry; the re-post is the same thread and a fresh gesture is a second one; before, `1 != 2` on the fresh gesture.
- `test_two_gestures_in_the_same_words_are_both_parked_while_the_anchor_lags`: two gestures park separately and land as two threads, the same gesture parks once; before, `1 != 2` on the second parked copy.
- `test_a_frame_without_a_create_id_still_keys_on_the_words`: the words identity stands for an unstamped frame and never meets a stamped one; before, `1 != 2` (the stamped gesture was answered from the words memo).

Three pin the guards around the id (green before and after; each goes red under the named change to the kernel):

- `test_a_retried_frame_with_the_same_create_id_is_one_thread_answered_twice`: the re-post side, one thread and two acks naming it.
- `test_a_repost_of_a_parked_gesture_is_refused_at_the_door_before_the_create_runs`: a spy on `_comment_create` sees no second call for a re-post of a parked stamped gesture; red if the reservation compares parked copies by their words (the re-post would run the create again and be saved only by the park-once check).
- `test_a_repost_landing_between_the_doors_release_and_its_park_is_parked_once`: the door releases its reservation before it parks, and a re-post of the same gesture that runs whole in that window parks first; this door then parks no twin. Red if the park-once check compares by the words.

`ui/webview/comments.test.ts`, six added (75 in the bundle): `newCommentCreate` and `commentCreateFrame` executed (two gestures in the same words get two ids; the frame carries the id and the picks, and a retry is the same frame; empty picks and an empty name stay empty); `retryCmtCreates` lifted from render.ts and run against a stubbed `vscodeApi` (every re-post carries the id minted at the send and is the held frame whole, red if the retry re-mints the id; a create with no transient nack yet is not re-posted; one past the attempt bound is dropped with the toast); the send site pinned. The pins on the old inline frames in three tests are retargeted. `ui/webview/comment-remote-routing.test.ts`: the id rides the route untouched. Before the change the test bundle fails to build on the two missing exports.

Targeted on this tree: the module 14 passed; the six comment bundles 121/121; `tsc --noEmit` clean.

Full suite on this tree (this commit on a7fd4ac3): pytest `-n 8` over `tests/`, 10647 passed, 40 skipped, 0 failed, 1688 subtests passed. `npm test` 3948 tests, 3948 pass, 0 fail; `npm run typecheck` clean.

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -13700,16 +13700,23 @@ def _comment_markers(sid):
 # alone starves; the client's frame-keyed re-post stays as the belt for a kernel restart that loses
 # this in-memory park). The typed transient nack keeps the client's optimistic mark alive meanwhile.
 ANCHOR_LAG_ERR = "that message isn't in the transcript yet; try again in a moment"
-_parked_creates = []                       # [{sid,uuid,exact,text,name,model,effort,fast,color,tries}]
+_parked_creates = []                       # [{sid,uuid,exact,text,name,model,effort,fast,color,createId,tries}]
 _PARK_MAX_TRIES = 30                       # pusher cycles (~15-90s) — past this the record isn't coming
-# A create's IDENTITY (T289): (parent sid, anchor uuid, passage, text) -> the thread it made. A lag-parked
-# create is retried by BOTH the pusher (above) and the client (its frame-keyed re-post), and a popover
-# whose ack was lost sends its create again; the second copy used to collide on its explicit name and
-# come back as the create door's refusal toast (the user 2026-09-09, on a remote session), or — with the
-# name now left to the kernel's default — would mint a SECOND thread for one comment. The memo answers a
-# repeat with the SAME thread's ack, and a parked copy is parked once. Bounded (oldest out), in memory:
-# a kernel restart forgets it, and the client's re-post after one creates exactly once.
-_recent_creates = {}                       # (sid, uuid, exact, text) -> tid
+# A create's IDENTITY (T289): the client's createId, minted by the popover at the send gesture and carried
+# by every re-post of it, under the parent sid -> the thread it made. A lag-parked create is retried by
+# BOTH the pusher (above) and the client (its frame-keyed re-post), and a popover whose ack was lost sends
+# its create again; the second copy used to collide on its explicit name and come back as the create
+# door's refusal toast (the user 2026-09-09, on a remote session), or, with the name now left to the
+# kernel's default, would mint a SECOND thread for one comment. The memo answers a repeat with the SAME
+# thread's ack, and a parked copy is parked once. The first cut keyed on the words (parent sid, anchor
+# uuid, passage, text), and so also answered a DELIBERATE second comment in the same words on the same
+# passage with the first thread: two acks, the second name nowhere on disk (review, 2026-09-09). The id
+# keys on the gesture instead; a frame from a client that sends none still keys on its words. A comment
+# the user posts again by hand while the first is still parked (after a viewer reload, or after the viewer
+# gave up at its own attempt bound) is a new gesture and lands as a second thread: a visible duplicate
+# the user can delete, where the words key lost the second comment silently. Bounded (oldest out), in
+# memory: a kernel restart forgets it, and the client's re-post after one creates exactly once.
+_recent_creates = {}                       # (sid, createId) or (sid, uuid, exact, text) -> tid
 _RECENT_CREATES_MAX = 256
 # The two doors that can run one create — the WS handler on a server thread and _retry_parked_creates on
 # the pusher — reserve the identity under ONE lock before creating (review, 2026-09-09): the pusher sends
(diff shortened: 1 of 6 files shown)


ksarma:
Merging as shipped. One follow-up below, nothing to fold in.

Reviewed inline: `_create_key` returns (parent sid, createId) for a stamped frame and the four-word tuple for an unstamped one. The two shapes never compare equal in the shared memo. The WS door, `_reserve_create`, the park-once check and `_retry_parked_creates` all key through `_create_key` or `_parked_key`. The door notes the id-keyed entry before it releases, and the pusher notes under the lock before it drops the park. A lost ack re-sent on the next frame is therefore answered with the first thread. The release-before-park window parks one copy. The id is minted client-side: Date.now plus a Math.random tail, base 36, about 50 bits. It is never echoed in an ack or the pusher's broadcast, so another viewer cannot learn one. A client that reuses an id collapses only its own comments under that parent sid. Federation passes the field through. `routeOutbound` spreads the frame and rewrites only id and sid. The extension host forwards the message whole. The kernel reads createId with no allowlist. The 256-entry bound and in-memory memo are unchanged, so a re-post after a restart behaves as before. The reverse case, a hand re-post while the first copy is still parked, lands as a visible second thread. The body states it and the parked-copies test covers it.

Follow-up:
- `tests/test_comment_create_idempotent.py:91`: setUp clears `_thread_msgs_cache` but not `_recent_creates`, `_inflight_creates` or `_parked_creates`, and the tests reuse the same ids under one sid. Isolation holds today only because `_repeat_create_tid` re-reads the store and thread ids are random. Clearing all three in setUp removes the ordering dependence.

Tests: at main 82dd9e9c with the PR's test file, the four named tests fail with 1 != 2 and the three guards pass. The four: a fresh gesture in the same words; a fresh gesture after a parked create landed; two gestures both parked; a frame without an id. At head the module is 14/14, `comments.test.js` plus `comment-remote-routing.test.js` 78/78, tsc clean. Each guard goes red under its named kernel mutation, and the UI retry test under a re-minting retry. CI at 00bc2e5f: all 11 checks green.

Review run by the maintainer automated review session; verified against head 00bc2e5f.

ksarma:
Head note: the head moved from 00bc2e5f to d585f2c0 by a merge of main into the branch (this session updated the branch to re-run the checks against current main after a long hold); the merge is mechanical (the parents' merge-tree equals the head's tree) and the PR's added and removed lines are identical to the reviewed ones, so the review above stands for this head.

Review run by the maintainer automated review session; verified against head d585f2c0.

PR #1235, opened 2026-09-10
Tests: a served lab's cfg.json carries only the environment the relaunched kernel needs, never the runner's whole environment

ksarma:
## Symptom

Running tests/test_ship_reship.py or tests/test_dashboard_reload_served.py on a machine whose shells carry API keys leaves those keys in a temp file for the run. Both labs kill their hermetic kernel from the browser driver and start it again, and the driver reads the kernel's environment from the lab's cfg.json; that file held the runner's whole environment.

## Cause

Each lab builds its kernel's environment as a copy of os.environ with the lab's roots and seams over it, which is right for the kernel the lab starts from the test process (a child gets its environment by process). The same mapping was written whole into cfg.json's relaunch.env for the driver's child_process.spawn.

## Fix

tests/test_ship_reship.py gains relaunch_env(env), the part of a lab kernel's environment the relaunched kernel needs: the ROMP_* and XDG_* names, CLAUDE_CONFIG_DIR, PATH (bin/romp-kernel runs under `env python3`) and HOME, plus what tests/conftest.py sets for every child of the run: TMPDIR and TMUX_TMPDIR, the private roots its temp files and tmux server live under, and GIT_CONFIG_GLOBAL and GIT_CONFIG_NOSYSTEM, which keep the kernel's boot-time git (the build sha, the release-tag probe) off the developer's git configuration. relaunch_cfg(env, klog) builds the cfg.relaunch stanza (the command, relaunch_env of the environment, the log), and both labs write their stanza through it: the ship lab builds its kernel's environment in kernel_env, and the dashboard lab imports the module. The lab's own kernel still gets the full environment by process, and what the relaunched kernel does with its environment is unchanged; only what goes to the file is narrowed.

## Tests

[...]

PR #1237, opened 2026-09-10
Tests: the live session-move test borrows the operator's apiKeyHelper under pytest again

ksarma:
## Symptom

The live session-move test (`tests/test_session_move_live.py`, opt-in under `ROMP_MOVE_LIVE=1`) authenticates its hermetic config dir with the `apiKeyHelper` command borrowed from the operator's own Claude Code settings when neither `ANTHROPIC_API_KEY` nor `ROMP_MOVE_LIVE_API_KEY_HELPER` is set (#946). Under pytest that route cannot work: on a machine whose settings name a helper, `ROMP_MOVE_LIVE=1 pytest tests/test_session_move_live.py -k LiveSetCwd -rs` skips with "no Claude Code auth for a hermetic config dir", while the skip message still lists "an apiKeyHelper in your Claude Code settings" as a way in. The class is opt-in, so nothing went red.

## Cause

#1065 floors `CLAUDE_CONFIG_DIR` to an empty temp directory for every test (so no test resolves the real `~/.claude`) and re-asserts it per test, and conftest scrubs `ANTHROPIC_API_KEY`. `_user_api_key_helper` read `CLAUDE_CONFIG_DIR`, so under pytest it read the floor's empty directory, found no settings, and returned "". Of the three routes the skip message names, only the explicit `ROMP_MOVE_LIVE_API_KEY_HELPER` could still work.

## Fix

Four test files change; no romp code does.

- `tests/conftest.py`: the location the run was handed is saved under `ROMP_TESTS_REAL_CLAUDE_CONFIG_DIR` immediately before the floor replaces it (captured after the floor, the saved value would be the floor itself). It is a `setdefault`, so an xdist worker keeps the controller's value rather than re-reading an environment the controller has already floored; the floor module's last two cases pin that. The name joins `_ENV_VALUE_PATH_NAMES` beside `CLAUDE_CONFIG_DIR`, a path a failure report may quote; an existing path is exempt already, so the entry matters only where the saved path does not exist.
- `tests/test_session_move_live.py`: `_user_api_key_helper` reads the saved location first, then `CLAUDE_CONFIG_DIR`, then `~/.claude`; an explicit `config_dir` argument still wins. The four existing `ApiKeyHelperPrecedence` tests clear the saved value inside their `mock.patch.dict`, since under pytest it names the operator's own settings and those tests must read only their temp dirs. The docstrings say where the value comes from.

Two choices:

- The borrow stays, rather than the explicit `ROMP_MOVE_LIVE_API_KEY_HELPER` becoming the only route: #946's point was a borrow that works wherever Claude Code is configured, and the skip message still describes it. This restores it under the floor.
- The new variable gets no per-test re-assert fixture: nothing pops the name outside a `mock.patch.dict`, which restores it, and a test that did would only make the live class skip. The floor's fixture guards a variable romp code reads at import; nothing in romp reads this one.

The saved value is a path under the operator's home. Nothing writes it to a file, and this PR does not quote it.

## Tests

Each of these fails on the base and passes with the change:

- `tests/test_session_move_live.py::ApiKeyHelperPrecedence::test_the_saved_real_config_dir_outranks_the_floored_one`: a "real" and a "floor" settings dir, `CLAUDE_CONFIG_DIR` at the floor and the saved variable at the real one; `_user_api_key_helper()` and `_api_key_helper()` return the real helper, and an explicit `config_dir` still wins. Before: `'/opt/example/floored-helper' != '/opt/example/real-helper'`.
- `tests/test_claude_config_floor.py::ClaudeConfigFloor::test_the_location_the_run_was_handed_is_saved_for_the_live_move_test`: the variable is set for every test and is neither the floor nor inside it. Before: not set. It does not assert which path it is, since a nested pytest run whose shell carries the outer run's value keeps it (the module's floor-override case runs that way).
- `tests/test_claude_config_floor.py::ClaudeConfigFloor::test_the_saved_location_is_the_root_the_shell_exported_before_the_floor`: a child pytest under `-p tests.conftest` whose shell exports a synthetic Claude root and carries no saved value runs one generated test, which passes only if the saved value is that root while `CLAUDE_CONFIG_DIR` is the floor. Before: `KeyError: 'ROMP_TESTS_REAL_CLAUDE_CONFIG_DIR'` in the child.
- `tests/test_claude_config_floor.py::ClaudeConfigFloor::test_an_xdist_worker_keeps_the_saved_location_the_controller_recorded`: the same child under `-n 2`, so the generated test runs in a worker whose environment the controller has already floored and stamped, and it still sees the handed root. Skipped where pytest-xdist is not installed (CI installs none). Before: the same `KeyError`, in the worker.
- `tests/test_env_value_redaction.py::RedactionRule::test_path_valued_shell_names_are_exempt_unless_credential_shaped`: the name is in the exempt-by-name mapping and `env_value_qualifies` is false for it. Before: `env_values_to_redact` returned the path.

One more pins the `setdefault` rather than the base's defect, and passes on the base, which never touches the variable:

- `tests/test_claude_config_floor.py::ClaudeConfigFloor::test_a_saved_value_the_shell_already_carries_survives_the_import`: the child's shell carries a saved value and a different `CLAUDE_CONFIG_DIR`; the generated test passes only if the saved value is the one handed in and `CLAUDE_CONFIG_DIR` is neither root. With the `setdefault` replaced by a plain assignment, this test fails (the child records the other root) and so does the xdist case (the worker records the controller's floor); the other new tests stay green under that change. Unlike the xdist case it needs no pytest-xdist, so it runs on CI.

Targeted run of `tests/test_session_move_live.py tests/test_env_value_redaction.py tests/test_claude_config_floor.py tests/test_key_source_floor.py tests/test_supervised_floor.py tests/test_state_isolation_order.py tests/test_tempdir_hygiene.py`: 93 passed, 1 skipped (the opt-in live class).

Full suite (`pytest -q -n 8 tests/`): 10570 passed, 115 skipped, 0 failed, 1688 subtests passed

Tier: docs

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -111,6 +111,14 @@ os.environ.pop("ROMP_STATE_DIR", None)  # a live kernel exports this to its sess
 # not reach a test either) before any test module loads, and re-asserted per test below so a
 # module-level pop or write in one test file cannot erase it for the run. A test that needs its own
 # Claude root sets the variable in setUp, after the fixture, exactly as the ones that do already do.
+# The location the run was handed is saved FIRST, before the floor replaces it: the one opt-in live
+# test that borrows the operator's apiKeyHelper command from their own settings
+# (tests/test_session_move_live.py) reads it through ROMP_TESTS_REAL_CLAUDE_CONFIG_DIR. Captured
+# after the floor it would name the run's empty temp dir, and that test would skip as "no auth"
+# while its skip message still named the borrow. setdefault, so an xdist worker keeps the
+# controller's value rather than re-reading an environment the controller has already floored.
+os.environ.setdefault("ROMP_TESTS_REAL_CLAUDE_CONFIG_DIR",
+                      os.environ.get("CLAUDE_CONFIG_DIR") or os.path.expanduser("~/.claude"))
 _CLAUDE_CONFIG = tempfile.mkdtemp(prefix="romp-tests-claude-")
 os.environ["CLAUDE_CONFIG_DIR"] = _CLAUDE_CONFIG
 
@@ -451,6 +459,9 @@ _ENV_VALUE_PATH_NAMES = frozenset((
     "PWD", "OLDPWD", "HOME", "PATH", "TMPDIR", "SHELL", "VIRTUAL_ENV", "PYTHONPATH", "LS_COLORS",
     "ROMP_SERVICE_ENV_FILE", "ROMP_SERVICE_ENV", "ROMP_DIR", "ROMP_STATE_DIR", "ROMP_CLAUDE_BIN",
     "ROMP_SYSTEMD_DIR", "ROMP_LAUNCHD_DIR", "CLAUDE_CONFIG_DIR", "TMUX_TMPDIR", "ROMP_TESTS_SYSTEM_TMPDIR",
+    # the Claude settings dir conftest saved ahead of its CLAUDE_CONFIG_DIR floor (above), for the live
+    # move test: a path a failure report may quote, like CLAUDE_CONFIG_DIR beside it
+    "ROMP_TESTS_REAL_CLAUDE_CONFIG_DIR",
     # GitHub Actions: the runner's workspace and tool cache, and the interpreter prefix setup-python
     # exports under six names (every stdlib and site-packages frame of a CI traceback is under it)
     "GITHUB_WORKSPACE", "RUNNER_WORKSPACE", "RUNNER_TEMP", "RUNNER_TOOL_CACHE", "pythonLocation",
--- a/tests/test_claude_config_floor.py
+++ b/tests/test_claude_config_floor.py
@@ -7,10 +7,17 @@ patching jd.PROJECTS wrote synthetic-sid directories under the developer's real 
 run's private temp root before any test module loads, and re-asserts it before every test, so a
 module-level pop or write in one test file cannot erase the floor for the rest of the run; a test
 that needs its own Claude root sets the variable in setUp, which runs after the fixture, as before.
+The location the run was handed is saved first, under ROMP_TESTS_REAL_CLAUDE_CONFIG_DIR, for the one
+opt-in test that borrows the operator's apiKeyHelper command from their own settings
+(tests/test_session_move_live.py); captured after the floor it would name the empty floor dir.
 
 Source pins in the style of test_supervised_floor.py, the floor observed from inside a test, from a
-module loaded under it, and from a subprocess whose shell exports a value of its own."""
+module loaded under it, and from a subprocess whose shell exports a value of its own; the saved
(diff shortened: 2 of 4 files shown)


ksarma:

Reviewed inline: the `setdefault` in `tests/conftest.py` sits on the line before the `CLAUDE_CONFIG_DIR` floor. Nothing above it alters `HOME` or `CLAUDE_CONFIG_DIR`. The saved location is therefore the one the run was handed, and the `~/.claude` fallback resolves the real home. `setdefault` keeps that value across an xdist worker and a same-process second execution; a probe confirmed both. The reader in `tests/test_session_move_live.py` takes an explicit `config_dir`, then the saved location, then `CLAUDE_CONFIG_DIR`, then `~/.claude`. Every `ApiKeyHelperPrecedence` case pins `HOME` and `CLAUDE_CONFIG_DIR` to temp dirs. Each pops or sets the saved name inside its patch, so none can read the operator's `settings.json`. `_skip_reason` returns at the `ROMP_MOVE_LIVE` gate, so the live class skips by default. No romp code reads `ROMP_TESTS_` names or scans `ROMP_` generically. `credentials.helper_env` forwards a fixed allowlist, so the `apiKeyHelper` child never sees the new name. The redaction exemption matches `CLAUDE_CONFIG_DIR` and `HOME` beside it. On CI the value is the runner's `~/.claude`. The `docs` label fits a tests-only diff, and the body matches the diff. The added lines carry no private strings, hostnames or home paths.

Follow-ups, neither a hold:
- `tests/test_ship_reship.py:142` (main, from #1235): `relaunch_env` keeps every `ROMP_*` name. After this merge the saved path therefore reaches a served lab's temp `cfg.json`, as `HOME` already does. Excluding `ROMP_TESTS_*` there would keep test bookkeeping out of the relaunched kernel's environment.
- `tests/test_claude_config_floor.py:125`: both child cases export `CLAUDE_CONFIG_DIR`. A third case, with it popped and a synthetic `HOME` and expecting `<home>/.claude`, would pin the fallback branch. A probe shows that branch correct today.

Red-first at base 82dd9e9c, running the PR's test modules over the base conftest with the helper hunk reverted: 5 failed, 1 skipped. Those are exactly the five named: the floored helper returned, the variable unset, `KeyError` in the child and in the xdist worker, the redaction returning the path. The setdefault pin is green at base and red under a plain assignment. Head: 94 passed, 1 skipped across the seven listed modules. CI is green on every check at d1b0430c.

Review run by the maintainer automated review session; verified against head d1b0430c.

ksarma:
Head note: the head moved from d1b0430c to b6d05b15 by a merge of main into the branch (this session updated the branch to re-run the checks against current main after a long hold); the merge is mechanical (the parents' merge-tree equals the head's tree) and the PR's added and removed lines are identical to the reviewed ones, so the review above stands for this head.

Review run by the maintainer automated review session; verified against head b6d05b15.

PR #1238, opened 2026-09-10
Tests: the served-page classes' copy of dist/ skips a concurrent build's staging files, and esbuild.js's failure line survives an empty errors list

ksarma:
## Symptom

This follows #1213 (merged) and the finding its post-merge review comment recorded: under a parallel local run (`pytest -n`), a served-page test class can error in `setUpClass` with a `shutil.Error` naming a hidden file under `dist/`, `.render.js.tmp-<pid>-<n>`, that no longer existed when the copy reached it. Local parallel runs hit it about one full run in two, three classes at once in one of them. CI runs pytest serially and is unaffected.

Two smaller findings from the same comment, in `vscode-extension/esbuild.js`: an error object with `errors: []` makes `failureSummary` throw a TypeError inside `main().catch`, so the process dies on the TypeError and the kernel's tail shows that stack in place of the summary line; and the 300-character cut of the last stderr line pairs with `_rebuild_dist`'s `.strip()[-300:]` in `kernel/kernel.py` with no test reading both.

## Cause

Each served-page class runs `node esbuild.js` into the shared `vscode-extension/dist/` and `shutil.copytree`s it into its lab. Since #1213 the script stages each output as a hidden sibling of its served name and renames it into place, and removes a staging file an exited build left behind. Two classes building at once: one build's rename or removal lands between the other's listing of `dist/` and its copy of that entry, and `copytree` collects the vanished file into a `shutil.Error`.

`failureSummary`'s per-error branch read `errors[0].location` with no length check. esbuild's API never yields an empty list, so no build reaches it; the throw sits in the handler that runs after a build has failed, where nothing catches it.

## Fix

- `tests/dist_copy.py`: `copy_dist(src, dst)`, a `shutil.copytree` with `ignore=shutil.ignore_patterns(".*.tmp-*")`, the shape `stagingPath` in esbuild.js writes (`.<name>.tmp-<pid>-<n>`). The served-page classes call it in place of their `copytree`, at seventeen call sites (the comment counted sixteen). The copy stays a copy: each class still builds for itself, as before.
- `vscode-extension/esbuild.js`: `buildErrors(e)` answers the error's messages, or null for a missing or empty list; `failureSummary` and `main().catch` both read it, so such an error gets the message line (`esbuild.js: build failed: <message>`) and its own stack printed, as any non-esbuild error does. The staging, the rename and the cut are unchanged.

## Tests

- `tests/test_dist_copy_staging.py`, 3 cases over scratch trees, no build, and a guard. First, a staging file present throughout (one beside a bundle, one beside an asset in its subdirectory) is skipped and the served files are copied byte for byte. Second, a staging file removed after the listing: `os.scandir` is replaced, for the source directory only, by a stand-in that takes the listing whole and then removes the staging file, which is where a concurrent build's rename lands. A plain `copytree` under it raises `shutil.Error` naming that file (the defect, asserted in the same case); `copy_dist` under it succeeds with the served files. Third, the `STAGING` pattern is read from esbuild.js, and every name it matches (each served name, three pids) is skipped while every served name is kept, a hidden name the script does not stage (`.keep`) among them, so the copy skips staging files and not every dotfile. Before the change the module these import is absent; with the copy routed through a plain `copytree` instead, the staging names appear in the copy in the first and third cases and the second raises the `shutil.Error`; with the pattern widened to `.*`, the third loses `.keep`. The guard reads every `tests/test_*.py` and refuses a `copytree(` call whose arguments name `dist`: the race fails no test of a class that copies raw until a parallel run lands on it, so a class put back on `shutil.copytree` would otherwise pass alone (checked by reverting one call site: the guard names its file and line).
- `vscode-extension/src/esbuild-build.test.ts`, two cases. `failureSummary` on an error with `errors: []` returns `esbuild.js: build failed: Build failed with 0 errors`; before: `TypeError: Cannot read properties of undefined (reading 'location')`. And `node esbuild.js` end to end under a stub esbuild (a preload replaces the package with one whose `build()` rejects with such an error, since the real API never yields one): exit 1, no TypeError, the error's own stack printed, the message line last, nothing written. Before the change the process died on the TypeError; with `main().catch` reading `Array.isArray(e.errors)` instead of `buildErrors`, the line survives but the stack is not printed, and the case fails on that.
- `tests/test_bundle_build_mode.py`, `FailureLineTail`: reads `_rebuild_dist`'s tail from kernel.py, `fit`'s cut and its ellipsis slice from esbuild.js, and `KERNEL_TAIL` from the two node test files that stand in for the kernel's tail, and holds them equal. A guard: green today, red when any number moves alone (checked by moving the kernel's to 200, the script's to 250 and a node test's to 299, each in turn).

Full runs on this head: `pytest -n 8 tests/` 10275 passed, 40 skipped, 0 failed, 1687 subtests passed; `npm run typecheck` clean; `npm test` 3780 tests, 3780 pass, 0 fail. An earlier full run of this branch had one served-page browser assertion fail under load, `test_ship_reship.py::ServedWedge` (a reload-timing check after a restart); its file's only change is the copy call in `setUpClass`, which succeeded, and the failing assertion reads nothing the change touches; it passed alone and in the following full run.

Tier: fix


ksarma:
Merging as shipped. One follow-up to record, nothing to fold.

Reviewed inline: `copy_dist` is `shutil.copytree` with `ignore_patterns('.*.tmp-*')`. `fnmatch` requires the leading dot. The glob therefore matches every name `stagingPath` writes (dot, served name, `.tmp-`, pid, counter) and no served asset or other dotfile. The third case reads `STAGING` from `esbuild.js` itself. All seventeen raw copytree sites at main are routed, and none remains under `tests/`. The only other dist copier in the repo, `tools/romp-lab/lab.sh`, is a dev script outside the race. Seventeen is right; the #1213 comment undercounted by one. In `esbuild.js`, `buildErrors` returns null for a missing or empty errors list, so `failureSummary` takes the message branch. `main().catch` prints the error's own stack, then the summary line last, then exits 1. By reading, the `BuildFailure` and non-esbuild paths are unchanged. The end-to-end node case rejects inside the stubbed `build()` before `mkdirSync` or any staging write, in a scratch cwd. No `dist/` appeared in either worktree after my runs. The served-page classes were checked by reading and by the guard, not run here. Body and commit message match the diff and quote no one. The privacy scan of the added lines is clean.

- `tests/README.md:63`, follow-up: the suite-wide conventions section records the hermetic temp and git rules and the credential redaction, each with the test that pins it. It does not record `copy_dist` or its guard, so a new served-page class learns the rule only when the guard fails. One sentence there would do.

Red-first at origin/main 82dd9e9c with the PR's test files: `tests/test_dist_copy_staging.py` fails to import (no `tests.dist_copy`). `FailureLineTail` is green at base, as a guard should be. `esbuild-build.test.ts` cases 6 and 10 fail with the TypeError at `esbuild.js:221` (8 pass, 2 fail). At head: the two pytest modules 12 passed, `esbuild-build.test.ts` 10 of 10, `npm run typecheck` clean. Every mutation the body describes was reproduced red. CI at 269d5eb1 is green on every check.

Review run by the maintainer automated review session; verified against head 269d5eb1.

ksarma:
Head note: the head moved from 269d5eb1 to 9cc809a9 by a merge of main into the branch (this session updated the branch to re-run the checks against current main after a long hold); the merge is mechanical (the parents' merge-tree equals the head's tree) and the PR's added and removed lines are identical to the reviewed ones, so the review above stands for this head.

Review run by the maintainer automated review session; verified against head 9cc809a9.

PR #1239, opened 2026-09-10
Tunnel rows: the recovery counter bumps once per Start, not once per answered poll under its hold (T291b)

ksarma:
## Symptom

During a Start (the popover's button on a no-kernel row), the row's recovery counter (`upSeq`, served on `/tunnels`) climbs by one on every supervisor pass whose poll answers while the Start is in flight, then once more when the hold clears: 1, 2, 3, 4 for one Start. The contract in `_note_recovery`'s docstring (#1209) is one bump per recovery. The dashboard is not affected today: `/tunnels` serves the row as `starting` while the hold stands, and federation's poll turns a counter change into `hostUp` only while the row reads `up`, so the user still sees one `hostUp` per Start. This is a defect in the counter's own contract, and a reader of `/tunnels` that does not apply the dashboard's status filter would see a recovery per poll.

## Cause

`_start_remote` holds the row at `status: starting` with `booting: True` while it updates and boots the far kernel, and the supervisor pass skips its status write under the hold (the `not r.get("booting")` guard) so the row does not flicker red mid-Start. The pass still runs `_note_recovery(r, st)` every tick, after `_note_poll` and before the status write. To the counter, an answered pass on a row whose status is not `up` is a not-up row coming up, and the held status never becomes `up`, so each answered pass under the hold bumps again.

For the record: #1209's body names an older `_note_recovery(r, st, had_miss)` with a pre-bookkeeping read of the misses; the merged function is `_note_recovery(r, st)`, called after `_note_poll`, reading the `_poll_miss` mark a silent pass leaves. This change is against the merged shape.

## Fix

`_note_recovery` returns without bumping while `r["booting"]` is set; the first answered pass after the hold clears finds the row not up and bumps once, the one recovery of that Start. The guard defers the counter's bookkeeping rather than discarding it: it sits after the not-answered branch, so a probe the far kernel refuses under the hold re-arms the demand path as it does off the hold, and it leaves the `_poll_miss` and `_demand_miss` marks on the row, so the pass after the hold spends them in its one bump, and a demand mark latches the demand path there as the first answered pass would have without the hold. The docstring carries the clause. Two guard lines and the docstring in `kernel/kernel.py`; no change to the call site, the `/tunnels` field or the dashboard.

## Tests

`tests/test_kernel_tunnel_truth.py`, class `RecoveryCounter`, through the class's `_pass` helper (the pass's order: `_note_poll`, then `_note_recovery`), with the status write held as the pass holds it:

- `test_a_start_hold_bumps_once_when_it_clears`: `booting` held across two answered passes, `upSeq` stays 0; the hold clears; the next answered pass bumps to 1; the steady pass after the status write does not bump. Before the change it fails on its first assertion: the first answered pass under the hold bumps (the unchanged code reads 1, 2, then 3 across the same passes).
- `test_a_silent_poll_under_a_start_hold_mints_no_second_bump`: answered, silent (the held status handed to `_note_recovery`, as the pass does on a silent poll), answered again under the hold: none bumps, and the `_poll_miss` mark the silent poll left is still on the row; the pass after the hold bumps once and spends it. Before the change it fails on its first assertion, for the same reason. A guard that spent the marks under the hold fails the mark assertion.
- `test_a_demand_mark_survives_the_hold_and_latches_in_the_one_bump`: a `_demand_miss` mark on the row when the Start begins (a request stalled, its woken pass found no kernel, which spends no mark, and Start was pressed) survives two answered passes under the hold; the bump after the hold spends it and latches the demand path, so the same request stalling again on the healthy link mints no second bump, and a real silent poll re-arms it. Before the change it fails on its first assertion (the first answered pass under the hold spends the mark and bumps). A guard that spent the marks under the hold mints the second bump.
- `test_a_not_up_probe_under_the_hold_still_rearms_the_demand_path`: a refused probe under the hold (`no-kernel`, an answered probe that is not up) pops `_demand_bumped` as it does off the hold, and the next demand mark after the hold bumps again. Green on the unchanged kernel: it pins the path the guard leaves alone, and a guard placed ahead of the not-answered branch fails it.

Red on the unchanged kernel with these tests: `tests/test_kernel_tunnel_truth.py` 3 failed, 30 passed. Green after: 65 passed across `tests/test_kernel_tunnel_truth.py`, `tests/test_kernel_remote_start.py` and `tests/test_remotes_stale_and_persist.py`. Full suite on this head (`pytest -q -n 8 tests/`): 10725 passed, 118 skipped, 0 failed, 1688 subtests passed.

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -17145,12 +17145,18 @@ def _note_recovery(r, st):
     timeout) mints one bump until a real silent poll or a status change re-arms it, so a request that keeps
     stalling on a healthy link cannot mint its own recovery, retry, stall, and mint again forever. A steady
     healthy row never bumps; a fresh boot starts at 0 (the counter describes THIS process, like the poll run
-    counters, and is not saved). Returns whether it bumped."""
+    counters, and is not saved). A Start in flight (`booting`, _start_remote's hold) owns the row's phase: the
+    pass skips its status write under the hold, so the row keeps reading "starting" however many polls answer
+    meanwhile, and every answered pass would count as a not-up row coming up and bump again (1, 2, 3, then
+    once more as the hold cleared). No pass under the hold bumps; the first answered pass after it clears
+    finds the row not up and bumps once, the one recovery of that Start. Returns whether it bumped."""
     answered = st == "up" and int(r.get("misses") or 0) == 0
     if not answered:
         if st != "up":
             r.pop("_demand_bumped", None)            # a status change re-arms the demand path
         return False
+    if r.get("booting"):
+        return False                                 # a Start hold: the pass after it clears bumps once (see above)
     real_miss = bool(r.pop("_poll_miss", False))
     demand_miss = bool(r.pop("_demand_miss", False))
     had_miss = real_miss or (demand_miss and not r.get("_demand_bumped"))
--- a/tests/test_kernel_tunnel_truth.py
+++ b/tests/test_kernel_tunnel_truth.py
@@ -100,6 +100,77 @@ class RecoveryCounter(unittest.TestCase):
         self.assertTrue(self._pass(r, answered=True))
         self.assertEqual(r["upSeq"], 1)
 
+    def test_a_start_hold_bumps_once_when_it_clears(self):
+        """_start_remote holds the row at "starting" (`booting`) while it updates and boots the far kernel, and the
+        pass skips its status write under the hold, so the row never reads "up" however many polls answer
+        meanwhile. Without a guard every answered pass under the hold looks like a not-up row coming up and bumps
+        again (1, 2, then 3 as the hold clears); the contract is one bump per recovery, when the hold clears."""
+        r = {"host": "TESTHOST", "status": "starting", "misses": 0, "booting": True}
+        self.assertFalse(self._pass(r, answered=True), "an answered pass under the hold is not the recovery")
+        self.assertFalse(self._pass(r, answered=True), "nor the next one: the hold owns the row's phase")
+        self.assertEqual(int(r.get("upSeq") or 0), 0, "no bump while the Start is in flight")
+        r["booting"] = False                    # the boot landed: the hold clears; the row still reads "starting"
+        self.assertTrue(self._pass(r, answered=True), "the first answered pass after the hold: the one recovery")
+        self.assertEqual(r["upSeq"], 1)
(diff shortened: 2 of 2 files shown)


ksarma:

Reviewed inline: the guard sits after the not-answered branch of `_note_recovery`. A probe the far kernel refuses under the hold therefore still pops `_demand_bumped`. The guard returns before the `_poll_miss` and `_demand_miss` pops. Both marks survive the hold and are spent in the one bump after it. The supervisor pass calls `_note_recovery` once per tick after `_note_poll`. It skips its status write while `booting` is set. The row therefore never read `up` under a Start, and every answered pass bumped again. With the guard, no pass under the hold bumps. The first answered pass after the hold clears bumps once and writes `up` in the same locked block. Federation turns that event into `hostUp`. The hold's success path clears `booting` and `detail` only. No writer sets `up` while `booting` stands, so the row still reads `starting` when the hold ends. A failed Start writes `no-kernel` and bumps nothing until a later answered pass. `_start_remote` has no try/finally around the hold, so an unexpected exception would leave `booting` set until a kernel restart. That is pre-existing. A row frozen at `starting` was already invisible to federation, so the counter freezing with it changes nothing the user sees. Body and commit match the diff. The fix label fits. The privacy scan of the added lines, body and commit is clean.

Tests:
- At base 82dd9e9c with the PR's test file, `tests/test_kernel_tunnel_truth.py` gives 3 failed, 30 passed. The failures are `test_a_start_hold_bumps_once_when_it_clears`, `test_a_silent_poll_under_a_start_hold_mints_no_second_bump` and `test_a_demand_mark_survives_the_hold_and_latches_in_the_one_bump`, each on its first assertion. The fourth is green at base, as stated.
- At head, that file plus `test_kernel_remote_start.py` and `test_remotes_stale_and_persist.py` give 65 passed.
- The two mutations the body describes fail exactly the tests it names.
- CI at 0aef7678: every required check green, Tier policy included.

Review run by the maintainer automated review session; verified against head 0aef7678.

ksarma:
Head note: the head moved from 0aef7678 to ba3069f4 by a merge of main into the branch (this session updated the branch to re-run the checks against current main after a long hold); the merge is mechanical (the parents' merge-tree equals the head's tree) and the PR's added and removed lines are identical to the reviewed ones, so the review above stands for this head.

Review run by the maintainer automated review session; verified against head ba3069f4.

PR #1240, opened 2026-09-10
Chat tabs: the no-strip `ready` holds for feed and timeline clients too, and the living-only ordered reader goes

ksarma:
Builds on #1218 (https://github.com/romp-on/romp/pull/1218), merged, which made the connect push the one source of the tab strip at `ready`. One commit.

## Symptom

None today. #1218 removed the strip the `ready` handler sent; this change keeps that strip from returning unnoticed on a feed or timeline socket and removes the reader it was built from. The removed strip went to every app's socket, and each pane's federation layer writes an inbound `tabOrder` frame into the stored arrangement (`federation.ts` `absorbHostReport`). A strip returning on a feed or timeline socket would prune the read-only reopened tabs from that store again at every page load, and #1218's test, which dispatches `ready` on a chat client only, would not notice.

## Cause

The test pinned the handler's behaviour for one app. `_ordered_alive`, the living-only ordered reader the removed strip was built from, stayed defined with no caller anywhere in the tree (`git grep`): chat tabs read `_chat_tab_sessions` and timeline lanes `_timeline_sessions`, both through `_ordered`. Only tests named it, as a stub or as the thing under test, so a strip could be rebuilt from it unnoticed.

## Fix

- `kernel/kernel.py`: `_ordered_alive` is removed; the `_tab_order_frame` docstring that named it says "a living-only liveness read".
- The tests that read the live order through `_ordered_alive` read the surfaces' own readers, `_timeline_sessions(live_only=True)` and `_chat_tab_sessions`. The tests that stubbed it against a returning strip stub `_alive_sessions`, the read such a strip would make.

## Tests

- `tests/test_kernel_tabs_first.py`: one `ready` per app through the real dispatcher, the connect push stubbed as a marker. `test_connect_ready_handler_sends_no_tab_order_of_its_own` (chat), `test_a_feed_clients_ready_yields_no_tab_order_frame` and `test_a_timeline_clients_ready_yields_no_tab_order_frame` each expect the push, then the caps frame, no `tabOrder`, and no `("taborder",)` dedup slot written (the slot is where a strip sent through `_send_client` records itself, so it reads the same frames a second way). The feed and timeline cases pass before this commit; they are guards. On the code before #1218 all three fail with `['_pushed', 'tabOrder', 'caps']` (run against that tree's kernel), and a strip re-added at `ready` for non-chat sockets only turns exactly the two new cases red. `test_no_living_only_ordered_reader_remains` fails before this commit because the module still defines `_ordered_alive`; it is a check on the name, and the behavioural coverage of the removal is the re-aimed cases below, which go red when either reader skips the shared order.
- `tests/test_kernel.py`: `test_alive_filter_empty_tmux_with_tmux_present_shows_nothing` asserts no chat tabs and no live timeline lanes through the two readers; `test_live_order_is_stable_under_activity` (was `test_ordered_alive_is_stable_under_activity`) and `test_session_order_roundtrip_and_sort` read the order as timeline lanes and as chat tabs; `SessionOrderStable` no longer stubs the removed name.
- `tests/test_tag_edit_ack.py`: `ReadyStripSource` stubs `_alive_sessions` in place of `_ordered_alive`; its three tests are unchanged. The `Capability` tests there already dispatch `ready` on a timeline socket and assert no `tabOrder`, so the timeline case was covered implicitly; the new cases state it per app and add the feed.
- `tests/test_kernel_order.py`: the module docstring names the readers that remain.

Targeted: 209 passed across `test_kernel_tabs_first.py`, `test_kernel_order.py`, `test_tag_edit_ack.py`, `test_chat_skeleton_reconnect.py`, `test_timeline_views.py` and the touched `test_kernel.py` cases; `test_kernel.py` whole: 485 passed (531 together with `test_kernel_order.py` and `test_kernel_tabs_first.py`). Full suite (`pytest -q -n 8 tests/`): 10208 passed, 62 skipped, 0 failed.

Tier: fix


ksarma:

Reviewed inline: the kernel change removes `_ordered_alive` and rewords one `_tab_order_frame` docstring phrase. `_ordered_alive` has no caller at the head; git grep finds the name only in the new test. The `ready` arm still calls `_push_one` and then `_send_caps`, with no strip of its own and no liveness read. The new per-app cases drive the real dispatcher with the connect push stubbed as a marker. The dedup-slot assertion reads the `sent` map `_send_client` writes, so it is a second reading of the same frames. Both re-aimed readers, `_timeline_sessions(live_only=True)` and `_chat_tab_sessions`, resolve through `_ordered`. A reader made to skip it turns `test_session_order_roundtrip_and_sort` red, so the shared-order property the removed docstring stated stays pinned. The tier is `fix`: kernel code changes, no capability is added, and one test is red before the commit. The body states that the feed and timeline cases are guards with no symptom today, a preventive guard of the kind #1200 merged as. Body and commit match the diff; the privacy scan of the added lines is clean.

Follow-up: tests/test_kernel.py:6546 gives both stubs the same input order. A timeline reader that skipped `_ordered` would therefore still pass the stability check. A second stub in mtime order ([B, A]) would make the check catch that.

Tests: at base 82dd9e9c with the PR's test files, `test_no_living_only_ordered_reader_remains` fails and the three `ready` cases pass, as the body says. A strip re-added at `ready` for non-chat sockets turns exactly the two new cases red. At head: 204 passed across the five listed modules, 485 in test_kernel.py. CI at 693bd0e2: all checks green.

Review run by the maintainer automated review session; verified against head 693bd0e2.

ksarma:
Head note: the head moved from 693bd0e2 to b93c9b4f by a merge of main into the branch (this session updated the branch to re-run the checks against current main after a long hold); the merge is mechanical (the parents' merge-tree equals the head's tree) and the PR's added and removed lines are identical to the reviewed ones, so the review above stands for this head.

Review run by the maintainer automated review session; verified against head b93c9b4f.

PR #1241, opened 2026-09-10
Timeline: a live lane's segment part is derived once and served while its inputs stand

ksarma:
## Symptom

Every bars build re-derived every live lane's segment part: the seam-aware segmentation, the awake-span excision, the caption lookups, the compact bar dicts and the judging marks were computed again for each live lane on each build, although between two builds most live lanes had not changed (a board of dozens of lanes has one or two writing at a time). The dead lanes have been served from their memo since #1131 and #1175; the live lanes, the ones being watched, paid the whole loop every cycle. Measured on an earlier form of this change (the same memo over long-key bars, with the captions keyed by object identity) on a 31-session state copy: the bars build fell from 477 ms to 93 ms at full hit, and a 30-build replay at the pusher's cadence served 99.3% of live lanes after the first build.

## Cause

`build_timeline` ran the per-lane turn loop inline with no memo for a live lane (the dead-lane memo is gated on `not live`), and `_derive_judging` applied the horizon inline beside the derivation, so nothing clock-free existed to hold across builds.

## Fix

Three commits, each droppable from the tail.

**1. The judging marks are derived apart from the horizon (a pure refactor).** `_derive_judging_marks(sid, caps, goals, seg_ends=None)` returns `(cap_marks, other_marks)` with no clock: the captioner's marks in caption-time order and `[(filter_t, mark)]` pairs for every other judge, each carrying the value the one-pass form compared against the horizon (the diary event's `ev_t` for a done, block or close; `distilledMt` or `briefedMt` for the distiller; the mark's own t otherwise), kept beside the mark because `seg_ends` moves a completion's plotted t to its segment's work end. `_judging_assemble(cap_marks, other_marks, t0, out, stamp=False)` applies the horizon and `JUDGE_CAP_LIMIT` per build; with `stamp` it appends copies carrying `_h`, since the pairs may be a memo's shared objects. `_derive_judging` is the two-line composition and keeps every caller, the dead-lane memo's stamped call included.

**2. The live-lane memo.** The inline loop moves to `_lane_segments(sid, session, goals, caps, live, bft, full_prompts=None)`, which emits the compact wire bar exactly as the loop did (the `BarLiteral` and bars-wire pins pass unchanged), derives the marks through `_derive_judging_marks` under the lane's own guard, and raises when a mark carries a non-numeric time, so a malformed row costs the lane its marks once per build and never the frame. `_lane_memo(sid, parsed, session, goals, caps, cap_key, live, bft, parse_ok, full_prompts)` serves a live lane's segment part from `_lanes_memo` while its inputs are the previous build's:

- the parse object by identity, after the live merge (`_parse` returns the cached object for an unchanged transcript and states file; a merged live tail is a new object and is derived, `live_tail`);
- the goal store by identity when a `FrozenStore`, or as "empty" when it has neither seams nor nodes (a private store with content is derived and not held, `unshared_skip`);
- the captions file's `_stat_key`, taken in `build_timeline` before `_captions` reads it. `_captions` builds a new dict per call, so the file is the input; stat before read means a row landing between the two is held under the old key and misses on the next build, never served stale (a stat after the read would hold the old rows under the new key). Rows read when the stat found no file have no key to be held under: derived, not held, and held on the next build;
- the branch clip, `tuple(_downtime)` and the archive's `_file_key`, taken before the derivation (a stat sentinel matches nothing and stores nothing).

A lane whose parse failed or whose seams or marks stage complained is derived and not held (`complain_skip`). The horizon and the caption cap are applied per build by `_judging_assemble`, on a hit as on a miss; the dead-lane path assembles its stamped marks the same way. A hit hands the binder the lane's whole prompts from the entry (T278b), so a delivery that names its sender on a later line of the prompt binds on a served build as on a derived one. Entries leave with a full build's lane set (`_lanes_forget`; a skeleton or live-only build evicts nothing), past 256 the least recently served first, when the parse object is no longer the build's, and when the dead-lane populate takes the lane (the entry holds the parse the populate releases). One lock covers get, put, evict and the counters; the derivation runs unlocked, so two threads deriving one lane both store an exact entry and the last wins; when both derived from one parse object the next build is served, otherwise it derives once more and holds again. The held bars and marks are shared by identity into `turns`, the bars wire cache, the delta parts and `semantic`; at this base none of those writes to them (`_bind_message_execs` mutates the messages only, `_timeline_skeleton` copies the frame, `_run_judging` reads, `_expand_bar` builds a new dict).

A dead miss is derived by `_lane_segments` directly and cached in the dead-lane memo, with a `failed` flag on a parse that raised. One dead-lane rule changes: a dead lane whose seams stage complained is derived every build and not cached, as a lane whose marks stage failed already was (the inline loop cached it after a seams complaint, so a malformed goals row was said once and then served). The populate also pops the live memo's entry. `_VIEW_STATS["laneServe"]` keeps counting a dead-lane serve under `views`. `GET /perf` gains `memos.lanes`: `hit`, `miss`, `live_tail`, `complain_skip`, `unshared_skip`, `evict`, `entries`, `segs_hit`, `segs_miss` for the live lanes and, so one block carries every lane, `dead_serve`, `dead_miss`, `dead_failed_serve` (`views.laneServe` counts a dead-lane serve whether or not its parse raised, so it equals `dead_serve` plus `dead_failed_serve`; retiring the older counter is a separate call). docs/reference.md describes the block.

**3. `_cached_timeline` reads the payload slot once.** It tested `e[1]` and returned `e[1]` as two reads of the shared list while the pusher assigns `_built_timeline[:]` on its thread, so a connect whose reads straddled the assignment was handed the replacement build, not the one its freshness test had seen.

Not changed: what a frame contains (a served lane is the rebuilt lane, byte for byte; the badge row is derived per build as before), the rebuild cadence, and the skeleton build, which still loads every lane's store (gating that load to dead lanes was left out: `tests/test_goal_store_fault_boundary.py` pins that a live lane's skeleton files the store fault row).

Pins moved: the frozen-view spy in `tests/test_kernel_goal_cache_wiring.py` aims at `_derive_judging_marks`; the stage pins in `tests/test_timeline_bars_resilience.py` read `build_timeline` and `_lane_segments`; `ui/timeline-lineage.test.ts` reads the clip's `bft` (no new kernel-text pin in TypeScript; the clip's behaviour, its `<=` boundary included, is pinned in Python).

## Tests

Each test was run on the sources before its commit first and failed as stated.

Commit 1, `tests/test_timeline_lane_memo.py` (DerivationSplit, 5 tests): a private copy of the one-pass form is the oracle; the wrapper and the two halves equal it at nine horizons, two of them inside a completion's evidence-to-work-end window, with and without `seg_ends` over a store carrying a planner mint and sub, a courier plant, an umbrella, a grouper op, done/block/close diary rows, a synth row, a `distilledMt` and a `briefedMt`, plus an archive and more captions than the cap; a timeless group op or diary row is dropped as before; the marks come out once with their filter times and the completion's plotted t at the work end, and an assembly at a horizon between the two drops the completion and keeps a later one; the stamped assembly copies and leaves the derived marks clean; the derivation reads no clock (the module's `time` patched to raise). Red: 4 of 5 with `AttributeError: _derive_judging_marks` (the wrapper-only case passed already). A mutation that filters the diary marks on the plotted work end instead of `ev_t` fails two of them.

Commit 2, `tests/test_timeline_lane_memo.py` (a LaneMemoBase fixture with a fresh state root per test and one live lane under a private synthetic sid; 44 tests in seven classes). Red on the tree before the commit: every one of the 44 at the fixture, which clears `_lanes_memo` and saves `_LANES_MEMO_MAX` (`AttributeError`). With those fixture references guarded for the measurement, the reds were `AttributeError: _lanes_memo_report` and `_lane_segments` in the bodies, `KeyError: 'failed'` on the dead entry, `KeyError: 'lanes'` on the snapshot, the guard's stderr line absent, and the behavioural cases: the seam stage ran on the second build (`2 != 1`), the second frame's bars list `is not` the first, a complaining dead lane was cached.

- HitsAndEquality: an unchanged live lane is served without walking its turns, with the same bars list and bar objects in `turns`, equal to the uncached derivation and to `_derive_judging`'s marks, hit 1 miss 1; a hit still runs the parse, the merge and the captions read; the held bars survive the postal join; the badge row changes on a hit; a hit hands the binder the whole two-line prompt and the sender named on its second line binds, where the wire's first line alone does not; the real judging band reader over a hit leaves the held marks unchanged; two threads deriving one lane from one parse object leave one exact entry and the next build is served (a `Barrier`, no timing).
- EveryInputBusts, one per key input: a transcript append, a states write, a store publish, a journal append, a captions append, a caption landing between the stat and the read (misses on the next build, holds again after), an archive rewrite and an archive appearing, a host suspension (a rebind of the same content hits), a suspension inside the segment splitting the bar on the next build, liveness (a lane going dead is derived once outside this memo, handed to the dead-lane memo, its live entry and parse dropped, then served with `laneServe` and `dead_serve` both counting), an open turn open only on a live lane, the branch clip moving for a live child when the parent leaves the build (a compaction inside the copied history and a turn whose work ends exactly at the fork time are the parent's while its lane is in the build and the child's after).
- NotHeld: a live tail, a failed parse (one stderr line), a private store with content, a complaining seams stage, a complaining marks stage, a caption row or an archive with a string time (the lane keeps its bars and loses its marks on both builds), a failed archive stat, captions read when the stat found no file (held on the next build); an empty store and an empty captions file hit under the empty rule; the skeleton build bypasses the memo; the assembly is guarded at its call site.
- ClockAtAssembly: a t-less caption is dropped on a hit as on a miss; the horizon and the caption cap apply per build on a hit; `_lane_segments` reads no clock.
- Eviction: with the lane set and not on a live-only build; least recently served first at the bound (the first-inserted entry is served, so insertion order and serve order differ); an entry whose parse is no longer the build's.
- DeadLanesOnTheSameBlock: a dead miss leaves this memo and the parse cache empty and counts `dead_miss`; a failed dead parse serves as `dead_failed_serve` until the transcript's stat moves; a complaining dead lane is derived every build and not cached.
- PerfWiring: `memos.lanes` on the snapshot with the live and dead counters apart.
- `tests/test_perf_stats.py`: the memos key set gains `lanes` with its twelve integer keys (red: `KeyError: 'lanes'`). `tests/test_kernel_goal_cache_wiring.py`: the spy re-aim (red: `AttributeError: _lanes_memo` in setUp). `tests/test_timeline_bars_resilience.py`: the stage pins over both functions (red: `AttributeError: _lane_segments`). `ui/timeline-lineage.test.ts`: the clip regex (red: `not ok 3`).

Mutations each caught by a named test: the LRU reinsert on a hit removed, the compaction markers not clipped at the fork time, the clip's `<=` made `<`, the prompts map not filled on a hit, the captions rows read with no file to stat held under the empty key, a seams complaint no longer blocking the dead-lane cache.

Commit 3, `tests/test_perf_stats.py` (PusherRecords): `_built_timeline` is replaced by a list whose payload slot answers the build on its first read and a replacement build after; a connect returns the build it tested, counts the serve and never rebuilds. Red: the replacement came back.

Full runs on the final tree: `pytest -n 8 tests/`: 10882 passed, 41 skipped, 1688 subtests passed, 151 s, no failures. `npm run typecheck` clean; `npm test`: 4052 pass, 0 fail; `node --test` on the lineage and bars-wire bundles: 3 and 6 pass.

Tier: fix

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1058,7 +1058,17 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   sessions that left the alive set. The interrupt tick drops from `intrMarks`
   and `statesOverlay` the entries of sessions outside its alive set each
   cycle; the `statesOverlay` cache is also cleared whole above 256 entries, a
-  drop `evict` does not count and `entries` shows.
+  drop `evict` does not count and `entries` shows. `lanes` is the timeline's
+  per-lane segment memo: a live lane's bars, segment ends, last activity,
+  compaction markers and judging marks, held while its parsed transcript and
+  goal store are the previous build's objects and its captions file, archive
+  file, branch clip and the host's recorded suspensions stand. One outcome per
+  live lane per bars build: `hit`, `miss`, `live_tail` (a live tail was merged,
+  so the lane was derived and not held), `complain_skip` (the parse or a stage
+  failed) and `unshared_skip` (a private store with content); `evict` and the
+  gauge `entries`; `segs_hit` and `segs_miss` count the segments served and
+  derived. `dead_serve`, `dead_miss` and `dead_failed_serve` are the dead-lane
+  memo's outcomes on the same block, so one block carries every lane.
 - `judge`: `passes`, `ms_sum`, `ms_last`, `ms_mean` (wall time; a pass waits
   on model calls), `cpu_ms_sum` (CPU time of the judge tier threads and every
   per-session worker they run; the workers' share is `cpu_ms_workers`).
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -407,7 +407,8 @@ class _PerfStats:
                           ("backref", jd.backref_memo_stats), ("captions", jd.captions_memo_stats),
                           ("goalArchive", jd.goal_archive_memo_stats), ("plannerSkip", jd.planner_skip_stats),
                           ("liftGate", _lift_gate_report), ("bgTops", _bg_tops_report),
-                          ("intrMarks", _intr_marks_memo_report), ("statesOverlay", _states_overlay_report)):
+                          ("intrMarks", _intr_marks_memo_report), ("statesOverlay", _states_overlay_report),
+                          ("lanes", _lanes_memo_report)):   # the timeline's per-lane segment memo, live lanes; the dead lanes beside
             try:
                 memos[key] = read()
             except Exception:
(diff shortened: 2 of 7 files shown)


ksarma:
Merging as shipped, under the `fix` tier. One follow-up to file, below. The change is a performance memo over an unchanged frame, plus a pure refactor and a race fix. That is the same class as #1131, #1175, #1228 and #1230, which merged as `fix`.

Reviewed inline: the memo key, checked against every input `_lane_segments` reads. The key holds the parse object, by identity after the live merge, and the `FrozenStore`, by identity or the empty rule. It holds the captions stat, taken in `build_timeline` before `_captions` reads, and the archive `_file_key`, taken before the derivation. It also holds the branch clip, the downtime tuple and liveness. A live tail, a failed parse, a private store with content and a complaining stage are derived and not held. The held bars and marks are shared by identity into turns, the bars wire cache, and the delta parts and semantic. Every consumer downstream of `build_timeline` reads or copies, and none writes in place: `_bind_message_execs`, `_timeline_skeleton`, `_run_judging`, `_compact_judging`, `_expand_bar`, `_delta_split`, `_bars_wire`. The memo path only reads the `FrozenStore`. One lock covers get, put, evict and the counters. The entry tuple is complete before the put. The LRU reinsert on a hit and the 256 bound happen under the lock. An entry whose parse is not the build's is popped. The dead-lane populate pops the live entry. A complaining dead lane is derived every build, as the body discloses. Commit 3 binds the payload slot once; the freshness test and the return use that one object. `HitsAndEquality` compares the served lane with the uncached derivation in wire form.

- `kernel/kernel.py:36636`, follow-up: a faulted store (`goals is None`) and an empty private store both key as "empty". `_lane_segments` derives no marks for `None`, but the empty dict still yields the captioner and archiver marks. So a live lane held across a fault-to-no-store transition, or the reverse, is served marks that differ from a fresh derivation. The served marks differ until the parse, captions, archive or store key changes. Reproduced both ways at head. The dead-lane path never caches a fault; treat `goals is None` as a skip here too, with a test for the transition.
- `kernel/kernel.py:37306`, note: `dead_serve` does not mirror `views.laneServe` as the body says. `laneServe` counts failed serves too, so it equals `dead_serve` plus `dead_failed_serve`.

Tests: red-first per commit in scratch worktrees. `DerivationSplit` at the base: 4 failed of 5, on the missing `_derive_judging_marks`. The four pytest files at 4275b838 with commit 2's tests: 54 failed, 78 passed, for the reasons the body lists; the lineage node test: not ok 3. `PusherRecords` at 1315abb1: 1 failed, with the replacement build returned. At head 924051e5: 133 passed over `tests/test_timeline_lane_memo.py`, `test_perf_stats.py`, `test_kernel_goal_cache_wiring.py` and `test_timeline_bars_resilience.py`; lineage 3 of 3; typecheck clean; four body mutations each caught by the named test. CI at this head: every check green.

Review run by the maintainer automated review session; verified against head 924051e5.

ksarma:
Head note: the head moved from 924051e5 to eb14583a by a merge of main into the branch (a branch update after the record posted); the merge is mechanical (the parents' merge-tree equals the head's tree) and the PR's added and removed lines are identical to the reviewed ones, so the review above stands for this head.

Review run by the maintainer automated review session; verified against head eb14583a.

PR #1242, opened 2026-09-10
Undo clear: an undo-reopen with nothing to restore leaves the state as it stands, so a completed top comes back completed, not Working

ksarma:
Follows the merged #1233 (the same-second replay fix): its test helpers (`SEAL_SID`, `_completed_top(sid=)`, `_seal_rows`, `_user_seals`) and the replay behaviour the second test depends on live there, and its body names the raw-snapshot shape below as not changed there.

## Symptom

A completed top the user cleared, undid, cleared again and undid, all inside one second, comes back in Working rather than Completed: uncleared, as it should be, but no longer completed. It happens when a triage pass that loaded the store after the first clear publishes afterwards through `save_goals`. The same happens when a clear and its undo, a second apart, are both lost to a pass save from a snapshot taken before either. Neither heals on a reload: the next load re-records nothing, and a further save and load add nothing.

## Cause

Two shapes put an undo-reopen into a node's log with no clear before it:

- `_rebase_onto_disk` keys verdict identity on (ev_t, src, kind) and filters the disk's rows against the writer's log only, never against each other. With the pass's copy holding the first clear at T, the disk's re-clear (T, user, clear) is a twin and drops, while both undo-reopens (T, user, reopen) are kept: the published log reads clear, reopen, reopen.
- Over a snapshot that predates a clear and its undo, the journal's last word for the node is the undo, so the clear row never replays (`last_seal`); the unclear row re-records its reopen on its own. #1233's body names this shape as unchanged there.

In `_fold_node` the undo branch restored the snapshot its clear took only when one stood (`e.get("undo") and clear_snap is not None`). With none, the row fell through to the plain reopen's path: state "open", the settle ended, and `rollup_status` filed the top as working. The override replay cannot heal either shape: every row's word is already in the log (`_newest_seal` and `last_seal` both skip), so nothing re-records and the published "working" stands.

## Fix

`kernel/judge.py`, `_fold_node`: the undo branch short-circuits whether or not a snapshot stands. With one it restores it as before; with none it changes nothing: the state, the settle stamp and `held` stay as they are, and the user's floor still advances. The plain reopen's `held` line drops its undo check, which no undo row reaches now.

Why this is the right reading of the log: an undo asserts nothing about doneness (the existing comment), and by the two writers' gates an undo-reopen that meets no clear is exactly a lost clear or a second undo for one clear. The live undo (`_mark_nodes_cleared`) records a reopen only when the flag was on, the replay's unclear arm only when the journal's last word for the node is not a clear, and `migrate_store` gives a legacy cleared node a synthesized clear. A node whose log was capped (`logTrunc`) with its clear dropped hands `_fold_node` the rows of the unit test's no-clear case (done, settle, undo-reopen) and now reads its state from them instead of opening, which is the intended reading.

Why not a dedup of the disk's rows against each other in `_rebase_onto_disk`: it would collapse the second reopen only when both undos share a second (with the last undo a second later the identities differ, the clear rows still never replay, and the top still opens), the raw-snapshot shape has no rebase at all, and it would change the dedup for every verdict kind (two same-second romp settles are legitimate rows today).

Not changed: `_rebase_onto_disk`, `_replay_overrides`, `record_verdict`, the kernel's writers, and the sort key (the shuffle-invariance property stands). `docs/goal-state.md` gains one clause.

## Tests

The first three cases below are red before this change and green after.

- `tests/test_judge_guarded_node.py`, `TheNewKinds.test_an_undo_reopen_with_nothing_to_restore_is_a_state_no_op`: over the real `_fold_node`, a log of done, settle, clear, undo-reopen, undo-reopen leaves the state done with the settle stamp and the floor at the last undo, and a log of done, settle, undo-reopen with no clear does the same. Fails before with `AssertionError: 'open' != 'done'`. A third case in the same test, on a node no verdict has landed on (clear, undo-reopen, undo-reopen), pins that the undo never holds the node open: a done node's `held` is masked at `_fold_node`'s return, so this is the shape where a branch that set `held` would show. It is green before the change too; a mutation that sets `held` in the undo branch fails it.
- `tests/test_kernel_goal_compaction.py`, in the class #1233's same-second tests live in, under its own sid with `time.time` pinned to one second: `test_a_stale_pass_save_that_collapses_the_re_clear_leaves_the_top_completed`. Clear, a pass loads its copy, undo, clear, undo; the pass publishes through `save_goals`. Premises asserted: four journal rows in order, the copy holding the first clear only, the live store completed and uncleared before the publish, the published log clear, reopen, reopen with the card uncleared. Fails before at the published store's status with `AssertionError: 'working' != 'completed'`; also asserts the next load reads it completed and re-records nothing, and a further save and load add no row.
- Same class, `test_an_undo_over_a_snapshot_taken_before_its_clear_leaves_the_top_completed`: clear at T, undo at T+1, overwrite with a snapshot taken before both, load. Premises: the two journal rows, the snapshot holding no seal and a completed top. Asserts the card uncleared, exactly one reopen re-recorded, and the status completed (fails before with `AssertionError: 'working' != 'completed'`); a further save and load add no row.

Mutation checks on the new branch, each run against the two test modules: ending the settle, skipping the floor advance, opening the node, and setting `held` for a user undo each fail at least one case above; restoring the plain reopen's `if not e.get("undo")` check changes nothing (it is dead code now).

Targeted run after the change: test_judge_guarded_node, test_kernel_goal_compaction, test_kernel_undo_restore_completed, test_ask_unit_cards, test_event_model_golden, test_reopen_next_word, test_kernel_goal_cache_wiring and test_judge_verdict_log (the verdict-log module), 209 passed. Full suite: `pytest -q -n 8 tests/`, 10758 passed, 118 skipped, 1688 subtests passed.

Tier: fix

### DIFF
--- a/docs/goal-state.md
+++ b/docs/goal-state.md
@@ -47,7 +47,9 @@ Sort events by evidence time (`ev_t`; arrival `at` breaks ties) and replay:
   clears.
 - **Snapshots**: provisional flips restore what they displaced. A `clear`
   snapshots the state it covers, and an undo-reopen restores it, so a
-  cleared completed card comes back completed, never "open". A msg-reopen
+  cleared completed card comes back completed, never "open". An
+  undo-reopen with nothing to restore (its clear lost, or a second undo
+  for one clear) leaves the state as it stands. A msg-reopen
   (the optimistic flip when you reply to a card) snapshots too, and a
   planner `dismiss` (the pivot verdict: that reply started its own thread)
   restores the original state and settle stamp. The pivot's new goal is its
--- a/kernel/judge.py
+++ b/kernel/judge.py
@@ -9090,7 +9090,8 @@ def _fold_node(nd):
     #                                       verdict: "that reply wasn't about this goal") restores it
     clear_snap = None                     # symmetric snapshot at `clear`: an undo-reopen restores the state
     #                                       the cross-off displaced (a cleared COMPLETED card comes back
-    #                                       completed, never "open"), instead of blindly opening
+    #                                       completed, never "open"), instead of blindly opening; an undo
+    #                                       with nothing to restore leaves the state as it stands
     for e in sorted(nd.get("log") or [], key=lambda e: (e.get("ev_t") or 0, e.get("at") or 0)):
         src, kind, t = e.get("src"), e.get("kind"), e.get("ev_t") or 0
         if kind == "reopen":
@@ -9105,9 +9106,17 @@ def _fold_node(nd):
                 # assert's own `t >= floor` equality-lands rule below: within one turn the reopen is the
                 # trigger, the wait is how the turn ENDED.
                 awaiting_why = awaiting_at = awaiting_kind = awaiting_peers = None
-            if e.get("undo") and clear_snap is not None:
-                state, cur_settle, prev_settle = clear_snap      # restore what the cross-off displaced
-                clear_snap = None
+            if e.get("undo"):
+                if clear_snap is not None:
+                    state, cur_settle, prev_settle = clear_snap  # restore what the cross-off displaced
+                    clear_snap = None
+                # An undo-reopen with NO clear before it has nothing to restore and is a state no-op: it
+                # asserts nothing about doneness, so a completed top stays done and its settle stands. Two
+                # shapes reach here (2026-09-10): a second undo row for one clear, when a same-second re-clear
+                # collapsed into the first clear as a rebase twin while both undo-reopens were kept; and an
+                # undo whose clear never landed, when a clear and its undo were both lost to a pass save and
+                # only the undo replays (the journal's last word). Both used to fall through to "open" like a
+                # plain reopen, and the completed top came back Working. The user's floor still advances.
                 if src == "user":
                     floor = max(floor, t)
                 continue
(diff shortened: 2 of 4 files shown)


ksarma:

Reviewed inline: the undo branch of `_fold_node` in `kernel/judge.py` now ends the row with or without a clear snapshot. An undo-reopen that meets no clear therefore leaves the state, the settle stamp and held unchanged while the user's floor still advances. The undo check dropped from the plain reopen's held line is dead, because no undo row reaches it. We read both undo writers. The live undo in `_mark_nodes_cleared` records a reopen only when the flag was on. The replay's unclear arm re-records only when the journal's last entry for the node is not a clear. `migrate_store` synthesizes a clear for a legacy cleared node. So a snapshot-less undo is either a lost clear or a second undo. Constructed logs at the head gave the expected column in every shape. A plain user reopen still opens and holds. An undo after a clear restores. An undo with no clear leaves a done node done and a blocked node blocked. A later judge verdict at or after the floor still lands. Each shape folded identically over hundreds of shuffles. The feed and the reopen gate read `cleared.jsonl`, a path separate from the fold, so a capped log that dropped its clear stays hidden. The docs clause matches the code and the body matches the diff. The fix label fits. The privacy scan of the added lines is clean.

Red-first is met by our own run. With the PR's two test files at the base, exactly the three named cases fail and 29 others pass: `test_an_undo_reopen_with_nothing_to_restore_is_a_state_no_op` with `'open' != 'done'`, and the two compaction cases with `'working' != 'completed'`. At the head the eight named modules pass, 209. Each of the four mutations of the branch fails a named case. CI at the head is green on every required check.

Review run by the maintainer automated review session; verified against head 5af375eb.

ksarma:
Head note: the head moved from 5af375eb to ade655e0 by a merge of main into the branch (this session updated the branch to re-run the checks against current main after a long hold); the merge is mechanical (the parents' merge-tree equals the head's tree) and the PR's added and removed lines are identical to the reviewed ones, so the review above stands for this head.

Review run by the maintainer automated review session; verified against head ade655e0.

PR #1243, opened 2026-09-10
Show completed: a cleared archived top whose only completion is a copied status leaves the list

ksarma:
Builds on #1234 (https://github.com/romp-on/romp/pull/1234, merged): this branch is cut from its head and adds one commit; the change to review is the last commit. #1234 fixed the compaction's copy from then on and named what it left: an archive the compaction wrote before it keeps the copied "completed" for a root only cleared.jsonl cleared, so that top stays listed until restored and cleared again. This closes that.

## Symptom

Under Show completed, a top the user cleared stays listed, struck through, with its subtree, when the compaction archived it before #1234: its clear survived only in cleared.jsonl, the compaction re-sealed the root and copied the pre-rollup "completed". The same row appears for a completion the rollup derived without a verdict (an umbrella, a settled store) whose root the ledger later cleared, among them a root whose completion was derived and whose flag a pass save erased before compactions re-sealed such roots (copied "completed", flag off, ledger row present); a root with its own done verdict keeps nodeComplete through that erasure and is not this case. A clear on the live card takes such a top off the list; #1187 states that as the intended outcome.

## Cause

`_fleet_archived_tops` qualifies an archived root by nodeComplete, a copied status of "completed", or a summary, and `_ledger_cleared_overlay` only marks a cleared root (and its subtree) cleared; it never drops a row. A re-sealed root's nodeComplete is False (the clear verdict's derivation), so the copied "completed" was its only qualification, and the overlay left it listed as done and cleared: the struck-through row. Nothing rewrites an archived status, nothing prunes the ledger, so the row persisted on every build.

## Fix

In `_ledger_cleared_overlay`, a depth-0 row that is cleared (copied flag or ledger row), whose `done` is derived, and that carries no summary (whitespace counts as none, as the qualification reads it) is dropped, and the descendant rows that follow it in the flat list go with it. The docstring carries the reasoning: a cleared flag rolls up to status cleared (`rollup_status` gives it precedence), so a copied "completed" beside a clear is a stale copy, not a completion; a live-card clear would have rolled the top to cleared before the copy and it would never have qualified. At depth 0 `derived` means the copied status or a summary and never an ancestor, so derived with no summary is status-only. The drop sits behind the overlay's existing early return, so it acts only while the ledger holds a live row; every such root has its own (the re-seal requires it, an undo removes the row and restores the node, nothing prunes the file). No payload field, no archive rewrite, no cache change; the cached rows are read, never mutated. One sentence at the end of `_fleet_archived_tops`'s docstring names the exclusion.

Two design points, stated so the choice is visible:

- A root completed by its own verdict (nodeComplete) that the ledger clears keeps the behaviour #1187 pinned: listed, struck through, subtree rolled down. #1187's body says a cleared completed top leaves the completed projection altogether; this change narrows to roots whose only completion is the copied status, so that test stays as written. One consequence: a clear taken from Show completed on an already archived top (ledger-only, since no live node exists) now has two outcomes by completion source: a verdict-completed root is struck through, a rollup-completed one (an umbrella, a settled store) disappears. Dropping every cleared archived root would match #1187's sentence literally, make the two agree, and change that test; say so if that is the preferred reading.
- A cleared top that carries a distiller takeaway stays listed, struck through, on every path today: `_distill_session` stamps completed tops and skips cleared ones, so a takeaway written before the clear persists, and the top qualifies through its summary whatever its status. That holds for older archives and for a live-card clear after #1234 (whose test's top has no summary). This change leaves that case as it is, so the archive matches the live-card path; the second test pins it. Dropping such tops would change the qualification rule itself, a separate decision.

Known limit: the drop runs after the projection's cap (`topids[:cap]`, cached by archive mtime), so a session whose most recent capped tops are all stale copies shows fewer rows than the cap while older completions stay behind it. Before this change those slots held the struck-through rows, so nothing visible is lost; moving the drop ahead of the cap would put the ledger in the cache key, a separate change.

## Tests

`tests/test_kernel_goal_compaction.py`, `ClearedLedgerIsAuthoritativeAcrossTheCompaction`, two methods appended after the class's last test:

- `test_an_archive_from_before_the_reseal_fix_does_not_list_a_status_only_root_the_ledger_clears`: writes the older archive shape directly (root g10 with the flag on, nodeComplete False and status "completed"; its child g10a; a control root g11 with nodeComplete True and status "completed"), reads the projection with an empty ledger (both roots list; the cache is primed), appends clear rows for both roots, and reads again cached and then with every in-process memo dropped. Asserts only g11 lists and reads cleared; g10 and g10a are gone; the first read's rows still hold g10 (the cached rows are not mutated). Before the change the second read lists g10, g10a and g11, g10 with done True and cleared True: the struck-through row.
- `test_only_a_status_only_root_is_dropped_a_takeaway_or_its_own_verdict_keeps_the_row`: five roots newest first: g12 with the flag off, status "completed", a whitespace summary and a ledger row (the erased-flag shape), with a child; g13 cleared with a summary and no status; g14 with nodeComplete True, nothing clearing it, with a child; g15 with status "completed", no flag and no ledger row, with a child; g16 with the flag on, status "completed" and no ledger row, with a child. Asserts the rows are exactly g13, g14, g14a, g15, g15a; g13 reads cleared; g14, g14a, g15 and g15a do not. Before the change the rows begin with g12 and g12a and end with g16 and g16a. g15 pins the cleared condition itself (an uncleared status-only root stays listed while the ledger holds other rows); g16 pins the copied-flag arm (dropped with no ledger row, and the drop of the last root runs to the end of the list); g12's whitespace summary pins that whitespace is not a takeaway.

Both fail before the change on the parent commit's `kernel/kernel.py` and pass after it. Three mutants of the drop condition (the cleared factor removed, the copied flag ignored, the whitespace strip removed) each fail the second test. The existing pins pass unchanged: `test_the_projection_reads_the_ledger_over_the_copied_flag_after_its_cache` (a nodeComplete root the ledger clears stays listed, cleared), `tests/test_fleet_archived_tops.py::test_status_completed_counts_as_done` and `test_mtime_cached` (both run with an empty ledger and pin the overlay's early return: the cached list object is returned as is).

Full suite on 6a83506e (pytest -q -n 8 tests/): 10566 passed, 115 skipped, 1 failed, 1688 subtests passed. The one failure, `tests/test_kernel_webpush.py::LandingRevealOffers::test_no_offer_for_the_session_already_in_front`, passes alone on the same tree and reads none of the two functions this change touches (the landing-reveal offer of the push notifier; a parallel-ordering interaction inside that module).

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -29609,7 +29609,9 @@ def _fleet_archived_tops(sid, cap=20):
     hierarchy, not just show a flat row (the user 2026-06-29). The Fleet filters depth==0 for the roots and
     merges the rest into its node map. "Completed" top = explicitly nodeComplete, rolled up to status
     'completed', or carries a distiller takeaway (summary). mtime-cached on the archive file (changes only on
-    a sweep), so this is ~free on the feed hot path. Caps the number of TOPS (each keeps its full subtree)."""
+    a sweep), so this is ~free on the feed hot path. Caps the number of TOPS (each keeps its full subtree).
+    A cleared top whose only completion is that copied status leaves the list (_ledger_cleared_overlay drops
+    it: a stale copy of a clear, not a completion)."""
     p = jd.GOALARCHDIR / (sid + ".json")
     try:
         mt = p.stat().st_mtime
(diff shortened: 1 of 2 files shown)


ksarma:

Reviewed inline: the drop in `_ledger_cleared_overlay` reads only the cached row's cleared, derived and summary fields and the ledger set. A row therefore leaves Show completed on a clear row and returns on an undo row. Nothing moves it between builds without new information. At depth 0 the projection runs `_proj` with `ancestor_done` False, so derived means the copied status or a summary, never an ancestor. Derived with a blank summary is the status-only case. Each root is emitted followed by its whole subtree before the next depth-0 row. The subtree drop thus rests on a real invariant of the flat list, the one the existing roll-down already reads. The drop sits behind the early return, the cached list is never mutated, and every ledger writer appends. As a result, the ledger-emptied case cannot arise on its own. The known limit, that the drop runs after the cap, is stated and acceptable: those slots held struck-through rows before.

On design point (a): the narrowed rule gives a clear taken from Show completed on an archived top two outcomes by completion source. The user cannot tell those sources apart. The literal #1187 reading (drop every ledger-cleared archived root without a takeaway) would match the live-card path for every source. I recommend that reading as a follow-up. This is not a hold: the change only removes rows main listed wrongly.

Red-first verified by my own run. With the PR's test file at the base 7b894883, both new tests fail; at the head they pass. The base failures: the stale root listed with its child, and the five-root list beginning with `g12`. The compaction module and the archived-tops sibling file pass at head (22 passed). Three mutants of the drop condition each fail the second test in the review's runs. CI at head: every check green.

Review run by the maintainer automated review session; verified against head 6a83506e.

ksarma:
Head note: the head moved from 6a83506e to 172ca91d by a merge of main into the branch (this session updated the branch to re-run the checks against current main after a long hold); the merge is mechanical (the parents' merge-tree equals the head's tree) and the PR's added and removed lines are identical to the reviewed ones, so the review above stands for this head.

Review run by the maintainer automated review session; verified against head 172ca91d.

PR #1245, opened 2026-09-10
Markdown images: the served-URL probe budget follows the URL, so a turn re-rendered while it streams keeps its three attempts, and the reconnect heal probes every remembered URL

ksarma:
## Symptom

A markdown image of a file an agent is still writing (a `/file` URL the kernel serves) fails while its turn streams, and its caption stays parked after one off-DOM probe instead of the three that #1222 promised for that case (its comment on `servedByKernel`: a file an agent is still writing may come back with no reconnect event, so it gets three bounded attempts on the per-message path). Nothing re-probes it until a reconnect-class event, though the file was finished a moment later.

Two related gaps come with it. A parked image whose turn has left the rendered window, or whose tab was closed, is never probed on a reconnect: its URL stays remembered for the page life, so scrolling back parks it again and the picture never lands. And a per-message probe that fails after a reconnect has already healed its URL re-arms an attempt, so the next push probes the healed URL once more.

## Cause

In #1222 the three-attempt budget is a closure over the img element that failed, registered in `failedPreviews` under that element. A streaming turn is re-rendered on every push: the tail path removes the turn's nodes and rebuilds them, so the element is off the page by the next push, `retryFailedPreviews` drops its entry unrun, and `mdImgPostPass` parks the fresh img with no budget. Trace for a served URL that 404s while its turn streams: error, park, budget 3; push 1 runs the probe (the img is still on the page); the re-render parks a fresh img; the probe fails and re-registers the old img, now detached; push 2 drops the entry. One probe in all.

`healMdImgs` collected URLs from the parked imgs on the page only, so a remembered URL with no img on the page was skipped. And the closure re-registered its img on every failure, whether or not the URL had healed meanwhile.

## Fix

`ui/webview/preview.ts`: the budget is keyed on the URL. `mdImgProbe` maps a served URL to the attempts left and `mdImgProbing` holds the URLs with a probe in flight, beside `mdImgFailed`. The capture listener still parks the img, and for a served URL arms three attempts when the URL has none. `retryFailedPreviews` first drains those budgets (`retryMdImgProbes`): one detached probe per URL per push, the next armed only once the previous has failed and only while the URL is still remembered, whatever img carries the URL now and whether or not one is on the page (the probe is off the DOM; a load re-queries the page, as before). A load clears the URL's budget along with its memory. `healMdImgs` seeds its URL set with `mdImgFailed` and still reads the page's parked imgs, so a reconnect probes the union. No markdown image is registered in `failedPreviews` any more; the figure previews' own registrations are untouched.

One behavior change to note: a reconnect-class event (romp:wsup, hostUp, romp:hostRelayUp) now probes every URL that failed during the page life, including URLs with no parked img on the page (turns evicted from the rendered window, closed tabs). It is one detached probe per distinct URL per event, and the events are rare.

## Tests

`ui/webview/md-img-park.test.ts`, executed over the file's fake DOM. Three cases fail before the change:

- "a served URL's budget follows the URL: a turn rebuilt on every push while it streams still gets three probes, one per push": fails before the change at push 2 with `probes.length` 1, not 2. It also asserts that a push while a probe is pending fires no second probe for the URL, that a failed probe never writes the img on the page, that a second img with the URL failing after push 2 keeps the one attempt left rather than re-arming three (three probes in all, not four), and that the reconnect heal then lands the picture on every img on the page with the URL.
- "the reconnect-class heal probes a remembered URL whose turn is outside the rendered window": fails before the change at its first assertion with `probes.length` 0. It also asserts the next reconnect probes the URL again while it is down, that a scroll-back renders the healed URL normally, and that a healed URL is not probed again.
- "a reconnect heal that lands while a per-message probe is pending stands: the stale failure re-arms nothing for the healed URL": fails before the change with 3 probes for 2 (the closure re-registered its img and the next push probed the healed URL).

One case passes before the change and pins the new bookkeeping: "a per-message probe that loads lands the picture on the img on the page now and clears the URL's budget with its memory; the URL failing again later starts afresh". Dropping either deletion from the load handler turns it red (a budget left behind fires a probe for a healed URL on the next push; an in-flight mark left behind blocks the URL's per-message probes for the page life).

`fresh()` now clears the module's memory through the public reconnect path (`refreshSettledPreviews` with every probe answered by a load) instead of resetting only the fake's arrays, and the header describes that. It is required once `healMdImgs` reads `mdImgFailed`: without it the existing bounded-probe case counts a URL left over from an earlier case (5 probes, not 4), checked by disabling the reset once.

The module: 8 of 8 pass with the change, 5 of 8 before. The source pins on this code are unchanged and pass: preview-heal (6), preview-retry-pace (4), chat-inline-preview (14), preview-core (5), heal-on-hostup (6), pr-links (55). `npm run typecheck` is clean.

Full `node --test` run over the built test bundle on this head: 4043 tests, 4043 pass, 0 fail.

Tier: fix

### DIFF
--- a/ui/webview/md-img-park.test.ts
+++ b/ui/webview/md-img-park.test.ts
@@ -5,8 +5,13 @@
 // stable caption); sixty pushes write the src zero times; the reconnect-class heal probes each parked URL once OFF the
 // DOM, a failed probe changes nothing, a successful one lands the picture on every parked img with that URL on the
 // page at that moment (a re-render during the probe included); a URL the kernel serves gets a bounded off-DOM probe on
-// the per-message path; and the chat's post-pass parks a re-rendered img with a remembered URL before the browser
-// fetches it. Every test builds its own state.
+// the per-message path, and that budget rides the URL, so a turn re-rendered on every push while it streams still gets
+// its three probes, a second img with the URL failing midway starts nothing over, a load clears the budget with the
+// memory (the URL failing again later starts afresh), and a heal that lands while a probe is pending is not undone by
+// that probe's failure; the reconnect-class heal probes every remembered URL, on the page or not (a turn evicted from
+// the rendered window, a closed tab); and the chat's post-pass parks a re-rendered img with a remembered URL before the
+// browser fetches it. The module remembers failed URLs for the page life, so fresh() heals every remembered URL off
+// the DOM through the public reconnect path before each test: every test starts with an empty memory.
 import { test } from "node:test";
 import * as assert from "node:assert/strict";
 import * as fs from "node:fs";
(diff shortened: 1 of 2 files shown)


ksarma:
Merging as shipped. One follow-up to file, nothing to fold into this PR.

Reviewed inline: the served-URL probe budget now lives in `mdImgProbe`, keyed on the URL, and `mdImgProbing` marks a probe in flight. A streaming turn's re-render therefore no longer discards the closure #1222 kept on the img element. The capture listener arms three attempts only when the URL has none. `retryMdImgProbes` fires one detached probe per URL per kernel push. It arms the next only after the previous fails and while the URL is still remembered. The load handler clears all three sets for the URL and is the only path that writes a page img's src. Every probe fires on a kernel push or a reconnect-class event (`romp:wsup`, `hostUp`, `romp:hostRelayUp`); no timer was added. `healMdImgs` seeds from `mdImgFailed` plus the page's parked imgs, so a remembered URL with no img on the page is probed on reconnect. Its load clears the memory, so a later scroll-back renders normally. A failing heal probe re-arms nothing. Two accepted limits remain, neither new. A per-message probe that never settles keeps its URL in `mdImgProbing`, which stops only that URL's per-message path until a reconnect heal loads it. `mdImgFailed` grows one short string per distinct failing URL for the page life, with no cap, as before. `fresh()` empties all three sets through one heal, so the eight cases share no state.

- `ui/webview/preview.ts:833`, follow-up: the `onFail` closure in `retryMdImgProbes` applies the `left` captured at fire time. It never checks whether a load has since retired the probe. Suppose a reconnect heal loads the URL while a per-message probe is pending, and the same URL fails again before the stale error arrives. The stale failure then deletes or shortens the fresh budget (reproduced: 4 probes not 5 with left=1, 2 not 3 with left=3). It can also clear a newer probe's in-flight mark. The loss is bounded to per-message attempts, and the reconnect heal still probes the URL. A per-probe token in `mdImgProbing`, compared in the callback, fixes it. A bare `if (!mdImgProbing.delete(u)) return;` misses the case where a fresh probe is already in flight.

Red-first is met locally. The PR's `md-img-park.test.ts` against the base `preview.ts` fails 3 of 8 (budget follows the URL: 1 probe not 2 at push 2; heal while pending: 3 not 2; off-page URL remembered: 0 not 1) and passes 8 of 8 at the head. Dropping either deletion in the load handler fails the bookkeeping case. `md-img-park` plus `preview-heal`, `preview-retry-pace`, `chat-inline-preview`, `preview-core`, `heal-on-hostup` and `pr-links`: 98 of 98; typecheck clean. CI at the head: all eleven checks green.

Review run by the maintainer automated review session; verified against head 0e9079e9.

PR #1246, opened 2026-09-10
Auto-reload: the refusal on a tab whose create failed and the queued edit's send refusal do not come back either

ksarma:
Follow-up to #1227 (merged): its review named one refusal in `sendComposer` that still rode the reload, and a second sits beside it.

## Symptom

A send into a tab whose create failed is refused with a toast saying the session never started, and the words stay in the box. When the reload after a kernel restart takes the page inside the toast's twelve seconds, the fresh page shows the toast again, naming a tab it does not have.

A queued edit sent while the session cannot be reached (its host is down, or its tab is still being created) is refused with a toast saying the edit was not sent and to send again when the link is back. Shown again by the fresh page, it asks the user to send again an edit that page no longer has, and a send there would post the words as a new message behind the unchanged queued one.

## Cause

Both refusals were plain `warnToast`s, so `persistNoticesForReload` kept them and the fresh page replayed them. Both report a state rather than an event: the provisional tab and the queued edit (`provisionalId`, `failedProvisionals`, `queuedEdits`) live in memory alone, so the fresh page has neither, and the replay is false.

The review named the never-started refusal. The queued edit's refusal sits beside it in `sendComposer` and reads as a state too. The plain send's refusal on a disconnected host stays unmarked because the message it speaks of is in the box on the fresh page too, and a send there sends it. The queued edit's "send again" cannot do what it says on a page with no edit in progress. The `isProvisionalId` half of its guard is reachable through the pencil on the dashed bubble of a tab still being created, so the same toast is raised on a provisional tab too.

## Fix

`ui/webview/render.ts`: the never-started refusal in `sendComposer`'s `deliver` and the queued edit's refusal ahead of it are raised through `ephemeralWarnToast`, each with a comment at the site; the census above `ephemeralWarnToast` lists them with the other state refusals and names the plain send's disconnected-host refusal as the unmarked one. Nothing else changes; the `stageComposer` refusals #1227 marked are untouched.

## Tests

`ui/webview/reload-notices.test.ts` lifts the provisional branch of the send's `deliver` and the send's queued-edit branch out of `render.ts` (between stable anchors, the file's pattern) and executes them over the fake DOM. Three cases join the state-refusals table:

- a send into a tab whose create failed: the toast is on screen, nothing is queued, the draft stays, `liveNotices` skips it;
- a queued edit sent while the session's host is unreachable: the toast is on screen, a re-dial is posted and no edit, the edit is still in progress, the draft stays, `liveNotices` skips it;
- a queued edit sent on a tab still being created: the same, with nothing posted.

Before the change each fails at the reading: `liveNotices` returns the refusal's text (18 tests, 14 pass, 4 fail). The source-pin test matches both sites through `ephemeralWarnToast`, pins the disconnected-host refusal as a plain `warnToast`, and counts ten `ephemeralWarnToast` sites (eight before this change, so it fails on the old count). `ui/webview/provisional.test.ts`: its pin of the never-started site follows the swap.

Full `npm test` on this head: 4055 tests, 0 failing (4033 in the main leg and 22 in ui/timeline-tags-scale.test.ts run alone as a second leg; both legs exit 0). `npm run typecheck` clean.

Tier: fix

### DIFF
--- a/ui/webview/provisional.test.ts
+++ b/ui/webview/provisional.test.ts
@@ -110,8 +110,9 @@ test("creating a session opens the provisional tab instead of a modal", () => {
 test("a send on a provisional tab is HELD, not posted to a session that doesn't exist", () => {
   assert.match(RENDER, /provisionalQueue\.push\(text\);\s*\n\s*registerOptimistic\(sid, text, attached\.filter\(\(p\) => previewKind\(p\) === "img"\)\);/,
     "the dashed bubble goes up now — with its dragged-image thumbnails — romp has it, it is not delivered");
-  // a FAILED tab has no pending spawn to queue onto: refuse loudly, the box keeps the only copy
-  assert.match(RENDER, /if \(sid !== provisionalId\) \{\s*\n\s*warnToast\("“" \+ \(sessions\.get\(sid\)\?\.name \|\| "this session"\)/);
+  // a FAILED tab has no pending spawn to queue onto: refuse loudly, the box keeps the only copy. The refusal reports
+  // a state the page after a reload does not have, so it is ephemeral (executed in reload-notices.test.ts)
+  assert.match(RENDER, /if \(sid !== provisionalId\) \{\s*\n\s*ephemeralWarnToast\("“" \+ \(sessions\.get\(sid\)\?\.name \|\| "this session"\)/);
 });
 
 test("adoption flushes the held messages FOR REAL and carries the draft across", () => {
--- a/ui/webview/reload-notices.test.ts
+++ b/ui/webview/reload-notices.test.ts
@@ -5,15 +5,17 @@
 // render.ts keeps the texts of the toasts on screen in this tab's sessionStorage on the core's synchronous hook and the
 // fresh page shows them once. The readings are pure and execute here. render.ts has import-time DOM side effects, so
 // its wiring is pinned to source the way reload-restore.test.ts pins the scroll record's, and the toast family with the
-// refusals that report a state is lifted out of it and executed over a fake DOM the way chat-exact-tail-exec.test.ts
-// lifts chatTail. The served scenario (a nack on the last ship across a kernel restart; the fresh page says it again,
-// once) is tests/test_ship_reship.py NackNoticeSurvivesReload. Synthetic only.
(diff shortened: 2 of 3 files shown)


ksarma:
Merging once the Python 3.13 leg reruns green. Nothing to fold.

Reviewed inline: the two swapped sites in `sendComposer`, the never-started refusal in `deliver`'s provisional branch and the queued edit's reachability refusal, now go through `ephemeralWarnToast`. So `persistNoticesForReload` leaves them out of the replay after a kernel-restart reload. The four staging marks from #1227 and every event toast are unchanged. The census comment above `ephemeralWarnToast` matches the code: ten `ephemeralWarnToast` occurrences including the definition, eight before. Every plain `warnToast` site reports an event or a plain message. The disconnected-host plain send is among them, named as the deliberately unmarked one. The queued-edit mark is right. The edit lives in `queuedEdits`, memory only, so the fresh page has no edit in progress. A replayed send-again would post the words as a new message behind the unchanged queued one. That makes the toast misleading advice rather than a stale but true report. The slash-command refusal two lines below it correctly stays plain. The queued message and its ✕ survive the reload, and typing the command does what the toast says. The lifted closures are the real `render.ts` text between anchors that occur once each in the file. The prelude's fake `vscodeApi` captures the redial the host-down case asserts. The fix label fits: a follow-up defect in #1217's mechanism, tests red at main and green at head.

Follow-ups, no hold:
- `ui/webview/render.ts:9966`: the census names the slash-command refusal in neither list. One clause would make the boundary explicit.
- `ui/webview/reload-notices.test.ts:261`: the host-down case runs on a bare id, so the redial's host is a truncated slice. A host-prefixed synthetic id would let the row pin the host.

Tests: at main 7b894883 with the PR's two test files, 33 tests, 28 pass, 5 fail. The failures are the three new state-refusal cases at the `liveNotices` reading and the two source pins. At head 3d9ffaf8, 33 of 33 pass and typecheck is clean. CI: every leg is green except Python 3.13, which failed on `tests/test_cut_turn_tree_kill.py`. That real-process test is untouched by this webview-only PR and passed on 3.10 through 3.12. Rerunning that job.

Review run by the maintainer automated review session; verified against head 3d9ffaf8.

PR #1247, opened 2026-09-10
Tests: the settings-dir save's home fallback has its own child case

ksarma:
Builds on #1237 (the settings dir the run was handed, saved ahead of the CLAUDE_CONFIG_DIR floor for the live session-move test); this branch is stacked on it, and the change to review is the last commit.

## Symptom

No user-visible symptom today. The gap is that a later conftest edit that floored HOME ahead of the save would go unnoticed: the saved location would become a run-private directory, the live session-move test would skip for want of auth on every machine that never exported CLAUDE_CONFIG_DIR, and the module's cases would stay green.

## Cause

All three child-pytest cases in tests/test_claude_config_floor.py export CLAUDE_CONFIG_DIR: one checks that the save comes before the floor, two that a saved value the shell already carries survives the import (one of them under -n 2). The save's other branch, .claude under HOME when the shell exports no root (the developer's and CI's case), had no case. The review of #1237 asked for one.

## Fix

One more case in the same idiom, test_the_saved_location_is_the_home_settings_dir_when_the_shell_exports_no_root: the child pytest's shell has CLAUDE_CONFIG_DIR cleared and HOME set to a synthetic temp dir, and the generated test passes only if ROMP_TESTS_REAL_CLAUDE_CONFIG_DIR is that home's .claude and CLAUDE_CONFIG_DIR is neither that value nor the home. The generated test gains the assertion that CLAUDE_CONFIG_DIR is not the saved value, for every case: the exported-root case already had the saved value among the exported values it checks, and without the assertion the new case could not tell a floor from no floor (a conftest that left CLAUDE_CONFIG_DIR at ~/.claude under the synthetic HOME would have passed it).

The helper _saved_location_seen_by_a_child treats a None value as "clear this variable from the child's shell" and leaves it out of the values the generated test checks the floor against. The module docstring and the generated case's comment and name mention the new case and the added assertion. Tests only; tests/conftest.py is unchanged.

## Tests

- tests/test_claude_config_floor.py::ClaudeConfigFloor::test_the_saved_location_is_the_home_settings_dir_when_the_shell_exports_no_root: a guard, green on the current conftest. Its red was shown by mutation in a scratch copy of the tree, each mutation applied to tests/conftest.py alone:
  - a HOME floor (os.environ["HOME"] = tempfile.mkdtemp(...)) inserted ahead of the save: the case fails (the saved value is the floored home's .claude, not the synthetic home's) while the existing exported-root case passes;
  - the save moved below the CLAUDE_CONFIG_DIR floor: both cases fail (the saved value is the floor);
  - the floor set to the saved value, or to ~/.claude, or left unset when the shell exports no root: the case fails on the added assertion that CLAUDE_CONFIG_DIR is not the saved value, and stays green with that assertion removed.
- The module: 9 passed (8 before).
- The meta-tests that scan test sources (tests/test_judge_parse_cache.py, tests/test_postal_marker_form.py, tests/test_state_isolation_order.py): 7 passed.
- Full suite on this tree (`pytest -n 6 tests/`): 10571 passed, 115 skipped, 0 failed, 1688 subtests passed.

Tier: docs

### DIFF
--- a/tests/test_claude_config_floor.py
+++ b/tests/test_claude_config_floor.py
@@ -13,8 +13,9 @@ opt-in test that borrows the operator's apiKeyHelper command from their own sett
 
 Source pins in the style of test_supervised_floor.py, the floor observed from inside a test, from a
 module loaded under it, and from a subprocess whose shell exports a value of its own; the saved
-location observed from inside a test, from a subprocess whose shell exports the root it saves, and
-from one whose shell already carries a saved value."""
+location observed from inside a test, from a subprocess whose shell exports the root it saves, from
+one whose shell exports no root and is handed the settings dir under a synthetic HOME, and from one
+whose shell already carries a saved value."""
 import importlib.util
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


ksarma:

Reviewed inline: the docs tier fits. The diff is one test module, `tests/test_claude_config_floor.py`, and `tests/conftest.py` is byte-identical to main. The new case runs the child pytest with `CLAUDE_CONFIG_DIR` removed from its shell and `HOME` set to a synthetic temp dir. The helper pops a None-valued variable from the child's env and leaves it out of the exported tuple. The generated test therefore checks the floor against the home alone. The child's env is this process's environment with `ROMP_TESTS_REAL_CLAUDE_CONFIG_DIR` and the pytest and xdist variables popped before the shell is applied. Conftest's expanduser fallback thus resolves under the synthetic home, never the real one. The added assertion that `CLAUDE_CONFIG_DIR` differs from the saved value is redundant for the three existing cases and required by the new one. Without it, a conftest that floored to the saved value passes. Body and commit match the diff and quote nobody. The added lines carry synthetic paths only.

- PR body, Tests section, follow-up (wording, optional): the third mutation clause holds for a floor set to the saved value, not for one left unset. It says a floor left unset fails the case on the added assertion and not without it. For a literal unset, the next line's read of `CLAUDE_CONFIG_DIR` raises `KeyError` as well, so that variant is caught either way.

Locally: the module at head, 9 passed. At the merge base with the PR's test file, 9 passed (a guard, green by design). A scratch conftest with `HOME` floored ahead of the save fails only the new case (1 failed, 8 passed). The save moved below the floor fails it and the three other saved-location tests. The floor set to the saved value fails on the added assertion alone and passes once it is removed. The source-scanning meta-tests, 7 passed. CI at head: every required check green, both tier checks included.

Review run by the maintainer automated review session; verified against head d911bb9a.

ksarma:
Head note: the branch was updated with main (3bebe6ba), so the head moved from d911bb9a to f39639c4. The new head is a mechanical merge of d911bb9a and 3bebe6ba: its tree equals the merge-tree of its two parents, and its added and removed lines against main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head f39639c4.

PR #1248, opened 2026-09-10
Usage modal: a hover crosshair over the spend chart, and the middle range is 7 days over 168 hourly buckets (T293)

henrypinkard:
Two changes to the usage modal's spend chart (T293).

**A hover crosshair.** Moving the pointer over the chart draws a hairline at the pointer's bucket and a stamp beside it naming the bucket in words: "Wed Sep 9, 3 PM" for an hourly bucket, "Wed Sep 9" for a daily one. The tooltip lists that bucket's sessions in spend order, each with a dot in its stack's colour and its value, the top six and then one "+N more" line. The bucket is the view's own: an hour in the 1-day and 7-day views, a day in the 90-day view.

**The middle range is "7 days · by hour"** over the hourly ledger's last 168 buckets. The ledger still keeps 192 hours, a day of slack past the view. The persisted preference value is unchanged, so a pick made under the old "8 days" label reads as the 7-day view with no rewrite.

Design points:
- **Over a bar, the same tooltip.** The one-segment tip is gone; a bar's hover shows the bucket box with that bar's row emphasised (and always listed, even past the fold). One shape, so nothing flips as the pointer crosses a bar's edge, and the bar's own reading is still there on its row. The header carries the bucket's total in the active measure, the other measure, and the stamp.
- **Pointer-inert, no layout change.** The hairline is an svg line and the stamp an absolutely placed element inside the chart box; the tooltip is the existing fixed box. All three take no pointer events, none is in the flow, so nothing moves under the pointer. They leave on pointerleave. Keyboard and touch gain nothing and lose nothing: only pointermove and pointerleave are read.
- **The stamp is read off the bucket KEY**, the recorder's local time, like every label on this chart; the existing zone note says when the viewer's clock differs. The tooltip sits below and to the right of the pointer (to its left near the right edge); when nothing fits below it goes above the CHART rather than the pointer, so it stays clear of the stamp at the chart's top, and only a window too short for either falls back to the pointer (the header repeats the stamp's text). The stamp flips to the line's left when it would run past the chart's right edge, decided in pixels, and a 24px floor keeps it off the y-axis ceiling label over the first buckets.
- **Pure functions, executed.** The bucket for a pointer offset, the stamp per view, and the top sessions of a bucket are plain functions in the landing script; `ui/webview/spend-crosshair.test.ts` runs them from the source, along with the range table, `spSeries` on a 192-key synthetic series, and `spLoadPrefs` against a stored pick from the old button.
- **Sizes and colours already on the surface.** The stamp wears the modal's 10px annotation size; the tooltip rows inherit the tip's 11px; the hairline and stamp have light-theme steps beside the tip's. An unattributed stack's dot is the hatch's stroke colour, "other" its neutral.

Tests: the node test above; `tests/test_spend_detail.py` pins the wiring and the rename; the headless served-page test (`tests/spend_modal_headless.js`, `tests/test_spend_modal_headless.py`) hovers the real chart in all three views and reads the line, the stamp, the rows, the fold, the pointer-inertness, the unchanged chart height, and that all three leave with the pointer; it also pins seven local midnights on the 7-day axis (192 hours carried eight).

Screenshots (served page, dark and light, 1-day and 7-day views): taken with the headless harness under `ROMP_SHOTS`.

**Review** (lean, 14 agents: four lenses over the branch diff, two skeptics per finding). Four confirmed, all fixed in the second commit: the short-window flip covered the stamp (now above the chart, pinned in a 600px window); the served order pin compared two rows that both rendered "$1" and could not fail (now the daily probe's token values, which differ); the stamp covered the ceiling label over the first buckets (a 24px floor, pinned at the first bucket); and a duplicate of the first. One plausible, a fixed 70% flip threshold on narrow charts, which one skeptic could not reproduce; fixed anyway by deciding the flip in pixels. Of eleven low items left unverified by the cap, the real cheap ones are folded: a rebuild hides the tooltip (pinned by a resize probe), the x-label pin counts labels rather than characters, two stale assertion messages and a stale test comment reworded, the stamp's CSS pin anchored to its rule, the seven-midnights pin's runner-zone assumption documented.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -44731,21 +44731,25 @@ rows.sort(function(a,b){return (b.v-a.v)||(a.si-b.si);});var top=rows.slice(0,ca
 if(pin>=0&&more>0&&!top.some(function(r){return r.si===pin;})){var p=null;rows.forEach(function(r){if(r.si===pin)p=r;});if(p){top.push(p);more--;}}return {rows:top,more:more,total:total};}
 function spDot(s){return s.kind==='unattributed'?'#8a97a6':s.kind==='other'?SP_OTHER:spColor(s);}   // the hatch's stroke stands in for its texture
 function spTipBox(){if(!spTip){spTip=document.createElement('div');spTip.id='rsp-tip';document.body.appendChild(spTip);}spTip.textContent='';return spTip;}
-// below and to the right of the pointer (the stamp sits at the chart's top, so the two never meet), to its left near
-// the right edge and above it near the bottom, kept inside the viewport; the box takes no pointer events
-function spTipPlace(x,y){spTip.style.display='block';var w=spTip.offsetWidth,hh=spTip.offsetHeight,left=x+14,top=y+18;if(left+w+6>window.innerWidth)left=x-14-w;if(top+hh+6>window.innerHeight)top=y-18-hh;
+// below and to the right of the pointer, to its left near the right edge; when nothing fits below, above the CHART
+// (rect, the svg's box) rather than the pointer, so the box stays clear of the stamp at the chart's top (review find:
+// a pointer-anchored flip landed a six-row box on the stamp in a short window); only a viewport too short for either
+// falls back to the pointer, and the header repeats the stamp's text. The box takes no pointer events
+function spTipPlace(x,y,rect){spTip.style.display='block';var w=spTip.offsetWidth,hh=spTip.offsetHeight,left=x+14,top=y+18;if(left+w+6>window.innerWidth)left=x-14-w;
+if(top+hh+6>window.innerHeight){var above=(rect?rect.top:y)-hh-6;top=above>=6?above:y-18-hh;}
 spTip.style.left=Math.max(6,left)+'px';spTip.style.top=Math.max(6,top)+'px';}
 // the bucket tooltip: the total leads (the other measure and the stamp follow), then a row per session — a dot in the
 // stack's colour, the name (user data: textContent), the value — the hovered bar's row emphasised, the fold line last
-function spBucketTipShow(x,y,head,total,top,stacks,many,meas,i,pin,fmt){var box=spTipBox(),h=document.createElement('div');h.className='rsp-tip-h';
+function spBucketTipShow(x,y,head,total,top,stacks,many,meas,i,pin,fmt,rect){var box=spTipBox(),h=document.createElement('div');h.className='rsp-tip-h';
 var b=document.createElement('b');b.textContent=total;h.appendChild(b);h.appendChild(document.createTextNode(' \u00b7 '+head));box.appendChild(h);
 top.rows.forEach(function(r){var s=stacks[r.si],row=document.createElement('div');row.className='rsp-tip-row'+(r.si===pin?' on':'');
 var dot=document.createElement('i');dot.style.background=spDot(s);row.appendChild(dot);var nm=document.createElement('span');nm.textContent=spStackText(s,many,meas,i);row.appendChild(nm);
 var v=document.createElement('b');v.textContent=fmt(r.v);row.appendChild(v);box.appendChild(row);});
 if(top.more>0||!top.rows.length){var m=document.createElement('div');m.className='rsp-tip-more';m.textContent=top.rows.length?'+'+top.more+' more':'nothing recorded';box.appendChild(m);}
-spTipPlace(x,y);}
+spTipPlace(x,y,rect);}
 function spTipHide(){if(spTip)spTip.style.display='none';}
-function renderChart(){var box=document.getElementById('rsp-chart');if(!box||!SP.data)return;
+// a rebuild (a resize under a still pointer, a toggle) takes the tooltip down with the svg it described (review find)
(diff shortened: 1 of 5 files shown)


ksarma:
Recommend merging as shipped. One follow-up to fold; nothing here holds an admin's feature.

Reviewed inline: the pointermove handler writes only the hairline, the stamp and the fixed tooltip. Only renderChart rebuilds the chart's innerHTML, on open, a range or measure change, or the resize frame. So bars survive a pressed pointer, and the hover path has no timer, debounce or delaying rAF. spBucketAt returns -1 for a negative offset, an offset at or past the width, a zero width, NaN or an empty series, and clamps to n-1. An empty or all-zero series leaves renderChart before any handler is attached. The range change is a view-only slice. SP_RANGE_BUCKETS reads 168 and _SERIES_HOURS stays 192; the rail's hover graph already windowed 168. The range id, the prefs key and the /spend/detail payload are unchanged. spLoadPrefs leaves an old stored pick on the 7-day view with no rewrite. The only 8-day wording left is in ledger comments that remain true. The stamp is formatted by hand from the bucket key, so it stays in the recorder's zone like the axis; the zone note is unchanged. spTipPlace flips against window.innerWidth and innerHeight, which is correct for a fixed box. Colours follow the tip's own literal-plus-light-step convention. The stamp uses the 10px annotation size, rows inherit the tip's 11px, and a bucket lists six one-line rows plus one "+N more". The feature label fits: a capability inside the existing modal with no contract change. Privacy and vocabulary scans of the added lines are clean. The body and commit message match the diff.

Follow-up:
- `kernel/kernel.py:44798`: renderChart replaces the chart's innerHTML and creates the new hairline and stamp hidden, but never calls spTipHide. `#rsp-tip` lives on document.body. Rest the pointer on the chart, then resize the window (a snap, F12) or press Enter on a focused range button. The svg is rebuilt with no pointerleave, so the tooltip stays up naming the old bucket while line and stamp are gone until the pointer moves. Reproduced in headless chromium against the PR's own served fixture. One line fixes it: spTipHide() at the top of renderChart. The resize half predates this PR; the two new marks make it visible.

Two nits, no action needed:
- The flip at pct>70 is a fixed fraction while the stamp's width is fixed in pixels. An un-flipped hourly stamp (about 101px at 10px) overflows a chart narrower than about 360px, by about 11px at the 320px floor (`kernel/kernel.py:44811`).
- In the 1-day view the axis prints 06:00 and 18:00 while the stamp above the same bucket prints 6 PM (`kernel/kernel.py:44725` and 44787).

Locally at the head: `tests/test_spend_detail.py` 23 passed; `tests/test_spend_modal_headless.py` 1 passed (chromium, ROMP_SHOTS unset); typecheck clean; spend-crosshair.test 5/5. That test is a real test: Math.floor to Math.ceil in spBucketAt fails only case 1, 168 to 192 fails only cases 4 and 5, and the base kernel.py fails all five. CI at 4a56b9e8: all eleven checks green, nothing pending.

Review run by the maintainer automated review session; verified against head 4a56b9e8.

ksarma:
Delta record: nothing is left to fold. The delta from the reviewed head 4a56b9e8 to the merged head fa75409a is one author commit plus two mechanical merges.

Reviewed inline: d060ab57 is the whole substantive delta. ccd45306 and fa75409a each equal the merge-tree of their parents. The commit answers the three review items. `renderChart` now calls `spTipHide()` as its first statement, so every rebuild hides the tooltip: the resize frame, the range, measure and order chips, the merge toggle and the open path. With that call removed in a scratch tree, the served test's new afterResize pin reproduces the stale tooltip and fails. The stamp flips by pixels, with a 24px left floor. The pixel test is the line x in svg pixels, plus the 6px translate and the measured offsetWidth, against the svg width. It therefore stays inside the chart at both edges for any chart wider than about 214px and clears the fixture's ceiling label. `spTipPlace` keeps the viewport test for the fixed box. When nothing fits below, it anchors above the svg's rect with a 6px guard. It falls back to the pointer only when neither fits, as the body says. The served pin measures the box bottom above the stamp top at a 600px window. The old order pin compared two hourly rows that both render as $1 and could not fail. The new one reads the daily token rows, asserts descending order and that the extremes differ, and fails under an ascending sort. Nothing in the commit falls outside its message. The 56 added lines carry no private strings, hostnames or home paths.

Follow-ups, none blocking:
- `tests/test_spend_detail.py:647`: the three behaviour pins live only in the served playwright test, which CI skips. Its docstring says so; the Python job installs pytest alone. A string pin on the `spTipHide()` call at the top of `renderChart` would give CI a guard.
- `kernel/kernel.py:45072`: the 24px floor clears ceiling labels of up to five glyphs. A daily view in dollars whose ceiling is $10000 puts the label's last digit about 3.5px under the stamp at the first buckets.
- The 1-day axis still prints 06:00 and 18:00 beside a stamp that reads 6 AM and 6 PM. This delta does not address it.

Locally at fa75409a: typecheck clean, spend-crosshair 5 passed, test_spend_detail 23 passed, test_spend_modal_headless 1 passed (the served class, alone). CI at fa75409a: all nine checks green. Merged as b649dc26.
Review run by the maintainer automated review session; verified against head fa75409a.

PR #1250, opened 2026-09-10
The reply-ready chip opens the thread it lands on, which is what marks the reply read

yatharth:
Tier: fix

Follow-up to #1119 and companion to #1249. The reply-ready chip scrolled to the nearest unread mark, pulsed it and stopped, deliberately posting no `commentSeen`: reading the reply meant a second click on the mark. That second click is the one #1249 fixes (nested marks on a twice-commented passage opened the wrong thread), and it is a click the chip can spare the user: the reply lives in the thread's popover, so opening it is the natural completion of the jump.

**Change**: `replyjump` now opens the thread by tid through `openCommentPopover` after the scroll and the pulse, so `commentSeen` posts, the ring drops and the chip's count falls by the thread it opened. The open does not hang on the landing (an anchor the route could not scroll to still shows its reply). The unread bit keeps its ONE clearer; the chip never writes it directly.

**Tests (fail first)**: `reply-ready.test.ts`'s click pin now requires the open (with the why) and its place after the scroll block; `comment-mark.test.ts`'s one-clearer pin requires the chip to reach the bit only through `openCommentPopover`.

`npm run typecheck && npm test && npm run build` green locally apart from the same pre-existing playwright-only KaTeX layout test noted on #1249 (fails identically on pristine `origin/main`; CI skips it).

### DIFF
--- a/ui/webview/comment-mark.test.ts
+++ b/ui/webview/comment-mark.test.ts
@@ -47,10 +47,12 @@ test("unread is the kernel's bit on an open thread, cleared by opening the threa
   assert.match(RENDER, /m\.classList\.toggle\("unread", !!th\.unread && th\.status === "open"\);/);
   assert.match(RENDER, /if \(th\) th\.unread = false;\s*\/\/ optimistic; the kernel's watermark reconciles/);
   assert.match(RENDER, /vscodeApi\?\.postMessage\(\{ type: "commentSeen", id: sid, tid \}\);/);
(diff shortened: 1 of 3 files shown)


yatharth:
Closing at the user's request: the reply chip should scroll to the mark and flash it, not open the thread on its own. Opening stays the user's click, which #1249 now aims at the thread that owns the ring.

PR #1251, opened 2026-09-10
Chat build: memoize the fixed costs a rebuild repeats, and name the tab and the input behind each rebuild in /perf

ksarma:
## Symptom

`romp perf` reports chat rebuilds as one number. On a kernel with many tabs open it cannot say whether the watched tab or the background tabs are rebuilding, or which input moved, so a rebuild storm has no attribution.

Each background rebuild also costs more than its trigger warrants. A judge pass that changed any store rebuilds every background tab (the judge generation is a component of the tab's signature), and that rebuild re-derives work whose own inputs did not move: the live merge's transcript-side sets over every atom of the parse (and again for the feed and the timeline builds of the same cycle), every sealed postal card re-hydrated although no caption moved, the ledger's goal-tree walk over the same parse and store, and the task fold over every turn's atoms. On a quiet board those re-derivations are most of what a background rebuild does.

## Cause

Four per-build costs had no memo although each is a pure function of inputs that rarely move between builds, the caption map's key missed two of its inputs, and the chat counter had no split:

- `_merge_live_atoms` derived `tx_uuids`, `tx_text_uuids`, `tx_texts`, `tx_text_t` and the human floor from every atom, per call; the chat, feed and timeline builds of one cycle each called it on the same parse object.
- The chat fold's postal gate keyed the sealed cards on `_judge_gen`, so any judge pass re-hydrated the tab's whole sealed raw list. A caption is the only judge-written value a card embeds; the rest comes from the raw event and the message log. The commit then hydrated the whole raw list again.
- `_msg_summaries` keyed each session's caption submap on the transcript mtime alone: a caption written after a segment's last transcript record did not reach the map until the session's next turn, and a seam change never did.
- The ledger's goal-tree walk ran on every build over the same parse, store, cleared set and anchor table.
- `_fold_tasks` scanned every turn twice per build, although a build over the parse cache's object changes at most the last turn (the live merge replaces its atoms list).
- `_PerfStats.build("chat", ...)` filed `cached`, `built`, `ms` only.

## Fix

Two commits, each one mechanism with its tests; the second can be dropped on its own. About 2,450 changed lines against the current main (+2,189, -267, `git diff --numstat`), 62% of them tests.

Commit 1, the memos and the counters:

- `_PerfStats.build_chat(cached, dt, active, miss)` at the pusher's serve site. `builds.chat` gains `active_built`, `bg_built` and `bg_miss`, a map from each labelled component of `_chat_build_sig` (`transcript`, `states`, `judge_gen`, `tasks`, `cut`, `row`, plus `cold` and `nosig`) to the background rebuilds it caused. `_chat_sig_miss` derives the labels from the tuple's shape, so the signature stays the flat tuple its callers compare; the push's row is one component either way (`None` without a row), so the tail has one shape. `romp perf` prints the split and the non-zero causes after the chat average. The watched tab's rebuilds are counted, not attributed.
- `_merge_tx_sets(session, sid)`: the five transcript-side values per sid, keyed on the parsed object's identity (the entry holds the object, so an id cannot be recycled onto a new parse), the three sets frozen; a withheld uuid builds a new set instead of subtracting in place. One entry serves the chat, feed and timeline builds of a cycle. Bounded at 512, oldest first, one at a time.
- The postal gate keys the sealed cards on the values they embed: the log's identity (`_chat_postal_key`) and, per card, its caption and its peer's name and colour (`_postal_card_deps`), read from one postal index and one caption map per build, which `build_session` hands to its hydrations as a getter (`_hydrate_postal(..., captions=)`). The commit reuses the sealed cards and hydrates only the raw events new since the seal; `_hydrate_postal` is per-event independent, which the tests pin. A build outside the pusher's names snapshot (a connect push on a handler thread) records its values as unverified, and the next scoped build re-hydrates once and records what it saw. `_msg_summaries` is re-keyed per session on the parse's key (transcript stat, pending cut, states file), the captions file and the goal store's identity (`jd._store_identity`), and read once per pusher cycle through a `_live_scope.msgsum` slot (`_msg_summaries_scoped`) that `_pusher_cycle` opens and its `finally` closes; the chat builds and the timeline's postal connectors (`_postal_messages`) read the same slot. The key's stats (the transcript, the states file, the captions file, the store's three files) therefore run once per cycle, and only in a cycle where some tab's sealed card carries a message id or the timeline rebuilds, since the map is fetched on first need. For a session the SDK backend does not hold live, the pending-cut component reads that session's registry file (the cut lives there): one small JSON read per dormant session per fetch, the same read `_parse` makes for a session it builds.
- The ledger memo: `build_session` keeps the goal-tree walk and the live roots per sid, keyed on the parsed object and the shared store object (both by identity; the shared store is frozen and replaced on any file change), the seams, `cleared.jsonl`'s stat taken before the read, and a per-sid warm-anchor revision (`_node_anchor_rev`, bumped after `_node_anchor_uuids` changes a table entry, read before the walk, so a resolve landing during the walk misses the next build rather than serving the pre-resolve tree). The seams stay a component beside the store's identity because the build loads the store twice, once at the top for the seg ids and once at the ledger for the nodes: a publish landing between the two loads pairs the old seams' seg maps with the new store object, and the next build, whose maps come from the new seams, has to miss. Bypassed for a live-merged parse, an armed rewind hold and an empty store. `_node_anchor_uuids` takes `sid=None`; the feed's three-argument call and its webview pin are unchanged.
- The pusher evicts the memos on the same keep set as the fold prefixes (the shown tabs plus this cycle's comment threads); the merge sets also stay for any alive session, since the feed and timeline merge every alive session. The counters bump under `_chat_fold_lock` (`_chat_memo_bump`).
- Counters under `memos.chatMergeSets`, `memos.chatPostal`, `memos.chatLedger`; `docs/reference.md` describes them.

Commit 2, the task fold: `_fold_tasks(session, sid=None)` splits into `_fold_tasks_turn(atoms)` (pure over one turn: results, rejected ids, ops) and a combine. The per-turn partials are kept per sid on the atoms list's identity plus `_chat_turn_fp`, the fingerprint the chat fold already trusts for a turn. A parse of any kind mints new turn dicts and atoms lists, so the memo serves builds over the parse cache's object and over the live merge's copy of it (which replaces the last turn's list only); the fingerprint guards against an in-place change to a held list. The combine runs in full and returns a fresh list. A call with no sid scans and memoizes nothing, so every direct caller keeps its meaning; `build_session` passes its sid. Counted per turn under `memos.chatFoldTasks`. The seven one-argument `_fold_tasks` stubs in `tests/test_kernel.py` take the `sid` keyword. #994 (user todos) stubs `_fold_tasks` with a one-argument lambda in `tests/test_user_todos.py`; whichever of the two lands second makes that a one-line edit.

Scope: these memos serve a build whose parse, store and log stand while another input moved. Today that is every background tab on every judge pass, states row and task-store write. A later change that replaces the judge generation in the signature with per-session store identities would remove most of those background rebuilds; the memos then serve mainly the watched tab's builds between transcript writes (a live-tail change over one parse), and the counters say what that is worth.

Measured with the script below on this head, a synthetic session under a temporary state root, medians of 25 runs, milliseconds (the machine carried other load, so the absolute figures vary between runs; the direction of every row held across three runs):

| | 200 turns | 1000 turns |
|---|---|---|
| task fold, every turn / one merged turn | 0.57 / 0.28 | 4.74 / 1.53 |
| live-merge sets, derived / served | 1.25 / 0.001 | 7.59 / 0.001 |
| 50 sealed postal cards, re-hydrated / values check | 0.14 / 0.04 | 0.13 / 0.04 |
| `build_session` over a warm prefix (300 nodes), ledger walked / served | 4.6 / 2.7 | 7.3 / 4.9 |

The two per-atom scans (the sets and the fold) grow with the transcript; the sets are derived up to three times per cycle per alive session. Nothing on the board changes: the tests compare a memoized build with a cold one at every step.

<details>
<summary>The measurement script (standalone; `python3 bench.py <repo root> [turns] [nodes] [cards]`)</summary>

```python
#!/usr/bin/env python3
"""A standalone measurement of the four chat-build costs this change memoizes, cold against warm, over a
synthetic session under a temporary state root. Usage: python3 bench.py <repo root> [turns] [nodes] [cards].
Prints medians in milliseconds. Nothing here is a test; the figures are for a PR body."""
import json
import os
import re
import statistics
import sys
import tempfile
import time
from datetime import datetime, timezone
from importlib.machinery import SourceFileLoader
from pathlib import Path

ROOT = sys.argv[1]
TURNS = int(sys.argv[2]) if len(sys.argv) > 2 else 200
NODES = int(sys.argv[3]) if len(sys.argv) > 3 else 300
CARDS = int(sys.argv[4]) if len(sys.argv) > 4 else 50
REPS = 25

os.environ["XDG_STATE_HOME"] = tempfile.mkdtemp()
os.environ.pop("ROMP_STATE_DIR", None)
os.environ["ROMP_KERNEL_NO_OPEN"] = "1"
os.environ.setdefault("ROMP_SERVE_TOKEN", "bench-token-DO-NOT-USE")
km = SourceFileLoader("romp_kernel_bench", os.path.join(ROOT, "bin", "romp-kernel")).load_module()
jd = km.jd

SID = "66666666-7777-8888-9999-bbbbbbbbbbb1"
PEER = "66666666-7777-8888-9999-bbbbbbbbbbb2"


def iso(t):
    return datetime.fromtimestamp(t, timezone.utc).strftime("%Y-%m-%dT%H:%M:%S.000Z")


td = Path(tempfile.mkdtemp())
cdir = td / "launchdir"; cdir.mkdir()
proj = td / "projects"
pdir = proj / re.sub(r"[^A-Za-z0-9]", "-", os.path.realpath(str(cdir))); pdir.mkdir(parents=True)
tpath = pdir / (SID + ".jsonl")
jd._rebind_state(td)
jd.PROJECTS = proj
jd.NAMES.mkdir(parents=True, exist_ok=True)
(jd.NAMES / SID).write_text("web\t%s\t#abcdef\n" % cdir)
(jd.NAMES / PEER).write_text("api\t%s\t#123456\n" % cdir)
km.NAMES = jd.NAMES
km._GLOBAL_CLAUDE_MD = td / "no-global-claude.md"
os.environ["CLAUDE_CONFIG_DIR"] = str(td / "claude")
now = int(time.time())
t0 = now - 3 * 86400
tm = {SID: {"state": "working", "since": now - 100, "model": "", "effort": "", "context": None, "compactPct": None, "color": None}}
km._tmux_sessions = lambda: tm
km._sdk = lambda: None

# the postal log: CARDS messages delivered to this session
mids = ["1788400000.%03d_1.TESTHOST" % i for i in range(CARDS)]
jd.MESSAGES.parent.mkdir(parents=True, exist_ok=True)
with open(jd.MESSAGES, "w") as f:
    for i, mid in enumerate(mids):
        f.write(json.dumps({"ev": "sent", "id": mid, "from": "api", "from_id": PEER, "to_id": SID,
                            "body": "message %d about the deploy" % i, "kind": "coordinate", "t": t0 + i}) + "\n")
caps = {mid: "api: message %d" % i for i, mid in enumerate(mids)}
km._msg_summaries = lambda: dict(caps)

# the transcript: TURNS complete turns, a TaskCreate every 10th, a delivered message every TURNS/CARDS-th
n = [0]
def uid():
    n[0] += 1
    return "bbbbbbbb-0000-0000-0000-%012d" % n[0]
t = [t0]
def tick():
    t[0] += 5
    return t[0]
last = [None]
recs = []
step = max(1, TURNS // CARDS)
for i in range(TURNS):
    u = uid()
    text = "step %d: tighten the notes-api search" % i
    if i % step == 0 and i // step < CARDS:
        text = "<!-- romp-msg-id: %s -->\nmessage %d about the deploy" % (mids[i // step], i // step)
    recs.append({"type": "user", "timestamp": iso(tick()), "uuid": u, "parentUuid": last[0], "promptSource": "typed",
                 "message": {"role": "user", "content": text}, "cwd": "/tmp/notes-api", "version": "2.1.0", "gitBranch": "main"})
    a = uid()
    content = [{"type": "text", "text": "Round %d." % i}]
    if i % 10 == 0:
        content.append({"type": "tool_use", "id": "tu_c%s" % a, "name": "TaskCreate",
                        "input": {"subject": "task %d" % i, "activeForm": "Doing task %d" % i}})
    content.append({"type": "tool_use", "id": "tu_%s" % a, "name": "Bash", "input": {"command": "uv run pytest -q"}})
    recs.append({"type": "assistant", "timestamp": iso(tick()), "uuid": a, "parentUuid": u,
                 "message": {"role": "assistant", "model": "claude-sonnet-4", "content": content, "stop_reason": "tool_use"},
                 "cwd": "/tmp/notes-api", "version": "2.1.0", "gitBranch": "main"})
    r = uid()
    results = [{"type": "tool_result", "tool_use_id": "tu_%s" % a, "content": "ok\n"}]
    if i % 10 == 0:
        results.insert(0, {"type": "tool_result", "tool_use_id": "tu_c%s" % a, "content": "Task #%d created" % (i // 10 + 1)})
    recs.append({"type": "user", "timestamp": iso(tick()), "uuid": r, "parentUuid": a,
                 "message": {"role": "user", "content": results}})
    b = uid()
    recs.append({"type": "assistant", "timestamp": iso(tick()), "uuid": b, "parentUuid": r,
                 "message": {"role": "assistant", "model": "claude-sonnet-4", "content": [{"type": "text", "text": "Step %d done." % i}],
                             "stop_reason": "end_turn"}, "cwd": "/tmp/notes-api", "version": "2.1.0", "gitBranch": "main"})
    last[0] = b
with open(tpath, "w") as f:
    for r in recs:
        f.write(json.dumps(r) + "\n")

parsed = km._parse(str(tpath), SID, now)
seg = km._segs_seam(parsed["turns"][0], {})[0]
nodes = {"g%d" % i: {"id": "g%d" % i, "text": "goal %d" % i, "t": t0 + i, "mt": t0 + i,
                      "parentId": None if i % 5 == 0 else "g%d" % (i - i % 5), "trail": [seg["id"]]} for i in range(NODES)}
p = jd.GOALDIR / (SID + ".json"); p.parent.mkdir(parents=True, exist_ok=True)
tmp = p.with_name(p.name + ".tmp"); tmp.write_text(json.dumps({"nodes": nodes, "status": {}, "seams": [], "lastNode": None})); os.replace(tmp, p)


def med(fn, reps=REPS):
    xs = []
    for _ in range(reps):
        a = time.perf_counter(); fn(); xs.append((time.perf_counter() - a) * 1000.0)
    return statistics.median(xs)


def build():
    km._live_scope.names = km._names_snapshot()
    try:
        return km.build_session(SID, now, tm)
    finally:
        km._live_scope.names = None


# warm everything: fold prefix, memos, parse
for _ in range(3):
    build()
m = build()
assert len([e for e in m["events"] if e.get("kind") == "postal-service"]) == CARDS, "cards did not render"
assert len(m["ledger"]["tree"]) == 80

# 1. the task fold: every turn (a direct call, the old per-build cost) vs one live-merged last turn
merged = dict(parsed, turns=[dict(x) for x in parsed["turns"]])
merged["turns"][-1] = dict(merged["turns"][-1], atoms=list(merged["turns"][-1]["atoms"]))
km._fold_tasks(parsed, SID)
fold_cold = med(lambda: km._fold_tasks(parsed))
fold_warm = med(lambda: km._fold_tasks(merged, SID))

# 2. the live merge's sets: derived vs served
def derive():
    km._merge_sets_memo.pop(SID, None)
    km._merge_tx_sets(parsed, SID)
sets_cold = med(derive)
sets_warm = med(lambda: km._merge_tx_sets(parsed, SID))

# 3. the sealed postal cards: a re-hydration of the sealed list vs the values check
entry = km._chat_fold_get(SID)
raw, cards = entry["postal_raw"], entry["postal_cards"]
assert len(raw) == CARDS, len(raw)
km._live_scope.names = km._names_snapshot()
idx = km._postal_index()
msum = lambda: dict(caps)
gate_old = med(lambda: km._hydrate_postal(list(raw), idx, SID, captions=msum))
gate_new = med(lambda: km._postal_card_deps(cards, idx, msum))
km._live_scope.names = None

# 4. the whole build_session over the warm prefix: the ledger memo cleared before each build vs held
def build_cleared():
    km._ledger_memo.clear()
    build()
led_cold = med(build_cleared, 15)
led_warm = med(build, 15)

print("synthetic session: %d turns, %d goal nodes, %d sealed postal cards; medians of %d runs, ms" % (TURNS, NODES, CARDS, REPS))
print("task fold           every turn %.2f    one merged turn %.2f" % (fold_cold, fold_warm))
print("live-merge sets     derived %.2f       served %.3f" % (sets_cold, sets_warm))
print("sealed postal cards re-hydrated %.2f   values check %.2f" % (gate_old, gate_new))
print("build_session (warm prefix)  ledger walked %.1f   ledger served %.1f" % (led_cold, led_warm))
```

</details>

## Tests

`tests/test_chat_fixed_cost_memos.py` (new, 40 cases), every one red on the tree before its commit: all 40 on the base, and on the commit-1 tree the 7 that commit 2 adds or extends (the five `TaskFold` cases, the four-memos report and the task-fold eviction):

- `SigLabels`: a real `_chat_build_sig` has one label per position and a row slot either way; every moved component is named, a reshaped key names the states section (fails before: `_chat_sig_labels` and `_chat_sig_miss` do not exist).
- `Collector`: `build_chat` splits watched from background and attributes the latter; the snapshot is a copy; the four memos report under `memos` (fails before: `build_chat` does not exist; `KeyError` on the memo keys).
- `MemoBump`: concurrent bumps land exactly; the bump waits on `_chat_fold_lock` (fails before: no `_chat_memo_bump`).
- `TwoTabAttribution`: the real `_push` over two tabs, one watched: first cycle `active_built 1, bg_built 1, bg_miss cold 1`; nothing moved: both served; the background transcript, a judge pass, the background states file each rebuild it under their label; the watched tab's own growth counts under `active_built`. The eviction of the merge sets (a sid neither shown nor alive), the ledger memo and the task fold memo (a tab no longer shown; this cycle's thread kept) (fails before: `KeyError: active_built`; the memos do not exist).
- `MergeSets`: the same parsed object hits, a fresh parse misses, sids do not share; the sets equal the unmemoized derivation and the three sets are frozen; a merge hands `prune_live` the memoized sets and a withheld uuid makes a new set; no live atoms skip the memo; the memo is bounded one eviction at a time (fails before: `_merge_tx_sets` does not exist).
- `PostalCardDeps`: each embedded value is a component and nothing else is; the caption map is fetched once and only for a card with a mid.
- `MsgSummariesKey`: a caption write, a store publish, a journal row, an archive write, a states row, the transcript's size and the pending cut each rescan that session alone; unchanged inputs rescan nothing; the key is taken before the scan (fails before: a caption write with no mtime change rescans nothing).
- `CycleCaptionSlot`: reads within one `_pusher_cycle` fetch the map once and the slot closes with the cycle; the slot is open while the jobs run and closed when they raise; the timeline's postal connectors inside a cycle read the cycle's map and fetch nothing new, and outside a cycle they read the map directly (fails before: no slot, `_msg_summaries_scoped` does not exist).
- `PostalGate`, over `build_session` on a synthetic session with a delivered message: a judge pass that moved no caption leaves the sealed card alone (a spy on `_hydrate_postal` sees the tail pass only; fails before: the gate re-hydrated the sealed raw list and the commit hydrated it again); a caption change re-hydrates exactly the sealed card and refreshes it; a peer's colour refreshes it while a caption for another message does not; a build outside the names scope seals unverified and the next scoped build verifies once; the commit hydrates only the raw events new since the seal. Each compares against a cold build.
- `LedgerMemo`: unchanged inputs hit and the served ledger equals a fresh walk; a store publish, a journal row, a clear, a transcript append and a seam change each miss once; a rewind hold and a live merge bypass; a warm anchor learned for this session's node misses once and another session's does not; a resolve landing during the walk misses the next build, never a stale hit; a warm resolve writes the table entry before it bumps the revision (a dict subclass records the revision at the write); a store published between the build's two loads misses the next build (a stub on `jd.load_goals_shared_or_fault` publishes new seams between them); a clear landing during the walk misses the next build; a muted tab still ships no tree (fails before: `_ledger_memo_stats` does not exist).
- `TaskFold` (commit 2): the same object serves every turn, a live merge rescans the last turn alone, a re-parse rescans all; a turn grown in place is rescanned; a rejected create and a late result fold as before; the result is a fresh list and the store reader leaves it unchanged; `build_session` hands its sid to the fold (fails before: `TypeError` on the two-argument call).

Four reversions of the mechanism, made one at a time with the tests in place, each go red: dropping the seams from the ledger key (one case), bumping the revision before the table write (one), reading the caption map directly from the timeline (one), and not opening the cycle slot (three).

Existing modules: `tests/test_chat_fold.py`'s postal cases follow the new contract (the caption's change alone refreshes the card, a judge pass re-hydrates a sealed send no longer, a log append does, the commit hydrates the tail pass only); `tests/test_postal_hydrate_scope.py` checks `build_session`'s sid through a spy on the hydrator instead of the call's source text and pins the per-event independence (`hydrate(A + B) == hydrate(A) + hydrate(B)` at every split, and the captions getter read once); `tests/test_perf_stats.py` and `tests/romp-perf.bats` carry the new fields; `tests/test_kernel.py`'s seven stubs take `sid=None`.

Targeted runs on this head: `tests/test_chat_fixed_cost_memos.py` (40), `tests/test_chat_fold.py`, `tests/test_postal_hydrate_scope.py`, `tests/test_perf_stats.py`, `tests/test_perf_bench.py`, `tests/test_stage0_log_readers.py`, `tests/test_chat_delta_resync.py`, `tests/test_kernel_rewind.py`, `tests/test_kernel_tool_result_scan.py`, `tests/test_judge_authoritative_plan_sync.py`, `tests/test_pusher_change_gates.py`, `tests/test_kernel_awaiting_lift.py`, `tests/test_kernel_bg_placed_tops.py`, `tests/test_kernel_goal_cache_wiring.py`: 458 passed; `tests/test_kernel.py`'s fold, todo and push-cache cases: 24 passed; `tests/romp-perf.bats` 15 of 15.

Full suite on this head (`pytest -q -n 8 tests/`): 10876 passed, 118 skipped, 0 failed, 76 s. No webview file changes, so the TypeScript lane is untouched.

Tier: fix

### DIFF
--- a/bin/romp
+++ b/bin/romp
@@ -1248,7 +1248,19 @@ print("stages    jobs %s   push %s  (chat %s  feed %s  timeline %s  send %s)"
 parts = []
 for k in ("chat", "feed", "timeline", "feedJson"):     # feedJson: GET /feed.json's own reads, apart from the pusher's feed
     built, cached, ms = dl("builds", k, "built"), dl("builds", k, "cached"), dl("builds", k, "ms")
-    parts.append("%s %d built / %d cached%s" % (k, built, cached, (" (%.0f ms avg)" % (ms / built)) if built else ""))
+    split = ""
+    if k == "chat":
+        # the chat builder's attribution: rebuilds of the watched tab against background tabs whose signature
+        # moved, then per signature component the background rebuilds it caused; a kernel from before the
+        # split reports neither field and prints the line as before
+        active, bg = dl("builds", k, "active_built"), dl("builds", k, "bg_built")
+        if active or bg:
+            labels = sorted(((b.get("builds") or {}).get("chat") or {}).get("bg_miss") or {})
+            causes = ", ".join("%s %d" % (lab, dl("builds", k, "bg_miss", lab)) for lab in labels
+                               if dl("builds", k, "bg_miss", lab))
+            split = "; %d watched, %d background%s" % (active, bg, (": " + causes) if causes else "")
+    parts.append("%s %d built / %d cached%s"
+                 % (k, built, cached, (" (%.0f ms avg%s)" % (ms / built, split)) if built else ""))
 print("builds    " + "   ".join(parts))
 parts = []
 for kind in ("full", "delta", "deduped"):
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1078,7 +1078,7 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   gauge `entries`; `segs_hit` and `segs_miss` count the segments served and
   derived. `dead_serve`, `dead_miss` and `dead_failed_serve` are the dead-lane
   memo's outcomes on the same block, so one block carries every lane.
-  Three memos cover the chat build's per-build fixed costs, each keyed on the
+  Four memos cover the chat build's per-build fixed costs, each keyed on the
   inputs it reads and evicted by the pusher with the tab set (a comment thread
   built this cycle is kept, like its fold prefix). `chatMergeSets` is the
   live-tail merge's memo of the sets it derives from a parsed transcript (the
@@ -1104,7 +1104,11 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   live atoms: the last turn's segments differ from the parse's),
   `bypass_hold` (an armed rewind hold filters a store copy per build),
   `bypass_empty` (a store with no nodes), `evict` (entries dropped for tabs
-  no longer shown) and the gauge `entries`.
+  no longer shown) and the gauge `entries`. `chatFoldTasks` is the per-turn
+  memo of the transcript's task fold, keyed per session on each turn's atoms
+  list and fingerprint: `hit` and `miss` count turns served from the memo
+  against turns scanned, so a build of a working session with one moved turn
+  is one miss, plus the gauge `entries` (sessions held).
 - `judge`: `passes`, `ms_sum`, `ms_last`, `ms_mean` (wall time; a pass waits
   on model calls), `cpu_ms_sum` (CPU time of the judge tier threads and every
   per-session worker they run; the workers' share is `cpu_ms_workers`).
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -973,6 +973,15 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   `push.*` stages count every push, including the one a connecting page gets,
   so they can add up to more than `push`.
 - `builds`: `chat`, `feed`, `timeline`, each with `cached`, `built`, `ms`.
+  `chat` also carries `active_built` and `bg_built` (rebuilds of the watched
+  tab, served while its exact key holds, against rebuilds of a background tab
+  whose signature moved) and `bg_miss`, a map from each labelled component of
+  the chat-build signature (`transcript`, `states`, `judge_gen`, `tasks`,
+  `cut`, `row`, plus `cold` for a tab with no cached build and `nosig` for
+  one whose signature could not be taken) to the background rebuilds it
+  caused. A rebuild with several moved components counts under each, so the
+  map's sum can exceed `bg_built`. `romp perf` prints the split and the
+  non-zero causes after the chat average.
 - `sends`: `full`, `delta`, `deduped`, each a map from slot name (`chat`,
   `feed`, `bars`, `taborder`, ...) to `count` and `bytes`. A deduplicated frame
   was built and compared, then not sent.
@@ -1069,6 +1078,33 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   gauge `entries`; `segs_hit` and `segs_miss` count the segments served and
   derived. `dead_serve`, `dead_miss` and `dead_failed_serve` are the dead-lane
   memo's outcomes on the same block, so one block carries every lane.
+  Three memos cover the chat build's per-build fixed costs, each keyed on the
+  inputs it reads and evicted by the pusher with the tab set (a comment thread
+  built this cycle is kept, like its fold prefix). `chatMergeSets` is the
+  live-tail merge's memo of the sets it derives from a parsed transcript (the
+  uuids and user texts the transcript already holds, and the newest human
+  turn's time), one entry per session keyed on the parsed session object's
+  identity and shared by the chat, feed and timeline builds of one cycle:
+  `hit` and `miss` (merges served against derived) and the gauge `entries`
+  (a session neither shown as a tab nor alive is dropped). `chatPostal` is
+  the chat fold's memo of a tab's sealed postal cards, keyed on the values
+  the cards embed from outside the transcript (the message log's identity
+  and, per card, its caption and its peer's name and colour): `gate` (gate
+  checks that re-hydrated a tab's sealed cards because one of those values
+  moved, or because the entry was sealed outside the pusher's names snapshot
+  and had to be verified), `hit` (checks that verified the sealed cards from
+  their recorded values without hydrating), and `commit_new` (raw postal
+  events hydrated at fold commits; each is hydrated once, when it is first
+  sealed). Before this memo every judge pass re-hydrated every tab's sealed
+  cards, although a caption is the only judge-written value a card carries.
+  `chatLedger` is the chat build's memo of a session's goal-tree walk and
+  live roots, keyed on the parsed transcript's identity, the store's
+  identity and seams, `cleared.jsonl`'s identity and the warm-anchor table's
+  per-session revision: `hit` and `miss`, `bypass_live` (a build that merged
+  live atoms: the last turn's segments differ from the parse's),
+  `bypass_hold` (an armed rewind hold filters a store copy per build),
+  `bypass_empty` (a store with no nodes), `evict` (entries dropped for tabs
+  no longer shown) and the gauge `entries`.
 - `judge`: `passes`, `ms_sum`, `ms_last`, `ms_mean` (wall time; a pass waits
   on model calls), `cpu_ms_sum` (CPU time of the judge tier threads and every
   per-session worker they run; the workers' share is `cpu_ms_workers`).
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -210,7 +210,14 @@ class _PerfStats:
       builds                       chat / feed / timeline / feedJson -> {cached, built, ms}: served
                                    from the build cache vs rebuilt, and the rebuild time. feedJson is
                                    GET /feed.json's own reads (_pure_feed), kept apart from `feed`,
-                                   the pusher's (review find, 2026-09-08)
+                                   the pusher's (review find, 2026-09-08). chat also carries
+                                   active_built / bg_built (rebuilds of the watched tab, served while
+                                   its exact key holds, against rebuilds of a background tab whose
+                                   signature moved) and bg_miss {transcript, states, judge_gen, tasks,
+                                   cut, row, cold, nosig}: per labelled _chat_build_sig component, the
+                                   background rebuilds it caused (one count per differing component,
+                                   so the sum can exceed bg_built; cold = no cached build, nosig = no
+                                   signature could be taken); see build_chat
       sends                        full / delta / deduped -> {slot: {count, bytes}} per dedup-slot
                                    name (chat, feed, bars, taborder, ...; at most SLOTS names, the rest
                                    under "other"). A deduped frame was built and compared, not sent
@@ -226,7 +233,18 @@ class _PerfStats:
                                    interrupt-marks memo (_intr_marks_memo_report: hit, miss, evict,
                                    entries) and the awaiting overlay's states-log fold
                                    (_states_overlay_report: hit, append, refold, fail, evict,
-                                   entries), both trimmed to the interrupt tick's alive set each cycle
+                                   entries), both trimmed to the interrupt tick's alive set each cycle;
+                                   the chat build's fixed-cost memos: chatMergeSets (the live merge's
+                                   transcript-side sets, one entry per sid on the parsed session's
+                                   identity, see _merge_tx_sets) -> hit / miss and the gauge entries;
+                                   chatPostal (the chat fold's sealed postal cards, keyed on the values
+                                   they embed, see _postal_card_deps) -> gate (gate checks that
+                                   re-hydrated a tab's sealed cards), hit (checks that verified them
+                                   from their recorded values), commit_new (raw postal events hydrated
+                                   at fold commits, each once); chatLedger (build_session's goal-tree
+                                   walk and live roots per sid, see _ledger_memo) -> hit / miss,
+                                   bypass_live (live atoms merged), bypass_hold (a rewind hold armed),
+                                   bypass_empty (a store with no nodes), evict, and the gauge entries
       judge                        passes (one per _producer pass), ms_sum / ms_last / ms_mean (wall:
                                    a pass is a join over the tier threads, so this is mostly model
                                    latency), cpu_ms_sum (CPU: the two tier threads' own time, from
@@ -250,6 +268,9 @@ class _PerfStats:
     SLOTS = 32
     STAGES = ("jobs", "push", "push.chat", "push.feed", "push.timeline", "push.send")
     BUILDS = ("chat", "feed", "timeline", "feedJson", "thread")
+    # builds.chat's bg_miss labels: _chat_build_sig's components (the transcript and states sections, then
+    # _CHAT_SIG_TAIL), a tab with no cached build, and a tab whose signature could not be taken
+    CHAT_MISS = ("transcript", "states", "judge_gen", "tasks", "cut", "row", "cold", "nosig")
     SEND_KINDS = ("full", "delta", "deduped")
 
     def __init__(self):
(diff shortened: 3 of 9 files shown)


ksarma:

Reviewed inline: each memo's key against the inputs its served value reads. The merge sets key on the parsed object's identity and hold the object. Every parse mints new turn dicts and atoms lists, so identity stands for content. The task fold adds a per-turn fingerprint against in-place growth and replaces its per-sid dict whole on each call. The postal gate keys on the log's identity plus, per card, the caption and the peer's name and colour. Reading `_hydrate_postal` end to end found no other value from outside the transcript. The ledger memo takes `cleared.jsonl`'s stat before the read and the anchor revision before the walk. It bypasses a live-merged parse, an armed rewind hold and an empty store. `_chat_memo_bump` takes `_chat_fold_lock` only from call sites outside it. The pusher's evictions run after the fold sweep's lock block, over list snapshots. The caption-map slot opens inside `_pusher_cycle`'s try and clears in its finally. A handler thread with no scope fetches directly. A memo-held versus memo-cleared comparison of chat, feed and timeline was equal on every surface. It ran over a synthetic multi-turn session with delivered messages and a goal tree.

Two notes, neither blocking:

- `kernel/kernel.py:30279`: the merge-set cap of 512 exceeds the parse cache's clear-at-256. A session with no live atoms returns before `_merge_tx_sets`. So after a cache clear, an idle alive session pins one older parse until it next streams or dies. Bounded; follow-up at most.
- The second commit's message names `TypeError` as the red mode of its seven cases. They fail on `AttributeError` in `TaskFold.setUp` and a `KeyError` on the counter. Red either way.

Tests: `tests/test_chat_fixed_cost_memos.py` fails 40 of 40 at main 3bebe6ba (individual AttributeErrors; the module imports). At 20161075 exactly the seven commit-2 cases fail. All 40 pass at head. `test_chat_fold`, `test_perf_stats`, `test_postal_hydrate_scope` and `test_kernel`: 8 red at main, 587 passed at head. `tests/romp-perf.bats`: case 5 red at main, 15 of 15 at head. CI green on all eleven checks at head.

Review run by the maintainer automated review session; verified against head ac482ced.

ksarma:
Head note: the branch was updated with main (b649dc26), so the head moved from ac482ced to 362609c4. The new head is a mechanical merge of ac482ced and b649dc26: its tree equals the merge-tree of its two parents, and its added and removed lines against main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 362609c4.

PR #1252, opened 2026-09-10
Awaiting overlay: a departed session's unreadable states file is named once per episode, not once per rebuild

ksarma:
## Symptom

When a session that has ended has a states file that exists but cannot be read, the kernel's stderr names it again on every rebuild that follows an interrupt tick, instead of once per unreadable episode, the contract stated in `_states_overlay_on`'s docstring and docs/reference.md. The file is still read after the session ends: by build_session when the session is kept open as a read-only tab or scrolled back with loadOlder, and by GET /classify for any sid. The awaiting chip's answer is the same either way (None on a failed read); the defect is log noise.

The rate is the reader's cadence: a background tab rebuilds when the judge generation or the task store moves, the active tab on any dirty mark, and loadOlder and GET /classify read per request, so one line per read that follows a tick's forget, at most one per cycle. The repeated failures need the shared reader's entry absent (after a kernel restart, an eviction, or a change to the file since its last good read), since an unchanged file is served from the reader's cache without being opened.

## Cause

The review-fix commit 21de3fc1 in #1228 made `_states_overlay_forget` discard a departed session's path from `_states_overlay_failed`, the once-per-episode latch, on the reasoning that nothing reads a departed session's file again. The three readers above do, none of them gated on liveness. The first read after the discard found the latch empty and wrote the line again.

## Fix

`_states_overlay_forget` leaves the latch alone. The set is bounded the way the fold cache is: `_states_overlay_on` clears it whole above 256 paths, under the lock it already holds on the fail branch, so a path stranded by a session whose file is never read again cannot pin it. A whole clear ends every open episode at once, so a still-unreadable file is named a second time after a clear, as the fold cache re-folds after its own clear. No per-path event but a good read ends an episode. The docstrings of both functions and the set's trailing comment say so; docs/reference.md already states once per episode and needs no change.

A liveness gate at build_session's `_session_awaiting` call was considered and rejected: it changes the chip's answer for a dead kept-open tab (a stale awaiting row would stop showing), covers neither loadOlder nor GET /classify, and moves a contract that belongs to the latch into its callers. The latch-side fix keeps once per episode for every reader, live or dead.

## Tests

tests/test_states_overlay_fold.py, each red on the unchanged kernel:

- `OverlayFold.test_a_departed_sids_open_episode_survives_the_forget_and_is_named_once` replaces `test_forget_also_drops_a_departed_sids_stranded_failed_episode` from 21de3fc1, which pinned the discard this change removes. It memoizes a readable file, appends a row and chmods it to 0 (so the next read is attempted rather than served from the reader's cache), reads three times with `_states_overlay_forget(set())` between the reads, and expects one "unreadable" stderr line, three fail counts, no memoized entry, and the path still in the latch; readable again, the overlay answers and the path leaves the set. Before: `3 != 1`.
- `OverlayFold.test_the_failed_set_is_cleared_whole_above_its_cap` seeds 256 synthetic paths into the latch, the cap itself, and runs the same recipe twice. The first failed read adds the failing path with nothing cleared (257 paths, one line), so 256 is the cap and not above it; the second clears the set whole down to that one path and, the clear having ended the episode, writes a second line for the same file. Before: the second read left all 257 paths in the set.
- `InterruptTickRetires.test_the_tick_leaves_a_departed_sids_open_fail_episode_alone` runs the real `_interrupt_block_tick` with the dead sid outside its alive set after a failed read and expects the path kept in the latch and the following read silent. Before: the path was gone from the set.

All three use the module's existing guard and skip for a user who reads through mode 000 (root). Changing the cap's `>` to `>=`, removing the cap lines, raising the cap, or checking `first` ahead of the clear fails the second test; restoring the discard loop fails the first and third.

Full suite (`pytest -n 6 tests/`): 10757 passed, 118 skipped, 1688 subtests passed.

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -24457,7 +24457,8 @@ def _states_awaiting_overlay(sid):
 
 _states_overlay_cache = {}    # str(states path) -> _fold_records entry over (last overlay row or None, working_after)
 _states_overlay_stats = {"hit": 0, "append": 0, "refold": 0, "fail": 0, "evict": 0}
-_states_overlay_failed = set()   # paths whose last read failed on a file that exists: one stderr line per episode
+_states_overlay_failed = set()   # paths whose last read failed on a file that exists: one stderr line per episode;
+#                                  cleared whole above 256 paths (_states_overlay_on), never per departed path
 _STATES_OVERLAY_LOCK = threading.Lock()   # the counters are bumped from the pusher, the connect-time builds on WS
 #                                           threads and GET /sessions at once, so a bare `+= 1` is a read-modify-write
 #                                           across threads (the _INTR_MARKS_STATS_LOCK precedent). The cache dict
@@ -24493,11 +24494,18 @@ def _states_overlay_on(path_s, kind):
     """The fold's `on` for one states file: count the path the fold took, and on "fail" (the file exists and
     could not be stat'ed, opened or read; the fold answered the empty state and memoized nothing) write one
     stderr line per episode, so a failed read is told apart from a rewrite in GET /perf and in the log. A
-    later good fold of the same file ends the episode. Cheap on purpose, since it runs inside every fold:
-    one locked increment, and the set is touched only on a failure or while an episode is open."""
+    later good fold of the same file ends the episode, and no per-path event does: the interrupt tick's forget
+    (_states_overlay_forget) leaves the set alone, since a departed session's file is still read. The set is
+    instead cleared whole above 256 paths, the fold cache's own bound (fold_records), so a path stranded by a
+    session whose file is never read again cannot pin it forever; a whole clear ends every open episode at
+    once, so a still-unreadable file is named a second time after it, exactly as the fold cache re-folds
+    after its clear. Cheap on purpose, since it runs inside every fold: one locked increment, and the set is
+    touched only on a failure or while an episode is open."""
     _states_overlay_bump(kind)
     if kind == "fail":
         with _STATES_OVERLAY_LOCK:
+            if len(_states_overlay_failed) > 256:          # the fold cache's cap: a stranded path cannot pin the set
+                _states_overlay_failed.clear()
             first = path_s not in _states_overlay_failed
             _states_overlay_failed.add(path_s)
         if first:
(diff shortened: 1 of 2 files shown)


ksarma:
Merging as shipped. One optional wording follow-up; nothing to fold into this PR.

Reviewed inline: the fail latch is now touched only inside `_states_overlay_on`. There the clear above 256 paths, the membership test and the add run as one block under `_STATES_OVERLAY_LOCK`. Two readers failing on the same path in one episode therefore write one line, and no reader sees a half-cleared set. The only unlocked access is the pre-existing truthiness check on the good-read branch. `_states_overlay_forget` no longer discards departed paths. That is right: `build_session` (a dead tab kept open, loadOlder) and `GET /classify` still read a departed session's file with no liveness gate. None was added, and a failed read still answers None. The order is clear, then test, then add. With 256 paths seeded, the first failure adds the 257th with nothing cleared. The second clears whole down to the failing path with a second line. The cap test pins both the set contents and the line count. The cost is bounded and keyed on a size event, not a timer. A clear needs 256 further distinct failing paths and names each still-unreadable file once more, the same bound `event_model.fold_records` uses. Both docstrings and the set's trailing comment describe the code. The once-per-episode sentence in `docs/reference.md` still holds. Added lines carry only the module's placeholder sids and synthetic paths. Body and commit match the diff.

- `docs/reference.md:1050`, optional follow-up: add a clause parallel to line 1060. It would say the fail latch is also cleared whole above 256 paths, and that a clear ends every open episode.

I confirmed the tests go red first with my own run. At the base with the head's test file, exactly three tests failed and 18 passed: `test_a_departed_sids_open_episode_survives_the_forget_and_is_named_once` (3 != 1), `test_the_failed_set_is_cleared_whole_above_its_cap` (257 paths left) and `test_the_tick_leaves_a_departed_sids_open_fail_episode_alone` (path gone from the set). At the head, `tests/test_states_overlay_fold.py` plus `tests/test_event_model_incremental.py` gave 42 passed. The five mutations the body names each go red. CI at 7b2b7909: all eleven checks green. Tier policy reports the fix tier merges on green.

Review run by the maintainer automated review session; verified against head 7b2b7909.

ksarma:
Head note: the branch was updated with main (cfc44a4d), so the head moved from 7b2b7909 to 07e56350. The new head is a mechanical merge of 7b2b7909 and cfc44a4d: its tree equals the merge-tree of its two parents, and its added and removed lines against main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 07e56350.

PR #1225, opened 2026-09-09
Guide: a tag group's header shows the tag's color and name, then the chevron and the count, the order the strip builds (T284)

ksarma:
## Symptom
The guide's Tags and groups paragraph (docs/guide.md) says each section header in the tab strip "shows a chevron, the tag's color, its name, and a member count". The header a reader sees shows the tag's chip (its color and name) first, then the chevron and the count.

## Cause
#1185 (T284) reordered makeGroupHead in ui/webview/render.ts to append the chip, then the caret, then the count, then (while the section is collapsed) the pip, and pinned that order in ui/webview/tab-groups.test.ts and tests/test_tab_groups_rows.py. The guide sentence, written for the earlier caret-first layout, was not updated. .tab-group-head is a flex row and no rule on it or on its parts (styles.css) sets order, flex-direction or margin-left: auto, so DOM order is what the reader sees.

## Fix
The one sentence now reads: "Each header shows the tag's color and name, then a chevron and a member count." The rest of the paragraph (the count while collapsed, the small dot after it) already matches the header and is unchanged. No code, CSS or kernel change. The render.ts header comment's "color as a short bar" phrase is a code comment and is left for a code change.

#1168, and #1201 which builds on it and carries the same two hunks, also edit this paragraph: the sentence two lines above this one (the untagged sessions' own row) and the paragraph's closing sentence (where four lines are added), with the header sentence carried as context. None of the three changes the same sentence, so after the first of them merges the others need a one-line rebase of this hunk, with no wording to reconcile.

## Tests
- ui/webview/tab-groups.test.ts, "the guide lists the header's parts in the built order: the tag's color and name, then the chevron and the count (T284)", beside the every-tag-rule guide pin. Three checks: assert.match on the new sentence across its line wrap; assert.doesNotMatch on the caret-first phrasing; and, over every .tab-group-head, -caret, -chip, -count and -pip rule in styles.css, that none sets order, flex-direction or margin-left: auto, so the build order the guide names stays the order the reader sees. On the unchanged guide the match fails (57 tests, 56 pass, 1 fail). With the guide edited, an order: -1 added to .tab-group-caret fails the CSS check and no other test in the module; a flex-direction on .tab-group-head or a margin-left: auto on .tab-group-count fails it too; a border-radius added to .tab-group-count does not trip it. After the edit, 57 of 57 pass.
- npm run typecheck: clean. Full npm test: 3761 tests, 3761 pass, 0 fail.

Tier: docs

[...]

PR #1253, opened 2026-09-10
Tab groups: the header-order guard also rejects flex-flow and rtl reorderings, and the guide names the pip's retrying state

ksarma:
## Symptom
Two follow-up notes from the review record on #1225 (merged as e70f818b). The test that keeps a tag group's header parts in their built order accepted a `flex-flow: row-reverse` on `.tab-group-head`, and a `direction: rtl` there; both reverse the row the guide describes. The guide's sentence on a folded header's small dot said the dot shows when a hidden tab is working or waiting on you, and left out the amber dot `sectionPip` shows when a hidden tab hit an API error and is retrying on its own.

## Cause
The T284 test in `ui/webview/tab-groups.test.ts` reads every `styles.css` rule on the header and its parts and refuses an `order:`, a `flex-direction:` or a `margin-left: auto` in the body. `flex-flow` is the shorthand for `flex-direction` and `flex-wrap`, and `direction: rtl` lays the row out from the right; the regex named neither. The guide sentence named two of `sectionPip`'s three states (`sectionPip` in `ui/webview/tab-state.ts` returns blocked for a member blocked or waiting on you, else working, else retrying); the amber retrying state was never in it.

## Fix
`ui/webview/tab-groups.test.ts`: the guard is a function, `headerReorderRules(css)`, returning the selectors of the header rules whose body carries an `order:`, a `flex-direction:`, a `flex-flow:` whose value has `row-reverse` or `column-reverse` (the wrap may come first), `direction: rtl`, or `margin-left: auto`, in any case, as CSS reads property names and keywords. The anchors keep `border:` from reading as `order:` and a custom property such as `--head-direction:` from reading as `direction:`. The T284 test asserts it finds nothing in `styles.css`. No CSS change.

`docs/guide.md`: the small-dot sentence says the dot shows when a hidden tab is busy or needs you (the dot is rendered only when `sectionPip` returns a state), and names the three states in the rule's order with their colors: red when one is blocked or waiting on you, otherwise gold when one is working, otherwise amber when one hit an API error and is retrying on its own.

## Tests
- `executed: the header-order guard rejects a flex-flow with a reverse value, wrap first or last, in any case; a wrap alone passes`: before the change, the guard returns no selector for `.tab-group-head { flex-flow: row-reverse; }`. The wrap-first and uppercase fixtures each hold one part of the regex: matching only the value's first token, or dropping the `i` flag, fails this case.
- `executed: the header-order guard rejects direction: rtl on the header, and accepts ltr and a custom property whose name ends in direction`: before the change, the guard returns no selector for `.tab-group-head { direction: rtl; }`. Dropping the anchor before `direction` fails it on `--head-direction: rtl`.
- `executed: the header-order guard's earlier forms still reject (order, flex-direction, a margin-left: auto push) and a border, a fixed margin or an unrelated rule passes`: a regression check on the forms the guard already had; passes before and after. Dropping the anchor before `order` fails it on a bare `border:`.
- `the guide's small-dot sentence says the dot shows only in the pip's three states and names them in the rule's order (tab-state.ts sectionPip): red for blocked or waiting on you, else gold for working, else amber for an API retry`: before the change, the guide has the two-state sentence.
- The T284 test still passes on the unchanged stylesheet.

On the unchanged guard and guide, `ui/webview/tab-groups.test.ts` runs 63 tests, 60 pass, 3 fail (the two new guard cases and the guide case). After the change, 63 of 63 pass. `npm run typecheck` is clean. Full `npm test`: 4034 pass, 0 fail, and `timeline-tags-scale.test.ts` run alone, 22 pass, 0 fail (4056 in all).

Tier: docs

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -150,9 +150,11 @@ tag order, each with a header in the tag's color, and the untagged sessions on a
 own at the end. A session with several tags appears under each of them; every copy is the same
 session (click either to open it, and closing either ends it). Each header shows the tag's color and name, then a chevron and a
 member count. Click a header, or press Enter on it, to fold its section down to the header
-alone; the count then says how many tabs are folded away, and a small dot after it says when
-one of them is working or waiting on you (hover it for their names). To keep one tab visible
-while its section is folded, right-click the tab and pick **Show when folded** under **Tags**;
+alone; the count then says how many tabs are folded away, and a small dot after it shows when
+one of them is busy or needs you: red when one is blocked or waiting on you, otherwise gold
+when one is working, otherwise amber when one hit an API error and is retrying on its own
+(hover it for their names). To keep one tab visible while its section is folded, right-click
+the tab and pick **Show when folded** under **Tags**;
 the header's count then leaves that tab out; when every tab in a section is set to
 show, the folded header shows the full count and its tooltip says nothing is hidden. Pick it
 again to fold the tab with the rest. A tab set to show when folded keeps that setting when its
--- a/ui/webview/tab-groups.test.ts
+++ b/ui/webview/tab-groups.test.ts
@@ -1836,15 +1836,61 @@ test("the guide states the every-tag rule (T264b)", () => {
   assert.doesNotMatch(GUIDE, /sits under the first of them in your tag order/, "the retired home-tag sentence is gone");
 });
 
+test("the guide's small-dot sentence says the dot shows only in the pip's three states and names them in the rule's order (tab-state.ts sectionPip): red for blocked or waiting on you, else gold for working, else amber for an API retry", () => {
+  const prose = (t: string) => new RegExp(t.replace(/[.()]/g, "\\$&").split(" ").join("\\s+"));   // the guide wraps its lines
+  assert.match(GUIDE, prose("a small dot after it shows when one of them is busy or needs you: red when one is blocked or waiting on you, otherwise gold when one is working, otherwise amber when one hit an API error and is retrying on its own (hover it for their names)."));
+  assert.doesNotMatch(GUIDE, prose("a small dot after it says when one of them is working or waiting on you"), "the two-state sentence is gone");
+});
+
 test("the guide lists the header's parts in the built order: the tag's color and name, then the chevron and the count (T284)", () => {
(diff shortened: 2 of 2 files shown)


ksarma:

Reviewed inline: the T284 guard is now `headerRules` plus `headerReorderRules` over a stylesheet string. I checked `REORDERS` claim by claim in a scratch copy. Matching only the first token of the `flex-flow` value fails the wrap-first fixture. Dropping the `i` flag fails the uppercase fixture. Dropping the anchor before `direction` flags the `--head-direction` fixture. Dropping the anchor before `order` flags the bare border fixture. A scratch `styles.css` with `flex-flow: row-reverse` or `direction: rtl` on `.tab-group-head` fails T284. On the real sheet the guard reads 16 header rules and flags none. The guide sentence is true at this head. `sectionPip` in `tab-state.ts` returns blocked, else working, else retrying. `render.ts` builds the pip only when that is non-null. The pip's classes resolve to `--st-blocked-bg` (red), `--st-working-bg` (gold) and `--st-retrying-bg` (amber). The pin reads the real `docs/guide.md`. `styles.css`, `tab-state.ts` and `render.ts` are byte-identical to the base, so the docs tier fits. Body and commit message match the diff. The 56 added lines carry no private strings, hostnames or home paths.

Two follow-ups, neither a defect of this change:
- `ui/webview/tab-groups.test.ts:1839`: the sentence pin reads the guide alone. A later change to `sectionPip`'s precedence or the pip's colour tokens leaves it green. Only `tab-state.test.ts` pins the order today.
- `ui/webview/tab-groups.test.ts:1865`: the guard still misses `margin-right: auto` and `margin-inline-start: auto` on a part, a `margin` shorthand ending in `auto`, `transform: scaleX(-1)`, and grid or absolute layouts. `headerRules` also skips a selector whose brace opens on the next line and a one-line `@media` wrapper. None is in `styles.css`.

Locally at head: typecheck clean; `tab-groups.test` 63 of 63 pass. With the base regex substituted into the head file, 60 pass and 3 fail (the two guard cases and the guide case), as the body states. The head file over the base alone fails only the guide case, since the guard is in that file. CI at b10f2f2f: all 11 checks green. The vscode-extension job ran 4071 tests: 4026 pass, 45 skipped, 0 fail.

Review run by the maintainer automated review session; verified against head b10f2f2f.

PR #1254, opened 2026-09-10
Feed: the session header's Clear all releases the hover-freeze gate the way a card's Clear does

ksarma:
## Symptom

In grouped mode, with the pointer resting on a session header row, clicking the row's Clear all starts the cards' exit and ghosts the row, but the next push does not apply. The kernel's confirmation of the clear, and every frame behind it, shows as pending on the headers until something else releases the hold: the local render of the 180 ms finalize, and only when no card is under the pointer, or a window blur. Under reduced motion the row is removed outright, so the release cannot come before that render.

## Cause

A header row holds the hover-freeze gate from its mouseenter (T285), keyed `h:<col>:<sid>`, and releases it on its own mouseleave. `clearSessionCards` dispatches a synthetic mouseleave on each leaving card, whose key is the card's own, and hands every header of the session straight to `startSessHeadExit`. `freezeLeave` releases only the key it holds, so the card dispatches leave the header's key held; the ghost is pointer-inert (`.sess-exit`, or removed under reduced motion), so no mouseleave of its own ever fires. A card's Clear has dispatched a synthetic mouseleave for this reason since 2026-07-03; the header path lacked it.

## Fix

`clearSessionCards` dispatches a synthetic mouseleave on every header of the session before starting its exit, the way the card loop just above it does. The release comes from the click: the flush's microtask applies the queued frame after the click's own handlers finish, so `pendingCleared` and the `askClearMany` post precede the render, as on the card path. `freezeLeave` ignores a key it does not hold, so the session's headers in the other columns are unaffected. The dispatch stays out of `startSessHeadExit`, which serves two other callers where the pointer is not on the row: `reconcileCol`'s render-time header exits, where the stale-freeze heal at the end of the render already covers a hovered element the render removed, and `dressHeaderIfLast` on a card's Clear, where the pointer is on the card and the card's own synthetic mouseleave releases the key it holds. The release belongs beside the card loop's, in the gesture that ghosts the row.

## Tests

- `ui/webview/feed-render-incremental.test.ts`, new case "a session header's Clear all releases the gate its row holds: the queued frame applies from the click, with no pointer leave and no render". With the pointer on the api session's header row and a push queued behind the hold, the row's Clear all is clicked through the delegate on the columns root (the stand-in's events do not bubble, so the click is dispatched on the root with the button shadowing its target). It asserts one `askClearMany` post for the session's card, the ghosted row, the frame still queued until the click's handlers finish, then the queued card rendered and the pending hint gone, with no timer advanced (the 180 ms finalize's render would heal the hold by itself). Before the change it fails on "the click released the gate: the queued frame applied with no pointer leave and no render": the queued card is still absent after the microtasks settle.
- Same file, the header badge test: its held-gate dispatch moves inside the `try`, so a throw there still reaches the release in its `finally` (noted in the review of #1232).
- `ui/webview/feed-sess-clear.test.ts`: the source pin on the header loop follows the new text; before the change it fails on the loop's old text.
- `npm run typecheck` clean. Full `npm test` over the built bundles: 4043 pass, 0 fail (4021 in the main run, 22 in the timeline-tags-scale file, run alone).

Tier: fix

### DIFF
--- a/ui/webview/feed-render-incremental.test.ts
+++ b/ui/webview/feed-render-incremental.test.ts
@@ -903,8 +903,8 @@ test("a hovered session header is one (column, session) row: the same session's 
   const rightOf = (h: El) => (win.innerWidth - h.getBoundingClientRect().right) + "px";
   assert.notEqual(rightOf(hw), rightOf(hb));
   hw.dispatchEvent(new Event("mouseenter"));                     // the pointer rests on the Working row
-  await dispatch(frame([...four, g1c]));                         // a push while it is held: one more web card, in Completed
   try {
+    await dispatch(frame([...four, g1c]));                       // a push while it is held: one more web card, in Completed
     assert.equal(card("g1c"), null, "the payload is queued, not rendered");
     assert.ok(badged(hb), "the same session's Blocked header carries its in-row badge");
     assert.ok(!badged(hw), "never inside the hovered row");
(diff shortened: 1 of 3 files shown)


ksarma:
Merging as shipped; nothing to change.

Reviewed inline: the header row's freeze listeners sit on the header element itself (`feed.ts` 3576-3577). The non-bubbling synthetic mouseleave therefore reaches `freezeLeave`, matching the card path's dispatch at 1377. `freezeLeave` returns unless it holds that exact key, so the session's headers in other columns are untouched. `tabScopeKey` only ever holds a card key, so that branch cannot fire for a header. `sessFreezeKey` reads `data-fcol` and `data-fsid`, which `startSessHeadExit` never changes: it re-keys `dataset.key`, adds `.sess-exit`, or removes the row under reduced motion. The dispatch precedes the call on both paths. The flush is a `queueMicrotask` and the delegate's click handler is synchronous. So `clearedStack.push`, `pendingCleared.add`, the `askClearMany` post and the 180 ms timer all precede the queued frame's render. The header carries no other mouse listeners, and its children use click handlers only, so nothing else reacts to the synthetic event. Keeping the dispatch out of `startSessHeadExit` is right: its other two callers run with the pointer off the row. The badge test's hunk moves the held-gate dispatch inside the try and changes no assertion. Body and commit match the diff, and the added lines carry synthetic names only.

I ran the red-first check myself:
- At base 79c7f969 with the head's two test files: 48 tests, 46 pass, 2 fail. The two failures are exactly the new case (the queued card still absent after the microtasks settle) and the feed-sess-clear source pin (the old loop text).
- At head: 48 of 48 pass, and `tsc` is clean.
- CI: all 11 checks green at 542ea4e7 (Python 3.10 to 3.13, bats, vscode-extension, tier label and Tier policy).

Review run by the maintainer automated review session; verified against head 542ea4e7.

PR #1255, opened 2026-09-10
Chat: switching to a loading tab takes the leaving tab's go-to-bottom and reply chips down with the loader

ksarma:
## Symptom

Switch from a tab you left at the top of a long transcript to a tab that is still loading (a skeleton after a reconnect, or a placeholder you just clicked) and the leaving tab's go-to-bottom chip, and any unread-reply chip, stay painted over the loader. A click on the stale reply chip looks its anchor up in the entering tab's view, parks it as the pending anchor, and answers with the couldn't-locate toast when the full lands. A click on the stale go-to-bottom chip sets follow mode on the entering tab's stale hidden view, which the landing full keeps, so a transcript of up to 300 rendered turns lands at the bottom instead of the saved spot.

A post-merge review comment on #1226 (https://github.com/romp-on/romp/pull/1226#issuecomment-5610743488) reproduced this in headless Chromium and Firefox and named this fix shape.

## Cause

The chips refresh only on events: the pane's scroll and resize, a land, an append. showActive's no-session branch hid every view, put the loader up and returned without running them. A leaving tab at scrollTop 0 whose switch changes no sibling of #content fires none of those events: the pane's height is the body minus its siblings, so hiding the views changes only its scroll extent, and there is no scrollTop to clamp.

Every other leaving tab heals on its own: one scrolled down, through the clamp's scroll event; one whose switch shows or hides the digest panel or the background-work panel, or changes the footer's height (the composer draft, the staged strip and the attachment strip are swapped for the entering tab before showActive runs), through the ResizeObserver on the resized pane. A tab left at the top with the same chrome on both sides, an empty composer being the common case, did not heal. #1226 narrowed the reply chips' read to liveSession, which hides them over a skeleton whenever they are re-read; nothing re-read them on the switch.

## Fix

ui/webview/render.ts, two places.

- showActive's no-session branch ends with updateJumpBtn(), its last statement once the loader or the no-sessions copy is in place (the test pins the order). In this branch updateJumpBtn measures nothing: its gate returns before the scroll extent is read.
- updateJumpBtn yields to the set before it measures: with no live session under the active tab (a loader, or no tab at all) it hides the go-to-bottom chip and runs updateReplyChips, whose liveSession read hides the reply chips and clears their signature, so the landing tab's first paint is not skipped as unchanged. Without this gate the added call would show the go-to-bottom chip over the loader on a short pane: .tab-loading-wait has min-height 60vh and #content 20px of vertical padding, so with the 2px at-bottom tolerance the loader overflows any pane under 60vh + 18px (a pane under roughly 320 to 450px depending on the chrome, or a tall composer). A window resize or a scroll during the load ran that measure before this change too; the gate covers that case as well.

The cost is one clientHeight read per run of the branch (a tab switch, dismissSession's fallback, a strip re-listing the active tab, an append arriving while the active tab is a skeleton). The stale go-to-bottom click's own views.get read is unchanged: with the chip gone, the click cannot happen.

## Tests

ui/webview/skeleton-tabs-wiring.test.ts. Three executing cases lift showActive, updateJumpBtn, updateReplyChips, liveSession and atBottom from render.ts with esbuild (the chat-exact-tail-exec.test.ts pattern) and drive them over a fake pane whose scroll extent is what its shown children stack to, as a browser's is; the loader's min-height is read from styles.css, and atBottomDist and isReplyReady are the real modules. The skeleton's stale copy keeps its hidden view and carries an unread open reply, so the liveSession read is the one clause of updateReplyChips' gate that hides the chips over it. Synthetic ids.

- run: switching from a tab left at the top of a long transcript to a loading tab takes both chips down with the loader. A 5000px transcript in a 600px pane, the reader at the top, a reply chip painted; the switch to a skeleton asks for the full, puts the loader up with the loading caption, hides both chips and clears the reply chips' signature. Before the change it fails at `no go-to-bottom chip over the loader` (false !== true).
- run: the loader itself overflows a short pane; the chip reads the set, not that measure. A 300px pane in a 500px window, where the loader's 300px min-height overflows the pane; the switch hides both chips and a further measure keeps the go-to-bottom chip hidden. Before the change it fails at `over a loading tab there is nothing to go to the bottom of` (false !== true), and with the updateJumpBtn() call but no gate it fails on the same line: the gate exists for this case.
- run: the gate reads the skeleton set, not the session map; the last tab closing drops the chips with the no-sessions copy. A loaded tab off the bottom keeps its chip; activeId null drops both chips with the no-sessions copy. Before the change it fails at `no chip over the no-sessions copy` (false !== true).
- Two source pins in the showActive test: the branch ends `updateStatusline(); updateJumpBtn(); return;` and updateJumpBtn carries the liveSession gate. Both fail before the change.
- The named liveSession list gains updateJumpBtn and updateReplyChips beside its seven siblings, so the read #1226 added is pinned by name; `updateJumpBtn reads the gated session` fails before the change.

On the file: 15 of 20 pass before the change, 17 of 20 with the call and no gate, 20 of 20 after. With the gate hiding the go-to-bottom chip but skipping the reply chips' re-read, all three run cases fail; with updateReplyChips reading the session map instead of liveSession, the named list and the first two run cases fail. `npm run typecheck` is clean. Full `npm test`: 4070 pass, 0 fail (4048 in the main run plus the timeline-tags-scale file run on its own, 22 pass).

Tier: fix

### DIFF
--- a/ui/webview/render.ts
+++ b/ui/webview/render.ts
@@ -10735,6 +10735,15 @@ function showActive(keep?: { uuid: string; y: number } | null) {
     } else { empty.style.display = ""; }
     document.body.style.removeProperty("--active-accent"); // no session → neutral window border
     updateStatusline();
+    // Take the leaving tab's go-to-bottom and reply chips down with it. They were measured against ITS transcript,
+    // and a tab left at the top of a long one fires none of the chips' events on the switch: no scroll clamp
+    // (scrollTop is 0 and stays 0) and no #content resize (the pane's height is the body minus its siblings; hiding
+    // the views changes only its scroll extent). Left up, a stale chip is a click into the wrong tab: the reply
+    // chip's anchor is looked up in the entering tab's view and lands as the couldn't-locate toast when the full
+    // arrives. With no live session under the active tab, updateJumpBtn's gate hides its chip without a measure and
+    // re-reads the reply chips, whose own gate hides them and clears their signature. The branch's last statement,
+    // after its state is settled (skeleton-tabs-wiring.test.ts pins the order).
+    updateJumpBtn();
     return;
   }
   document.getElementById("tab-loading")?.remove();   // the payload landed — the real view takes over in place
@@ -11134,6 +11143,11 @@ jumpBtn.innerHTML = '<svg viewBox="0 0 24 24" width="16" height="16" fill="none"
 function updateJumpBtn(): void {
   const c = document.getElementById("content");
   if (!c || c.clientHeight <= 0) { jumpBtn.hidden = true; return; }   // hidden pane measures 0 — no chip
+  // No live session under the active tab (a loading or skeleton tab's loader, or the no-sessions copy): nothing to
+  // go to the bottom of, so the set decides before the measure does. The loader's min-height (60vh, styles.css)
+  // overflows a pane under 60vh + 18px, and the measure alone would paint the chip over it (skeleton-tabs-wiring.test.ts).
+  // The reply chips still get their re-read: their own gate hides them over a loader and clears their signature.
+  if (!liveSession(activeId)) { jumpBtn.hidden = true; updateReplyChips(); return; }
   const off = c.scrollHeight > c.clientHeight + 2 && !atBottom(c);   // the chip: shown the moment the reader leaves the true bottom
   jumpBtn.hidden = !off;
   if (off) jumpBtn.style.bottom = (Math.max(0, window.innerHeight - c.getBoundingClientRect().bottom) + 8) + "px";
--- a/ui/webview/skeleton-tabs-wiring.test.ts
+++ b/ui/webview/skeleton-tabs-wiring.test.ts
@@ -2,12 +2,17 @@
 // skeleton-tabs.test.ts). After a redial the kernel sends only the active tab in full and lists the rest as
 // `skeleton` on the tab strip, each carrying a status frame instead of a transcript; the page keeps the stale
 // pre-outage sessions underneath and must never DISPLAY one as current. No jsdom for the webview, so these are
-// source-level pins (the prebuild-wiring.test.ts convention). Synthetic ids only.
+// source-level pins (the prebuild-wiring.test.ts convention), plus three executing cases at the end that lift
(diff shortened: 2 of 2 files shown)


ksarma:

Reviewed inline: `liveSession` is the gate's only input. It filters on the skeleton set and the session map and has no status field. So it is falsy only for no active tab, an id the kernel is withholding, or an id with no session entry yet. `showActive`'s no-session branch already sends those states to the loader or the no-sessions copy. Each writer of the two inputs (`applyTabOrderSkeleton`, `onFull`, `onDismiss`, `dismissSession`'s `sessions.delete`) re-runs `showActive` when the active tab is affected. A closed read-only session, a remote tab during a host outage, and a renamed or moved tab keep their chips. On the scroll and resize listeners, the gate costs one `Set.has` and one `Map.get`. `updateReplyChips` does not call `updateJumpBtn`. Its gate path clears `replyChipSig`, so the landing tab's first paint is not skipped. `updateStatusline` sets nothing the gate reads. The gated path returns before any measure, so `dismissSession`'s fallback, a re-listed active tab and an append on a skeleton active measure nothing. `jumpBtn` and `replyChips` are module-level nodes, never null. The test reads the worktree's `render.ts` and `styles.css` at run time. Its loader regex matches the real 60vh rule. The fix label fits. The added lines carry only the synthetic placeholder id.

Local runs of the head's `skeleton-tabs-wiring.test.ts`:
- Over base 3bebe6ba: 15 of 20 pass. The order pin, the named `liveSession` list and the three executing cases fail.
- With the `updateJumpBtn()` call and no gate: 17 of 20 pass. The short-pane case fails.
- At head b2331c78: 20 of 20 pass, tsc clean.

CI at head is green on all eleven checks.

Review run by the maintainer automated review session; verified against head b2331c78.

ksarma:
Head note: the branch was updated with main, so the head moved from b2331c78 to 7ee1028b. The new head is a mechanical merge of b2331c78 and 7136c205: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 7ee1028b.

PR #1256, opened 2026-09-10
Markdown images: a stale probe failure leaves a fresh budget and a newer probe alone

ksarma:
## Symptom
A markdown image the kernel serves (a `/file` or `/media` URL) fails, its per-message probe is in flight, a reconnect heal loads the URL, and then the URL fails again before that probe's error arrives (the file moved away, a relay blipped). The second failure arms three attempts afresh, but the late error takes them: one push later the image probes nothing, or once or twice instead of three times, and its caption waits for the next reconnect. The same late error can also let two probes for one URL overlap.

## Cause
`retryMdImgProbes` (`ui/webview/preview.ts`, from #1245) marked a URL in flight in a Set, and its failure callback applied the attempts-left it captured at fire time, with a check of the failed-URL memory standing in for "was this probe retired". A load of the URL by any probe clears the in-flight mark and the memory, and the URL's next failure re-arms both, so the stale error passed the memory check with the OLD count: it deleted the fresh budget (one attempt left at fire time) or cut it to two (three left), and its delete of the in-flight mark let the next push fire a second probe while the newer one was pending. A bare "still in flight" check is not enough either: the newer probe has re-marked the URL.

## Fix
`mdImgProbing` is a `Map<string, number>` from URL to the token of the probe in flight. Each probe `retryMdImgProbes` fires takes a fresh token, and its failure callback returns first when the URL's token is no longer its own, so a stale failure spends nothing and clears nothing. The memory check goes, since the token check covers it: the only path that forgets a URL is a load, and a load retires the in-flight token. The load path is unchanged (it deletes the URL from the map as it did from the set).

Two limits stay as they were, on purpose: a per-message probe that never settles keeps its URL marked in flight until some probe of it loads (a failing reconnect probe passes no failure callback, so it never clears the mark), and the failed-URL memory is page-life and uncapped.

## Tests
Three new cases in `ui/webview/md-img-park.test.ts`, each executed over the file's fake DOM through the public paths (`retryFailedPreviews`, `refreshSettledPreviews`, the capture-phase error listener), each red before the change:
- a stale failure from a probe fired with the last attempt does not delete the budget the URL's re-fail armed afresh: fails before on "the fresh budget survives the stale failure: the next push probes" (4 probes, expected 5)
- a stale failure from a probe fired with three attempts does not shorten the budget the URL's re-fail armed afresh: fails before on "push 3 after the re-fail: one probe, the fresh budget intact" (4, expected 5)
- a stale failure does not clear a newer probe's in-flight mark: no two probes for one URL overlap: fails before on "a push while the newer probe is pending fires no second probe for the URL" (4, expected 3)

A bare in-flight check in place of the token check (return when the URL has no probe in flight) turns only the third case red: the newer probe has re-marked the URL, so the stale failure passes it and shortens the fresh budget as it clears the mark. Removing the token check turns all three red, with the existing heal-while-pending case (the token check now covers what the removed memory check did).

The file's eight existing cases stay green (11 of 11 after). `npm run typecheck` clean. Full `npm test` on this head: 4070 tests, 4070 pass, 0 fail.

Tier: fix

### DIFF
--- a/ui/webview/md-img-park.test.ts
+++ b/ui/webview/md-img-park.test.ts
@@ -8,7 +8,9 @@
 // the per-message path, and that budget rides the URL, so a turn re-rendered on every push while it streams still gets
 // its three probes, a second img with the URL failing midway starts nothing over, a load clears the budget with the
 // memory (the URL failing again later starts afresh), and a heal that lands while a probe is pending is not undone by
-// that probe's failure; the reconnect-class heal probes every remembered URL, on the page or not (a turn evicted from
+// that probe's failure, nor does that probe's failure, arriving after the URL has failed AGAIN, shorten or delete
+// the fresh budget or clear a newer probe's in-flight mark (a failure counts only for the probe whose token
+// is current); the reconnect-class heal probes every remembered URL, on the page or not (a turn evicted from
 // the rendered window, a closed tab); and the chat's post-pass parks a re-rendered img with a remembered URL before the
 // browser fetches it. The module remembers failed URLs for the page life, so fresh() heals every remembered URL off
 // the DOM through the public reconnect path before each test: every test starts with an empty memory.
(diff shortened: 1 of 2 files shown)


ksarma:

Reviewed inline: the probe lifecycle in `ui/webview/preview.ts`. The token is a module-level counter incremented before each per-message probe, so it cannot repeat in a page life. The in-flight map is set only in `retryMdImgProbes`. It is deleted only by a fresh failure there or by the shared load path. The heal's probes also run that load path, so a load by any probe retires the token. The map's only readers are the per-push gate's `has(url)` and the callback's `get(url)`. The heal and the post-pass read the failed-URL memory, not the map. The stale callback returns before touching the mark, the budget or the memory. The fresh callback still clears the mark and decrements or deletes the budget. The dropped memory check was redundant: only the load path forgets a URL. A heal probe that fails while a per-message probe is pending leaves the token in place. That probe's later failure therefore still spends an attempt. A probe that never settles keeps its URL marked in flight until a load, as before this change. The `fix` label fits: a race defect with tests that fail before the change.

Test evidence:
- The head's `md-img-park.test.ts` over the base 3bebe6ba fails exactly the three new cases (4 vs 5, 4 vs 5, 4 vs 3) and passes the eight others.
- At the head, 11 of 11 pass and typecheck is clean.
- Both mutation claims in the body reproduce.
- Privacy scan of the added lines: clean.
- CI at head: all 11 checks SUCCESS.

Review run by the maintainer automated review session; verified against head 643e7039.

PR #1259, opened 2026-09-10
Chat pane: fail loudly when the federation manager is missing and never write the arrangement from that page; a session frame adds no duplicate id

amith-ananthram:
## What was wrong

A kernel-served chat page loads three classic scripts in order: the pane shim, `/dist/federation.js` (the multi-kernel manager), then `/dist/render.js` (the chat pane). If the middle one fails to fetch or to evaluate, nothing stops the pane: the shim's `deliver()` falls to its window-dispatch branch and the pane consumes the kernel's **raw** `tabOrder` frames. The browser's tab arrangement (`view-order.ts`, `romp:vieworder`) is applied only inside the manager's merge, so such a pane:

- shows the kernel's seed order instead of the arrangement, and snaps back to it on every push;
- on a drag, writes that seed (plus the one move) over the shared arrangement, so every other pane of the dashboard follows it.

Both fallbacks were designed for pages that never have a manager (a VS Code webview, the timeline page) and so degraded silently, against the fail-loudly rule. It surfaced on 2026-09-10 as one chat column of a split dashboard disagreeing with the other and reshuffling its tabs with no drag: the kernel had stalled connections for about three minutes during a dashboard reload, so one frame's bundle fetch failed. Reproduced twice in a hermetic browser (abort the bundle for one frame) and matched row for row in the order audit: every non-drag permutation had the raw-dispatch stack and landed on exactly the persisted seed order.

A second, smaller finding from the same trace: a new session's `tabOrder` frame lands before its session frame, so the pane's order already carries the id when the session frame's upsert pushes it again. `renderTabs` hid the duplicate; a drag's commit wrote it into the arrangement.

## The fix

- `frame-listener.ts` gains three pure helpers: `federationMissing(w)` (shim present **and** manager absent is *missing*, not "not part of this page"; a VS Code webview has no shim and is untouched), `federationLoadEntry(entries)` (the browser's resource entry for the bundle: status, transfer size, body size, duration, the measurement that tells a failed fetch from a bundle that failed to evaluate) and `fedRetryKey(location)` (the retry marker's key, per document: `sessionStorage` is shared by every same-origin frame of a tab, so two chat columns must not share a retry).
- `render.ts` at boot: when missing, reload **once** (the marker also carries the first pass's load entry; blocked storage means no retry at all, so a bundle that keeps failing cannot loop), file one `clientDiag` row per incident (`what: "federation-missing"`) from the pass that stays up (a failed retry: both fetches; a recovered boot: the first fetch), consume the marker with it so a later load gets its own retry, and past a failed retry show a bar **in flow above the tab strip**, never over it: the tabs under it still select, open their menu and close; only their order is wrong. While the page is in this state every drag starter refuses at `dragstart`, every `draggable` flag is off, a drop is committed only when the reorder happened (a refused one FLIPs the strip home), and the arrangement writer (`commitTabOrder`) stands down: an order that never passed through the arrangement is not one.
- The upsert's push becomes the pure `adoptArrival(order, id, existed)` in `tab-order.ts`; the revive skeleton's own push is guarded the same way.

## Tests

- `ui/webview/federation-missing.test.ts`: the three helpers executed (kernel page, healthy page, VS Code webview, non-function shim slot; aborted fetch vs evaluated body vs no status; `/chat` vs `/chat?col=2`), and the boot check, the guards, the drop's commit, the banner's placement and copy pinned at source.
- `ui/webview/federation-missing-exec.test.ts`: `commitTabOrder` and `reorderTo` lifted from `render.ts` and executed under both states. These two guards are unreachable from a browser story (no drag starts without a manager), and the review's mutation pass found them uncovered.
- `ui/webview/tab-order.test.ts`: `adoptArrival` executed (already carried, existing session, genuinely new).
- `tests/test_federation_missing_served.py`: a real browser against a hermetic kernel, 8 checks. With the manager, a synthetic drag writes the arrangement exactly once and persists the reverse of the kernel's order. Then the bundle is aborted for the chat frame: the pane retries once and shows the bar in flow above the strip; the strip shows the kernel's seed, not the arrangement it had just persisted; a plain click on a tab under the bar selects it; no tab is draggable; a synthetic drag is refused at `dragstart` with zero arrangement writes, also after Rename then Escape (which used to re-arm the drag); a second failing load gets its own single retry; the kernel's `client-diag.jsonl` gains one row per incident carrying both fetches. The bar's Reload heals it and files nothing. A bundle that fails only once is healed by the single retry with no bar, and its incident is still filed. A session planted mid-run and then dragged is persisted once. The retry's bound is the event it approximates, a third bundle request within one leg, not a harness timeout.
- Existing pins adjusted only to admit the new import names, the guard and the drop's commit shape (`perf-telemetry`, `federation-direct-delivery`, `view-order-wiring`, `subagent-view`, `skeleton-tabs-wiring`, `tab-drag-live`, `tab-strip-skip`, `tab-strip-skip-exec`, `revive-loader`).

For maintainers: the next time a pane reports "tabs won't stay put", `client-diag.jsonl` will hold a `federation-missing` row saying whether the bundle never arrived or arrived and failed to run.

Follow-ups noted by the review, not in this PR: the feed, Outline and timeline panes still degrade silently without their manager (nothing corrupts there; a lane drag persists nothing and reverts).


amith-ananthram:
Opened this because I noticed my tabs were reordering themselves.  Asked Claude to investigate the cause.  TLDR:

When a chat pane's federation manager fails to load (the bundle didn't arrive or didn't run), the pane used to fall back silently to the kernel's raw session order and, on any drag, overwrite the browser's shared tab arrangement with it, so every other pane followed. This PR closes that write path: a pane without its manager never writes the arrangement and never starts a drag. It then reloads itself once, which recovers the common case of a transient stall, and if the bundle fails a second time it shows a one-line bar above the tab strip with a Reload button and files a diagnostic row with the fetch details, so a maintainer can tell a failed download from a bundle that failed to run. Separately, a new session's frame no longer pushes a duplicate id into the pane's order, which a drag used to persist.

PR #1260, opened 2026-09-10
Chat: a pending bubble's ✕ drops the entry that owns its id, never the first same-text entry pressed in the same millisecond

ksarma:
## Symptom

Send the same text twice in one go (two staged items with the same text, each posted as its own send: a slash command, or an item with a goal cite; or two identical texts typed into a new session's tab before the kernel created it) and press ✕ on the second bubble: the first bubble disappears instead. On the next push the surviving bubble hides the kernel's copy of the first send; a second ✕ on it is answered with the too-late toast and the composer restore is reverted; a third ✕, on the kernel's own copy, cancels it. This is the follow-up recorded on #1224 after its merge.

## Cause

`dropPending` (ui/webview/send-pending.ts) read the press time and the text before the id: the `ts !== undefined` branch came first, and the id branch ran only for a ✕ with no press time. Our bubble's ✕ carries both (`data-qts` and `data-qid`, set in render.ts from the entry), and `registerOptimistic` stamps `Date.now()` once per send, so two same-text sends registered in one synchronous loop can share the stamp and differ only by id (`flushStaged` posts each staged slash command and goal-cited item as its own send; `adoptProvisional` posts each text a new session's tab held the same way). `findIndex` returned the first entry, while the cancel the same handler posts named the second's id and the kernel removed that copy and no other (`_cancel_parked` and `unqueue` look the id up; neither relocates by body). The client and the kernel then disagreed about which copy remained: the survivor's id was nowhere in the next frame, `cover` took the leftover copy by text, `hideQueuedCopy` hid it behind the survivor's bubble, and the next ✕ posted an id neither queue held.

## Fix

`dropPending` resolves the id first whenever the ✕ carries one: the entry that owns it, else the entry that copy covered by text on the last push. An id neither owns removes nothing, even when a same-text entry shares the ✕'s press time: that entry is another send, with its own ✕. A ✕ with no id keeps the older readings, by press time and text for an id-less bubble (older data; `newPending` always mints an id) and by text for an id-less kernel copy. The docstring says why, and the `PendingSend.ts` field comment now says the same (it had called the press time the bubble's identity). No caller changes: the qx handler already passes both, and the source pins in send-pending.test.ts hold.

Three corrections in the same files:

- `_wire_qid` (kernel/kernel.py) matches the id shape with `fullmatch`. With `match`, `$` admitted `"echo:" + 32 hex + "\n"`; held nowhere, that id was returned by `_client_qid` and would have ridden the parked op, the mirror and every chip. The comment above `_CLIENT_QID_RE` notes it.
- `_client_qid`'s refusal when the hold cannot be checked (a queue read or a live-echo read that raises) had no test. It has one now.
- docs/read-side.md said the tmux route's CLI-held copies are cancelled by index and body. They have no ✕ at all: `TmuxBackend` has no `unqueue`, so `build_session` ships them `cancelable: false` and the chat draws none. The index-and-body path belongs to a copy whose ✕ names no id (an op the kernel parked itself, a ✕ from an older client), and the sentence now says so.

Not in this change: the ✎ on a queued bubble. `applyQueuedEditLocally` retargets by press time and text, and the edit wire and the kernel's `editQueued` arms carry no id, so a client-only retarget cannot make the client and the kernel agree on which copy changed. That is the edit by id the #1224 body names as the natural continuation. render.ts is untouched: its two comments that describe the press-time-first order (above `mk`, and in the qx handler) are stale by one clause and can move with the edit-by-id change; #994 and #1126, both open, touch that file.

## Tests

- ui/webview/send-pending.test.ts, new test: two same-text entries share the press stamp; the ✕ carrying the stamp and the id drops the entry that owns the id; the same id again removes nothing, not the same-text neighbour; an id-less ✕ still names its entry by stamp and text and leaves a same-text neighbour pressed at another time. Before the change the first assertion fails, `dropPending` returns the first entry (1 fail, 38 pass in the file); after, 39 pass.
- tests/test_queued_copy_identity.py: the bad-id tuple in `test_the_handler_takes_a_fresh_id_in_the_echo_form_and_refuses_one_the_session_holds` gains `"echo:" + "f" * 32 + "\n"`. Before the change `_client_qid` returns it and the `assertIsNone` fails (1 failed, 24 passed in the module); after, 25 passed.
- tests/test_queued_copy_identity.py, new test `test_the_handler_refuses_a_fresh_id_when_the_hold_cannot_be_checked_and_names_the_sid_only`: on a real SdkBackend, a `pending_queued_meta` that raises, then one that answers `[]` with a `live_atoms` that raises, each refuse a fresh id; the stderr line names the sid and neither the id, the message text nor the fault's text; once both reads answer, the same id is taken. Green before the change: a guard for the branch.
- Full runs on the branch: pytest 10833 passed, 41 skipped; `npm run typecheck` clean; `npm test` 4053 pass, 0 fail.

Tier: fix

### DIFF
--- a/docs/read-side.md
+++ b/docs/read-side.md
@@ -258,10 +258,13 @@ it with the send, the kernel parks the copy under it (the parked op's fourth slo
 or queues it under it, and the ✕ names it, so the kernel cancels exactly the copy
 the bubble stands for, never a same-text neighbour by index or body. That holds
 wherever the copy carries the id: a parked send on any backend, and the SDK
-route's queue. A copy without one (the tmux route's queue, which the CLI holds)
-is still cancelled by index and body. Every copy the kernel queues itself (mail,
-a nudge, a re-delivery) is still minted an id where it enters the backend's
-queue. The CLI extracts
+route's queue. A copy whose ✕ names no id (an op the kernel parked itself, such
+as a nudge or a re-delivery; a ✕ from an older client) is still cancelled by
+index and body. The tmux route's queue, which the CLI holds, has no ✕ at all:
+`TmuxBackend` has no `unqueue`, so `build_session` ships those copies
+`cancelable: false` and the chat draws no ✕ for them. Every copy the kernel
+queues itself (mail, a nudge, a re-delivery) is still minted an id where it
+enters the backend's queue. The CLI extracts
 no image paths on the stream-json route (its only image-path test belongs to the
 interactive composer's paste handler), so an image path in an SDK send lands as
 typed and the echo's text matches. `_path_bearing` and the extension set it tests
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -28683,7 +28683,8 @@ def _is_slash_command(text):
 # "echo:" + hex, so isKernelEchoUuid on the client, the landed-record stamp's echo skip (build_session), the
 # re-queue's prefix test (sdk_backend _enqueue_with_id) and the echo-in-queue reading (_echo_queued_in) all
 # treat it as the kernel's. Bounded: the kernel mints 32 hex digits (uuid4().hex); a client id is admitted in
-# the same shape, never an arbitrary string that would ride the wire, the mirror and every chip.
+# the same shape, never an arbitrary string that would ride the wire, the mirror and every chip. Matched whole
+# (fullmatch): `$` alone admits a trailing newline, and an id with one is held nowhere, so it would be taken.
 _CLIENT_QID_RE = re.compile(r"^echo:[0-9a-f]{16,64}$")
 
 
@@ -28691,7 +28692,7 @@ def _wire_qid(msg):
     """The copy id a ws message names (`qid`), or None when it carries none or one in another form. A cancel's
     id only has to be looked up (an unknown id is the honest miss), so the shape is all a cancel checks."""
     q = msg.get("qid") if isinstance(msg, dict) else None
-    return q if isinstance(q, str) and _CLIENT_QID_RE.match(q) else None
+    return q if isinstance(q, str) and _CLIENT_QID_RE.fullmatch(q) else None
 
 
 def _client_qid(msg, sid, be):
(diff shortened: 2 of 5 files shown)


ksarma:

Reviewed inline: `dropPending` now resolves a cancel by its id whenever it carries one. It takes the entry that owns the id. Failing that, it takes the entry a kernel copy of that id covered by text on the last push (`cover`, which `reconcilePending` only ever sets to a kernel-minted id). Failing both, it drops nothing, even when a same-text entry shares the press time. Only an id-less cancel falls to the old press-time-and-text and text-only readings. The `qx` handler in `render.ts`, the sole caller, passes both `data-qts` and `data-qid` and posts the same `data-qid` on the cancel. `routeUserMessage` and `adoptProvisional` mint one id per send and hand it to both the post and `registerOptimistic`. So the entry the client drops and the copy the kernel cancels agree by construction. A kernel copy's cancel carries no `qts` and takes the same path as before. Both id producers emit `echo:` plus lowercase hex with no whitespace, so `fullmatch` in `_wire_qid` rejects only malformed ids. The refusal test asserts the sid is in stderr and the id, the message text and the fault text are not. `TmuxBackend` defines no `unqueue`, `build_session` ships those copies `cancelable` false, and the chat draws no cancel for them, so the corrected docs sentence is true. Body and commit message match the diff; the 90 added lines scan clean.

Follow-ups, neither blocking:
- `ui/webview/render.ts:204`, 445, 4269 and 16652 still call the press time the entry's identity; the body names two of the four.
- `kernel/kernel.py:28722` logs the same line for an unreadable hold as for a held id. A fixed sid-only suffix would keep the test's privacy property.

Red-first confirmed by my own run at base 82dd9e9c with the head's test files. `send-pending.test`: 38 pass, 1 fail (the new test's first assertion; `dropPending` returned the first entry). `test_queued_copy_identity`: 24 passed, 1 failed (the trailing-newline id passed `_client_qid`). At the head: 39 pass, 25 passed, typecheck clean. CI: all eleven checks green at the head.

Review run by the maintainer automated review session; verified against head 6b7f4ae4.

ksarma:
Head note: the branch was updated with main, so the head moved from 6b7f4ae4 to 7cadd855. The new head is a mechanical merge of 6b7f4ae4 and 15ad1d26: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 7cadd855.

PR #1261, opened 2026-09-10
Held input echoes: a slash or skill command's echo retires on the transcript's command wrapper record

ksarma:
## Symptom

A slash or skill command typed into the chat composer with its arguments on the line below keeps its "sending" bubble after the turn has run and replied, on every client, until the bubble is dismissed by hand. After a kernel restart the send is delivered again and the command runs twice. Reported from a live dashboard.

## Cause

Every landing check compares the echo's typed text with a transcript record's text under one key, `echo_text_key` (outer whitespace stripped, nothing else), and a slash send's record never carries the typed text. Claude Code records the send as a wrapper (`<command-message>`, `<command-name>`, `<command-args>`; a skill adds `<skill-format>` and a second record with the skill body). A verbatim copy is not guaranteed: some CLI versions write a raw twin record beside the wrapper, which the event model already drops as an invocation echo, and the reported transcript carried none. The event model reads the wrapper as a command atom whose text is `/name args`, with one space between the name and the arguments.

So `/deploy`, a newline, `staging now` never met `/deploy staging now`, and four readers failed at once:

- the kernel's prune (`prune_live`, through `_merge_live_atoms`) kept the echo, so the chat showed the pending bubble;
- the thread's held count (`_comments_frame`, `_landed`) kept owing a reply;
- the backend's boot and spawn scan (`_text_landed`) reads the raw records, where the wrapper's text equals no typed text at all, so it called the send lost and re-queued it, and the next spawn ran the command again;
- the fed-copy pairing (`qids_for_landing`) left the landed event without the copy's id, so the chat's own pending bubble, which retires by id, stayed too.

Measured on the unchanged sources with the new module's fixtures: the backend scan found no slash send of any shape; the kernel prune retired a typed command only when its whitespace already equalled the parsed form (`/deploy staging now` on one line) and kept `/deploy  staging now` and `/deploy` + newline + `staging now`.

Carrying the composer's press id through to the record would identify the send exactly, but the transcript never records it. Matching the wrapper's parsed form is the practical route.

## Fix

`session_backend.command_text_key` is a second key rule for slash sends: the whitespace-delimited tokens joined by single spaces, applied only when the first token is command-name-shaped (a slash, then no further slash or whitespace; a message that starts with a path keeps the plain rule). The same words in the same order match; a different command or different arguments never do. `echo_keys` returns the plain key and, for a slash send, the command key; it is the one place the either-key rule is written. Both live in `session_backend.py` because that module already holds the shared key rule (`sdk_backend.py` imports `echo_text_key` from there so the scan can never find what the prune cannot retire).

Both sides key both ways:

- Kernel: `_atom_user_texts` adds the command key for a slash-shaped record text. `_echo_landed_in` asks for either key at `build_session`'s queued count, `_merge_live_atoms`' display dedup and `_tmux_echo_prune`; `_comments_frame`'s held count asks the same keys against per-atom sets built once per frame.
- Backend: `_landed_texts` reads a wrapper record as the `/name args` the event model shows (`_command_invocation`, shared with `msg_to_atom`, which built the same string inline before) and adds the command key to every text; `_text_landed`, `prune_live` and `qids_for_landing` ask for either key.

The `command` flag stays off a typed send's echo; this PR explains it rather than changes it. The flag marks the CLI's own feedback atoms (the streamed wrapper and stdout, the picker chip), which owe no landing and are never flagged lost or re-delivered. A typed slash send is a message the CLI takes and records, so its echo keeps the full lifecycle, and flagging it would hide its loss. A comment at the echo's construction in `send` says so.

The rejudging latch in `build_feed` is unchanged: it arms off the parsed reply turn, never the echo, so a stranded echo cannot pin it either way; its comment now points at the new key. The Codex backend is unchanged: its echo readers (`_rec_texts`, its `prune_live`) compare under the plain key alone, and no wrapper reading exists for its records. `docs/read-side.md` names the second key where it described the one-key rule.

## Tests

`tests/test_kernel_slash_echo_retire.py` is new: 19 tests (four of them iterate subtests) over the real code paths, with synthetic fixtures (an invented `/deploy` command, a private synthetic sid, the notes-api demo domain). On the unchanged sources pytest reports 19 failures (9 tests and 10 subtests) and 10 passes: the 7 negative and plain-text cases that pin what the rule must not do, and the 3 parents of failing subtests.

- `TheSharedKey` (4): the key rule on both modules' copies. Three fail before on `AttributeError` (`command_text_key` and `echo_keys` do not exist); `test_the_plain_rule_is_unchanged` passes before and after.
- `WrapperReadingMatchesTheEventModel` (2): `_command_invocation` agrees with the event model's parse on six wrapper shapes (both tag orders, no args, a name without its slash, multi-line args, padded args), and prose quoting a tag is not an invocation. Both fail before: the helper does not exist.
- `KernelRetire` (5): the wrapper record retires the typed echo through `_merge_live_atoms`, in both tag orders (fails before: the echo stays). Arguments typed with irregular whitespace (`/deploy`, newline, `staging  now`, against a record whose arguments read `staging  now`) land on both sides, the scan and the prune (fails before: the scan misses, the echo stays). Another command, other args, no args, a wrapper written before the send, and a plain send differing in whitespace all keep the echo (pass before and after).
- `BackendScanAgrees` (3): `_text_landed` finds the send in its wrapper record; the scan and the kernel prune give the same verdict on five records; `_mark_dropped_echoes` at spawn marks the echo `_landed` and re-queues nothing. Before: the scan finds nothing, both positive records read as unlanded on both sides, and the spawn scan re-queues the send.
- `LandingCarriesTheCopysId` (3): `qids_for_landing` pairs the wrapper's parsed text with the typed copy (fails before: `[None]`). Another command does not spend the copy, and a plain copy still pairs by exact text (pass before and after).
- `TmuxEchoAgrees` (2): `_tmux_echo_prune` on the parsed command atom's texts retires the typed echo (fails before); another command's atom keeps it.

Cases added to existing modules:

- `tests/test_kernel.py::ViewBuilder::test_tmux_slash_sends_echo_is_not_counted_as_queued_once_its_wrapper_record_lands`: `build_session`'s queued count for a busy tmux-owned session, a typed `/deploy`, newline, `staging now` echo, and its wrapper record landing in the same second. Fails before: the typed text is listed under the queued indicator. That count runs only for a backend without `unqueue`, so none of the SDK-side cases reach it.
- `tests/test_comment_threads.py::ThreadProjection::test_a_slash_sends_echo_is_landed_by_its_wrapper_record`: the thread's held count reads the landing off the wrapper record, and another command's record is not this send's landing. Fails before: `queued` stays 1 after the record.
- `tests/test_queued_copy_identity.py::TheChatCarriesTheIds::test_a_slash_sends_wrapper_record_carries_the_typed_copys_id`: a real `send` through the backend, then the wrapper and skill-body records on disk. The landed event carries the typed copy's id and the echo leaves the live store. Fails before: the landed event's `qid` is None.
- `tests/test_kernel_fed_echo_absorbed.py::OneTextRuleAcrossKernelAndBackend`: three slash-shaped records join the scan-vs-prune parity pin (a command with irregular whitespace, two command blocks in one record, a path-leading message). This pin passes before and after; it holds the two sides to the same key set. A raw wrapper record is not in its list by design (the kernel keys the parsed command atom, the scan derives `/name args` from the raw record itself; `BackendScanAgrees` pins that agreement through the real parse), and the class docstring says so.

Each key site is held by a test that goes red when that site alone is reverted to the plain key: the queued count by the `test_kernel.py` case, the record-side key in either module by the irregular-whitespace case, with the parity pin excluded from the run.

Full suite: `pytest -q -n 6 tests/` on this head (tree `ca0759f5`): 10856 passed, 118 skipped, 0 failed, 1705 subtests passed (99 s, xdist 6 workers).

Tier: fix

### DIFF
--- a/docs/read-side.md
+++ b/docs/read-side.md
@@ -283,9 +283,14 @@ reads the whole file. The found verdict is recorded on the echo (`_landed`), and
 `prune_live` and the chat merge retire the echo on it without a text match, so a
 found echo always has an exit and a later boot never re-scans it. Every by-text
 comparison of an echo against a record (the guard's scan, `prune_live`'s retire, the
-kernel's `_atom_user_texts` and its folds, and the tmux echo's prune
-`_tmux_echo_prune`) uses one key, `echo_text_key` in `session_backend.py`: outer
-whitespace stripped, nothing else.
+kernel's `_atom_user_texts` and its folds, the fed-copy pairing `qids_for_landing`, and
+the tmux echo's prune `_tmux_echo_prune`) uses the two keys in `session_backend.py`:
+`echo_text_key` (outer whitespace stripped, nothing else) and, for a slash send,
+`command_text_key` (the tokens joined by single spaces). The second exists because the
+CLI records a slash or skill command as its `<command-name>` wrapper, which parses to
+`/name args` with one space whatever the sender typed between the name and the
+arguments; both sides key a slash-shaped text both ways, so the scan and the prune
+agree on the same records.
 
 **The ledger is a table of contents** (pure projection of captions + archive):
 - top: the archiver's one-sentence headline for the session,
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -13704,11 +13704,12 @@ def _comments_frame(sid, tmux=None):
                 # deliberate re-send repeat earlier texts, which read as "already in the transcript" and hid a
                 # send the CLI still held (round-5 review). A `dropped` echo (the backend adjudicated the send
                 # LOST — a reconnect with it in flight; the popover shows "never delivered") owes nothing.
-                user_atoms = [a for tr in turns for a in (tr.get("atoms") or []) if a.get("type") == "user"]
+                user_atoms = [(float(a.get("t") or 0), set(_atom_user_texts(a)))       # (stamp, its texts), built once per frame
+                              for tr in turns for a in (tr.get("atoms") or []) if a.get("type") == "user"]
                 def _landed(e):
-                    et = sb.echo_text_key(e.get("_echo_text"))
+                    keys = set(sb.echo_keys(e.get("_echo_text")))     # the plain key and, for a slash send, its words
                     since = float(e.get("t") or 0)                     # the send's own stamp: the record the CLI writes for
-                    return any(et in _atom_user_texts(a) for a in user_atoms if float(a.get("t") or 0) >= since)   # it is at or after it
+                    return any(t >= since and not keys.isdisjoint(texts) for t, texts in user_atoms)   # it is at or after it
                 floor = _human_turn_floor({"turns": turns}) if turns else 0
                 # `_landed` on the atom: the backend's boot/spawn scan read the landing off the transcript
                 # itself (sdk_backend._mark_dropped_echoes) — delivered, so nothing is held for it
@@ -29896,7 +29897,14 @@ def _atom_user_texts(a):
     sdk_backend.prune_live floors no echo (2026-09-06; before that the floor was narrowed to path-bearing
     echoes, 2026-07-20), so a genuinely dropped send stays visible. Per-block EXACT match, never a
     substring test: a bundled block is the very string that was
-    echoed, so this retires the delivered nudge without ever guessing about containment."""
+    echoed, so this retires the delivered nudge without ever guessing about containment.
+
+    A slash-shaped text also yields its command key (sb.command_text_key: the tokens joined by single
+    spaces). The CLI records a slash or skill send as a wrapper record, which the event model reads as a
+    command atom whose text is "/name args" with one space, whatever the sender typed between the name and
+    the arguments; the typed echo meets that atom under the command key whatever whitespace it carried
+    (2026-09-10). The backend's _landed_texts adds the same key to the raw records its landing scan reads,
+    so the two agree."""
     if a.get("type") != "user":
         return ()
     out = []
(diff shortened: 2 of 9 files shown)


ksarma:

Reviewed inline: `command_text_key` applies only when the first whitespace-delimited token is a slash with no further slash or whitespace. A path-leading message therefore keeps the plain rule. The only widening over the plain rule is whitespace-insensitivity for slash-leading texts. A record with different words never retires a typed echo. Both sides key both ways. The kernel's `_atom_user_texts` adds the command key to slash-shaped record texts. `_landed_texts` reads a wrapper record as the `/name args` string through `_command_invocation`. That is the `msg_to_atom` parse moved verbatim, so the live command atom's text is unchanged. `_text_landed` matches the echo's key set against each record at or after the send stamp. So a wrapper record marks the echo landed, and `_mark_dropped_echoes` re-queues nothing. A slash send with no wrapper record still reads as lost and is re-queued. Two identical sends retire on one record, as identical plain sends already did. `qids_for_landing` still pairs one copy per landing. `_comments_frame` builds its per-atom key sets once per frame. `docs/read-side.md` names both keys and the same readers the code has. Fixtures are synthetic. The added lines scan clean.

Follow-up, not folded: the `echo_text_key` docstring (`kernel/session_backend.py:34`) and the `kernel/codex_backend.py:43` comment still describe a single shared rule. Comment-only.

Test evidence: at the base with the head's test files, `tests/test_kernel_slash_echo_retire.py` reports 19 failed, 10 passed, 7 subtests passed. The three named cases in `test_kernel.py`, `test_comment_threads.py` and `test_queued_copy_identity.py` fail there too. The fed-echo parity pin is green there by design. At the head, the five modules report 452 passed, 17 subtests passed, and `test_event_model_golden.py` 110 passed. CI: all 11 checks SUCCESS at fac32b35.

Review run by the maintainer automated review session; verified against head fac32b35.

ksarma:
Head note: the branch was updated with main, so the head moved from fac32b35 to c34eba76. The new head is a mechanical merge of fac32b35 and 15ad1d26: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head c34eba76.

PR #1262, opened 2026-09-10
Tests: a lab kernel starts from a list of names, never a copy of the runner's environment, with a postal bus of its own that is never started

ksarma:
Builds on #1235 (merged), which narrowed what a served lab writes to its cfg.json for the driver's relaunch of the kernel (`relaunch_env`). This change narrows what the lab's own kernel starts with, one layer out, and gives every lab kernel a postal bus of its own.

## Symptom

Run the browser tests from a shell on a machine that is running romp (a session's shell, or one that carries the kernel's exports) and the hermetic lab kernels misbehave in three ways. A lab kernel exits in the middle of a test when the live manager restarts: it inherited ROMP_MANAGER_PID, which the kernel's `_parent_watch` treats as its own supervisor. A lab kernel binds where the live kernel serves (ROMP_SERVE_HOST), which need not be loopback. And each lab kernel's boot-time `romp-postal-service ensure` runs against the machine's bus port: on a machine with a bus, the lab kernel dials the live bus from then on; on one without, ensure starts a detached bus that no tearDown stops. The remote-comment lab set ROMP_POSTAL_PEERS=0 with a comment saying it stays off the machine's bus; that value only selects the legacy singleton scheme, and ensure still started a bus on the default port.

## Cause

Sixteen served test modules built their lab kernel's environment as `dict(os.environ, ...)` with the lab's roots and seams over it, popping ROMP_STATE_DIR and, in some, two key names, and the ship lab's `_ShipLab.kernel_env` did the same. Everything else in the runner's environment reached the lab kernel: the live kernel's ROMP_MANAGER_PID, ROMP_SERVE_HOST, ROMP_SID and ROMP_SESSION_NAME, its ROMP_POSTAL_PORT (or, with none, the default 25302), and whatever else the shell carried. `postal_service.is_client_only` returns False whenever `peers_on()` is true, and ROMP_POSTAL_PEERS=0 alone leaves ROMP_POSTAL_CLIENT_ONLY unset, so `ensure()` fell through to starting a bus.

## Fix

One function, `kernel_env` in tests/test_ship_reship.py beside `relaunch_env`, builds every lab kernel's environment from names. From the runner it takes PATH and HOME, the XDG_* names (kernel/credentials.py resolves the service.env default under XDG_CONFIG_HOME) and, of what tests/conftest.py sets for every child of the run, TMPDIR and TMUX_TMPDIR, GIT_CONFIG_GLOBAL and GIT_CONFIG_NOSYSTEM, and the four ROMP_ names of its floor the lab does not set itself (ROMP_SERVICE_ENV_FILE and ROMP_SERVICE_ENV, ROMP_CLAUDE_BIN, ROMP_CLI_SCOPE). Over those go the lab's roots and seams, any seams a lab adds by keyword, and a postal bus of its own: ROMP_POSTAL_PORT at a free port with ROMP_POSTAL_PEERS=0 and ROMP_POSTAL_CLIENT_ONLY=1 (both names, because client-only applies in the legacy scheme alone), so the kernel's boot-time ensure pings the private port and returns without starting anything. The seventeen modules that boot a kernel call it; their pops of ROMP_STATE_DIR and the key names go, since nothing beyond the list is taken. `relaunch_env` is unchanged; the probe the served labs plant for its cfg.json check is now planted in the lab kernel's environment directly, since the runner's no longer reaches it.

## Tests

- `LabKernelEnv` in tests/test_ship_reship.py (no kernel, no browser, runs everywhere) plants in os.environ a live kernel's exports (ROMP_MANAGER_PID, ROMP_SERVE_HOST, ROMP_STATE_DIR, ROMP_SID, ROMP_SESSION_NAME, ROMP_SUPERVISED), a synthetic RUNNER_SECRET_PROBE, the conftest floor, an XDG_CONFIG_HOME of the runner's and the machine's bus port, then builds a lab kernel's environment. Three tests: none of the exports or the probe is present, the floor, the runner's XDG_CONFIG_HOME and the lab's own names are, and every ROMP_ name is the lab's or the floor's; ROMP_POSTAL_PORT is a port of its own, with the runner naming the machine's bus and with the runner naming none, next to ROMP_POSTAL_PEERS=0 and ROMP_POSTAL_CLIENT_ONLY=1; a lab's own seams go over the names, including a name kernel_env sets for every lab (ROMP_KERNEL_NO_OPEN) and one of the floor (ROMP_CLI_SCOPE). Before the change all three fail: the module-level function is absent, and run against the earlier `_ShipLab.kernel_env` they fail on ROMP_MANAGER_PID present, ROMP_POSTAL_PORT equal to 25302, and the keyword seams rejected. Each clause of the helper has a case that fails without it: dropping the XDG_* pass-through fails the first test, and applying the seams before the lab's names fails the third.
- `RelaunchEnv` keeps its assertions and now plants its probe in the lab kernel's environment, as the served labs do.
- The seventeen kernel-booting modules ran locally against Chromium, serially: 43 passed, no skips. Under parallel workers their dist copies can hit the race `copy_dist` (#1238) fixes on main after this branch's base; it runs before `kernel_env` is called.
- A served lab whose kernel never serves /healthz skips rather than fails, so a module that went back to copying the runner's environment would read as skips on a machine running romp, not as a failure; `LabKernelEnv` is the executing check, on the helper itself.
- Full suite on this tree (six workers): 10774 passed, 41 skipped, 1 failed: `tests/test_ship_reship.py::ServedWedge::test_restart_between_ship_and_ack_reships_heals_and_releases_the_held_send`, the Chromium restart-heal leg, on `reloadHeldWhileBusy` None with the kernel healed and the reload fired. It passes alone on this tree (4 of 4 runs) and in the serial run above, and it is known to fail under load on main without this change; the environment path it drives (boot, kill, relaunch from the written env, heal) succeeded.

Tier: docs

### DIFF
--- a/tests/test_awaiting_box_sync.py
+++ b/tests/test_awaiting_box_sync.py
@@ -28,6 +28,7 @@ import re
 import shutil
 import socket
 import subprocess
+import sys
 import tempfile
 import time
 import unittest
@@ -37,6 +38,9 @@ HERE = os.path.dirname(os.path.realpath(__file__))
 ROOT = os.path.dirname(HERE)
 BIN = os.path.join(ROOT, "bin")
 EXT = os.path.join(ROOT, "vscode-extension")
+sys.path.insert(0, HERE)
+import test_ship_reship as _lab   # noqa: E402  the lab kernel's environment (the module, not its classes: an
+#                                   imported TestCase would be collected here a second time)
 
 RENDER = open(os.path.join(ROOT, "ui", "webview", "render.ts")).read()
 
@@ -162,11 +166,7 @@ class ServedSync(unittest.TestCase):
                                     "stop_reason": "end_turn"}}) + "\n")
         cls.port = _free_port()
         cls.token = "testtok-awaitsync"
-        env = dict(os.environ, XDG_STATE_HOME=os.path.join(cls.lab, "xdg"), CLAUDE_CONFIG_DIR=claude,
-                   ROMP_MANAGER_PORT="1", ROMP_KERNEL_NO_OPEN="1", ROMP_SERVE_TOKEN=cls.token,
-                   ROMP_KERNEL_PORT=str(cls.port), ROMP_DIST_DIR=dist,
-                   ROMP_MODEL_CATALOG="off")   # hermetic: never reach the network
-        env.pop("ROMP_STATE_DIR", None)
+        env = _lab.kernel_env(cls.lab, claude, dist, cls.port, cls.token)
         cls.klog = os.path.join(cls.lab, "kernel.log")
         cls.kernel = subprocess.Popen([os.path.join(BIN, "romp-kernel")],
                                       stdout=open(cls.klog, "w"), stderr=subprocess.STDOUT, env=env)
--- a/tests/test_dashboard_reload_served.py
+++ b/tests/test_dashboard_reload_served.py
@@ -34,8 +34,8 @@ SID = "11111111-2222-4333-8444-000000000201"
 
 import sys
 sys.path.insert(0, HERE)
-import test_ship_reship as _lab   # noqa: E402  the served lab's cfg.relaunch stanza (the module, not its classes:
-#                                   an imported TestCase would be collected here a second time)
+import test_ship_reship as _lab   # noqa: E402  the lab kernel's environment and the cfg.relaunch stanza (the module,
+#                                   not its classes: an imported TestCase would be collected here a second time)
 
 
 def _free_port():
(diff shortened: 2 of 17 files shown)


ksarma:

Reviewed inline: `kernel_env` builds a lab kernel's environment from `KERNEL_ENV_NAMES` plus the `XDG_*` names, then the lab's roots and seams, then the caller's keyword seams. `KERNEL_ENV_NAMES` is `CLAUDE_CONFIG_DIR`, `PATH`, `HOME`, `TMPDIR`, `TMUX_TMPDIR`, the two `GIT_CONFIG` names, `ROMP_SERVICE_ENV_FILE`, `ROMP_SERVICE_ENV`, `ROMP_CLAUDE_BIN` and `ROMP_CLI_SCOPE`. `ROMP_MANAGER_PID`, `ROMP_SERVE_HOST`, `ROMP_SID`, `ROMP_SESSION_NAME`, `ROMP_SUPERVISED`, `ROMP_STATE_DIR`, the runner's `ROMP_POSTAL_PORT` and every key name stay out. All sixteen served modules and `_ShipLab` pass `(lab, claude, dist, port, token)` in signature order, from the variables the removed dicts used. The remote-comment and reload labs keep their extra names as keyword seams. No module builds a kernel environment from `os.environ` any more. The remaining `dict(os.environ)` calls are the node driver's, unchanged. The kernel's boot-time `ensure` subprocess inherits the lab environment. `postal_service` reads `ROMP_POSTAL_PORT` at import. With `ROMP_POSTAL_PEERS=0` and `ROMP_POSTAL_CLIENT_ONLY=1`, `ensure()` returns `ping()` before any `Popen`. Every other bus dial in the kernel uses the same `BUS_PORT`. `relaunch_env` carries the private port and both flags into the relaunch stanza. The only kernel path under `XDG_CONFIG_HOME` is `credentials.service_env_path`, floored by `ROMP_SERVICE_ENV_FILE`. `HOME` passing through is pre-existing and unchanged here. A lab kernel can therefore still read the operator's `~/.config/romp` knobs and `~/.ssh/config` on the paths that resolve them. That is a follow-up, not a defect of this PR.

Ran locally: `LabKernelEnv` and `RelaunchEnv` at head, 4 passed, 9 deselected. At base 6f2238b5 the three `LabKernelEnv` tests fail against the old `_ShipLab.kernel_env`: `ROMP_MANAGER_PID` present, port 25302, keyword seam rejected. They also fail with the module-level function absent. Both mutation claims hold. The served legs were not run here, by rule: they start and kill real kernels, and CI has no playwright. Their call sites are checked by reading. The serial 43-passed run is the author's evidence. CI: all 11 checks green at head.

Review run by the maintainer automated review session; verified against head 72799667.

ksarma:
Head note: the branch was updated with main, so the head moved from 72799667 to bcea2e7e. The new head is a mechanical merge of 72799667 and 15ad1d26: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head bcea2e7e.

PR #1263, opened 2026-09-10
Timeline: a live lane whose goal store faulted is derived, not served from the memo

ksarma:
Follow-up to #1241, from its review: the finding that a faulted goal store and a store with no file shared one key in the live-lane memo.

## Symptom

A live lane whose goal store could not be read and was then removed, or whose missing store appeared unreadable, kept the wrong judging marks in its band until its transcript, captions, archive or store moved: the archiver's and the captioner's marks stayed missing after the store was gone, or stayed shown while it could not be read. One lane's judging band only; healed by the next transcript append.

## Cause

`_lane_memo` keyed a faulted store (`goals` None, after `build_timeline` complained) and a store with neither seams nor nodes both as `"empty"`, on the reasoning that the two fields read are empty either way. But `_lane_segments` derives no marks at all for None (`_derive_judging_marks` is not called), while for the empty store it still derives the captioner's marks from the captions file and the archiver's from the archive. A lane held across the transition hit the entry the other side had stored and was served its marks.

## Fix

In `_lane_memo` (kernel/kernel.py), a faulted store is a skip: derived and not held, counted under `complain_skip` (the goals stage complained, as `_bars_complain` says; the counter set on `/perf` `memos.lanes` is unchanged). Its key stays the empty store's, and a skip never compares or stores a key, so the shared key no longer reaches the memo; the comment on that arm, the input comment above `_lanes_memo` and the function's docstring say so. The dead-lane path already derives a faulted lane every build and never caches it; the live path now does the same. docs/reference.md's description of `complain_skip` (the parse or a stage failed) already covers it.

## Tests

Two on the `LaneMemoBase` fixture in tests/test_timeline_lane_memo.py (`NotHeld`), both faulting the store through the real loader (an unreadable goals file), with an archive written so the two derivations differ:

- `test_a_faulted_store_is_derived_and_not_held`: the store unreadable, then removed. Before the change the faulted build counted a miss and stored an entry ({'miss': 1, 'complain_skip': 0}), and the store-less build was served it with no archiver mark ([] where ['retry loop'] is derived). After: one `complain_skip`, no entry, and the store-less build derives the archiver's mark.
- `test_a_held_lane_whose_store_faults_is_derived`: a lane held with no store, whose store then appears unreadable. Before the change the faulted build was served the entry's archiver mark. After: derived without marks, one `complain_skip`, the entry stands, and once the store reads as absent again the entry serves a frame equal to the first build's derivation (the lane's judging and turns compared as JSON).

Full pytest on this head (`pytest -n 6 tests/`): 10836 passed, 118 skipped, 0 failed, 1688 subtests passed.

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -36482,9 +36482,10 @@ def _dead_lane_marks(marks, t0):
 #             object every build (live_tail).
 #   goals     the store the loop reads seams from (_segs_seam) and nodes from (_derive_judging_marks): a FrozenStore
 #             by identity (load_goals_shared serves one per file version, so a publish or a journal append is a new
-#             object); a store with neither seams nor nodes, or None after a fault, as "empty", since the two
-#             fields read are empty whatever its identity; any other private store is not held (unshared_skip: a
-#             mutable store could change under the entry).
+#             object); a store with neither seams nor nodes as "empty", since the two fields read are empty
+#             whatever its identity; any other private store is not held (unshared_skip: a mutable store could
+#             change under the entry). None after a fault is derived and not held (complain_skip): the derivation
+#             skips the marks for it, so it is not the empty store, and a skip compares and stores no key.
 #   captions  _stat_key of captions/<sid>.jsonl, taken by build_timeline BEFORE _captions reads the file (_captions
 #             builds a new dict per call, so the file is the input). Stat before read: a row appended between the
 #             two is read by this build and held under the OLD key, so the next build's stat misses and derives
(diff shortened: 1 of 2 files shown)


ksarma:

Reviewed inline: in `_lane_memo` the faulted-store skip is decided before the memo get. The hit branch requires no skip, and the put requires a plain miss. So the key still computed for a `None` store is never compared or stored. The counter increments once per faulted build, under the lock. The standing entry is popped only when the parse object moved. A held lane therefore survives a fault episode and serves again once the store reads as absent. That entry is the empty store's derivation on the same parse, captions and archive, so it equals a fresh one. The second test checks that equality as JSON. The dead-lane path assigns no lane key when the goals read faulted, so it never serves or caches such a lane. The live path now matches. The delta entry memo reuses an encoding only for the same entry object. The skeleton is a shallow copy of the cached build. So no third path holds a per-lane frame. The `/perf` counter set is untouched. The `complain_skip` wording in `docs/reference.md` covers the goals stage. The comment rewrites match the code. The body and commit message match the diff. The added lines carry no private strings, hostnames or home paths.

Follow-up, `tests/test_timeline_lane_memo.py:1041`: the new tests fault the store with `chmod(0)` and have no root skip. Under uid 0 they would fail rather than skip. The module's earlier `chmod(0)` test at line 182 shares the gap. CI runs unprivileged.

Red-first by my own run: the head's test module at base 3bebe6ba fails exactly the two new tests, `NotHeld::test_a_faulted_store_is_derived_and_not_held` and `NotHeld::test_a_held_lane_whose_store_faults_is_derived`. The readings match what the body states (2 failed, 62 passed). At head the module passes 64/64 and `tests/test_perf_stats.py` 49/49. CI at head: all 13 checks green.

Review run by the maintainer automated review session; verified against head 947994e5.

ksarma:
Head note: the branch was updated with main, so the head moved from 947994e5 to 6e0faee6. The new head is a mechanical merge of 947994e5 and 15ad1d26: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 6e0faee6.

PR #1266, opened 2026-09-10
Tests: the raw-copytree guard reads each test module's call tree, and tests/README.md records the copy_dist rule

ksarma:
## Symptom

A served-page test class that copies the built `vscode-extension/dist/` with a plain `shutil.copytree` passes the guard in `tests/test_dist_copy_staging.py` when the call's arguments nest two levels deep (`os.path.join(os.path.dirname(EXT), "dist")`), when the function is imported under another name (`from shutil import copytree as ct`), or when the path sits in a name such as `DIST`. Under `pytest -n` such a class still errors in `setUpClass` on a concurrent build's staging file, which is what the guard exists to refuse before a parallel run lands on it. Separately, the conventions section of `tests/README.md` records the hermetic temp and git rules and the report redaction, each with the test that pins it, and not the `copy_dist` rule, so a new class's author learns it only from the guard's failure (the follow-up recorded on #1238).

## Cause

The guard matched each module's raw text against a regular expression that accepted one level of nested parentheses, anchored on the literal `copytree(`, and looked for `"dist"` case-sensitively. A second level of nesting breaks the match and `re.finditer` skips the whole call; an alias never carries the anchor; `DIST` fails the case.

## Fix

`raw_dist_copies(text, filename)` reads the module's call tree with `ast`: a call of `copytree` as an attribute of whatever it is reached through, of the name itself, or of an alias `from shutil import copytree as ...` binds is a hit when the source of any argument or keyword value names dist in any case, wherever the call sits and however its arguments nest; hits come in line order. The walk over `tests/test_*.py` is `raw_dist_copies_under(directory)`, whose roster names the module, the call's first line and the call as written. A module whose text lacks the token is not parsed, so the guard over the 626 test modules costs 0.10 s against the text scan's 0.03 s (one machine; the three modules that carry the token are parsed); text that carries the token and does not parse raises `SyntaxError` under the module's name (collection fails on such a module first anyway). A string holding a call is not a call, so the guard module's own synthetic snippets do not trip it. Case-insensitive `dist` is a choice: it catches a name-bound `DIST`, and no `copytree` call under `tests/` names anything else containing the letters. An assignment alias (`ct = shutil.copytree`) and a `getattr` call are outside the scan, and the docstrings say what it reads. `tests/README.md` gains one paragraph in the conventions section, in the section's form: the rule, the race behind it, and the test that pins it.

## Tests

- `tests/test_dist_copy_staging.py::EveryTestModuleCopiesTheBuiltDistThroughCopyDist::test_the_scan_reads_a_nested_call_an_aliased_import_and_a_module_alias` (new) runs `raw_dist_copies` over synthetic modules assembled at run time. Nine raw shapes (one-level, two-level, multi-line, keyword-argument, destination-only, aliased import, star import, module alias, name-bound `DIST`) each yield one hit at the call's line; five clean ones (a `copy_dist` call, a `copytree` without dist, one with `ignore=`, a `mock.patch.object` naming `copytree`, a string holding a call) yield none; the roster's source is the call as written; a call inside a function and a module-level call below it come back in line order; text without the token is not parsed (`def (` yields `[]`), and text with it that does not parse raises `SyntaxError` carrying the given filename. Before the change: `NameError` on `raw_dist_copies`. With the regex scan extracted under that name: the two-level, aliased-import and name-bound modules yield `[]` where `[2]` is expected, the string yields `[1]`, and no `SyntaxError` is raised.
- `test_the_walk_names_the_module_line_and_call_of_a_raw_copy` (new) writes a routed module and a module with a two-level plain call inside `setUpClass` into a scratch directory and reads `raw_dist_copies_under` over it: the roster is exactly `test_served.py:6: <the call>`. Under the regex scan the roster is empty.
- `test_no_test_module_copies_dist_with_a_plain_copytree` (existing, green before and after; the tree holds no raw copy). Mutations, each reverted: a two-level `shutil.copytree(os.path.join(os.path.dirname(...), "dist"), dist)` at `tests/test_ship_reship.py:293` and `from shutil import copytree as ct` with `ct(...)` at `tests/test_remote_comment_name_served.py:177` both pass the previous guard (traced with its regex) and are each named with file, line and call by the new one. Scanner mutations (keyword values dropped, first argument only, prefilter removed, walk never calling the scanner, hits unsorted, filename not passed to the parser, case-sensitive dist, from-import aliases ignored, bare-name calls ignored) each turn one of the new cases red.
- The module also passes under Python 3.10 and 3.14 via `unittest`; `tests/test_tempdir_hygiene.py`, which reads every test file, passes on the changed tree.
- Full run on this head (`pytest -n 6 tests/`): 10836 passed, 118 skipped, 0 failed, 1688 subtests passed

Tier: docs

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -108,6 +108,14 @@ the developer's git configuration (CI has none), and the env identity outranks
 exports its own `GIT_AUTHOR_*` after the floor. `tests/test_tempdir_hygiene.py`
 and `tests/git-hermetic.bats` pin all of it.
 
+**A served-page class copies the built `vscode-extension/dist/` with
+`tests.dist_copy.copy_dist`, never `shutil.copytree`.** Under `pytest -n` a
+sibling class's build renames or removes its staging files
+(`.<name>.tmp-<pid>-<n>`, `stagingPath` in `esbuild.js`) between the listing
+and the copy, and a plain copytree raises `shutil.Error` before the class's
+first test; `copy_dist` skips that shape. `tests/test_dist_copy_staging.py`
+pins the copy and refuses a raw copytree of dist in any test module.
+
 **No test report shows a process-environment value or a credential-shaped
 token.** An assertion whose container is an environment mapping prints the
 whole mapping when it fails (`assertNotIn("X", os.environ)` renders every
--- a/tests/test_dist_copy_staging.py
+++ b/tests/test_dist_copy_staging.py
@@ -14,10 +14,12 @@ copied; one removed after the listing (a stand-in for os.scandir removes it once
 taken) fails a plain copytree, the defect, and not copy_dist; and every name the script's own
 STAGING pattern matches is skipped while the served names are kept, a hidden name of another shape
 among them, the pattern read from the script so the copy's shape cannot drift from the script's
-unnoticed. A guard closes the module: no test module copies the built dist/ with a plain copytree,
-since the race fails no test of the class that does until a parallel run lands on it. All fixtures
-synthetic.
+unnoticed. A guard closes the module, reading each test module's call tree: none copies the built
+dist/ with a plain copytree, reached as an attribute, by the bare name or by an imported alias and
+however its arguments nest, since the race fails no test of the class that does until a parallel run
+lands on it. All fixtures synthetic.
 """
+import ast
 import glob
 import os
 import re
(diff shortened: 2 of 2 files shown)


ksarma:

Reviewed inline: `raw_dist_copies` parses a module only when its text carries the `copytree` token. It walks every `copytree` call reached as an attribute, by bare name, or by an alias bound by a from-import of `shutil`. It reports a call when the source of any positional or keyword argument contains `dist` in any case. Hits are sorted by line. Text with the token that does not parse raises `SyntaxError` under the module's filename instead of being skipped. `raw_dist_copies_under` opens and parses `tests/test_*.py` and never imports or runs a module. Under the old regex, extracted under the new name, the four cases behave as the body says. The two-level, aliased-import and name-bound shapes yield no hit; the string module yields a false one. The new scan reverses all four. An assignment alias, a `getattr` call, a wrapper with plain-named arguments, and a copy of a parent directory stay outside the scan by design. None exists in the tree, and `tests/dist_copy.copy_dist` is the only sanctioned route. The two other `copytree` calls under `tests/` name nothing containing the letters `dist`, and no tracked path has a `dist` component. The case-insensitive match therefore has no false positive today. The README paragraph matches `esbuild.js` and `tests/dist_copy.py` and sits in the conventions section in its form. Only `tests/README.md` and `tests/test_dist_copy_staging.py` change, so `docs` fits.

For the maintainers: open #1126 adds a one-level raw `copytree` of `dist` in a new served module, which both guards name. Main's guard goes red when #1126 merges unrouted.

Locally:
- `tests/test_dist_copy_staging.py` and `tests/test_tempdir_hygiene.py` at head: 25 passed, 63 subtests.
- The base module with the two new cases appended fails both with `NameError` and keeps the existing guard green.
- The guard costs 0.10 s over 626 modules, three of them parsed.
- The tree mutations at `test_ship_reship.py:293` and `test_remote_comment_name_served.py:177` pass the old regex. The new guard names each with file, line and call (scratch copy, parsed only).
- CI at head: eleven of eleven checks green.

Review run by the maintainer automated review session; verified against head 2859e03c.

ksarma:

Reviewed inline: `raw_dist_copies` parses a module only when its text carries the `copytree` token. It walks every `copytree` call reached as an attribute, by bare name, or by an alias bound by a from-import of `shutil`. It reports a call when the source of any positional or keyword argument contains `dist` in any case. Hits are sorted by line. Text with the token that does not parse raises `SyntaxError` under the module's filename instead of being skipped. `raw_dist_copies_under` opens and parses `tests/test_*.py` and never imports or runs a module. Under the old regex, extracted under the new name, the four cases behave as the body says. The two-level, aliased-import and name-bound shapes yield no hit; the string module yields a false one. The new scan reverses all four. An assignment alias, a `getattr` call, a wrapper with plain-named arguments, and a copy of a parent directory stay outside the scan by design. None exists in the tree, and `tests/dist_copy.copy_dist` is the only sanctioned route. The two other `copytree` calls under `tests/` name nothing containing the letters `dist`, and no tracked path has a `dist` component. The case-insensitive match therefore has no false positive today. The README paragraph matches `esbuild.js` and `tests/dist_copy.py` and sits in the conventions section in its form. Only `tests/README.md` and `tests/test_dist_copy_staging.py` change, so `docs` fits.

For the maintainers: open #1126 adds a one-level raw `copytree` of `dist` in a new served module, which both guards name. Main's guard goes red when #1126 merges unrouted.

Locally:
- `tests/test_dist_copy_staging.py` and `tests/test_tempdir_hygiene.py` at head: 25 passed, 63 subtests.
- The base module with the two new cases appended fails both with `NameError` and keeps the existing guard green.
- The guard costs 0.10 s over 626 modules, three of them parsed.
- The tree mutations at `test_ship_reship.py:293` and `test_remote_comment_name_served.py:177` pass the old regex. The new guard names each with file, line and call (scratch copy, parsed only).
- CI at head: eleven of eleven checks green.

Review run by the maintainer automated review session; verified against head 2859e03c.

PR #1268, opened 2026-09-10
Tunnel rows: a Start whose step raises releases the row's hold instead of leaving it set until a kernel restart

ksarma:
## Symptom

A Start on a no-kernel tunnel row (the popover's button) whose update or boot step raised an exception left the row stuck until a kernel restart. The row kept reading "starting" with "updating + starting the kernel" as its detail, never returned to up or no-kernel, showed no Start button on any later render (the click handler's Retry relabel lasts only until the next refresh, which rebuilds the list and renders Start only on no-kernel rows), kept the popover's 600 ms fast poll running for as long as the label read starting, and its recovery counter (upSeq) stayed still. The click itself got the generic alert "Start on <host> failed to reach the kernel." with nothing naming the failure. A tunnel death meanwhile relabelled the row down or error, still with no Start button; a kernel restart's load reset was the only way back.

No live raise path is known: every ssh seam in both legs catches Exception and returns a value. My merge notes on #1239 recorded the missing guarantee as pre-existing, so this PR closes a gap, not a reproduced outage.

## Cause

`_start_remote` sets the row's hold (`booting`, with status "starting") and, before this change, released it on two exits only: every planned failure returns through the nested `_fail`, which writes no-kernel with the failure as the detail and clears the flag, and the success tail clears it after the token refresh. An exception escaping any step between them left the hold set, and nothing else in the process clears it; the load-time reset runs only across a kernel restart. Three readers defer to the hold, which is why the row froze rather than healed: the supervisor pass skips its status write and the no-kernel hint while it stands, so the poll's no-kernel never reached the row, and since #1239 `_note_recovery` skips its bump. The route (`POST /tunnels/start`) has no exception handler of its own, so `do_POST`'s catch-all answered 500 with a text/plain traceback, which the popover's `r.json()` cannot parse, hence the generic alert.

## Fix

The body of `_start_remote` after the hold is set runs in a `try`; an `except BaseException` routes the exception through `_fail`, so the row reads no-kernel with "Start on <host> failed unexpectedly (<type>): <message>" as its detail (the message stripped and cut at 160 characters, and the ": <message>" tail omitted when the exception carries none), and re-raises, so the route's answer is unchanged. Two choices in it:

- Through `_fail` rather than clearing the flag alone. With only `booting` cleared, the next pass writes no-kernel but keeps the stale "updating + starting the kernel" text, because the pass keeps a specific parked failure over its generic hint until the row heals (its 2026-07-11 rule, so a refused update's reason stays visible). Parking the failure text makes the row read like every other failed Start, and the row's detail is where the user reads what went wrong, since the alert cannot show it.
- `BaseException`, so the release has a `finally`'s coverage: a SystemExit or KeyboardInterrupt in the handler thread releases the hold too, and the detail names the exception type. The alternative is `except Exception` plus a `finally` that clears the flag, at the cost of the failure text.

The ordinary returns and the success tail are unchanged, re-indented into the try (`git diff -w` shows the added lines alone). The `_fail` call runs after every `with _remotes_lock` block in the body has exited, so it takes the plain lock safely. One docstring sentence. No change to `_note_recovery`, the supervisor, the route or the popover.

## Tests

tests/test_kernel_remote_start.py, class StartRemote (synthetic host, every ssh seam stubbed, run under pytest). Five executing tests over the real `_start_remote`, each red on the unchanged kernel:

- `test_an_escaped_exception_releases_the_hold_and_parks_the_failure`: a stand-in update step raises `RuntimeError("stand-in step failed")`. The exception propagates, the hold is released, the row reads no-kernel with the exact detail "Start on TESTHOST failed unexpectedly (RuntimeError): stand-in step failed", and the boot leg never ran. Before the change: `AssertionError: True is not false : the hold must not outlive the call`.
- `test_the_recovery_counter_moves_on_the_first_answered_pass_after_an_escaped_exception`: the same raise, then `_note_recovery(r, "up")` bumps upSeq to 1. Before the change: `False is not true : no hold left standing to freeze the recovery counter` (the #1239 guard returns early under the hold).
- `test_an_escaped_exception_in_the_boot_leg_releases_the_hold_too`: the update leg reports up to date and a stand-in boot step raises a `RuntimeError()` with no message; the hold is released, the row reads no-kernel, and the detail ends at the type, "Start on TESTHOST failed unexpectedly (RuntimeError)". Before the change: `True is not false` on the hold.
- `test_an_exit_that_is_not_an_exception_releases_the_hold_too`: the stand-in update step raises `SystemExit("stand-in exit")`; it propagates, the hold is released, the row reads no-kernel and the detail names the type. Red under `except Exception` as well as before the change: `True is not false : the hold must not outlive a non-Exception exit either`.
- `test_a_long_failure_message_is_stripped_and_cut_in_the_parked_detail`: a 400-character message with surrounding whitespace is parked stripped and cut at 160 characters. Before the change: the detail still reads "updating + starting the kernel".

Each detail assertion is an exact match, so dropping the type name, the message, the strip or the cap from the parked text fails a test. Module: 20 passed (15 before, plus these five). tests/test_kernel_tunnel_truth.py (the #1239 tests): 33 passed, unchanged. Full pytest on this head: 10840 passed, 118 skipped, 1688 subtests passed, 0 failed (96 s, -n 6).

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -21634,7 +21634,9 @@ def _start_remote(host):
       2. If the kernel still isn't answering (the up-to-date path), _start_remote_kernel + the same
          port wait attach's bootstrap uses.
     Refreshes the stored token after boot (a first-ever kernel just wrote its serve-token).
-    Returns (ok, detail); mirrors _update_remote's contract."""
+    Returns (ok, detail); mirrors _update_remote's contract. An exception that escapes any step after the
+    hold is set releases the hold through _fail (the row reads no-kernel, the failure as its detail) and
+    propagates."""
     host = str(host or "").strip()
     if not host:
         return False, "no host"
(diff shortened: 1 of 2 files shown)


ksarma:

Reviewed inline: every step between the hold and the two planned exits, the success tail included, runs inside the `try`. The hold is set inside the first `_remotes_lock` block, the nested `_fail` follows, and the `try` is the next statement. `git diff -w` shows only the added `try:`, the except arm and one docstring sentence, so no logic moved. The except arm runs unlocked, `_fail` takes the lock safely, and no other lock is involved. `_remotes_lock` is a plain `threading.Lock` used only through `with` blocks, which release on exception. The re-raise is a bare `raise`, so SystemExit and KeyboardInterrupt propagate as before. `do_POST`'s catch-all still answers a text/plain 500, and the popover's alert is unchanged. The parked detail does not start with "no kernel answering", so the supervisor keeps it under the 2026-07-11 rule until the row heals. `_note_recovery` bumps upSeq on the first answered pass after release. The detail is `str(e)` cut at 160. It can carry an ssh argv, the host alias or a local path, never the serve-token.

Two notes for the record, neither a change request:
- `kernel/kernel.py:21689` evaluates `str(e)` before `_fail`. An exception whose `__str__` itself raises would still leave the hold set. No current class does; a `try/finally` would close it.
- A follow-up could have `/tunnels/start` answer JSON on an escaped exception. The click's alert would then name the failure instead of the generic text.

Red-first by my own run: the head's test file over base cfc44a4d fails exactly the five new StartRemote tests (5 failed, 15 passed) and passes at head. At head, `tests/test_kernel_remote_start.py` 20 passed and `tests/test_kernel_tunnel_truth.py` 33 passed. The `except Exception` mutation, run in a scratch copy, fails only the SystemExit case. CI at head: all 11 checks green. Privacy scan of the 104 added lines: TESTHOST only.

Review run by the maintainer automated review session; verified against head 0b6d90dc.

ksarma:
Head note: the branch was updated with main, so the head moved from 0b6d90dc to 548f34d8. The new head is a mechanical merge of 0b6d90dc and a9704cb8: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 548f34d8.

PR #1269, opened 2026-09-10
Merged board: a clear only the viewer holds drops a remote archived top whose only completion is a copied status

ksarma:
## Symptom

On a board that merges several kernels, under Show completed, a remote top named by the viewer's own cleared.jsonl stays listed, struck through, with its subtree, when its only completion is the copied "completed" status. The same top, cleared through the owning kernel, leaves the list since #1243 (https://github.com/romp-on/romp/pull/1243). The two disagree whenever the clear reached the viewer's cleared.jsonl and not the owning kernel's: a cleared.jsonl copied between machines, a session whose goal store moved to another machine, an older client. A current client routes a remote card's clear to the owning kernel (or drops it with a notice when that kernel is unreachable), so the case is narrow; when it holds, nothing forwards the clear, and the row persists on every build.

## Cause

`applyViewerClears` in `ui/webview/federation.ts` (the viewer's cleared ids over remote rows, #1190 https://github.com/romp-on/romp/pull/1190, with #1195 https://github.com/romp-on/romp/pull/1195 for the row's sid) marks a remote archived top the viewer's ids name as cleared and rolls the flag down its subtree; it never removes a row. `_ledger_cleared_overlay` in `kernel/kernel.py` (#1243) drops a cleared depth-0 row whose `done` is derived and whose summary is blank, with the descendants that follow it. The call-site comment promises the rows read exactly as the owning kernel's own overlay would have read them; after #1243 that no longer held.

## Fix

`applyViewerClears` applies the same rule over the remote rows, from the projection fields the frame already carries (`derived`, `summary`, `cleared`): at depth 0, `rootCleared` is the row's copied flag or a foreign id, as before, and `drop = rootCleared && !!n.derived && !String(n.summary || "").trim()`. A dropped root and every non-depth-0 row that follows it (its subtree, the flat-list order the roll-down already reads) leave the list; otherwise the mark-and-roll-down is unchanged. The entry is still rebuilt fresh, so the host payload's own rows are never mutated. The doc comment and the call-site comment carry the reasoning from #1243: a cleared flag rolls up to status cleared, so a copied "completed" beside a clear is a stale copy, not a completion; a root with its own verdict (`derived` false) or a takeaway (a non-blank `summary`) stays listed and reads cleared. No type, payload or kernel change.

Two points, stated so the choice is visible:

- The drop reads `rootCleared`, which includes the remote row's own copied flag, as the kernel's `root_cleared` does; the viewer's ids stand in for entries the owning kernel's cleared.jsonl would have held, and the existing `if (!foreign.size) return` plays the part of the kernel's empty-file early return. So while the viewer holds any foreign id, a status-only row that reached the merged board struck through by its own flag (from an owning kernel older than #1243, or one whose cleared.jsonl is empty) also leaves the list, which is what a current owning kernel with those entries would show. Gating the drop on the viewer's ids alone would leave that row struck through; the reported case drops under either.
- #1243's review recommends widening the kernel rule to the literal #1187 reading (drop every cleared archived root without a takeaway, whatever its completion source). If that lands, this condition widens the same way (the `derived` factor goes in both), or the two overlays disagree again.

## Tests

`ui/webview/federation-cleared-overlay.test.ts`, two cases appended.

The first, "a foreign clear of a remote archived top whose only completion is a copied status drops the row and its subtree; a verdict or a takeaway keeps it": a remote frame carries six roots, every one `done: true`: g1 (`derived: true`, no summary) with a child; g2 (`derived: false`) with a child; g3 (`derived: true`, a takeaway); g4 (`derived: true`, a whitespace summary) with a child; g5 (`derived: true`, no summary) with a child; g6 (`derived: true`, no summary, its copied `cleared` flag already on) with a child. The local payload names g1 to g4. Through `mergeHostFeeds` the remote entry's rows must be exactly g2 and g2a (cleared), g3 (cleared), g5 and g5a (uncleared); the remote payload's own eleven rows must be untouched (g1 first, cleared false); a local entry of the same shape must pass through as the same object. Before the change the assertion fails with eleven rows led by g1 and g1a, cleared, and g4, g4a, g6 and g6a cleared: the struck-through rows.

The second, "a drop with nothing else to mark still takes effect: the entry is rebuilt with the row and its subtree gone": `applyViewerClears` over one entry of a status-only root and its child, with the root the only id named, must leave the entry's `archivedTops` empty, replace the entry object, and leave the two original rows in place. Before the change it fails with both rows present and cleared. It pins the rebuild the drop alone has to cause: without it the entry passes through by reference and the row stays listed, uncleared, which the first case cannot see because its other marks rebuild the entry anyway.

Four mutants of the drop arm each fail the file: the `derived` factor removed (three cases fail, since the existing fixtures carry no `derived`), the trim removed (g4 and g4a stay), the subtree skip removed (g1a, g4a and g6a stay in the first case, g1a in the second), and the drop taken without setting `changed` (the second case: both rows stay, uncleared). The three existing cases pass unchanged, as does `routing-clear-ids.test.ts`.

`npm run typecheck` clean. Full `npm test` on the head (`node --test` over every bundle, with the timeline-tags-scale file run alone as a second leg): 4069 tests, 4069 pass, 0 fail.

Tier: fix

### DIFF
--- a/ui/webview/federation-cleared-overlay.test.ts
+++ b/ui/webview/federation-cleared-overlay.test.ts
@@ -1,6 +1,7 @@
 // The viewer's ledger over remote rows (review find, 2026-09-09): a remote kernel's feed and archive projection read
 // only their own cleared.jsonl, so an id the LOCAL ledger clears for a remote card (clearedForeign on the local
-// payload, bare) must drop that remote ask/item and read that remote archived top cleared, subtree with it, while
+// payload, bare) must drop that remote ask/item and read that remote archived top cleared, subtree with it (a top
+// whose only completion is the copied status leaves the list instead, as the owning kernel's overlay drops it), while
 // local rows and untouched remote rows pass through exactly as before. Same harness as feed-clock-anchor.test.ts.
 import { test } from "node:test";
 import * as assert from "node:assert/strict";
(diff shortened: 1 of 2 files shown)


ksarma:

Reviewed inline: `applyViewerClears` in `ui/webview/federation.ts` against the kernel's `_ledger_cleared_overlay` at this head. The predicates match over the same projection fields. At depth 0, a root is cleared by its own copied flag or by an id the viewer holds. It drops when derived is set and the summary is blank. Descendants follow their root in the flat list until the next depth-0 row, which resets the drop. The kernel's empty-ledger return and the viewer's empty-foreign return are the same gate. The archived-tops producer projects each root with no ancestor and appends it before recursing. So derived at depth 0 means the copied status or a summary, and the order the subtree skip relies on holds. `prefixInbound` passes `archivedTops` through by reference. The kernel drops a root struck by its own flag whenever its ledger holds any entry. Design point (a) therefore mirrors the kernel rather than widening it. The residual the body names is accepted. An owner older than #1243, or with an emptied ledger, lists such a row struck through where the merged board drops it. A widening of the kernel rule must carry over, as point (b) says. Rows are rebuilt fresh, and the host payload is never mutated. The second test pins the rebuild when a drop is the only change. The `fix` label fits. Privacy scan of the 75 added lines is clean. Body and commit message match the diff.

Verification:
- Red-first met by this session's runs: the head's `federation-cleared-overlay.test.ts` over base cfc44a4d gives 3 pass, 2 fail. The two new cases fail with the readings the body states (eleven rows led by g1 and g1a cleared; both rows present and cleared). At head: 5/5, plus `routing-clear-ids` 3/3, typecheck clean, and all four mutants fail the file.
- CI: eleven checks SUCCESS at c01b1ac1.
- Merges clean onto current main.

Review run by the maintainer automated review session; verified against head c01b1ac1.

ksarma:
Head note: the branch was updated with main, so the head moved from c01b1ac1 to eb30999e. The new head is a mechanical merge of c01b1ac1 and a9704cb8: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head eb30999e.

PR #1270, opened 2026-09-10
Auto-reload: the plain send's refusal on a disconnected host does not come back on the page that follows the reload either

ksarma:
Follow-up to #1227 and #1246 (both merged), which marked the refusals that report a state so the page that follows a restart's reload does not repeat them, and named this one as the refusal that stays unmarked. This PR argues the other way, and marks it.

## Symptom

A send to a session whose host is disconnected is refused with a toast: the host is disconnected, so this wasn't sent; the message is still in the composer, romp is re-dialing the link, send again when it's back. When a kernel restart's reload takes the page inside the toast's twelve seconds, the fresh page shows the toast again. The reload waits while the composer is focused with text in it, so the case is a send refused, the composer left, and the restart's reload inside the twelve seconds. At that moment the fresh page shows every host as connected: no tab mark, no transcript foot. The toast says a host is disconnected on a page that shows no such thing, and says romp is re-dialing when nothing on that page asked for a re-dial.

## Cause

The refusal is a plain `warnToast`, so `persistNoticesForReload` keeps it and the fresh page replays it at module load.

#1227 and #1246 left it unmarked on the reading that what it says stays true after the reload: the message was not sent, it is still in the composer, and a send there sends it. That holds for one of the toast's three claims. The other two are claims about a state.

- "The host is disconnected" is the state the staged sends' "Can't send yet" reports, on the same `hostIsDown` predicate, and that refusal is marked at both of its sites. `hostIsDown` reads the federation manager's disconnected set, which starts empty on a fresh page and has one writer, the page's poll of the kernel's `/tunnels`. The replay runs at module load, before the first poll, so it tells a page that shows every host as up that one of them is disconnected. Once the poll lands the toast is redundant (the host is still down, and the tab mark and the transcript foot say so) or false (the kernel that restarted dials every attached host as it comes up, so a link that was dead on this side comes back once that dial completes).
- "Romp is re-dialing the link now" is made true by the `redial` the refusal posts as it is raised. A replay posts none.
- The event claim, that the message was not sent and is still in the composer, is shown by the composer for itself: the input listener persists the draft and `restoreActiveDraftOnce` puts it back on the fresh page. That is the reason #1227 gave for marking the picker, attachment and roster refusals: the fresh page shows the state, so a replayed toast repeats it.

So the refusal reports a state, by the standard the census above `ephemeralWarnToast` already applies, and belongs with the marked ones.

## Fix

`ui/webview/render.ts`: the disconnected-host branch of `sendComposer`'s `deliver` raises its refusal through `ephemeralWarnToast`, with a comment at the site. The census above `ephemeralWarnToast` lists it with the other state refusals, says what the fresh page shows for itself (the host mark and the transcript foot, from the kernel's tunnel health, which the page reads afresh; the persisted draft, with the refused message in it) and that a replay posts no redial; the unmarked list keeps the nack, the dismissal and the other-tab ack. The toast's text and the `redial` it posts do not change; the `stageComposer` and `sendComposer` refusals #1227 and #1246 marked are untouched. No change to `reload-notices.ts`, the kernel or the docs.

Behaviour change: this one toast no longer comes back on the page that follows a reload inside its twelve seconds.

## Tests

`ui/webview/reload-notices.test.ts`, 19 cases under `node --test` (18 before):

- The lifted slice of the send's `deliver` now starts at its disconnected-host branch (`if (hostIsDown(sid)) {`, unique in the file) instead of the provisional branch, and runs through the first step of a send that passed both guards (`lastSent.set`, which remembers the text for a Ctrl+C restore), so a refusal that fell through would show as a remembered send. The page world gains a `remote` option that puts the active session on a host (`lab:web`), so the refusal has a host to name, and a `lastSent` map.
- One row joins the state-refusals table, "a send while the session's host is unreachable": the toast is on screen and starts with "lab is disconnected, so this wasn't sent.", a `redial` is posted and nothing else, nothing is queued, nothing is remembered as sent, the draft stays where it was, and `liveNotices` skips it. Before the change it fails at the reading, which returns the refusal's full text. With the `return` after the marked toast removed it fails on the remembered send (a variant that passed before the slice ran through `lastSent`). The "send into a tab whose create failed" row gains the same remembered-send check.
- The source-pin test matches the site through `ephemeralWarnToast`, tied to the branch head and the `redial` it posts, and counts eleven `ephemeralWarnToast` sites (the definition, the two reachability sites and the eight state refusals). Before the change it fails on the site pin and on the count (ten).
- `ui/webview/host-offline-ux.test.ts` pins the toast's text without the function name and is unchanged; `provisional.test.ts`, `queued-edit.test.ts` and `staged-list-cap.test.ts`, which pin neighbouring sites, pass unchanged (43 cases).

On this head: `npm run typecheck` clean; full `npm test` 4091 tests, 0 failing (4069 in the main leg and 22 in `ui/timeline-tags-scale.test.ts` run alone as a second leg; both legs exit 0).

Tier: fix

### DIFF
--- a/ui/webview/reload-notices.test.ts
+++ b/ui/webview/reload-notices.test.ts
@@ -5,10 +5,11 @@
 // render.ts keeps the texts of the toasts on screen in this tab's sessionStorage on the core's synchronous hook and the
 // fresh page shows them once. The readings are pure and execute here. render.ts has import-time DOM side effects, so
 // its wiring is pinned to source the way reload-restore.test.ts pins the scroll record's, and the toast family with the
-// refusals that report a state (the staging refusals, the branch jump's, the send into a tab whose create failed, the
-// queued edit's send on a session that cannot be reached) is lifted out of it and executed over a fake DOM the way
-// chat-exact-tail-exec.test.ts lifts chatTail. The served scenario (a nack on the last ship across a kernel restart; the
-// fresh page says it again, once) is tests/test_ship_reship.py NackNoticeSurvivesReload. Synthetic only.
+// refusals that report a state (the staging refusals, the branch jump's, the send on a disconnected host, the send into
+// a tab whose create failed, the queued edit's send on a session that cannot be reached) is lifted out of it and
+// executed over a fake DOM the way chat-exact-tail-exec.test.ts lifts chatTail. The served scenario (a nack on the last
+// ship across a kernel restart; the fresh page says it again, once) is tests/test_ship_reship.py
+// NackNoticeSurvivesReload. Synthetic only.
 import { test } from "node:test";
 import * as assert from "node:assert/strict";
 import * as fs from "node:fs";
@@ -50,11 +51,11 @@ test("the toasts on screen read as their texts, in order, blanks dropped; none w
 
 test("the reading asks for the toasts without the ephemeral mark: a refusal about a state the fresh page shows for itself stays behind", () => {
   // a refusal that reports a state (the staged sends' "Can't send yet": the host unreachable, the tab still being
-  // created; the send into a tab whose create failed; the queued edit's send on a session that cannot be reached; the
-  // staging refusals; the branch jump to a session not on this dashboard) is about something the fresh page shows for
-  // itself or no longer has; render.ts marks those toasts where they are raised (executed below) and the
-  // selector skips the mark. The mark's effect on a real DOM is executed by tests/test_ship_reship.py
-  // NackNoticeSurvivesReload.
+  // created; the plain send on a disconnected host; the send into a tab whose create failed; the queued edit's send on
+  // a session that cannot be reached; the staging refusals; the branch jump to a session not on this dashboard) is
+  // about something the fresh page shows for itself or no longer has; render.ts marks those toasts where they are
+  // raised (executed below) and the selector skips the mark. The mark's effect on a real DOM is executed by
+  // tests/test_ship_reship.py NackNoticeSurvivesReload.
   const asked: string[] = [];
   assert.deepEqual(liveNotices({ querySelectorAll: (sel: string) => { asked.push(sel); return [{ textContent: "kept" }]; } }), ["kept"]);
   assert.deepEqual(asked, [SEL]);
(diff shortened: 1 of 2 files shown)


ksarma:

Reviewed inline: one call changes in `sendComposer`'s `deliver`, the disconnected-host refusal raised through `ephemeralWarnToast`, and the census comment above it moves the site to the state refusals. The text and the redial post are unchanged, and the #1227 and #1246 sites are untouched. The argument holds against the code. `hostIsDown` reads the federation manager's `downHosts`, a fresh empty Set on every page. Its only writer is the tail of `poll()`, after an awaited `/tunnels` fetch. The replay is a synchronous statement at module load, so it always runs before the page knows any host is down. Only the gesture posts the redial that makes the re-dialing claim true; a replay posts nothing. The refused message sits in the persisted draft that `restoreActiveDraftOnce` restores, the reason #1227 gave for marking the picker and attachment refusals. Counter-case: the host is still down after the reload. The user gets the draft at once, the dimmed tab and its hover note when the first poll lands, and a fresh refusal with a fresh redial on the next send. A replayed toast would therefore be false before the poll and redundant after it. This reverses a per-site reading in the census, not the mark's rule, so `fix` fits. The lifted slice runs between two unique anchors and covers `lastSent.set`. `lab:web` is the host:sid shape `hostIsDown` splits at the colon.

Follow-up, pre-existing: at `ui/webview/render.ts:6511` the `romp-hosts` listener calls `renderTabs` alone, so the transcript foot the census names repaints only on the next `showActive` or `appendActive`; the tab mark is the immediate cue. Calling `syncHostOfflineFoot` beside `renderTabs` would make the comment exact.

Tests:
- Base 9109afac with the head's `reload-notices.test.ts`: 17 of 19 pass. The new row fails at the `liveNotices` reading; the source pin fails on the site and on the count of ten.
- Head: 19 of 19, with host-offline-ux, provisional, queued-edit and staged-list-cap unchanged at 62 total.
- The return-removed mutation fails only the remembered-send check.
- Typecheck clean. CI is green on every check at head.

Review run by the maintainer automated review session; verified against head 3872a037.

PR #1271, opened 2026-09-10
The pre-push hook reads each symlink's target in a pushed tip, which git grep skips

ksarma:
## Symptom

A branch whose tip carries a symlink pointing into a home directory pushes clean when the commit that added the link is already on a remote. The suite's tip case shows the asymmetry: a branch cut from main while main carried `leak.txt` holding such a path, before main redacted it, is refused at the tip and told to merge the main that carries the redaction; the same branch with a `node_modules -> /home/<user>/...` link in place of the file passes. The hook's header promises that a banned string anywhere in the tip's tree refuses the push, and for a symlink that was not true. The #968 body listed it as a limit.

## Cause

The tip pass is `git grep -i -I -l -F ... "$local_sha"`, and git grep reads regular-file blobs only. A committed symlink is a blob whose content is its target, so the grep never opens it. The added-lines pass does see a NEW link (`git diff-tree -p` prints `new file mode 120000` and the target as an added line), which is why a new link on a new branch is caught today. Not caught: a link the tip inherits from a commit some remote already has, and a link whose string entered the denylist after it was pushed.

## Fix

After the tip grep and before the added-lines loop, the hook lists the tip's `120000` entries with `git ls-tree -r`, reads each target with `git cat-file -p`, and on a match refuses the push with a tip line of the same shape as a regular file's, followed by `  in the SYMLINK TARGET of <path> -> <target>`.

- The read is `target=$(git cat-file -p "$blob" 2>/dev/null) || target=""`: under `set -e` a failed substitution would end the hook with no message, before its verdict on everything else. Such a link counts as empty; a test removes a link's loose object and checks that the verdict still prints.
- The ls-tree line (`<mode> <type> <sha>\t<path>`) is parsed with parameter expansion rather than `--format`, which needs a newer git; every construct is bash 3.2.
- One `cat-file` per link per pushed ref. The repo carries fourteen relative `bin/` links; they hold no home path and pass (checked over the real tree with a denylist).
- A new link at the tip of a new ref is reported twice, by the tip pass and by the ADDS pass, the way a regular-file leak is today; a test asserts both lines so the double report is deliberate.
- The header's tip line reads "in a regular file or a symlink's target"; the `added_lines` comment says a new link is an added line; CLAUDE.md's sentence on the hook gains "(regular files and symlink targets)".

## Tests

`tests/pre-push-hook.bats`: seven cases on the file's inline setup and helpers, plus a `branch_inheriting_mains_symlink_leak` helper mirroring `branch_inheriting_mains_leak`; `main_redacts_and_branch_merges` takes an optional path (default `leak.txt`) so the symlink twin reuses it. Every target is synthetic (`/home/zzsynthuser/...`, `/mnt/nas...`).

- "a branch that only INHERITED main's SYMLINK leak passes once its tip merged the removal": green before and after; pins the no-false-refusal side.
- "the same branch is refused while its TIP still carries an inherited SYMLINK leak": the symlink twin of the existing inherited-leak case. Before the change the hook exits 0 (fails at `[ "$status" -ne 0 ]`); after, refused naming `the tip of refs/heads/main`, `SYMLINK TARGET`, `node_modules -> /home/zzsynthuser/` and the merge-main remedy, with no `ADDS` line.
- "a RELATIVE symlink with no identifier passes": green before and after; the repo's own `bin/` links have this shape.
- "a NEW symlink at the tip is named with its link and target, by both passes": the target is mixed case (`/home/ZZSynthUser/notes`). Before the change the push is refused by the ADDS pass alone (fails at the `SYMLINK TARGET of notes-link -> ...` assertion); after, both lines appear. Dropping `-i` from the new grep fails this case.
- "an identifier in an INTERMEDIATE commit's SYMLINK TARGET is caught by the added-lines pass": green before and after; pins the ADDS coverage the comment claims, and that the tip names no link.
- "a dot in a denylist entry matches only a dot, in a file and in a symlink target": a denylist of `nas.zzsynth.invalid` against `/mnt/nasXzzsynthXinvalid/share` in a file and in a link passes. Green before and after; dropping `-F` from the new grep fails it.
- "a symlink whose blob cannot be read does not end the hook before its verdict": a leak in `leak.txt` plus a link whose loose object is removed; the hook exits 1 with `BLOCKED` and `leak.txt` named. Green before and after; without the `|| target=""` guard the hook exits 128 after the first tip line and the case fails.

Before: 2 of 22 fail. After: 22/22. `tests/install-sh.bats` (the hook installer and its real-push cases) 29/29. shellcheck clean at warning level. A real `git push` to a bare remote with the hook installed refuses the inheriting branch with the link named, and accepts it once the link is removed.

Tier: fix

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -11,7 +11,8 @@
 # Two checks per pushed ref, both about what THIS push changes on the remote:
 #
 #   the tip's tree    the content the push exposes. A banned string anywhere in
-#                     it refuses the push.
+#                     it, in a regular file or a symlink's target, refuses the
+#                     push.
 #   each new commit   the lines it ADDS: against its parent, or for a merge the
 #                     lines in none of its parents. A string introduced in an
 #                     intermediate commit and removed by a later one is still
@@ -81,10 +82,12 @@ rev_range() {   # <local_sha> <remote_sha> → a rev-list argument list
 # was clean (the #968 review). In a combined diff every content line carries one
 # column per parent, and a line in no parent has a `+` in each; -c also omits a
 # file that matches any parent outright, which holds no such line. -M keeps a
-# rename from re-adding every line of the file. The path comes from the
-# `+++ b/<path>` header, read only between a `diff --git` or `diff --combined`
-# line and the first hunk: inside a hunk, a line that begins with `+++ ` is added
-# content (`++ b/x` as written) and belongs to the current file.
+# rename from re-adding every line of the file. A new symlink shows up here too:
+# a link's target is its blob content, so it is an added line. The path comes
+# from the `+++ b/<path>` header, read only between a `diff --git` or
+# `diff --combined` line and the first hunk: inside a hunk, a line that begins
+# with `+++ ` is added content (`++ b/x` as written) and belongs to the current
+# file.
 added_lines() {   # <rev>
     local n
     n=$(git rev-list --parents -n 1 "$1" | awk '{ print NF - 1 }')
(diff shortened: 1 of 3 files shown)


ksarma:

Reviewed inline: the new block sits between the tip grep and the added-lines loop. It lists the tip's 120000 entries with ls-tree, cuts the blob sha at the tab that always follows it, and takes the path after that tab. A name with a tab, a newline, a leading space or C-quoting therefore still reads the right blob. Such a name is reported in git's quoted form, never skipped. The block reuses grep_args, which the hook builds as fixed strings. The hook exits at lines 38 and 48 when the denylist is absent or empty, so the symlink pass never runs with an empty pattern list. An unreadable blob counts as empty through the guarded substitution, so set -e cannot end the hook before its verdict. A tip with no links makes the grep exit 1, and the if skips the block. A new link is reported twice, by the tip pass and the ADDS pass; a test asserts both lines. The header, the added_lines comment and the CLAUDE.md sentence match the code. The added lines use no ls-tree --format, mapfile or case-folding expansion, so the bash 3.2 claim holds by reading. The fix label fits: the header promised a tip scan that a symlink target escaped, and two new cases are red on the base hook.

- `.githooks/pre-push:128`, follow-up: the unreadable-blob branch is silent. A one-line stderr note that the link was not scanned would match the fail-loudly rule and change no test outcome.

Test evidence: the head's `tests/pre-push-hook.bats` over the base hook at a9704cb8 fails exactly case 11 (inherited SYMLINK leak, hook exits 0) and case 13 (new link, no SYMLINK TARGET line), 20 passing. At head, 22/22. `install-sh.bats` 29/29. shellcheck clean at warning level. Each of the body's three mutations fails the case it names. Privacy scan of the 144 added lines: no denylist, hostname or non-synthetic home-path hits. CI at head: all 11 checks SUCCESS.

Review run by the maintainer automated review session; verified against head d40924af.

ksarma:
Head note: the branch was updated with main, so the head moved from d40924af to ee1e38f7. The new head is a mechanical merge of d40924af and 8786f5f1: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head ee1e38f7.

PR #1273, opened 2026-09-10
Chat: a late stamp reads an event that names the send as this send's, whatever its stamp

ksarma:
## Symptom

On a client whose clock runs ahead of the kernel's, a send pressed against a tab with no resident frame (a placeholder tab whose payload is still loading, or a provisional tab whose session does not exist yet) misreads its own records on the first frame. The user sees the kernel's echo and our dashed bubble both drawn until the landing (the message twice), or the kernel's never-delivered bubble beside ours still saying "sending", or, when the first frame already holds the landing, a bubble that never ends.

## Cause

A late stamp (`stampBase` in `ui/webview/send-pending.ts`) orders events by the kernel's stamps against the press's second: an event stamped before that second is background, and the last stable one among them is the anchor. The kernel stamps the send's own echo at its receipt and the landed atom at the CLI's record, both on the kernel's clock, so a client a second ahead (or a press that crosses a second boundary) puts those stamps before the press's second. The stamp then filed the send's own echo in `seen`, and the id path in `reconcilePending` read that echo as spoken for, so it covered nothing and the never-delivered verdict on it counted for nothing; and it took the landed atom, or a reply after it, as the anchor, so the landing sat before the scan and nothing could retire the bubble. The docblock named this cost as accepted, from before the frame could name the send. Since the press mints the id, the frame does name it: the echo's uuid is the id, and the landed atom carries it as `qid` or `qids`.

## Fix

`namesSend(e, qid)` says whether a user event names the send: its uuid is the id (the echo, delivered or flagged never-delivered), its `qid` is the id, or its `qids` include it; never for an entry with no id. `stampBase` finds the first such event: the anchor is looked for below it only, and the background loop skips it and every event after it, whatever their stamps. Queued groups keep their count and the queued presumption. Below that event, and throughout a frame that names the send nowhere, a record that carries no id (an older kernel, the tmux route) or another send's id is still read by stamp, as before. The header and the docblock state the rule; the accepted-cost sentence is gone. Two files change, `send-pending.ts` and its test; the `PendingSend` type and `dropPending` are untouched.

## Tests

`ui/webview/send-pending.test.ts`, section 16: one test over the pure module (no DOM) with the client's press a second after the kernel's stamps. Each record wears the send's id, built from the entry's `qid`.

- (a) the echo: `seen` is `[]`, the echo is hidden for our bubble (`echoHide` `[1]`, one inject), `received` is set. Before the change `seen` lists the echo's id, `echoHide` is `[]` and `received` is unset.
- (b) the landed atom: anchor `a1`, landed `[1]`. Before: anchor `u-landed`, nothing landed, the bubble kept.
- (c) the landed atom with a reply after it: anchor `a1`, landed `[1]`. Before: anchor `a2`, nothing landed.
- (d) the landing as the frame's only event: anchor `null`, landed `[0]`. Before: anchor `u-landed`, nothing landed.
- (e) a record of several sends with one block wearing the id: anchor `a1`, landed `[1]`. Before: anchor `uXY`, nothing landed.
- (f) the never-delivered echo: `lost` is the entry. Before: `lost` is `[]`, the bubble kept.
- (g) a same-text id-less record after the landed atom, stamped inside the same skew: anchor `a1`, `seen` `[]`, landed `[1]`. This pins that every event from the naming one on is after the send: a variant that skips only the naming event lists `u-after` in `seen`.
- (h) two records naming the send, the echo and then the landed atom: anchor `a1`, `seen` `[]`. This pins that the first one bounds the frame: a variant that takes the last lists the echo's id in `seen`.
- Guards, green before and after the change: (i) a same-text record wearing another send's id, stamped early, is background by text (anchor and `seen` name it, one bubble kept); (j) an id-less record stamped early, likewise; (k) an event of another kind wearing the id neither bounds the frame nor anchors the send (anchor `u-old`); (l) an entry with no id names nothing, so an early id-less same-text record is background for it and not its landing (anchor `u-old`, one bubble kept). Dropping the `kind` guard from `namesSend` turns (k) red; dropping the `qid` guard turns (l) red.

Red on the unchanged source at (a): `seen` deepEqual `[]` fails with the echo's id. Green with the change; the guards and the existing section-13 test (a foreign echo at the press's second) stay green throughout.

`npm run typecheck`: clean. Full `npm test`: 4092 tests, 0 failures (4070 in the main run plus the 22 of the timeline-tags-scale test run alone).

Tier: fix

### DIFF
--- a/ui/webview/send-pending.test.ts
+++ b/ui/webview/send-pending.test.ts
@@ -17,7 +17,9 @@
 // bubble's own entry; (13) a stamp taken late reads the events' own times. Round 4: (14) a late stamp
 // presumes the frame's newest queued copy of its text is its own (the queued bubble carries no stamp).
 // T252c gave every kernel copy an id; now (15) the id is the CLIENT's, minted at the press and posted with the
-// send, so the kernel's copies wear it from their first appearance and nothing is latched by text.
+// send, so the kernel's copies wear it from their first appearance and nothing is latched by text. And (16) at
+// a late stamp an event that NAMES the send (wears its id) is this send's whatever its stamp: the id outranks
+// the clock.
 // The decisions are executed through send-pending.ts; the DOM half is pinned in render.ts/styles.css.
 import { test } from "node:test";
 import * as assert from "node:assert/strict";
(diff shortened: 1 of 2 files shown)


ksarma:

Reviewed inline: `namesSend` and the two bounds it puts on `stampBase`. The predicate fires only for a user event whose `uuid`, `qid` or `qids` carry the entry's own id. It never fires for an entry with no id, so stamp still reads id-less records and other sends' records. The anchor loop starts at `firstNamed` minus one, and the background loop skips `i` at or past `firstNamed`. So the naming event is neither anchor nor background, and a frame that names the send nowhere reads exactly as before. Queued groups are counted before the skip. Against the kernel: the echo is merged among the last turn's atoms in kernel-clock order. `SdkBackend.send` keys the echo on the client's qid and never re-mints an admitted id. `_client_qid` refuses an id the session already holds, so an event carrying the id is this send's. Each entry runs its own `stampBase`, so a several-sends record retires this entry without taking a later send's anchor. A send named at index 0 gets a null anchor and an empty background, which is correct. The behind-clock reading the docblock keeps is unchanged. In `render.ts`, `echoHide` hides the echo, and `landed`, `lost` and `received` retire or clear the bubble. So (a) to (h) mean one bubble that ends at the landing. The `fix` label fits. `PendingSend` and `dropPending` are untouched. The added lines scan clean.

Test evidence:
- The head's section-16 test over base `a9704cb8` fails at (a) with `seen` holding the echo's id. It passes at head (send-pending 40/40; with queued-edit, staged-list-cap, provisional and provisional-cancel 77/77).
- Typecheck is clean.
- Six mutants of `stampBase` and `namesSend` each fail at the scenario the body names. The kind guard at (k) and the qid guard at (l) are among them.
- CI: all 11 checks green at head.

Review run by the maintainer automated review session; verified against head e64b4df6.

ksarma:
Head note: the branch was updated with main, so the head moved from e64b4df6 to 33346869. The new head is a mechanical merge of e64b4df6 and c2423b2a: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 33346869.

PR #1275, opened 2026-09-10
Tests: bats suites that start the real manager isolate their state root, and the suite floors poison the kernel and serve ports beside the manager port

ksarma:
## Symptom

On a machine that runs romp, a bats run of `tests/romp-manager-ensure.bats` or `tests/romp-manager-origin.bats` starts one fake kernel per kernel registered in the live `~/.local/state/romp/kernels.json`, each with that registry entry's `stateDir` in its environment, and its drain polls read the live `serve-token`. The suites pass either way, so nothing shows. From a shell of a profiled kernel's session, the three manager tests in `tests/romp.bats` do the same through the inherited `ROMP_STATE_DIR`. Under pytest, a test that dials the kernel through an inherited or absent `ROMP_KERNEL_PORT` or `ROMP_SERVE_PORT` reaches the developer's live kernel, and a `python -m unittest tests.test_x` run has no port floor at all, the manager's included.

## Cause

`bin/romp-manager` resolves `STATE_ROOT` as `ROMP_STATE_DIR || XDG_STATE_HOME/romp || ~/.local/state/romp`; `up` loads `STATE_ROOT/kernels.json` and calls `spawnKernel` per spec, `specEnv` hands each child the spec's `stateDir`, and `kernelToken` reads `STATE_ROOT/serve-token`. The manager writes nothing there and dials no live kernel, which is why the leak is silent. The two manager suites set neither variable in `setup()`, and `romp.bats` exports `XDG_STATE_HOME` under its fixture HOME but never unsets `ROMP_STATE_DIR`, which outranks it (a profiled kernel's sessions inherit it).

On the pytest side `tests/conftest.py` poisons `ROMP_MANAGER_PORT` only. `kernel/kernel.py` (`PORT`) and `postal/postal_service.py` (`KERNEL_BASE`) resolve `ROMP_KERNEL_PORT` at import, `bin/romp` reads it in every kernel subcommand, `hooks/romp-wake.sh` at every wake, and `bin/romp-manager` reads `ROMP_SERVE_PORT` first; each maps an absent variable to the live default 29855, so a pop is not safe and only a set value is. `tests/__init__.py`, the floor for unittest runs, poisons no port.

## Fix

Tests only.

- `tests/romp-manager-ensure.bats`, `tests/romp-manager-origin.bats`: `unset ROMP_STATE_DIR` and `export XDG_STATE_HOME="$TEST_DIR/state"` in `setup()`, beside the tmux socket isolation, with the reason in a comment.
- `tests/romp.bats`: `unset ROMP_STATE_DIR` beside the existing `XDG_STATE_HOME` export.
- `tests/conftest.py`: `ROMP_KERNEL_PORT` and `ROMP_SERVE_PORT` set to `"1"` at import beside the manager port, and the autouse fixture re-asserts all three per test (a module-level write in a test file also runs at collection and would otherwise hold for the rest of the run). A test that needs a port of its own sets it in setUp or passes it to the process it starts, as the ones that start a kernel already do.
- `tests/__init__.py`: the three ports set to `"1"` for `python -m unittest` runs.
- `tests/README.md`: two sentences.

## Tests

- New `tests/bats-state-isolation.bats` (9 tests), the ratchet for the bats side as `tests/test_state_isolation_order.py` is for the pytest modules. It resolves every `load`, `source` and `.` spelling (`load name`, `load "$BATS_TEST_DIRNAME/name"`, `source ...`) and each helper's own loads, every file once. A suite starts the manager when it, or a helper it loads, executes the repository's own `bin/romp-manager`: a path climbing from the tests directory (`pwd)/romp-manager"` or `../bin/romp-manager"`), or a `"$VAR/.../romp-manager"` of any depth under the variable where a command starts, with `VAR=value`, `node`, `run`, `exec`, `nohup`, `env` and `timeout N` prefixes allowed. Comment lines and lines that only talk about the binary (a grep, an assert, a `[[ ]]` or `[ ]` operand) are dropped, and a stand-in under a test directory does not match, so `romp-refresh-audit.bats`, which drives `bin/romp` against a fake `$TEST_DIR/bin/romp-manager`, is not listed. Each detected suite must carry both isolation lines before its first `@test`, in the suite head or a loaded helper; equivalent spellings are accepted (an exported `ROMP_STATE_DIR` under a shell variable, the assignment and `export` on separate lines, the unset with a second variable), and the check reads the line's shape, not where the path resolves, as the header says. The detected list is pinned exactly (`romp-manager-ensure`, `romp-manager-origin`, `romp-manager-tmux-scope`, `romp.bats`), so a regex drift can neither pass vacuously nor list a mention. Seven self-checks run the rule and the detector over synthetic suites written in `setup()` and removed in `teardown()` whether or not the check passes: a leaky suite; a manager started through a loaded helper; floors set in a helper; a floor written inside the first test body on its own lines (shaped as `setup()` would carry it, so only the cut at the first `@test` keeps it out) or only in a comment; `$BATS_TEST_DIRNAME`, `source` and chained (looping) loads; a node-less exec, the `../bin` spelling and a path under a root variable (`node "$ROOT/bin/romp-manager"`) beside grep-only mentions, the heredoc mock and the refresh-audit stand-in shape; and the SIGPIPE-ignored runner (the suite text is collected whole and matched after, never piped into `grep -q`, which on a runner that starts children with SIGPIPE ignored prints `grep: write error: Broken pipe` into `$output`). Red before the setup changes: 8 of 9 pass and the rule test fails listing five lines, both lines for each manager suite and the unset for `romp.bats`; the detector test is green on the unchanged tree, so the red is the rule, not the detector. Three drifts checked by hand each turn one self-check red: returning the whole file instead of the head cut fails the in-test-floor check, dropping the depth group from the regex fails the root-variable check, and a regex that matches nothing fails the exact-list test and leaves no synthetic directory behind. One spelling the detector does not see: a real binary assigned through a variable rooted somewhere other than the tests directory and then run as `node "$MGR"`; the header says a new spelling is added to the regex with the suite that introduces it. No suite on the tree spells it that way.
- `tests/test_state_isolation_order.py::test_the_suite_wide_floors_stay_in_place` reads `conftest.py` and `__init__.py` as ASTs and pins each half of the port floor apart, as `tests/test_cli_scope_floor.py` does for the cli-scope floor: a module-level statement `os.environ["<VAR>"] = "1"` for each of the three ports in both files, and in `conftest.py` an autouse fixture whose body re-asserts each. Red before the `conftest.py` and `__init__.py` changes (`conftest.py must set os.environ["ROMP_KERNEL_PORT"] = "1" at module level`), and red alone when the import-time statement is removed with the fixture's copy kept, when the fixture's re-asserts are removed with the import-time statement kept, or when the fixture loses `autouse=True`. The import-time half is a source pin by design, as the file's XDG and pop pins are: every module is collected before any test runs and several set the same values at import, so a run-time read cannot tell that floor from a module's own set.
- New `tests/test_state_isolation_order.py::DeadPortsHoldPerTest`, the run-time half of the fixture claim: `setUpClass` writes a second dead port to the three variables and the test reads `"1"` back, a value only the fixture's per-test re-assert can produce (pytest sets a class up before its function-scoped fixtures run). The saved values are restored in `tearDownClass`, and the test skips under a bare unittest run, which has no fixture. Red on the unchanged `conftest.py` (`ROMP_KERNEL_PORT is not re-asserted per test`) and under each fixture mutation above; green on this tree, alone and in the full run.
- Green after the change: `bats tests/bats-state-isolation.bats` 9/9, `bats tests/romp-manager-ensure.bats` 7/7, `bats tests/romp-manager-origin.bats` 1/1, `bats tests/romp.bats` 119/119; `pytest tests/test_state_isolation_order.py tests/test_tempdir_hygiene.py tests/test_cli_scope_floor.py` 25 passed, 63 subtests; `python -m unittest tests.test_state_isolation_order` 2 passed, 1 skipped, and after `import tests` the three ports read `1` with `ROMP_STATE_DIR` absent.
- Full pytest suite on this tree (`-n 6`, the served modules running against Chromium): 10957 passed, 41 skipped, 0 failed, 1688 subtests passed. The import-time poison changes what every test sees, and every module that starts a kernel or a server sets its own ports.
- Leak checked by hand against a scratch state root seeded with a synthetic `kernels.json` (one aux entry, placeholder paths): a manager started the way the suites start it, with that root inherited through `ROMP_STATE_DIR`, listed the synthetic kernel in `/status` beside `main`; with the two isolation lines it listed `main` only.

Tier: docs

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -45,7 +45,16 @@ Every bug fix or feature change lands with a test (repo rule). Five suites:
   (`ROMP_CLI_SCOPE_MEMORY_MAX` and the others): the kernel hands them to every
   session's CLI and a tool shell inherits them, so a suite run from a session on
   a self-hosted install would otherwise see them at every backend construction
-  and in every exact argv pin.
+  and in every exact argv pin. `conftest.py` and `__init__.py` set
+  `ROMP_MANAGER_PORT`, `ROMP_KERNEL_PORT` and `ROMP_SERVE_PORT` to a dead port
+  (never unset: to every reader an absent variable means the live default), so
+  no test dials a live manager or kernel through an inherited value.
+  Any suite that starts the real `bin/romp-manager` also gives it a state
+  root of its own before its first `@test` (`unset ROMP_STATE_DIR` plus
+  `export XDG_STATE_HOME="$TEST_DIR/state"`, or an exported
+  `ROMP_STATE_DIR`), since the manager boots from its state root's
+  `kernels.json` and reads the serve token there; `bats-state-isolation.bats`
+  is the ratchet.
   Any test whose subject binds a loopback port picks it with `load
   free-port` + `free_port VAR...`, never a literal: a literal shared by two
   files collided within one run (`romp-manager-ensure.bats` once used
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -67,3 +67,9 @@ os.environ["XDG_STATE_HOME"] = tempfile.mkdtemp(prefix="romp-tests-state-")
 # thread method ends a hung run that way), so a hung run leaves this dir beside the root.
 STATE_DIR = os.environ["XDG_STATE_HOME"]
 os.environ.pop("ROMP_STATE_DIR", None)  # a live kernel's export outranks the XDG floor
+# No unittest run may reach a REAL manager or kernel either: the same three ports conftest.py poisons
+# for pytest (its comment has the readers and the incident). Set to a dead port, never popped: every
+# reader maps an absent variable to the default port, which on a machine running romp is the live one.
+os.environ["ROMP_MANAGER_PORT"] = "1"
+os.environ["ROMP_KERNEL_PORT"] = "1"
+os.environ["ROMP_SERVE_PORT"] = "1"
--- a/tests/bats-state-isolation.bats
+++ b/tests/bats-state-isolation.bats
@@ -1,0 +1,338 @@
+#!/usr/bin/env bats
+
+# Every bats suite that starts the REAL bin/romp-manager gives it a state root of its own. The
+# manager's STATE_ROOT is ROMP_STATE_DIR || XDG_STATE_HOME/romp || ~/.local/state/romp, and a suite
+# whose setup() sets neither hands the manager the developer's live root: `up` loads that root's
+# kernels.json and spawns one child per registered kernel (the suite's fake launcher, with the
+# registry entry's stateDir as its ROMP_STATE_DIR: specEnv), and the drain poll's token comes from
+# that root's serve-token (kernelToken). The manager writes nothing there and dials no live kernel,
+# so the leak is silent: the suite passes while a fake kernel runs for every kernel registered on the
+# machine. This is the ratchet for the bats side, as tests/test_state_isolation_order.py is for the
+# pytest modules: the isolation lines must be present in each such suite, before its first @test
+# (setup() runs before every test; a floor set inside one test leaves the others on the live root).
+#
+# The two lines, as the suites spell them:
+#     unset ROMP_STATE_DIR                    # a profiled kernel's sessions inherit it; it outranks the XDG floor
+#     export XDG_STATE_HOME="$TEST_DIR/state"  # or another path under a shell variable
+# A suite that instead exports ROMP_STATE_DIR to such a path satisfies both, and so do the equivalent
+# spellings: `X="$TEST_DIR/..."` on one line with an `export X` on another, or the assignment unquoted.
+# The check reads the line's shape, not where the path resolves: every suite today uses `$TEST_DIR`, or
+# `$HOME` after re-exporting HOME under its test directory (tests/romp.bats); a variable left pointing
+# at a live path would pass it. A helper the suite `load`s counts as part of the suite, both ways: a floor
+# set in a helper's function isolates, and a helper that starts the manager makes the suite one that
+# does. Any spelling of the load counts (`load name`, `load "$BATS_TEST_DIRNAME/name"`, `source`),
+# and a helper's own loads are read in turn.
+
+TESTS="$(cd "$(dirname "$BATS_TEST_FILENAME")" && pwd)"
+
+setup() { SYN="$(mktemp -d)"; }   # the self-checks' synthetic suites, removed whether or not the check passes
+teardown() { rm -rf "$SYN"; }
+
+# The helper files a suite brings in, one per line, nothing for a suite that loads none: every `load`
+# (bats resolves `load name` to name, else name.bash, beside the suite; a path as given when absolute),
+# `source` and `.` line, with `$BATS_TEST_DIRNAME` or `${BATS_TEST_DIRNAME}` in the argument expanded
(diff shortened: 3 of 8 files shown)


ksarma:

Reviewed inline: the two manager suites' `setup()` unsets `ROMP_STATE_DIR` and exports `XDG_STATE_HOME` under `TEST_DIR` before the fake launcher and every `@test`. `romp.bats` unsets `ROMP_STATE_DIR` beside its `HOME` and `XDG` exports. `bin/romp-manager` resolves `STATE_ROOT` as `ROMP_STATE_DIR`, else `XDG_STATE_HOME/romp`, else the home default. It reads only `kernels.json` and `serve-token` there, so both land under the test directory. Its one other home read is a `romp-serve` candidate, which every suite bypasses with `ROMP_SERVE_BIN`. `conftest.py` sets `ROMP_KERNEL_PORT` and `ROMP_SERVE_PORT` to 1 at import and in the autouse fixture, beside the manager port; `__init__.py` sets all three. Every test module that reaches a kernel passes its own port or sets `KERNEL_BASE`, so none relied on the absent-variable default. The ratchet executes no romp binary: it greps suite text and writes synthetic suites under a `mktemp` dir it removes. It collects each suite's text whole and matches it with a here-string, never piped into `grep -q`. Run by hand over every bats file, its regexes match exactly the four pinned suites, each through the `pwd)/romp-manager` line. The other `romp-manager` mentions are heredoc targets, `chmod` operands or `ROMP_MANAGER_BIN` values, and no suite runs the binary through a variable rooted outside `tests/`. All eight files sit under `tests/`, so the `docs` tier fits. The added lines carry no private strings, hostnames or home paths.

One follow-up, not a blocker. `tests/bats-state-isolation.bats:76` drops lines that start with `[`, `[[`, `test` or `!`, and `CMD_START` has no `if`, `while` or `until`. A future suite whose only manager start sits after such a guard would pass undetected. Applying `TALK_ONLY_RE` only to lines without a `&&`, `;` or `|` continuation would close it.

Run at the base, the head's tests fail: the ratchet ran 8 of 9, the rule test listing the five missing lines, and `test_state_isolation_order.py` failed 2 of 3. At head the ratchet is 9/9. The pin module with `test_tempdir_hygiene.py` and `test_cli_scope_floor.py`: 25 passed, 63 subtests. unittest: 3 ran, 1 skipped by design. The four suites that start the real manager were reviewed by reading only, not run here. All eight required checks pass at head.

Review run by the maintainer automated review session; verified against head a326eb6e.

ksarma:
Head note: the branch was updated with main (be76aaff), so the head moved from a326eb6e to e2071bd1. The new head is a mechanical merge of a326eb6e and be76aaff: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head e2071bd1.

ksarma:
Head note: after #1278 landed, the branch was brought up to date by a merge of main (9ac871fc) into the branch, so the head moved from e2071bd1 to d7d60eb6. Its added and removed lines against the merge base with main are identical to the reviewed diff (the tests/__init__.py trailing block now follows the romp_load registration #1278 added). The review above stands for this head.

Review run by the maintainer automated review session; verified against head d7d60eb6.

PR #1277, opened 2026-09-10
Chat build: one complete per-session signature serves every tab, and a judge pass rebuilds only the tab whose store it moved

ksarma:
Builds on https://github.com/romp-on/romp/pull/1251 (the chat build's fixed-cost memos and the per-input rebuild attribution), which has merged: this branch is rebased onto main and its three commits are the whole diff.

## Symptom

Publish one session's goal store and every open chat tab rebuilds, byte-identical. On a board with many tabs the judge's passes made the pusher spend most of its time re-reshaping transcripts nothing had changed in: in the replay on the branch this work was derived from, 420 of 435 chat builds per 120 s were rebuilds of background tabs whose only moved input was the global judge-pass counter (that figure was not re-measured here). The watched tab had a second key of its own (an exact key beside the background tabs' file-stat key) and a start-keyed dirty watermark standing in for the in-memory stamps neither key named. An input that key did not name left the tab stale until some file moved (a tmux session's composer echo, until 2026-09-05), and every edit inside the exact key had to be expressed in two places.

## Cause

`_chat_build_sig` included `_judge_gen[0]`, the counter every producer pass advances when any judge-written store moved, as a proxy for the judge's outputs, so a publish in session B moved session A's key. The watched tab's second key (`_active_chat_sig`, with `_clock_predicates`, `_external_sig` and `_ACTIVE_SIG_FILES`) stat'd a fixed list of side files and digested the tail's atoms for a backend without a revision, and the `_views_dirty` watermark covered whatever was left. Two writers of the SDK backend's live tail (`unqueue`, `edit_queued`) and one flag write (`_landed`) did not advance the tail's revision, and the tmux echo store had no revision at all.

## Fix

Three commits.

**1. The live-tail revision.** Every change to a session's in-memory tail advances one per-session revision on both backends. The SDK backend's `_touch_live` gains `unqueue`, `edit_queued` and the `_landed` flag write; `prune_live` and `retire_live_work` bump once per settling call; the two flag writes `_mark_dropped_echoes` makes outside the lock each take it for their bump; `live_rev` reads under the live-tail lock; the rule is written on `_touch_live`, and `tests/test_live_tail_rev.py` pins the writer set by source, so a queue or echo mutator that touches `_live` fails until it bumps. The kernel's tmux echo store gets `_tmux_echo_rev`, bumped by `_tmux_echo_add`, `_tmux_echo_prune`, `_tmux_echo_settle` and `TmuxBackend.dismiss_echo` only when something was added, dropped or marked; `TmuxBackend.live_rev` reads it. `Sessions.live_rev` dispatches to the owning backend and serializes the tail for a backend with no counter. The bench's constructor-free backend carries the revision slot.

**2. One signature for every tab.** `_chat_build_sig` is a flat tuple with one labelled component per input `build_session` reads (`_CHAT_SIG_LABELS`, 35 labels), and every tab is served while its signature holds. The components, in groups:

- Files by identity: the transcript and states files; the session's own goal store with its journal and archive (`jd._store_identity`, so a publish busts the owning tab at once and no other); the archive headline; the episode log; the registry and the death marker; the task store; the instruction files on the chain.
- In-memory state by value or revision: the rewind hold; the pending cut; the live tail by revision; the liveness row minus its per-cycle stamp; the backend's brackets and queue with each copy's identity, its recallability and launch error; the parked ops; the account hold, read only while a queued bubble can render (a queue, parked ops, or a tmux echo in flight); the retry state; the live background-task rows after the build's own expiry filter; the watches as rendered; the awaiting-stamp view; the warm-anchor revision; the branches forked from the session; the cwd's branch and repository and the trees it names.
- The clock, as the booleans the payload renders: the interrupt and compaction caps, the model-switch cap, the faded hour. The key moves at each crossing and at no other tick.
- Shared by every tab and taken once per push (`_chat_sig_shared`): the suspension count, a digest of the names registry, the flags, bell and colormap files, the login label with the availability half `_auth_avail_status` returns, `cleared.jsonl`, the host name.
- Three dependency components re-evaluated each cycle over the record the last build left (`_chat_build_deps`): the files whose tails the payload embeds, noted by `_read_task_output`, `_subagent_meta_map`, `_subagent_file` and `_agent_steps` as they read them (the sibling-directory fallback in `_subagent_file` records each sibling's `subagents/` directory, the directory its glob reads: a file landing there moves that directory's mtime, not the sibling's); the unresolved path tokens, re-resolved only for the messages whose candidate directories moved (`_chat_pl_precheck`, sharing `_repo_index_key` with the repo file index); and the postal cards' embedded values beside the log's identity, where a raw postal event that did not hydrate flags the log as a dependency too.

A build is cached only when the static components held across it (compared on a post-build signature taken without the dependency tail); otherwise it counts under `builds.chat.moved` and the next cycle builds it again. The cache entry is `(signature, payload, serialization, dependency record)`; index 2 stays the lazy serialization, so `tests/test_chat_skeleton_reconnect.py`'s pins hold. The pusher consumes the record after each build and leaves none on its thread; `_push_session_now` and the scroll-back slice, which cache nothing, leave none either. The comment threads' cache keys on the same signature without the dependency tail plus the thread's own states row, and keeps the dirty watermark for what that tail would have carried.

The liveness map the push holds is served to the signature's nested reads through the existing `_serve_live` scope, so `_bg_live_norm`, the awaiting sources and the watch rows read the snapshot the build will, and a connect push on a handler thread forks no tmux per tab; the liveness helpers keep their shapes. A signature that raises (one component's read failed) is said once per fault episode, on stderr with the traceback and as a refused bell row, the rule the chat-build fault follows; while the fault stands the tab is built every cycle and never cached (counted under `nosig`), and a signature that is taken ends the episode.

Retired: `_active_chat_sig`, `_clock_predicates`, `_external_sig`, `_ACTIVE_SIG_FILES` and the chat cache's `_views_dirty` watermark. `_judge_gen` stays for the feed and timeline view signature. `/perf`'s `bg_miss` carries one counter per label plus `cold` and `nosig`; `builds.chat` gains `moved`; `romp perf` prints the moved count beside the split when it is non-zero.

**3. Docs and the two webview comments** that explained the active-tab re-arm through the retired key. They now date the streamed-nowhere symptom to that key and say what a missing hint costs today: the watched tab's place at the head of the build order and the first flush.

Design points a reviewer may want to weigh:

- A bare `_mark_views_dirty` no longer rebuilds the watched tab. The post-build signature is the event-based form of the start-keyed watermark: the cache asks whether any named input differed between the signature before the build and the one after it. The accepted gap is an input moving between the post-build signature and the send, which the next cycle's signature sees (one cycle, the watermark's own granularity). `test_kernel.py`'s push-cache test is rewritten for this and pins that a live-tail echo, which the payload renders, does rebuild.
- Every tab now pays the signature per push. Measured with `tools/perf-bench.py` against one copy of this machine's state directory (8 live sessions, all tabs served, `--iters 5`), main at 8786f5f1 against this head: `push_steady` 19.52 ms to 24.04 ms per push (+4.5 ms, about 0.6 ms per tab); `push_connect:chat` 37.45 ms to 42.43 ms (+5.0 ms: the connect push now takes the signature and opens its shared scopes on the handler thread). `build_session_warm` moved within 1.5 ms per session and `push_cold_cycle` within 0.8 % (noise). On one copy of the state, base and head produce the same chat bytes. A first compare, on an earlier copy, differed by 243 bytes between its two runs; a second base run on the same copy as the head matched the head and the first base run, so the difference was a running task's output tail, read outside the copy, growing between the runs. The bench runs no judge pass between pushes, so it does not measure the rebuilds a publish no longer causes; that mechanism is pinned by the pusher tests below.
- The pathlink pre-check vouches for a message's unresolved tokens while the cwd, the repo-index key and the identities of its candidate directories hold, so a quiet cycle re-resolves nothing; a directory that moved re-resolves only the messages that named it.
- The signature is compared by position with `==` (its values include the liveness row's lists and the stamp view's frozenset), never hashed.

## Tests

Executing tests, each red before the change it covers:

- `tests/test_live_tail_rev.py` (new, 16 tests): every writer of either tail bumps once per changing call, by behaviour and by source; a read, a prune that retired nothing, a settle over marked echoes and a queue miss do not; `Sessions.live_rev` dispatches to the owning backend and serializes the tail for one without a counter. On main before commit 1, 11 fail (`unqueue`, `edit_queued` and the `_landed` write bump nothing; `prune_live` and `retire_live_work` bump per atom; `_tmux_echo_rev`, `TmuxBackend.live_rev` and `Sessions.live_rev` do not exist).
- `tests/test_chat_build_sig_inputs.py` (new, 47 tests; before commit 2, 43 fail, the four passing being the census's self-checks):
  - The census derives every helper call, dotted call and module global of `build_session` by AST and requires each to be classified, every `sig` label to be a component of the signature, and the retired names to be absent.
  - A differential suite over a hermetic world moves one input at a time and requires the miss under exactly that input's label: a transcript append, a states row, the store and its journal and archive, another session's publish plus the producer's generation bump, a rewind hold, the archive and episode log, the registry and death marker, the task store, a tmux echo and its dropped mark, the liveness row and its per-cycle stamp, each clock boolean at its crossing, parked ops and the account hold, the retry state, the task rows, a watch and one re-armed under the same note, the awaiting stamp and a postal-log landing, the anchor revision, the suspension list and the names digest, the shared files, the billing readers, the cwd and instruction files, a recorded task output that grows or appears, a pending path token whose file appears, a postal card's caption and the log. The handed liveness map is served to every nested read: a fresh liveness read during the signature fails the test.
  - The pusher with stubbed builders: a session's store publish rebuilds that tab alone under `store`, a bare dirty mark rebuilds nothing, a live-tail echo rebuilds its tab, the entry is the four-tuple, a build whose signature moved is not cached and counts under `moved`, a signature that raises is said once per episode (one stderr line, one bell row, the tab built and uncached under `nosig` until the read succeeds), and a push closes the scopes a previous push on its thread left open.
  - The pre-check and the scopes: a quiet cycle re-resolves no token, a moved directory re-resolves the message that named it, a push outside a cycle opens and closes its own scopes, and the cycle's finally closes what its jobs left open.
  - The real `build_session` under `_push`: a running task's output file and the agent's own transcript are recorded under the identity they were read under, and each growing rebuilds the tab under `taskout`; an agent file found under a sibling directory records that sibling's `subagents/` directory, and a file landing there rebuilds the tab; a fold hit over the same parse carries the sealed card's output into the record; a raw postal marker the index does not hold flags the log, and the log landing rebuilds under `postal`; a targeted push leaves no record on its thread.
- `tests/test_stage0_log_readers.py`: the watched-tab key class is rewritten as `EveryTabIsServedOnTheCompleteKey`, the same observables through the one key, each attributed to its component. The `base=` reuse test went with the parameter; the captions store is covered by the postal component; the in-flight judge set is not an input of `build_session`. Commit 1 also routes that class's tmux drop through the store's own settle and expects a new key after the dismissal, since a revision counts changes and never returns.
- `tests/test_kernel.py`: the push-cache test's dirty-mark step (the first design point).
- `tests/test_chat_fixed_cost_memos.py`, `tests/test_perf_stats.py`, `tests/romp-perf.bats`: the label pins, the `moved` field and the CLI fixture follow the new labels (`store` where `judge_gen` was); the CLI fixture's window carries three moved builds and the chat line asserts `; 3 moved`, red without the `bin/romp` change.

Mutation checks, each turned red by a test above: a reader that stops noting its file (`_read_task_output`, `_agent_steps`, `_subagent_meta_map`, `_subagent_file`); the sibling fallback noting the sibling instead of its `subagents/`; a fold hit dropping the sealed prefix's outputs; `build_session` ignoring the raw tail for the postal flag; `_chat_build_sig` serving no handed map; the pusher or `_push_session_now` leaving the record on the thread; the push scopes left open at the top of a push or by the cycle's finally; the signature fault written every cycle; `romp perf` printing no moved count.

Runs on the final tree: `pytest -n 6 tests/` 11071 passed, 41 skipped, 0 failed, 1705 subtests passed (162 s); `npm run typecheck` clean; `npm test` 4094 passed (4072 in the main leg, 22 in the timeline-tags-scale leg run alone), 0 failed; `bats tests/romp-perf.bats` 15 of 15.

Not included: the working note and the user todos are not inputs of `build_session` here, so they have no component; a later change that reads one in the build fails the census until it is classified and keyed.

Tier: fix

### DIFF
--- a/bin/romp
+++ b/bin/romp
@@ -1254,11 +1254,13 @@ for k in ("chat", "feed", "timeline", "feedJson"):     # feedJson: GET /feed.jso
         # moved, then per signature component the background rebuilds it caused; a kernel from before the
         # split reports neither field and prints the line as before
         active, bg = dl("builds", k, "active_built"), dl("builds", k, "bg_built")
+        moved = dl("builds", k, "moved")             # builds not cached because an input moved while they ran
         if active or bg:
             labels = sorted(((b.get("builds") or {}).get("chat") or {}).get("bg_miss") or {})
             causes = ", ".join("%s %d" % (lab, dl("builds", k, "bg_miss", lab)) for lab in labels
                                if dl("builds", k, "bg_miss", lab))
-            split = "; %d watched, %d background%s" % (active, bg, (": " + causes) if causes else "")
+            split = "; %d watched, %d background%s%s" % (active, bg, (": " + causes) if causes else "",
+                                                          ("; %d moved" % moved) if moved else "")
     parts.append("%s %d built / %d cached%s"
                  % (k, built, cached, (" (%.0f ms avg%s)" % (ms / built, split)) if built else ""))
 print("builds    " + "   ".join(parts))
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -973,15 +973,31 @@ The snapshot's fields, all plain numbers (`ms` is milliseconds of wall time):
   `push.*` stages count every push, including the one a connecting page gets,
   so they can add up to more than `push`.
 - `builds`: `chat`, `feed`, `timeline`, each with `cached`, `built`, `ms`.
-  `chat` also carries `active_built` and `bg_built` (rebuilds of the watched
-  tab, served while its exact key holds, against rebuilds of a background tab
-  whose signature moved) and `bg_miss`, a map from each labelled component of
-  the chat-build signature (`transcript`, `states`, `judge_gen`, `tasks`,
-  `cut`, `row`, plus `cold` for a tab with no cached build and `nosig` for
-  one whose signature could not be taken) to the background rebuilds it
-  caused. A rebuild with several moved components counts under each, so the
-  map's sum can exceed `bg_built`. `romp perf` prints the split and the
-  non-zero causes after the chat average.
+  Every chat tab, the watched one included, is served from its cached build
+  while one complete per-session signature holds: one component per input the
+  build reads (the transcript and states files, the session's goal store and
+  its journal and archive, the task store, the backend's live tail by
+  revision, its queue and brackets, the liveness row, the clock crossings the
+  payload renders, the parked ops, the account hold behind a queued bubble,
+  the retry state, the live background-task rows, the watches, the awaiting
+  stamp, the shared files, the cwd's branch and repository, the instruction
+  files, and the files and postal values the last build embedded). `chat`
+  also carries `active_built` and `bg_built` (rebuilds of the watched tab
+  against rebuilds of a background tab), `moved` (builds not cached because
+  an input moved while they ran; the next cycle builds them again) and
+  `bg_miss`, a map from each labelled component of that signature
+  (`transcript`, `states`, `store`, `hold`, `archive`, `episodes`, `reg`,
+  `gone`, `tasks`, `cut`, `live`, `row`, `clock`, `backend`, `ops`, `limit`,
+  `retry`, `bg`, `watch`, `stamp`, `anchors`, `downtime`, `names`, `flags`,
+  `ncards`, `colormap`, `acct`, `cleared`, `host`, `cwd`, `claudemd`, `fork`,
+  `taskout`, `pathlink`, `postal`, plus `cold` for a tab with no cached
+  build and `nosig` for one whose signature could not be taken) to the
+  background rebuilds it caused. A rebuild with several moved components
+  counts under each, so the map's sum can exceed `bg_built`. One session's
+  goal-store publish moves that session's `store` component and no other
+  tab's; the judge-pass generation busts the feed and timeline caches only.
+  `romp perf` prints the split and the non-zero causes after the chat
+  average, and the moved count when it is non-zero.
 - `sends`: `full`, `delta`, `deduped`, each a map from slot name (`chat`,
   `feed`, `bars`, `taborder`, ...) to `count` and `bytes`. A deduplicated frame
   was built and compared, then not sent.
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -171,6 +171,20 @@ def _process_stats():
             "pid": os.getpid()}
 
 
+# The chat-build signature's components, in the order _chat_build_sig appends them. One label per position:
+# the signature is a flat tuple of exactly this length, so a miss is attributed by comparing positions
+# (_chat_sig_miss) and /perf's builds.chat.bg_miss carries one counter per label. The last three
+# (_CHAT_SIG_DEPS) are the build-time dependencies a cached build recorded (_chat_build_deps), re-evaluated
+# per cycle over that record. _chat_build_sig's docstring says what each component is;
+# tests/test_chat_build_sig_inputs.py maps every read build_session makes to one of them.
+_CHAT_SIG_LABELS = ("transcript", "states", "store", "hold", "archive", "episodes", "reg", "gone", "tasks", "cut",
+                    "live", "row", "clock", "backend", "ops", "limit", "retry", "bg", "watch", "stamp", "anchors",
+                    "downtime", "names", "flags", "ncards", "colormap", "acct", "cleared", "host",
+                    "cwd", "claudemd", "fork",
+                    "taskout", "pathlink", "postal")
+_CHAT_SIG_DEPS = ("taskout", "pathlink", "postal")
+
+
 class _PerfStats:
     """Always-on counters behind GET /perf (`romp perf`): where the kernel's threads spend their time,
     kept cheap enough to leave running. Every writer takes the lock and does a few dict operations;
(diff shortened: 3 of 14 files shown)


ksarma:
Post-merge record: the owner merged this PR as be2e14de at head b3ee756b while this review was running. The merge stands as shipped; one should-fix goes to a follow-up, nothing needed folding.

Reviewed inline: the live-tail revision on both backends and the one complete signature. The SDK backend bumps the revision in unqueue, in edit_queued and in the dropped-echo writes, once per settling call in prune and retire. The tmux echo store keeps its own counter, and `Sessions.live_rev` dispatches to the owner. The signature keys each tab on its own store identity, so a publish busts only the owning tab. In-memory inputs are folded in by value or revision, the clock as the booleans the payload renders. The shared components are taken once per push. I read every `_mark_views_dirty` caller: each change it signals is carried by a component or is state the chat payload does not render. A bare mark therefore no longer needs to rebuild the watched tab, and nothing it covered goes uncovered. The pusher, its fault branch, `_push_session_now` and the scroll-back slice consume the dependency record, and no thread holds it. The comment-thread cache keeps its own dirty watermark. The post-build compare counts a moved input under `builds.chat.moved`, and the next cycle sees it. A raising signature is reported once per episode, and that tab builds uncached under nosig until a signature is taken. Every component is an immutable value or a fresh tuple, so the positional compare is sound. At dozens of tabs the signature costs tens of ms per push, against tens to hundreds of ms per avoided rebuild. The retired names have no dangling reference. `docs/reference.md` lists the 35 labels in code order. The two webview changes are comments only, and the added lines carry no private strings.

Follow-up, `tests/test_chat_build_sig_inputs.py:574`: the differential clock test never arms `_interrupt_clicked`. It moves the faded hour, the compaction cap and the model-switch stamp. The stage0 case that pinned the 120 s interrupt cap was dropped, and `_interrupting` is outside the census too. Dropping the first clock element would therefore pass every suite while a watched tab kept showing interrupting past the cap. The component is present at head: arming moves the key under clock, and the cap running out moves it back with the stamp popped. Add the arm-and-cap case, and correct the body line that says each clock boolean is moved.

Nits:
- `ui/webview/render.ts:13603` still describes the retired live change key.
- `kernel/kernel.py:15521`, `28718` and `28932` still say no signature sees the interrupt stamp, the parked ops or the cwd; the clock, ops and cwd components now carry them.

Tests: red-first by this review's own runs at base 8786f5f1 with the head's test files. `test_live_tail_rev.py` failed 11 of 16, the stage0 complete-key class 10 of 10, `test_chat_fixed_cost_memos` 4, `test_perf_stats` 1, plus `test_kernel`'s push-cache test and `romp-perf.bats` test 5. At commit 620c88c4, `test_chat_build_sig_inputs.py` failed 43 of 47; the four census self-checks pass. All green at head: 16, 47, 36, 89 and 342 passed, bats 15 of 15, typecheck clean, and the compiled relay-active-rearm test 5 of 5. CI: all 11 checks SUCCESS at b3ee756b.

Review run by the maintainer automated review session; verified against head b3ee756b.

PR #1278, opened 2026-09-10
Kernel: every file-path import goes through load_source, so the kernel loads clean under Python's load_module() deprecation

ksarma:
## Symptom

Every file-path import in `kernel/kernel.py`, `kernel/judge.py`, `kernel/sdk_backend.py` and `kernel/codex_backend.py` raises `DeprecationWarning: the load_module() method is deprecated and slated for removal in Python 3.12; use exec_module() instead` on Python 3.10 through 3.13 (the 3.14 interpreter's text names 3.15). Python's default filter drops it on a plain kernel start (the warning is attributed to importlib, not `__main__`), so it shows under pytest, `-W default` or `-X dev`; the test run today reports thousands of them. With the warning promoted to an error, `kernel/kernel.py` stops importing at line 26; on the release that removes the call it stops importing outright.

## Cause

The four kernel modules import their sibling modules with `SourceFileLoader(name, path).load_module()`: six loads at kernel.py's head plus the two lazy backend loads, `judge.py`'s event-model and guarded credentials loads, `sdk_backend.py`'s palette, guarded credentials and guarded session-backend loads, and `codex_backend.py`'s codex_events, codex_runtime and guarded session-backend loads. `load_module()` has been deprecated since Python 3.4, emits a DeprecationWarning on 3.10 and later (verified 3.10 through 3.14), and its removal is documented for 3.15. Nothing in the repo replaced it, and the test run has no warning filter, so the warnings have been summary noise.

## Fix

`kernel/loadsource.py` (new, stdlib only): `load_source(name, path)` builds the module from `spec_from_loader`, `module_from_spec` and `exec_module` over an explicit `SourceFileLoader` (the `bin/` names carry no `.py` suffix, which `spec_from_file_location` alone cannot map) and keeps the `sys.modules` properties the kernel and the tests rely on:

- registered in `sys.modules[name]` before it executes, so a module that loads another under the same name during its own import gets the same object;
- a name already in `sys.modules` is re-executed into the SAME module object rather than replaced, so `km.jd` is a test's `jd` and a `_rebind_state` on one is seen by the other (the semantics the shared-judge rule in `tests/conftest.py` describes), with `__file__`, `__spec__`, `__loader__` and `__cached__` re-stamped the way `load_module()` did;
- a failed first load leaves no half-built entry;
- a module that puts a stand-in for itself in `sys.modules` while its body runs hands the stand-in to its importer, as `load_module()` did.

Each of the four kernel modules bootstraps it with a four-line block (three `importlib.util` calls and the `load_source` binding) and calls `load_source` for every load (`kernel.py` also for the lazy `sdk_backend` and `codex_backend` loads). The guards that reuse an already-loaded `romp_credentials` (judge.py, sdk_backend.py) and `romp_session_backend` (both backends) are unchanged; `sdk_backend.py` gains `_HERE` in place of three inline `Path(__file__).resolve().parent` spellings.

`tests/romp_load.py` (new) exposes the loader to test modules as `from romp_load import load_source`, and `tests/__init__.py` registers it under that bare name so the import resolves under pytest and `python -m unittest tests.test_x`; a direct script run finds the file on `sys.path`. The existing test modules keep their own `SourceFileLoader` loads, which still work; converting them is a separate change.

Two test modules stubbed `km.SourceFileLoader` and now stub `km.load_source`, since the name is gone from the kernel: `tests/test_kernel_headless_ops.py` (CodexRuntimeSelection, SdkSingleFlight) and `tests/test_api_health_hover.py` (the boot_at construction test). `tests/test_state_isolation_order.py` counts a `load_source` call as a load, so a module that loads through it before setting the state floor is caught like one that loads through `SourceFileLoader`. The comments and doc lines that named the old call now name `load_source` or say "loaded by file path": `kernel/kernel.py` (the restart-classifier docstring), `kernel/judge.py` (the active_runs registry note), `kernel/codex_backend.py` (module docstring), `tests/conftest.py` (the shared-judge rule), `kernel/README.md`, `tests/README.md` and `docs/judges.md`.

Not in this change: the test modules' own loads (above), and `tools/perf-bench.py`, which still loads the kernel with `load_module()` outside both the kernel and the test tree.

## Tests

`tests/test_loadsource.py` (new, 10 tests):

- `KernelModulesLoadWithoutTheDeprecatedCall` (4): each of the four kernel modules is imported by path in a fresh interpreter (`sys.executable`) with `warnings.filterwarnings("error", message=r".*load_module\(\).*")`, the message rather than the category, so an unrelated deprecation in the kernel's imports cannot trip it, and with `XDG_STATE_HOME` pointed at a temp dir in the child. One module per child, so every guarded load executes instead of short-circuiting on an already-loaded module, and nothing touches this process's shared modules. Before the change all four fail and the child traceback names the site: `kernel/kernel.py` line 26, `kernel/judge.py` line 65, `kernel/sdk_backend.py` line 49, `kernel/codex_backend.py` line 41, verified on CPython 3.10.20, 3.11.15, 3.12.3, 3.13.14, 3.14.6 and 3.14.6 free-threaded. After the change all four pass on the same interpreters.
- `LoadSource` (6): a fresh load is registered before it executes (the module sees itself in `sys.modules` while its body runs); a second load under the same name re-executes the same object and re-stamps `__file__`, `__spec__` and `__loader__`; a failing first load leaves no entry; a failing reload keeps the existing module registered; a module that replaces itself in `sys.modules` is returned as the replacement; a path without a `.py` suffix loads. Before the change the module cannot collect (`romp_load` does not exist). Each loader line has a test that fails without it: removing the pre-registration fails four, disabling the reuse branch or making it replace the object fails two, and removing the failed-load pop, the `__file__` re-stamp, the `__spec__` re-stamp or the `sys.modules` lookup on return fails one each.

`tests/test_state_isolation_order.py::test_scan_counts_a_load_source_call_as_a_load`: `scan()` over a module that calls `load_source` before its state-floor assignment reports the call as the module's first load. Before the change it finds no load.

`tests/test_kernel_headless_ops.py` (CodexRuntimeSelection, SdkSingleFlight) and `tests/test_api_health_hover.py` (the boot_at construction test) run against the converted kernel through the `load_source` stub; left unconverted they would fail with AttributeError at the `SourceFileLoader` lookup once the name is gone.

Full suite on this head: `pytest -q -n 6 tests/` on Python 3.12.3: 11002 passed, 41 skipped, 1705 subtests passed, 0 failed, in 176 s, with the served-page modules running against Chromium. The run's 5504 warnings are all the `load_module()` DeprecationWarning, attributed to importlib's shim and to test modules and never to `kernel/`: the test modules' own `load_module()` calls (and `tools/perf-bench.py`'s, exercised by `tests/test_perf_bench.py`), unchanged here.

Tier: fix

### DIFF
--- a/docs/judges.md
+++ b/docs/judges.md
@@ -514,4 +514,4 @@ permission/API-error floors: one interrupt at a time, the present event first.
   `STATE/judge-auth.json` (the per-session judge-auth-down latch; see
   "Billing" above).
 - Debugging: run the judge's own code against the live store
-  (`SourceFileLoader` on `kernel/judge.py`) rather than inferring from logs.
+  (loaded by file path from `kernel/judge.py`) rather than inferring from logs.
--- a/kernel/README.md
+++ b/kernel/README.md
@@ -43,6 +43,6 @@ kernel before anything is spawned when a retired key path is still configured
 kernel's own environment), naming variable names and paths only. See
 `docs/reference.md` for setup and rotation.
 
-Everything here is loaded by file path (`SourceFileLoader`), not installed as a
-package; the repo runs straight from a git clone. Python tests live in
-`tests/test_*.py`.
+Everything here is loaded by file path (`loadsource.load_source`, the
+`spec_from_loader` + `exec_module` idiom), not installed as a package; the repo
+runs straight from a git clone. Python tests live in `tests/test_*.py`.
--- a/kernel/codex_backend.py
+++ b/kernel/codex_backend.py
@@ -5,7 +5,7 @@ Drives OpenAI Codex sessions through the official openai-codex Python SDK's sync
 (JSON-RPC to `codex app-server` over stdio) and materializes each thread as Claude-transcript-shaped
 JSONL via kernel/codex_events.ThreadNormalizer, so romp's entire read side parses Codex sessions
 unchanged. Duck-types the SessionBackend ABC exactly like SdkBackend does (the kernel loads backends
-via SourceFileLoader; a conformance test asserts every abstract method exists).
+by file path; a conformance test asserts every abstract method exists).
 
 Shape of the machine:
 - ONE CodexClient per backend — the app-server hosts many threads, unlike Claude's one-CLI-per-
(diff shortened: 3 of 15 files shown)


ksarma:
Merging as shipped: one follow-up, nothing to fold.

Reviewed inline: `kernel/loadsource.py` builds the module from `spec_from_loader`, `module_from_spec` and `exec_module` over an explicit `SourceFileLoader`. It keeps the `load_module()` properties the body lists. Registration happens before exec. Re-execution reuses the same object, with `__spec__`, `__loader__`, `__file__` and `__cached__` re-stamped. A failed first load leaves no entry. A module that replaces itself returns the stand-in. Against importlib's `_load_module_shim` path on 3.14.6, the differences are these. There is no per-name import lock. The reused entry is not moved to the end of `sys.modules`. A self-deleting module returns instead of raising `KeyError`. `__cached__` is None on a reload from a suffix-less path. No kernel or test site depends on any of them. The two lazy loads run under `_sdk_lock` and `_codex_lock`; import-time loads are single-threaded. The bootstrap block is identical in the four modules and uses no deprecated call. No `load_module()` or `SourceFileLoader` remains in `kernel/` outside `loadsource.py`. The child probes write only under the temp XDG state dir and open no socket. `tests/__init__.py` registers `romp_load` with `setdefault` after the state floors; nothing on the venv path carries that name.

Finding:

- `kernel/loadsource.py:53` (follow-up): the `__cached__` re-stamp has no test that fails without it; 10 passed with the line deleted. The body's per-line coverage claim overreaches for this line. One assert in the same-object test closes it.

Red-first holds by the review's own run. At base 6d1231fc with the head's test files, the four child probes fail at `kernel.py:26`, `judge.py:65`, `sdk_backend.py:49` and `codex_backend.py:41`. The six loader tests cannot import. The converted stubs fail with `AttributeError`. At head, `tests/test_loadsource.py` 10 passed. The changed modules plus `tests/test_kernel.py::ViewBuilder`, `tests/test_judge_captions.py` and `tests/test_codex_echo_merge.py`: 414 passed. CI is green at head on Python 3.10 to 3.13, bats, extension and tier checks. This head merges clean onto current main. It conflicts with the open #1275 in `tests/__init__.py`, so whichever lands second rebases.

Review run by the maintainer automated review session; verified against head d0775475.

Head note: the branch was updated with main (be76aaff) after this review ran, so the head moved from d0775475 to b956d6cf. The new head is a mechanical merge of d0775475 and be76aaff: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above covers this head.

Review run by the maintainer automated review session; verified against head b956d6cf.

PR #1282, opened 2026-09-10
The kernel runs the interpreter its SDK venv was built for, and a venv for another python is named as such, not as a missing install

ksarma:
## Symptom

A newer Python arriving on the machine takes the SDK backend down at the next kernel restart. On 2026-09-06 a `uv python install 3.14` in an unrelated shell put a `python3.14` shim in `~/.local/bin`. The next respawn ran the kernel on it while `sdkvenv` was still built for 3.12, and for two hours every SDK session showed "romp's Agent SDK backend isn't installed" over a venv that was present and intact; `romp new` refused with the same claim. The remedy both named, `bin/romp-sdk-setup`, rebuilt the venv for 3.14 without restarting the kernel, after which the session card still said mismatch while `romp new` went back to "isn't installed". A `ROMP_PYTHON` pin that named a removed path was worse: `romp-serve` died at its `exec` with bash's own "No such file or directory", the pin unmentioned, and the manager respawned it every few seconds.

## Cause

`pick_python` (one copy each in `bin/romp-serve`, `bin/romp-sdk-setup` and `bin/romp-codex-setup`) echoed `ROMP_PYTHON` unchecked and otherwise walked `python3.14` down to `python3.10` newest first, never reading the venv it was about to import into the kernel. `_ensure_sdk_on_path` added every `sdkvenv/lib/python3.*/site-packages` to `sys.path` whatever interpreter ran, so the 3.12 venv's compiled extensions failed deep inside the import under 3.14, and the fixed "not found" texts blamed a missing install. The setup scripts' rebuild check compared `pyver "$VENV/bin/python"` with the pick; `bin/python` is a symlink to the base interpreter, which follows a repointed `python3`, so the check could read as a match over a stale `lib/`. The session card and the two creation refusals were fixed strings with no view of the disk.

## Fix

- `pick_python`, byte-identical in the three scripts and pinned by one bats case that diffs them. The order: `ROMP_PYTHON` first, checked to be an executable interpreter (a bare name resolves via `type -P`; anything else exits 1 naming the pin and `service.env`). Else the interpreter the SDK venv's `pyvenv.cfg` records (`executable`, or `home` plus `version`/`version_info`), which must run and report the recorded X.Y and the build the venv's `lib/python3.X{t}` directory names; the build comes from `sys.abiflags`, never the file name, because uv links `python3.14` to `python3.14t`. Else another python of that same tag on `PATH` or in `~/.local/bin`, with one stderr line saying the venv still matches. Else the existing newest-first walk, with a stderr line asking for the rebuild.
- Each candidate the picker runs is run under `timeout 5` where coreutils has it. The picker first executes the interpreter and reads its exit; the clock ends only a probe that never answers (a stalled mount, a `sitecustomize` reaching for the network), which would otherwise hang the launch with the manager watching a live pid.
- `bin/romp-sdk-setup` and `bin/romp-codex-setup`: `venv_built_for` reads the venv's record (the cfg's version plus the `lib` directory's tag) and `pytag` reads the pick's tag; a difference rebuilds with `REBUILDING ... for python X (path): it was built for python Y` lines that name the restart that must follow. `romp-codex-setup` reads the SDK venv's cfg, not its own, so the two venvs and the kernel cannot disagree three ways.
- `kernel/kernel.py`: `_running_python_tag`; `_ensure_sdk_on_path` adds only `sdkvenv/lib/python<tag>/site-packages`, records other tags in `_SDK_VENV_BUILT_FOR` and names the mismatch once on stderr with both remedies; `_sdk_import_notice` prints the "not found" line only when no mismatch explains it; the two creation refusals (POST /new and the WS create) call `_sdk_setup_hint`, which asks the backend's verdict and falls back to the mismatch text with the rebuild remedy alone when no backend loaded. `SDK_SETUP_HINT` keeps its words and becomes that function's default.
- `kernel/codex_backend.py`: the same tag-gated join for `codexvenv`.
- `kernel/sdk_backend.py`: `sdk_venv_verdict` (`none`, `broken`, `present` or `mismatch`, with the recorded interpreter and whether it runs, a bounded `-c pass` probe), `sdk_unavailable_text` and `sdk_creation_refusal` built from that one verdict. `SdkBackend.unavailable_verdict` caches it on the venv's on-disk fingerprint, so a card render costs a few stats and the probe runs once per disk state; `unavailable_text` and `creation_refusal` read it, and so do `launch_error` and `_record_launch_error`, at request time. A venv rebuilt while the kernel runs therefore makes the card and the refusal both say "set up after romp started, restart romp". The `ROMP_PYTHON` pin is named only for an interpreter seen to run; otherwise the rebuild. The backend's construction-time log line for an unimportable SDK carries the same verdict text instead of the fixed install remedy, and `SDK_SETUP_REFUSAL` holds the plain refusal, the kernel's `SDK_SETUP_HINT` in the same words.
- Docs: `docs/reference.md` gains "The kernel's Python" (the order, the pin, the four-step move); `docs/install.md` gains "Which Python runs the kernel" (the order, the `uv python install --no-bin` hazard, the move, with the rebuild step run as `ROMP_PYTHON=<path> bin/romp-sdk-setup` because the script reads the pin from its own environment); `bin/README.md` rows for `romp-serve` and `romp-sdk-setup`.

One existing test changes its fixture. `pick_python: ROMP_PYTHON override wins verbatim` pointed `ROMP_PYTHON` at a path that does not exist and expected it echoed. The pin is now checked, so the fixture is an executable stub; the verbatim echo still holds for an executable, which is what the test was about.

## Tests

Every case runs the real code paths: fake interpreters on a bare `PATH`, a temp state root, `find_spec` stubbed so the machine's own SDK does not decide. Before the change, on the unchanged production files:

- `tests/romp-serve.bats`: 24 new cases plus the three-way picker pin, which replaces the two-way one. 23 of 41 fail before. The headline case, `a venv's recorded interpreter (executable key) beats a newer python on PATH`, prints the fake `python3.14`; the pin-check cases exit 0 on a missing path; `romp-serve: a bad ROMP_PYTHON stops the launch` sees exit 127 and bash's own error; the same-tag, abi-flags and hang-bound cases pick the wrong interpreter or say nothing. The hang case asserts on which interpreter was picked (a stub that sleeps past the bound and then exits 0 would be taken unbounded), not on elapsed time.
- `tests/install-optional-deps.bats`: 9 new cases (6 for `romp-sdk-setup`, 3 for `romp-codex-setup`). All 9 fail before: no `REBUILDING` line, a re-run rebuilds despite a matching venv, the record-vs-`bin/python` case reads a match, the missing-pin case is called a too-old python, `romp-codex-setup` builds with the newest python. The suite's setup now also unsets `ROMP_PYTHON`, sets `ROMP_STATE_DIR` to the test's own root and poisons the manager port.
- `tests/test_sdk_venv_abi.py` (new, 9): the tag-gated join, the once-per-verdict stderr line (a second call with the same venv writes nothing), the free-threaded tag, `_sdk_import_notice`, `_sdk_setup_hint` with and without a backend. 8 of 9 fail before, and the module collects on the unchanged tree: `_ensure_sdk_on_path` adds the 3.99 venv and writes no line. The no-venv case passes either way.
- `tests/test_sdk_launch_error.py`: `VenvBuiltForAnotherInterpreter` (10) and `OneVerdictForEverySurface` (7). 14 of 17 fail before: the card reads `SDK_MISSING_TEXT` over a 3.99 venv, and `creation_refusal` and the verdict functions do not exist. The three that pass either way expect the plain install text (no venv; a venv for this python with no SDK in it; a late import failure in a process that had the SDK). Among the cases: `interpreter_runs` on a stub that never exits within the bound returns False (asserted on the return value); a `pyvenv.cfg` with `home` and `version` but no `executable` names the pin; the construction-time log line carries the verdict.
- `tests/test_codex_backend.py::EnsureCodexSdk` (5): 4 of 5 fail before, the same way as the kernel's.
- `tests/test_kernel.py` (2): POST /new and the WS `createSession`, each against a backend whose `creation_refusal(default)` returns a distinct sentence, assert that sentence comes back as the error and the warn text. Both fail before: the doors returned the fixed hint.

After the change: `bats tests/romp-serve.bats` 41/41, `bats tests/install-optional-deps.bats` 23/23, the pytest modules above plus `test_loadsource.py` and `test_state_isolation_order.py` 159 passed, and the SDK-refusal cases in `test_kernel.py` 9 passed. Full pytest on this head (`pytest -q -n 6 tests/`, with the extension's `node_modules` installed so the served-page legs ran): 11148 passed, 41 skipped, 0 failed, in 2:35.

Tier: fix


ksarma:
Merging as shipped. Two follow-ups and a nit below, none a hold.

Reviewed inline: `pick_python` is byte-identical in the three scripts. It tries `ROMP_PYTHON` first, as an executable file or a bare name through `type -P`. A pin that resolves to neither exits 1 naming the pin and service.env, a clean exit under `set -e` before the exec. Second, it tries the interpreter pyvenv.cfg records, accepted only when it runs and reports the recorded X.Y and the build the lib directory names through `sys.abiflags`. Third, a same-tag python on PATH or `~/.local/bin`. Last, the newest-first walk, with its stderr line. Each probe runs under `timeout 5` where coreutils has it. With a 3.12 venv and a python3.14 shim ahead on PATH, the picker returns the 3.12 binary. A pin naming a removed path exits 1 with the romp line on the manager's inherited stderr, and the manager keeps respawning on its backoff. In the kernel, `_ensure_sdk_on_path` adds only the matching tag's site-packages and names a mismatch once per venv state. `sdk_venv_verdict` feeds the card, the log line and both creation refusals from one cache keyed on the venv's on-disk fingerprint, so a rebuild while running moves every surface to "restart romp". The `-c pass` probe is bounded and its output goes to DEVNULL. `codex_backend` carries the same gate. `install.md`, `reference.md` and `bin/README.md` match the code. The privacy scan of the 2025 added lines is clean. The fix label fits: every new test fails at the base for the reason the body gives.

- `kernel/sdk_backend.py:2843`, follow-up: the mismatch remedy offers the `ROMP_PYTHON` pin whenever the recorded interpreter exits 0 on `-c pass`. It does not check that the interpreter still reports the venv's tag; `pick_python`'s `_runs_as` does. So a recorded unversioned path repointed to another minor gets the pin, the picker refuses that binary, and the card shows the same remedy after the restart. Compare the tag in the probe and fall to the rebuild remedy otherwise, with a case in `test_sdk_launch_error.py`.
- `docs/architecture.md:67`, follow-up: still says the venv is built against the newest Python 3.10+ and rebuilt when that Python changes. That is the reverse of what `install.md` and `reference.md` now say.
- `kernel/kernel.py:14630`, nit: `_sdk_setup_hint` drops any exception from `creation_refusal` with a bare `pass`. One stderr line there keeps a future verdict failure visible.

Red-first holds. This session ran the head's test files at f22e080b: `romp-serve.bats` 23 of 41 red, `install-optional-deps.bats` all 9 new setup cases red, `test_sdk_venv_abi.py` 8 of 9, `test_sdk_launch_error.py` 14 of 17, `test_codex_backend.py::EnsureCodexSdk` 4 of 5, `test_kernel.py` 2 of 2, each for the reason the body gives. At the head: `romp-serve.bats` 41/41, `install-optional-deps.bats` 23/23, 159 passed across the five pytest modules, `test_kernel.py` refusal cases 7 passed, shellcheck clean on the three scripts. CI at head: all 11 checks green, Tier policy reporting the fix gate.

Review run by the maintainer automated review session; verified against head 60f68072.

Head note: the branch was updated with main after this review ran, so the head moved from 60f68072 to 3bec21ae. The new head is a mechanical merge of 60f68072 and e068d628 (its tree equals the merge-tree of its two parents): its added and removed lines against the merge base with main are identical to the reviewed diff. The review above covers this head.

Review run by the maintainer automated review session; verified against head 3bec21ae.

PR #1285, opened 2026-09-10
SECURITY.md names every token-exempt kernel route, the web manifest and install icons included, and the Origin check's vscode-webview:// acceptance

ksarma:
Follows https://github.com/romp-on/romp/pull/1280 (merged), whose cookie-scope paragraph this edits. The change to review is the last commit: one file, prose only.

## Symptom

SECURITY.md tells a reader that `/healthz`, `/version` and `/busy` are the only kernel routes served without the serve token. The kernel's GET handler also answers `/manifest.webmanifest` and three `/media/` icons (`romp-touch-180.png`, `romp-app-192.png`, `romp-app-512.png`) before the token gate, so the document understated the unauthenticated surface. The paragraph on the cookie's Origin check said "the VS Code webview" where `_origin_ok` accepts any origin starting with `vscode-webview://`: every VS Code webview, not only this extension's.

## Cause

The manifest and icons are served exempt of necessity: browsers fetch a web app manifest and its icon list with credentials omitted, so a gated manifest returns 403 at the moment "Add to Home Screen" consults it. The inventory sentence was not extended when they were added. The webview wording named one webview where the check matches a scheme.

## Fix

Two edits in the trust-model section, both above `## What is already hardened`:

- The inventory names the liveness probes (`/healthz`, `/version`, `/busy` on the kernel, `/ping` on the bus) and the install files (the manifest and its three icons), says why the browser needs them without credentials, and says what they are: static, reading no session state, a fixed JSON literal (app name, colors, start URL and icon list) and three PNG files. The icon route is described as a fixed allowlist of three names, not a path prefix, which is what `do_GET` checks.
- The cookie paragraph says any `vscode-webview://` origin (every VS Code webview, not only romp's own). The paragraph is rewrapped around the longer clause; the wording change is that parenthetical.

A token-less `GET /` still gets the static login page the section already describes. That is the gate's denial, not a route served before it, so the inventory does not list it.

No code or test change.

## Tests

Prose only, so no executing test can fail before it. The manual check: `git grep -n manifest.webmanifest -- SECURITY.md` is empty on the base and finds one line on this head. No source-text pin was added.

- `ui/webview/md-sanitize.test.ts` reads SECURITY.md from `## What is already hardened`, below this edit; run alone: 13 tests, 13 pass. With SECURITY.md restored to the base text it also passes (13 of 13), which confirms the slice is untouched.
- `vscode-extension`: `npm ci && npm run typecheck` pass.

No pytest or full npm run for a one-file prose change.

Tier: docs


ksarma:

Follow-up, outside this diff: `docs/guide.md:680-681` and `docs/read-side.md:728-729` still say only the liveness probes are exempt.

Reviewed inline: the kernel's GET handler at this head answers exactly five paths before `_authorize`: `/healthz`, `/version`, `/busy`, `/manifest.webmanifest`, and `/media/<name>`. The `?drain=1` and `?park=` writes on `/busy` require `_write_token_ok`, which the paragraph above covers. For `/media/<name>`, the segment after `/media/` must equal one of the three `_INSTALL_ICONS` entries, so a nested or renamed path falls through to `_authorize`. HEAD, POST and OPTIONS authorize first. OPTIONS answers an unauthenticated preflight with 403. The exempt set is therefore GET-only, and the inventory is complete without being over-inclusive. `_APP_MANIFEST` is a module-level `json.dumps` of a literal (name, two colors, a `start_url` of `/`, two icons) with no port, host or token. The bus serves only GET `/ping` before `_authorized()`. `_origin_ok` accepts any lowercased origin that starts with `vscode-webview://`, a scheme prefix match, so the cookie paragraph is now exact. Both hunks sit above `## What is already hardened`. The slice from that heading to the end of the file is byte-identical to the base, so `md-sanitize.test.ts` reads unchanged input. Body and commit message match the diff. The 13 added lines carry no private strings, hostnames or home paths.

Locally, in capped scopes, `npm run typecheck` passes and `ui/webview/md-sanitize.test.ts` alone passes 13 of 13. CI at head is 11 of 11 green. The Tier policy check reports docs merges on green for every author.

Review run by the maintainer automated review session; verified against head 11596467.

ksarma:
Head note: the branch was updated with main, so the head moved from 11596467 to b939fb5c. The new head is a mechanical merge of 11596467 and 0658d2aa: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head b939fb5c.

PR #1286, opened 2026-09-10
A SIGTERM nobody asked for leaves a row and a cut reason, and the manager notes every kill it sends

ksarma:
## Symptom

A kernel that receives a SIGTERM nobody requested through the manager (a stray `kill`, a `systemctl --user stop`, a `launchctl bootout`) exits with an empty `reason` on its restart-cuts.jsonl row and writes nothing to restart-audit.jsonl. The manager respawns it with the same `exited (code= sig=) respawning` line it prints for a restart it sent itself, so the log gives no hint the restart was unrequested. A kernel whose manager disappears exits from `_parent_watch` through a bare `os._exit(0)` with no row in either file. An outage that begins with such a restart has no first cause on disk: one did, when a direct signal to the kernel pid respawned it onto a different python and every SDK session died at import behind a message that blamed a missing install.

## Cause

Three gaps. `bin/romp-manager` sent SIGTERMs without a row of its own, so a `romp-manager restart` or a service stop reached the kernel indistinguishable from a stray kill, and the kernel could not tell "the manager asked" from "nobody did". `_graceful_term` read the audit tail AFTER the drain, so a row landing during the drain (the manager's own stop, a second signal) was taken as the request for the signal that preceded it, and with nothing on record the cut row's reason was `""`. `_recent_restart_audit` walked past only `_NO_RESTART_ACTIONS`, so a request older than this kernel (the self-update that restarted its predecessor, still inside the 90 s window) or a predecessor's own exit row would name a later cut. A second SIGTERM mid-drain nested a second handler inside the first, which wrote a second, partial cut row.

## Fix

`bin/romp-manager`: `auditSigterm(spec, pid, reason, trigger)` appends `{t, action: manager-sigterm, kernel, pid, reason, trigger}` to the kernel's own state root's restart-audit.jsonl before every SIGTERM in `restartKernel` (trigger `restart`), `restartAll` (`restart-all`), the stale-manager self-bounce (`refresh`) and `stopKernel` / `shutdownAll` (`stop`, from `stopTrigger()`, decided once per shutdown so every kernel's row names the same cause). The registry record remembers a requested restart (`requested`), and `exitLine` (pure, exported) says `stopped`, the plain respawn line, or `exited without a restart request (signal or crash); respawning`. The note is a mechanism row, not an intent: the requester's own row (bin/romp's refresh row, the kernel's self-update, the dashboard's http-restart) is written before it, and the reader treats the note accordingly.

`kernel/kernel.py`: `_graceful_term` takes `_EXIT_ONCE` (a second SIGTERM returns at once), sets `_TERMINATING` first as before, reads `_recent_restart_audit()` at SIGNAL time and hands the row to `_drain_and_exit`, the drain-write-exit tail now shared with `_parent_watch`, which files a `parent-gone` row (`_audit_parent_gone`) before it. `_recent_restart_audit(window, now, started)` walks the last 200 rows newest first: it skips `_NO_RESTART_ACTIONS`; stops at rows older than this process's start (`_STARTED`, whole seconds) except a parked `when: quiet` request, which the manager delivers to whichever kernel runs at the window (dropped once a manager note for a restart or for the manager's own shutdown, a manager-gone verdict or a consuming cut row shows the park delivered or gone; a `stop` note with trigger `stop` aimed at another kernel's pid settles nothing, since `POST /stop` naming one kernel leaves the park armed and the row cannot be told from the stop that took every kernel down, so the park stays on record, the error to prefer: a held converge self-heals at the backstop bound, a released one cuts the turns the quiet window was to spare); skips the kernel's own verdict rows (`EXIT_VERDICT_ACTIONS`: `signal`, `parent-gone`); a request row wins unless a cut row already consumed it (the T240 `auditT` rule, kept); a `manager-sigterm` note about this pid answers only when no request row precedes it, with `manager-sigterm: <trigger>` from `_audit_reason_text`; an unlabeled row (bin/romp's refresh row) is skipped, except an unlabeled `when: quiet` row, the park `_parked_quiet_deploy` reads through this same walk, which is returned when nothing labels the exit and yields to the note when there is one. With nothing on record, `_unrequested_signal_reason` decides `managerStopped` from `ROMP_MANAGER_PID` liveness (read before the wait and once more after it, since a manager older than the note writes none and can exit during it) or from the manager's own stop note landing after the signal (`_manager_sigterm_row_for_us`, over the same 200-row tail the reader uses, so a dozen end-on-idle rows landing after the note do not hide it), files a `signal` row (`signal`, `pid`, `ppid`, `managerPid`, `managerRequested: false`, `managerStopped`, `managerRestartPending` from the backend's drain lease) and returns the reason the cut row carries: `signal, not requested through the manager` or `signal; the manager was stopped too (a service stop or restart)`. The sender's pid is not recorded; a Python signal handler does not receive it, and the wording claims only what the kernel can see.

On the one time constant: `SIGNAL_MANAGER_NOTE_WAIT_S` (0.5 s) bounds how long an unexplained SIGTERM waits for the manager's stop note. The note is the event the code looks for first: the manager writes it before it kills, and node dispatches its own SIGTERM handler after the kernel's Python one has run, so during a service stop the note lands moments after our signal. The wait only caps that look; a genuinely stray signal pays it once and reads as unrequested.

Unchanged: `_TERMINATING` and the T240 stand-down in `_parent_watch`; `main()`'s `signal.signal(SIGTERM)` followed by `jd._cred.check_boot_environment()`; the 800 ms exit in `shutdownAll`; the `_DEPLOY_RESTART_REASONS` cool-down (a note is not a deploy reason, and a request row beneath a note is still the request, so a self-update's cut row keeps `auditT` equal to the request's `t`).

Docs: `docs/reference.md` (Where things live) describes both files' rows and the reader's rules; `bin/README.md`'s romp-manager row names the note and the unrequested-exit log line.

The test module's loader is `from romp_load import load_source` (tests/README.md), in place of the deprecated `SourceFileLoader(...).load_module()` the module had.

## Tests

Every kernel test runs against the module's own rebound state root (`_own_state_root`) with fake or reaped pids (`os.kill(pid, 0)` only, or `_pid_alive` stubbed), `_broadcast_restarting` stubbed and `os._exit` patched to raise, so nothing signals a real process; the conftest floors the manager and kernel ports.

- `tests/test_restart_cuts.py::UnrequestedSignal` (23). `test_the_handler_files_the_row_and_the_cut_reason_when_nothing_asked` runs the real SIGTERM handler over an empty file; before the change it fails on `Lists differ: [] != ['signal']` (no audit row, and the cut row's reason was `""`). The others: the row's fields; `managerRestartPending` read from a backend's `drain_holding()` (False for a backend without it); the wording for a live-and-silent manager, a gone manager pid, a manager whose pid goes during the wait without a note (`_pid_alive` stubbed True then False), a stop note landing after the signal, a restart note landing after it (not a stop), a note about another kernel, a predecessor's stop note for a reused pid, the `_STARTED` default, a stop note behind twelve end-on-idle rows; a second SIGTERM mid-drain writes one row; the request is read at signal time, not after the drain; an unlabeled row alone is neither the request nor consumed (the cut row has no `auditT`; a `signal` row is filed); a note alone is consumed like a request row (`auditT`); a predecessor's `signal` or `parent-gone` verdict is not this kernel's request.
- `tests/test_restart_cuts.py::RequestOnRecord` (21). The reader over exact row orders: the CLI's unlabeled refresh row plus the manager's note, a labeled request beneath a note, a labeled request beneath an unlabeled row (the request answers and is what the cut consumes), the note alone, an unlabeled row alone (nothing), old rows, a note for another pid, each trigger label, the start bound (an older request is a predecessor's; the same second is kept; a request consumed inside this kernel's lifetime is not named again), quiet requests older than the kernel (yield to the note; name the cut when no note does; dropped when a restart note, a `refresh` stop note or a manager-gone verdict shows them delivered or gone; kept past a single-kernel `stop` note for another pid and past a stray-kill verdict with the manager alive; the `RESTART_EXPECT_MAX_S` bound), verdict rows never the request, and the unlabeled quiet park (returned while nothing labels the exit; the note labels its delivery; a delivery to a predecessor is not this kernel's).
- `tests/test_restart_cuts.py::ParentGone` (3). `test_the_exit_leaves_both_rows` runs `_parent_watch` with a reaped pid as `ROMP_MANAGER_PID`; before the change it fails on `0 != 1` (no `parent-gone` row, no cut row). Also the standalone no-op and the stand-down under the handler's lock.
- `tests/test_restart_cuts.py::OwnStateRoot` (1): the module's state-root helper.
- `tests/test_restart_cuts.py::CutRow`: two existing pins move with the code. `test_sigterm_handler_writes_the_ledger` pins the new handler shape (`_drain_and_exit(_audit_reason_text(rec), signum=signum`, `reason = _unrequested_signal_reason(signum, be)`, `audit_reason=reason`, the `_EXIT_ONCE` acquire) in place of `audit_reason=_audit_reason_text(rec)`; `test_reason_joins_the_recent_audit_tail_only` passes `started=900`, since its synthetic rows predate the process and the reader now bounds at the kernel's start.
- `tests/manager-exit-attribution.test.js` (17; runs under CI's `node --test tests/manager-*.test.js`). `exitLine`'s three verdicts and the crash case; `auditSigterm`'s row, an aux kernel's own root, never throws; `restartKernel`, `restartAll` and `stopKernel` seed the registry with a stand-in child whose `kill` records the audit rows as they stood, so row-before-signal is asserted on order, with the trigger per door and `requested` / `stopping`; `stopTrigger` and a caller-supplied trigger. In child node processes, with the state root at a temp dir and the ports poisoned: `shutdownAll` (it exits the process) with trigger `stop` and with a caller's; the stale-manager self-bounce (`restartAllOrSelf` on a copy of the manager whose mtime moves after the require, under `ROMP_SUPERVISED`) writes `{reason: stop, trigger: refresh}` before the kill; and `spawnKernel`'s exit handler, with `ROMP_SERVE_BIN` at a launcher that exits at once, logs `exited without a restart request (signal or crash); respawning` for an exit nothing requested, the plain line after `requested` is set, and `stopped` for a record marked stopping. Before the change all 17 fail: none of the functions were exported. Temp roots are removed at exit.

Runs: `pytest tests/test_restart_cuts.py` with this branch's tests over the unchanged `kernel/kernel.py` and `bin/romp-manager`: 50 failed (4 of them subtests of one case), 10 passed; after: 56 passed, 12 subtests. `node --test tests/manager-*.test.js`: 96 passed (17 new). Each of the reader and handler rules named above was checked by removing it and watching a test go red (the unlabeled-row skip, the in-lifetime consumption stop, the drain-lease read, the note tail depth, the liveness re-read, the single-kernel-stop park rule, the exit handler's use of `exitLine`, the self-bounce's `refresh` trigger). Full `pytest -n 6 tests/` on this tree with the extension's node_modules installed: 11163 passed, 41 skipped, 1717 subtests passed in 2:52, no failures.

Tier: fix


ksarma:
Merging as shipped. The four findings below are follow-ups; none blocks the merge.

Reviewed inline: by reading only, since the change sends signals to processes; CI at the head ran the tests. The only local runs were `ast.parse` of `kernel/kernel.py` and `node --check` of `bin/romp-manager`. `_graceful_term` takes `_EXIT_ONCE` non-blocking, so a second SIGTERM returns instead of nesting a handler. It sets `_TERMINATING`, reads the audit tail at signal time, and hands the row to `_drain_and_exit`. Its `finally` writes one cut row, then calls `os._exit`. `_parent_watch` takes the same lock and files its parent-gone row before the shared drain. I traced `_recent_restart_audit` by hand for the CLI row plus the manager's note, a labeled request beneath a note, and an other-pid stop note over a parked quiet request. Each returns what the body says. The manager's `auditSigterm` is a synchronous append inside try/catch to the kernel's own state root, called before every kill. `stopTrigger` is decided once per shutdown. The self-bounce writes reason `stop` with trigger `refresh`. `requested` resets on each spawn. `exitLine` has its three verdicts. An older kernel reads a manager-sigterm row as a plain request without error. A newer kernel under an older manager waits the half-second look once. That constant caps the wait for the manager's note and decides nothing on its own. Docs match the code. The added lines carry no private strings, hostname or home path. The `fix` label fits.

Follow-ups:
- `kernel/kernel.py:21871`. The window and start-bound breaks run before the walk-past rules. A stop note for another pid, or a signal verdict with `managerStopped` false, therefore ends the walk when it is older than 90 s or older than this kernel. That hides a live quiet park beneath it. The pinned test probes only inside both bounds. Classify such rows with `continue` before the bounds and pin the two out-of-bound cases.
- `kernel/kernel.py:53483` (present at base). A stray SIGTERM with a quiet park on record returns the park and stamps its `t` as `auditT`. A sibling kernel under the same state root then consumes a park the manager still holds, and the released stand-down converges early. Skip the stamp when the row is a quiet park with no manager note for this pid.
- `tests/manager-exit-attribution.test.js:82`. The never-throws case names a path `mkdirSync` recursive can create, so the catch is not exercised. Use an existing regular file as a path component.
- `kernel/kernel.py:21638`. The stderr write in `_unrequested_signal_reason` is unguarded and runs before the cut row. Wrap it as `_audit_unrequested_signal` wraps its own write.

Test evidence: by rule, nothing ran here. CI at 98b77e68 is green on all 11 checks. The Python jobs cover 3.10 to 3.13; on 3.12, 11084 passed and 121 skipped, with `tests/test_restart_cuts.py`'s 56 cases. The Shell job's `node --test tests/manager-*.test.js` step passed 96 of 96, with the 17 new cases. The red-first claim holds by reading: the base manager exported none of the functions the node tests call, and the headline kernel cases assert rows the base handler and `_parent_watch` never wrote.

Review run by the maintainer automated review session; verified against head 98b77e68.

ksarma:
Head note: the author merged main (0658d2aa) into the branch as 4ce08138 (parents 98b77e68 and 0658d2aa; the reviewed commit is not rewritten). Verified against the merge base: the added and removed lines are identical to the reviewed diff in `bin/README.md`, `bin/romp-manager`, `docs/reference.md`, `kernel/kernel.py` and `tests/manager-exit-attribution.test.js`. In `tests/test_restart_cuts.py` the two loader-import lines (`SourceFileLoader` replaced by `load_source`) left the diff because main carries them from #1283; nothing else changed there. The merge commit's tree differs from a mechanical merge of its parents only in `bin/README.md`, where the conflict between #1282's `romp-serve` row and this PR's `romp-manager` row is resolved by keeping both rows in table order. Auto-merge is armed on this head; it lands when the checks finish.

Review run by the maintainer automated review session; verified against head 4ce08138.

ksarma:
Post-merge delta record: the owner merged this PR as f0597caa at head f36ddba8 (the review-fix commit pushed after the verified main-merge 4ce08138) while this delta review was running; auto-merge was never armed past 4ce08138. The commit is sound and closes follow-up 4 of the earlier review. The items below are follow-ups; none would have held it.

Reviewed inline: by reading only, since the code signals processes. CI ran the tests. In `_drain_and_exit` the inner try wraps only the helper call. The row build, `auditT`, `drainError` and `_append_restart_cut` run on both paths. The fallback `SIGNAL_REASON_UNREQUESTED` is the constant `_audit_unrequested_signal` returns when the manager is not stopped. The row therefore reads as the docs describe for that case. The helper's stderr guard wraps the one write with its format inside; `return reason` stays outside it. The test patches the helper on the loaded module. The call site names it as a module global at call time, so the patch takes. `_fire` calls the handler in-process with the broadcast stubbed, no backend and `os._exit` patched to raise. It sends no signal, forks nothing and starts no manager. With the audit file unlinked and no manager pid in the environment, the reader returns None. The handler takes the unrequested branch, and an empty drain result gives `cutTurns` of `[]`. At 4ce08138 the same raise reached the ledger's outer except, which printed and skipped the row. The test fails there on 0 != 1. The added lines carry no private strings, hostname or home path, and the loader lines are main's.

Follow-ups:
- `kernel/kernel.py:53574`. The new except records the swallowed fault nowhere: no stderr line, no row field. The same block records a drain fault as `drainError`. When the helper raises before `_audit_unrequested_signal` runs (a malformed `ROMP_MANAGER_PID`), the cut row carries the plain verdict. No `signal` row exists, and nothing on disk says why. Set `row["reasonError"]` from the exception class, print inside a guarded write, and pin the field in the new test.
- `kernel/kernel.py:21721`. The stderr guard has no test; the new test replaces the helper wholesale. Without the guard, a raising stderr after the audit row lands leaves that row saying the manager was stopped. The cut row then says unrequested. A direct call with a reaped manager pid and a raising `km.sys.stderr` pins it.
- `kernel/kernel.py:53557` (present at base). The drain announcement sits inside the try before `be.drain`. A raising stderr therefore skips the drain in the scenario the new comments name. The ledger handler's write at 53584 runs unguarded before `os._exit`. Guard both. The manager spawns the kernel with inherited stdio, so a supervisor closing stderr, the cause the comments name, does not close the descriptor. A reset journal stream or a closed tty does.
- The PR body counts `UnrequestedSignal` at 23 cases and the module at 56 passed. After this commit they are 24 and 57.

Test evidence: locally only `py_compile` of both files in a capped scope, by rule. Red-first holds by reading, as above. CI at f36ddba8 is green on all 9 checks, Python 3.10 to 3.13 included. The 3.12 job reports 11143 passed, 121 skipped, 1719 subtests.

Review run by the maintainer automated review session; verified against head f36ddba8.

PR #1288, opened 2026-09-10
Comment threads: a signature that raises on the thread path is said once per fault episode, the chat loop's rule

ksarma:
## Symptom
A comment thread whose change signature cannot be taken (one of the reads behind `_chat_build_sig` raises: a torn state file, a directory that cannot be listed) is rebuilt on every pusher cycle, and nothing says so: no stderr line, no bell row. The same exception from the same function on a chat tab is reported once per fault episode, with the traceback on stderr and a refused bell row on the dashboard.

## Cause
`_thread_events` (kernel/kernel.py) takes the thread's signature inside a `try` whose handler only sets `key = None`. The chat loop takes the same signature under `_chat_sig_ok` / `_chat_sig_fault`. On the thread path a `None` key is always a swallowed exception: `_sdk_sess` always hands a thread a transcript path, so `_chat_build_sig`'s own `None` return (no path) is unreachable there.

## Fix
The handler becomes `except Exception as e: _chat_sig_fault(sess, e); key = None`, and a signature that is taken calls `_chat_sig_ok(tsid)` so the episode closes. The thread still builds uncached until the read succeeds; the fault is said once per fault text per thread, on stderr with the traceback and as one refused bell row. `sess` is the `_sdk_sess` dict (`sid` is the thread sid, `name` the reg's name or the sid's first eight characters), the shape `_chat_sig_fault` reads.

`_chat_sig_faults` is keyed by sid, and the two writers never meet on one. A thread stays out of the chat loop's session list on both of its sources: its names entry is withheld, so discover never lists it, and the backend's live set skips a threadOf reg. Once promotion writes the names entry and clears threadOf, `_comments_frame` stops handing the thread to `_thread_events`. The promote step does those two writes in sequence, so one pusher cycle inside it could run both paths for the sid; the worst case there is one report said twice or suppressed once in that cycle.

`_chat_sig_fault` itself is unchanged, so the bell row reads "the pane for <thread name> cannot be cached"; a noun for the thread case would touch the chat path and is left out of this fix. The `_thread_events` docstring gains a sentence.

## Tests
- `tests/test_comment_threads.py::ThreadProjection::test_a_signature_that_raises_on_the_thread_path_is_said_once_per_episode` (new). `_watch_awaiting`, a read the signature makes unconditionally, is replaced by one that raises for the thread. The first frame writes one `push build: chat signature <sid>` line to stderr, followed by the traceback and the fault text, and files one `refused` bell row naming the thread; the second and third frames write nothing more, while `build_session` is attempted on each frame and nothing is stored under the thread; a different fault text is a new episode (a second line, a second row); a signature that is taken ends the episode (the thread leaves `_chat_sig_faults` and the build is cached); the first fault re-armed after that is said anew. The injected read also fails the real `build_session` for the thread (through `_session_awaiting`), so the test asserts nothing on the frame's events while the fault stands; every assertion holds either way. Every replaced function and module table is restored in `finally`. Before the change it fails at the first assertion: the first frame writes nothing (`0 != 1`), and no bell row is filed. Two further checks were run and turn it red: without the `_chat_sig_ok` call it fails at "a signature that is taken ends the episode"; with a plain stderr write in place of `_chat_sig_fault` it fails at "the second frame, same fault: not again".
- The three existing thread-cache tests in the class stay green: a build that raises under a good signature is still uncached; that path is outside this fix.
- Targeted: `tests/test_comment_threads.py`, `tests/test_chat_build_sig_inputs.py`, `tests/test_pusher_change_gates.py`: 173 passed.
- Full `pytest tests/` on this commit (six workers, the served-page modules running against Chromium): 11116 passed, 41 skipped, 0 failed, 1705 subtests passed.

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -13466,7 +13466,12 @@ def _thread_events(tsid, cut_uuid, now, tmux):
     the task store, the pending cut, the backend's live revision and queue, the snapshot row, the
     shared files), without the dependency tail (a thread records no build dependencies), plus the
     thread's own states row; the serve yields to _views_dirty for the dependencies that tail would
-    have carried. A thread with no keyable input (no transcript yet) is built every time, never cached."""
+    have carried. A thread with no keyable input (no transcript yet) is built every time, never cached.
+    A signature that RAISES (one of its component reads failed) is said once per fault episode, on
+    stderr with the traceback and as a refused bell row, the chat loop's rule for the same signature
+    (_chat_sig_fault; a signature that is taken ends the episode, _chat_sig_ok), and the thread is
+    built uncached until a signature is taken: before this, the fault was swallowed into an uncached
+    build and left no trace anywhere."""
     reg = _thread_reg(tsid)
     if reg.get("forkOf"):
         return []
@@ -13477,6 +13482,7 @@ def _thread_events(tsid, cut_uuid, now, tmux):
     key = None
     try:
         sig = _chat_build_sig(sess, tm, now, tmux=tmux, deps=False)
+        _chat_sig_ok(tsid)                          # a signature that was taken ends its fault episode
         if sig is not None:
             # plus the thread's OWN state rows (review 2026-09-08): the backend writes states/<tsid>.jsonl under
             # the romp sid, while the key above stats states/<fsid>.jsonl for the reg's lastSid (_sdk_sess hands
@@ -13489,7 +13495,10 @@ def _thread_events(tsid, cut_uuid, now, tmux):
             except OSError:
                 states = None
             key = sig + (states,)
-    except Exception:
+    except Exception as e:
+        _chat_sig_fault(sess, e)                # once per fault episode: stderr and a bell row, the chat loop's rule.
+        #                                         _chat_sig_faults is sid-keyed; the chat loop builds only a PROMOTED
+        #                                         thread and the frame never hands one here, so no entry is shared.
         key = None                              # an input we cannot key → build, never cache
     hit = _built_thread.get(tsid)
     if key is not None and hit is not None and hit[0] == key and hit[1] == cut_uuid and _views_dirty[0] <= hit[3]:
--- a/tests/test_comment_threads.py
+++ b/tests/test_comment_threads.py
@@ -10,6 +10,7 @@ order. All fixtures SYNTHETIC: invented text, placeholder UUIDs.
 """
 import contextlib
 import errno
+import io
 import json
 import os
 import shutil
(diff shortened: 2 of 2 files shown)


ksarma:

Reviewed inline: `_thread_events` now calls `_chat_sig_fault` in the handler that used to set only `key = None`. A taken signature calls `_chat_sig_ok`. A thread whose signature read raises is therefore reported once per fault text: one stderr traceback and one refused bell row. It still builds uncached until the read succeeds. The sid-keyed ledger is not shared between the chat loop and the thread path. The chat loop lists sessions from discover and the backend's live set. Both skip a thread (names entry withheld, threadOf reg skipped). `_comments_frame` stops calling `_thread_events` once the store row reads promoted. The only overlap is the promote step's few consecutive writes, where one report is said twice or suppressed once in that cycle. `_sdk_sess` hands the thread a sid, a name and a non-empty transcript path, the shape `_chat_sig_fault` reads. `_chat_build_sig`'s None return is therefore unreachable there, and a None key is always a reported fault. The ledger holds one short string per faulting sid, popped when a signature is taken. The added lines carry no private strings, hostnames or home paths.

Two wording nits, no code change:

- The body and commit name the names and threadOf writes as what stops `_comments_frame`. The gate is the store status at `kernel/kernel.py:13772`, written at `:14428` after `promote_thread` returns. The window therefore spans three writes.
- The docstring at `kernel/kernel.py:13469` keeps a sentence about a no-transcript thread. The body calls that case unreachable.

Tests: at base f22e080b, the head's `tests/test_comment_threads.py` fails only the new case, at its first assertion (0 != 1). The three existing thread-cache tests stay green (1 failed, 3 passed). At head, the three targeted files give 173 passed. Both stated mutations go red where the body says. CI: 13 of 13 checks SUCCESS at head.

Review run by the maintainer automated review session; verified against head f4fd5087.

ksarma:

Reviewed inline: `_thread_events` now calls `_chat_sig_fault` in the handler that used to set only `key = None`. A taken signature calls `_chat_sig_ok`. A thread whose signature read raises is therefore reported once per fault text: one stderr traceback and one refused bell row. It still builds uncached until the read succeeds. The sid-keyed ledger is not shared between the chat loop and the thread path. The chat loop lists sessions from discover and the backend's live set. Both skip a thread (names entry withheld, threadOf reg skipped). `_comments_frame` stops calling `_thread_events` once the store row reads promoted. The only overlap is the promote step's few consecutive writes, where one report is said twice or suppressed once in that cycle. `_sdk_sess` hands the thread a sid, a name and a non-empty transcript path, the shape `_chat_sig_fault` reads. `_chat_build_sig`'s None return is therefore unreachable there, and a None key is always a reported fault. The ledger holds one short string per faulting sid, popped when a signature is taken. The added lines carry no private strings, hostnames or home paths.

Two wording nits, no code change:

- The body and commit name the names and threadOf writes as what stops `_comments_frame`. The gate is the store status at `kernel/kernel.py:13772`, written at `:14428` after `promote_thread` returns. The window therefore spans three writes.
- The docstring at `kernel/kernel.py:13469` keeps a sentence about a no-transcript thread. The body calls that case unreachable.

Tests: at base f22e080b, the head's `tests/test_comment_threads.py` fails only the new case, at its first assertion (0 != 1). The three existing thread-cache tests stay green (1 failed, 3 passed). At head, the three targeted files give 173 passed. Both stated mutations go red where the body says. CI: 13 of 13 checks SUCCESS at head.

Review run by the maintainer automated review session; verified against head f4fd5087.

PR #1291, opened 2026-09-10
File browser: a browse click the viewer's discard confirm refuses stands down whole

ksarma:
## Symptom

A browse click that the viewer's discard confirm refuses still builds the browser beneath the kept viewer. Browse files while the viewer holds unsaved edits (a folder link in the chat, the tab menu's Browse files row, or a viewer's own directory link), and answer the discard confirm with cancel. The viewer stays, as asked, but the browser's overlay is built beneath it, a listing is fetched for the folder and painted under the opaque viewer, the body reads as browsing, and a keydown handler is bound for a browser hidden under the viewer. Closing the viewer later lands on a listing that was never wanted. When the browser was already up beneath the viewer (a file opened from its listing, then edited), the refused click swaps the listing under the viewer to the other folder and session, resets Hidden, and paints the loader over the rows the person will come back to.

## Cause

`openFileBrowse` builds the overlay, calls `closeFileView()` and asks for the listing unconditionally. `closeFileView` returns nothing; its `closeGuard` (the discard confirm) can leave the viewer up, and nothing read that outcome.

## Fix

After the close attempt, a viewer still up means the click stands down whole. An overlay built for this click is taken down again by `unbuild`: the element, the body class and the keydown handler. It sends no `browseClosed` (the notice a browser that was up sends on closing) since nothing opened, and resets no latch since no ask went out. Nothing else runs either: the session, the Hidden reset and the ask move after that check, so a refused re-invoke over an existing listing leaves it as it was, its session and Hidden state included. No notice follows; the person has just answered the confirm. The overlay is still built before the close attempt, as before, so a browse that goes through is unchanged, and the module header names the stand-down beside the one-directional stack it belongs to.

## Tests

`ui/webview/filebrowse.test.ts` gains three cases that run the real `openFileBrowse` and `openFileView` against a DOM stand-in (the `file-view-text-size.test.ts` idiom: a tree, dataset, bubbling events, a selector matcher, a fetch answering the kernel's headers), taking the viewer into edit mode through its own Edit button and the plain textarea, with `window.confirm` answered by the test. Synthetic fixtures only.

- A browse under a kept editor, with a framing shell listening for the browser's close notice: no overlay, no listing asked, no keydown handler bound, no `browseClosed` to the shell, the edits kept; then, edit mode left, the next browse opens as ever, and closing that browser sends the shell its one `browseClosed` (the same shell stand-in that heard nothing at the stand-down). Before the change it fails at "no browser overlay beneath the kept viewer": the overlay is built under the viewer. With `closeFileBrowse()` in place of `unbuild()` it fails at "no close notice to the shell for a browser that never opened".
- A browse over a listing the kept viewer covers, for another folder and session: the same overlay, no listing asked, rows, crumbs and Hidden unchanged; the viewer closed, a Hidden press re-asks the original path for the original session. Before the change it fails at "no listing asked for the other folder": a `listDir` goes out for the other folder and session. With the session or the Hidden reset moved back ahead of the check it fails at the re-ask's session or at "Hidden still on".
- The confirm accepted: the viewer closes and the listing is asked, so the stand-down keys on the refusal and never on the edits alone. Passes before and after.

The file's 22 existing source pins are unchanged and pass. `npm run typecheck` passes. Full `npm test`: 4099 tests, 4099 passed, 0 failed, 0 skipped (the main run at concurrency 8, 4077 tests, plus `timeline-tags-scale.test.js` alone, 22 tests; both exit 0).

Tier: fix


ksarma:

Reviewed inline: `openFileBrowse` at the head builds the overlay, tries `closeFileView`, then reads `document.getElementById('romp-fileview')` a second time. That read is settled, because `closeFileView` is synchronous end to end. Its `closeGuard` is `confirmDiscard`, a blocking `window.confirm`, and `wrap.remove()` runs inline with no animation or deferred removal. So the read is true exactly when the confirm was refused. Before the close attempt the function sets only three things: the overlay element with element-local listeners, the `filebrowse-open` body class, and the document keydown handler in `onKeyRef`. The new unbuild removes those three and nothing else, with the same ref and target the bind and `closeFileBrowse` use. It sends no `browseClosed` and resets no latch, correctly: nothing opened and no ask went out. The session assignment, the Hidden reset and `ask()` now sit below the check. A refused re-invoke over a listing already beneath the viewer therefore fetches nothing, paints no loader, and keeps its session and Hidden state. The second test pins this through a later Hidden press that re-asks the original path and session. The refused click's only acknowledgement is the confirm the person just answered with cancel. They chose that outcome, viewer kept and edits intact, so no further notice is owed. The module header names the stand-down beside the one-directional rule it extends rather than reverses, so the fix label fits. The body matches the diff and the added lines scan clean.

Test evidence:
- The head's `filebrowse.test.ts`, run against base f22e080b in a scratch worktree, fails exactly the two new cases: "no browser overlay beneath the kept viewer" and "no listing asked for the other folder" (25 tests, 23 pass, 2 fail; the confirm-accepted case and the 22 pins pass).
- At head it passes 25/25, and 83/83 with the file-view neighbours; `tsc --noEmit` is clean.
- The three described mutations, and two more on unbuild, each fail at the named assertion.
- CI at head: all 11 checks SUCCESS.

Review run by the maintainer automated review session; verified against head 0bde04f8.

ksarma:
Head note: the branch was updated with main, so the head moved from 0bde04f8 to f1214b91. The new head is a mechanical merge of 0bde04f8 and 41c2d5f4: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head f1214b91.

PR #1292, opened 2026-09-10
Judges: a Fast judging setting adds Claude Code's fast-mode opt-in to Opus judge calls

ksarma:
A kernel setting, **Fast judging** in the gear's Judges section (off by default), runs judge calls whose model is Opus in Claude Code's fast mode. The judges are `claude -p` subprocesses, and the CLI refuses fast mode to a non-interactive client unless the flag-settings layer carries the `fastMode` opt-in; sessions get that key through `flag_settings_path`, and the judges had no way to ask for it, so a triage or distilling tier pinned to Opus always ran at normal speed.

## Design

- **A kernel setting in the shape of the tmux-backend and comment-fast settings.** `STATE/judge-fast` holds `on` or `off`; `_set_judge_fast` validates and gesture-stamps it; it sits in `_JUDGE_SETTING_FIELDS` and `_GT_STORES`; `/version` reports it raw at the top level and in the cross-machine `settings` dict; `/judge-settings` applies and acks it; the `setJudgeFast` socket op reads a boolean `enabled` through `_as_bool`, stores on/off, and fans the applied value out to every linked kernel under its gesture stamp, standing down on a stale one with a `settingStale` frame. The gear row is a label-row checkbox with the mixed mark, a stamped emitter under its own store name, `STALE_LABELS` and `STALE_TYPE` entries, and `KERNEL_SETTING` carries `setJudgeFast`, so one click writes every attached kernel like the other judge settings.
- **Decided per call, by model family.** `_judge_cmd` adds `fastMode: true` when the setting is on and the call's model reads as the opus family through `_model_family_version`: the bare `opus` alias or a pinned Opus version id, since the judge tiers accept version ids. A call on Sonnet, Haiku or Fable gets the argv it always got, byte for byte. There is no triage-model gate on the checkbox: the mechanism is per call, and the distilling or indexing tier may be Opus while triage is not. `cmtFastGate` is the pattern if a gate is wanted.
- **One `--settings` overlay.** `--settings` takes one value, so the fast opt-in and a login-billed call's `apiKeyHelper: ""` share one `json.dumps` overlay when both apply. The login-only overlay is the same string as before, byte for byte, so the auth-billing pins hold unchanged. No settings file is written.
- **The CLI's answer is recorded.** Whether fast engaged is per account: an account with extra usage turned off, or an organisation with fast mode disabled, reports it off with the setting on. Each `judge-usage.jsonl` row now keeps the envelope's `fast_mode_state` as `fast` (`on`, `off`, `cooldown`, or `null` when the CLI reported none), so the setting has a readback and a checkbox that reads on beside rows that read off names the account, not the setting.
- **Cost.** Fast mode bills Opus at a premium (twice the standard Opus rate at the time of writing) and draws on fast mode's own rate limits, the pool sessions' fast toggles share. The row copy and docs/reference.md say so. The setting ships off, and nothing turns it on for an existing install.
- **Considered: a per-install setting** (the thinking-summaries shape). The judge model and effort picks already follow to every connected machine, and a judge tier that runs Opus on one machine and Opus in fast mode on another is the per-machine surprise the propagation rule exists to prevent, so this follows too. If per-install is preferred, the change is dropping the `KERNEL_SETTING`, `_JUDGE_SETTING_FIELDS` and fan-out entries; the rest stands.
- **Left out on purpose.** No triage-model gate on the checkbox, no static settings file, no third state beyond on/off.

Docs: docs/reference.md gets a Fast judging section after Judge concurrency; docs/judges.md lists `STATE/judge-fast` under Ops and knobs and names the usage row's `fast` field.

## Tests

Executing tests, each red before the change:

- `tests/test_kernel_judge_model.py::FastJudging` (5 tests): off by default with the argv untouched and `/version` reporting `off`; the opt-in on the `opus` alias and on two pinned Opus version ids with the sonnet, haiku and fable argv equal to the toggle-off argv, `--effort` still beside the overlay, and a login-billed Opus call riding one overlay with both keys while the sonnet login argv keeps the verbatim helper-off string; the setter's value space (`yes`, `true`, `1`, `""`, `ON` refused unwritten), `_GT_STORES`, `_JUDGE_SETTING_FIELDS` and `/version`; the socket op storing on/off, fanning out `{"judgeFast": ..., "gt": ...}`, standing a stale gesture down with a `settingStale` frame naming `judge-fast`, and refusing a non-boolean with a `warn`; the usage row keeping `fast`. Before the change: `AttributeError: module 'romp_judge' has no attribute '_judge_fast'`, and `_judge_cmd` emits no `--settings` outside the login branch.
- `tests/test_judge_auth_billing.py`: `test_a_fast_login_billed_call_rides_one_overlay_with_both_keys` (one overlay parsing to `{"fastMode": true, "apiKeyHelper": ""}`; the sonnet login argv still ends in `HELPER_OFF`; key-billed and unpicked Opus calls carry the opt-in alone) and `test_a_fast_login_pick_on_opus_launches_with_one_overlay_and_logs_the_readback` (end to end through `_judge_run` with a fake CLI: one overlay, the login tokens, no key, and a usage row reading `("opus", "on")`). Before the change the Opus login argv carries only the helper key and the usage row has no `fast` field.
- `tests/test_judge_settings_sync.py::ApplySettings::test_judge_fast_applies_and_reports_raw`: `{}` acks `off`; `on` lands and acks; `true` is ignored with the ack still `on`; `off` clears. Before the change: `KeyError: 'judgeFast'`. The fan-out pin gains `args=({"judgeFast": _jfv, "gt": _jgt},)`.
- `tests/test_setting_gesture_order.py`: the gt-store count 16 -> 17, `_set_judge_fast("on", gt=T)` reported under `judge-fast` with nothing else moving, and `setJudgeFast` in the frames-and-report vocabulary walk.
- `tests/test_kernel_session_flags.py`: a `setJudgeFast` row in the `_as_bool` table (every non-boolean refused unwritten with the warn frame; real booleans apply and flip back).
- `tests/test_kernel_settings_sections.py`: `rs-judgefast` between Judges and Updates & debug; a label row in the Fast comment threads shape with the mixed mark; the `.rs-jrow` count unchanged at nine.
- `ui/webview/multi-kernel-merge.test.ts`: `setJudgeFast` in the routeOutbound list, so the op reaches every attached kernel with the same message to each; red with the op missing from `KERNEL_SETTING` (it falls through to the local kernel alone).
- `ui/webview/gear.test.ts`: the stamped emitter under `judge-fast`, the `STALE_LABELS` and `STALE_TYPE` entries, the mixed-mark pair, the fill from the raw value, and the `KERNEL_SETTING` member; the existing completeness pin over federation's set covers the emitter as well.

Runs on the final tree: the six modules above plus the suite's loader rule, 208 passed; `npm run typecheck` clean; `gear.test.js` 18/18 and `multi-kernel-merge.test.js` 56/56; full `pytest -n 6 tests/` and full `npm test` (the main leg plus the timeline-tags-scale file alone): full `pytest -n 6 tests/` 11148 passed, 41 skipped, 0 failed, 1707 subtests passed (the served-page modules running against Chromium); full `npm test` 4097 tests, 4097 pass, 0 fail (4075 in the main run plus the 22 of the timeline-tags-scale file run alone).

Tier: feature

### DIFF
--- a/docs/judges.md
+++ b/docs/judges.md
@@ -459,6 +459,8 @@ permission/API-error floors: one interrupt at a time, the present event first.
 
 - Toggles: `CLOSER_ON`, `GROUPER_ON`, `DISTILLER_ON`, `CONSOLIDATE_ON`.
   Models: `STATE/judge-model` (triage), `STATE/index-model`.
+  Fast judging: `STATE/judge-fast` (`on` | `off`, off by default; read per
+  call, and the fast-mode opt-in rides only a call whose model is Opus).
   Pool width: `STATE/judge-concurrency` (the gear's Judge concurrency, 1..16,
   read fresh each pass; empty = `ROMP_JUDGE_CONCURRENCY` as read at load,
   else 6). Every pool reads it at call time (`_conc`, or `_judge_concurrency()`
@@ -486,7 +488,8 @@ permission/API-error floors: one interrupt at a time, the present event first.
   the sessions the evidence gate ran. Outside
   a pass frame (`romp-judge --plan`) the evidence gate stamps nothing, and
   the inner gate does the skipping.
-- Logs: `STATE/judge-usage.jsonl` (per-call cost, one name per prompt),
+- Logs: `STATE/judge-usage.jsonl` (per-call cost, one name per prompt, and
+  the CLI's `fast_mode_state` as `fast`),
   `STATE/judge-errors.jsonl` (the row contract above; kinds are parse,
   call, give-up, sweep-cut, cite-miss, rate-limited, task-store, history-unreadable,
   task-key-collision (a duplicated to-do mirror key, reconciled per node
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -485,6 +485,24 @@ through to `install.sh`:
   and follows to every connected machine like the other judge settings; its
   Default option clears the setting back to the variable, else 6.
 
+### Fast judging
+
+- **Fast judging** (the gear's Judges section; off by default) runs the judges
+  in Claude Code's fast mode, an Opus-only research preview billed at a premium
+  over standard Opus rates. The setting is read per call: a judge call whose
+  model is Opus, by the bare alias or a pinned Opus version, carries the CLI's
+  fast-mode opt-in in its per-call settings; a call on any other model runs
+  exactly as before, so with every tier on Sonnet and Haiku the setting changes
+  nothing until a tier is pinned to Opus. Fast requests draw on fast mode's own
+  rate limits, the pool your sessions' fast toggles share. Whether fast engaged
+  is the CLI's answer, per account (an account with extra usage turned off, or
+  an organisation with fast mode disabled, reports it off with the setting on):
+  each row of `judge-usage.jsonl` keeps that answer in its `fast` field (`on`,
+  `off` or `cooldown`; `null` when the CLI reported none), so a checkbox that
+  reads on beside rows that read off names the account, not the setting. Like
+  the other judge settings, a change applies on the judges' next pass with no
+  restart and follows to every connected machine.
+
 ### Session backends
 
 - **Enable Claude Code tmux backend** (the gear's Updates & debug section; off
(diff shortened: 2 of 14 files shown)


henrypinkard (approved):
Approved on head e40ac1d4. A follow-up on top will show the checkbox only where a judge tier's model can run fast mode, say when it is inert, and take up the review's four concerns (key-billed permission, silent refusal, readback, cost view).

henrypinkard (approved):

PR #1295, opened 2026-09-10
Docs: the architecture page states the kernel's Python and the SDK venv's rebuild cases, bin/README lists romp-codex-setup, and bats covers a uv-built venv's pyvenv.cfg

ksarma:
## Symptom

`docs/architecture.md`'s installer section said the SDK venv is "built against the newest Python 3.10+ on the machine and rebuilt when that Python changes". `docs/install.md` ("Which Python runs the kernel") and `docs/reference.md` ("The kernel's Python") say the opposite: the kernel follows the venv, installing a newer interpreter does not move it, and `ROMP_PYTHON` is the way to move it. A reader crossing the three pages found two rules.

`bin/README.md`'s table of real files has no row for `bin/romp-codex-setup`, which has been there since the Codex backend landed and was edited when the venv-first picker shipped. (`bin/romp-uninstall` also has no row; this change does not add one.)

`tests/romp-serve.bats` has twenty `pyvenv.cfg` fixtures and every one writes `version =` or no version line; none writes `version_info =`, the key uv writes (uv also writes no `executable =` line). The `version_info` alternative in `pick_python` can be deleted from all three copies (`bin/romp-serve`, `bin/romp-sdk-setup`, `bin/romp-codex-setup`) with the suite still 41/41 green. `tests/install-optional-deps.bats` has the same gap for `venv_built_for`'s read of that key.

## Cause

The architecture clause predates the venv-first picker and was not updated with `install.md` and `reference.md`. The README row was never added. The bats fixtures were written from the stdlib's `venv` output, so the uv layout (which the picker's comment and `kernel/sdk_backend.py` both name as supported) had no case.

## Fix

Commit 1, docs only:
- `docs/architecture.md`: the clause now reads "built on one Python 3.10+ and rebuilt by `bin/romp-sdk-setup` only when you move romp to another Python by setting `ROMP_PYTHON`, when the interpreter it was built with is gone and no other of its version and build remains, or when an interrupted build left it without pip; the kernel runs `ROMP_PYTHON` if set, else the interpreter the venv's `pyvenv.cfg` records." Each of the three cases is the script's: a pin of another tag fails the mismatch check and rebuilds (a pin of the same minor and build is kept); a recorded interpreter that is gone with no same-tag python on PATH or in `~/.local/bin` sends `pick_python` to the newest-first walk, whose tag differs; the pip gate rebuilds a venv with `bin/python` but no `bin/pip`. The sentence matches the wording the same clause has in the downstream clone that ported it, so the two copies stay identical. The section names `ROMP_PYTHON` before `pyvenv.cfg` and no longer contains the word "newest".
- `bin/README.md`: a `romp-codex-setup` row after `romp-sdk-setup`, in the table's shape: what it provisions (`codexvenv` with the pinned `openai-codex` SDK, and the pinned Codex runtime under the state directory), that it shares the python pick with `romp-serve` and `romp-sdk-setup` by reading the SDK venv's record, the same rebuild for a different interpreter tag, the `~/.local/bin/codex` link for login when no `codex` is on PATH, and that `install.sh` does not run it (`docs/codex.md`'s once-per-machine step). No version pins in the row.

Commit 2, tests only, nothing under `bin/` changes:
- `tests/romp-serve.bats`: `pick_python: a uv-built venv (home plus version_info, no executable) still names its interpreter`. A cfg with `home`, `implementation`, `uv`, `version_info` and `include-system-site-packages` (a real uv venv's keys), a `python3.14` first on PATH and the venv's `python3.12` under `home`: the pick is `home/python3.12` with nothing on stderr.
- `tests/install-optional-deps.bats`: `romp-sdk-setup: a uv-built venv (home plus version_info, no executable) is followed and kept, not rebuilt`. The same cfg shape with a newer python first on PATH and no `lib/` directory under the venv (with one, `venv_built_for` takes the tag from `lib/python3.X` and the cfg read would go untested; the case says so): exit 0, no `REBUILDING`, no newest-python line, no venv built.

## Tests

- Commit 1: none; prose only. No test reads `docs/architecture.md` or `bin/README.md`, so nothing can fail before the change.
- `tests/romp-serve.bats`, new case: fails before on the pick assertion when the `version_info` alternative is removed from `pick_python`: the picker returns `python3.14` from the newest-first walk and writes "was built with <home>/python?, which is gone, will not run, or is no longer python ?, and no other python ? was found; picking the newest python instead". With the read in place: `home/python3.12`, stderr empty. The unchanged suite stayed 41/41 with the read removed from all three copies, which is the gap.
- `tests/install-optional-deps.bats`, new case: fails on `[[ "$output" != *"REBUILDING"* ]]` with the `[_a-z]*` key match dropped from `venv_built_for`'s sed (rebuilds "for python 3.12", "built for python ?"), and again with the picker's alternative dropped in that script's copy (rebuilds for 3.14).
- Green with the copies restored: `tests/romp-serve.bats` 42/42, `tests/install-optional-deps.bats` 24/24. No TypeScript is touched, so no typecheck; no pytest module reads the two docs, so no pytest; no full suites run for a docs change plus two bats cases.

Tier: docs


ksarma:

Reviewed inline: the new architecture clause against `bin/romp-sdk-setup` and `bin/romp-serve` at the head. Each of the three rebuild cases maps to a code path. A `ROMP_PYTHON` pin of another tag fails the built-versus-wanted compare. A recorded interpreter that is gone, with no same-tag python on PATH or in `~/.local/bin`, sends `pick_python` to the newest-first walk. That walk's tag then differs. The pip gate rebuilds a venv with `bin/python` but no `bin/pip`. The kernel's order in the clause (`ROMP_PYTHON`, else the interpreter `pyvenv.cfg` records) matches `pick_python`. The word newest is gone from the page. Every claim in the new `bin/README.md` row checks against `bin/romp-codex-setup`, `install.sh` and `docs/codex.md`. The uv fixture matches the real layout. A venv built here with uv 0.12.1 wrote home, implementation, uv, version_info and include-system-site-packages, with no executable line. Leaving out `lib/` in the install-optional-deps case is the right construction. `venv_built_for` takes the tag from `lib/python3.X` when one exists, and reads the cfg only when none does. The three `pick_python` copies stay byte-identical. Nothing under `bin/` or `kernel/` changes. The `docs` label fits a change of docs plus tests. Privacy scan of the 48 added lines: clean.

Both new cases fail under the mutations the PR body names; no other test does:
- version_info alternative dropped from the picker: `tests/romp-serve.bats` is 41 ok, 1 not ok on the pick assertion.
- sed key match dropped from `venv_built_for`: `tests/install-optional-deps.bats` is 23 ok, 1 not ok on REBUILDING.

At the head: romp-serve.bats 42/42, install-optional-deps.bats 24/24. CI at head: 11 of 11 checks green.

Review run by the maintainer automated review session; verified against head 84a00985.

PR #1297, opened 2026-09-10
The dashboard drops ?token= from its address once the cookie is set, and every page the kernel serves carries Referrer-Policy: same-origin

ksarma:
## Symptom

The dashboard's token rides in the document URL for as long as the page is open. A browser opens the dashboard as `/?token=<token>` (the login page's `location.replace`, `romp url`, the CLI's open); the response sets the `romp_token` cookie; the URL keeps the token. A document URL is what the page's requests send as `Referer`, what every same-origin pane iframe reads as `document.referrer`, and what the address bar shows. The shell itself makes no cross-origin request today, so nothing leaked from it, but a pane page opened bare as `/chat?token=` renders transcript `` loads from other hosts, and under the older `no-referrer-when-downgrade` default an `http://` dashboard sends the full URL, token included, to each image host. SECURITY.md says a cross-site page cannot obtain the token; that held only because modern browsers default to `strict-origin-when-cross-origin`. The kernel enforced nothing.

## Cause

Nothing removes the token from the URL after the response spends it, and the kernel sets no referrer policy. `do_GET` on `/` with a valid `?token=` calls `_authorize`, which returns the token as the cookie to set, and serves `_landing()` directly through `_send`, which adds `Set-Cookie` to that same 200. No redirect strips the query (Jupyter, whose model the token gate follows, redirects). The only `history.replaceState` in the shell is the push deep-link strip in `_LANDING_REVEAL_JS`, which deletes `push-reveal` and `push-card` and leaves `token`. `_send` sends `X-Content-Type-Options`, `X-Frame-Options` and `Content-Security-Policy: frame-ancestors` and no `Referrer-Policy`; no served page carries a `<meta name=referrer>` or a `referrerpolicy` attribute.

## Fix

Two hunks in `kernel/kernel.py` and one sentence in SECURITY.md.

The shell's head script, the one that mints the dashboard id before the parser reaches an `<iframe>`, gains one statement: delete the `token` param and `history.replaceState` the rebuilt URL (path, remaining query, hash), inside a `try`. The response that served the page already turned the token into the cookie, so the URL copy is spent; the statement runs in the head, ahead of `<link rel=manifest>` and the first `<iframe>`, so no request this document makes carries the token. The other params survive (`panes`, `wid`, a `push-reveal` deep link the reveal script strips later in the body), re-serialized by `URLSearchParams`: a comma becomes `%2C`, which the `panes` reader's `searchParams.get` decodes. It sits inside the existing `<script>`, so the shell's script count and the mint-before-`<body>` pins hold, and the reveal block is untouched (#1219 edits that line).

`_send` adds `Referrer-Policy: same-origin` after the CSP line. Every document the kernel serves (the shell, the pane pages, `/sw.js`, the login page) then sends the full `Referer` to its own origin and nothing cross-origin, whatever the browser's default; the manifest, the install icons and the other assets carry the header too, since every response through `_send` does. The value is `same-origin` rather than `no-referrer` because of #1219: its `_request_page_origin(headers)` reads the `Referer`'s origin when a request carries no `Origin` (a same-origin GET such as the page's `/push/pending`, the worker's fetch) for the Apple Declarative Web Push `navigate` URL. `same-origin` keeps that header on same-origin requests; `no-referrer` would remove it. The `/file` GET, HEAD and range paths and `_remote_file` write their own headers outside `_send` and do not carry the new header; their URLs carry no token, and the SECURITY.md sentence is scoped to pages for that reason.

SECURITY.md's sentence now says what makes the claim hold: the dashboard drops `?token=` from its address as it loads, and every page the kernel serves carries `Referrer-Policy: same-origin`, so the token never reaches another origin in a `Referer`. It says nothing about browser history: the guaranteed effect of `replaceState` is on the document URL.

A reload after the scrub rides the cookie. The pane iframes already do (their `src` is a bare pane path, no token), so a browser that refuses the cookie never had a working dashboard, and none regresses. Checked at this tree: nothing reads the token from `location` after load (the kernel-served scripts read `wid`, `push-reveal`, `push-card` and `panes`; the webview reads `wid` and the host-injected `__rompKernelToken`); nothing reads `Referer` or `document.referrer`, #1219's same-origin fallback aside. Not changed here: a pane page opened bare as `/chat?token=` keeps the token in its address bar; the header keeps it out of any cross-origin `Referer`, and a same-origin request from that page still carries the full URL to the kernel itself.

## Tests

`tests/test_kernel_auth_hardening.py`, new class `TokenLeavesTheUrl`, both executing:

- `test_the_shell_drops_the_token_from_its_address_before_any_request`: the shell's first `<script>` runs under node against stubs of the globals it touches (`sessionStorage`, `document`, `navigator` through `Object.defineProperty`, `location`, `history`), one subtest per address. On `/?token=abc&keep=1#frag`, `replaceState` is called once with `/?keep=1#frag`; with no token it is not called; with the token alone it is called with `/`; with `&push-reveal=S1` the deep link survives; with `&panes=chat,feed#frag` the rebuilt query is `panes=chat%2Cfeed` (the re-serialization the fix relies on; a rebuild that decoded the query fails this case alone). The scrub text sits inside the first script and ahead of `<link rel=manifest` and `<iframe`. Before the change: `replaceState` is never called (`[] != ['/?keep=1#frag']` and the three other token cases).
- `test_every_page_the_kernel_serves_carries_referrer_policy_same_origin`: the real `do_GET` serves `/?token=<token>` (the bootstrap response, which also carries `Set-Cookie`), `/chat` and `/media/romp-swirl-glyph.svg` with `Referrer-Policy: same-origin`. Before the change: the header is absent (`None != 'same-origin'`); with `no-referrer` as the value it fails too.
- `ResponseHardeningHeaders.test_send_sets_nosniff_and_frame_guards` pins the header's line in `_send` beside the other three. Before the change: red.

`_serve_get` in that module now wraps a `_serve_get_full` that also returns the response headers.

Targeted run of `tests/test_kernel_auth_hardening.py tests/test_kernel_mobile.py tests/test_per_viewer_focus.py tests/test_kernel_webpush.py tests/test_kernel_cors.py tests/test_state_isolation_order.py`: 214 passed (9 subtests). The mobile module's `<script>` count, the mint-before-`<body>` pin, the reveal block's `replaced == ['/?keep=1#frag']` pin and the loader check are in that run.

Full `pytest tests/` on this commit: 2 failed, 11173 passed, 41 skipped, 1714 subtests passed in 174 s (`pytest -q -n 6 tests/`, with the extension's node modules installed so the served-page tests run). The two failures are `tests/test_perf_stats.py::PushStages::test_stages_and_chat_builds_are_recorded_by_the_real_push` and `tests/test_notification_tap_resume_browser.py::ServedTapResume::test_a_notification_shown_but_never_tapped_is_offered_when_the_app_comes_forward`, timing tests in modules this change does not touch. Both pass alone (2 passed, and 3 of 3 in two of three repeats), and both also fail intermittently on the current main without this change, on the same machine at a load average near 30 (1 of 3 repeats each). The browser test does load the shell and so runs the head script; its failure is a reveal the kernel parked because the chat pane had not booted yet, a timing path, and the earlier full run of the identical kernel hunks passed it (11174 passed, 41 skipped, no failures). Neither is called green here; both are reported as load failures.

Tier: fix


ksarma:

Reviewed inline: the scrub is one try-wrapped statement in the shell's existing head script, ahead of the manifest link and the first iframe. It rebuilds the URL from pathname, remaining query and hash in a single `replaceState`. A throw leaves the address untouched, and no partial rewrite is possible. No kernel-served script, webview module or extension source reads the token from the location after load; the readers are wid, panes, push-reveal and push-card. The reveal script's own `replaceState` still yields its pinned URL with the token gone. Every HTML document the kernel serves goes through `_send`, including the login page, the 413 page and `/sw.js`. The writers outside it emit no document: file HEAD and range replies, downloads, the preflight, the websocket upgrade, the remote relay. The `SECURITY.md` sentence is accurate. `same-origin` is the right value over `no-referrer`: the pane iframes are same-origin documents, and no cross-origin request the pages make depends on a Referer. `no-referrer` would blank #1219's page-origin read of same-origin GETs. A reload rides the `Path=/ SameSite=Lax HttpOnly` cookie and the manifest `start_url` is a bare `/`, so nothing regresses. The node harness stubs every global the head script touches and unlinks its temp file in `finally`.

Verification:
- Red-first at the base with the head's test file: `TokenLeavesTheUrl`'s two tests and the `ResponseHardeningHeaders` pin fail for the stated reasons (`replaceState` never called, header `None`, pin substring absent); 48 others pass.
- At head, the module plus `test_kernel_mobile`, `test_per_viewer_focus`, `test_kernel_webpush`, `test_kernel_cors` and `test_state_isolation_order` give 214 passed, 9 subtests.
- A `no-referrer` mutation fails only the header test and its source pin; a `decodeURIComponent` mutation fails the panes case alone.
- Privacy scan of the 120 added lines is clean.
- CI at 9abb73c2: all eleven checks green.

Review run by the maintainer automated review session; verified against head 9abb73c2.

ksarma:
Head note: the branch was updated with main, so the head moved from 9abb73c2 to 85be4c6e. The new head is a mechanical merge of 9abb73c2 and 6480f4da: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 85be4c6e.

PR #1299, opened 2026-09-10
The mismatch remedy names the ROMP_PYTHON pin only for a recorded interpreter that still runs as the venv's python, and a creation refusal whose verdict raises is said on stderr

ksarma:
## Symptom

An SDK session's card over a venv built for another python can prescribe a fix that does not work. The venv's `pyvenv.cfg` records an interpreter path; when an upgrade repoints that path at another minor (a `python3` the distro moved from 3.12 to 3.14, or a recorded binary replaced in place by another build), the card and the `romp new` refusal say to set `ROMP_PYTHON` to that path and restart romp. `bin/romp-serve` honors the pin as given, the kernel comes up on the same other python, and the same card returns after the restart with the same remedy. The rebuild, the remedy that works, is never shown.

A smaller hole beside it: when the backend's `creation_refusal` raised, `_sdk_setup_hint` swallowed it with a bare `except` and fell to the plain install hint, so a failing verdict downgraded `romp new` and the browser's create with no trace of why.

## Cause

Three predicates decided whether "the venv's interpreter still runs", and the one that chose the remedy was the loosest. `bin/romp-serve`'s picker (`_runs_as`) follows the recorded interpreter only when it reports the venv's X.Y and build; the kernel adds the venv's site-packages only for its own running tag (`_ensure_sdk_on_path`); but the kernel's remedy probe (`interpreter_runs`) accepted the recorded path on a clean exit of `-c pass` alone, reading no version or build. Reachability is narrow: CPython's venv records a versioned realpath, uv's cfg has no `executable` line and the kernel's fallback from its `home` and version is a versioned `home/python3.X`, so the trigger is a hand-edited or third-party cfg, an in-place build swap, or a 3.14/3.14t swap at one path. The value is one predicate across the three probes.

## Fix

- `kernel/sdk_backend.py`: `interpreter_tag(path, timeout=5.0)` replaces `interpreter_runs`. It runs the interpreter with the one-line program `running_python_tag()` evaluates in this process and `bin/romp-sdk-setup`'s `pytag` runs (`%d.%d%s` from `sys.version_info` and `sys.abiflags`), bounded as before, and returns the tag when the exit is 0 and the stripped last stdout line is a whole tag (`3\.\d+t?`), else `""`. A sitecustomize that prints ahead of the tag does not defeat it; a script that prints anything else does not pass for a python. `sdk_venv_verdict` records `interp_tag` (test seam `probe`) and drops `interp_runs`; nothing else read the key. `_mismatch_remedy` names the pin only when `interp_tag` is one of the venv's built tags, the same test `_ensure_sdk_on_path` applies to the kernel's own interpreter, and the rebuild otherwise. `sdk_venv_fingerprint` and `SdkBackend.unavailable_verdict` are unchanged: the fingerprint keys on the recorded interpreter's path, mtime and size, read through a link, so a repointed link re-derives the cached verdict (tested below).
- `kernel/kernel.py` `_sdk_setup_hint`: the bare `except Exception: pass` becomes one stderr line naming the exception type and message (`sdk-backend: the backend's creation_refusal failed (RuntimeError: ...); refusing from this process's own reading of the venv`), then the unchanged fallback. The write is guarded on its own, so a message that will not format still leaves the create door its answer.
- `docs/reference.md`, "The kernel's Python": the remedy sentence now says the pin is named when the recorded interpreter still runs as the venv's python (the kernel runs it and reads its version and build), the rebuild otherwise.

Not in this PR: the bash pickers (`pick_python`'s pin branch honors `ROMP_PYTHON` as given, by design; the docs say to pin a versioned path), `bin/README.md`, and `docs/architecture.md`'s newest-Python sentence (a docs change of its own).

## Tests

Executing tests, each red on the unchanged tree unless said otherwise:

- `tests/test_sdk_launch_error.py::VenvBuiltForAnotherInterpreter::test_a_recorded_interpreter_that_reports_another_tag_names_the_rebuild`: the class's 3.99 venv records an executable stub that prints this process's tag (a `#!/bin/sh` stub receives `-c ...` as positional parameters, so it prints and exits 0 under the old probe and the new). Before: `'ROMP_PYTHON' unexpectedly found in "... Set ROMP_PYTHON=<state>/py399/python3 in service.env, then restart romp ..."`. After: the rebuild line, and the verdict's `interp_tag` is this process's tag.
- `tests/test_sdk_launch_error.py::OneVerdictForEverySurface::test_mismatch_with_an_interpreter_of_another_tag_offers_the_rebuild_on_both`: the card and `creation_refusal` both name the rebuild and neither names the pin. Before: the same pin on both surfaces.
- `tests/test_sdk_launch_error.py::OneVerdictForEverySurface::test_mismatch_with_an_interpreter_of_the_other_build_offers_the_rebuild_on_both`: the recorded interpreter prints `3.99t` over a venv built for `3.99`; both surfaces name the rebuild and `interp_tag` is `3.99t`. This is the build half of the predicate, the case `_runs_as` refuses as well. Before: the pin on both surfaces.
- `tests/test_sdk_launch_error.py::OneVerdictForEverySurface::test_a_repointed_recorded_interpreter_moves_the_cached_verdict_to_the_rebuild`: the trace on one backend. The cfg records `python3`, a link to a stub that prints `3.99`, and both surfaces name the pin; the link is repointed at a stub of another size that prints this process's tag, and the next ask on the same backend names the rebuild on both surfaces with `interp_tag` this process's tag. Before: the pin again after the repoint.
- `tests/test_sdk_launch_error.py::VenvBuiltForAnotherInterpreter::test_the_remedy_follows_the_tag_the_probe_reports`: through `sdk_venv_verdict`'s `probe=` seam, four answers over the 3.99 venv: `3.99` names the pin; this process's tag, `3.99t` and `""` name the rebuild; the probe is asked about the recorded path. Before: `TypeError` (no `probe` keyword).
- `tests/test_sdk_launch_error.py::VenvBuiltForAnotherInterpreter::test_the_probes_program_is_the_tag_this_process_computes`: `_TAG_PROGRAM` run in this process prints `running_python_tag()`; under a patched free-threaded 3.99 (`sys.abiflags`, `sys.version_info`) it prints `3.99t`, the build half of the expression, which the probe cannot exercise through a default-build machine's own interpreter. Before: `AttributeError` (no `_TAG_PROGRAM`).
- `tests/test_sdk_launch_error.py::VenvBuiltForAnotherInterpreter::test_the_probe_reads_a_tag_from_the_last_line_and_nothing_else`: noise ahead of the tag still yields the tag; padding is stripped whether the tag is the only line or the last of several; a blank line after the tag is not the answer; a free-threaded tag is read whole; a non-tag answer, an empty answer, a tag followed by more words, and a tag with a failing exit all report `""`. Before: `AttributeError` (no `interpreter_tag`).
- `tests/test_sdk_venv_abi.py::EnsureSdkOnPath::test_a_raising_creation_refusal_is_said_on_stderr_and_the_hint_still_answers`: a backend whose `creation_refusal` raises `RuntimeError('cfg unreadable')`; the hint is `SDK_SETUP_HINT` and stderr holds exactly one line naming `creation_refusal`, `RuntimeError` and the message; then the same over a mismatch this process saw itself (the rebuild-only fallback, no pin). Before: `0 != 1 : one stderr line: ''`.
- `tests/test_sdk_venv_abi.py::EnsureSdkOnPath::test_a_verdict_failure_the_report_cannot_format_still_gets_the_fallback`: the raised exception's `__str__` raises; the hint is still `SDK_SETUP_HINT`. Green on the unchanged tree (its bare `except` swallows everything); red against this change with the stderr write unguarded (`ValueError` out of `_sdk_setup_hint`).

Fixture updates in the same module: the stubs that stand for the venv's own python print its tag (`#!/bin/sh` then `echo 3.99`) in the present-and-running case, the home-plus-version case, the boot-log case and `OneVerdictForEverySurface._interp`'s default; the present-and-running case also asserts the verdict's `interp_tag` and that `interp_runs` is gone; the not-running case's probe assertions read `interpreter_tag` (`""` for a failing exit, a missing path, and a stub that prints the tag and then never exits, with `timeout=0.5`; this interpreter's tag for `sys.executable`); the cache test wraps `interpreter_tag`. Red before the change: 10 of the 70 cases in the two modules and the seam table's four subtests (the four behavioural cases on the remedy text and the stderr line; the rest on the probe's new name, the program's name, the seam and the verdict's new key).

What these tests catch when the code is altered: dropping the per-line strip or the whole-output strip in `interpreter_tag` (the parsing case), dropping the `abiflags` clause from `_TAG_PROGRAM` (the program case), comparing the minor alone in `_mismatch_remedy` (the other-build case and the `3.99t` subtest), dropping `timeout=` from the probe's `subprocess.run` (the not-running case, which then reads `3.99` from the hang stub after 30 s), and removing the guard around the stderr write (the unformattable-message case). Dropping the `os.access` pre-check changes no outcome: `subprocess` raises and the `except` returns `""` the same way.

Targeted run: `tests/test_sdk_launch_error.py tests/test_sdk_venv_abi.py tests/test_codex_backend.py` plus every module that reads `docs/reference.md` (`test_api_health_hover.py`, `test_api_health_rail.py`, `test_perf_stats.py`, `test_postal_hydrate_scope.py`, `test_reference_billing_label.py`): 321 passed, 16 subtests passed.

Full suite (`pytest -q -n 6 tests/`, with the extension's node_modules installed so the served-page legs run): 11183 passed, 41 skipped, 1711 subtests passed in 178 s, no failures.

Tier: fix


ksarma:

Reviewed inline: `interpreter_tag` runs the recorded interpreter with the same one-line tag program that `running_python_tag` evaluates in-process and `bin/romp-sdk-setup`'s `pytag` runs. It runs with stdin at DEVNULL and the same 5 s bound as the old exit-0 probe. It returns a tag only when the exit is 0 and the stripped last stdout line is a whole 3.X or 3.Xt. A non-zero exit, a hang, a missing or non-executable path, or any other output reads as empty. `_mismatch_remedy` names the ROMP_PYTHON pin only when that tag is one of the venv's `lib/python3.*` tags. That is the same membership test `_ensure_sdk_on_path` applies to the kernel's own interpreter. A pin it names therefore brings up a kernel that matches the venv. A repointed python3, a build swap or an empty answer names the rebuild. The verdict stays cached on the venv fingerprint, and every caller goes through `unavailable_verdict`. The probe therefore runs once per disk state, not per card or request. In `_sdk_setup_hint` the bare pass becomes one guarded stderr line naming the exception type and message. The fallback text is unchanged. `interp_runs` has no other reader at head. The `reference.md` sentence matches the code, and `install.md` already agreed. Privacy scan of the 279 added lines is clean. Fixtures are shell stubs under temp dirs with synthetic tags.

Red-first, by our own runs: at base 41c2d5f4 with the head's two test files, 14 failed, 60 passed. Those are exactly the ten cases and four seam subtests the body names, with the stated errors. The unformattable-message case is green at base, as claimed. At head, both modules: 70 passed, 4 subtests. Each of five mutations was caught by exactly the named test. The body's targeted set: 321 passed, 16 subtests. CI is green on all 11 checks at head. The Tier policy check reports the fix gate.

Review run by the maintainer automated review session; verified against head 6816f93a.

ksarma:
Head note: the branch was updated with main, so the head moved from 6816f93a to 16a0c7ef. The new head is a mechanical merge of 6816f93a and 5c8ccebe: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 16a0c7ef.

PR #1302, opened 2026-09-10
Restart audit: a live quiet park survives an aged verdict or note, a stray SIGTERM does not consume it, and the manager's never-throws test runs its catch

ksarma:
Follows the merged #1286 and lands on main: the reader rules that PR added to `_recent_restart_audit`, the unrequested-signal path beside them, and the manager-side `auditSigterm` test are what the three commits below correct, one item per commit. The romp down feature PR that follows is stacked on this one; it extends the same reader with `down` and `down-failed` rules and will rebase onto this. For that rebase: the first commit creates a block at the top of the walk holding every rule that continues without answering, before the window and the start bound. The `down-failed` continue and the superseded-`down` continue belong in that block (both decide without the bounds, and a superseded pair older than 90 s above a live park would otherwise end the walk the same way), while an unsuperseded `down` stays below the bounds as a request row.

### Symptom
A quiet restart request parked with the manager (a quiet converge's `main-converge` row, a `romp refresh --quiet`) read as nothing on record once a row above it had aged past the 90 s window or had been filed before this kernel started, when that row was one that never answers: another kernel's stop note (`POST /stop` naming one kernel notes the stop for that pid and leaves the manager's park armed) or a predecessor's `signal` verdict with the manager alive. A SIGTERM arriving then carried no request on its cut row, and `_parked_quiet_deploy` returned 0, so `_main_drift_check` said the parked deploy was no longer pending and converged at once, cutting the turns the quiet window was to spare. Inside the bounds the same rows kept the park, as the reader's docstring and `test_a_quiet_request_survives_a_stop_of_one_other_kernel` said they should.

### Cause
`_recent_restart_audit` applied the window break and the start-bound break to every row before the rules that pass over verdict rows and manager notes, so a row that could never be the answer ended the walk whenever it sat out of bounds, with the live park beneath it unread.

### Fix
The verdict and `manager-sigterm` classification now runs first, wherever the row sits; `via_manager` is set only for a note about this pid inside the window and at or after the start, the two bounds its old position applied implicitly. The window and the start bound apply only to rows that can answer (a request, a park, an unlabeled row). A design point that widens behaviour: `park_settled` is consulted for a predecessor's park only, as before, so a park this kernel filed is its own request past a settling row at any age, which inside the window it already was. Consulting it for in-lifetime parks would relabel the filing kernel's cut at the quiet window from the park's text to the manager's note, and `_last_deploy_restart_t` counts a quiet `main-converge` as a deploy landing only through that text; so it stays predecessor-only, and the docstring and docs/reference.md say so.

No code change in `_unrequested_signal_reason`: its stderr line has been guarded since #1286's review commit, and `_drain_and_exit` falls back to the plain verdict if the helper raises. That guard had no test that reached it (the existing test replaces the helper), which #1286's post-merge record lists as a follow-up. This commit adds the test and a docstring sentence naming what the guard protects: without it, a stderr that raises after the `signal` row has landed leaves that row saying the manager was stopped while the cut row gets the fallback's plain unrequested verdict, two records of one exit disagreeing.

### Tests
tests/test_restart_cuts.py, against the module's own state root, fake pids, `os._exit` patched. Red-first figures are from this PR's tests run against main's `kernel/kernel.py` (03feab4d).
- `RequestOnRecord::test_a_quiet_request_survives_a_stop_of_one_other_kernel` probes `(now, started)` in `(1080, 1003), (1101, 1003), (1080, 1011)` for both the other-pid stop note and the `signal` verdict. Before the change the four subtests for the two new pairs fail with `'' != 'kernel-asks-manager-restart-all: self-update'` (the walk breaks at the window, respectively at the start bound).
- `test_a_park_this_kernel_filed_is_its_own_past_a_settling_row_at_any_age` pins the design point above: the now=1101 subtest fails before the change; and the same rows for a kernel started after the park are a settled predecessor's park at either age.
- `test_a_quiet_request_delivered_or_dropped_before_this_kernel_is_consumed`, widened to the same pairs, is green before and after: a settling row out of bounds still settles a predecessor's park.
- `UnrequestedSignal::test_a_stderr_that_refuses_the_write_keeps_the_cut_row_agreeing_with_the_signal_row` fires the handler with a reaped manager pid under a stderr whose write raises EPIPE, then calls the helper directly under the same stream: the `signal` row says the manager was stopped, the one cut row carries the same verdict, and the direct call returns it. Green on main as it stands. With the guard removed (the write left bare) the cut row reads the plain unrequested verdict while the row says stopped, and the direct call raises: 1 failed. The cut row also carries a `drainError` from the drain's own writes under that stream; not asserted.

### Symptom
With a quiet restart request parked with the manager, a SIGTERM the manager did not send took the park as its request: a stray kill of this kernel, or of a dynamic (`/ensure`) or stateDir-less aux kernel sharing the state root and so the park. The cut row read `main-converge` with the park's t as `auditT`, no `signal` row was filed (the case that row exists for), and `_consumed_audit_t` then hid the park from every reader under the root: `_parked_quiet_deploy` returned 0, `_main_drift_check` logged the parked deploy as no longer pending, and `_last_deploy_restart_t` counted the stray cut as a deploy landing, anchoring the converge cool-down to a kill that deployed nothing, while the manager still held the park (the stray-exit path in `spawnKernel` clears no `pendingQuiet`).

Under default settings this restarted nothing by itself: the cool-down (`ROMP_CONVERGE_COOLDOWN`, 25 min) outlasts the park's 15-minute backstop, so the released stand-down posted no immediate restart-all. The cost was the misleading "no longer pending" line, a cool-down anchored to a stray kill (the next legitimate auto converge waited up to 25 min on a restart that was no deploy), and a park lost to every reader under the root. Only a `ROMP_CONVERGE_COOLDOWN` below the park's remaining time turned it into the immediate restart the quiet window was to prevent. In the single-kernel case the successor runs the checkout, so the stand-down never engaged and the cost was the attribution alone.

### Cause
`_graceful_term` handed whatever `_recent_restart_audit` returned to `_drain_and_exit`, which stamps `auditT` for any row. The reader returns a quiet park with or without a manager note, because the label and the drift stand-down need it, and nothing checked that this SIGTERM was the manager's delivery of the park. The manager notes every kill it sends before sending it (`auditSigterm`), so a SIGTERM with no note for this pid is not one. The reader's own `test_a_quiet_request_survives_a_stop_of_one_other_kernel` already described a `signal` verdict with `managerStopped: false` above a live park; no code path produced one, because the handler consumed the park instead.

### Fix
`_graceful_term` drops a `when: quiet` pick when `_manager_sigterm_row_for_us(reasons=("restart", "stop"))` finds no `manager-sigterm` note for this pid (or naming none) inside the window and this kernel's lifetime, and `_drain_and_exit` takes its existing unrequested path: `_unrequested_signal_reason` files the `signal` row (managerStopped decided as before, pid liveness plus the bounded look for a stop note) and the cut row carries `SIGNAL_REASON_UNREQUESTED` with no `auditT`, so the park stays on record for the kernel the quiet window will restart. `_manager_sigterm_row_for_us` gains the `reasons` parameter; its default `("stop",)` keeps the service-stop look's semantics for its existing caller, `_unrequested_signal_reason`, and for the tests that call it. The `_drain_and_exit(_audit_reason_text(rec), signum=signum` call shape is unchanged. Rejected shape: skipping only the `auditT` stamp would leave the stray kill's cut row labeled as the deploy, file no `signal` row, and let `_last_deploy_restart_t` count it as a deploy landing. Mixed-version note: a manager that writes no notes delivering a quiet restart is now filed as a stray signal and the park stays unconsumed until it expires; the successor runs the checkout, so the stand-down never engages, and the cost is one cut's attribution. Docstrings of `_graceful_term`, `_recent_restart_audit` and `_manager_sigterm_row_for_us`, and one sentence in docs/reference.md.

### Tests
tests/test_restart_cuts.py, `UnrequestedSignal`, with the note wait patched to 0.
- `test_a_stray_sigterm_with_a_quiet_park_on_record_does_not_spend_the_park` parks a `main-converge` quiet row (the shape `_run_main_update` writes), sets `ROMP_MANAGER_PID` to a live pid that wrote no note and fires the handler. Before the change: the audit file holds `['main-converge']` with no `signal` row (`['main-converge'] != ['main-converge', 'signal']`), the cut row reads `main-converge` with `auditT` set, and `_parked_quiet_deploy` returns 0. After: the `signal` row (managerStopped false) is filed, the cut row carries the unrequested reason and no `auditT`, `_recent_restart_audit` still returns the park for a sibling or the successor, and `_parked_quiet_deploy` returns the park's t.
- `test_the_managers_delivery_of_a_quiet_park_consumes_it`, the same park under the manager's `restart-all` note for this pid, is green before and after: the cut row names and consumes the park, no `signal` row, nothing parked afterwards.

### Symptom
tests/manager-exit-attribution.test.js's case "auditSigterm never throws: an unwritable root is a lost note, not a lost restart" was vacuous: with `auditSigterm`'s try/catch removed it still passed, so it pinned nothing about the property it names.

### Cause
Its root, `STATE/no/such/file.txt/dir`, did not exist yet, and Node's recursive `mkdirSync` creates every missing component, `file.txt` included; the row was written, nothing threw, and the catch never ran.

### Fix
The case writes an empty regular file under STATE and uses it as a component of the root's path (`a-file.txt/dir`): recursive `mkdirSync` fails with ENOTDIR, so the catch is what keeps the call from throwing. An `assert.throws` on the same `mkdirSync` keeps the fixture uncreatable (the case cannot go vacuous again without failing there first), the return is still a number, and no `restart-audit.jsonl` appears under the root. The comment says why the path is a file. `bin/romp-manager` is unchanged. The stray file sits directly under STATE; the row-counting cases read `STATE/restart-audit.jsonl`, not the directory listing, and stay green. The error code is not pinned (ENOTDIR here; another platform may report EEXIST).

### Tests
Red first with `auditSigterm`'s catch mutated to rethrow: the old case passed (17 of 17); the rewritten case fails with `ENOTDIR: not a directory, mkdir '.../a-file.txt/dir'` thrown out of `doesNotThrow` (16 of 17). With the real catch all 17 pass, and `node --test tests/manager-*.test.js` passes 96 of 96.

## Verification
- Targeted: tests/test_restart_cuts.py with test_main_drift_notice, test_restart_seam, test_restart_audit, test_kernel_headless_ops and test_kernel_tunnel_truth: 176 passed, 26 subtests passed.
- `node --test tests/manager-*.test.js`: 96 passed.
- This PR's tests against main's kernel.py (03feab4d), the tree it lands on: 6 failed (the four new pairs of the stop-of-one-other-kernel test, the in-lifetime park at now=1101, the stray-SIGTERM park test), 5 passed, 15 subtests passed; the stderr test, the delivered-or-dropped test and the manager's-delivery test pass there.
- Full pytest suite on the final tree (vscode-extension dependencies installed, so the served-page legs ran): 11230 passed, 41 skipped, 1740 subtests passed in 163 s, no failures.

Tier: fix

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1759,7 +1759,10 @@ read. A row with no action (the `romp refresh` row) is skipped, and the
 manager's `restart-all` note written after it is what names the refresh; a
 `romp refresh --quiet` row is the parked deploy that holds the automatic
 converge until the window opens, and the note written at the window names its
-delivery the same way.
+delivery the same way. A SIGTERM that reaches a kernel with a quiet-window
+request parked and no manager note for its pid (a note naming no pid counts as
+its own) is not that request's delivery: the kernel files a `signal` row and
+leaves the request on record for the kernel the window will restart.
 
 When no row qualifies, the kernel writes a row with action `signal`: the signal
 name, its pid and its parent's pid, the manager pid it was started with,
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1752,11 +1752,14 @@ request, and are passed over. The manager's `manager-sigterm` row is a note
 that the manager sent the signal, not a request: it answers only when no
 request row written before it lies within the window and this kernel's
 lifetime, with `manager-sigterm: <trigger>` as the reason, and a note aimed at
-another kernel's pid is ignored. A row with no action (the `romp refresh` row)
-is skipped, and the manager's `restart-all` note written after it is what
-names the refresh; a `romp refresh --quiet` row is the parked deploy that
-holds the automatic converge until the window opens, and the note written at
-the window names its delivery the same way.
+another kernel's pid is ignored. Verdicts and notes are passed over wherever
+they sit, an aged one included: one older than the window or older than this
+kernel never ends the walk, so a quiet-window request beneath it is still
+read. A row with no action (the `romp refresh` row) is skipped, and the
+manager's `restart-all` note written after it is what names the refresh; a
+`romp refresh --quiet` row is the parked deploy that holds the automatic
+converge until the window opens, and the note written at the window names its
+delivery the same way.
 
 When no row qualifies, the kernel writes a row with action `signal`: the signal
 name, its pid and its parent's pid, the manager pid it was started with,
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -21649,15 +21649,18 @@ def _audit_unrequested_signal(signum, pending=False, now=None, manager_stopped=F
 SIGNAL_MANAGER_NOTE_WAIT_S = 0.5   # how long an unexplained SIGTERM waits for the manager's own stop note
 
 
-def _manager_sigterm_row_for_us(now=None, window=90, started=None):
-    """Whether the audit tail holds a `manager-sigterm` STOP note aimed at THIS kernel (its `pid` is
-    ours, or it names none) within `window` seconds and not before this kernel's start (`started`; the
-    default is the process start, _STARTED, in whole seconds as the rows are): the manager's note that
-    it is stopping us, which is what _unrequested_signal_reason takes as the manager going down
-    alongside us. Only `reason: stop` counts. A `restart` note means the manager is alive and about to
-    respawn the kernel, and one landing during the drain of an unrelated signal (a stray kill, then the
-    rail's restart button) must not file that exit as a service stop with `managerStopped: true` while
-    the manager's own log shows a requested restart. The start bound is the one _recent_restart_audit
+def _manager_sigterm_row_for_us(now=None, window=90, started=None, reasons=("stop",)):
+    """Whether the audit tail holds a `manager-sigterm` note aimed at THIS kernel (its `pid` is ours, or
+    it names none) with a `reason` in `reasons`, within `window` seconds and not before this kernel's
+    start (`started`; the default is the process start, _STARTED, in whole seconds as the rows are).
+    With the default, only `reason: stop` counts: the manager's note that it is stopping us, which is
+    what _unrequested_signal_reason takes as the manager going down alongside us. A `restart` note means
+    the manager is alive and about to respawn the kernel, and one landing during the drain of an
+    unrelated signal (a stray kill, then the rail's restart button) must not file that exit as a service
+    stop with `managerStopped: true` while the manager's own log shows a requested restart. With
+    `reasons=("restart", "stop")` the question is whether the manager noted ANY kill of this kernel,
+    which is how _graceful_term tells the delivery of a parked quiet request from a signal nobody sent
+    through the manager. The start bound is the one _recent_restart_audit
     applies: the note this waits for lands AFTER our signal, so a stop note for our pid from before the
     process existed is a predecessor's, left behind when a reboot reused the pid (the machine back inside
     the window, the kernel up under the pid the previous one had, then a stray kill), and without the
@@ -21679,7 +21682,7 @@ def _manager_sigterm_row_for_us(now=None, window=90, started=None):
             continue
         if t0 - rec["t"] > window or rec["t"] < born:
             return False
-        if rec.get("pid") in (None, os.getpid()) and rec.get("reason") == "stop":
+        if rec.get("pid") in (None, os.getpid()) and rec.get("reason") in reasons:
             return True
     return False
 
@@ -21904,8 +21907,11 @@ def _recent_restart_audit(window=90, now=None, started=None):
         the deploy as the reason the service went down. The one exception is a `when: quiet` request: the
         manager parks it and restarts whatever kernel is running at the quiet window, so it outlives the
         kernel that filed it. With the manager's note for us on record the note answers (below); with none
-        (a manager build that wrote no notes) the parked row itself names the cut, unless a row above it
-        shows it delivered or dropped: a `manager-sigterm` note for a restart (every restart the manager
+        the parked row itself is returned, unless a row above it shows it delivered or dropped. That
+        no-note return serves the label and _parked_quiet_deploy, not consumption: _graceful_term
+        consumes a quiet row only when the manager noted a kill of this pid (its delivery), and files a
+        SIGTERM with a park on record and no such note as unrequested, the park untouched. Delivered or
+        dropped: a `manager-sigterm` note for a restart (every restart the manager
         sends clears the park in passing) or for the stop of this kernel or of every kernel (the manager's
         own shutdown takes the park down with it), a verdict that the manager was gone (`parent-gone`, or
         `signal` with `managerStopped`), or a cut row that already joined it (`auditT`: the restart it asked
@@ -53570,7 +53576,10 @@ def _graceful_term(signum, frame):
     mutation, and a cut turn keeps its 'working' state tail: the NEXT kernel's boot reconcile
     resumes exactly those. Bounded (~2s) so `romp refresh` stays snappy. Never construct the
     backend here: no SDK sessions were running if it doesn't exist. A second SIGTERM while the first
-    is draining returns at once (_EXIT_ONCE): the first finishes and exits."""
+    is draining returns at once (_EXIT_ONCE): the first finishes and exits. A parked quiet request on
+    record is consumed only when the manager noted a kill of this kernel (its `manager-sigterm` row for
+    this pid, or one naming none): the manager delivers the park, so a SIGTERM with no such note is not
+    its delivery, and the park stays on record for the kernel the quiet window will restart."""
     if not _EXIT_ONCE.acquire(blocking=False):
         return
     _TERMINATING[0] = True                          # the parent-watch stands down (see _parent_watch)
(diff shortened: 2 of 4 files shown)


ksarma:
Merging as shipped. The one finding below is a follow-up: it predates this PR and does not hold the merge.

Reviewed inline: by reading only, since the handler path signals processes and probes pids. CI ran the tests. The delivery check in `_graceful_term` matches the note's `reason` field by exact membership. The manager writes that field only as `restart` (restartKernel, restart-all included) or `stop` (stopKernel, the refresh self-bounce included). So every delivery the manager sends matches, and the trigger vocabulary never enters the check. The note names the child pid, and bin/romp-serve execs into the kernel, so that pid is the kernel's own. A note about the old kernel carries the old pid and cannot answer for the new one. The reordered walk classifies verdicts and manager notes before the window and the start bound. It sets `via_manager` only inside both and bounds request, park and unlabeled rows as before. So a stop note for this pid and a `signal` verdict with the manager stopped behave as they did. The predecessor-only settled rule is sound. `_last_deploy_restart_t` counts a quiet converge only through the cut row's reason text, so settling an in-lifetime park would lose the landing. The stderr-guard test probes a reaped child's pid with signal 0 and sends nothing. The node case's file-as-path-component fixture makes recursive mkdirSync fail, self-checked by assert.throws, so the catch runs. Docs match the code. The added lines carry no private strings, hostname or home path. The fix label fits.

Follow-up:
- `kernel/kernel.py:53595`. A single-kernel `POST /stop` note for this pid counts as the park's delivery. A shared-root sibling started before the park consumes a park the manager still holds (stopKernel clears no pendingQuiet). The main kernel's readers then see the park as spent. Pre-existing on main and narrowed here, not closed. Exclude a stop note whose trigger is `stop` from the delivery check and keep the self-bounce's `refresh` trigger. Then decide what the service-stop case files.
- Nit, `tests/test_restart_cuts.py:459`: the `stop` member of the delivery check (the self-bounce, reason stop, trigger refresh) has no handler-level case. A one-row variant of the delivery test would pin it.

Test evidence: nothing from the PR ran here. Local runs were py_compile and node --check in capped scopes. Red-first holds by reading. On main, the four new (now, started) subtests of the stop-of-one-other-kernel test return an empty reason at the window or start-bound break. So does the now=1101 subtest of the own-park test. The stray-SIGTERM test finds no `signal` row. CI at 3d9f37f6 is green on all 11 checks. Python 3.12 reports 11150 passed, 121 skipped, 1740 subtests. The Shell job's `node --test tests/manager-*.test.js` step passed 96 of 96, the rewritten case included.

Review run by the maintainer automated review session; verified against head 3d9f37f6.

ksarma:
Merging as shipped. The one finding below is a follow-up: it predates this PR and does not hold the merge.

Reviewed inline: by reading only, since the handler path signals processes and probes pids. CI ran the tests. The delivery check in `_graceful_term` matches the note's `reason` field by exact membership. The manager writes that field only as `restart` (restartKernel, restart-all included) or `stop` (stopKernel, the refresh self-bounce included). So every delivery the manager sends matches, and the trigger vocabulary never enters the check. The note names the child pid, and bin/romp-serve execs into the kernel, so that pid is the kernel's own. A note about the old kernel carries the old pid and cannot answer for the new one. The reordered walk classifies verdicts and manager notes before the window and the start bound. It sets `via_manager` only inside both and bounds request, park and unlabeled rows as before. So a stop note for this pid and a `signal` verdict with the manager stopped behave as they did. The predecessor-only settled rule is sound. `_last_deploy_restart_t` counts a quiet converge only through the cut row's reason text, so settling an in-lifetime park would lose the landing. The stderr-guard test probes a reaped child's pid with signal 0 and sends nothing. The node case's file-as-path-component fixture makes recursive mkdirSync fail, self-checked by assert.throws, so the catch runs. Docs match the code. The added lines carry no private strings, hostname or home path. The fix label fits.

Follow-up:
- `kernel/kernel.py:53595`. A single-kernel `POST /stop` note for this pid counts as the park's delivery. A shared-root sibling started before the park consumes a park the manager still holds (stopKernel clears no pendingQuiet). The main kernel's readers then see the park as spent. Pre-existing on main and narrowed here, not closed. Exclude a stop note whose trigger is `stop` from the delivery check and keep the self-bounce's `refresh` trigger. Then decide what the service-stop case files.
- Nit, `tests/test_restart_cuts.py:459`: the `stop` member of the delivery check (the self-bounce, reason stop, trigger refresh) has no handler-level case. A one-row variant of the delivery test would pin it.

Test evidence: nothing from the PR ran here. Local runs were py_compile and node --check in capped scopes. Red-first holds by reading. On main, the four new (now, started) subtests of the stop-of-one-other-kernel test return an empty reason at the window or start-bound break. So does the now=1101 subtest of the own-park test. The stray-SIGTERM test finds no `signal` row. CI at 3d9f37f6 is green on all 11 checks. Python 3.12 reports 11150 passed, 121 skipped, 1740 subtests. The Shell job's `node --test tests/manager-*.test.js` step passed 96 of 96, the rewritten case included.

Review run by the maintainer automated review session; verified against head 3d9f37f6.

PR #1303, opened 2026-09-10
Tests: the chat signature's _World fixture resets the mention-pins latch per world, so a torn-down world's state root is never recreated

ksarma:
Follow-up to the merge comment on #1296.

## Symptom
A run of `tests/test_chat_build_sig_inputs.py` could recreate `<temp dir>/state/mention-pins/assoc` under a test world's temp directory after that world was torn down, and leave it behind. Under pytest the conftest's removal of its temp root takes the recreated chain with it, and under `python -m unittest tests.test_chat_build_sig_inputs` the package's atexit hook removes it (every `TemporaryDirectory` goes through the `mkdtemp` the package records). The leak shows only in a direct script run or a unittest run from inside `tests/`, and those runs already leave the module's own import-time state root beside it. No verdict was affected: `_pin_assoc` reads an absent sidecar as `{}` under either path, every resolved token in the module is a `.md` that `_pin_mention` refuses before touching the pin directory, and no test enumerates the state directory.

## Cause
`_pin_dir()` latches `jd.STATE / "mention-pins"` in `km._MENTION_PINS` at first use and keeps it for the process. `_World.setUp` rebinds every judge-derived directory per world through `jd._rebind_state`, but the latch is kernel-owned and was not among the values the fixture saved and restored. So the first test that resolved a path token latched its own world's root, and that path outlived the world. Every later world's first resolve missed the sidecar memo (`tearDown` pops `_PIN_ASSOC_MEMO` for the sid) and reloaded through `_pin_assoc_dir()`, which mkdirs `<latched>/assoc` with parents, under the torn-down root. The sidecar test (`test_a_fresh_worlds_pin_sidecar_is_read_under_a_reused_id`) worked around this in-test with a save, a `None` and a `try/finally` restore, and the value it restored was always a torn-down world's path.

## Fix
`tests/test_chat_build_sig_inputs.py` only.
- `_World.setUp` saves `km._MENTION_PINS` beside the other saved values and sets it to `None` after `jd._rebind_state(state)`, so each world's first `_pin_dir()` latches its own root. `tearDown` restores it beside the other restores, before the temp dir is removed.
- The sidecar test drops its in-test latch lines (the save, the `None`, the `try/finally`). Its verdict is unchanged: `{'notes/report.md': 'pin1'}`.
- `RecordedDependencies.setUp` and `tearDown` mirror the same save, reset and restore, so the two fixtures stay in step. Its two-world test otherwise left the latch on its removed second world; nothing recreates under it today because no later test in that class resolves a token, so this mirror has no test of its own.

## Tests
- New: `Differential.test_each_world_latches_its_own_pin_directory`, beside the sidecar test. It resolves `notes/report.md` in one world, records that world's directory and tears the world down. It then asserts that `km._MENTION_PINS` no longer points at that directory (the restore), sets the latch back to that directory's `state/mention-pins` (the value a fixture that never reset would inherit; an `addCleanup` puts the preceding value back after the final `tearDown`), sets up a second world and resolves again, and asserts that `km._MENTION_PINS` is the second world's `jd.STATE / "mention-pins"` (the reset) and that the first directory no longer exists. It asserts on the global rather than on `_pin_dir()`, so a resolve that skipped the reload cannot latch at the assertion and pass.
  - Before the change, alone: the first assertion fails, `PosixPath('<first>/state/mention-pins') == PosixPath('<first>/state/mention-pins')`, since nothing restored the latch. With that assertion removed for the record, the equality fails (`PosixPath('<first>/state/mention-pins') != PosixPath('<second>/state/mention-pins')`) and `first.exists()` is `True`.
  - Before the change, in the whole module: pytest runs unittest methods in sorted name order, so the sidecar test runs first and its `finally` restores its own torn-down first-world path, a different one; the first assertion passes, the unreset fixture inherits the stale path set on the latch, and the equality fails on it; 1 failed, 49 passed.
  - After the change: 50 passed, serial and under `-n 4`.
  - Deleting only the fixture's reset (`km._MENTION_PINS = None` in `_World.setUp`) fails the test at the equality; deleting only its restore (`km._MENTION_PINS = pins` in `_World.tearDown`) fails it at the first assertion. Both alone and in the whole module.
- The sidecar test keeps its verdict with its latch lines gone.
- Full pytest suite on this head: 11227 passed, 41 skipped, 1726 subtests passed (`-n 6`).

Tier: docs

### DIFF
--- a/tests/test_chat_build_sig_inputs.py
+++ b/tests/test_chat_build_sig_inputs.py
@@ -415,8 +415,9 @@ class _World(unittest.TestCase):
         state.mkdir()
         self.saved = (jd.STATE, jd.PROJECTS, km.NAMES, km.WORKING_DIR, km._GLOBAL_CLAUDE_MD, km._tmux_sessions, km._sdk,
                       os.environ.get("CLAUDE_CONFIG_DIR"), os.environ.get("ROMP_HOST_NAME"),
-                      len(km._downtime), km._claude_account_label, km._auth_avail_status)
+                      len(km._downtime), km._claude_account_label, km._auth_avail_status, km._MENTION_PINS)
         jd._rebind_state(state)                       # every STATE-derived dir (goals, states, episodes, gone, sdk, ...)
+        km._MENTION_PINS = None                       # the pin dir latches at first use (_pin_dir): this world's root, not a prior one's
         jd.PROJECTS = proj
         jd.NAMES.mkdir()
         (jd.NAMES / SID).write_text("web\t%s\t#1EA1EB\twhite\n" % self.cdir)
@@ -435,11 +436,12 @@ class _World(unittest.TestCase):
         self.sess = {"sid": SID, "name": "web", "path": str(self.tpath), "anchor": SID}
 
     def tearDown(self):
-        (state, proj, names, wdir, gmd, tmux, sdk, cfg, host, ndown, acct, avail) = self.saved
+        (state, proj, names, wdir, gmd, tmux, sdk, cfg, host, ndown, acct, avail, pins) = self.saved
         jd._rebind_state(state)
         jd.PROJECTS = proj
         km.NAMES, km.WORKING_DIR, km._GLOBAL_CLAUDE_MD, km._tmux_sessions, km._sdk = names, wdir, gmd, tmux, sdk
         km._claude_account_label, km._auth_avail_status = acct, avail
+        km._MENTION_PINS = pins
         for k, v in (("CLAUDE_CONFIG_DIR", cfg), ("ROMP_HOST_NAME", host)):
             if v is None:
                 os.environ.pop(k, None)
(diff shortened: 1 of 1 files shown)


ksarma:

Reviewed inline: the pin directory latch and the two fixtures that now own it. `_pin_dir` latches `jd.STATE / "mention-pins"` on first use, and `_pin_assoc_dir` mkdirs `assoc` under it. `_World.setUp` rebinds the state root and sets the latch to None, so every mkdir in the module lands under that world's temp directory. The module also points XDG_STATE_HOME at a mkdtemp before loading the kernel, so nothing reaches the live state root. The tearDown restore runs before the temp directory is removed. `RecordedDependencies` carries the same save (11 values, matching unpack), reset after its rebind, and restore. The new test's `addCleanup` is right: unittest runs cleanups after tearDown and binds the setattr argument at registration. The final tearDown therefore restores the planted stale value, and the cleanup then puts the pre-test value back. Asserting on the global rather than `_pin_dir()` is correct: a resolve that skipped the reload would otherwise latch at the assertion. The no-verdict-affected claim holds. `_pin_mention` returns None for a non-image extension before touching the pin directory. `_pin_assoc` reads an absent sidecar as {}, and the module resolves only `.md` tokens. The docs label fits: one test file, fixtures and one test. Privacy scan of the 42 added lines: no hits.

Tests: at head the module (50) and its two neighbours (85) pass under a scratch state root, with nothing left under the temp root. Fixture variants in scratch copies reproduced the failure shapes the body gives:
- the base fixture plus the new test fails at the first assertion alone and at the equality in the whole module;
- removing only the reset fails at the equality;
- removing only the restore fails at the first assertion;
- a probe showed the module latch unchanged across the new test.

CI: all eleven checks green at head.

Review run by the maintainer automated review session; verified against head ea472040.

ksarma:

Reviewed inline: the pin directory latch and the two fixtures that now own it. `_pin_dir` latches `jd.STATE / "mention-pins"` on first use, and `_pin_assoc_dir` mkdirs `assoc` under it. `_World.setUp` rebinds the state root and sets the latch to None, so every mkdir in the module lands under that world's temp directory. The module also points XDG_STATE_HOME at a mkdtemp before loading the kernel, so nothing reaches the live state root. The tearDown restore runs before the temp directory is removed. `RecordedDependencies` carries the same save (11 values, matching unpack), reset after its rebind, and restore. The new test's `addCleanup` is right: unittest runs cleanups after tearDown and binds the setattr argument at registration. The final tearDown therefore restores the planted stale value, and the cleanup then puts the pre-test value back. Asserting on the global rather than `_pin_dir()` is correct: a resolve that skipped the reload would otherwise latch at the assertion. The no-verdict-affected claim holds. `_pin_mention` returns None for a non-image extension before touching the pin directory. `_pin_assoc` reads an absent sidecar as {}, and the module resolves only `.md` tokens. The docs label fits: one test file, fixtures and one test. Privacy scan of the 42 added lines: no hits.

Tests: at head the module (50) and its two neighbours (85) pass under a scratch state root, with nothing left under the temp root. Fixture variants in scratch copies reproduced the failure shapes the body gives:
- the base fixture plus the new test fails at the first assertion alone and at the equality in the whole module;
- removing only the reset fails at the equality;
- removing only the restore fails at the first assertion;
- a probe showed the module latch unchanged across the new test.

CI: all eleven checks green at head.

Review run by the maintainer automated review session; verified against head ea472040.

PR #1304, opened 2026-09-10
Kernel exit path: a dead stderr no longer skips the drain or the exit, and the cut row names the reason helper's fault (reasonError)

ksarma:
Follows the merged #1302 and lands on main (rebased onto 5c8ccebe on 2026-09-10, three commits replayed unchanged); the change to review is the three commits. All three items are the follow-ups recorded on #1286 after its merge, against its review-fix commit f36ddba8.

## Symptom

Three gaps on the kernel's exit path (`_drain_and_exit`, shared by `_graceful_term` and `_parent_watch`):

1. When `_unrequested_signal_reason` raised before filing its `signal` row (a `ROMP_MANAGER_PID` the kernel cannot use as a pid: a digit string `int()` rejects, or a pid past the C int range, which `_pid_alive` does not catch), the inner guard fell back to the plain verdict and recorded nothing else: no stderr line, no field. The cut row was indistinguishable from a healthy unrequested exit, and the `signal` row docs/reference.md promises was missing with nothing on disk to say why. The same block already records a drain fault as `drainError`.
2. The helper's own stderr line sits after the `signal` row lands and before the return, so a stderr that raises there would carry the verdict away (the plain wording on the cut row beside `managerStopped: true` in the audit row). The write is guarded, and the existing broken-pipe test drives an OSError through `_graceful_term` and then calls the helper alone asserting the return only. Two things were unpinned: the guard's breadth beyond OSError, and that the audit row filed on the direct call agrees with the verdict returned.
3. The kernel writes on the stdio the manager spawned it with (stdio inherit). A stream that raises at write time (a reset journal stream, a closed tty on a kernel that outlived the hangup, a full log disk) hit the drain announcement, the first statement of the drain's `try`, so the `except` ran before `be.drain`: no session was drained, their claude children orphaned with in-flight turns uninterrupted, and the cut row read as a clean drain (`cutTurns` empty) with a `drainError` naming the stderr fault, not a drain's. The "drain failed" line raised again inside the except. And when the row builder raised, the outer except's own line stood before `os._exit(0)` unguarded: with stderr dead it raised out of the handler and the exit never ran. From `_graceful_term` that ends the interpreter with no row; from `_parent_watch` it ends that thread alone and the kernel lives on with the exit lock held, so every later SIGTERM returns at once.

## Cause

The inner `except Exception:` only reset the reason. Three `sys.stderr.write` calls on the exit path were unguarded, one of them inside the `try` whose body is the drain; only the reason helper's own line had a guard.

## Fix

- Commit 1: the except keeps the plain verdict, keeps the traceback's last line (200 characters, the recipe `drainError` uses two lines below) and writes it to the row as `reasonError` beside the `drainError` line, and says so on stderr best-effort. The pinned `reason = _unrequested_signal_reason(signum, be)` line is unchanged. One sentence in docs/reference.md's restart-files passage describes the field.
- Commit 2: tests only (below).
- Commit 3: `_exit_log`, a best-effort stderr line defined above `_graceful_term` (outside the pinned `_graceful_term`..`main()` block), carries every line on the exit path. The announcement now precedes the `try`, whose body is the drain alone, so `drainError` names a fault of the drain only. The row-failed line goes through it, so `os._exit(0)` is reached whatever stderr does, and the line that names the reason helper's fault does too, so its raise cannot leave the inner except for the outer one, which writes no row. The drain-failed line follows for hygiene: the finally reaches `os._exit(0)` either way. With a healthy stderr the output is unchanged. The `_drain_and_exit` docstring states the property.

## Tests

All in tests/test_restart_cuts.py. Every case runs the real code in-process under the module's own state root with `sys.stderr` swapped for the call and restored, and only reaped pids as `ROMP_MANAGER_PID`; the handler-level cases run `_graceful_term` or `_parent_watch` with `os._exit` patched to raise SystemExit, the direct-call case swaps stderr only.

- `UnrequestedSignal::test_a_raising_reason_helper_still_leaves_the_cut_row` (extended): asserts `reasonError == "OverflowError: pid"`, no `drainError`, no `signal` row on this path, and that the stderr line names the field. Before commit 1: `KeyError: 'reasonError'`.
- `UnrequestedSignal::test_a_closed_stderr_does_not_undo_the_service_stop_verdict` (new, commit 2): a closed `io.StringIO` as stderr (its write raises ValueError, the kind a reset stream leaves behind), a reaped pid, the helper called with `wait=0`; asserts `SIGNAL_REASON_MANAGER_STOPPED`, one audit row with `managerStopped` true, and the row's reason equal to the return. Red by mutation: with the helper's guard removed, `ValueError: I/O operation on closed file` out of the call (the broken-pipe test beside it goes red through the handler as well; every other case stays green). No kernel change in this commit.
- `UnrequestedSignal::test_a_raising_stderr_does_not_skip_the_drain` (new): a `DeadTty` stderr (OSError EIO) and a `LiveDrain` backend through `_graceful_term`; asserts the drain ran, `cutTurns` names the turn, and no `drainError`. Before commit 3: `False is not true: the drain runs whatever stderr does` (cutTurns empty, drainError naming the OSError).
- `UnrequestedSignal::test_a_raising_stderr_when_the_cut_row_fails_still_exits` (new): `_restart_cut_row` raising RuntimeError under the dead stderr; asserts `os._exit(0)` was reached once and no row. Before: `OSError: [Errno 5] Input/output error` raised out of the handler in place of the exit.
- `ParentGone::test_a_raising_stderr_does_not_skip_the_drain` (new): the same through `_parent_watch` with a reaped manager pid; asserts the drain ran and both rows land. Before: the drain was skipped.
- `UnrequestedSignal::test_a_raising_reason_helper_under_a_dead_stderr_still_leaves_the_cut_row` (new): the helper raising OverflowError under a closed `io.StringIO` stderr; asserts one cut row with the plain verdict and `reasonError == "OverflowError: pid"`, no `signal` row. Red by mutation: with that one line a bare `sys.stderr.write`, no row lands (`0 != 1`); with `_exit_log` narrowed to `except OSError`, ValueError out of the handler.
- `UnrequestedSignal::test_the_handler_files_the_row_and_the_cut_reason_when_nothing_asked` (extended): a clean exit's row carries neither `drainError` nor `reasonError`. Red with the field set unconditionally.

Targeted run on the final tree (tests/test_restart_cuts.py, test_main_drift_notice, test_restart_seam, test_restart_audit, test_kernel_headless_ops): 148 passed, 26 subtests. The source pins in CutRow, test_restart_seam, test_kernel_headless_ops and test_main_drift_notice hold unchanged. Full pytest (`-n 6`, extension node modules installed so the served-page legs run): 11235 passed, 41 skipped, 1740 subtests passed.

Added after the review (2026-09-10, commit 3d4188b3, the fourth commit): the SDK backend's `log=` wire in `_sdk_locked` was a bare `sys.stderr.write` lambda, and `SdkBackend.drain` logs its summary line after every session is shut down, joined and reaped, immediately before it returns, so under a dead stderr that one line raised out of `be.drain` with the work already done and `_drain_and_exit` filed an empty `cutTurns` row with a `drainError` naming the stderr fault, the misfiled row the guards in commit 3 exist to close. The wire now goes through a `_backend_log` callback that writes the `sdk-backend:` line through `_exit_log`, so the guard covers the backend's whole life, not the exit alone. Tests: the `LiveDrain` fake logs its summary through the wired callback, `UnrequestedSignal.test_a_raising_stderr_does_not_skip_the_drain` therefore exercises the summary line under the raising stderr, and `CutRow.test_the_backends_log_wire_is_best_effort` pins the wire (`log=_backend_log`) and that `_backend_log` swallows a raising stderr. Not changed: the codex backend's `log=` wire keeps the bare lambda; `CodexBackend` has no drain and `_drain_and_exit` drains the SDK backend alone, so it is off the exit path.

Tier: fix

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1780,7 +1780,11 @@ receive it. A kernel whose manager disappears writes a row with action
 `parent-gone` before it exits. `restart-cuts.jsonl` gets one row per exit
 naming the turns the drain cut and the reason: the audit row's `action:
 reason`, the `signal` row's reason, or `parent-gone: the manager exited; the
-kernel followed it`. A second SIGTERM during the drain is ignored; the first
+kernel followed it`. When the helper that files the `signal` row fails (a
+`ROMP_MANAGER_PID` the kernel cannot use as a pid), no `signal` row is written,
+and the cut row carries the plain `signal, not requested through the manager`
+verdict plus a `reasonError` naming the fault, so the missing row is explained
+on disk. A second SIGTERM during the drain is ignored; the first
 writes the row. The manager's log says `exited without a restart request
 (signal or crash); respawning` when a kernel exits that it did not ask to stop
 or restart.
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -53614,6 +53614,7 @@ def _drain_and_exit(reason, signum=None, what="SIGTERM", audit=None):
     (_unrequested_signal_reason: an empty reason used to be all restart-cuts.jsonl had for such a restart)."""
     res = {}
     err = ""
+    reason_err = ""
     be = _sdk_backend or None
     try:
         sys.stderr.write("romp-kernel: %s, draining SDK sessions\n" % what)
@@ -53630,17 +53631,28 @@ def _drain_and_exit(reason, signum=None, what="SIGTERM", audit=None):
             if not reason and signum is not None:
                 # the helper's own guard: it reads ROMP_MANAGER_PID and writes to stderr, and a raise
                 # there (a pid too large for os.kill, a stderr the dying supervisor closed) must not
-                # skip the cut row this exit exists to leave; the reason falls back to the plain verdict
+                # skip the cut row this exit exists to leave. The reason falls back to the plain verdict
+                # and the row names the fault (reasonError, drainError's recipe): the helper's raise
+                # paths precede its `signal` row, so this exit has none, and a row with the plain verdict
+                # alone read as a healthy exit of that kind with the row it promises missing
                 try:
                     reason = _unrequested_signal_reason(signum, be)
                 except Exception:
                     reason = SIGNAL_REASON_UNREQUESTED
+                    reason_err = traceback.format_exc().strip().splitlines()[-1][:200]
+                    try:                              # best-effort: a closed stderr is one of the causes
+                        sys.stderr.write("romp-kernel: the signal's reason helper failed, the cut row "
+                                         "carries the plain verdict and reasonError: %s\n" % reason_err)
+                    except Exception:
+                        pass
             row = _restart_cut_row(res, watches_armed=len(_pr_watches) + len(_watches),
                                    audit_reason=reason)
             if audit:
                 row["auditT"] = int(audit["t"])     # the audit row this cut CONSUMED (see _recent_restart_audit)
             if err:
                 row["drainError"] = err.strip().splitlines()[-1][:200]
+            if reason_err:
+                row["reasonError"] = reason_err     # the reason helper's fault: no `signal` row was filed
             _append_restart_cut(row)
         except Exception:
             sys.stderr.write("romp-kernel: cut ledger failed: %s\n" % traceback.format_exc())
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -53546,6 +53546,17 @@ _TERMINATING = [False]   # set the moment an exit path takes the lock: the watch
 #                          know an exit is underway ask this instead of touching the lock
 
 
+def _exit_log(text):
+    """One stderr line on the exit path, best-effort. The kernel writes on the stdio bin/romp-manager
+    spawned it with (stdio inherit), so a supervisor closing its own stderr does not close this
+    descriptor, but a reset journal stream, a closed tty or a full log disk makes the write raise, and
+    nothing said here is worth skipping the drain, filing the fault as the drain's, or skipping the exit."""
+    try:
+        sys.stderr.write(text)
+    except Exception:
+        pass
+
+
 def _parent_watch():
     """Exit if the manager that spawned us (ROMP_MANAGER_PID) dies, so a supervisor crash doesn't
     orphan the kernel. No-op when launched standalone (no ROMP_MANAGER_PID). The exit leaves the same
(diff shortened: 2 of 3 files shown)


ksarma:
Post-merge record: the owner merged this PR as 025b032a at head 3d4188b3 while this review was running; nothing to fold. The bar was met, and nothing below would have held it.

Reviewed inline: by reading only, since the exit path signals processes and probes pids; CI ran the tests. The four commits close the three items recorded after #1286. Commit 1 (8011742e) puts the reason helper's fault on the cut row as `reasonError`, the traceback's last line cut to 200 characters. That is the recipe `drainError` uses two lines below. The docs sentence matches the code: the helper's raise paths precede its audit write, so no signal row exists on that path. Commit 2 (3561e4c7) pins the helper's existing stderr guard with a closed StringIO, a ValueError rather than an OSError. Commit 3 (dfd19f81) adds `_exit_log`, one write under try/except Exception. It also moves the announcement ahead of the try, so the try's body is the drain alone and `drainError` names a drain fault only. Commit 4 (3d4188b3) covers the write those guards missed. `SdkBackend.drain` logs its summary after its work through the log callback. `_sdk_locked` wired that callback as a bare stderr write, and `SdkBackend._log` calls it unguarded, so a dead stderr raised out of a finished drain. The wire is now `_backend_log`, pinned by source text and by direct calls under a raising and a healthy stderr, whose output is unchanged. At the head, every Exception inside the finally lands in an outer except whose only statement is a guarded write. `_append_restart_cut` never raises, and `os._exit(0)` follows on every path. The two helpers sit above `_parent_watch`, outside the pinned `_graceful_term..main()` block. The source pins in CutRow, test_restart_seam, test_kernel_headless_ops and test_main_drift_notice hold on the head text. The ledger's readers index reason, t, cutTurns and auditT and ignore the new key. The tests use reaped pids only, strip ROMP_MANAGER_PID, and swap stderr and `os._exit` inside context managers. The diff against 5c8ccebe is exactly the three files, and the added lines carry no private strings, hostname or home path.

Follow-ups:
- PR body and commit 3d4188b3: the body describes three commits, and the fourth's message is a headline. The merged history therefore records the backend wire change and its test only in that commit's diff. The body is still editable.
- `kernel/kernel.py:14810`: the codex backend keeps the bare stderr lambda. It has no drain, and `_drain_and_exit` drains the SDK backend alone, so it is off the exit path. A guard of the same shape would be a hygiene change.

Test evidence: locally, only py_compile of `kernel/kernel.py` and `tests/test_restart_cuts.py` ran, in a capped scope, by rule. Red-first holds by reading at 5c8ccebe. `test_a_raising_reason_helper_still_leaves_the_cut_row` and its dead-stderr sibling fail on KeyError `reasonError`. Both `test_a_raising_stderr_does_not_skip_the_drain` cases fail on `be.called`, since the announcement inside the try raised before `be.drain`. `test_a_raising_stderr_when_the_cut_row_fails_still_exits` sees an OSError in place of SystemExit. `test_the_backends_log_wire_is_best_effort` fails on the missing `log=_backend_log` wire. `test_a_closed_stderr_does_not_undo_the_service_stop_verdict` is a mutation pin of an existing guard, as the body says. CI at 3d4188b3 is green on all 9 checks, Python 3.10 to 3.13 included.

Review run by the maintainer automated review session; verified against head 3d4188b3.

PR #1307, opened 2026-09-10
A stop of one kernel no longer consumes a parked quiet converge: the manager's stop note with trigger stop names the cut and the park stays on record

ksarma:
## Symptom

The manager's stop of one kernel consumed a parked quiet converge it does not deliver. The park is the `when: quiet` row `_run_main_update` writes; the manager holds it and restarts whatever kernel runs at the quiet window. When the manager stopped one kernel (`POST /stop` naming it, `stopKernel` in bin/romp-manager), that kernel's `_graceful_term` took the park as the request behind its SIGTERM: the cut row read `main-converge` with `auditT` = the park's t. `_consumed_audit_t` then answered the park's t for every reader under the root, so the park read as spent while the manager still held it: `_parked_quiet_deploy` returned 0, `_main_drift_check` dropped its stand-down and logged the parked deploy as no longer pending, and `_last_deploy_restart_t` counted the stop as a deploy landing. Which row was filed depended on birth order: a sibling kernel born after the park (a dynamic kernel or a stateDir-less aux shares the main root) got `manager-sigterm: stop` from the reader's walk, with the park untouched. A stop of every kernel (`shutdownAll`) writes the same note and consumed the park the same way.

## Cause

`_graceful_term` kept a `when: quiet` pick whenever `_manager_sigterm_row_for_us(reasons=("restart", "stop"))` was truthy, and that helper matched a note on pid and reason alone; the note's `trigger` never entered it. `stopKernel` writes reason `stop` with trigger `stop` and does not clear the manager's park (`clearPendingQuiet` is called by the quiet apply, `shutdownAll`, `/restart-all` and `/restart`). The notes that follow a clear are the restart notes (trigger `restart` or `restart-all`) and the stale-manager self-bounce's stop note with trigger `refresh`. A stop note with trigger `stop` follows no clear, and it cannot say whether one kernel or every kernel is stopping.

## Fix

Two commits.

Commit 1 (kernel/kernel.py, docs/reference.md, tests). `_manager_sigterm_row_for_us` returns the matching row or None instead of True or False; its two callers and the six tests that call it check truthiness only. In `_graceful_term`'s quiet pick, a note whose reason is `stop` and whose trigger is `stop` (the trigger defaults to the reason, as the manager's writer does with `trigger || reason`) becomes the cut's request: the row reads `manager-sigterm: stop` with `auditT` = the note's t, no `signal` row is filed, and the park stays on record. That is the row a sibling born after the park already got, so both birth orders now file the same one. A restart note and the stop/refresh note deliver and consume the park as before, and a quiet pick with no note takes the unrequested path as before. The walk in `_recent_restart_audit` is unchanged. The docstrings of `_graceful_term`, `_manager_sigterm_row_for_us` and `_recent_restart_audit` say so; the restart-audit paragraph in docs/reference.md gains the sentence.

Commit 2 (tests only). The self-bounce's stop/refresh note is pinned as a delivery at the handler; see Tests.

Design points:

- A service stop during a parked quiet converge now files `manager-sigterm: stop` instead of the park's text and leaves the park on disk. The successor reads it as a live predecessor's park (the stop note names the predecessor's pid, and a stop/stop note for another pid settles nothing in the walk). That is inert: `_main_drift_check` consults `_parked_quiet_deploy` only while the kernel does not run the checkout, and the successor does; the park expires at `RESTART_EXPECT_MAX_S`. A restart note for the successor inside that window settles the park in the walk, so the successor's cut row reads `manager-sigterm: restart` and `_parked_quiet_deploy` answers 0 afterwards; a stray kill of the successor files an unrequested `signal` row with no `auditT`, as before, and the park stays visible to `_parked_quiet_deploy`. Both are tested (the two successor cases under Tests).
- The one cost: `_last_deploy_restart_t` no longer anchors the converge cool-down to a service restart that landed a parked quiet converge. The cut row's text was its only source for a quiet main-converge; its audit-row source counts `when: now` only.
- Rejected, `rec = None` for the stop/stop note: the unrequested path's default look (`reasons=("stop",)`) finds the same note and files `managerStopped: true`, a false verdict on a single-kernel stop with the manager alive. And a main kernel respawned after a crash, whose park is a predecessor's that the manager still holds (`spawnKernel`'s exit handler clears nothing), would read its park as settled by that verdict, so the park is lost in that sub-case.
- Rejected, a manager-side trigger split so that `shutdownAll` writes a trigger of its own: the principled disambiguation, but a bin/romp-manager change with a mixed-version window, and the kernel must still read `stop` ambiguously from an older manager. A candidate follow-up.
- Rejected, a post-drain re-check of the manager pid: timing inference over the manager's 800 ms exit.
- Same-second t: `_consumed_audit_t` keys on t alone, so a stop note landing in the same second as the park marks the park spent too. Pre-existing on the sibling's path; every test here puts the park and the note in distinct seconds so the `auditT` stamp says which row was consumed.

## Tests

All in `tests/test_restart_cuts.py`, class `UnrequestedSignal`, through `_graceful_term` with `os._exit` patched to raise, against a rebound state root with the test process as the manager pid; a test whose park predates its note pins the process start (`_STARTED`) before the park so the park sits inside the kernel's lifetime.

Commit 1:

- `test_a_stop_of_this_kernel_alone_does_not_spend_a_quiet_park_on_record` (the fix's test): a park at t-5, a stop/stop note for this pid at t, the manager alive. Asserts no `signal` row; one cut row reading `manager-sigterm: stop` with `auditT` = t; `_consumed_audit_t` not the park's t; `_recent_restart_audit` still returning the park; `_parked_quiet_deploy` still t-5. Before the change: `'main-converge' != 'manager-sigterm: stop'`; the cut row carried `auditT` = t-5, `_consumed_audit_t` answered t-5, `_parked_quiet_deploy` returned 0 and the reader's walk returned the note.
- `test_a_stop_note_without_a_trigger_reads_as_the_stop_of_this_kernel_alone`: the same rows with no `trigger` key on the note (the shape the reader already reads by its reason); the same assertions. Before the change: the same mismatch. With the handler's default to reason dropped (`note.get("trigger") == "stop"` alone), this is the one test in the module that turns red (1 failed, 72 passed).
- `test_a_sibling_born_after_the_park_files_the_same_row_for_the_stop_of_itself`: the same rows read by a kernel whose start sits between them. Asserts the cut reads `manager-sigterm: stop` with `auditT` = t, and `_parked_quiet_deploy` still t-5 for the kernel that filed the park. Green before and after: the row the change makes both birth orders file.
- `test_the_managers_restart_of_this_kernel_alone_delivers_a_quiet_park_too`: the same park under a restart/restart note (`POST /restart` naming one kernel). Asserts the cut reads `main-converge` with `auditT` = the park's t and `_parked_quiet_deploy` 0. Green before and after: the member the change keeps.
- `test_a_restart_of_the_successor_settles_the_park_a_service_stop_left_on_record` and `test_a_stray_kill_of_the_successor_leaves_the_park_a_service_stop_left_on_record`: the record a service stop now leaves (the park at t-20, the stop/stop note at t-15 for a pid nothing owns, that kernel's cut row naming the note with `auditT` = t-15), read by a successor started at t-10 under a live manager. The first asserts `_parked_quiet_deploy` answers t-20, then, under a restart/restart note for the successor's pid at t: no `signal` row, a second cut row reading `manager-sigterm: restart` with `auditT` = t, and `_parked_quiet_deploy` 0. The second, with no note for the successor: a `signal` row with `managerStopped` false, a second cut row with the unrequested reason and no `auditT`, and `_parked_quiet_deploy` still t-20. Both green before and after: the first design point's two claims.

Commit 2:

- `test_the_stale_manager_self_bounce_delivers_a_quiet_park_too`: a variant of the existing delivery test with the self-bounce's stop/refresh note, the park at t-5 and the note at t. Asserts no `signal` row, the cut reading `main-converge` with `auditT` = t-5, `_parked_quiet_deploy` 0. Green as it stands. Red by mutation with the predicate narrowed to `reasons=("restart",)`: `['main-converge', 'manager-sigterm', 'signal'] != ['main-converge', 'manager-sigterm']`. On the code before commit 1 under that mutation, with commit 1's two stop/stop tests deselected (they fail there on the defect itself), this pin is the only test in the module that turns red (1 failed, 71 passed); on this tree the two stop/stop tests turn red with it (3 failed, 71 passed). With the handler's trigger check reduced to the reason alone (every stop note taken as the cut), the pin fails at the cut-row reason: `'manager-sigterm: refresh' != 'main-converge'`.

The existing `test_a_stray_sigterm_with_a_quiet_park_on_record_does_not_spend_the_park` (no note) and `test_the_managers_delivery_of_a_quiet_park_consumes_it` (a restart-all note) stay green, so each of the four note shapes the manager writes has a handler-level case.

Targeted run at the final tree (`tests/test_restart_cuts.py`, `test_main_drift_notice.py`, `test_restart_seam.py`, `test_restart_audit.py`, `test_kernel_headless_ops.py`): 156 passed, 26 subtests passed. Full suite at the final tree (`pytest -q -n 6 tests/`, with the extension's node_modules installed): 11253 passed, 41 skipped, 1744 subtests passed in 171.96s, no failures.

Tier: fix

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1762,7 +1762,13 @@ converge until the window opens, and the note written at the window names its
 delivery the same way. A SIGTERM that reaches a kernel with a quiet-window
 request parked and no manager note for its pid (a note naming no pid counts as
 its own) is not that request's delivery: the kernel files a `signal` row and
-leaves the request on record for the kernel the window will restart.
+leaves the request on record for the kernel the window will restart. The
+manager's stop of one kernel (a `stop` note with trigger `stop`, which leaves
+the manager's parked request armed; a stop of every kernel writes the same
+note) is not the delivery either: that cut is named by the note,
+`manager-sigterm: stop`, and the request stays on record. A restart note, or
+the self-bounce's `refresh` note, is the delivery: the cut row names the
+request and consumes it.
 
 When no row qualifies, the kernel writes a row with action `signal`: the signal
 name, its pid and its parent's pid, the manager pid it was started with,
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -21657,9 +21657,10 @@ SIGNAL_MANAGER_NOTE_WAIT_S = 0.5   # how long an unexplained SIGTERM waits for t
 
 
 def _manager_sigterm_row_for_us(now=None, window=90, started=None, reasons=("stop",)):
-    """Whether the audit tail holds a `manager-sigterm` note aimed at THIS kernel (its `pid` is ours, or
+    """The newest `manager-sigterm` note in the audit tail aimed at THIS kernel (its `pid` is ours, or
     it names none) with a `reason` in `reasons`, within `window` seconds and not before this kernel's
-    start (`started`; the default is the process start, _STARTED, in whole seconds as the rows are).
+    start (`started`; the default is the process start, _STARTED, in whole seconds as the rows are), as
+    the row itself, or None when there is none; a caller reads the note's `trigger` off the row.
     With the default, only `reason: stop` counts: the manager's note that it is stopping us, which is
     what _unrequested_signal_reason takes as the manager going down alongside us. A `restart` note means
     the manager is alive and about to respawn the kernel, and one landing during the drain of an
@@ -21667,17 +21668,20 @@ def _manager_sigterm_row_for_us(now=None, window=90, started=None, reasons=("sto
     stop with `managerStopped: true` while the manager's own log shows a requested restart. With
     `reasons=("restart", "stop")` the question is whether the manager noted ANY kill of this kernel,
     which is how _graceful_term tells the delivery of a parked quiet request from a signal nobody sent
-    through the manager. The start bound is the one _recent_restart_audit
-    applies: the note this waits for lands AFTER our signal, so a stop note for our pid from before the
-    process existed is a predecessor's, left behind when a reboot reused the pid (the machine back inside
-    the window, the kernel up under the pid the previous one had, then a stray kill), and without the
-    bound it would read as the manager going down now. The tail is as deep as _recent_restart_audit's:
+    through the manager; the note's `trigger` then tells the stop of this kernel alone (`stop`, which
+    leaves the manager's park armed, so no delivery) from a restart or the self-bounce's `refresh` (the
+    delivery), which is why the row comes back and not a flag. The start bound is the one
+    _recent_restart_audit applies: the note this waits for lands AFTER our signal, so a stop note for our
+    pid from before the process existed is a predecessor's, left behind when a reboot reused the pid (the
+    machine back inside the window, the kernel up under the pid the previous one had, then a stray kill),
+    and without the bound it would read as the manager going down now. The tail is as deep as
+    _recent_restart_audit's:
     a service stop notes every kernel, and each session's own end-on-idle row lands in the same file,
     so a short tail would miss the note behind a dozen such rows and file the stop as a stray signal."""
     try:
         tail = (jd.STATE / "restart-audit.jsonl").read_text().strip().splitlines()
     except Exception:
-        return False
+        return None
     t0 = int(now if now is not None else time.time())
     born = int(_STARTED if started is None else started)
     for line in reversed(tail[-200:]):
@@ -21688,10 +21692,10 @@ def _manager_sigterm_row_for_us(now=None, window=90, started=None, reasons=("sto
         if not (isinstance(rec, dict) and rec.get("action") == "manager-sigterm" and isinstance(rec.get("t"), int)):
             continue
         if t0 - rec["t"] > window or rec["t"] < born:
-            return False
+            return None
         if rec.get("pid") in (None, os.getpid()) and rec.get("reason") in reasons:
-            return True
-    return False
+            return rec
+    return None
 
 
 def _unrequested_signal_reason(signum, be=None, wait=None, sleep=time.sleep, now=None, started=None):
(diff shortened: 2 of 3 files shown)


ksarma:

Reviewed inline: by reading only, since the handler fires the signal-exit path and probes pids. Commit 201da6fd closes item 1 of the #1302 review record; commit 34c6f4fe closes item 2. The manager writes every note's trigger as `trigger || reason`. Stop notes carry `stop`, or `refresh` from the self-bounce. Restart notes carry `restart` or `restart-all`. So the handler's check matches exactly the single-kernel stop and shutdownAll's stop, and no other note in main (`romp down`, #1306, is unmerged). `_audit_reason_text` reads the note as `manager-sigterm: stop`. `_drain_and_exit` stamps the note's t, and the walk never marks a note spent. `_DEPLOY_RESTART_REASONS` excludes the text, so the cool-down loses only the anchor the body names. `_parked_quiet_deploy` has one reader, `_main_drift_check`, behind the running-vs-checkout guard. A park left by a service stop therefore surfaces nowhere and expires at RESTART_EXPECT_MAX_S (20 minutes). The successor's restart note settles it in the walk. The rejected `rec = None` path would file `managerStopped: true` through the unrequested helper's default `stop` look. All eight callers of `_manager_sigterm_row_for_us` check truthiness only. The four source pins on `_graceful_term` hold. The added lines carry no private strings, hostname or home path. The fix label fits.

Follow-up:
- Nit, `kernel/kernel.py:53622`: the docstring's last sentence says the park stays on record for the kernel the quiet window will restart. After a stop of every kernel, the manager dropped its park, so the row is stale until it expires, as the body says.

Test evidence: nothing from the PR ran here. Local runs were py_compile of both files in a capped scope and the privacy scan. Red-first holds by reading. At main, the two stop/stop tests find the note truthy, keep the park, and fail at `main-converge` against `manager-sigterm: stop`. CI at 34c6f4fe is green on all 11 checks. The Python jobs report 11173 passed, 121 skipped, 1744 subtests.

Review run by the maintainer automated review session; verified against head 34c6f4fe.

ksarma:
Head note: the branch was updated with main, so the head moved from 34c6f4fe to 0cfa6b97. The new head is a mechanical merge of 34c6f4fe and 77cf5726: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 0cfa6b97.

PR #1308, opened 2026-09-10
Model catalog: a boot with a cache serves it and runs no helper; only an install's first boot fetches

henrypinkard:
Tier: fix

## What

The kernel's boot no longer runs the apiKeyHelper to refresh the model catalog when a catalog cache exists. The refresh runs Claude Code's configured helper in-process, and on the user's laptop that helper is a password-manager read that raises a desktop prompt; every kernel boot there, several deploy converges a day, put up a prompt nobody answered and the run timed out after 15 s (two boots minutes apart this morning, both timing out). The user asked for romp's unnecessary asks of the password manager to go; this was the one romp itself made.

**Change.** `_model_catalog_boot` installs the last fetched list as before and fetches from the Models API only when no cache exists (an install's first boot). With a cache the list serves as is, one stderr line says so, naming the cache's fetch time and that the list refreshes when a session reports a model it lacks, and `/version`'s `modelCatalog.source` reads `cache` until an event refreshes it. The refresh trigger is the existing staleness event alone (an unknown `claude-*` id from a reg or a pick, once per id per kernel life): the designed trigger, event over time; a boot is a clock. No periodic refresh was added. Every existing failure path of the fetch is unchanged and loud. An unreadable cache or one with no rows counts as no cache and fetches.

No card-move rule is involved.

## Review (lean, by hand)

The helper is the only boot-time caller of the refresh (pinned: no other `"boot"` fetch remains); the staleness event's dedup and single-flight are untouched. The cache loader's zero return covers absent, unreadable and empty caches alike, so the first-boot fetch still happens whenever nothing usable is on disk. The stderr line formats the cached `fetchedAt` when it is a number and says "an unknown time" otherwise, never raising. The off switch (`ROMP_MODEL_CATALOG=off`) still stops the first-boot fetch inside `_refresh_model_catalog`, as before.

## Tests

`tests/test_model_catalog.py`, new class on the hermetic fixture (per-test `XDG_STATE_HOME` root and `CLAUDE_CONFIG_DIR` with a counting fixture helper, the Models API a local fake): a boot with a cache present runs the helper zero times, asks the API nothing, serves the cache with source `cache` and its fetch time, and logs the line; a first boot with no cache fetches once and writes the cache; an unreadable cache and an empty one count as none; an unknown id still fires the refresh with a cache present; the kernel boots through the policy and no other boot-time fetch remains. Red before the change (the helper ran at boot), green after. `docs/reference.md`: the catalog sentence.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -707,7 +707,10 @@ token included. Its output must be one non-empty line with no whitespace, at
 most 16 KiB; a trailing newline is forgiven. A helper that fails is a problem
 row in the Log panel in static words. With no helper configured the catalog
 serves its cached list, or its built-in one, and the kernel log says why at
-each refresh attempt (boot, and once per model id it does not know); the
+each refresh attempt (an install's first boot, when no cache exists, and once
+per model id it does not know; a boot with a cache serves it, says so with the
+cache's fetch time, and never runs the helper: the helper can be a desktop
+prompt, and a boot is not an event); the
 pickers still work, and Claude Code's own alias table still tracks each
 family's newest. The fast-mode probe then leaves the CLI's own check standing
 and says nothing.
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -1913,6 +1913,26 @@ def _refresh_model_catalog(reason, _async=True):
     return True
 
 
+def _model_catalog_boot(_async=True):
+    """The catalog at kernel boot (T296, the user 2026-09-10): install the last fetched list, and fetch from
+    the Models API ONLY when no cache exists (an install's first boot). The fetch runs Claude Code's
+    apiKeyHelper, which on some boxes is a password-manager read that raises a desktop prompt; a boot is
+    a clock, not an event, and several deploy converges a day each put up a prompt nobody answered and
+    timed out. With a cache the list serves as is and refreshes on the designed trigger alone, the
+    staleness event (_note_unknown_model: an unknown claude-* id from a reg or a pick, once per id per
+    kernel life). One stderr line says so, naming the cache's fetch time; /version's modelCatalog reads
+    "cache" until an event refreshes it. Returns whether a fetch started."""
+    n = _load_model_catalog_cache()
+    if n:
+        fetched = _catalog_status.get("fetchedAt")
+        when = (datetime.fromtimestamp(int(fetched), timezone.utc).strftime("%Y-%m-%d %H:%M UTC")
+                if isinstance(fetched, (int, float)) and fetched else "an unknown time")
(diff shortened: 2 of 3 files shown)


ksarma:
Post-merge record: the owner merged this PR as 77cf5726 at head 2ec31d25 while this advisory review was running; no objection to the change. The 3 items below are all follow-ups.

Reviewed inline: the kernel diff is two hunks, the new `_model_catalog_boot` and its single call in `_sdk_locked`. `_refresh_model_catalog` (off switch, single-flight), `_note_unknown_model` (once per id per kernel life) and every fetch failure path are unchanged. The only remaining "boot" refresh call sits inside `_model_catalog_boot`. It runs only when `_load_model_catalog_cache` returns 0, which covers absent, unreadable, malformed and empty caches. Rows without a fetchedAt serve with "an unknown time". The "cache" source value already existed and has no reader outside `kernel.py` (`ui/` and `vscode-extension/src` grepped). On the design point: a boot carries no information about the model list. An unknown id from a reg or a pick does. Retiring the boot trigger therefore follows the repo's event rule. The cost is a cache that serves until a session reports an id it lacks. The seed moves with each release, so no picker or default goes wrong. The test fixture is hermetic: a per-test CLAUDE_CONFIG_DIR with its own counting helper, a scratch XDG_STATE_HOME, a loopback fake Models API. Privacy scan of the 91 added lines: zero hits. The body paraphrases and quotes no one. The fix label fits.

- `tests/test_api_health_hover.py:280` (follow-up): the stub `km._refresh_model_catalog = lambda why: None` predates the new call `_refresh_model_catalog("boot", _async=_async)`. At the head, `_sdk_locked`'s except swallows the TypeError. The construction test passes through the guard's failure branch with a traceback on stderr (reproduced with -s; absent at base). Give the stub an `_async=True` parameter, or stub `_model_catalog_boot` instead and add it to the saved names.
- `kernel/kernel.py:1688` and `tests/test_model_catalog.py:9` (nit): both still describe the refresh as keyed on boot.
- `kernel/kernel.py:1928` (nit): a cache whose fetchedAt is NaN, Infinity or past year 9999 raises in the formatter. The call site's except contains it, so only the one-line notice is lost.

Red-first verified. At base 025b032a with the head's `tests/test_model_catalog.py`, the five BootPolicy cases fail (four AttributeError on the missing `_model_catalog_boot`, one on the source pin) and 44 pass. A probe at the base running the old two-line boot with a cache present recorded one helper run and one API request. At the head the module and `tests/test_model_catalog_floor.py` pass (51). CI: eleven checks SUCCESS.

Review run by the maintainer automated review session; verified against head 2ec31d25.

PR #1309, opened 2026-09-10
Postal cards fold whenever the gist does not carry the whole message, and the head's meta never shrinks (T294)

henrypinkard:
## Summary

A postal card whose message is one paragraph could not be opened, and nothing past the head's ellipsis was reachable. Since the notice rewrite the card's head is one nowrap line, and the fold under it exists only when the gist differs from the message. For mail without a recipient caption the gist was the message's first line, unclipped, so a one-paragraph message compared equal to itself: no fold, no click. That is most sent mail, and every inbound message the recipient's judge has not captioned yet, the inbound question that is meant to open by default included. The pre-rewrite card rendered the whole body in exactly that case. Alongside it the head's meta text (the kind, the delivery state) shrank to a letter ("q…") beside a long gist.

## The fix

- `ui/webview/gist.ts` (new): the shared gist rule (`gistOf`, `collapseWs`, moved out of render.ts) and `postalHead(ev)`: the gist is the caption, else the first line CLIPPED by `gistOf`; the fold holds the full message whenever the gist does not carry all of it. A short one-liner keeps no fold, since nothing is hidden.
- `ui/webview/render.ts`: `renderPostalService` builds its head through `postalHead`; the in-place rule and `postalServiceSummary` are gone.
- `ui/webview/styles.css`: the POSTAL card's `.notice-meta` never shrinks (the gist is the one flexible slot of its head); scoped to the postal card, because an API-error head carries a sentence-long meta and a compact notice-group head a whole gist there, and both must still yield to the actions beside them on a narrow pane. A head-only postal line wraps instead of truncating (a bare link included, `overflow-wrap: anywhere`), with its glyph pinned to the first line.

**From the adversarial review, folded in before arming.** The no-shrink meta was global at first and would have pushed an API-error card's buttons off a narrow pane; it is the postal card's alone now. The wrap rule had no break opportunity for an unbroken token, so a one-line link message still ran off the head; it breaks anywhere now. The fold rendered the TRIMMED message, which flattened a leading indented code block into prose; the fold holds the message as sent and only the comparison trims. The comment above the head rule quoted the user verbatim and described the old rule; it paraphrases and describes the rule that is there. Measuring at a 390 px pane: the compact notice-group head ("2 notices" beside a 90-character gist in its meta slot) had its gist squeezed to 20 px by proportional flex shrink, before and after this change alike, so the gist keeps a floor of `min(100%, 10ch)` on the group head and the postal card (not as the base rule: a floor on the API-error head pushed its buttons off the pane). And on a phone-width postal head even the short kind and delivery words do not fit beside that floor, so the postal head is a size container and under 360 px its meta yields again to a floor of its own inside the card (`@container (max-width: 360px)`). Every postal meta measured inside its head at 390, 520 and 1000 px; whole at 520 and 1000.

## Tests

- `ui/webview/postal-head.test.ts` (new): unit tests of the rule (a one-paragraph message without a caption folds; uncaptioned inbound mail folds; a short one-liner has no fold; a multi-line message folds; a caption is the gist), red on the old rule, plus pins on the render.ts call site and the two CSS rules.
- `ui/webview/postal-expand.test.ts`: the rule pins moved to gist.ts.

Verified on a hermetic served chat with synthetic mail (inbound delegation, coordination and question, a sent one-paragraph message, a sent multi-line message): before, four of five cards were slim head-only lines with no click; after, all five fold and open, the inbound question opens by default, and the kind and delivery state read in full.

Tier: fix

### DIFF
--- a/ui/webview/gist.ts
+++ b/ui/webview/gist.ts
@@ -1,0 +1,30 @@
+// The gist rule every text-derived notice head shares, and the postal card's head built on it. A pure module
+// (no DOM) so the rule is unit-testable: render.ts imports it, and the one-line heads it builds, a peer's mail,
+// a background report, a romp notice, a harness note, all clip the same way.
+
+// The first meaningful line of a message, clipped for a head.
+export function gistOf(text: string, max = 90): string {
+  const lines = text.split("\n").map((l) => l.trim());
+  const first = lines.find((l) => l && !l.startsWith(">")) || lines.find((l) => l) || text.trim();
+  return first.length > max ? first.slice(0, max - 2).replace(/\s+\S*$/, "") + "…" : first;
+}
+
+export const collapseWs = (s: string): string => s.replace(/\s+/g, " ").trim();
+
+// A postal card's head: the gist the head shows, and the message the fold holds beneath it, as sent (null when
+// the gist already carries the whole message, so nothing is hidden and there is nothing to open).
+export interface PostalHead {
+  gist: string;
+  body: string | null;
+}
+
+export function postalHead(ev: { body?: string; summary?: string }): PostalHead {
+  const raw = ev.body || "";
+  const full = raw.trim();
+  const cap = ev.summary && ev.summary.trim();
+  const gist = cap || gistOf(full);   // CLIPPED: an unclipped first line equalled a one-paragraph message and left it no fold (T294)
+  // the fold holds the message AS SENT, not the trimmed copy the comparison uses: a leading indent is markdown (four
+  // spaces open a code block) and the trim would have flattened it into prose
+  const body = !!full && collapseWs(full) !== collapseWs(gist) ? raw : null;
+  return { gist, body };
+}
--- a/ui/webview/postal-expand.test.ts
+++ b/ui/webview/postal-expand.test.ts
@@ -6,7 +6,7 @@
 //
 // 2026-07-25 (the user, from a real sent card): three more guarantees pinned here —
 //   1. the expand is KEYED (openFolds), because the unkeyed toggle was silently re-collapsed by the
-//      next kernel push ("it expands for like a second and then something collapses it");
+//      next kernel push (the user watched a card open and snap shut a moment later);
 //   2. expanded shows the full message ALONE — the summary line was repeating the same words right
 //      above the body;
 //   3. the collapsed fallback is clamped by CSS to two full lines, not pre-truncated at 100 chars,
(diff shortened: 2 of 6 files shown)


ksarma:
Post-merge record: the owner merged this PR as e3b599a6 at head c3cf184e before this advisory review posted; no objection to the change as shipped. Two items below: a layout edge to follow up and a PR-body correction.

Reviewed inline: `postalHead` in `gist.ts` takes the caption, else `gistOf`'s 90-character clip of the first line. It returns the message as sent whenever the collapsed-whitespace gist differs from it. `gistOf` is byte-identical to the base's `render.ts` copy, and `collapseWs` gains only a return annotation, so no other head's gist changed. `renderPostalService` keeps the fold key `notice:postal:<mid or uuid>` in `openFolds`. It seeds an inbound question open once through `noticeSeeded`. It toggles through `data-act=noticetoggle` on the `document.body` delegate. So the open state survives re-renders, and the click is safe. The fold body still goes through `md()` with the sender's repo: the same sanitizer path and the same raw string as before. `postalServiceSummary` has no caller left. In `styles.css`, the rigid meta and the 10ch gist floor are scoped to the postal card and the group head. The base `.notice-meta` and `.notice-gist` rules are untouched. `container-type` on the postal head does not collapse it: the head is a block child of the card, and nothing inside it is positioned. No new font-size or colour appears. `feed.css` needs no mirror, since only `render.ts` builds these cards, on the chat page. The 174 added lines scan clean, and the fixtures are the notes-api demo domain.

- Follow-up, `ui/webview/styles.css:2323`: with a long peer name, the head overflows the card between the 360px breakpoint and about 480px. Above 360px the source chip, the meta and the gist floor are all rigid, so a long name has nothing to take space from. Measured in headless Chromium on synthetic markup, base CSS against head CSS: a 25-character name at a 420px pane overflows the head by 72px, with the meta 59px past the card edge. At 380px it overflows by 26px. The base ellipsized the meta there. A 3-character name never overflows. A 13-character one overflows by at most 6px, inside the padding. The fonts differ from the webview's, so the edges of the band shift.
- Nit, PR body, Tests section: under the old unclipped rule, only the one-paragraph and uncaptioned-inbound tests go red. The short one-liner, multi-line and caption cases pin behaviour the base already had. The CSS test now pins seven rules, not two.

Tests: typecheck and the test bundle are clean at c3cf184e. The head was reconstructed from the PR diff on its parent 025b032a, with all six blobs matching the remote tree. The 14 affected compiled suites pass 162/162. A mutation restoring the unclipped first-line rule reds exactly tests 1 and 2 of `postal-head.test.ts` (control 8/8). At the base the new tests cannot bundle, since `gist.ts` does not exist there. CI at c3cf184e: all nine checks green.

Review run by the maintainer automated review session; verified against head c3cf184e.

PR #1311, opened 2026-09-10
Catalog boot residues: the health-hover stub covers the boot helper, the words follow the first-boot rule, a corrupt fetch time cannot raise

henrypinkard:
Tier: fix

## What

Three residues of the catalog boot change (the post-merge review):

1. **A stale stub hid a swallowed error.** The health-hover test that builds the backend with the kernel's own start stubbed the cache loader and the refresh with their old signatures; on main the boot path calls the new boot helper, which handed the stubbed refresh an unexpected keyword, and the boot's own `except` swallowed the traceback to stderr while the test stayed green. The test now stubs the boot helper itself and asserts that stderr carries no catalog-boot line, so the leg is proven clean. Red first: the assertion fails against the stale stubs.
2. **The words follow the rule.** The catalog's events header, the cache loader's docstring, the staleness event's inflight example and the test module's docstring said boot was a refresh event; they now say what holds: an install's first boot with no cache fetches, then only the staleness event, and never a boot that has a cache.
3. **The fetch-time line cannot raise.** A corrupt `fetchedAt` in the cache (out of range, NaN, negative beyond the epoch, a string, null) made the timestamp formatting raise inside the boot's try after the cache had already installed, a swallowed traceback; the formatting is guarded and says "an unknown time".

No behaviour of the catalog changes beyond the guarded line; no card-move rule is involved.

## Review (lean, by hand)

The stub replaces exactly the boot's one catalog call with the same return shape (no fetch); the stderr capture wraps only the build. The guard catches the exceptions the formatting can raise (`ValueError`, `OverflowError`, `OSError`, `TypeError`) and nothing that would hide a different fault. The comment rewrites name the same functions the code calls.

## Tests

`tests/test_api_health_hover.py`: the backend build asserts no catalog-boot stderr line (red against the stale stubs). `tests/test_model_catalog.py`: a cache whose `fetchedAt` is a string, out of range, NaN, hugely negative or null still serves as the cache, runs no helper, says "an unknown time" and writes no traceback.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -1684,9 +1684,11 @@ _models_rev = [int(time.time() * 1000)]
 # new version ids into the families, ADD-ONLY: an id the API omits but the seed knows stays (key-scoped
 # visibility differs per account); a dated snapshot, a suffixed variant (-fast) or the pre-4 naming
 # never joins (routing ids, not versions — the alias-table lesson); every family stays newest-first.
-# Cached durably (STATE/model-catalog.json) so a dead API never blanks a picker. Refreshed on exact
-# EVENTS — boot, and a claude-* id reaching a set path or the pick store that the merged list does not
-# know (the very moment staleness bites) — never on a timer. Unreachable API = serve seed+cache and
+# Cached durably (STATE/model-catalog.json) so a dead API never blanks a picker. Fetched on exact
+# EVENTS — an install's first boot, when no cache exists, and a claude-* id reaching a set path or the
+# pick store that the merged list does not know (the very moment staleness bites) — never on a timer,
+# and not at a boot that has a cache (T296: the fetch runs the apiKeyHelper, a desktop prompt on some
+# boxes; a boot is a clock). Unreachable API = serve seed+cache and
 # SAY SO (stderr + /version), never a quietly stale list. ROMP_MODEL_CATALOG=off disables the fetch
 # outright — hermetic labs must never reach the network; ROMP_MODELS_URL points tests at a fake.
 MODEL_CATALOG_FILE_NAME = "model-catalog.json"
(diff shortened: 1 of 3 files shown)


ksarma:
Post-merge record: the owner merged this PR as 3abe4592 at head daa2c2f6 before this review ran; no objection. The 2 items below are nits, record only.

Reviewed inline: the diff closes the three follow-ups of the #1308 record. The hover construction test replaces the two stale stubs with `_model_catalog_boot = lambda _async=True: False`. That stub matches the helper's signature and its no-fetch return, which `_sdk_locked` discards. The test captures stderr around the `_sdk_locked()` call alone. It asserts the exact prefix that call's except writes, `model catalog boot:`, which the healthy line `model catalog (boot):` cannot match. Neither old name appears elsewhere in the module. The four prose rewrites (the events header, the `_load_model_catalog_cache` and `_note_unknown_model` docstrings, the test module docstring) name the functions the code calls. They state the rule the helper implements: a first boot with no servable cache fetches, the staleness event refreshes, a boot with a cache never fetches. The guard in `_model_catalog_boot` wraps the single `when =` expression; `sys.stderr.write` sits outside it, so it hides nothing beyond the formatter. Both fixtures were read before running. The catalog module's fixture is a per-test CLAUDE_CONFIG_DIR with a counting fixture helper and a loopback fake; the hover module's stubs the boot helper whole. Neither reaches the real helper, the live state or the network. Privacy scan of the 41 added lines: zero hits; the body quotes no one. The fix label fits: a test-hygiene defect and a guard, each with a red-first test.

- `tests/test_api_health_hover.py:261` (nit): the docstring still lists the catalog cache and refresh among the stubbed names. The tuple now stubs the boot helper.
- `kernel/kernel.py:1934` (nit): TypeError in the except tuple is unreachable. The isinstance check sends every non-number to the else branch. Dropping it leaves the corrupt-stamp test green. Harmless.

Red-first verified. At base 77cf5726 with the head's two test files, the corrupt-fetchedAt case fails with OverflowError at the 1e30 stamp and 83 pass. The hover assertion is red against the stale stubs: the base file plus only the new capture and assertion fails on the swallowed TypeError traceback. At head both modules plus tests/test_model_catalog_floor.py pass (86). Mutation: dropping ValueError, OverflowError or OSError lets NaN, 1e30 or -1e18 escape respectively. CI: eleven checks SUCCESS at daa2c2f6.

Review run by the maintainer automated review session; verified against head daa2c2f6.

PR #1312, opened 2026-09-10
Restart-classifier tests: the fixture's git forbids background work, so no child writes into .git during teardown (T298)

henrypinkard:
Fixes the flake in `tests/test_restart_classifier.py` where `RealDiffShapes`' teardown raised `OSError: [Errno 39] Directory not empty: '.git'` from `TemporaryDirectory.cleanup` (the Python 3.10 job of a webview-only PR, 2026-09-10).

**Writer, found at source.** `git commit` spawns `git maintenance run --auto`. On current git that child detaches from its parent (`maintenance.autoDetach`, on by default in recent git; on older git `gc --auto` detached once it had work), so it can still be writing into `.git` when the fixture's `rmtree` runs a few milliseconds after the class's last commit. Git's own trace (`GIT_TRACE`) shows the spawn on every fixture commit: 450 `run_command: git maintenance run --auto --quiet` lines over thirty runs locally. Locally (git 2.43) the maintenance process does not detach and the race never fired in thirty niced runs, so the flake is not reproducible here; the pin is red-first on the mechanism instead.

**Fix.** Every git the fixture runs carries `-c maintenance.auto=false -c maintenance.autoDetach=false -c gc.auto=0 -c gc.autoDetach=false -c core.fsmonitor=false`, and `setUpClass` writes the same keys into the repo's own config so a git the kernel runs against the repo obeys them too. Nothing retries the `rmtree`.

**Pin.** `test_the_fixture_forbids_background_git_work` reads each key back from the repo's local config, checks the helper's flags carry it, and traces a commit through the helper: the trace names the commit and no maintenance or gc child. Before the fix the trace named `git maintenance run --auto --quiet`, so the test was red.

Five further traced runs of the class after the fix show zero `run_command` children of any kind.

### DIFF
--- a/tests/test_restart_classifier.py
+++ b/tests/test_restart_classifier.py
@@ -29,8 +29,19 @@ load_source("romp_judge", os.path.join(BIN, "romp-judge"))
 km = load_source("romp_kernel_rclass", os.path.join(BIN, "romp-kernel"))
 
 
-def _git(repo, *args):
-    r = subprocess.run(["git", "-C", str(repo)] + list(args), capture_output=True, text=True)
+# T298: every git the fixture runs forbids BACKGROUND work. `git commit` spawns `git maintenance run --auto`,
+# which on current git detaches from its parent (maintenance.autoDetach, on by default in recent git; older git's
+# `gc --auto` detached once it had work) and can still be writing into .git while tearDownClass removes the
+# temp repo — the CI flake "Directory not empty: '.git'" raised by TemporaryDirectory.cleanup's rmtree (the
+# Python 3.10 job, 2026-09-10). The repo's own config carries the same keys (setUpClass), so a git the KERNEL
+# runs against the repo obeys them too; fsmonitor is off for the same reason on hosts where it has a daemon.
+GIT_NO_BACKGROUND = {"maintenance.auto": "false", "maintenance.autoDetach": "false", "gc.auto": "0",
+                     "gc.autoDetach": "false", "core.fsmonitor": "false"}
+_GIT_C = [x for k, v in GIT_NO_BACKGROUND.items() for x in ("-c", "%s=%s" % (k, v))]
+
+
+def _git(repo, *args, env=None):
+    r = subprocess.run(["git", "-C", str(repo)] + _GIT_C + list(args), capture_output=True, text=True, env=env)
     assert r.returncode == 0, r.stderr
     return r.stdout.strip()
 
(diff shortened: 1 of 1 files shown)


ksarma:
Post-merge record: the owner merged this PR as b359cbf4 at head ef05b8e3 before this review ran; no objection.

Reviewed inline: the diff is `tests/test_restart_classifier.py` only, +34/-2, and every clause of the body matches it. It has the five-key `GIT_NO_BACKGROUND` map, with `_GIT_C` on every helper call. `setUpClass` writes the same keys into the temp repo's local config. `_git` takes an `env` parameter. `tearDownClass` is unchanged, so nothing retries rmtree. The writer is confirmed rather than assumed. The flake is CI run 34499858395 attempt 1 (Python 3.10, runner git 2.55.0), where rmdir of `.git` itself hit ENOTEMPTY. On git 2.55.0 `git commit` spawns `git maintenance run --auto --quiet --detach`. It daemonizes and kept writing under `.git` for about 170 ms after the commit returned (strace). A tight repro on Python 3.10 raised the exact CI error in 33 of 150 pre-fix iterations and 0 of 150 with the flags. On the box's git 2.43 the child does not detach, as the body says. `maintenance.auto=false` alone removes the child on 2.43, 2.47 and 2.55. `gc.auto=0` alone does not on 2.43, so both keys are needed. `maintenance.autoDetach`, unknown to 2.43, is accepted silently. The kernel's git calls under test (diff and show) are read-only, waited, and run with cwd at the temp repo. So the local config governs them and no other writer remains. `GIT_TRACE` appends by path and the env is per call. Privacy scan of the 34 added lines: clean.

Follow-ups, nits at the owner's discretion:

- `tests/test_restart_classifier.py:225`: `assertNotIn("maintenance", t)` is a bare substring check. The trace also carries hook and repo paths (`GIT_TRACE` reaches git's children). A hooks path or `TMPDIR` containing that word gives a false red. `"run_command: git maintenance"`, mirroring line 226, pins the spawn exactly.
- `tests/test_restart_classifier.py:215`: with the fix absent, the first red is `_git`'s bare returncode assert on `git config --local --get` (exit 1, empty stderr), not the trace the body cites. `--default ""` (git >= 2.18) would let the assertEqual carry the diagnosis.

Red-first: at base 77cf5726 with the head's test file, the module is green by construction (fixture and pin share the file). With the fix removed, the pin fails at the config assertion. With only the trace leg kept, it fails on `run_command: git maintenance run --auto --quiet` on git 2.43 and 2.55. Head: 23 passed, with `HOME` unchanged and with `GIT_CONFIG_GLOBAL=/dev/null`. CI: 11 of 11 checks SUCCESS at ef05b8e3.

Review run by the maintainer automated review session; verified against head ef05b8e3.

PR #1313, opened 2026-09-10
perf-bench: the kernel and its SDK backend load through load_source, so the tool runs clean under the load_module() deprecation

ksarma:
## Symptom

tools/perf-bench.py loaded the kernel and its SDK backend with `SourceFileLoader(...).load_module()`, the last two such calls outside the tests. Python's default filters drop the deprecation warning (it is raised from importlib's own frame), so a plain run shows nothing and CI never did, although every interpreter CI runs (3.10 to 3.13) deprecates the call. Under a filter that surfaces DeprecationWarnings (`PYTHONWARNINGS=default`, `-X dev`) the tool warns at the kernel load. With warnings promoted to errors the run dies before its first row, with a traceback in place of the tool's own error line. On Python 3.15, where `load_module()` is removed, the tool loads no kernel at all: `AttributeError: 'SourceFileLoader' object has no attribute 'load_module'` at the first load.

## Cause

`load_kernel` still loaded kernel/kernel.py and kernel/sdk_backend.py with `SourceFileLoader(...).load_module()`. kernel/ and tests/ load by file path through kernel/loadsource.py's `load_source` (#1278), which keeps `load_module()`'s sys.modules semantics without the deprecated call; the tool was left out of that change.

## Fix

`load_kernel` bootstraps `load_source` from kernel/loadsource.py with the four-line idiom kernel/kernel.py uses (`spec_from_file_location`, `module_from_spec`, `exec_module`) and loads both modules through it. The existing `sys.modules` guard on the backend is unchanged and now has a test: `load_source` executes a name already in `sys.modules` again, in place, so the guard is what keeps a backend the kernel registered to one execution. Two placement choices:

- The loader comes from the tool's own checkout, not from `--repo`. It is stdlib glue independent of the revision under measurement, so a candidate checkout that predates kernel/loadsource.py stays benchable, and a `--repo` holding only kernel.py and sdk_backend.py is still refused with "this kernel lacks _live_scope" (`test_a_kernel_lacking_a_builder_is_refused`) rather than failing on a missing loader.
- The bootstrap sits inside `load_kernel`, not at module level, because the tool module loads no romp code at import; tests/test_perf_bench.py loads the tool in-process for its unit tests on that basis.

kernel/loadsource.py's docstring gains a clause naming the tool among the module's loaders.

## Tests

tests/test_perf_bench.py drives the tool as a subprocess, so the child's `PYTHONWARNINGS` is the executing witness; the module runs under pytest against a rebound state root and dead ports, never a live kernel.

- New: `test_the_tool_loads_the_kernel_clean_under_the_load_module_deprecation_as_an_error` plants a `--repo` holding only kernel.py (`x = 1`) and an empty sdk_backend.py, runs the tool with `PYTHONWARNINGS=error:the load_module() method is deprecated:DeprecationWarning`, and asserts rc 1, "this kernel lacks _live_scope" on stderr, no "DeprecationWarning", no "Traceback". Before the change the child raises at the first load, main's BaseException handler prints "perf-bench: DeprecationWarning: the load_module() method is deprecated ..." followed by the traceback, and the refusal line never appears, so the assertIn fails.
- New: `test_an_sdk_backend_the_kernel_already_loaded_is_not_executed_again` plants a kernel that registers its sibling sdk_backend.py as `romp_sdk_backend` at import, with the backend appending a line to a marker file on every execution, and asserts the marker holds one line once the run has ended on the refusal. Replacing the guard with an unconditional `load_source` of the backend makes it two lines and fails the test.
- `run_tool` now appends that same filter to every child's `PYTHONWARNINGS` (the message alone, so no other warning is promoted; the process environment's own filters stay, and a `PYTHONWARNINGS` passed through `env_extra` replaces the variable as every `env_extra` key does), so every subprocess-driven test covers the full bench path, including the real kernel and SDK backend loads. Before the change 18 of the module's 44 tests fail under it; after, all 44 pass. The named tests are the ones that say what broke.
- The filter matches the message by literal prefix, the same text on 3.10 through 3.14 ("... slated for removal in Python 3.12" through 3.12, "... 3.15" from 3.13). Checked by hand with a planted `--repo` on 3.10, 3.11, 3.12, 3.13, 3.14 and 3.15.0b4: under default filters the previous tool prints no warning on 3.10 to 3.14 and ends on the refusal line, under `PYTHONWARNINGS=default` it prints the warning, and under the error filter it dies with the traceback; on 3.15.0b4 it fails with the AttributeError under every setting. The changed tool ends on the refusal line with no warning and no traceback in every cell.
- Full suite on this head (Python 3.12): 11281 passed, 41 skipped, 1744 subtests passed.

Tier: fix


ksarma:

One wording nit, left as is: a copy of the tool outside its checkout now fails with a raw `FileNotFoundError` traceback. The missing-kernel case beside it raises `BenchError`.

Reviewed inline: `load_kernel` bootstraps `load_source` from `kernel/loadsource.py` beside the tool's own file, through `spec_from_file_location`, `module_from_spec` and `exec_module`, with no `sys.modules` insert. `_TOOL_FILE` is `realpath(__file__)` at module level, so the path holds as a script and when the tests load the tool in-process. Nothing else registers `romp_loadsource`. The base held exactly the two `load_module()` calls the PR body names, `tools/perf-bench.py:736` and `:739`. No code-line call remains outside `tests/`. `load_source` re-executes an existing name in place. The unchanged `sys.modules` guard on `romp_sdk_backend` therefore keeps a kernel-registered backend to one execution. A second `load_kernel` in one process re-executes the kernel in place, as `load_module` did. The appended `PYTHONWARNINGS` entry is a literal prefix (`warnings._setoption` re.escapes it), identical on 3.10 through 3.14. It is joined after the inherited value, and `env_extra` still replaces the variable. Planted kernels are synthetic. Every child runs against a `mkdtemp` state with the manager variables stripped.

Tests: at the base with the head's test file, 18 of 44 fail. Each fails on the DeprecationWarning raised at line 736, with a traceback and no refusal line. At the head all 44 pass. Dropping the backend guard fails only `test_an_sdk_backend_the_kernel_already_loaded_is_not_executed_again`. Bootstrapping the loader from `--repo` fails four tests, led by `test_a_kernel_lacking_a_builder_is_refused`. Privacy scan of the 72 added lines clean. CI at head: 12 of 12 checks green.

Review run by the maintainer automated review session; verified against head a9c86a91.

ksarma:
Head note: the branch was updated with main, so the head moved from a9c86a91 to 77c32128. The new head is a mechanical merge of a9c86a91 and 05c8d21f: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 77c32128.

PR #1314, opened 2026-09-10
Tests: one shared git runner for throwaway-repo fixtures forbids background work; the restart-classifier fixture moves onto it (T299, 1 of 2)

henrypinkard:
Follow-up to the restart-classifier flake fix (the `Directory not empty: '.git'` teardown error): the no-background git configuration moves into one shared runner for every test fixture that builds a throwaway git repository. This is the first of two PRs; the second moves the twelve other files onto it (kept separate so each stays reviewable).

**The helper**, `tests/git_fixture.py`: `GIT_NO_BACKGROUND` (maintenance.auto, maintenance.autoDetach, gc.auto, gc.autoDetach, core.fsmonitor), `git(repo, *args, env=None, ident=None, timeout=60, check=True, text=True)` running every command with the keys as `-c` flags, `init_repo(path, *init_args, ...)` writing them into the repo's own config at init, and `forbid_background(repo)` for a clone or worktree the kernel will run git against. The docstring carries the why: `git commit`, fetch and merge spawn `git maintenance run --auto`, which on recent git detaches from its parent and can still be writing into `.git` while a fixture's `TemporaryDirectory` removes the repo. `tests/__init__.py` registers the module under its bare name the way it does `romp_load`; `tests/README.md` says where fixture git goes.

**Pins**, `tests/test_git_fixture.py`: the keys ride every invocation even against a repo that lacks them; `init_repo` and `forbid_background` land them in the repo; a commit through the runner, and a plain fetch (no flags, as the kernel's own subprocess would run it) in a `forbid_background` clone, spawn no maintenance or gc child, read off `GIT_TRACE`; failures raise `CalledProcessError` with the streams attached and `check=False` returns the process; identity, env and init args pass through. The restart-classifier fixture's runner delegates to the shared one, keeping its shape and its own per-file pin (its kernel-run git reads the repo). No assertion changes beyond that pin's constant rename.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -130,6 +130,23 @@ the developer's git configuration (CI has none), and the env identity outranks
 exports its own `GIT_AUTHOR_*` after the floor. `tests/test_tempdir_hygiene.py`
 and `tests/git-hermetic.bats` pin all of it.
 
+**A fixture that builds a throwaway git repository runs git through
+`tests/git_fixture.py`** (`from git_fixture import git, init_repo,
+forbid_background`, registered under the bare name like `romp_load`). Every
+command it runs carries `-c` flags that forbid background work
+(`maintenance.auto`, `maintenance.autoDetach`, `gc.auto`, `gc.autoDetach`,
+`core.fsmonitor`), `init_repo` writes the same keys into the repo's own config,
+and `forbid_background` does so for a clone or worktree the kernel will run git
+against: `git commit`, fetch and merge spawn `git maintenance run --auto`,
+which on recent git detaches from its parent and can still be writing into
+`.git` while the fixture's `TemporaryDirectory` removes the repo (the CI flake
+`Directory not empty: '.git'` from `rmtree`, 2026-09-10). A file keeps its own
+thin runner (its identity, timeout and return shape) and delegates the body;
+read-only git against the real checkout stays a plain `subprocess.run`.
+`tests/test_git_fixture.py` pins the runner with git's own trace: a commit
+through it, and a plain fetch in a `forbid_background` clone, spawn no
+maintenance or gc child.
+
 **A served-page class copies the built `vscode-extension/dist/` with
 `tests.dist_copy.copy_dist`, never `shutil.copytree`.** Under `pytest -n` a
 sibling class's build renames or removes its staging files
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -81,3 +81,7 @@ os.environ["ROMP_SERVE_PORT"] = "1"
 # sys.path itself. One module object either way.
 from . import romp_load as _romp_load  # noqa: E402
 sys.modules.setdefault("romp_load", _romp_load)
+# `from git_fixture import git, init_repo` (tests/git_fixture.py, T299): the throwaway-repo fixtures' one git
+# runner, registered the same way for the same reason.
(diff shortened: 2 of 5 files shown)


ksarma:
Post-merge record: the owner merged this PR as 3815ed6f at head 266b2ad8 before this review ran; no objection, two follow-ups below at the owner's discretion.

Reviewed inline: the diff is tests only, five files, +198/-17, and every clause of the body matches it. `tests/git_fixture.py` has the five-key `GIT_NO_BACKGROUND` map and its `-c` expansion `GIT_C`. `git()` runs `git -C <repo> <GIT_C> [identity flags] <args>` with both streams captured. With `check` set it calls `check_returncode()`, so a `CalledProcessError` carries both streams. `timeout` passes straight through, and a `TimeoutExpired` propagates. `env=None` inherits the environment. Identity flags follow `GIT_C`, so a mapping could override a key, harmlessly. `forbid_background` writes the keys into the repo's own config. `init_repo` runs init, then `forbid_background`, then identity. `tests/__init__.py` registers the bare name with the same `sys.modules.setdefault` as `romp_load`. The import resolves under pytest, under `python -m unittest tests.x`, and in a direct script run. The restart-classifier `_git` delegates with `check=False` and keeps its returncode assert and stripped stdout. `setUpClass` uses `init_repo`. The only assertion change is `_GIT_C` to `GIT_C`. The README paragraph and the docstring match what the #1312 record verified on git 2.43, 2.47 and 2.55. The fetch test clones a local path. Privacy scan of the 198 added lines: clean.

Follow-ups:

- `tests/test_git_fixture.py:102`: `assertNotIn("maintenance", t)` on the fetch trace is a confirmed false red. The trace names the fetch source by absolute path (`built-in: git upload-pack <src>`). The conftest nests its temp root under the incoming `TMPDIR`. A temp root whose path contains the word therefore fails this test on a green tree. Reproduced with a `.../maintenance-probe` root: 1 failed, 1 passed, no maintenance child in the trace. `"run_command: git maintenance"`, the same form as the gc pin on line 103, matches the real spawn and cannot match a path. The same shape is at `tests/test_git_fixture.py:81` and `tests/test_restart_classifier.py:217`. There a commit trace carries no path, so exposure needs a hook path under a bare unittest run.
- `tests/test_restart_classifier.py:39` (carried from #1312): a missing key fails with an empty message. With a key missing from the repo config, the pin at line 207 first fails in the runner's bare returncode assert. That message is empty because `git config --local --get` exits 1 with empty stderr. The assertEqual also never names the key.

Red-first: at base b359cbf4 with the head's test files, both modules fail at import (`ModuleNotFoundError: git_fixture`). Three mutations at head each fail tests. Dropping `maintenance.auto` fails four tests: the dict pin, both trace pins, and the restart-classifier pin. The plain fetch then spawns the maintenance child on git 2.43. `forbid_background` writing nothing fails the two landing pins and the restart-classifier pin. `GIT_C = []` fails the dict pin, the invocation pin and the restart-classifier pin. The commit-trace pin stays green under that mutation because the config `init_repo` wrote covers it. Head: 29 passed (6 in `test_git_fixture`, 23 in `test_restart_classifier`) in a capped scope with `HOME` unchanged. The conftest floor already nulls the global config under pytest. CI: 10 of 10 checks SUCCESS at 266b2ad8.

Review run by the maintainer automated review session; verified against head 266b2ad8.

PR #1315, opened 2026-09-10
Tests: twelve throwaway-repo fixtures move onto the shared no-background git runner (T299, 2 of 2)

henrypinkard:
Second of two PRs (T299): the twelve remaining test files that build throwaway git repositories move onto `tests/git_fixture.py`, the shared runner that forbids background git work (the `Directory not empty: '.git'` teardown flake).

**Mechanical, and no assertion changes.** Each file keeps its local runner's name, signature and what it returns or raises (a stripped stdout, a `CompletedProcess`, `None`; `AssertionError` or `CalledProcessError`); only the body delegates to `git()`, passing the file's own identity, environment, timeout and text-ness. Fixture inits go through `init_repo`; clones, worktrees and bare repos the kernel runs git against get `forbid_background` right after they are made; inline git calls on fixture repos move onto the runner. Read-only git against the real checkout (`git --version`, `--exec-path`, rev-parse on the repo running the tests) stays a plain `subprocess.run`. The touched assertion lines in the diff are exactly the pins below.

**Per-file pins.** Eleven of the twelve make the kernel, or a script under test, run its own git against a fixture repo; each carries one small test that the repos the kernel touches hold `maintenance.auto=false` in their local config. `test_tempdir_hygiene` runs all its git through its own runner, so the shared pin in `tests/test_git_fixture.py` covers it and it adds none.

Files: test_kernel_update, test_kernel_remote_update, test_converge_main_branch, test_github_repo, test_file_github, test_kernel_fork_burn, test_git_pointer_file_bytes, test_kernel_worktree_display, test_tempdir_hygiene, test_sdk_kernel, test_stage0_log_readers, test_perf_bench.

### DIFF
--- a/tests/test_converge_main_branch.py
+++ b/tests/test_converge_main_branch.py
@@ -11,9 +11,9 @@ checkout, a second clone to advance the remote's main.
 
 Hermetic state, synthetic content; the restart leg is stopped before the manager POST."""
 import os
-import subprocess
 import tempfile
 import unittest
+from git_fixture import git as fixture_git, init_repo, forbid_background
 from romp_load import load_source
 from pathlib import Path
 from unittest import mock
@@ -35,8 +35,13 @@ else:
     os.environ["ROMP_STATE_DIR"] = _PREV_STATE_DIR
 
 
+# T299: every git here goes through the suite's shared runner (tests/git_fixture.py), which forbids BACKGROUND
+# work: `git commit`, fetch and merge spawn `git maintenance run --auto`, which on recent git detaches from its
+# parent and can still be writing into .git while the TemporaryDirectory removes the repo (the CI flake
+# "Directory not empty: '.git'", tests/test_restart_classifier.py, 2026-09-10). init_repo writes the same keys
+# into each fixture repo's own config, so the git the KERNEL runs against the checkout obeys them too.
 def git(cwd, *args):
-    r = subprocess.run(["git", *args], cwd=str(cwd), capture_output=True, text=True, timeout=30)
+    r = fixture_git(cwd, *args, timeout=30, check=False)
     if r.returncode != 0:
         raise AssertionError("git %s failed: %s%s" % (" ".join(args), r.stdout, r.stderr))
     return r.stdout.strip()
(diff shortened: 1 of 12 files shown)


ksarma:
Post-merge record: the owner merged this PR as 54baf25e at head fdfd4c1f before this review ran; no objection to what shipped. The two items below are follow-up nits.

Reviewed inline: every hunk in the twelve files against the body's claims. Each local runner keeps its name, signature, identity, environment and return or raise shape. Only the body delegates to `git()` from `tests/git_fixture.py`. Fixture inits go through `init_repo`. Every clone, bare clone and linked worktree the kernel or a script under test runs git against gets `forbid_background` right after it is made. No assertion line is removed. The eleven added assert lines are the per-module pins on `maintenance.auto=false` in local config; `test_tempdir_hygiene` adds none, as stated. The git left as plain `subprocess.run` is read-only against the real checkout and touches no fixture: `--version` and `--exec-path` in `test_file_github.py:99` and `:616`, `rev-parse` in `test_perf_bench.py:331` and `test_sdk_kernel.py:524`. No module makes the kernel clone at run time, so no kernel-made repo escapes the fixtures' configuration. All twelve modules run in-process against temp repos with a rebound state root and dead manager ports, so they were safe to run here. The 277 added lines carry no private strings, hostname or home path; identities are synthetic.

Follow-ups:
- `tests/test_github_repo.py:156`: the pin reads the five class fixtures, not the bare clones at `:798` and `:812`, so deleting both `forbid_background(bare)` calls leaves the module green. The kernel's git there is `rev-parse` only, so no flake follows. Adding one bare clone to the pin would close the gap. Also, `forbid_background` on a linked worktree writes the parent clone's config, which `init_repo` already set, so those calls are no-ops the pins cannot see.
- Body wording: runners that had no timeout before now inherit `git()`'s 60 s default rather than the file's own (eight of the twelve files). Where a timeout existed it is kept exactly. No outcome changes.

Red-first was checked by mutation, since the pins do not exist at base 3815ed6f. In a scratch worktree at fdfd4c1f, three mutations turned exactly those three pins red (3 failed, 160 passed across four modules): `forbid_background` removed on the clone in `test_converge_main_branch.py` and on the install clone in `test_kernel_update.py`, and `init_repo` reverted in `test_file_github.py`. At the head the twelve modules pass: 384 passed, 63 subtests, in a capped scope. CI at fdfd4c1f: all ten checks SUCCESS.

Review run by the maintainer automated review session; verified against head fdfd4c1f.

PR #1316, opened 2026-09-10
/file sandboxes an SVG: Content-Security-Policy sandbox on every image/svg+xml success, local and relayed

henrypinkard:
A tab navigated to `/file?path=x.svg` parses the SVG as a document and runs its inline script at the kernel's origin with the dashboard's session cookie. The own-tab opener (`openFileTab`, since the PDF-only gate came off in the file-viewer links change) hands that route any path on a Cmd/Ctrl or middle click, so a path link to an SVG inside a shown file (a cloned repository's README, say) is a script-execution hole. Found in the review of that change (2026-09-10).

## What changes

- `_media_policy_headers(mime)` in `kernel/kernel.py`: `Content-Security-Policy: sandbox` when the mime is `image/svg+xml`, nothing otherwise. A sandboxed document runs no script and gets an opaque origin, so it can reach nothing of the dashboard's.
- Sent on every SVG success shape of `/file` (HEAD, 206 Range, 200) and of the `/remote/<host>/file` relay, which rebuilds its interpretation headers from our mime rather than mirroring the remote's, so it has to restate the policy.
- Never gated on `_is_navigation`: harmless on a fetch or an `` load (no document is created, so the chat's thumbnails, the viewer's inline preview and the lightbox keep rendering), and closing the hole must not hinge on Sec-Fetch headers a plain-http dashboard never sends.
- On the 200 branch the header rides beside `_send`'s `frame-ancestors 'self'` as a second policy header, enforced additively; the framing policy is untouched. `nosniff` alone does not help here because the type is declared.

## Tests (red before the change)

- `tests/test_kernel_preview.py` `SvgSandboxPolicy`: a GET of a synthetic SVG carrying an inline script is `200 image/svg+xml` with the exact `sandbox` policy (and the framing policy still present); the HEAD probe and a `bytes=1-` 206 carry it; a `.png` (GET, HEAD, Range) and a `.md` text response carry no sandbox. 25 to 29 cases.
- `tests/test_kernel_remote_file_relay.py`: a remote `.svg` through the relay carries the relay's own sandbox policy on GET, HEAD and 206, and a remote PNG does not. 16 to 17 cases.
- Assertions check the exact header value, so a weakened `sandbox allow-scripts` would fail.

Tier: fix

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -40953,9 +40953,33 @@ _img_cache = {}                                  # "path:mtime:size" → dataURL
 #      the actual bytes over HTTP (behind _authorize, like everything else) instead of a data-URL round
 #      trip — the browser lazy-loads, caches, and renders a PDF natively in the lightbox iframe. The
 #      allowlist is RENDERABLE media only; anything else 404s and the client shows a plain link. SVG is
-#      served as an image (an <img> never runs its scripts); the files are the user's own, written by
-#      their own agents, on their own machine.
+#      served as an image (an <img> never runs its scripts; a tab NAVIGATED to one is a document, which
+#      _media_policy_headers below sandboxes); the files are the user's own, written by their own agents,
+#      on their own machine.
 _PREVIEW_MIME = dict(_IMG_MIME, **{".pdf": "application/pdf"})
+
+
+def _media_policy_headers(mime):
+    """The extra headers a /file SUCCESS carries for its media type: `Content-Security-Policy: sandbox`
+    on image/svg+xml, nothing on anything else.
+
+    An SVG is the one type on the allowlist that is ALSO a document. Served to an <img> it is a picture
+    and its scripts never run; but the own-tab opener (ui/webview/preview.ts openFileTab) hands this route
+    ANY path on a modified click since the PDF-only gate came off, and a tab NAVIGATED to /file?path=x.svg
+    parses it as a page and runs its inline <script> at the kernel's origin, with the dashboard's session
+    cookie attached (the 1204 review, 2026-09-10). nosniff is no help there: the type is declared, and
+    image/svg+xml is the scriptable one. `sandbox` closes it: a sandboxed document runs no script and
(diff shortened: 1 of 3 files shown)


ksarma:
Post-merge record: the owner merged this PR as ddbba4bb at head 404079fa before this review ran; no objection. It closes a real script-execution hole. At the base, a tab navigated to `/file?path=x.svg` ran the SVG's inline script at the kernel's origin with the session cookie. `openFileTab` hands that route any path on a modified click.

Reviewed inline: `_media_policy_headers` returns `Content-Security-Policy: sandbox` for `image/svg+xml` only. All six success shapes that serve a user file as SVG emit it. Those are HEAD, 206 and 200 in `_file_preview` and in the `/remote/<host>/file` relay. On the 200 branch `_send` writes `frame-ancestors 'self'` with `send_header` and appends the extra headers the same way. So the sandbox is a second policy header beside the framing one; the mutation run showed both in the response. The relay keys the sandbox on the mime it derives from the requested extension, so a remote PNG carries none. The only other `image/svg+xml` site is the `/media` static route. It serves romp's bundled glyphs through ``, `<link rel=icon>` and CSS, not user data. No `<object>` or `<embed>` exists in `ui/webview` or `kernel.py`, and the lightbox iframe is PDF-only. No rendered surface changes. The `fix` label fits. Privacy scan of the 177 added lines: clean.

Optional follow-up: the hardened list in `SECURITY.md` (Output sanitization, line 92) does not mention the `/file` SVG sandbox. One sentence there would record it.

Test evidence:
- Base b65ce0ec with the head's test files: 4 failed, 42 passed. The failures are the three SvgSandboxPolicy cases and the relay's `test_a_remote_svg_is_a_sandboxed_document_by_the_relays_own_policy`.
- Head: 46 passed (29 + 17, matching the body).
- Weakening the policy to `sandbox allow-scripts` fails the same 4.
- CI at 404079fa: all 12 checks green.

Review run by the maintainer automated review session; verified against head 404079fa.

PR #1318, opened 2026-09-10
Postal cards: the kind as coloured text, a delivery-state icon, both ends in their sessions' colours, no wash

henrypinkard:
## What

The postal cards in the chat pane, as the user ruled after seeing the old and new renderings side by side (2026-09-10), building on the gist-and-fold fix already on main. Kept: the envelope glyph, the FROM / TO words, the one-line gist with its fold.

1. **The kind is coloured TEXT, never a chip** (chips read as tags now): "Delegation" `#b08cff`, "Coordination" `#14b8a6`, "Question" `var(--st-working-bg)`, the colours the pre-rewrite chips wore, placed in the head's meta slot, which stays rigid so the word is never truncated.
2. **The delivery state is an icon at the head's right edge**, the way messaging apps show sent / delivered / read, each with a worded title and the clock ("Delivered 9:41 AM"). The states are the ones the kernel actually files: a sent card is stamped at send time (delivered | parked | nothing on error), and the postal ledger records, per message id, `exec` (the recipient's inbox drain consumed it: a real read event), `relayed` (the far host's end-to-end ack), `bounced` (+ why), `recall`. The kernel's postal index now folds those onto the message's record and the sent card carries them as `receipt` (with `remote` for a send across the peer bus). Map (`ui/webview/postal-state.ts`): one check = sent (handed to the relay), two dim checks = delivered, two accent checks = read, a clock = parked, a red mark = bounced (the title carries the reason), a return arrow = recalled; an incoming card in hand shows no icon, one that waited while the session was offline shows the clock. The word "delivered" left the meta text.
3. **Both ends wear their sessions' colours, and no card wears a wash**: the head reads FROM <peer chip, peer colour> TO <this session's chip, its own colour> for incoming, TO / FROM for sent; under 520 px this session's chip collapses to its coloured dot so both colours still show. The 6% peer-colour wash is gone (it read as this session's colour). Incoming messages keep their box (border + rail) even when their one line carries the whole message; sent ones stay slim lines.
4. **Amendment (the user, same day): a sent message that has not landed wears the pending send's own provisional dress.** A sent card in the sent or parked state carries the queued bubble's class (`queued-bubble`), and the bubble's dashed rule reaches it by a selector on the same declaration block, not a copy; it turns solid when the receipt says delivered, relayed or read; bounced keeps its red mark. The icon and the dress agree by construction (both read `deliveryOf`).

## Tests

- `ui/webview/postal-state.test.ts`: the pure rules (kind word, state per receipt, titles).
- `ui/webview/postal-card.test.ts`: pins for the kind text and colours, the icon-per-state map and its right-edge placement, the both-ends head and its narrow collapse, the no-wash rule, the boxed-incoming rule, the shared provisional class and rule.
- `tests/test_postal_receipts.py`: the index folds exec / unexec / relayed / bounced / recall; a sent card carries `receipt` and `remote`.
- `tests/test_postal_cards_served.py`: the real /chat page of a hermetic kernel over a synthetic chat (web with peers api and tests, nine cards, every kind and state) at 1000 px and 520 px: kind words and colours, icon per state with its title, both chips with their colours, no wash, boxed vs slim, the provisional dress on the two not-landed cards only, the narrow collapse. With `POSTAL_SHOTS` set it writes the screenshots.
- Two existing pins updated (the intent's meta-text pin; the notice-vocabulary retirement of the colours, which now return as text colours).

Design points: the two surfaces share one provisional vocabulary (the class, not a lookalike); the read state is real (the ledger's exec event), never inferred; the fold-through of ledger outcomes costs one pass over the same append-incremental log the index already reads.

Tier: feature

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -26389,17 +26389,49 @@ def _postal_index():
     if hit is not None and hit[0] == key:
         return hit[1]
     idx = {}
+    later = {}                                        # mid -> its outcome rows, in log order (folded after the scan)
     for o in _messages_rows(p):                       # append-incremental rows (2026-09-03): a send no
         if not isinstance(o, dict):                   # longer re-decodes the whole log on the active tab
             continue
-        if o.get("ev") == "sent" and o.get("id"):
-            idx[o["id"]] = {"id": o["id"], "from": o.get("from", "?"), "fromId": o.get("from_id", ""),
-                            # the sender's host as the log stamped it: "" for this kernel's own sessions,
-                            # a peer's name for relayed mail — and None when the row carries NO field, a
-                            # row from before the field existed, whose sender could be either (2026-09-06)
-                            "fromHost": o.get("from_host"),
-                            "toId": o.get("to_id", ""), "body": o.get("body", ""), "kind": o.get("kind", ""),
-                            "t": o["t"] if isinstance(o.get("t"), (int, float)) else 0, "park": bool(o.get("park"))}
+        ev, mid = o.get("ev"), o.get("id")
+        if ev == "sent" and mid:
+            idx[mid] = {"id": mid, "from": o.get("from", "?"), "fromId": o.get("from_id", ""),
+                        # the sender's host as the log stamped it: "" for this kernel's own sessions,
+                        # a peer's name for relayed mail — and None when the row carries NO field, a
+                        # row from before the field existed, whose sender could be either (2026-09-06)
+                        "fromHost": o.get("from_host"),
+                        "toId": o.get("to_id", ""), "body": o.get("body", ""), "kind": o.get("kind", ""),
+                        "t": o["t"] if isinstance(o.get("t"), (int, float)) else 0, "park": bool(o.get("park"))}
+            continue
+        # The message's LATER outcomes ride the same record (T302): the sent card's delivery icon reads them.
+        # Each is the ledger's own event, never inferred — `exec` is the recipient's inbox drain consuming
+        # the message (a REAL read), `unexec` a claimed-then-rolled-back drain (not read after all),
+        # `relayed` the far host's end-to-end ack, `bounced` a return (with the refusal's why), `recall`
+        # the sender unsending it. Collected here and folded after the scan, the postal service's own
+        # reader's shape (_sent_receipts): deliver() publishes the file before it appends the sent row, and
+        # a drain can log its exec in that instant, so an outcome may sit BEFORE its sent row. Row order
+        # within one id still decides (exec then unexec is not read).
+        if mid and ev in ("exec", "unexec", "relayed", "bounced", "recall") and isinstance(o.get("t"), (int, float)):
+            later.setdefault(mid, []).append(o)
+    for mid, rows in later.items():
+        rec = idx.get(mid)
+        if rec is None:                               # an outcome for a message this log never sent
+            continue
+        for o in rows:
+            ev = o.get("ev")
+            if ev == "exec":
+                rec["read"] = o["t"]
+            elif ev == "unexec":
+                rec.pop("read", None)
+            elif ev == "relayed":
+                rec["relayed"] = o["t"]
+            elif ev == "bounced":
+                rec["bounced"] = o["t"]
+                why = str(o.get("why") or "")
+                if why:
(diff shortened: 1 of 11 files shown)


ksarma:
Post-merge record: the owner merged this PR as 90c0ef74 at head 649f92c5 before this review ran; 3 items below. All three items are follow-ups.

Reviewed inline: the kernel's postal index collects the exec, unexec, relayed, bounced (with its reason) and recall rows per message id during the one scan it already makes. It folds them onto the sent record after the scan, so an outcome logged before its sent row still joins. The hydrator attaches the receipt only to a sent card whose send did not error. A refused send therefore cannot inherit its retry's outcomes through the body-keyed join. The read state is the ledger's exec event, written when the recipient's drain consumes the file. A cross-host read is logged under the sender's relay id; nothing is inferred. postal-state.ts has no side effects. It ranks bounced over recalled over read over relayed over parked over delivered. A remote send still awaiting its ack shows as sent. The card puts the kind as coloured text in the rigid meta slot. The delivery icon sits at the head's right edge with a textContent title (a peer-supplied bounce reason cannot inject markup). The peer chip and this session's own chip are in the head. A sent or parked card gets the queued bubble's own class rather than a copy of its rule. Against ui/CLAUDE.md: no new font size; the kind colours are theme tokens with light-theme values; the 520 px collapse keeps both colours on narrow layouts; the own chip is not a clickable dead end. The feature label fits: an additive receipt field and a new card surface inside the existing postal model, by the owner, so it merged on green. The privacy scan of the 962 added lines is clean, the fixtures use placeholder ids and the demo domain, and the body paraphrases the user.

Follow-ups:

- `ui/webview/styles.css:1888`. The provisional styling reaches only a slim sent card. A sent message with a fold (first line over the 90-character gist clip, or several lines) is boxed by the density rule at `render.ts:1576`. `render.ts:4795` adds queued-bubble anyway. The later `.notice` rule then wins border, background and radius; max-width 72% and opacity 0.85 survive. Measured in a browser against the head's stylesheet, the in-flight or parked card is 72% of the column wide, dimmed, with a solid border. It becomes full width when the receipt arrives. Fix: add `.notice.queued-bubble` to the shared block and to the width reset at `:1894`. Add one sent card over 90 characters in the sent or parked state to `postal-card.test.ts` or the served world (its six sent bodies are 34 to 64 characters).
- `tests/test_postal_cards_served.py:264`. The lab kernel's environment is a copy of the runner's minus three names. The `tests/README.md` rule says every hermetic kernel is built with `kernel_env` from `tests/test_ship_reship.py`. On a machine running romp the lab kernel inherits `ROMP_MANAGER_PID` (it exits if the live manager restarts mid-test) and `ROMP_SERVE_HOST`. `kernel_env` sets the same eleven names, so the fix is one import and one line. `tests/test_federation_missing_served.py:243` predates the rule with the same pattern and can go in the same change.
- `docs/guide.md:290`. The guide says the card shows the kind as a chip, draws a pill legend, and shows a 2026-07-23 screenshot. For the legend, `docs/stylesheets/extra.css:284` cites `.postal-service-intent`, which no longer exists in `styles.css`. The chip sentence was stale before this PR. This PR settles the kind as coloured text and adds the delivery icon and the both-ends head without touching docs. A docs-tier change: reword the sentence, restyle the legend, regenerate `postal-chat.png` (the served test writes screenshots when `POSTAL_SHOTS` is set).

Two nits. The comment at `ui/webview/styles.css:2320` still calls the peer chip the one coloured element in the notice vocabulary. The body counts two updated pins, which undercounts: `postal-head.test.ts`, `postal-intent.test.ts` and `romp-bubble.test.ts` all changed.

Locally at head 649f92c5: `tests/test_postal_receipts.py` 4 passed; `npm run typecheck` and `node esbuild.js --tests` exit 0; `node --test` over 21 compiled files on the postal render path 219 passed, 0 failed. Red-first was run, though tier 2 does not require it. At base ac44ee51 with the PR's test files checked out, the pytest module fails 4 of 4 (KeyError: 'read') and `postal-state.test.ts` fails to import. The changed pins fail exactly where changed (postal-card 5 of 5, postal-head 1 of 8, postal-intent 1 of 3, romp-bubble 1 of 5). `tests/test_postal_cards_served.py` was not run here: a served fixture boots a real kernel and a browser, and it skips on CI without a browser, so it was reviewed by reading. CI at head: all ten checks green, including Tier policy.

Review run by the maintainer automated review session; verified against head 649f92c5.

Issue #1317, opened 2026-09-10
Sessions that survive a kernel restart: a per-session host process the kernel re-attaches to


henrypinkard:
## Summary

The maintainer's goal is a deploy restart that does not cut a running turn and lets the session keep working unattended. Today that is impossible by construction: the session's Claude Code process is a child of the kernel, talks to it over two pipes, and the kernel's shutdown interrupts every turn and kills every Claude Code process (the `claude` CLI) within about two seconds (kernel/kernel.py:53638, kernel/sdk_backend.py:8377). The next kernel resumes the conversation from the transcript, but the turn that was running is gone.

Recommendation: build a per-session host process that owns the CLI over its stdio pipes and exposes a local Unix socket the kernel attaches to, and plug it into the SDK through the custom transport hook the SDK already exposes (`ClaudeSDKClient(transport=...)`, claude_agent_sdk/client.py:70-151). This is candidate (a) below. It is the only candidate that keeps the running turn alive across a kernel death, works the same on macOS and Linux, and can be reached in stages that each pay for themselves. Handing the live pipes across kernel generations (candidate b) has no macOS path and no manager path either, because the manager is a Node process and cannot pass pipe descriptors. The CLI's own background sessions (candidate c) keep a process alive but offer no machine protocol to re-attach the SDK, so they do not reach the goal on their own.

Before any host exists, three things must be true, or the host adds a new orphan class instead of removing one: an ownership lease every host writes and every kernel checks; the reaper and the interrupt escalation rewritten to work by session identity rather than by "direct child of this kernel"; and one owner named for every per-session file. Those three are useful on their own and are the first stages of the plan.

## What a restart loses today

What is lost, per restart:

- The running turn. The SDK client interrupts it and the Claude Code process (the `claude` CLI) exits; the model's in-flight generation and every tool call in progress die with the process. The transcript keeps what the CLI had flushed. The resumed CLI starts a new turn from the continuation notice; it does not continue the interrupted thought.
- Tool calls in flight. They run inside the CLI process; a Bash tool mid-command dies with it (and, on Linux, with the CLI's whole scope, which is what T276 made deliberate).
- Permission prompts and questions awaiting the maintainer. `can_use_tool` and the ask flow park inside the kernel (kernel/sdk_backend.py:6349, 7292); a pending ask is recorded in the registry as `pendingAsk` and re-raised as a notice on boot (kernel/sdk_backend.py:8035 onward), but the specific prompt is gone.
[...]

PR #1320, opened 2026-09-10
Pin what the Claude Code CLI does when its parent goes quiet (T303, stage 3 of the restart-surviving sessions program)

henrypinkard:
Part of #1317 (stage 3: the hermetic probe of the Claude Code process's control-request patience).

The session host of stage 4 will hold a CLI whose kernel is absent for a while. This change writes down what the CLI itself does in that situation, measured on Claude Code 2.1.257 against throwaway config directories (no live session, no romp module loaded), and pins the two facts a test can hold without a model call.

Measured (the full numbers and method are in the program's stage 3 report; the reference's "What survives a restart" section carries the summary):
- a permission request (`can_use_tool`) waits for its answer with no expiry within ten minutes, and the turn continues normally on a late answer;
- a hook callback waits 600 seconds by default, or the matcher's `timeout` seconds when one is set, then the CLI cancels the request (`control_cancel_request`), records a hook-timeout error as the tool's result and goes on with the turn;
- a second `initialize` on the same stdin is accepted and its hook table replaces the first;
- stdin end-of-file ends an idle CLI at once (0.02 s) and a busy one after its turn (a 30 s tool call ran to completion first);
- an unread stdout does not stall the CLI: the pipe's 64 kilobytes fill, the rest buffers inside the process, the turn completes;
- `--resume` takes no lock, and two processes on one session id both append to the one transcript;
- `claude --bg` runs an interactive session on a pseudo-terminal under a daemon that stays in the launcher's cgroup, and refuses `--print`, so a background session has no stream-json channel.

Test: `tests/test_cli_control_protocol_probe.py` pins the second-initialize acceptance, the prompt idle exit on stdin end-of-file, and the `--bg`/`--print` refusal. It runs the real CLI, so it follows the suite's standing rule that no test reaches the live CLI unasked: it runs only with `ROMP_CLI_PROBE_LIVE=1` and a `claude` on PATH (or `ROMP_CLI_PROBE_CLAUDE`), against a throwaway `CLAUDE_CONFIG_DIR` and `XDG_STATE_HOME` with an empty settings file and no key (so it costs no model call), every `CLAUDE_*` and `ROMP_*` variable scrubbed from the child's environment, the child killed with its process group in tearDown. It skips otherwise, on CI too.

Docs: one paragraph under "What survives a restart" in `docs/reference.md` naming the measured values and the CLI version.

Verified: the pin test passes live on the devbox (2 passed, 0.7 s) and skips by default; the docs-reading tests pass (167 passed, 2 skipped).

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -759,6 +759,26 @@ check whether each is still running before relaunching it. A kernel restart has
 never touched work a session deliberately detached: tmux servers, `setsid`
 children and other processes that outlive their shell.
 
+What the CLI itself does when its parent goes quiet was measured on Claude Code
+2.1.257 (2026-09-10, the restart-surviving sessions program's stage 3 probe, run
+against a throwaway config directory): a permission request (`can_use_tool`)
+waits for its answer with no expiry within ten minutes and the turn continues
+normally on a late answer; a hook callback waits 600 seconds by default, or the
+matcher's `timeout` seconds when one is set, then the CLI cancels the request
+(`control_cancel_request`), records a hook-timeout error as the tool's result
+and goes on with the turn; a second `initialize` on the same stdin is accepted
+and its hook table replaces the first; stdin end-of-file ends an idle CLI at
+once (0.02 s) and a busy one after its turn (a 30 s tool call ran to completion
+first); an unread stdout does not stall the CLI (the pipe's 64 kilobytes fill,
+the rest buffers inside the process, the turn completes); `--resume` takes no
+lock, and two processes on one session id both append to the one transcript;
+`claude --bg` runs an interactive session on a pseudo-terminal under a daemon
+that stays in the launcher's cgroup, and refuses `--print`, so a background
+session has no stream-json channel. `tests/test_cli_control_protocol_probe.py`
+re-checks the two facts that need no model call (the second initialize, the
+`--bg` refusal) when run with `ROMP_CLI_PROBE_LIVE=1` and a `claude` on PATH; it
+skips otherwise, as every test that would reach the live CLI must.
+
 A message the kernel cannot handle does not end the session's CLI. The kernel
 handles each streamed message on its own: when a handler raises, it logs the
 exception type and the failing frame (file, line and function, first on the line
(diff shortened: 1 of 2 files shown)


ksarma:
Post-merge record: the owner merged this PR as 89ebeb3d at head 66bedbd5 before this review ran; four items below. Three follow-ups and one nit.

Reviewed inline: the diff is a 187-line opt-in test module and a 20-line paragraph in `docs/reference.md`; no romp code changes. The test class skips unless `ROMP_CLI_PROBE_LIVE=1` is set and a `claude` is on PATH. No workflow sets either, so CI skips it, as the body and the docs say. The first test sends two initialize control requests and closes stdin, so it needs no credential and starts no turn. The group it kills in `tearDown` is its own child's. The second test is an argv probe the measured CLI refuses before any network call. The child environment drops `CLAUDECODE`, `CLAUDE_*`, `ROMP_*` and `XDG_*` and gets a throwaway config directory with an empty settings file. Every value in the docs paragraph matches the body, and the paragraph names the CLI version.

- `tests/test_cli_control_protocol_probe.py:45`, follow-up. The scrub keeps `ANTHROPIC_*`, so the child has no credential under pytest only because `tests/conftest.py` pops the auth variables. The module's own unittest entry point loads no conftest. The `--bg` probe at line 161 also passes a real prompt through `subprocess.run` with no process group of its own. A CLI that accepts the pair (the case the assertion exists to catch) would run a billed turn in a daemon that `tearDown` cannot kill. Fix: scrub `ANTHROPIC_`, drop the prompt, and give that child its own group.
- `docs/reference.md:766`, follow-up. The hook-timeout sentence covers every hook callback but was measured for a tool hook. The hook romp registers is `UserPromptSubmit`. On the installed 2.1.266 (the paragraph measures 2.1.257), that hook blocks the prompt when it times out instead of letting the turn go on. Fix: scope the sentence to tool hooks, and state the prompt-hook outcome or mark it unmeasured.
- Tier label, follow-up. `fix` means a bug fix with a test that fails before it. A docs paragraph plus a probe test that skips on CI is tier 0, `docs`. Both tiers merge on green for every author, so the gate was unchanged. Relabel for the record.
- `docs/reference.md:778`, nit. The paragraph says the test re-checks two facts. It asserts three, the idle exit on stdin end-of-file included.

Nothing from the PR ran here: the probe module drives the real CLI, so the standing rule applies. The evidence is a syntax parse of the new file (ok), a privacy scan of the 207 added lines and the PR text (clean), and CI at 66bedbd5. All twelve checks there are green (Python 3.10 to 3.13, bats, vscode-extension, gitleaks, both tier checks). The red-first bar does not apply: no romp code changed.

Review run by the maintainer automated review session; verified against head 66bedbd5.

PR #1321, opened 2026-09-10
The deep-link landing test pins the address the dashboard token scrub leaves; the tap itself was landing all along (T307)

henrypinkard:
The browser-backed deep-link landing test (`tests/test_notification_tap_resume_browser.py`, the link road) fails on a pristine `upstream/main`: after the page boots on `/?token=…&push-reveal=…&push-pid=…`, the address is the bare origin, and the test pinned that `token=` must stay. That pin is stale, not a regression: the same morning the test landed (mobile-tap-3, 09:47 UTC), the dashboard token scrub landed in parallel (dashboard-token-scrub-offer, 10:14 UTC, neither commit an ancestor of the other) and made the shell's head drop `?token=` the moment the page runs, once the response that served it has turned the token into the `romp_token` cookie every later request rides. Everything the test asserts about the tap itself passed on main before the token line: the chat pane's active tab became the linked session, exactly one `/reveal` via `link` with the boot flag, the ledger row settled, the push params stripped. The click and vanish scenarios pass on main as well.

What the user sees after a notification tap: the session opens and the address bar shows a clean origin with no token in it; the cookie keeps them signed in (the second, in-place arrival the same test drives rides that cookie and lands too). No login prompt, no dead landing.

The change: the driver reports the browser context's cookie names after boot; the test now pins the scrub's contract instead of the pre-scrub address: no `token=` in the address, the `romp_token` cookie present (a clean address WITHOUT the cookie would be a logout, which is the regression this line now guards), and nothing else left on the address. Red-first: the file was red on main for the stale line alone; with the re-pin all three scenarios pass (3 passed, 6.2 s, Chromium via playwright, hermetic kernel per class).

Why CI could not catch it: the Python jobs install no node dependencies and no browser, so every served-page test (ten files, `tests/*browser*.py` and `tests/*served*.py`) raises SkipTest there; the extension job does install playwright's Chromium, but only for the pane bench (`node --test tests/ui-bench.test.mjs`), and never runs the Python browser tests. Recommendation, for a separate change: run those ten in the extension job right after `npx playwright install chromium` (python3 and the built dist are already there), as one step with the same skip-to-failure stance the bench uses, so a runner missing the browser turns the job red rather than green with the coverage gone; they took about six seconds per file here, so the cost is minutes. This PR changes no CI.

Tier: fix.

### DIFF
--- a/tests/test_notification_tap_resume_browser.py
+++ b/tests/test_notification_tap_resume_browser.py
@@ -30,9 +30,10 @@ Three scenarios against the REAL shell, the REAL worker and the REAL kernel (her
   2. THE LINK ROAD (what an iOS tap produces): the page is navigated to the deep link the declarative message's
      `navigate` names — '/?push-reveal=<api>&push-pid=<pid>' — as iOS does on a tap. The page must land it at boot:
      ONE /reveal via 'link' with boot:true, parked by the kernel and consumed on the chat pane's ready (the tab
-     becomes `api`), /push/landed for the pid, the params stripped from the URL, the 'deeplink' row on file. Then the
-     open page GAINS the params without a load (history.pushState + pageshow, the window iOS navigates in place)
-     for a second push: landed live via 'link' (no boot flag), settled, stripped.
+     becomes `api`), /push/landed for the pid, the params stripped from the URL (and the ?token= it was opened
+     with, which the shell drops once the cookie is set), the 'deeplink' row on file. Then the open page GAINS the
+     params without a load (history.pushState + pageshow, the window iOS navigates in place) for a second push:
+     landed live via 'link' (no boot flag), settled, stripped.
   3. THE VANISH ROAD (what a LIVE iOS app leaves): three test pushes (`api`, `tests`, `docs`) filed at the kernel and
      acked shown by pid, as the phone's worker acks them; the page stubs getNotifications by data.pid and comes forward
      three times. Two of three displayed → exactly the missing one lands: ONE /reveal via 'vanish', the tab on `api`,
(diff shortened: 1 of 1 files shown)


ksarma:
Post-merge record: the owner merged this PR as 52e8ac61 at head fd0b50e6 while this advisory review was running; no objection. Nothing to fold in.

Reviewed inline: one test file, +16/-5, matching the body clause by clause. I traced the boot link through the shell at main. The head script drops token (kernel.py:49090). The reveal script drops the push params (kernel.py:48557). Each adds a `?` only for a non-empty query. The address therefore ends as the bare origin with a trailing slash, and the rstrip equality holds. A lingering `?` would fail it loudly. The cookie is real. A valid `?token=` makes `_authorize` return the token, and `_send` sets `romp_token` with Path=/, SameSite=Lax, HttpOnly on the first response. Playwright's `context.cookies(url)` is the context API and returns HttpOnly cookies, so the value equality against the hermetic token holds. In DRIVER_LINK, cfg and context are bound before the new line. The old assertIn fails at the base on that line alone, since the scrub merge is an ancestor of the base. The three replacements pass by the same trace. Neither the pin's branch tip nor the scrub commit is an ancestor of the other: #1297 merged 12:12Z, #1219 17:05Z. Privacy scan of the 16 added lines: zero hits. The logged cookie value is the hermetic test token, and the sids are synthetic. Label fix fits: tests only, a stale pin corrected.

Findings:
- Nit, follow-up: `tests/test_notification_tap_resume_browser.py:333` writes a token key into cfg.json that no driver reads. The Python side compares against `self.token`. Harmless dead config.
- The CI suggestion in the body is sound as a separate change. The extension job installs no cryptography, which this module's push helper needs. Turning the skip into a failure would go red until that is added.

Test evidence: read only. The module is a served browser fixture (hermetic kernel plus headless Chromium) and never runs on this box. CI skips it too: the Python jobs install no node deps or browser, so setUpClass raises SkipTest. CI green therefore does not execute the changed test. Execution evidence is the owner's stated local run (3 passed, 6.2 s) and the CI-run TokenLeavesTheUrl pin in `tests/test_kernel_auth_hardening.py`. py_compile of the test file passed here. CI at head: all ten checks SUCCESS.

Review run by the maintainer automated review session; verified against head fd0b50e6.

PR #1324, opened 2026-09-10
Gear: the judges' fast-mode box says Fast mode, sits on the Triage model row, and greys out while no judge tier is on Opus

henrypinkard:
Follow-up to the Fast judging setting that merged this morning, on the maintainer's call (2026-09-10): keep the feature, change how it presents.

## What changes

- **The words.** The box says **Fast mode**, the term the chat's statusline badge and its tip already use for the same CLI feature, instead of "Fast judging". The stood-down-gesture toast calls it "Fast mode (judges)" so it cannot be confused with "Fast comment threads".
- **The place.** The box rides the **Triage model row**, after the picker, as a compact checkbox and word at the row's size; the standalone row and its paragraph are gone. The help shrinks to the row-hint style its neighbours use, and hovering shows one description at a time (the box's own over the box, the row's elsewhere).
- **The inert state is visible.** With the shipped defaults (triage on Sonnet, indexing on Haiku, the rest following triage) the box did nothing and the gear did not say so, a finding in the review of the original change. Now, while no judge tier is on Opus, the box is greyed and its hint says why and what to pick. It keeps the stored value rather than unchecking, so pinning Opus later revives it without a second click.

## What does not change

The element ids, the `setJudgeFast` op and its gesture stamp, the `judgeFast` fill, the kernel's stored `judge-fast` value and its propagation to linked kernels. The edits to `kernel/kernel.py`, `kernel/judge.py` and `ui/webview/federation.ts` are comments naming the new label.

## Tests

`ui/webview/gear.test.ts`: the renamed setting test, and a new test pinning the row placement, the words, the hint content, the gate wiring, the mixed-mark scoping and the hover and wrap rules (19 tests). `tests/test_kernel_settings_sections.py`: the old label-row test replaced by one that finds the box on the Triage model row. Both green here, along with the gear select matrix test. Docs (`docs/reference.md`, `docs/judges.md`) renamed to match.

Tier: feature

### DIFF
--- a/docs/judges.md
+++ b/docs/judges.md
@@ -459,8 +459,9 @@ permission/API-error floors: one interrupt at a time, the present event first.
 
 - Toggles: `CLOSER_ON`, `GROUPER_ON`, `DISTILLER_ON`, `CONSOLIDATE_ON`.
   Models: `STATE/judge-model` (triage), `STATE/index-model`.
-  Fast judging: `STATE/judge-fast` (`on` | `off`, off by default; read per
-  call, and the fast-mode opt-in rides only a call whose model is Opus).
+  Fast mode for the judges (the gear's Fast mode box beside the Triage model
+  picker): `STATE/judge-fast` (`on` | `off`, off by default; read per call, and
+  the fast-mode opt-in rides only a call whose model is Opus).
   Pool width: `STATE/judge-concurrency` (the gear's Judge concurrency, 1..16,
   read fresh each pass; empty = `ROMP_JUDGE_CONCURRENCY` as read at load,
   else 6). Every pool reads it at call time (`_conc`, or `_judge_concurrency()`
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -485,18 +485,21 @@ through to `install.sh`:
   and follows to every connected machine like the other judge settings; its
   Default option clears the setting back to the variable, else 6.
 
-### Fast judging
+### Fast mode for the judges
 
-- **Fast judging** (the gear's Judges section; off by default) runs the judges
-  in Claude Code's fast mode, an Opus-only research preview billed at a premium
-  over standard Opus rates. The setting is read per call: a judge call whose
-  model is Opus, by the bare alias or a pinned Opus version, carries the CLI's
-  fast-mode opt-in in its per-call settings; a call on any other model runs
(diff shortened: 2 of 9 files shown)


ksarma:
Post-merge record: the owner merged this PR as f5d00733 at head 91b7a28f before this review ran. The review found 2 items, listed below.

Reviewed inline: the rename to Fast mode, the move onto the Triage model row after the picker, and the help reduced to a hover hint match the body. The element ids, the `setJudgeFast` op and stamp, the `judgeFast` fill and the stored judge-fast value are unchanged. The `kernel.py`, `judge.py` and `federation.ts` edits are comment-only. `judgeFastGate` reads the three model selects and resolves Follow triage to the triage pick, as `_distill_model()` does. Its opus substring test agrees with `_model_family_version` for every value the pickers offer. An unloaded tier counts as unknown. The gate re-runs from `fill()` after the tiers are written and from the three pickers' change handlers. The version facade dispatches a real change event, so a facade pick re-evaluates without a reload. `fillMixedMarks` scopes to the control's own label first, so the box's mixed mark is its own. The CSS adds no font-size and no hex. The one-hint-at-a-time `:has()` rules resolve by specificity, and a headless Chromium probe confirmed each hover shows one hint. Docs match, the feature label fits, and the privacy scan of the 135 added lines is clean.

- Follow-up, `ui/webview/gear.css:86`: dim only the input and the word. `opacity .4` on `.rs-fastin.rs-off` applies to the label that contains the hint span. So in the shipped default state (no tier on Opus), the hint that explains the greying renders at about 40%. The rows beneath show through its translucent background. Headless Chromium measured the brightest hint pixel at 210 live against 92 greyed. The pins at `ui/webview/gear.test.ts:91` and `tests/test_kernel_settings_sections.py:129` change with it.
- Nit: the retired name Fast judging remains in comments and assertion messages in six other test files. Examples: `tests/test_kernel_judge_model.py:180` and `tests/test_setting_gesture_order.py:798`.

Locally at head: `tests/test_kernel_settings_sections.py` 10 passed, typecheck clean, and 164 compiled gear-related UI tests pass across 17 files. The PR's two test files fail at the merge base (1 of 10, 2 of 19). CI at head was green.

Review run by the maintainer automated review session; verified against head 91b7a28f.

PR #1325, opened 2026-09-10
Pane shim: declare a redial (reconnect=1) only once the bundle's ready has left on a socket

ksarma:
## Symptom

A pane can declare a redial (`&reconnect=1`) for sessions it never held. Two shapes reach it: the first socket opened and died before the bundle evaluated (the shim dials during HTML parse, while the bundle can still be downloading), or the bundle said ready while the socket was down. Every later dial then carries the flag, and the kernel serves a client that holds nothing as one that held every session: a skeleton strip, the active tab in full and one status frame per tab, sent into a document that has no message listener yet. Today the bundle's `ready` pops the skeleton set and the reconnect flag unconditionally (`_client_reset_chat_base`), so the screen recovers. The cost is that skeleton strip and those status frames, all redone whole at ready, and a false statement on the wire.

## Cause

The dial term was `everConnected?"&reconnect=1":""`. `everConnected` is set at the first `onopen`, before any bundle has evaluated, and `send()` kept no record of the bundle's `ready`: a ready handed to it while the socket was down queued and flushed onto the redial socket like any other message.

## Fix

Two bits beside `everConnected` in the shim core (`_shim` in kernel/kernel.py):
- `bundleReady`: the bundle's own `{type:"ready"}` has passed through `send()` on this page (set at the top of `send()`).
- `readyQueued`: that ready is waiting in `queue` for an open (set in `send()` when the socket is not open; cleared in `onopen` right after the flush loop, which carried it).

The dial term is now `(everConnected&&bundleReady&&!readyQueued)?"&reconnect=1":""`. A socket that died before the bundle said ready, or while its ready was queued, redials as a fresh page and gets everything whole; the queued ready goes out exactly once on the redial socket, by the flush. No new template placeholder, so the format args and `_shim_core_js` are untouched, and the `onopen` text after the flush (`queue=[];queuedDiag=0;` followed by the reload core's `ended()`) is unchanged, so the pin in tests/test_dashboard_auto_reload.py holds. The redial gloss, the accept-time comment, `ui/webview/skeleton-tabs.ts` and `docs/read-side.md` are reworded to the new condition.

One case stays open, and the accept-time note says so: a ready that left on an open socket the kernel never processed, the socket dying before any frame came back, still redials with the term. No shim bit sees it, and the page is served skeletons that fill on click or the idle prefetch.

## Tests

`tests/test_pane_shim_return.py` `ReconnectFlag` runs the real shim core in node (`_run`) and sends the bundle's ready through `window.__rompLocalSend`, the shim's alias of `send` (the harness's `__rompFed` fake has only `inbound`).
- `test_a_ready_that_queued_while_the_socket_was_down_redials_without_the_flag_and_goes_out_once_on_the_redial` (new): open, keepalive, drop; the ready is sent while the socket is down and nothing goes out on the dead socket; the redial URL has no `reconnect` and ends with `&active=S1`; the redial's flush carries exactly one ready; after a frame on that socket the next redial ends with `&active=S1&reconnect=1`. Before the change it fails at the first URL assertion: the URL ends with `&active=S1&reconnect=1`. It is the one case that fails on an `everConnected&&bundleReady` term, on a `send()` that never sets `readyQueued`, and on an `onopen` that never clears it, so it pins the third bit at both ends.
- `test_a_first_socket_that_died_before_the_bundles_ready_redials_without_the_flag` (new): two sockets die before any ready, both redials carry no flag and keep the active hint; the ready goes out on the third socket; the fourth dial ends with `&active=S1&reconnect=1`. Before the change it fails at `midLoad`: the URL carries `&reconnect=1`.
- `test_the_first_dial_has_no_flag_and_a_redial_after_the_bundles_ready_left_on_a_socket_carries_it_after_iid_and_active` (renamed from `..._and_every_redial_carries_it_...`): now sends the ready on the first socket before the drop. The case as previously written (no ready before the drop) fails under the gate, as the new name says it should: the second URL ends with `&active=S1` and carries no flag. All three cases fail on a `send()` that never sets `bundleReady`.
- `tests/test_chat_skeleton_reconnect.py`: the exact-string pin follows the new term; the module docstring says a redial after the bundle's ready declares itself.

Red run on the unchanged shim, `ReconnectFlag` alone: 2 failed, 1 passed. Green: 84 passed across `tests/test_pane_shim_return.py`, `tests/test_chat_skeleton_reconnect.py` and `tests/test_dashboard_auto_reload.py`; `npm run typecheck` clean (skeleton-tabs.ts changes a comment only). Full pytest suite on this tree, served-page legs included: 1 failed, 11390 passed, 41 skipped, 1744 subtests passed. The one failure, `tests/test_notification_tap_resume_browser.py::ServedTapLanding::test_a_navigation_to_the_deep_link_lands_at_boot_and_again_when_the_open_page_gains_it` (`'token=' not found in 'http://127.0.0.1:<port>/'`), fails the same way alone with the unchanged kernel/kernel.py swapped in, so it is not attributed to this change; it is a playwright-driven served-page test that continuous integration skips (its module docstring says so).

Tier: fix


ksarma:

Reviewed inline: `send()` sets `bundleReady` before the `readyState` check and `readyQueued` only on the queue branch. A ready on an open socket therefore leaves `readyQueued` false. A ready while connecting or closed queues with the bit set. `onopen` clears it after the flush loop and before `queue=[]`, so the auto-reload pin's text is intact. The dial term is built inside `connect()` at dial time. The exact-string pin in `test_chat_skeleton_reconnect.py` matches it byte for byte. Every shim-served pane bundle posts exactly `{type:'ready'}` through the shim's send alias, directly or via federation's local route. The shell page's own socket never used the flag. The kernel side is untouched: the ready handler still pops the skeleton set, and the `/ws` reconnect read is unchanged apart from reworded comments. The implicit behaviour change is correct. A bundle that never says ready is now served whole on redial rather than as skeletons, which is the accurate outcome. The stated residual reproduces by probe and is scoped correctly. Privacy and vocabulary scans of the added lines are clean. The `skeleton-tabs.ts` change is comment only.

Follow-ups, neither for this PR:
- `tests/test_pane_shim_return.py:597`: no direct case covers a ready posted while the first socket is still connecting. A probe shows it queues, flushes once, and the next redial carries the flag. A pinned case would guard that path directly.
- `kernel/kernel.py:45273`: the kernel answers every processed ready with a caps frame. A shim bit latched on that frame would close the stated residual on an exact event.

Red-first at base 54baf25e with the head's test files: ReconnectFlag 2 failed, 1 passed. The first-socket-died case failed at midLoad, the queued-ready case at its first URL assertion, and the renamed case stayed green. Elsewhere only the exact-string pin was red. At head the three modules ran 84 passed, all four mutations failed where stated, and `npm run typecheck` is clean. CI is 12 of 12 green at head.

Review run by the maintainer automated review session; verified against head be11455c.

ksarma:
Head note: the branch was updated with main, so the head moved from be11455c to 104c157b. The new head is a mechanical merge of be11455c and 90c0ef74: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 104c157b.

PR #1328, opened 2026-09-10
Chat: the mention chip wears the awaiting chip's dress, the bare name in its session colour on a dark backing

henrypinkard:
Follow-up to the composer @-mention picker that merged earlier today, on the maintainer's call (2026-09-10): the chip a sent @name becomes should look like the Awaiting box's session chip, not a colour-filled pill, and should read just the name.

## What changes

- **The look.** The mention chip drops its identity-colour fill, ring and 9px radius and takes the awaiting chip's dress: the 85% black backing, 7px radius and 0 5px padding, declared **once** in a rule shared by `.chip-peer-name` and `.mention-chip` so the two cannot drift. The mention chip's own rule sets only what is its own (display, line height, weight, and the session's identity colour as the text colour), none of the shared properties, so cascade order does not matter.
- **The text.** The chip shows the bare session name; the `@` is gone from the chip. The typed token rides a `data-token` attribute, and what is sent is untouched: the composer still inserts `@name `, the segmenter and the postal address rules are unchanged. A remote session's `host:` prefix wears the same quiet prefix treatment the awaiting chip gives it. The closed-session state keeps its last colour and its hover title.
- **Docs.** The guide's paragraph on naming another session describes the chip's new look.

## Tests

- `ui/webview/awaiting-peer-name.test.ts`: its pin on the awaiting chip's rule follows the shared rule (this was the one existing test the change broke before review caught it).
- `ui/webview/composer-mention-pane.test.ts`: the chip cases read the bare name and the token, cover a remote session's prefix, and pin the shared stylesheet rule and that the chip's own rule sets none of its properties.
- The pure segmenter tests are unchanged since what is sent is unchanged. Typecheck clean; the three touched files pass (32 tests) under the capped wrapper, browser cases executed.

Render of the two chips side by side, from a synthetic fixture, is at the reviewer's disposal on the maintainer's machine.

Tier: feature

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -65,7 +65,9 @@ its peers; if the bare name is ambiguous there, the session's mail tools refuse 
 list the candidates as `host:name`, and the session picks one. **Escape** closes the list
 without inserting, and it stays closed for that `@` until you delete it: more letters, or a
 caret move away and back, do not reopen it. In the sent message, a name that matches a live
-session is shown as a chip in that session's color.
+session is shown as a chip: the name without its `@`, in that session's color on a dark
+backing, the way the Awaiting chip names the session it waits on. Hover it for how that
+session is doing; the message itself still carries the `@name` you typed.
 
 **A message that has not gone yet.** Send to a busy session and your message waits as a
 dashed bubble under an hourglass until the session takes it — while it compacts, while a
--- a/ui/webview/awaiting-peer-name.test.ts
+++ b/ui/webview/awaiting-peer-name.test.ts
@@ -47,8 +47,11 @@ test("the chat pane names the peer: box in identity colour, pill with the colour
   assert.match(chip, /nm\.style\.color = chipPeers\[0\]\.color\.bg/);
   assert.ok(!/nm\.textContent = .*host/.test(chip), "no one-string concatenation of host and name");
   assert.match(chip, /chipPeers\.length \+ " peers"/, "several peers keep the one-line rule as a count");
-  assert.match(CSS, /\.chip-peer-name \{ background: rgba\(0, 0, 0, 0\.85\); color: #fff; border-radius: 7px;/,
+  // the backing is ONE rule shared with the transcript's mention chip since 2026-09-10 (the user, who wanted a
+  // typed @name to wear this chip's look); the peer name keeps its own #fff default beside it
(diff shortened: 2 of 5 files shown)


ksarma:
Post-merge record: the owner merged this PR as 634563b2 at head 6c50907a before this review ran; 3 items below. All three are follow-ups.

Reviewed inline: the shared rule at `styles.css:1373` is one declaration carrying the backing, radius and padding for both chips. The mention chip's own rule at `:1240` sets only display, line height, weight and the identity colour as text. It sets none of the shared properties, so cascade order does not matter. The chip text goes through `hostNameNodes`, so a remote `host:` prefix gets the `.host-prefix` class. Its textContent stays the full `host:name` that the roster re-key reads. The typed token is set by a dataset property assignment, never innerHTML. Nothing in `render.ts` reads it. What is sent is unchanged: `composer-mention.ts` and its tests are byte-identical to the base. The edit-and-resend path reads `ev.md`, not the bubble. No new font-size, no menu rule, no click handler. The docs sentence matches the code. The body paraphrases the request and quotes no one. The feature label fits: a self-contained visual change by the owner, merged on green. Privacy scan of the 78 added lines: no private strings, hostname, home path or banned vocabulary.

- `ui/webview/styles.css:1240`, follow-up: the swatch as text on the 85% black backing puts five dark swatches in the hawaiiS and romaO palettes (`kernel/palette.py`) at 2.1 to 2.8:1. The old fill with the palette's white foreground read at 6.9:1 or better. The romp, phase and pastel sets all clear 3.0. The awaiting chip and the tab labels already paint those swatches this way, so the fix belongs to the shared pattern. Two options: a lightness floor in the shared rule, or a re-curation of those two palettes' dark swatches.
- `ui/webview/styles.css:3225`, follow-up: in the light theme `.host-prefix` takes `--dim` (#5D574E), about 2.7:1 on the shared backing. A remote chip's `host:` prefix is therefore hard to read. This is inherited from the awaiting chip; a scoped override for both chips would fix it.
- `ui/webview/render.ts:16401`, follow-up: selecting and copying a sent bubble now yields the bare name, because the @ is present only on `data-token`. A pasted resend therefore names no session. Accept it, or substitute `data-token` in a copy listener.

Locally at head 6c50907a: typecheck clean. The 24 compiled UI tests referencing the mention code pass 318/318 with 0 skipped, so the Chromium cases CI skips ran here. The three files the body counts pass 32/32 (4+14+14). At the base 90c0ef74 with the PR's test files checked out, exactly the three changed or added tests fail. CI at head: every required check green, including Tier policy and the secret scan.

Review run by the maintainer automated review session; verified against head 6c50907a.

PR #1329, opened 2026-09-10
Tests: the bats git floor forbids background maintenance in fixture repos and their bare remotes, so a teardown's rm -rf never races a detached git child

ksarma:
## Symptom

A bats suite that commits, merges, fetches or pushes into a repo its teardown removes can fail on that teardown: `rm -rf "$TEST_DIR"` exits 1 on a `.git` it could not empty, and bats fails the test whose teardown failed. Six suites write into such repos: install-sh, release-sh, gitleaks-config, docs-serve, pre-push-hook and bootstrap-sh (docs-serve's teardown returns 0 after the rm, so there the race leaves a directory behind instead of failing). The same race showed on the python side as `Directory not empty: '.git'` from rmtree (#1312).

## Cause

`git commit`, fetch and merge spawn `git maintenance run --auto`, and receive-pack runs the receiving repository's auto gc or maintenance after every push. On git 2.47 and later the maintenance child takes `.git/maintenance.lock` and detaches from its parent, so it can still be writing under `.git` after the test's git call has returned. CI's runners have git 2.55.0. #1314 and #1315 forbid that work for the python fixtures per call (`tests/git_fixture.py`, `GIT_NO_BACKGROUND`); the bats floor in `tests/git-hermetic.bash` exported no maintenance or gc key, and `tests/pre-push-hook.bats` did not load the floor at all.

## Fix

- `tests/git-hermetic.bash`: `git_hermetic` carries the five `GIT_NO_BACKGROUND` keys (maintenance.auto=false, maintenance.autoDetach=false, gc.auto=0, gc.autoDetach=false, core.fsmonitor=false; the same order as `tests/git_fixture.py`) two ways, because neither alone reaches every git a test starts:
  - as `GIT_CONFIG_COUNT=5` runtime pairs in the environment (`GIT_CONFIG_KEY_n` / `GIT_CONFIG_VALUE_n`; git 2.31 and later, inside the floor's existing 2.32 requirement), which every git the test, a script under test (install.sh, release.sh, bootstrap.sh, docs-serve.sh, the pre-push hook) or a hook runs inherits; they override every config file and yield to any `-c` a test passes. A test that needs pairs of its own re-exports the whole set, since a count naming a missing pair is a fatal error for every git call.
  - in a global config file of the floor's own under `BATS_TEST_TMPDIR`, which `GIT_CONFIG_GLOBAL` names in place of `/dev/null`. A push over a local path starts receive-pack with `GIT_CONFIG_COUNT` stripped (the same rule keeps a `-c` from crossing into the other repository), so the pairs never reach the receiving side; `GIT_CONFIG_GLOBAL` survives, and the file is what receive-pack in a bare fixture remote reads. On a git whose receive-pack still spawns `gc --auto` directly, that child reads gc.auto=0 and exits before taking any lock.
  The header comment says all of this. Because the floor writes a file, `git_hermetic` runs inside a test (setup, body or teardown), not at file load or in setup_file; every current caller is in setup or a test body.
- `tests/pre-push-hook.bats`: loads git-hermetic and calls `git_hermetic` first in setup, in place of the empty `GIT_CONFIG_GLOBAL` file it kept for itself: nothing read that file, and keeping it would hide the floor's file from this suite's own bare remote. No test in the file asserts on author or committer and the hook reads no identity field, so the floor's exported identity changes no outcome; the setup comment says the repo-local `user.*` lines give the repo a configured user for anything that reads one.
- `tests/README.md`: the hermetic paragraph names both mechanisms and points at `tests/git_fixture.py`.

No production code changes; no assertion in any existing test changes.

## Tests

`tests/git-hermetic.bats` gains three cases behind its existing 2.32 skip, and its setup unsets `GIT_CONFIG_COUNT` alongside `GIT_CONFIG_GLOBAL` so the pins measure `git_hermetic`, not the caller's shell.

- "with git_hermetic the no-background keys read back from the environment and the floor's global file, not from the repo": each of the five keys reads back through `git config --show-scope --get-all` once at the `global` scope and once at the `command` scope; `GIT_CONFIG_COUNT` is 5 and the floor's file lists exactly five keys; `config --local --get maintenance.auto` exits nonzero with no output; a repo-local `maintenance.auto=true` still reads back false (the pairs beat a file value) and `-c maintenance.auto=true` reads back true (`-c` beats the pairs). Before the change it fails at the first read (git exits 1, nothing printed). This is the deterministic pin.
- "with git_hermetic a push's receive-pack reads the no-background keys in the receiving repository": commits, inits a bare repo, and pushes into it through a `--receive-pack` stand-in that records `maintenance.auto` and `gc.auto` as receive-pack's own environment sees them (untraced) and then runs the real receive-pack; asserts the push landed, both reads say `false` and `0`, the trace names `built-in: git receive-pack` and carries no `maintenance` line. Before the change both reads exit 1 with nothing printed; with the pairs alone (no file) they still do, which is what this pin exists to catch. The trace part is a smoke check: a git whose receive-pack spawns `gc --auto` directly shows that child either way.
- "with git_hermetic a commit spawns no maintenance or gc child": a commit under `GIT_TRACE` writes a trace containing `built-in: git commit` and no line containing `maintenance` or `run_command: git gc` (the substrings `tests/test_git_fixture.py` reads; a bare `maintenance` also covers the `--detach` argument recent git appends). Before the change it fails on the `maintenance` grep, the trace carrying `run_command: git maintenance run --auto --quiet`. A smoke check on the git running the suite; the config pin is the red-first that does not depend on what this git spawns.

Runs on git 2.43.0, bats 1.10.0, stdin from /dev/null: git-hermetic 8/8, pre-push-hook 22/22, install-sh 44/44, release-sh 29/29, gitleaks-config 9/9 (gitleaks installed, nothing skipped), docs-serve 6/6, bootstrap-sh 13/13 (its origin is a local fixture repo, no network, nothing skipped). Removing the file half of the floor turns the config and push pins red; removing the pairs turns the config pin red while the file still covers the commit and the push. No python file changes, so no pytest run.

Tier: docs

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -124,8 +124,15 @@ stderr: `[tests] not removed at run end: <path>`, instead of the run ending
 green over it. The same conftest gives git no global or system config
 (`GIT_CONFIG_GLOBAL`, `GIT_CONFIG_NOSYSTEM`) and a synthetic identity through
 `GIT_AUTHOR_*` / `GIT_COMMITTER_*`; bats suites that run git get the same from
-`load git-hermetic` + `git_hermetic` in `setup`. A fixture must not depend on
-the developer's git configuration (CI has none), and the env identity outranks
+`load git-hermetic` + `git_hermetic` in `setup`, with a global config file of
+the floor's own in place of none. That floor also forbids background git work:
+the five no-background keys of `tests/git_fixture.py` (below) ride the
+environment as `GIT_CONFIG_COUNT` pairs, inherited by every git a test, a script
+under test or a hook runs, and sit in the floor's global file, which receive-pack
+in a bare fixture remote reads (a push over a local path starts it without the
+pairs); so no detached `git maintenance` writes into a fixture repo while its
+teardown removes it. A fixture must not depend on the developer's git
+configuration (CI has none), and the env identity outranks
 `git config user.*` and `-c user.*` — a test that must pin a particular author
 exports its own `GIT_AUTHOR_*` after the floor. `tests/test_tempdir_hygiene.py`
 and `tests/git-hermetic.bats` pin all of it.
--- a/tests/git-hermetic.bash
+++ b/tests/git-hermetic.bash
@@ -1,7 +1,9 @@
-# tests/git-hermetic.bash: git in a bats test reads none of the developer's configuration.
+# tests/git-hermetic.bash: git in a bats test reads none of the developer's configuration and
+# spawns no background work.
 #
 # Use from a bats file: `load git-hermetic` at the top and `git_hermetic` as the first line of
-# setup(), before any git command runs.
+# setup(), before any git command runs. It writes a config file under BATS_TEST_TMPDIR, so it runs
+# inside a test (setup, the test body or teardown), not at file load and not in setup_file.
 #
 # Why (2026-09-06): the fixture repos here are built with `git init` + `git commit` in temp dirs,
 # and those commands honoured the developer's global git config. On one box a global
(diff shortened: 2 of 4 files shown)


ksarma:
Merging as shipped. One follow-up to fold later.

Reviewed inline: the floor now carries the five no-background keys two ways. They ship as GIT_CONFIG_COUNT=5 runtime pairs and in a global config file under BATS_TEST_TMPDIR, which GIT_CONFIG_GLOBAL names. On git 2.43 here, the pairs report as the command scope and sit between the repo config and -c. A push over a local path strips them from receive-pack: the push trace's spawn line starts with `unset GIT_CONFIG_COUNT`. GIT_CONFIG_GLOBAL survives, and the stand-in's reads inside receive-pack return false and 0 from the file. The `${BATS_TEST_TMPDIR:?}` guard fails loudly in setup_file and at file load. All seven callers run git_hermetic in setup or a test body. No consumer re-exports GIT_CONFIG_GLOBAL afterwards. No script under test sets GIT_CONFIG_* or runs git under env -i. The pre-push hook reads no identity field. Red-first holds with scratch floors. The base floor reds all three new cases at the stated lines; pairs only reds the config and push pins; file only reds the config pin. Tests only, so the docs label fits; body and commit message match the diff.

Finding:

- `tests/git-hermetic.bats:147` (follow-up; :122 for consistency): the bare `grep 'maintenance'` matches the fixture paths the push trace carries on six lines. So a TMPDIR whose path contains the word reds the pin on a green tree. Reproduced: 7 passed, test 8 failed at line 148, no maintenance child spawned. `grep 'run_command: git maintenance'` matches the spawn line, including the 2.47+ `--detach` form, and no path. #1327 is fixing the same shape on the python side.

Locally at head (git 2.43.0, bats 1.10.0, stdin closed): git-hermetic 8/8, pre-push-hook 22/22, install-sh 44/44, release-sh 29/29, gitleaks-config 9/9, docs-serve 6/6, bootstrap-sh 13/13. CI: all 12 checks green at d8c82d84; Tier policy reports the docs gate.

Review run by the maintainer automated review session; verified against head d8c82d84.

PR #1335, opened 2026-09-10
Dead-owner sweep of the suite's temp roots: conftest marks each root with its pid, boot reconcile removes the ones whose owner died

henrypinkard:
A test run that dies without its run-end cleanup leaves its whole `romp-tests-*` temp root behind; the kernel's boot reconcile now removes such roots when their owning pid is dead.

## Tier

Tier: fix

## Why

`tests/conftest.py` mints one private `romp-tests-*` root per run under the system temp dir and removes it at run end (PR #944 and follow-ups). A run that dies without reaching that removal (pytest-timeout's `os._exit`, a kernel restart cutting the tool shell, the cut-turn reaper's kill) leaves the root standing with every worker root and child scratch inside it. On the shared devbox those roots, together with sessions' scratch, piled into millions of small files, and the next boot's `/tmp` cleanup spent 39 minutes deleting them while ssh, Tailscale and the romp manager all waited behind it (2026-09-10).

## What

- `tests/conftest.py`: writes an owner marker (`romp-tests-owner.json`: the run's pid and start time) into the root and into the package state dir at mint time.
- `kernel/sdk_backend.py`: `sweep_dead_test_roots(tmpdir)` removes `romp-tests-*` directories whose marker names a dead pid. A live pid stays; a root with no marker or an unreadable one stays (refusing is the safe direction: the sweep cannot tell a foreign directory or a pre-marker root from a leak). `_boot_reconcile` calls it beside the orphan-CLI reap and the leftover-scope sweep and reports the count in its summary line. Under the suite the kernel's tmpdir is the run's own root (TMPDIR is redirected), so a test kernel never reaches the real system temp dir.

Not in scope: sessions' worktrees and scratch parked under `/tmp` by hand (a fleet-brief rule, not code) and the pr-watch skill's own scratch sweep (its skill file).

## Tests

`tests/test_test_root_sweep.py` (new, 9 tests): a dead-owner root goes and is counted; a live-owner root, an unmarked root, an unreadable marker, a marker without a pid, a foreign prefix and a symlink all stay; the running suite's own root carries a marker naming this process and survives a sweep of its parent; the marker name agrees between conftest and the kernel; `_boot_reconcile` calls the sweep. Everything is built under the test's own temp dir, never the real system temp dir.

Ran `python3 -m pytest tests/test_test_root_sweep.py tests/test_sdk_backend.py tests/test_cut_turn_tree_kill.py tests/test_tempdir_hygiene.py -q`: 274 passed, 37 skipped, 3 failed; the same 3 fail identically on untouched main on this macOS machine (BSD `mktemp` ignores `TMPDIR` without `-t`; a scope-count assertion), so they are pre-existing there and pass in CI's Ubuntu.

### DIFF
--- a/kernel/sdk_backend.py
+++ b/kernel/sdk_backend.py
@@ -29,6 +29,7 @@ import shutil
 import signal
 import subprocess
 import sys
+import tempfile
 import threading
 import time
 import traceback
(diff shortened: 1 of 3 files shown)


ksarma:
Post-merge record: the owner merged this PR as 8f4b9f1f at head b907073b before this review ran; 3 follow-ups below. The head moved from 47884e2e to b907073b (the review-fixes commit) during the review. Both heads were read and tested here.

Reviewed inline: conftest writes an owner marker (pid, start time, argv basenames) into the run root and the package state dir at mint time. Only this sweep reads it. The sweep takes direct children of `tempfile.gettempdir()` named `romp-tests-*` and skips symlinks. It keeps any root without a readable marker naming an int pid. It keeps a live pid (signal 0; EPERM counts as alive). It renames a dead root to `<name>.sweeping` before deleting. A partial failure then leaves a tombstone a later boot finishes, rather than a marker-less root refused forever. It chmods and retries a resisting child once. It logs a survivor every boot. It stops at 30 s per boot and logs the remainder. Identity is pid-only: the start time is written but never compared. A dead root whose pid a live process reused therefore stays until that process exits, which is the safe failure mode. At the merged head the sweep starts last in `_boot_reconcile`, after the orphan-CLI reap and the resume loop. It runs on a daemon thread whose starter catches every exception, so many dead roots delay neither. Three findings confirmed at 47884e2e are fixed by the second commit and pinned by its new tests. They were the synchronous sweep ahead of the reap; rmtree with `ignore_errors` deleting the marker and stranding a root later boots skip silently; and the suite-marker test sweeping and diffing the real system temp dir in a serial run. The `fix` label fits a leak with red-first tests. The remedy is a new kernel-boot mechanism, so `feature` is the stricter reading; the gate is the same under the owner's role. For the record: a conftest-side dead-owner sweep at the next run's start needs no kernel code but runs only when the suite runs again. The two could coexist.

Follow-ups (none blocks):
- `kernel/sdk_backend.py:2155` (b907073b): `_rmtree_stubborn`'s `onexc` chmods `dirname(path)` and `path` to 0o700 for whatever path failed. It does not check that either is inside the tombstone. A failure at the tombstone itself makes `dirname(path)` the system temp dir. Two causes: rmdir ENOTEMPTY from a leftover process of the dead run still writing, or a tombstone a peer sweeper already removed. A user kernel on root-owned `/tmp` gets EPERM, swallowed. A user-owned TMPDIR or a root-run kernel gets a 0700 temp dir with no log. `os.chmod` also follows symlinks, so a link inside a read-only child chmods its target outside the root. Both reproduced on private arenas (an outside file 0o644 to 0o700; a 0o555 parent to 0o700). Fix: chmod `path` only when it is a non-link directory, and skip the parent chmod when `path` is the tombstone root. Add a fixture asserting the parent's and a link target's modes are unchanged.
- PR body: it describes 47884e2e, not the merged head. It says 9 tests (14). It says the sweep is called beside the reap with the count in the summary line (it starts last, on a thread, budgeted, with its own log line). It says the suite's root survives a sweep of its parent (that assertion was removed). It does not mention the tombstone, the chmod retry or the budget. The second commit's message has all of it; a body edit closes this. Nit: the Not-in-scope line uses a word the repo's vocabulary rule bans in new prose; "a session-brief rule" says the same.
- `tests/README.md:95`: the hermetic-temp paragraph describes only run-end removal and the stderr notice. A few sentences on the owner marker, the kernel's dead-owner sweep and the `.sweeping` tombstone would tell a developer who finds either what they are.

Test evidence: at the merge base 90c0ef74 with the head's `tests/test_test_root_sweep.py`, collection fails with AttributeError on `TEST_ROOT_OWNER_MARKER` (the sweep and its constants do not exist at base). At 47884e2e 9 passed. At b907073b 14 passed, serially and under `-n 2`, from a scratch worktree under a private TMPDIR with no residue. Mutations at 47884e2e: skipping the live-pid check turns the live-owner and mixed-arena tests red; removing conftest's marker write turns the suite-marker test red. Dropping the islink guard survives, since `shutil.rmtree` refuses a symlink itself; rmtree rather than the guard pins the symlink behaviour. `tests/test_tempdir_hygiene.py` at 47884e2e: 19 passed. Privacy and vocabulary scan of both commits' added lines: clean. CI at b907073b: all ten checks green; merged 2026-09-10T20:14:50Z.

Review run by the maintainer automated review session; verified against head b907073b.

PR #1337, opened 2026-09-10
CI runs the browser-backed served-page tests in the extension job, where the browser is; a switch turns their skips into failures there (T308)

henrypinkard:
This PR touches `.github/workflows/ci.yml`: it adds one step to the extension job. No other CI change.

CI never saw the deep-link landing regression the tap test caught on main (the token pin fixed earlier today) because the ten browser-backed served-page tests (`tests/test_*_browser.py`, `tests/test_*_served.py`; nine files, sixty-odd tests) skip on every Python matrix runner: those runners have no node dependencies and no browser. The extension job installs playwright's Chromium for the pane bench, so that job now runs them too.

What changes:
- `tests/conftest.py`: `ROMP_SERVED_TESTS_REQUIRE=1` turns any skip inside those files into a red result that carries the skip's own reason (a class-level skip reports as a setup error, a method-level one as a failure; both fail the run). This is the stance the pane bench takes with `ROMP_UI_BENCH_REQUIRE`, in one place instead of nine files. Off (the default) nothing changes: contributors and the matrix jobs skip as before. One exception a test can claim for itself: a skip whose reason begins with `optional:` stays a skip.
- `tests/test_pane_hidden_word_browser.py`: its Firefox and WebKit legs say `optional:` when `ROMP_SERVED_TESTS_ENGINES` names the engines the runner installed and theirs is not among them. CI installs Chromium only (WebKit's system libraries would need apt, which that step deliberately does not run), so those two legs stay skips there and the Chromium leg is required.
- `tests/test_served_tests_require.py`: pins the switch by running pytest in a child on synthetic files outside `tests/` with the conftest loaded as a plugin: off, all skip; on, the served file's class-level skip is an error and its method-level skip a failure, each naming the switch and quoting the reason, the `optional:` skip stays a skip, and a plain-named file's skip is untouched. Red before the conftest change, green after.
- `.github/workflows/ci.yml`, extension job: `actions/setup-python` (the image's pip refuses the system interpreter), `pip install pytest pytest-timeout cryptography` (what the Python jobs install; the push tests' kernel needs cryptography), then one pytest process over the nine files with `ROMP_SERVED_TESTS_REQUIRE=1` and `ROMP_SERVED_TESTS_ENGINES=chromium`, `--timeout=600`, `--durations=10`. The matrix jobs are unchanged.

Measured locally on the devbox, one process, Chromium: the nine files ran 66 tests in 82 s wall (the slowest single test 22 s, the dashboard reload leg). A hidden browser fails loudly with the skip's reason instead of passing empty.

Proof from this PR's own run (the extension job on the final head 95cab3d5, 2026-09-10 2:07 PM Pacific): the step "Browser-backed served-page tests (pytest)" ran 70 s end to end (21:06:50 to 21:08:00 UTC, pip install included) and its log reads `67 passed, 2 skipped in 63.36s`; the two skips are the pane-hiding test's Firefox and WebKit legs, reported as `optional:` because the runner declares Chromium only. Every other browser-backed test executed there, and every matrix job is green. Along the way the step caught two things in the tap-landing file on the runner: a boot race in the chat pane that can land a notification tap on the wrong session (its one assertion is `optional:` under the switch until the fix lands, filed as T312 with the fix already verified), and a pin on the order of two kernel log lines that ride separate connections (dropped: not a fact of the design).

Tier: fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -266,3 +266,30 @@ jobs:
         env:
           ROMP_UI_BENCH_REQUIRE: "1"
         run: node --test tests/ui-bench.test.mjs
+      - uses: actions/setup-python@v5
+        # For the served-page tests below: the image's python3 would do, but its pip refuses to install
+        # into the system interpreter, and pytest is not on the image.
+        with:
+          python-version: '3.12'
+      - name: Browser-backed served-page tests (pytest)
+        # tests/test_*_browser.py and tests/test_*_served.py boot a hermetic kernel and drive the real
+        # dashboard pages in the headless Chromium installed above; on the Python matrix runners (no
+        # node deps, no browser) they skip, and say why, so a served-page regression never turned CI
+        # red (the deep-link landing pin, 2026-09-10: red on main, CI green). This is the one job that
+        # has the browser and the built dist, so they run HERE, one pytest process, with
+        # ROMP_SERVED_TESTS_REQUIRE=1 turning any skip in those files into a failure that carries the
+        # skip's reason (tests/conftest.py; the same stance as ROMP_UI_BENCH_REQUIRE above): a runner
+        # without the browser, the deps or a serving kernel turns the job red rather than green with
+        # the coverage gone. cryptography is what the push tests' kernel needs, as in the Python jobs.
+        # ROMP_SERVED_TESTS_ENGINES names the one engine this job installs: the pane-hiding test drives
+        # Chromium, Firefox and WebKit, and its legs for engines a runner declares it does not have skip
+        # as "optional", which the switch leaves alone (WebKit's system libraries would need apt, which
+        # this step does not run; see the install step above). The matrix jobs are unchanged: still no
+        # browser, still skipping these.
+        working-directory: ${{ github.workspace }}
+        env:
+          ROMP_SERVED_TESTS_REQUIRE: "1"
+          ROMP_SERVED_TESTS_ENGINES: chromium
+        run: |
+          python -m pip install --upgrade pip pytest pytest-timeout cryptography
+          python -m pytest tests/test_*_browser.py tests/test_*_served.py -q -rs -p no:cacheprovider --durations=10 --timeout=600 --timeout-method=thread
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -712,8 +712,14 @@ def _redact_report(rep) -> None:
 
 @pytest.hookimpl(hookwrapper=True)
 def pytest_runtest_makereport(item, call):
+    # ONE implementation per hook per module: a second `def` of this name would silently replace this one
+    # (it did, for an afternoon on 2026-09-10, and every report printed its values again). Anything else
+    # that shapes a test report joins here: the served-tests switch first (its message quotes the skip's
+    # reason), the redaction last, so whatever any step wrote is read for values before it is printed.
     outcome = yield
-    _redact_report(outcome.get_result())
+    rep = outcome.get_result()
+    _require_served_test_ran(item, rep)
+    _redact_report(rep)
 
 
 @pytest.hookimpl(hookwrapper=True)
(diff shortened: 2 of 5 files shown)


ksarma:
Advisory review (the owner merges): 7 items below. One item blocks at the current head: the required extension check is red there (item 1). One blocker, one resolved since staging, four follow-ups, one nit.

Reviewed inline: the review was staged at 6d912a46. The head has since moved to ac067614 (one commit rebased onto main 634563b2) and then to 54172863; the checks below cover all three. The conftest switch rewrites a skipped runtest report from a `test_*_browser.py` or `test_*_served.py` file into a failure carrying the skip's own reason. A setUpClass skip reads as a setup error, a method skip as a failure. A reason beginning with `optional:` stays a skip. Nothing changes in other files or when the switch is off. `ROMP_SERVED_TESTS_ENGINES` is split on commas and read only by the pane-hiding test. The CI step sits after the pinned Chromium install and the pane bench, with setup-python and the Python jobs' exact pip line. A runner without the browser therefore turns the job red instead of passing empty. The ten files import stdlib, cryptography and repo modules. Their node drivers load playwright and esbuild from the extension's node_modules, and dist is built earlier in the job. The `fix` label fits: the PR closes a CI coverage gap with test infrastructure and a red-first pin, no product code. The `.github/` approval rule does not apply to an admin author, as the Tier policy check says. Privacy scan of the added lines is clean.

- `.github/workflows/ci.yml:295`, blocker at the current head. The required extension check is red inside this step at both post-rebase heads. Each run fails a different order-sensitive assertion in `tests/test_notification_tap_resume_browser.py`. At ac067614 it is line 486, the deep-link active tab, which 54172863 makes optional under the switch. At 54172863 it is line 463, the shown ack logged after the clicked ack (1 failed, 66 passed, 2 skipped). Neither assertion is this PR's code, and the rebase touched no landing or push path. Still, the module has gone red in 2 of the step's 3 runs, so a re-run alone leaves a required gate that flakes for every PR. Suggest making the class's cross-connection order pins optional, or keying them on the response, before this gates. Also, the T312 skipTest ends the method, so it skips the token-scrub pin the body names too.
- `tests/conftest.py:784` at 6d912a46, resolved at ac067614. The staged head defined pytest_runtest_makereport twice, so the redaction hookwrapper never registered. All four Python jobs failed 11 HookEndToEnd cases. The rebased conftest folds the served check into the one hook ahead of _redact_report, and the file passes 50/50.
- `.github/workflows/ci.yml:210`, follow-up. The extension job keeps timeout-minutes 15 while the step adds --timeout=600. Setup plus the pytest phase plus 600 s leaves about 80 to 130 s under the cap. The file's own T230b note rejected that margin for the Python jobs. Raise the cap to 25, or size the per-test timeout to the step (slowest test 22 s; the drivers cap at 300 s).
- `tests/conftest.py:778`, follow-up. The selector and the globs cover ten files. Thirteen more files carry the same browser-gated skips and still skip on every runner. Among them: `test_ship_reship.py`, `test_awaiting_box_sync.py`, `test_scroll_marks_click.py`, `test_spend_modal_headless.py` and nine others. So the body's claim that every browser-backed test executed overstates. A marker or module attribute set by the served classes would let a test opt in where it is written. test_ship_reship's lab writes the runner's environment into cfg.json, so widening needs a read of each file first.
- `tests/README.md:91`, follow-up. ROMP_SERVED_TESTS_REQUIRE, ROMP_SERVED_TESTS_ENGINES, the selector and the `optional:` prefix are documented only in comments. The sibling bench switch has a line here and in `CONTRIBUTING.md:131`.
- PR body, follow-up. The globs match ten files with 69 tests, not nine and 66. The proof paragraph cites a run whose four Python jobs were red. The body predates the hook fix, T312 and the fifth changed file.
- `tests/test_pane_hidden_word_browser.py:84`, nit. The retained skip text says CI installs no browser, and the docstring at line 12 says CI sets no node path. Both are false after this PR.

Red-first holds. At the base e8b5be8e with only the PR's pin file checked out, `tests/test_served_tests_require.py` reads 1 failed, 1 passed; in the on case the child pytest exits 0 with 4 skipped and no hook. At 6d912a46, ac067614 and 54172863 it reads 2 passed. `tests/test_env_value_redaction.py`: 11 failed, 39 passed at 6d912a46; 50 passed at ac067614 and 54172863. The served modules were not run here; the CI step is the run of record. It read 67 passed, 2 skipped at 6d912a46, then 1 failed, 66 passed, 2 skipped at ac067614 and again at 54172863. CI at 54172863: Python 3.10 to 3.13, bats, gitleaks and both tier checks green; the extension check red; merge state blocked.

Review run by the maintainer automated review session; verified against head 54172863.

PR #1338, opened 2026-09-10
API health across every connected kernel: one dot in the spend readout, a plain-words reading, a graph (T301)

henrypinkard:
## Summary

The API health indicator on the shell's rail is redesigned to the user's spec (T301). It covers every connected kernel, it is one small dot inside the spend readout, it reads in plain words instead of the state machine's vocabulary, and its popup carries a graph instead of per-window percentage rows.

## Design points

**Every connected kernel.** Each kernel serves its last local apiHealth shell frame at `GET /api-health/frame` (its local half only, never its view of its peers, so two kernels attached to each other never nest). The tunnel supervisor's pass polls every attached host's frame (once per pass, about every 15 s; kept on a blip; kept and marked with a `fault` when the read is refused, so a rotated token names the machine instead of losing it; cleared when the host answers that it has none) and the shell frame carries them under `hosts`, a per-host map with a `stale` mark (a down tunnel or a refused read). Each frame also carries `quiet` (no API event in that kernel's longest window) and `errs` (failed attempts in it), from the aggregator every cycle, so the frame changes and is pushed the moment the last failure ages out. The remote's frame, its poll stamp and its fault are per process (`_NOT_SAVED`): never written to the 0600 remotes file, never restored at boot. For the history the browser reads `GET /remote/<host>/api-health`, a read relay beside the `/ws` and `/file` relays (404 unknown host, the remote's token rewritten in, JSON passed through, 502 and a redial on a dead tunnel). The merge is client-side in `ui/webview/api-health-merge.ts` (loaded by the shell as `/dist/api-health-global.js`, the age-colour module's pattern): per-host maps in, one line per machine out, worst state wins for the dot, no count or clock added to or compared with another kernel's. The dot follows the frames alone (their state, `quiet` and `errs`); a history reading only words a machine's line, so a reading taken at an earlier hover can never pin the dot. A machine whose tunnel is down or whose frame could not be read is named with what it last said and has no say in the dot. Why the split: the dot must be live without the browser polling N hosts and the supervisor already dials every host, so the frame rides the push the shell listens to; the history is on demand and a few KB.

**The rail cell.** `#rail-api` is a dot alone (no second API word, no "ok"). It moves into the spend readout's slot right after that readout's `API` label and left of its 1-day segment whenever the readout renders (the stable node is re-parented, so its listeners survive the readout's rebuilds), and sits after the usage bars when there is no readout. Colours through tokens with fallbacks: the accent when every kernel is fine, the blocked red when errors are being met anywhere (429 storm, 5xx, offline, paused), the label gray when no kernel has API traffic in the windows. The usage tip yields while the pointer is on the dot and comes back when the pointer slides onto the figures (the dot's mouseenter and mouseleave, no timer); the dot's click and keyboard activation stop at the dot, so they never open the spend modal; the readout's minute repaint re-parents the dot with focus kept and a focus-shown hover intact; the slot is `display:contents`, so a hidden dot costs the readout no gap.

**The reading.** The state machine and `/api-health`'s schema are unchanged. The popup reads what happened over the longest window: traffic with no errors is "N requests in the last 15 min, all succeeded" (with "too few requests to call a trend yet" as a sub-line while the machine still says `unknown`), errors are counted in plain words ("Rate-limit storm: 9 rate-limited attempts, 1 server error among 45 attempts in the last 15 min; 1 turn gave up."), no traffic is quiet. A storm in the window reddens the dot even with nothing waiting right now (the frame's `errs`). The word `unknown` appears nowhere the user reads. With several machines the head sums them up ("Errors on TESTHOST") and each machine gets its own line.

**The popup.** The per-window rows are gone. The snapshot gains an additive per-bucket `series` (attempts per minute over the 900 s window: ok / 429 / 5xx / no status / other; schema stays 1), and the popup draws it in the usage hover's graph grammar (polyline + fill, one ceiling label, a tick every five minutes) with 429 attempts in the blocked red and 5xx in the warn amber (a 1-2-5 ceiling at any magnitude), followed by one legend sentence for the codes. Each machine's document lands on its own, so a hung tunnel never holds back this machine's numbers. The "worst of N buckets" and "kernel up" caveats are gone; State changes are capped at four and worded plainly (rate-limit storm, API failing, recovering, fine, quiet); the pause control and its plain-words reasons stay.

## Tests

- `tests/test_api_health_hosts.py` (new): the frame's hosts map, its quiet and errs flags through the patched `_sdk` seam (no backend built), a refused read's fault, `/api-health/frame` through the handler (503 then the local half), the supervisor poll's rate gate and clear/keep/fault contract (403 pinned), the relay (auth, 404, 502, token rewrite, passthrough), the series bins from a synthetic ring through `snapshot()`, the per-process keys never reaching the remotes file or coming back at boot, the CSS pins for the dot's three states, the slot, the tip hooks and the focus-keeping move.
- `ui/webview/api-health-merge.test.ts` (new): worst-state merge, per-host lines, the dot from the frames alone (a stale reading cannot pin it), a down or refused machine named and not counted, no cross-host arithmetic (a source pin too), the reading rule (traffic and no errors reads fine; none reads quiet; errors in plain words), the histories merge with pending and failed reads, the per-document series sum, the shell's module load.
- `tests/test_api_health_browser.py`, `tests/test_api_health_hover_browser.py`, `tests/test_api_health_hover.py`, `tests/test_api_health_rail.py`, `ui/webview/rail-spend.test.ts`: pins moved to the dot, the plain-words reading, the graph and legend, the four-row tail, a new served scenario with an attached host read through the relay, and the served cell inside the readout: its click never reaches the readout's click, the two tips never show at once, the repaint's move keeps focus and the hover.

## Docs

`docs/reference.md`: the `series` field, an "On the dashboard" subsection (the dot's colours and reading, the multi-host merge), the frame block gains `hosts`, `quiet`, `errs` and the per-host `fault`; the indicator section describes the new reading, the poll cadence (once per pass, about every 15 s) and what has no say in the dot. `docs/guide.md`: the bottom bar paragraph.

## Review fixes folded before arming

The adversarial review confirmed nine findings on the first head, all folded here with pins: the remote frame keys never persisted to the 0600 remotes file (they rewrote it every pass and restored a dead host's storm at boot); the dot's click no longer opens the spend modal; the fixtures patch the `_sdk` seam so no test builds a real backend; the two tips never show at once; the dot follows the frames instead of a hover's reading; the empty slot costs no gap; a refused frame read keeps the machine with its fault; the series is pinned from `snapshot()`; the docs give the real poll cadence. Also from the review: a down machine's last frame no longer votes, a hung tunnel no longer holds back this machine's history, the graph's ceiling scales past 1000, and the readout's repaint keeps focus.

Tier: feature

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -807,17 +807,20 @@ separating your sessions from the judge pipeline. You can also reconfigure the
 judges from the gear: the high-volume indexing tier defaults to Haiku, and the
 judgment tier defaults to Sonnet.
 
-The bottom bar's **API** cell, a dot and a word, shows how the API is treating
-your sessions. Gray **ok** means no session is waiting on the API. Amber shows
-how many sessions are waiting and names the problem: **rate limited**,
-**overloaded**, **offline** (this machine cannot reach the API), or **errors**.
-Red **paused** means auto-retry and the judges are stopped, and says why: a
-usage limit, the monthly spend cap, or that you stopped them. Hover for the same
-reading with the waiting sessions listed, and the history under it: the API's
-state over the last 1, 5 and 15 minutes (attempts, the 429 and 5xx shares,
-give-ups, sessions that retried) and the most recent state changes with how long
-each held. A kernel restart shows as its own line there, because the counts
-start over with the kernel. Click the cell, or press Enter on it,
-for the detail: each waiting session (click one to open that session), a button
-that stops auto-retry for every session while sessions are waiting and resumes
-it while paused, and links to the usage figures and the Log.
+The bottom bar's API readout carries one small dot, right after its **API**
+label, that shows how the API is treating your sessions on
+every connected kernel. Blue (the romp accent) means everything is fine. Red means errors are
+being met somewhere: a 429 rate-limit storm, 5xx failures, a machine that
+cannot reach the API, or auto-retry paused (a usage limit, the monthly spend
+cap, or you stopped it). Gray means the API is not being used right now: no
+traffic in the last 15 minutes on any machine. Hover for the reading in plain
+words (for example, 4 requests in the last 15 minutes, all succeeded), one
+line per machine when several are connected, the waiting sessions listed, and
+the history under it: a graph of attempts per minute over the last 15 minutes
+with rate-limited attempts in red and server errors in orange, one sentence
+explaining the codes, and the most recent state changes with how long each
+held. A kernel restart shows as its own line there, because the counts start
+over with the kernel. Click the dot, or press Enter on it, for the detail:
+each waiting session (click one to open that session), a button that stops
+auto-retry for every session while sessions are waiting and resumes it while
+paused, and links to the usage figures and the Log.
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -809,11 +809,14 @@ judgment tier defaults to Sonnet.
 
 The bottom bar's API readout carries one small dot, right after its **API**
 label, that shows how the API is treating your sessions on
-every connected kernel. Blue (the romp accent) means everything is fine. Red means errors are
-being met somewhere: a 429 rate-limit storm, 5xx failures, a machine that
-cannot reach the API, or auto-retry paused (a usage limit, the monthly spend
-cap, or you stopped it). Gray means the API is not being used right now: no
-traffic in the last 15 minutes on any machine. Hover for the reading in plain
+every connected kernel. The accent colour (blue in the dark theme, clay in the
+light one) means everything is fine. Red means errors are being met somewhere:
+a 429 rate-limit storm, 5xx failures, a machine that cannot reach the API, or
+auto-retry paused (a usage limit, the monthly spend cap, or you stopped it), and
+it stays red while a failed attempt sits in the last 15 minutes anywhere. Gray
+means the API is not being used right now: no traffic in the last 15 minutes on
+any machine. A machine whose link is down is named in the popup with what it
+last said and does not colour the dot. Hover for the reading in plain
 words (for example, 4 requests in the last 15 minutes, all succeeded), one
 line per machine when several are connected, the waiting sessions listed, and
 the history under it: a graph of attempts per minute over the last 15 minutes
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1750,11 +1750,22 @@ orange, and one sentence explains the codes.
 The signal covers every connected kernel, not only the one serving the page.
 Each kernel serves its own last shell frame at `GET /api-health/frame` (its
 local half only, never its view of its peers), and the tunnel supervisor polls
-every attached host's frame (every 10 seconds, kept on a blip, cleared when the
-host answers that it has none) and carries them in the shell frame under
-`hosts`, a map keyed by host name with each machine's `state`, class, headline,
-waiting count, since, pause reason and a `stale` mark when that tunnel is not
-up. The hover's history reads each attached host's document through `GET
+every attached host's frame (once per supervisor pass, about every 15 s; kept on
+a blip; kept and marked with a `fault` when the read is refused, a 403 from a
+rotated token or a 500; cleared when the host answers that it has none) and
+carries them in the shell frame under `hosts`, a map keyed by host name with
+each machine's `state`, class, headline, waiting count, since, pause reason,
+its `quiet` and `errs` flags when that kernel sends them, and a `stale` mark
+when that tunnel is not up or the read was refused (the frame's own `type`,
+`sessions` and `seq` stay on their kernel). The frame's `quiet` says that
+kernel saw no API event in its longest window and `errs` counts the attempts
+that failed in it; both come from the aggregator every cycle, so the frame
+changes, and is pushed, the moment the last failure ages out. The dot follows
+the frames alone: red when any reachable machine's frame is degraded, paused or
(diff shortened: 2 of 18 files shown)


ksarma:
Advisory review (the owner merges): 9 items below. The first four are wrong colours or hover states on the new dot; fix them before or soon after merging.

Reviewed inline: every body claim checks out against the diff except two, the storm-in-window red (item 1) and the usage tip yielding to the dot (item 4). `GET /api-health/frame` sits after the `_authorize` gate. It serves the local frame without its hosts map and answers 503 before the first cycle. The relay at `GET /remote/<host>/api-health` sends only the remote's own token in `X-Romp-Token`, never the browser's credential. It answers 404 for an unknown host, and 502 plus a redial on a dead tunnel. The supervisor poll runs on a 10 s wall-clock gate, a timer of the same shape as the existing usage poll, with keep-on-blip and clear-on-empty. The frame builder excludes peers, so two attached kernels never nest. The series is additive: 15 one-minute bins over the 900 s window, schema 1 unchanged. The dot re-parents one stable node into the readout's slot, so its listeners survive rebuilds. The popup keeps the dot, hover, pinned-detail disclosure. The graph reuses the usage hover's grammar, and there are no new font sizes. The `feature` tier fits. A scan of the 1348 added lines found no private strings, hostnames or home paths.

All nine are follow-ups.

1. `kernel/kernel.py:47475`. `READINGS` is written only by `load()` and never cleared in `hide()` or `close()`. Every frame repaints through it, so after one hover the dot keeps that hover's reading until the next hover: red after a storm has ended, gray after traffic resumes. Before any hover the frame alone paints, so the body's storm-in-window red holds only after a hover. Carrying the window verdict in the frame beside `quiet` would let the frame decide the dot.
2. `ui/webview/api-health-merge.ts:162`. Any error count above zero in the window reads level `errors`, whatever the state machine says. The `recovering` arm at line 173 is unreachable because `errors === 0` already returned. One retried 429 among 200 requests under a healthy verdict paints the rail red for 15 minutes. Gate the level on the machine's verdict or a rate threshold, and keep the counts in the words.
3. `ui/webview/api-health-merge.ts:93`. A stale row ranks like a live one. A peer that was degraded or paused when its tunnel dropped keeps the merged dot red, with the head reading Errors on that host, until it reconnects. The usage rows include a remote only when its status is up. Exclude or down-rank stale rows while keeping the line.
4. `kernel/kernel.py:47036`. The dot now lives inside the readout, whose only pointer hooks are `mouseenter` and `mouseleave`. Sliding from the dollars onto the dot leaves both tips open. Entering through the dot and then sliding to the dollars shows no usage tip until the pointer leaves and returns. Both reproduced with a real pointer in Chromium; the base layout had neither. A bubbling `mouseover`/`mouseout` on the readout keyed on the target, or `show()` hiding the usage tip, closes both.
5. `kernel/kernel.py:49807`. The fine dot uses `var(--accent)`, which `ui/CLAUDE.md:83` reserves for chrome. The two tests that pinned the rule were inverted, and the rule file is untouched. The comments at 49792 and 49826 still say the opposite of the CSS beneath them. In the light theme the accent is clay, so fine and errors are two warm reds, and `docs/guide.md:812` (blue means fine) is false there. Record the T301 exception in `ui/CLAUDE.md` and make the guide's colour word theme-neutral, or use a status token such as `--st-ready-bg`.
6. `kernel/kernel.py:43805`. `_apih_quiet` calls `_sdk()`, which builds the real `SdkBackend` (boot-reconcile thread, model-catalog boot). It does so inside the rail fixture and the new kernel-side module that reuses it, both of which stub everything else the frame reads. A bare run without conftest's floors reaches the apiKeyHelper and the Models API. Read the singleton without building it, or stub `_sdk` in the fixture. Also pin `quiet` False through a stubbed backend: the one assertion, `quiet` is True, also holds when the `except` branch fires.
7. `kernel/kernel.py:51363`. `GET /api-health/frame` is never requested over HTTP; the one test calls the helper. Moving the route above the auth gate or renaming the path passes the suite (checked by mutation). Three requests in the module's existing `Relay` class would pin it.
8. `kernel/kernel.py:47511`. `graphHTML` runs under no test: no fixture carries a series, and the storm test asserts zero graphs. An inverted Y mapping and a broken tick loop pass everything (checked by mutation). Add one fixture bucket with a series, or a node test over the extracted function. The `per` variable at 47520 is unused.
9. `kernel/kernel.py:47394`. With the merge module absent (a stale dist or a failed script fetch), the popup headline reads `Fine.` during a storm. History shows no reading and no notice, though the comment above says the popup reports it in the kernel's words. One `.ah-err` line when `MERGE` is null meets the loud-failure rule.

Nits: the new kernel-side test module's file name, one test name and two docstrings (`kernel/kernel.py:43818`) add the word the vocabulary rule bans. `docs/reference.md:1931` says the frame is unchanged, while the section above documents it gaining `quiet` and `hosts`.

Ran locally at head, in a capped scope with the Anthropic keys removed and scratch state. The PR's six non-browser pytest modules: 166 passed. The same files at the merge base fail 36 of 166, so the tests exercise the change. `npm run typecheck` and `node esbuild.js --tests` are clean. `node --test` on api-health-merge, rail-spend and spend-windows-hover: 28 pass, 0 fail. The two browser modules were reviewed by reading only; they also skip in CI. CI at 897e22bf: all ten checks SUCCESS.

Review run by the maintainer automated review session; verified against head 897e22bf.

PR #1340, opened 2026-09-10
Restart monitors: what kernel restarts do to the sessions, measured (stage 0 of #1317)

henrypinkard:
Part of #1317, stage 0 ("Measurement comes first"): before any restart behaviour changes, the monitors exist, a baseline is recorded over a deploy week, and the same reader keeps running after the change. Everything here is additive: no restart behaviour changes.

## What lands

**Ledgers (kernel/sdk_backend.py).** Two append-only files under the state directory, written best-effort and rotated at 32 MB with one predecessor kept:
- `session-events.jsonl`: one flat row per thing that went wrong with a session's process (an orphaned CLI ended at boot, a leftover scope stopped, two CLIs holding one conversation, a crash heal or loop, a session the drain's bound left closing) and one summary row per boot sweep that had sessions to reconcile. Written through `problem_row()`, which also puts the prose on the problem ring (the bell) and the kernel-log line as `<prose> ;; problem-row {json}`; the row shape is shared with the lease work (T305), which writes its `lease.*` kinds through the same helper. `SdkBackend._log` gains `ring_text` so the ring stays readable while the log line carries the object.
- `turns.jsonl`: one row per settled turn with event stamps only: the feed pop (`fedT`), the first streamed work atom (`firstOutT`, new: there was no durable stamp for it), the ResultMessage (`resultT`), the CLI's own `durationMs` and `apiMs`, the spend fold's dollars and tokens, and `resumeNotice` for a turn that redoes cut work.

**Restart rows carry the kernel's own size.** `restart-cuts.jsonl`'s cut and boot rows gain `rssKb` and `cpuS`, sampled at the two events the ledger already records, so the kernel's growth between restarts is a series with no sampler of its own.

**The manager's quiet window writes a row.** `quiet-window` in `restart-audit.jsonl` at apply: the wait from parked to restart, whether the fifteen-minute backstop fired, the coalesced requests and the park's drain-hold counts. A note, not a request: the kernel's restart-reason walk lists it in `_NO_RESTART_ACTIONS`.

**`GET /session-events`** (token-gated): the ledger newest first since a stamp, each row with `host`, and `count` since this kernel's boot (never summed across kernels; the dashboard cue is the existing bell, per the T301 vocabulary).

**`romp restart-metrics`** (cli/restart_metrics.py, read-only, no kernel module loaded): turns cut per restart and per window, outage and reconcile times, quiet-window waits and backstop firings, orphans and reaps, duplicate CLIs, crash heals, drain leftovers, lease problems, continuation notices and redo cost, turn latency (feed to result and to first output, beside the state log's one-second series for turns before `turns.jsonl`), machine cuts by cause, per-session memory and CPU from the `romp-session-*` cgroups (ps where there is none), the kernel's own CPU, resident size and idle-cycle share from `/version` and `/perf`, and a live check for two CLIs on one conversation. Day or week windows, `--json` for the figure script, a missing ledger named loudly.

**`scripts/restart_metrics_report.py`**: the before-versus-after figures with cleanplots under uv (not a romp dependency; the script says so when absent): cut turns, restart timing, quiet-window wait, sessions gone wrong, redo cost, turn-latency distribution, session and kernel memory, the kernel's memory at each exit and boot.

## Design points
- Every stamp is an event's time; nothing is sampled by a timer.
- Redo cost comes from `turns.jsonl`; the spend ledger has hour and day buckets per session and no per-turn rows, and the summary says so.
- A boot with no sessions writes no summary row, so a read-only route's lazy backend build leaves the state directory untouched.
- Limits: `turns.jsonl` at about 300 bytes a turn and the 32 MB rotation holds roughly a month here.

## Tests
`tests/test_session_events_ledger.py` (every row kind, the log-line parse, the ring text, rotation, the pure duplicate-CLI helper, the turn row), `tests/test_session_events_route.py` (the route, the audit-walk exclusion, the kernel sample on restart rows), `tests/test_restart_metrics.py` (every parser, the window arithmetic, the cut-to-boot join, the summary text, the live helpers on synthetic cgroup files and ps lines), `tests/test_restart_metrics_report.py` (data shaping always; drawing end to end under cleanplots, skipped with a named reason without it), `tests/manager-quiet-window-audit.test.js`, `tests/romp-restart-metrics.bats`. Docs: `docs/reference.md` (the command table, the ledger and row formats, a Restart metrics section).

Merge order with the stacked lease pull request (T305): this one first; it rebases.

### DIFF
--- a/bin/romp
+++ b/bin/romp
@@ -658,6 +658,7 @@ EOF
     _romp_cmd "romp up [--foreground]"     romp-manager "Start the kernel manager: through the login service when installed, else in the foreground"
     _romp_cmd "romp version"               romp-version "Versions across the parts (tree, kernel, served vs built bundles)"
     _romp_cmd "romp spend-rebuild"         romp-spend-rebuild "Recount the spend ledger's token columns from the transcripts (dry run; --apply writes, dollars untouched)"
+    _romp_cmd "romp restart-metrics [--json|--window day|week]" romp-restart-metrics "What kernel restarts do to the sessions, from the state ledgers and the kernel's routes: turns cut, outage and reconcile times, quiet-window waits, orphans and reaps, crash heals, redo cost, turn latency, per-session memory and CPU"
     _romp_cmd "romp uninstall"             romp-uninstall "Remove romp's installed footprint (--purge also deletes recorded state)"
     _romp_cmd "romp help"                  -            "Show this help"
     cat <<'EOF'
@@ -753,6 +754,17 @@ fi
 if [[ "${1:-}" == "spend-rebuild" ]]; then
     shift
     exec romp-spend-rebuild "$@"
+fi
+
+# `romp restart-metrics [--json] [--window day|week] [--anchor D] [--since D] [--until D] [--tz Z] [--no-live]`
+# — the restart monitors (stage 0 of the restart-surviving sessions program, 2026-09-10): what kernel
+# restarts do to the sessions, read from the state ledgers (restart-cuts, restart-audit, session-events,
+# turns, the state logs, spend) and the running kernel's /version and /perf; a text summary per window, or
+# the whole document as JSON for the figure script. Read-only; never loads a kernel module. Delegates to
+# romp-restart-metrics (python, cli/restart_metrics.py).
+if [[ "${1:-}" == "restart-metrics" ]]; then
+    shift
+    exec romp-restart-metrics "$@"
 fi
 
 # The kernel serve token — required on EVERY kernel/bus request, loopback included (Jupyter's
--- a/bin/romp-manager
+++ b/bin/romp-manager
@@ -266,6 +266,31 @@ function auditSigterm(spec, pid, reason, trigger) {
 // kernel's row names the same cause.
 function stopTrigger() { return downMarkerHeld() ? 'cli-down' : 'stop'; }
 
+// T304: the quiet window's OWN row in restart-audit.jsonl, written when the parked refresh APPLIES, so the
+// wait from parked to restart and the fifteen-minute backstop's firing are machine-readable (until now they
+// were the `applying deferred refresh … waited Ns` log line alone). A mechanism note like manager-sigterm,
+// never a request: the kernel's reason walk lists `quiet-window` in _NO_RESTART_ACTIONS and passes it over;
+// the restart it releases writes its own manager-sigterm note. `since` and `t` are epoch seconds; `reason`
+// is quietGate's verdict verbatim ('quiet', 'backstop cap', or the unreachable-kernel sentence), `backstop`
+// the one bit a reader wants most; the drain counts are the park's evidence (queueQuietRestart records them).
+function quietWindowRow(park, g, now) {
+  const p = park || {}; const since = Number(p.since) || now;
+  return { t: Math.floor(now / 1000), action: 'quiet-window', since: Math.floor(since / 1000),
+           waitedS: Math.max(0, Math.round((now - since) / 1000)), reason: String((g && g.reason) || ''),
+           backstop: !!(g && g.reason === 'backstop cap'), coalesced: p.count || 1, mode: String(p.mode || 'all'),
+           lastInflight: p.lastInflight === undefined ? null : p.lastInflight, misses: p.misses || 0,
+           drainRefusedCount: p.drainRefusedCount || 0, drainArmedCount: p.drainArmedCount || 0,
+           drainRefusedAfterArm: p.drainRefusedAfterArm || 0 };
+}
+function auditQuietWindow(park, g, now) {
+  const row = quietWindowRow(park, g, now || Date.now());
+  try {
+    fs.mkdirSync(STATE_ROOT, { recursive: true });
+    fs.appendFileSync(path.join(STATE_ROOT, 'restart-audit.jsonl'), JSON.stringify(row) + '\n');
+  } catch (e) { /* a lost note, not a lost restart */ }
+  return row;
+}
+
 // The exit handler's verdict (pure, exported for tests): what to log and whether to respawn. A
 // kernel we asked to stop is reaped; one we SIGTERMed for a restart keeps the plain respawn line;
 // one that exited with NEITHER on record (a signal from elsewhere, or a crash) says so, because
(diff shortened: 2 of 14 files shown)


ksarma:
Advisory review (the owner merges): no objection. 7 items below, none a blocker. The head moved during the review. a13bd0c3 rebases the same seven commits onto main and adds one fix commit. The rebase removes the `bin/romp-manager` conflict that made faba36d7 unmergeable. Everything below is against faba36d7; where I read the fix commit's diff, I note what it changes.

Head note: the head moved twice while this review ran, to a13bd0c3 (the rebase the text above describes, plus one fix commit) and then to 9c698105 (one more commit to the turn ledger's ResultMessage branch, 11 lines). The rebased seven commits were compared here line for line with the reviewed patch and are identical, so the items below carry over. The two fix commits are reviewed separately; a delta comment follows.

Reviewed inline: both ledgers go through one guarded append helper, so a failing write costs the row and logs it. They rotate at 32 MB by `os.replace` with one predecessor. The reconcile sweep, crash heal and loop, drain and settle paths that write them are otherwise unchanged, so the no-behaviour-change claim holds. `rssKb` and `cpuS` on the cut and boot rows come from a guarded sample of the kernel's own process. `GET /session-events` sits after the `_authorize` gate beside `/api-health`. It adds only the host field and counts since this kernel's boot. The manager writes the quiet-window row at apply, before the restart it releases. The kernel lists the action in `_NO_RESTART_ACTIONS`, so the restart-reason walk continues past it. The other two audit readers filter by action. `cli/restart_metrics.py` imports the standard library only. It reads the ledgers, `/version` and `/perf` without writing. The report script exits with a message when cleanplots is absent. `firstOutT` is stamped in `_forward` on a fed turn's first streamed work atom: two attribute reads and one time call. Nothing samples on a timer. Against the #1317 list: boot reaps and scope stops, the parked wait, the kernel's memory and CPU, and per-turn and first-output time now have rows. Per-session memory is a live read only. A cut turn's own wasted spend has no row; redo cost is the redo turn's cost via `resumeNotice`. `cutTurns` is still the drain's in-flight list, with no `inflightAtExit` or transcript-verified count. The label fits. The privacy scan of the added lines found nothing.

- `kernel/sdk_backend.py:5760`, follow-up (fixed at a13bd0c3 by reading): a turn the CLI opened itself carried the previous fed turn's `fedT` and `firstOutT`. `parse_turns` turned them into latency samples. The fix omits both stamps when the turn had no feed. Rows a pre-fix kernel already wrote still carry them, so a reader gate on fedTexts would clean the baseline.
- `kernel/kernel.py:15123`, follow-up: count uses `int(_STARTED)` and the default since uses the float. A row stamped in the boot second is therefore counted and missing from rows; a predecessor's `drain.unjoined` or this kernel's first reconcile row are examples. Every row carries pid. Compare that, or use `int(_STARTED)` for both.
- `scripts/restart_metrics_report.py:374`, follow-up: `summary.txt` is written from the raw documents. It therefore names the three largest sessions and any duplicate CLIs beside the anonymized figures. The Render test checks only its header.
- `docs/reference.md:1815` and `kernel/sdk_backend.py:8656`, follow-up: the docs, the header comment and the body say every non-summary kind goes through `problem_row`. `drain.unjoined` is appended directly, with no text field and no marker log line. Route it through `problem_row` with `ring=False`, or name the exception.
- `kernel/sdk_backend.py:12026`, follow-up: the `firstOutT` capture in `_forward` has no test. a13bd0c3 pins the append by a source scan. Nothing feeds a work atom and a `ResultMessage` through the handler and reads `turns.jsonl`.
- `scripts/restart_metrics_report.py:37`, follow-up: the default output directory is a sibling of romp's state directory, and `romp uninstall --purge` leaves it behind. The anonymize boundary is the whole XDG state root. The docs call it the state root where they elsewhere say state directory.
- `bin/README.md:57` and `cli/README.md:12`, follow-up: no row for the new verb.

Nothing from the PR ran here. The reaper, scope and drain paths are read-only on this machine, so CI is the execution evidence. At faba36d7 only the tier checks completed before the head moved. At a13bd0c3 bats, gitleaks and the tier checks pass; Python 3.10 to 3.13 and the extension job are still running.

Review run by the maintainer automated review session; verified against head faba36d7.

ksarma:
Post-merge record: the owner merged this PR as 15c3d1d1 at head 9c698105 while this review ran; no objection. 6 items follow, none a blocker. This is the delta over the advisory posted at faba36d7. The head is now 9c698105: the same seven commits rebased onto main, plus the two fix commits reviewed here.

Reviewed inline: `_LEDGER_LOCK` is a leaf lock. It is held only around the stat, the rename and the append, with no callback under it. No appender holds another lock at the call; the crash-heal path releases the backend lock first. The SIGTERM drain appends from the main thread, which otherwise never takes the lock. Every fed turn goes through the single fresh-feed pop, which stamps `_fed_t` beside `since`. The boot and crash resume notices, the move settle and the rename ping therefore all carry the stamp. The ResultMessage branch is one try again. `delta` and `turn_u` are read only inside the `total > 0` block. The row is written from the finally under its own guard. The three stamps are cleared unconditionally. The settle runs whether the spend fold raised or the append's guard caught. `parse_turns` accepts any positive `fedT`, so a pre-fix integer stamp parses. Each of the twelve folds is coherent. Every reader fold has a test that fails at 081ff511 by reading. The kernel folds are pinned at the row builder and by source text. Of the earlier seven items, (a) is resolved on the writer side. Item (e) still applies and now covers the new pop stamp. Items (b), (c), (d), (f) and (g) still apply, because the fix commits do not touch those lines. No em-dash remains in any user-facing string; the one in the added lines is a kernel code comment. The privacy scan of the 214 added lines found nothing.

- `cli/restart_metrics.py:270`, follow-up: `parse_turns` has no `fedTexts` gate. A pre-fix kernel wrote rows for self-opened turns with `fedTexts` 0 and the previous fed turn's stamps. Those rows still become hours-long feed-to-result samples and duplicated first-output samples in the baseline. Skip the interval when the row carries `fedTexts` and it is 0. Add such a row to the Parsers fixture.
- `kernel/sdk_backend.py:5535` and `:12290`, follow-up: the `_fed_t` stamp at the pop and the `_first_out_t` capture in `_forward` are the only writers of the two stamps. No test reaches either; the ledger tests preset both on a `__new__`-built double. Deleting either line leaves the suite green and every row without the key. The reader then reports no latency with no error.
- `tests/test_session_events_ledger.py:316`, follow-up: the finally-path row and the stamp resets are pinned by source-text indices only, one on a 15-space comment alignment. A test could push a ResultMessage through `_on_message` with a raising `_record_spend` and read the row and the cleared attributes. That would pin the behaviour and cover the item above.
- `tests/test_restart_metrics_report.py:51`, follow-up: the small-n points-or-box choice and the measured/unmeasured annotation run only under the Render class. CI skips that class: the Python job installs no plotting library. Even there the fixture never reaches five samples, and nothing reads the annotation. A pure frames-layer helper would make both testable without matplotlib. The a13bd0c3 message says every fold is pinned; for these two folds and the two stamp sites it is not.
- `kernel/sdk_backend.py:6573`, nit: the ledger guard reports its fault with a bare `_log` inside the finally, ahead of the stamp resets and the settle. That violates the finally's own ordering rule; the settle tail guards its own `_log` for that reason. The kernel's log callback swallows through `_exit_log`, so no shipped path reaches it. Wrap it, or push the fault onto `failed`.
- `tests/test_restart_metrics.py:154`, nit: the new machine-cut test was inserted above the last line of `test_events_turns_and_state_log`. The `spend_by_day` assertion now runs inside the machine-cut test. Move it back.

Nothing from the PR ran here: the kernel paths stop scopes and read cgroups, so CI is the execution evidence. At 9c698105 all ten checks pass: Python 3.10 to 3.13, bats, gitleaks, the extension job and both tier checks.

Review run by the maintainer automated review session; verified against head 9c698105.

PR #1342, opened 2026-09-10
Session leases: a CLI is owned by pid, start time and heartbeat, not by its parent (stage 1 of #1317)

henrypinkard:
Part of #1317 (sessions that survive a kernel restart), stage 1 of the staged plan: ownership by lease and identity, no host yet.

## What changes

**A lease file per session** at `STATE/leases/<sid>.json`, written by the kernel the moment the SDK connect hands it the CLI, removed when the CLI's client closes (so only a kernel death leaves one behind). Fields: the romp sid and the conversation id the argv carries, the CLI's pid and start time, the holder's pid and start time (the kernel in this stage; the host in stage 4), the kernel's code version, a spawn stamp, a heartbeat.

**Heartbeat on the drain hold's cadence, exactly.** One backend thread beats every 3 s while any lease is held; a lease is fresh for 12 s (`LEASE_TTL_S == DRAIN_HOLD_TTL`: four beats, so it outlives a missed beat and not a dead holder). Holder identity is the primary check and the heartbeat the second line, against a holder that is alive but wedged.

**The reaper reads leases** (`lease_census`, wired into `find_orphan_clis` when given the leases). Per SDK CLI carrying one of our conversation ids:
- a valid lease (fresh beat, holder alive by pid and start time, CLI alive by pid and start time) means owned by its holder, whatever its parent; a lease from another code version is owned and reported;
- else a direct child of this kernel is owned (the spawn-to-lease-write window, by design, no row);
- else a child of a live kernel is kept and reported (`lease.cli-without-lease`): the previous code version's kernel wrote no leases, and its sessions survive the upgrade boot as they did before;
- everything else is an orphan and is ended with its tree, its lease removed.
Since the kernel is the holder, a crashed kernel's leases are invalid at the next boot and its CLIs are reaped as before, keeping one writer per transcript.

**Identity is pid plus start time, never pid alone**: `/proc/<pid>/stat` field 22 on Linux, `ps -o lstart=` elsewhere (`proc_start`).

**The scope sweep** spares an owned CLI's scope. **The interrupt escalation** (`find_session_cli`) signals the leased CLI first when its identity still matches, so a re-parented CLI is stoppable. **The drain** drops the lease of any CLI it reaps.

**Every anomaly is a problem row**: prose on the error center's ring, the same prose plus one JSON object on the kernel log line, one JSON line in `STATE/session-events.jsonl`. Kinds: `lease.cli-without-lease`, `lease.no-live-process`, `lease.holder-gone`, `lease.stale-heartbeat`, `lease.version-skew`. Two CLIs on one conversation is the boot sweep's own row (`reconcile.duplicate-cli`), filed once. The row shape and the `problem_row` helper come from the restart monitors' work (T304, branch `romp_metrics-t304`); the lease code carries no helper of its own. The CLI takes no lock on a transcript it resumes (the stage 3 probe), so the one writer per conversation is entirely the lease's to keep.

## Merge order

This layer was stacked on the restart monitors' work (T304, #1340) while that was open; #1340 has merged, so this pull request now targets `main` and its four commits are rebased onto it. That work owns the `problem_row` helper, the `session-events.jsonl` ledger and `_log`'s `ring_text` seam; the lease code carries no helper of its own. One of its tests (`test_session_events_ledger.py`, the boot reconcile rows) gains the `lease.cli-without-lease` row its live-kernel-child fixture now earns. Nothing here changes the reap signalling itself (`_end_cli_tree`) or the parentage rule when no leases exist: `find_orphan_clis` without `leases` is byte-for-byte the previous verdict.

## Review fixes (fourth commit)

The heartbeat thread starts under the lock (two sessions connecting at once could both adopt one unstarted thread and start it twice); a beat re-checks membership and writes under the lock (it could rewrite a lease a concurrent close had just removed and file a false row at the next boot); the real-process test mints its sids per run (two checkouts running it at once would reap each other's fake CLIs). Each with a pin.

## Tests

- `tests/test_sdk_lifecycle_hardening.py::LeaseRules`: every census rule as pure fixtures (valid lease owned whatever the parent, holder gone by absence and by pid reuse, stale heartbeat at the TTL boundary, pid reuse on the CLI, own child, a live kernel's unleased child kept and reported, version skew, two CLIs on one conversation), `lease_state`'s ordering, `find_session_cli` with a lease, the lease file round trip, `proc_start` on a live pid and a fake one above pid_max plus the no-procfs path, the problem row's three writes, the backend's lease open, beat and close with a fake transport, source pins for the connect write and the close drop.
- `tests/test_cut_turn_tree_kill.py`: the boot reconcile with a leased re-parented CLI spared (scope too), a stale one reaped with its scope and lease, a dead lease dropped, the rows filed; and real processes on Linux, re-parented through an exiting intermediate, one surviving the boot by its lease and one reaped, the escalation reaching the survivor through its lease.
- Hermetic: per-test state dirs, synthetic ids, fake pids above pid_max, the cgroup seam pinned empty, every process the tests start killed by them.

## Docs

`docs/reference.md`, "What survives a restart": the lease, the reaper's rule and the problem rows.

Tier: feature

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -971,10 +971,12 @@ its CLIs are reaped as before, keeping one writer per transcript. The scope swee
 spares an owned CLI's scope, and the interrupt escalation signals the leased CLI
 first, so a re-parented CLI is still stoppable. Every anomaly the census meets (a
 CLI without a lease, a lease without a live process or holder, a stale
-heartbeat, two CLIs on one conversation, a lease from another code version) is
-a problem row: prose in the error center, the same prose with a JSON object on
-the kernel log line, and one JSON line in `session-events.jsonl` under the state
-directory, the shape the restart monitors read.
+heartbeat, a lease from another code version) is a problem row: prose in the
+error center, the same prose with a JSON object on the kernel log line, and one
+JSON line in `session-events.jsonl` under the state directory, the shape the
+restart monitors read. Two CLIs on one conversation is the boot sweep's own row
+there. The CLI takes no lock on a transcript it resumes, so the one writer per
+conversation is entirely the lease's to keep.
 
 A message the kernel cannot handle does not end the session's CLI. The kernel
 handles each streamed message on its own: when a handler raises, it logs the
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -951,6 +951,31 @@ re-checks the two facts that need no model call (the second initialize, the
 `--bg` refusal) when run with `ROMP_CLI_PROBE_LIVE=1` and a `claude` on PATH; it
 skips otherwise, as every test that would reach the live CLI must.
 
+Who owns a running CLI is a lease, not its parent process. The kernel writes
+`leases/<sid>.json` under the state directory the moment the SDK connect hands
+it a CLI: the CLI's pid and start time, the kernel's own pid and start time as
+the holder, the kernel's code version, and a heartbeat the kernel refreshes
+every three seconds while the CLI runs; the lease holds for twelve seconds past
+its last beat (the deploy drain hold's cadence: four beats, so it outlives a
+missed beat and not a dead holder). The lease is removed when the CLI's client
+closes, so only a kernel death leaves one behind. A lease is valid when its beat
+is fresh, its holder is alive and its CLI is alive, each identified by pid and
+start time together, never pid alone. The boot reaper reads the leases: a CLI
+with a valid lease is owned by its holder whatever its parent, so a CLI
+re-parented by a wrapper or a debugger (and, later, one kept by a per-session
+host) survives the boot; a CLI parented to a live kernel without a lease is kept
+and reported, so the sessions of a kernel from before leases survive the upgrade
+boot; every other CLI of ours is an orphan and is ended with its tree. Since the
+kernel is the holder, a crashed kernel's leases are invalid at the next boot and
+its CLIs are reaped as before, keeping one writer per transcript. The scope sweep
+spares an owned CLI's scope, and the interrupt escalation signals the leased CLI
+first, so a re-parented CLI is still stoppable. Every anomaly the census meets (a
+CLI without a lease, a lease without a live process or holder, a stale
+heartbeat, two CLIs on one conversation, a lease from another code version) is
+a problem row: prose in the error center, the same prose with a JSON object on
+the kernel log line, and one JSON line in `session-events.jsonl` under the state
+directory, the shape the restart monitors read.
+
 A message the kernel cannot handle does not end the session's CLI. The kernel
 handles each streamed message on its own: when a handler raises, it logs the
 exception type and the failing frame (file, line and function, first on the line
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -14750,7 +14750,8 @@ def _sdk_locked():
                 # would sort under a row the previous kernel filed in that same second, and the tail would read
                 # the old kernel's last state as current above the restart row. /version's `started` is the
                 # same start in whole seconds.
-                boot_at=_STARTED)
+                boot_at=_STARTED,
+                code_version=_kernel_sha())   # stamped on every session lease this kernel writes (T305)
             # a limit-shaped judge error envelope pokes ONE exact usage poll (get_usage rides turn
             # ends, so an idle fleet's usage.json goes stale — measured ~15h — and the rate gate is
             # only as good as that file); the backend picks any live login session to ask
--- a/kernel/sdk_backend.py
+++ b/kernel/sdk_backend.py
@@ -8607,15 +8607,16 @@ class SdkBackend:
                  "spawnedAt": int(now), "t": now}
         with self._lock:
             self._leases[str(sess.sid)] = (sess, lease)
-            need = self._lease_thread is None or not self._lease_thread.is_alive()
-            if need:
+            if self._lease_thread is None or not self._lease_thread.is_alive():
+                # started UNDER the lock: an unstarted thread reads as not alive, so a second session
+                # connecting in the first's write window would otherwise adopt it too and both would call
+                # start() on one Thread (RuntimeError out of the connect; the review of T305, 2026-09-10)
                 self._lease_thread = threading.Thread(target=self._lease_beat_loop, name="sdk-lease-beat", daemon=True)
+                self._lease_thread.start()
         try:
             write_lease(self.state_dir, lease)
         except Exception as e:
             self._log("lease (%s): write failed: %s" % (sess.name, e), problem=True)
-        if need:
-            self._lease_thread.start()
 
     def _lease_close(self, sess) -> None:
         """The CLI is gone or being ended on purpose (the session's client closed, a drain reap): drop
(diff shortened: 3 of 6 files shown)


henrypinkard:
Closing and reopening to trigger the CI workflow on the rebased head (the head push after the base refresh produced only the tier run).

ksarma:
Advisory review (the owner merges): no objection; 6 items below. All six are follow-ups; none is a blocker. The first is the #1317 design review's stale-beat point checked against the code.

Reviewed inline: the lease delta only (e6dd81b1..158aabd1). The #1340 ledger commits underneath were reviewed separately. The review was by reading, because the code signals processes and stops scopes. The head moved to c1085485 during the review. With hunk headers stripped, the two diffs have zero differing lines and the same six files and counts. Everything here applies to the current head. The lease is written right after the SDK connect. It is dropped in the connect block's finally and at the drain reap. `lease_state` checks CLI identity, then holder identity, then the beat. The census owns a valid lease whatever the parent. It owns this kernel's direct children silently (the spawn-to-write window). It keeps and reports a live kernel's unleased child. It reaps the rest with their trees before the resume loop runs. In the manager's flow the holder is dead at every boot, so one writer per transcript holds. `find_orphan_clis` without leases is unchanged. The start-time reader splits `/proc/<pid>/stat` after the last `)`, so a comm with spaces or parentheses parses. `proc_start` falls back to `ps -o lstart=` without procfs. The five `lease.*` kinds reach the ring, the kernel log and the ledger through `problem_row`. The docs paragraph and the body match the code, and the feature label fits. The body does not address the #1317 review's stale-beat should-fix. That review was posted 19:20Z; the lease commits were authored earlier.

Follow-ups:
- `kernel/sdk_backend.py:3372` and `3437` (pinned by `tests/test_sdk_lifecycle_hardening.py:252`): a lease whose CLI and holder are both alive by pid and start time can still be reaped. That happens when its beat is over 12 s old and the CLI is not a kernel's child. That is a time threshold. The event it approximates, holder identity, is checked two lines earlier. A stalled beat thread, a suspended or swapping holder, or a clock jump then kills a live turn. The boot resumes a second writer. The rule is unreachable in the manager's stage-1 flow: the holder is dead at boot, and a live kernel's direct child is kept. Stage 4's host inherits it. Suggest owning the CLI and filing the `lease.stale-heartbeat` row instead of reaping. The test's first assertion flips.
- `kernel/sdk_backend.py:8350-8358`: `_lease_open` decides `need` under the lock but calls `start()` after releasing it. `is_alive()` is `False` on an unstarted Thread. Two connects landing while no beat thread is alive (the boot resumes three at once) can both take `need=True`. The second `start()` on the same Thread raises RuntimeError inside the connect block after `connected=True`. That session's connect dies and its fresh CLI is closed. Create and start the thread under the lock.
- `kernel/sdk_backend.py:8367-8380` and `8360-8365`: the beat snapshots the leases under the lock and writes them without it. `_lease_close` pops under the lock and unlinks outside it. A close landing between the snapshot and that lease's write recreates the file. The next boot then files a false `lease.no-live-process` row. This is most likely after a drain, which closes every session inside one beat cycle. Hold one lock across the write, and one across the pop plus unlink.
- `kernel/sdk_backend.py:3404-3405` and `3442-3447`: the census admits a process to `clis` only when its argv carries a reg's `lastSid`. After a `/clear`, or a resume that lands on a new conversation id inside the same process, the argv names the old id. Lines 6331-6335 adopt the new fsid, and the beat refreshes `lease["fsid"]`. The CLI falls outside `clis`, so `lease_state` is never consulted. The census files `no-live-process` (the pid is alive with the lease's start time) and drops the lease without reaping the CLI. The missed reap predates this PR; the false row and the discarded identity are new. Judge a lease outside `clis` by `lease_state` when its identity still matches a live SDK process.
- `kernel/sdk_backend.py:3569-3579`: the lease branch of `find_session_cli` never reads `lease["holder"]`. The docstring still says it may only signal our own child. With two kernels on one state root (an ensure without a stateDir, or a hand-run kernel), both beat the same file. The escalation or the drain can then SIGINT or SIGKILL the other kernel's CLI. Require the holder to be this kernel's pid and start. The fixture at `tests/test_sdk_lifecycle_hardening.py:310-314` (holder 777, kernel 31337) then needs `holder=OWN`.
- `kernel/sdk_backend.py:8382`: `_lease_beat_loop` is never executed by a test (patched to a no-op at `tests/test_sdk_lifecycle_hardening.py:380`). It calls `time.sleep` directly. Its contract (a beat per lease per interval, exit when none is held, restart at the next open) therefore has no test. `_end_cli_tree` in the same file takes `sleep` and `now` seams; the loop could too.

Two smaller points:
- `docs/reference.md:817-818` and the body say a pre-lease kernel's sessions survive the upgrade boot. Under the manager that kernel is dead at the boot, so its CLIs are reaped and resumed as before. The kept-and-reported rule fires only for a still-running kernel.
- `docs/pr-tiers.md:28` says every layer of a stack must live in this repository. This layer's head is on the fork. CI ran at both heads, so nothing broke.

Evidence: nothing from the PR ran here (read-only, by the reaper rule). `py_compile` of the five changed Python files passed. The 737 added lines carry no private string, hostname or home path, and no banned vocabulary. CI at 158aabd1 was all green (each Python job reports 11379 passed, 136 skipped). At c1085485, CI, PR tier and Tier policy completed green as of 20:17Z. The PR targets the stack branch. The head that reaches main after #1340 lands is new and needs a fresh look.

Review run by the maintainer automated review session; verified against head 158aabd1.

ksarma:
Post-merge record: the owner merged this PR as e02f99f2 at head bc6d2929 while this review ran; no objection; 1 item below. Since the advisory posted at 158aabd1, the head has moved to bc6d2929. It is the same lease commits rebased onto main after #1340 merged, plus one fix commit reviewed here.

Reviewed inline: the fix commit only (751f2efc..bc6d2929), by reading, because the lease code stops scopes and signals processes. `_lease_open` now creates and starts the heartbeat Thread inside the lock. `Thread.start()` returns only once the thread has signaled started. A second connect in the first's write window therefore sees a live thread and starts nothing. `_lease_beat_once` re-checks each lease's membership by object identity and runs the whole `write_lease` under the same lock. `_lease_close` still unlinks outside the lock, and that suffices. Its pop is under the lock, so a beat's write can never follow a pop. A write that came first is removed by the close's own unlink. The lock is held for one small write per lease per 3 s beat. The drain releases it before calling `_lease_close`. Connects therefore wait microseconds, and the non-reentrant Lock is never nested. The two new tests run the real `_lease_open` and `_lease_beat_once` with only `write_lease` and the loop patched. By reading, both fail at 751f2efc and pass at bc6d2929 in either thread order. The real-process test's per-run `uuid.uuid4()` sids are synthetic, and its cgroup seam is still pinned empty. Items (2) and (3) of the earlier advisory are resolved. Items (1), (4) and (5) stand unchanged at `kernel/sdk_backend.py:3611`, `3643` with `3696`, and `3809`. Item (6) is below.

Follow-up:
- `kernel/sdk_backend.py:8651`: no test executes `_lease_beat_loop`; every lease test patches it away (`tests/test_sdk_lifecycle_hardening.py:380`, `:425`, `:442`). A regression in its exit handshake (the re-check of `_leases` before clearing `_lease_thread`) would pass the suite. Every held lease would then stop beating and read stale after 12 s. The loop also calls `time.sleep` directly. `_end_cli_tree` takes `sleep` and `now` seams, and the loop could take the same. One test could then step a beat, an empty exit and a restart at the next open.

Two nits, neither introduced by this commit:
- `kernel/sdk_backend.py:8615`: `Thread.start()` is unguarded. If the OS refuses a thread, it raises into the connect after `connected` is set. The finally then drops the lease, and the session crash-heals a working CLI. The docstring's rule for a missing pid, run unleased and file a problem row, fits this failure too.
- The commit message and body say each of the three fixes has a pin. The diff adds two tests, and the per-run sids have none. Wording only.

Evidence: nothing from the PR ran here. `compile()` of the three changed files passed. The 85 added lines carry no private string, hostname, home path or banned word. CI at bc6d2929 is all green: Python 3.10 to 3.13, bats, gitleaks, the extension, the tier label and Tier policy checks.

Review run by the maintainer automated review session; verified against head bc6d2929.

PR #1343, opened 2026-09-10
The host-down notice's swirl spins only while romp is dialing the host, keyed on the kernel's tunnel row and the relay socket, never a timer (T309)

henrypinkard:
Design points (feature tier: a self-contained capability inside romp's existing model).

The host-down notice at a disconnected remote's transcript tail ("<host> is disconnected — this is the last romp got from it. Reconnecting.") now carries the romp swirl after "Reconnecting", spinning only while romp is actually trying to reach the host, still between attempts. The sentence is unchanged; the swirl sits after it so the words read first; the notice resolves as before when the host returns.

What "trying right now" means, and why the kernel says it: the browser never dials a host the kernel reports down (federation connects a relay socket only while the /tunnels row reads up, and the notice shows exactly when it does not), so a spinner keyed on the relay socket alone would never spin while the notice is on screen. The honest signal is the kernel's own retry. The tunnel supervisor already marks a row `starting` while an ssh dial is spawned and not yet healthy and `down` while it backs off until `nextTry`; for a checked-in peer the attempt is its health poll of the peer's port. So:
- `kernel/kernel.py`: the /tunnels row gains `dialing` (`_row_dialing`): true while the status is `starting` (or a checked-in row's birth state `connecting`) or while the pass's health request to the host is in flight (a per-pass mark set around the request, never saved to the remotes file). The four supervisor round-trips to a silent host take their timeouts, so the mark is on for seconds per pass and off between passes; against a refused port it is on for milliseconds.
- `ui/webview/federation.ts`: the same /tunnels poll that publishes the down set now publishes the dialing set and fires `romp:hostDial` only when it changes; the relay socket's own dial, open and close fire the same event (three transitions, no timer anywhere). `__rompFed.dialing(host)` answers from both states.
- `ui/webview/host-prefix.ts`: `hostDialLive(rowDialing, readyState)` is the pure rule (the kernel's word, or the socket in CONNECTING), and `hostIsDialing(sid)` its reader, false with no manager, for a local session and for a manager too old to publish the state.
- `ui/webview/render.ts`: the foot rebuilds on `romp:hostDial`, keyed on the state, appends the swirl inside the gist after the text only when dialing, and passes `live` for the same state.
- `ui/webview/styles.css`: the swirl spins with the placeholder tab's motif (same keyframes, 12 px) and rests under reduced motion.

Tests: `ui/webview/host-dial-live.test.ts` executes the pure rule and its reader against a stub manager and pins the wiring at source (federation's publish and its four dispatch sites, the foot's order of words then swirl, the CSS); `tests/test_kernel_tunnel_truth.py` pins `_row_dialing`, the unsaved mark and where the pass sets and clears it; two existing pins re-anchored (the single-instance import line now names the third pure helper; the host-offline notice's gist is an element). Suite runs stayed behind the machine's test lock in capped scopes: the extension suite, the tunnel-row Python tests, and the focused re-run of the re-anchored pins.

Seen by eye from a hermetic served page with a fake checked-in host (a checkin_peer row pointing at a stub server that answers the kernel's polls and pushes one `api` session over the relay): pausing the stub makes the kernel's health request hang to its timeout, the row reads `no-kernel` with `dialing` true and the notice shows the swirl (`~/.local/state/romp/drops/romp_feed-reconnecting-spinning.png`, with a tight crop `-spinning-foot.png`); killing the stub makes the port refuse at once and the notice sits still (`-still.png`, `-still-foot.png`). Same sentence, same place, in both.

### DIFF
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -18911,6 +18911,35 @@ def _expected_restart_status(r, st, rsha, now):
     return st
 
 
+def _row_dialing(r):
+    """Is romp trying to reach this host RIGHT NOW? True while an ssh dial is spawned and not yet confirmed
+    (status "starting"; "connecting" is the checked-in row's birth state) or while the supervisor's health
+    request to the host is in flight (_dialing, set around the poll below). False while the row waits out
+    its backoff (status "down" until nextTry). The dashboard's host-down notice spins
+    its swirl on exactly this (the user 2026-09-10, who wanted a spinner that means romp is trying right
+    now, never one that spins whatever happens); federation reads it from the /tunnels row every few
+    seconds and repaints on a change. The mark is on for ANY polled row during its pass's requests, an up
+    row included (milliseconds against an answering host); the notice only ever shows for a down one."""
+    return bool(r.get("_dialing")) or (r.get("status") or "") in ("starting", "connecting")
+
+
+@contextlib.contextmanager
+def _dialing_mark(r):
+    """The row wears the in-flight mark (_dialing, read by _row_dialing) for the supervisor pass's health
+    requests to its host: on from the first round-trip until the last returns, HOWEVER the block ends. The
+    clear is the context manager's exit, a finally by construction, because a mark left on by an exception
+    (a socket the port check could not build) would read as dialing through the row's whole next backoff, the
+    one state the mark exists to distinguish (review find, 2026-09-10). The polls themselves stay inline in
+    _tunnel_supervisor, whose source a dozen tests pin line by line."""
+    with _remotes_lock:
+        r["_dialing"] = True
+    try:
+        yield
+    finally:
+        with _remotes_lock:
+            r["_dialing"] = False
+
+
 def _remote_public(r):
     """The API view of a remote row — everything the browser needs, minus the Popen and minus the remote's
     credential. The browser reaches a remote through /remote/<host>/ws, where _remote_ws injects that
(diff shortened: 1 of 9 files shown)


ksarma:
Advisory review (the owner merges): three findings below. Recommend merging as shipped. The one item of substance is a follow-up test.

Reviewed inline: the /tunnels row's `dialing` comes from `_row_dialing`, true on `starting`, `connecting`, or the per-pass `_dialing` mark. That mark is in `_NOT_SAVED`, so it never reaches the remotes file. federation.ts builds the dialing set from the same /tunnels poll as the down set. It fires `romp:hostDial` only when size or membership changes, and dispatches the same event when the relay socket is constructed, opens, and closes. `__rompFed.dialing` delegates to `hostDialLive`, true only when the kernel reports dialing or readyState is 0. `hostIsDialing` reads false for a local sid, with no manager, and against a manager without the accessor. `syncHostOfflineFoot` keys its rebuild on the sentence plus the dialing bit, so a repaint with nothing changed leaves the DOM alone. It appends the 12 px swirl after the unchanged sentence only while dialing, with `live` set for the same state. The CSS reuses the placeholder tab's keyframes and does not animate under reduced motion. There is no timer anywhere. The feature label fits, and the body matches the diff clause by clause. The added lines carry no private strings, hostnames, home paths, or ruled-out vocabulary.

- `ui/webview/host-dial-live.test.ts:57`, follow-up: federation's publish (the set from the poll, the change-only event, the three socket dispatches) is pinned by source regexes only. Two mutations tried here kept every regex matching while the event never fired from the poll: `this.dialingHosts = dialing` hoisted above `dialChanged`, and an early return before the publish block. A poll() case in federation-reconnect.test.ts's existing harness would execute the path in a few lines: a `dialing: true` row twice, counting `romp:hostDial`, plus openRemote with ws.open and die.
- `kernel/kernel.py:22387`, nit: the `_dialing` set and clear are two lock blocks with no try/finally around the five calls between them. Every path between them swallows its exceptions today (the one raise is a non-numeric `local_port` in `_port_open`). A raise would leave the row reading `dialing: true` through its backoff until the next pass.
- `kernel/kernel.py:18879`, nit: the mark is set for every non-skipped row, so an `up` host reads `dialing: true` during its routine health round-trips. The docstring says that never happens. The foot is unaffected (it shows only for a down host), but a /tunnels poll landing in that window fires `romp:hostDial` with no real change. Gate the mark on status, or amend the docstring.

Run here at the head in capped scopes: typecheck, the test build, the 54 compiled ui/webview tests that import host-prefix or federation (651/651), and pytest tests/test_kernel_tunnel_truth.py (34/34). For the record (not required for a feature): at the merge base with the PR's test files checked out, the Python test fails on the missing `_row_dialing`, the UI test build fails on the missing exports, and the two re-anchored pins fail on their new regexes. CI at bccf5dac: all 12 checks green, Tier policy included.

Review run by the maintainer automated review session; verified against head bccf5dac.

PR #1345, opened 2026-09-10
Comment threads: one solid outline in the notch's yellow around the whole unread passage

henrypinkard:
An unread comment thread (a landed reply the user has not opened) wore a dashed outline on its highlight mark. An outline on an inline mark paints once per line fragment, so a wrapped passage read as a stack of dashed boxes, one per line. The user's ruling (2026-09-10): ONE outline around the WHOLE highlighted area, in the same yellow the scroll rail's notch uses for that thread, never a per-line treatment.

## Design

- **One box per unread thread, drawn from the marks.** CSS cannot merge a mark's line fragments (`box-decoration-break: clone` keeps them apart), so the cue is an absolutely positioned child of the anchor turn, `.cmt-outline`, sized by `paintCommentOutlines` in render.ts to the union of the thread's mark fragments' client rects. Turn-relative coordinates: the box scrolls with the text and needs no repaint on scroll.
- **Nothing moves, nothing tints.** An outline on an empty box (nothing reflows when it comes and goes), 1px clear of the glyphs, `pointer-events: none` so hover and click land on the marks beneath. The fill tiers are untouched: 45% yellow still says landed, the green wash still says a reply is being written.
- **The notch's colour, exactly.** Solid 1.5px in `var(--cmt-hl)`, the token `.cmt-tick` wears, so box and notch agree in both themes.
- **Event-driven repaint, no timers.** After every marks pass (each transcript rebuild and comments frame), on the rail's rAF scheduler (a re-render, the view's resize observer) and on window resize, writing only what changed. The box goes with the unread bit (the popover opening, a resolve) and with the marks (a windowed-out turn, a deleted thread).

## Tests

- `ui/webview/comment-outline.test.ts`: the rule (solid, the tick's token, positioned, pointer-transparent, no hex), the painter's shape (unread marks only, grouped by thread, union of `getClientRects`, a child of the turn, stale boxes removed, hidden view keeps none) and its three hooks.
- `ui/webview/comment-mark.test.ts`: no mark rule carries an outline any more; the fill tiers and the unread bit's one clearer stay pinned.
- `tests/test_comment_outline_served.py`: the real /chat page of a hermetic kernel with a two-line and a five-line unread passage. Exactly one box per thread covering every fragment, hugging the union; outline colour byte-equal to the thread's scroll tick in the dark and the light theme; no fragment with an outline of its own; opening one thread removes its box on the same pass and leaves the other's. With `COMMENT_SHOTS=<dir>` it writes the dark and light screenshots.

Tier: feature

### DIFF
--- a/tests/test_comment_outline_served.py
+++ b/tests/test_comment_outline_served.py
@@ -3,10 +3,13 @@
 scroll notch's yellow around the WHOLE highlighted passage — never a dashed box per line.
 
 A hermetic kernel serves a synthetic chat (the notes-api demo world: session web) whose one answer carries two open
-comment threads with a landed, unseen reply each: a passage that wraps over two lines and one over five. Asserted on
+comment threads with a landed, unseen reply each: a passage that wraps over two lines and one over five
+(on a 640 px page; the kernel ships a thread's exact text capped at 500 characters, so the mark covers at most that),
+plus an incoming question card whose long body scrolls, with a third thread on its last sentence: that box exists only
+while its passage shows inside the body (cut to the scrolling container, repainted when the body scrolls). Asserted on
 the page: every unread thread has EXACTLY one outline box, positioned (the text under it never moves) and transparent
 to the pointer, whose rect covers every line fragment of the thread's marks; the box's outline is solid and its
-colour is byte-equal to the thread's scroll tick, in the dark theme and the light one; no mark fragment wears an
+colour is byte-equal to the thread's scroll tick and reads at 3:1 or better against the page, in the dark theme and the light one; no mark fragment wears an
 outline of its own; opening a thread (a click on its mark) removes ITS box on the same pass and leaves the other's.
 With COMMENT_SHOTS=<dir> the driver also writes a dark and a light screenshot. Skips LOUDLY without the extension
 deps or a Playwright browser. SYNTHETIC fixtures only (placeholder UUIDs, invented prose, host TESTHOST)."""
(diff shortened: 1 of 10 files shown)


ksarma:
Advisory review (the owner merges): two items below. Neither blocks, and one is already fixed at the current head.

Head note: the head moved to 8e476f85 (two review-fix commits) while this review ran; the text below covers a160acbe and read the newer commits without re-running them. The two fix commits are reviewed separately; a delta comment follows.

Reviewed inline: the new `.cmt-outline` box is an absolute child of the anchor `.turn`. That turn is `position: relative`, with no border on any variant and no transformed ancestor. Subtracting the turn's client rect from the union of the unread marks' `getClientRects` therefore places the box on the glyphs. A headless-page probe reproduced one box sitting exactly 1px outside a four-line mark on every side. The painter runs at both exits of `applyCommentMarks`, in the rail's rAF callback, and on window resize, with no timer. The view's ResizeObserver and `showActive` repaint a re-shown view. A stale box (thread read, marks moved to another turn, no laid-out fragment) is removed on the next pass. The view's MutationObserver watches childList only, and `#cmt-pop` mounts on `document.body`, so the box never feeds back into the painter. The fill tiers and the `.busy` wash are byte-for-byte unchanged. The rule uses the tick's token with no hex literal. Privacy and vocabulary scans of the 416 added lines found nothing. The `feature` label fits a UI change inside the existing comments model.

- `ui/webview/comment-outline.test.ts:30` (follow-up): every CI-run assertion about the painter is a regex over `render.ts`. The turn-relative subtraction is not among them. Changing it to viewport coordinates passes all 102 node tests. The served test would catch it, but it skips in CI, where the Python job has no playwright. Factoring the union-and-offset arithmetic into a pure function with fixed-rect tests would close this, as `reply-ready.ts` does for `placeMark`. The skip pattern is repo-wide, not this PR's.
- `tests/test_comment_outline_served.py:192` (follow-up, fixed at 8e476f85): the lab kernel's environment was `dict(os.environ)` minus three names, against the `kernel_env` convention in `tests/README.md`. The current head uses `_lab.kernel_env` with `ROMP_HOST_NAME` set. The PR body still names `var(--cmt-hl)` as the box's colour and lists three test files. The head draws box and tick in `--cmt-hl-outline` (amber in the light theme) and also re-pins `comments.test.ts` and `reply-ready.test.ts`.

Locally at a160acbe: typecheck clean, esbuild test build clean. `node --test` on the eight compiled UI files that reference the painter or its rule: 144 pass, 0 fail. At the merge base with the PR's four test files checked out, 7 of 102 fail (the new pins). The served test was read, not run. CI at a160acbe: all ten checks green. The PR has since moved two commits to 8e476f85. Those changes were read but not re-run here, and their CI is still running.

Review run by the maintainer automated review session; verified against head a160acbe.

ksarma:
Advisory review (the owner merges): three items below. One is a blocker at 8e476f85 that the next commit, 244909f9, already fixes; that commit was read, not run. The other two are follow-ups from the base review that the fix commits changed. Delta over the advisory posted at a160acbe: the head is now 8e476f85, and its two review-fix commits are reviewed here.

Reviewed inline: `paintCommentOutlines` cuts each mark fragment to the bounding rects of every scrolling ancestor between the mark and the turn. It drops a fragment the cut empties and removes the box when no fragment remains. So a passage scrolled out of a notice body has no box, and a partly visible one gets a box cut to the container. The pane's scroll listener is registered once on `#content` at module init, passive and in the capture phase. A scroll inside a notice body or a wide formula therefore reaches the same rAF scheduler; nothing is attached per rebuild. The rAF and resize hooks call the painter only while `hasUnreadOpenThread` is true, a store read. The marks pass stays ungated and remains the removal path. Box, tick and unread ring share the new `--cmt-hl-outline` token, with a 3:1 theme-parity pin. It is the fill yellow in the dark theme and amber in the light theme. Amber measures 4.2:1 on the cream page, where the fill yellow measures 1.1:1. The served test boots the lab kernel through `_lab.kernel_env` with `ROMP_HOST_NAME` set, which closes the base review's second follow-up. Privacy and vocabulary scans of the 179 added lines found nothing. The `feature` label still fits.

- **Blocker at 8e476f85, fixed at 244909f9 pending its CI.** `ui/webview/rail-sticky-stamp.test.ts:53` and `ui/webview/scroll-marks.test.ts:209` both pin source lines ff1ea6c1 changed: the listener's `{ passive: true }` and the tick's `var(--cmt-hl)`. So the extension check fails 2 of 4284 tests at 8e476f85. 244909f9 rewrites those two regexes, and they match the head's source.
- **PR body (follow-up; the base review's third item, now wider).** It still names `var(--cmt-hl)` and says the box needs no repaint on scroll. It does not say the light-theme scroll tick changed from the fill yellow to amber. That is a visible change to an existing element, stated only in ff1ea6c1's message. The Tests list omits five changed test files and the served test's scroll and contrast scenarios. The `render.ts:8399` docblock and the `styles.css:3326` comment carry the same two stale sentences.
- **`ui/webview/comment-outline.test.ts:39` (follow-up; the base review's first item, now wider).** The clip intersection has no pin that runs in CI. Replacing it with the unclipped rect passes typecheck and every node test. Only the served test, which skips in CI, would catch it. The contrast pair is executed by theme-parity.

Locally at 8e476f85: typecheck clean, test build clean. The PR's three changed node files: 83 pass, 0 fail; at a160acbe, 6 of 83 fail (the new pins). The two pins above fail locally as in CI. The served test was read, not run. CI at 8e476f85: the extension check is red on those two pins. Tier checks, gitleaks and Python 3.11 are green; the push cancelled the rest. At 244909f9 the extension check is queued.

Review run by the maintainer automated review session; verified against head 8e476f85.

PR #1348, opened 2026-09-10
Chat: a queued message is edited in place, and the edit holds the message until it is done (T306)

henrypinkard:
Editing a QUEUED message happens in place, inside its provisional bubble, and the message holds until the edit is done (T306).

**In place.** The ✎ turns the bubble's text into a field where it sits: the same width, the dashed provisional look kept, the ✕ still in its corner, Save and Cancel under the text. Enter saves, Shift+Enter breaks a line, Escape cancels. Save posts the existing `editQueued` with the new words and the kernel replaces the entry in place (same slot, a follow-up's wrapper kept); Cancel leaves it as it was. The composer is never touched: the earlier design pulled the text into the composer under an editing pill, and a message that went meanwhile left its words there as a stray context chip.

**Hold, event-based.** Opening the field posts `holdQueued`, a new drive op beside `editQueued` with the same three arms (parked slot, backend index, body only), owned by the connection that opened it (every ws client now carries a `cid`). The SDK backend's feeder pops the first UNHELD copy and the parked walk takes the first unheld op, so a held entry keeps its slot while the queue behind it keeps moving; the in-flight guard finds the op the backend holds by scanning for it instead of assuming slot 0. The hold ends with the Save (the edit itself releases it), the Cancel (`holdQueued` with `hold:false`), or the client's socket closing (the ws handler's `finally` hands the client to `_release_client_holds`: the disconnect is the event, no timer). A hold on an entry no queue holds any more (fed already) is refused through the existing `editResult` ok:false path with the too-late text: the field closes and the bubble says so. Every client sees a held copy as "editing": the queued event carries the mark from both queues.

**Design points.**
- The editor's state is keyed by the entry (session, body, the copy's id and stamp) and lives in memory: the tail rebuilds on every push and repaints the field from the state, with its words, caret and focus, so nothing is lost to the rebuild; a fresh page has no edit in progress, and the kernel released the hold with the old socket.
- Save and Cancel are delegated through the one body delegate like the ✕ and the ✎ (click-safe across re-renders); the field acknowledges the click by appearing.
- A refused Save reverses the optimistic repaint and hands the typed words back in a sticky toast, never the composer; a refused Save while the host is down keeps the field open with the words and asks for a re-dial.
- Holds are in memory only; a kernel restart or a dropped link drops them, and the page re-holds every open field on the socket-up frame (and on a relay coming back, for remote sessions); a copy that fed meanwhile answers the refusal and the field closes with the typed words in a toast. The first editor keeps a copy: another connection's hold or Save on it is refused with "another client is editing this message". A Save the kernel refuses releases the saver's own hold. A tab close releases the holds it drops (the session and the socket live on).
- Only messages pass a message being edited: a later message queued behind a held one goes on (the held one keeps its slot), but a command, a compaction or a setting parked behind it waits, so the shown order stays the run order for everything but messages. This is the one place the design changes what runs when; it is stated on the held bubble's title and here for the user's call.
- Sizes and colours already on the surface: the buttons and labels wear the queued header's 0.82em; the field inherits the bubble's type; the accent ring marks an editing bubble.

**Review** (lean, 16 agents: four lenses over the working tree, two skeptics per finding, read-only under the machine's load rules). Five confirmed, all fixed: the editor was keyed by the copy's stamp, which the ✎ carries only for the page's own optimistic copies, so a kernel-rendered copy (after a reload, in a second pane, from the CLI) was held but never showed a field (now keyed by the copy's id, else the stamp); the pending group's cached node ignored the editor, so the ✎ on the user's own just-sent message painted nothing; no client re-held after a socket reconnect although the kernel comments promised it (now on the socket-up and relay-up events); the last two were duplicates. Of the unverified items, folded: a hold's acknowledgement consumed the Save's restore stash; the blur the rebuild itself fires cleared the focus flag so focus was lost on every push (now ignored when the field has left the DOM); a keystroke in the rebuild gap could land in the composer (type-from-anywhere stands down while a field is open); a tab close left the hold in place; the ✕ on an editing bubble discarded the edit (it now hands back the edited words); two same-words id-less parked sends shared one hold key; the second client's silent takeover; the walk letting commands and settings run ahead of a held message; a refused Save leaving the hold; the field painting one row and growing a frame later; distinct refusal text for a session that is not running; the guide's paragraph. Not folded, noted: the sender's own bubble does not show "editing" when another client holds the copy (our optimistic copy hides the kernel's, which carries the mark).

**Second review** (the manager's, on the open pull request): five confirmed, all fixed in the third commit. A refused Save's words lived only in a fading toast: a copy still queued now reopens the field with the typed words and the refusal beside them (re-held), and a copy already gone puts them in a toast that never fades with a Copy button (the kernel's frame says which). Parked relocation by body with two id-less same-words sends and a drifted slot landed the edit on the unheld twin: it now prefers the twin an editor holds and refuses when nobody can tell them apart. A closed editor's "another client is editing" note never retired: it lasts as long as the held mark. The drain's drop arm left the parked holds behind: it pops them, and so does an emptied queue. Opening the field widened the content box by the ✎'s room and re-wrapped the words: the editing bubble keeps the idle padding. The kernel's frame says `gone` only when the copy is gone, so an ok frame keeps its pinned shape.

**Tests.** `tests/test_queued_edit_hold.py` executes the backend (a held copy is skipped and keeps its slot; Save, Cancel and disconnect release; a hold on a fed copy is refused; the persisted queue mirrors an edit) and the kernel (the three hold arms through `_drive`, the parked walk's skip, the in-flight slot guard, the disconnect release, the routing pins). `ui/webview/queued-edit.test.ts` pins the field, Save/Cancel/keys, the hold ops, the verdict handling, and that the editor's code paths never write the composer or its chips (and that the composer's chip strip knows nothing of queued messages). The reload-notices refusals lift the in-place Save. Extension suite and full Python suite green.

**Screenshots** (hermetic served chat page with a synthetic parked send; the user's eye): `~/.local/state/romp/drops/romp_timeline-queued-edit-idle.png`, `-editing.png`, `-saved.png`.


ksarma:
Advisory review (the owner merges): no objection. Below are three follow-ups and two smaller points. The first two follow-ups are small enough to fold before the merge.

Reviewed inline: the connection id is minted in `_new_ws_client` and never read from a payload. Another connection's hold, Save, Cancel or release on a held copy is therefore refused. The SDK feeder pops the first unheld copy, and the parked walk takes the first unheld op. The in-flight guard finds the op by identity, so a held entry keeps its slot while the queue behind it moves. A hold ends with Save, Cancel, the ✕ or the socket's finally. A hold on a copy that already fed is refused with the too-late text. A command, compaction or setting parked behind a held message waits until the editor closes or the socket drops. That wait is user-initiated and event-based, so `feature` fits; the label is yours. The editor state is keyed by the copy's id and survives rebuilds. Save and Cancel go through the body delegate. The accent appears only as the editing ring and the Save fill, and no new size or hex is added. The head moved to a85d8e1f after this review: a merge of main (#1340) into 0443d20c. Its tree equals the merge-tree of its parents, so there was no manual resolution. The PR's per-file diff against the new base is unchanged. The three sites below are the same there; kernel lines shift by 34 and 50.

- Follow-up, `kernel/kernel.py:15855-15860`: the body-only `holdQueued` arm keeps the FIFO's too-late text whenever `_hold_backend_queued` returns any text. So the sender's own bubble (its ✎ carries md and qid only) is told the message already reached the session. That happens when another connection holds the copy or the session is not running. The message is still queued at that point. The Save's body-only arm at 15828-15832 propagates the backend's text. Mirror it, and add a body-only test against another owner and a dormant session.
- Follow-up, `ui/webview/render.ts:14552`: Enter saves unconditionally. The composer gates Enter on `!isCoarsePointer()`, because a soft keyboard has no Shift+Enter. At the merge base the edit went through the composer, so on a phone this is a regression: the first Enter saves, and a line break cannot be typed. Add the guard (the Save button covers that case) and update the pin at `ui/webview/queued-edit.test.ts:59`.
- Follow-up, `kernel/kernel.py:28103` and `28111`: the two hold folds in `_chat_build_sig` have no test. With both removed, all 184 PR tests pass. The chat pusher serves the cached payload on a signature hit. Other clients would then not repaint "editing" and their hold would be refused on a bubble that showed nothing. A differential test on the census harness (template at `tests/test_chat_build_sig_inputs.py:617`) closes this.
- Two smaller points. The PR body says the ordering change is stated on the held bubble's title. The title at `render.ts:4324` says only that the held message goes when the edit is done. The user gets no hint that a parked setting behind it waits. The walk's exception path at `kernel.py:30706` pops the queue, the in-flight op and the drain hold, but leaves `_park_holds[sid]`, whose id-less keys are object addresses.

Locally at 0443d20c: the four PR Python files, 184 passed; typecheck clean; the six touched node test files, 77 of 77. At the merge base 304489cb with the PR's tests checked out: 25 Python failures (all 21 in the new module), and all six node files red. Three mutations of the hold paths were each caught: the feeder ignores the hold, `_parked_held` always false, a non-send passes a held send. CI: every check green at 0443d20c. At a85d8e1f, all green except Python 3.11, still running when this was written.

Review run by the maintainer automated review session; verified against head 0443d20c.

PR #1352, opened 2026-09-10
Pane shim: declare a redial only once the kernel's caps frame has answered the bundle's ready

ksarma:
Follows #1325, which has merged; this branch is cut from main, and the change to review is its two commits: the test pin and the caps latch that #1325's review record named as follow-ups.

## Symptom

A pane whose bundle posted its `ready` on an open socket that then died before the kernel processed it (a keepalive the only frame back) dialled the next socket with `&reconnect=1`. The kernel served a page that had never received its sessions as one that held them: the skeleton strip, the active tab in full and a status frame per tab, with the withheld tabs filling on click or in idle. #1325's accept-time note stated this residual, and its review named the caps frame as the event to close it on.

The review also named a gap in the tests: no case covered the bundle's `ready` posted while the FIRST socket was still connecting.

## Cause

The dial term keyed on the ready having LEFT (`everConnected&&bundleReady&&!readyQueued`); nothing in the shim recorded whether the kernel had processed it. The kernel already says so on an exact event: the `ready` arm pops the skeleton set, serves the page whole (`_push_one`), and only then sends the caps frame (`_send_caps`, called from that arm alone), so the frame is the kernel's word that the page was served.

## Fix

Commit 1 (tests only): one ReconnectFlag case posts the ready before any `open()` while the first socket is CONNECTING, reads the shim's queue bits by name (no dial runs while a socket is connecting, so no URL could show them), and asserts the flush once at the first open, the cleared bit, and the next redial's flag. It is red under each of three shim mutations that every existing case survives (the clear moved into the reconnect branch; the queue bit set only for a closed socket; a ready dropped while connecting) and under a deleted clear (also red at the existing queued-ready case). One docstring sentence names the order.

Commit 2: `_shim` gains `readyAcked`, set in `ws.onmessage` on a `caps` frame (placed after the parse and before the keepalive line, so the heartbeat, disconnect-banner and auto-reload source pins stand). The dial term becomes `everConnected&&bundleReady&&readyAcked&&!readyQueued`: a fourth conjunct rather than a replacement, so the term reads as the four events in order and a frame with no prior ready from this bundle arms nothing. In production the caps frame implies the other three (ready is posted once per renderer life and the kernel answers only a ready it processed), so the extra conjuncts bite only on injected frames; for the same reason a latch gated on `bundleReady` would behave identically in production, and no test distinguishes the two. If you prefer the single bit, drop the conjuncts and the four-events test case goes with them.

Cost, stated in the accept-time note, `_send_caps`'s docstring and the ReconnectFlag docstring: a caps frame that never arrives (the socket died between the pushes and the frame, the ready lost on a half-dead socket, or the ready arm raised into the dispatch except) leaves the page dialling fresh for its LIFE, not for one redial: the bundle posts ready once, so no later socket carries one and no caps frame follows. Every redial of such a page is served whole, the cost before 2026-09-07, never a false skeleton.

Comments this change made stale, corrected here: the KERNEL_WS_CAPS comment's "the shim's re-send on a reconnected socket" (the socket that re-sends `ready` at every open is the shell page's own, `shellWS`; a pane's shim never re-sends, and a reconnected pane socket carries a ready only when it queued across the drop); `_send_caps`'s stale-banner sentence, which presumed a caps frame on every reconnect; the ready arm's comment; the reconnect-skeletons header; docs/read-side.md's redial sentence, its caps paragraph ("the caps frame is also the reconnect signal"), and a closing paragraph to its viewsSeq section scoping the restore adoption and the in-flight drop to a socket whose ready the frame answered (the shell's at every open, a pane's at load and, on a reconnect, only when the bundle's ready queued across the drop); ui/webview/skeleton-tabs.ts's header. Left for a separate comment-only change, since the UI files are outside this change: the same "re-send on a reconnected socket" claim in ui/webview/timeline-boot.ts:55, ui/webview/render.ts:853-854 and ui/romp-timeline-view.js:3674-3675. A reader of those still expects a caps frame on a pane's reconnect, which comes only when a queued ready flushed onto it, so the in-flight-write drop they describe fires for a pane only in that case; that is a latent issue of its own, not this change's.

## Tests

All in tests/test_pane_shim_return.py ReconnectFlag, running the real shim core under node; the harness gains `caps()`, the frame in the shape `_send_caps` sends.

- `test_a_ready_posted_while_the_first_socket_is_still_connecting_queues_flushes_once_at_its_open_and_the_next_redial_declares_itself` (commit 1): green on the unchanged shim; red under the mutations above.
- `test_a_ready_the_kernel_never_answered_redials_without_the_flag_and_the_page_dials_fresh_until_a_caps_frame_arrives`: fails before commit 2 at its first assertion, the redial URL ending `&active=S1&reconnect=1`. A second leg pins that a later socket with no ready earns no flag either; a third, harness-only leg pins that the frame, and the frame alone, arms the gate.
- `test_a_pushed_view_frame_after_the_ready_is_not_the_kernels_answer_only_the_caps_frame_is`: fails before commit 2 the same way. The pusher serves a socket from its registration at accept, ahead of the ready's processing, so a view frame after the ready is no evidence it was processed.
- `test_a_caps_frame_arms_nothing_on_its_own_neither_before_the_bundles_ready_nor_while_that_ready_is_still_queued`: harness only (a kernel sends no caps frame before a ready); green before and after; it pins the term's shape. Red if the term drops `bundleReady`, if it drops `!readyQueued`, or if the queue bit is set only for a closed socket.
- The four cases before commit 2 now feed a caps frame after the ready on the socket whose next redial must carry the flag: a bare ready plus a keepalive no longer earns it. The first is renamed to say the kernel answered; its third leg (a later redial with only keepalives still carries the flag) is the page-lifetime control, red if the latch is cleared at a reconnect open.
- tests/test_chat_skeleton_reconnect.py: the exact-string pin follows the new term and pins the latch line.

Mutations run against the finished class: the latch never setting fails six cases; the term without `readyAcked` fails the two cases above that are red before commit 2; the term without `bundleReady`, the term without `!readyQueued`, a latch on any non-keepalive frame, and a latch cleared at a reconnect open each fail one; the queue bit set only for a closed socket fails two (the connecting pin and the four-events case).

Red before commit 2 (its tests on the unchanged shim): 3 failed, 85 passed across tests/test_pane_shim_return.py, test_chat_skeleton_reconnect.py and test_dashboard_auto_reload.py (the two cases above at their URL assertions; the exact-string pin). At head: those three plus test_kernel_ws_heartbeat.py, test_kernel_disconnect_banner.py and test_new_session_dir.py (which reads docs/read-side.md), 158 passed; `npm run typecheck` clean (skeleton-tabs.ts is comment only). Full pytest at head, with the extension's node modules installed so the served-page browser legs run: 11505 passed, 43 skipped (environment-conditioned cases, among them the two ControlProtocolPins cases that need a live CLI), 1744 subtests passed, 0 failed. An earlier full run of the same two commits on the main of a few hours before (without #1326) had one failure, tests/test_notification_tap_resume_browser.py ServedTapLanding's deep-link landing, a Chromium-driven leg that had passed in the run before it on the same code, passed 4 of 5 times alone under a load average near 10, and reads nothing this change touches; it passed on the current head.

Tier: fix

### DIFF
--- a/docs/read-side.md
+++ b/docs/read-side.md
@@ -93,13 +93,16 @@ completed); the feed just paints columns. (Reflected in `docs/judges.md`.)
   for a reload or a tab the browser discarded; a `resent: true` copy of the
   `return` row means the kept socket proved dead and the row was re-filed onto
   the redial.
-  A redial declares itself (`reconnect=1` on the `/ws` URL) once the bundle's
-  ready has left on a socket; before that, or with the ready still queued, it
-  dials as a fresh page. The kernel then sends the active tab in full and lists
-  every other session as a `skeleton` on the tab strip with one small `status`
-  frame each, and the chat pane loads a skeleton on click or one at a time in
-  idle, never while the tab is hidden; one `skeleton` client-diag row (count,
-  active) records the regime.
+  A redial declares itself (`reconnect=1` on the `/ws` URL) once the kernel's
+  caps frame has answered the bundle's ready; before that, with the ready still
+  queued, or after a socket that died before the caps frame came back, it dials
+  as a fresh page. A page whose ready was never answered dials fresh for its
+  life (the bundle posts ready once), so each of its redials is served whole.
+  On a declared redial the kernel sends the active tab in full and lists every
+  other session as a `skeleton` on the tab strip with one small `status` frame
+  each, and the chat pane loads a skeleton on click or one at a time in idle,
+  never while the tab is hidden; one `skeleton` client-diag row (count, active)
+  records the regime.
 - **The Outline pane's ages run on the kernel's clock.** Its timestamps are the
   kernel's, so the pane never reads the browser's clock against them: it anchors
   on the frame's `now` paired with the moment that frame arrived from the wire
@@ -611,12 +614,19 @@ once per distinct error, and stops retrying until the file changes or a
 write succeeds.
 
 The kernel announces what it can do in a `{type: "caps", caps, viewsSeq}` frame
-in reply to every `ready` and lists the caps on `/version`; `tagEdit` covers the
-targeted op, the acks and the `seq`. A client uses the targeted op only when the
-cap is present and posts the whole blob otherwise. A message no handler takes is
-answered `{type: "unknownOp", op, writeId}`, which the client treats as a refusal
-of that write and as withdrawing the cap. The caps frame is also the reconnect
-signal, and the `ready` handler sends it after its own connect push. `viewsSeq`
+in reply to every `ready` (a pane's bundle posts one per page life; the shell
+page's own socket posts one at every open) and lists the caps on `/version`;
+`tagEdit` covers the targeted op, the acks and the `seq`. A client uses the
+targeted op only when the cap is present and posts the whole blob otherwise. A
+message no handler takes is answered `{type: "unknownOp", op, writeId}`, which
+the client treats as a refusal of that write and as withdrawing the cap. The
+`ready` handler sends the caps frame after its own connect push, and the pane
+shim latches on it as the kernel's word that the ready was processed and the
+page served whole: a redial declares itself only once the frame has arrived.
+It is not a reconnect signal for a pane: a pane's shim re-sends no `ready` on a
+reconnect, so a reconnected pane socket gets a caps frame only when the
+bundle's ready queued across the drop and flushed onto it; the shell's socket,
+which posts `ready` at every open, learns the caps again each time. `viewsSeq`
 is the write seq of the views blob that push put on the socket: the tabOrder
 frame's, the timeline skeleton's or the feed frame's, the highest when the push
 carried more than one. When the push carried no views frame (a chat page that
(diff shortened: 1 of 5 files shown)


ksarma:
Merging when CI finishes green; nothing to fold.

Reviewed inline: the redial term gains `readyAcked` as a fourth conjunct, set in `ws.onmessage` on a caps frame. `_send_caps` has one call site, in the ready arm after `_push_one`. So a caps frame means the kernel processed this bundle's ready and served the page whole. The latch sits after the JSON parse and before the keepalive line, so the heartbeat, disconnect-banner and auto-reload source pins hold. A caps frame with no prior ready from this bundle arms nothing, because `bundleReady` and `!readyQueued` stay in the term. The shell page's `shellWS` is its own socket, so its per-open ready never reaches a pane's shim. The stated cost is accurate. A ready with no caps frame leaves the page dialling fresh for its lifetime, served whole on every redial: the cost before 2026-09-07, never a false skeleton. The `docs/read-side.md` hunks, the `KERNEL_WS_CAPS` comment, the `_send_caps` docstring and the `skeleton-tabs.ts` header match the code. The `fix` label and the body's tier line agree. Privacy scans of the 187 added lines are clean.

Follow-ups, neither a blocker:
- `kernel/kernel.py:45518`: the wsclose client-diag row carries `everConnected` but neither `bundleReady` nor `readyAcked`. So client-diag cannot tell a page in the lifetime fresh-dial regime from a fresh page. Adding both bits would make it visible.
- `kernel/kernel.py:53101`: the latch rests on `_send_caps` having one caller, and no test pins that. A source pin on the call count would make a second sender fail loudly.

Tests: at base 304489cb with the head's test files, 3 failed / 85 passed. The failures are the two new caps cases at their redial URL assertions ending `&active=S1&reconnect=1`, plus the exact-string pin in `test_chat_skeleton_reconnect.py`. At head, 158 passed across the six modules and `npm run typecheck` is clean. Nine shim mutations each fail exactly the cases the body names. CI at 7390404b: Python 3.13, bats, gitleaks, the tier-label check and Tier policy are green. Python 3.10, 3.11 and the extension job are in progress; 3.12 is queued.

Review run by the maintainer automated review session; verified against head 7390404b.

ksarma:
Head note: the head moved from 7390404b to accdbb04, a merge of main (c625cb77) into the branch. Verified here as a mechanical merge: the merge-tree of its two parents equals the head's tree, and the PR's diff against its new merge base is identical line for line to the reviewed one (235 lines, five files). The review above carries over unchanged. Auto-merge is armed on this head, so it lands when CI is green.

Review run by the maintainer automated review session; verified against head accdbb04.

PR #1319, opened 2026-09-10
Docs: the stack/** namespace is exempt from the branches-arrive-as-PRs ruleset, for stacked pull requests

henrypinkard:
Records a repository setting changed today, 2026-09-10, on the maintainer's decision in the stacked-pull-requests issue (#1011): the ruleset that turns every branch into a pull request ("branches arrive as PRs only") now exempts the `stack/**` namespace, so GitHub's stacked pull requests are possible here (every layer of a stack must live in this repository; a fork cannot host one).

- `docs/pr-tiers.md` gains a paragraph: how a member uses the namespace, that stack branches are unprotected staging refs deleted on merge, that `main` keeps every rule, that a rebased layer needs a fresh approval (approval is on the current head), and that the Actions default token stays read-only because an in-repository head runs workflows with the repository's token.
- `CLAUDE.md`'s worktree section names the one exception to "rulesets block every direct branch push".

No code change. Pre-checks made before the ruleset edit: the Actions default workflow permission is read-only and no workflow needs more than it declares; `protect main` and `tier policy` rulesets are untouched.

Tier: docs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1354, opened 2026-09-10
Docs: a stack layer is held to main's requirements, an in-repository head runs its own workflow copy with the permissions it declares, and the publish step admits the stack/** push

ksarma:
## Symptom

The stacked-pull-requests paragraph in `docs/pr-tiers.md` tells a reader that no check is required on a pull request into a `stack/<name>` branch, and that the repository's read-only default Actions token is what bounds a pull request whose head lives in this repository. The first understates the gate on a stack layer; the second names a safeguard that does not apply to a workflow with a `permissions` key. In `CLAUDE.md`, publish step 1 says the server rejects every push to romp-on/romp, a few lines after the same bullet names the `stack/**` exception, and the tiers document tells a member to push `stack/<name>` to this repository.

## Cause

The paragraph describes the staging refs (no ruleset, deleted on merge) but not the gate GitHub applies to a layer of a stack. It names the read-only default token as the bound on an in-repository head, which holds only for a workflow that declares no `permissions` key. The post-merge record on https://github.com/romp-on/romp/pull/1319 lists the three sites.

## Fix

Three sentences in two files.

1. `docs/pr-tiers.md`, the staging-ref sentence: no ruleset targets a stack branch; GitHub evaluates every layer of a stack against the stack base's required checks and reviews, so, in a stack based on `main`, a layer into `stack/<name>` merges only when it, and every layer below it, meets `main`'s requirements; no check holds a pull request into `stack/<name>` unless it is a layer of such a stack. The last sentence is conditioned on the stack's base rather than saying that only a plain, non-stacked pull request is unchecked, because GitHub lets a stack be based on any branch: a stack whose bottom layer targets another `stack/<other>` branch is judged against that branch's rules, of which there are none, so its layers are unchecked too (its code still meets `main`'s gate when `stack/<other>` is pulled into `main`). Confirmed against GitHub's documentation before writing:
   - "Each pull request in a stack is evaluated against the base of the stack (typically `main`), rather than the branch it directly targets." and "Required reviews, required status checks, and CODEOWNERS are all enforced against the stack's base branch for every pull request in the stack." (https://docs.github.com/en/pull-requests/tutorials/roll-out-stacked-prs)
   - "The merge requirements for any pull request in the stack are determined by the bottom pull request's base branch, typically `main`." and "The first, or bottom, pull request targets the stack's trunk ... though it can be any branch, such as a release branch." (https://docs.github.com/en/pull-requests/get-started/about-stacked-prs)
   - "A pull request in a stack can only merge when it, and every pull request below it, meets all merge requirements, and the stack has a fully linear history." (https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/troubleshooting-stacked-pull-requests)

   The sentence that follows (`main` keeps every rule above; a rebased layer needs a fresh approval) is unchanged.
2. `docs/pr-tiers.md`, the Actions token sentence: an in-repository head runs its own copy of every `pull_request` workflow with the repository's token and whatever `permissions` that copy declares. The read-only default bounds only a workflow with no `permissions` key (`ci.yml`; `tier-policy.yml` declares write scopes and posts the Tier policy check with them), and a head here can add or edit a workflow. The sentence now states the bounds that hold: only whoever holds write access can push `stack/**`; no ruleset has a bypass actor, so the token cannot push `main`; the repository holds no Actions secrets of its own; the `github-pages` environment deploys from `main` only. The list does not claim to be exhaustive: no ruleset targets tags, for one, and write access can push a tag directly in any case.
3. `CLAUDE.md`, publish step 1: "the server rejects it" becomes "the server rejects a push to every branch but `stack/**` (the stacked-pull-request staging area above)", rewrapped inside the numbered item at the bullet's existing width. "Every branch" rather than "every ref", because no ruleset targets tags.

Facts re-read from the live repository at the time of writing, through the read-only API:

- Rulesets 19846744 ("branches arrive as PRs only": include `~ALL`, exclude `~DEFAULT_BRANCH` and `refs/heads/stack/**`; rules creation and update), 19842293 ("protect main") and 22560299 ("tier policy") all target branches, are active, and have no bypass actors. The effective rules on a `stack/<probe>` branch are none; on `main` they are deletion, non_fast_forward, pull_request and two required_status_checks.
- Secrets: repository Actions secrets 0, Dependabot secrets 0, the `github-pages` environment's secrets 0, and that environment's deployment branch policy lists `main` alone. Organization-level Actions secrets could not be read with the token used (403), which is why the sentence says the repository holds none "of its own".
- The read-only default workflow permission is kept from the original sentence; the setting itself is not readable with the token used. The reworded sentence keeps the statement and narrows what it bounds.

## Tests

None: prose only, no code. `mkdocs build --strict` passes on the branch (the paragraph renders as one paragraph). The two modules that pin sentences in these files pass: `tests/test_tier_policy.py` (docs pins: "label" present in `docs/pr-tiers.md`, the retired fetcher sentence absent, "Declaring the tier" followed by `Tier: fix`, `label` and `wins`; `CLAUDE.md` names `Tier: fix`) and `tests/test_injected_voice.py` (`CLAUDE.md` carries the injected-voice rule and its exceptions): 136 passed, 190 subtests passed, run twice at the head and once with both files reverted to the base (same figures, since no test pins the replaced sentences). The full suite was not run for a prose change.

Tier: docs


ksarma:
Merging when the last CI check finishes; no changes requested.

Reviewed inline: the diff is three prose lines in `CLAUDE.md` and `docs/pr-tiers.md`, so the docs tier fits. Each new claim was checked at its source. GitHub's stacked-PR pages carry the cited sentences verbatim: every layer is evaluated against the stack's base; required reviews, checks and CODEOWNERS are enforced against that base for every layer; a layer merges only when it and every layer below it meet the requirements. The live rulesets agree: all three target branches only, are active, and have no bypass actors. The branches-arrive-as-PRs ruleset excludes the default branch and `refs/heads/stack/**`. `stack/probe` has no effective rules; `main` has deletion, non-fast-forward, pull request and two required-check rules. At the head, `ci.yml` is the only `pull_request` workflow without a `permissions` key; `pr-tier.yml` declares read; `tier-policy.yml` is `pull_request_target` with write scopes. Repository, Dependabot, github-pages and shared organization secrets all count zero. The pages deployment policy lists `main` alone, and delete-branch-on-merge is on. The rewritten `CLAUDE.md` sentence reads correctly against the `stack/**` exception above it.

Follow-up, `CLAUDE.md:137` and `docs/pr-tiers.md:28`: the `stack/**` exclusion exempts one path segment. The effective-rules API returns creation and update for `stack/a/b` and none for `stack/probe`, so a layer named `stack/feature/x` would be rejected. The wording predates this PR. A later edit can say one segment, or the ruleset can add `refs/heads/stack/**/*`.

Locally: `tests/test_tier_policy.py` and `tests/test_injected_voice.py`, 136 passed and 190 subtests passed at head and base. `mkdocs build --strict` is clean. The three added lines carry no private strings, host or home paths, or banned vocabulary. CI at the head: eleven checks green, Python 3.11 still running at the last read.

Review run by the maintainer automated review session; verified against head 14469090.

PR #1362, opened 2026-09-10
Session events: the default since is the boot's whole second, and a drain row is a problem row

ksarma:
## Symptom

`GET /session-events` with no `?since=` can answer a `count` larger than the number of problem rows it lists: a row stamped in this kernel's boot second is counted and missing from `rows`, and the echoed `since` names a fraction of a second while `bootAt` names the whole second. The realistic case is the previous kernel's `drain.unjoined` row: it is written as that kernel exits, the manager spawns the next kernel on that exit, and `_STARTED` often lands in the same wall second. An unparseable `?since=` falls back to the same fraction.

A `drain.unjoined` row itself has no `text` and no `<prose> ;; problem-row {json}` kernel-log line, although the header comment in `kernel/sdk_backend.py` and `docs/reference.md` promise both for every kind but the boot summary. A log reader parsing on the marker never sees a drain row.

## Cause

The handler defaulted `since` to `_STARTED`, a float, both when the parameter is missing and when it does not parse, while `_session_event_rows` counts with `int(_STARTED)` and every row's `t` is a whole second (`append_session_event` stamps `int(time.time())`). A row with `t == int(_STARTED)` passes the count's `t >= boot_t` and fails the listing's `t >= since`.

The drain loop appended its rows with `append_session_event` directly instead of `problem_row`, which is what gives a row its `text` and the marker line.

## Fix

- `kernel/kernel.py`: the default `since` is `int(_STARTED)` for a missing and for an unparseable `?since=` alike, the same whole second `count` filters by and `bootAt` reports, so the default rows and `count` are one predicate but for the boot summary and the `limit` cap (`count` is uncapped). An explicit `?since=` still parses as a float. The helper's docstring and the handler's comment say why the whole second, and that a previous kernel's row stamped in this kernel's boot second is therefore both listed and counted (the resolution the row stamps have). Switching only `count` to a pid compare would reopen the disagreement the other way (a previous kernel's boot-second row listed but not counted).
- `kernel/sdk_backend.py`: the drain writes each unjoined session's row through `problem_row(..., log=self._log, ring=False, ...)`, so the row carries its prose as `text` and the kernel log gets the parseable line. `ring=False` stays: the kernel is exiting and the bell has no reader for a ring entry. The prose reads `drain: session <name> was still closing when the shutdown's wait ran out`, with `; its claude process was ended` when the reap ran. The header comment names the one exception, the ring.
- `docs/reference.md`: the `session-events.jsonl` paragraph says every kind but the boot summary carries `text` and the log line, every kind but the boot summary and `drain.unjoined` is a ring entry too, and the route's default `since` is the boot in whole seconds.

The echoed `since` changes from a float to an int by default; no reader consumes it today (`cli/restart_metrics.py` reads the file, not the route), and it now equals `bootAt`.

## Tests

- `tests/test_session_events_route.py::SessionEventsRoute::test_a_row_in_the_boot_second_is_counted_and_listed`: `_STARTED = BOOT + 0.6` and a previous kernel's `drain.unjoined` row at `t == BOOT`; for a GET with no query and for `?since=garbage` (two subtests), asserts the row is listed, `count` equals the number of non-summary rows, and `since == bootAt == BOOT`. Before the fix both subtests fail with `'drain.unjoined' not found in ['crash.heal', 'reconcile.boot', 'reconcile.orphan-reaped']` (count 3 against 2 listed); with only one of the two handler defaults changed, the other subtest still fails.
- `tests/test_session_events_ledger.py::DrainRows::test_unjoined_row_is_a_problem_row_without_the_ring`: drains the two doubles with a logging backend; asserts the row has a `text` naming the session, exactly one log line carries `PROBLEM_ROW_MARK` and parses to the same row, the drain's own summary still logs, and the ring is empty. Before the fix: the row has no `text`.
- `tests/test_session_events_ledger.py::DrainRows::test_reaped_row_says_so_in_its_prose`: the CLI-pid seam answers a pid above pid_max and a stub `kill` reports it gone at the first existence poll (nothing real is signalled); asserts `reaped` is true and the prose says the process was ended. Before the fix: `KeyError: 'text'`.
- The existing `test_unjoined_sessions_get_a_row_each` keeps its assertions, on a drain helper the three tests share.

Full suite on this head: 11650 passed, 45 skipped, 1746 subtests passed, 1 failed. The failure is `tests/test_notification_tap_resume_browser.py::ServedTapLanding::test_a_click_the_worker_saw_lands_the_page_via_the_message_and_settles_the_row`, a served-page browser test of the notification tap landing that reads none of the code this change touches; run alone right after, it and its module pass (1 passed; 3 passed).

Tier: fix

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -2133,12 +2133,16 @@ sweep summary of every boot that had a session to reconcile: `sessions`, `resume
 as the boot's process listing stood: `fsid`, `pids`, `n`), `crash.heal` and
 `crash.loop` (`attempt`), `drain.unjoined` (a session the drain's bound left
 closing: `inflight`, `reaped`), and the lease work's `lease.*` kinds. Every kind
-but the boot summary is also a problem-ring entry (the bell and error center
-show its prose) and a kernel-log line of the form `<prose> ;; problem-row
-{json}`, the same object after the marker, so a log reader parses it with a
-split on the marker. `GET /session-events?since=<epoch s>&limit=<n>`
-(token-gated) returns the rows newest first since the stamp (default this
-kernel's boot; at most 1000), each with `host`, and `count`, this kernel's
+but the boot summary carries `text` and is also a kernel-log line of the form
+`<prose> ;; problem-row {json}`, the same object after the marker, so a log
+reader parses it with a split on the marker; every kind but the boot summary
+and `drain.unjoined` (written as the kernel exits, when the bell has no reader)
+is a problem-ring entry too (the bell and error center show its prose).
+`GET /session-events?since=<epoch s>&limit=<n>` (token-gated) returns the rows
+newest first since the stamp (default this kernel's boot in whole seconds, the
+resolution every row's `t` has and the `bootAt` the response names, so the
+default rows and `count` are one predicate but for the boot summary and the
+`limit` cap; at most 1000), each with `host`, and `count`, this kernel's
 problems since its boot, never a sum across kernels. `turns.jsonl` gets one
 row per settled turn: `t`, `sid`, `name`, `fedT` (the feed pop, when the text
 left the queue for the CLI's stdin, at millisecond resolution), `firstOutT`
--- a/kernel/kernel.py
+++ b/kernel/kernel.py
@@ -15109,9 +15109,14 @@ def _session_event_rows(since=0.0, limit=200, tail=SESSION_EVENTS_TAIL):
     CLIs holding one conversation, a crash heal or loop, a session the drain left closing, and the boot
     sweep's summary), newest first, rows with t >= `since`, at most `limit`, each carrying `host` (this
     kernel's own name, _self_host) so a federated shell merges per-host maps and never sums across kernels.
-    `count` is the problems since THIS kernel's boot on THIS host: every row at or after _STARTED except the
-    boot summary (reconcile.boot, which every boot writes). Reads the file's tail only; a missing or
-    unreadable ledger is ([], 0)."""
+    `count` is the problems since THIS kernel's boot on THIS host: every row at or after int(_STARTED) except
+    the boot summary (reconcile.boot, which every boot writes). The route's default `since` is that same whole
+    second, the resolution every row's `t` has (append_session_event stamps it to the second), so the default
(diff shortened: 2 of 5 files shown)


ksarma:
Merging when CI is green; nothing to fold.

Reviewed inline: the `/session-events` handler now defaults `since` to `int(_STARTED)` for a missing or unparseable `?since=`. That is the whole second `_session_event_rows` counts from and `bootAt` reports. So the default rows and count are one predicate apart from the boot summary and the limit cap. An explicit value still parses as a float. Nothing in `ui/`, `cli/` or the extension reads the echoed `since`; `cli/restart_metrics.py` reads the file. The drain's unjoined rows go through `problem_row` with `log=self._log` and `ring=False`. At the head, `problem_row` writes the prose as the row's text, logs the `;; problem-row` line, and wraps the log call in try/except. `_log` with `problem=False` skips the ring. So nothing new can raise out of the drain, whose own summary line already uses the same sink. Every other problem kind already uses `problem_row`, and the drain is the only `ring=False` call. So the header comment and `docs/reference.md` read true against the code. The prose carries only the session name, which the drain's summary line already logs. Privacy scan of the 113 added lines is clean. The one `fix` label and the `Tier: fix` body line agree.

Test evidence:
- Base 7d48b6d3 with the head's two test files: 4 failed, 31 passed. The failures are both DrainRows tests on the missing text, and both route subtests with `drain.unjoined` absent from the listed kinds.
- Head: 33 passed, 2 subtests passed.
- Four single-line mutations (each handler default alone, `ring=True`, `log=` dropped) each fail the test that guards them.
- CI at 6e4f315a: tier checks, Python 3.11 and 3.13, bats and gitleaks green. Python 3.10 and 3.12 still running.
- The vscode-extension job is red on the served tap-landing browser test. That test fails identically on main's own run at feb57836 and reads none of this PR's files. It needs a re-run, not a change here.

Review run by the maintainer automated review session; verified against head 6e4f315a.

ksarma:
Head note: the branch was updated with main, so the head moved from 6e4f315a to 2766381b. The new head is a mechanical merge of 6e4f315a and dc90c876: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 2766381b.

PR #1365, opened 2026-09-10
Postal cards: a sent card boxed by its fold wears the whole provisional dress, not a solid box at 72%

ksarma:
## Symptom

A sent postal card with a fold rendered, while its message had not landed, as a solid box three-quarters of the column wide and dimmed, then snapped to the full width when the receipt arrived. The design is one provisional dress for every send that has not yet landed: a sent card handed to the relay or parked for an unreachable host wears the queued bubble's dashed border and wash at the column's full width, and turns solid when the receipt says delivered, relayed or read. Only the slim card wore it. A sent card whose first line runs past the head's 90-character clip, or that has a second line, has a fold, so the density rule boxes it, and the boxed card fell outside the dress. A delegation is usually more than one line, so this was the common case. The review record on #1318 lists it as the first follow-up.

## Cause

The shared rule names the bubble as `.queued-bubble` and the slim card as `.notice.notice-slim.queued-bubble`, nothing in between. On a boxed sent card only `.queued-bubble` (0,1,0) matches, and `.notice` (0,1,0), later in the sheet, takes the border, background and radius. The bubble's `max-width: 72%` and `opacity: 0.85` survive because `.notice` sets neither, and the width reset requires `.notice-slim`, so it never reaches that card. The result is a solid box at 72% of the column, dimmed, which becomes full width when `deliveryOf` stops returning sent or parked.

## Fix

Two selector additions in `ui/webview/styles.css`, no new declarations and no colours:

- The shared block becomes `.queued-bubble, .notice.queued-bubble, .notice.notice-slim.queued-bubble`. `.notice.queued-bubble` (0,2,0) beats `.notice` (0,1,0) whatever the order. The slim selector (0,3,0) stays, because `.notice.notice-slim` (0,2,0) is later in the sheet and would otherwise beat `.notice.queued-bubble` on the slim card.
- The width reset becomes `.notice.queued-bubble, .notice.notice-slim.queued-bubble { max-width: none; display: block; }`, so the boxed card keeps the column too and nothing moves when the receipt lands.

One visible consequence: while provisional, the boxed card loses the 2px peer-colour rail to the dashed border, as the slim provisional card and the user's own queued bubble already do. The rail returns with the receipt. The rail dot and time marker already sit at the boxed offset on a provisional turn (`.turn-postal-service.postal-provisional > .dot { top: 18px }`), so their position is unchanged.

## Tests

- `tests/test_postal_cards_served.py`, the Chromium harness over the real /chat page of a hermetic kernel at 1000, 520 and 340 px plus the light theme: an eleventh card, a parked delegation to `tests` with a 117-character line, so it has a fold and is boxed. The driver now records each card's computed `max-width` and measured width. The new assertions: the card is collapsible and not slim, provisional, its border is dashed, its background and width equal the slim parked card's, and its max-width is `none`. Before the change it fails at the border, `'solid' != 'dashed'`, with `maxWidth: '72%'`, the box background and a width of 645 px at 1000 px. The "sent stays slim" pin becomes slim-iff-no-fold (the new card fails the old pin too), and the card counts move to eleven, three provisional at phone width.
- `ui/webview/postal-card.test.ts`: the provisional test pins the two selector lists and says why the slim selector stays; before the change it fails by text mismatch.
- At this head, locally: `npm run typecheck` clean; `npm test` 4315 passed, 0 failed (4293 in the main run plus the 22 of `ui/timeline-tags-scale.test.ts`, run on its own); `pytest tests/ -n 6` 11687 passed, 45 skipped, 1744 subtests passed, with the served-page modules executing.

Tier: fix

### DIFF
--- a/tests/test_postal_cards_served.py
+++ b/tests/test_postal_cards_served.py
@@ -7,8 +7,9 @@ colours), never a chip; the DELIVERY STATE is one icon per state at the head's r
 (delivered, read, parked, bounced, recalled, and sent for a message handed to the relay), from what the kernel
 files (the send-time stamp, and the postal ledger's exec / relayed / bounced / recall rows joined by message id);
 both ENDS wear their sessions' colours (the peer's chip, then this session's own chip); no card wears a background
-wash; incoming cards are boxed, sent ones slim; a sent card whose message has not landed wears the pending
-send's own provisional dress (the queued bubble's class). With POSTAL_SHOTS=<dir> the driver also writes
+wash; incoming cards are boxed, sent ones slim unless a fold boxes them; a sent card whose message has not landed wears
+the pending send's own provisional dress (the queued bubble's class) at either density, slim or boxed, at the column's
+full width. With POSTAL_SHOTS=<dir> the driver also writes
 screenshots at 1000 px, 520 px, 340 px dark and 1000 px, 340 px light, named romp_chat-postal-cards-<theme>-<width>.png (the phone width: the head wraps, the ends first, the kind word and the icon on
 that line or the next as one unit, the gist last on its own full-width line, T313) and the light theme. Skips LOUDLY without the extension deps or a Playwright browser. SYNTHETIC
 fixtures only (the notes-api demo world: web / api / tests; host TESTHOST)."""
@@ -95,7 +96,14 @@ def sent_row(mid, frm, frm_id, to_id, body, t, kind, park=False):
 KINDLESS = "Merged the fixtures branch; nothing else is pending on my side."
 
 
-# The world: ten cards, every kind and every state, plus one legacy card with no kind. Bodies are invented notes-api chatter.
+# a sent one-liner OVER the 90-char clip: its gist cannot carry the whole line, so the card has a fold and is boxed; parked,
+# it must wear the same provisional dress as the slim parked card beside it (the box rule used to win its border and
+# background, and the bubble's 72% cap its width)
+LONG_PARKED = "Once the fixtures land, run the whole integration suite against notes-api and write every flaky case into the README."
+
+
+# The world: eleven cards, every kind and every state, among them one legacy card with no kind and one parked card with a
+# fold. Bodies are invented notes-api chatter.
 def world(t0):
     msgs = {
         "in-deleg": ("api", API, "delegate", "Take the retry-loop rewrite in notes-api: exponential backoff with jitter, cap at two minutes, tests included."),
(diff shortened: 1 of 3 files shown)


ksarma:
Merging when CI is green on an updated head. Merge main first: the branch conflicts with it in one docstring.

Reviewed inline: the two selector additions in `ui/webview/styles.css` and the cascade they change. At the base a boxed sent card matched only `.queued-bubble` (0,1,0). The later `.notice` rule at line 2370 took border, background and radius. The bubble's 72% max-width and 0.85 opacity survived. `.notice.queued-bubble` (0,2,0) now outranks it. The slim (0,3,0) selector still beats the later `.notice.notice-slim`. The width reset reaches both densities, with no new declarations or colours. The fold rule at line 2497 is untouched. `display: block` is `.notice`'s default, so the collapsible head and hidden body behave as before. The rail dot offset is the one a boxed turn already has. A static Chromium probe over the base and head stylesheets, without a kernel, measured the boxed provisional card. At the base: 72% width, solid border, box background. At the head: max-width none, the slim provisional card's width, dashed border, the bubble's background. The slim provisional card, the boxed solid card and the user's own queued bubble (72% kept) were identical between the two. The rail hidden while provisional matches the slim card's styling and returns with the receipt. Added lines are clean of private strings, hostnames and home paths. Label and body tier line agree.

- `tests/test_postal_cards_served.py:10`, follow-up: the only conflict with main is this module docstring. #1358 rewrote the screenshot-naming sentence beside the one this PR rewrote; `styles.css` and `postal-card.test.ts` auto-merge. Merge main and keep both sentences; no assertion changes.

Red-first: the head's `ui/webview/postal-card.test.ts` fails at base feb57836 on the first selector regex (4 pass, 1 fail) and passes at the head. The nine compiled UI tests referencing these classes pass 79/79 at the head and on the merged stylesheet; typecheck clean. The served test was not run here (a browser fixture over a real kernel). By reading, its first failing assertion at the base is the border check, solid against dashed, as the body says; CI runs it. CI has not started at this head; only the tier checks are present.

Review run by the maintainer automated review session; verified against head b1ea9d03.

ksarma:
Head note: the branch was rebased onto 0cfa6bfb (the #1358 merge) to resolve the docstring conflict; the head is now a942836c. Verified here: the PR's diff against its new base is identical line for line to the reviewed one (66 lines, three files), and a merge-tree against current main is conflict-free. The review above carries over unchanged. Auto-merge is armed on this head, so it lands when CI is green.

Review run by the maintainer automated review session; verified against head a942836c.

ksarma:
Head note: the branch was updated with main, so the head moved from a942836c to 8580fb1e. The new head is a mechanical merge of a942836c and 1af162be: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 8580fb1e.

PR #1373, opened 2026-09-10
Restart metrics: a turn the CLI opened itself measures no latency, and the feed stamps get driven tests

ksarma:
## Symptom

`romp restart-metrics` reports an hours-long feed-to-result and a duplicated feed-to-first-output sample wherever turns.jsonl holds a row that a kernel before the 2026-09-10 writer fix wrote for a turn the CLI opened by itself (a channel message, a task notification, a scheduled prompt). The summary's "feed to result" line, the day and week buckets and the report's latency frames all carry the sample. The current kernel no longer writes that row shape (a fedT now implies fedTexts >= 1), so only rows recorded under the earlier writer are affected, and a baseline read over them is wrong.

The post-merge review of #1340, the PR that added restart metrics, named this reader check and two test gaps; the gaps are closed here as well. The two writers of the stamps the latency figures depend on, the pop's `_fed_t` in inputs() and the first work atom's `_first_out_t` in _forward, had no executing test: deleting either line left the suite green and the reader reporting n 0 with no error. The finally that writes the turn's row and spends the stamps was pinned by source-text indices only. And in tests/test_restart_metrics.py the spend_by_day assertion ran under test_state_log_pair_breaks_at_a_machine_cut, whose docstring never claimed it, so a spend_by_day regression would have reported under the machine-cut test's name.

## Cause

parse_turns derived the seconds from any row with fedT > 0 and resultT >= fedT. The earlier writer set fedT from `since` and firstOutT from `_first_out_t` unconditionally, and both were reset only at a fresh feed, so a self-opened turn's row carried the previous fed turn's stamps with fedTexts 0, and the reader took the interval from one turn's feed to another turn's result. The misplaced assertion dates from the commit that inserted the machine-cut test between test_events_turns_and_state_log's last two statements.

## Fix

- cli/restart_metrics.py: parse_turns keeps a row whose fedTexts is 0 as a turn and derives no seconds from it. The gate is the count 0 alone: a row without the key is read by its stamps as before, and a bool is not a count. The bucket count still takes the row as a turn; only its feed latency is unmeasured.
- tests/test_restart_metrics.py: the spend_by_day assertion is back as the last statement of test_events_turns_and_state_log, and the machine-cut test ends at its cuts assertion, as its docstring says.
- tests/test_session_events_ledger.py: the index-arithmetic pin (now test_the_result_branch_stays_one_try) keeps two assertions that are checkable on the source alone: the branch's one-try shape, and the feed stamps reset ahead of the settle inside the finally (plain assignments with no await between them, so the order is not observable from outside the finally). The row's content under a raising spend write, the spend tuple set before that write, and the stamps spent afterwards are driven by the new tests below.

## Tests

- tests/test_restart_metrics.py: a fourth turns row with the self-opened shape (fedTexts 0, the 01:00 turn's fedT and firstOutT, a 02:01 resultT). Before the gate, Parsers.test_events_turns_and_state_log fails on turns[3] carrying feedToResultS 3657.0 and feedToFirstOutS 2.5; Document.test_week_buckets_carry_every_metric fails on feedToResultS n 3 and max 3657.0; Document.test_summary_text_names_the_numbers fails on the summary's feed-to-result count (3 for 2); tests/test_restart_metrics_report.py Frames.test_frames_from_two_documents fails on latencySamples [37.0, 10.0, 3657.0]. The week bucket's turns count moves from 3 to 4.
- tests/test_restart_metrics.py, Parsers.test_parse_turns_gates_on_a_zero_count_alone: parse_turns over four direct rows with the same stamps and no fedTexts key, fedTexts 0, fedTexts false and fedTexts 1 derives (5.0, 2.0), nothing, (5.0, 2.0) and (5.0, 2.0). It fails without the gate (the 0 row derives), with a gate keyed on falsiness (`if not n:`, the keyless row derives nothing), with a gate that also takes a missing key (`n is None or ...`), and with the bool exclusion dropped (the false row derives nothing).
- The spend_by_day move, shown executing: with rm.spend_by_day replaced by a function returning {}, the machine-cut test fails before the move and test_events_turns_and_state_log fails after it, read from those two tests alone; a deliberate wrong expected value (99.0 for 12.5) fails test_events_turns_and_state_log at the moved line.
- tests/test_sdk_stream_survives_handler_errors.py, TheDeferredReconnectTakesTheHeldQueueWithIt.test_a_fed_turn_gets_the_pop_stamp_and_its_row_carries_both_event_stamps: the real _amain on the stand-in SDK runs a fed turn; the pop stamp, the first-output stamp and the settled row carrying both (fedT <= firstOutT <= resultT, fedTexts 1, opener human, usd present) are read through the class's _wait, and the stamps are None afterwards. Removing the `self._fed_t = time.time()` line in inputs() fails it (timed out waiting for the pop stamp); removing the `sess._first_out_t = time.time()` line in _forward fails it too.
- tests/test_session_events_ledger.py, TurnLedgerDriven.test_forward_stamps_the_first_work_atom_once: a real SdkSession with inflight 1; the first work atom stamps, the second keeps the first's stamp, nothing in flight stamps nothing, a command line never stamps. Removing the _forward stamp fails it (None is not a float).
- tests/test_session_events_ledger.py, TurnLedgerDriven.test_the_settle_writes_the_row_and_spends_the_stamps_when_the_spend_write_raises: a real SdkSession travels the ResultMessage branch through _handle_stream_message with _record_spend raising. The row lands with fedT, firstOutT, fedTexts 1, usd 0.5, durationMs 10, apiMs 8 and the token counts; the stamps and _turn_spend are None afterwards; inflight is 0 and the state log reads waiting; the handler returns False. Removing the three reset lines fails it (_fed_t still 1700000000.125); moving the append out of the finally to after the spend write fails it (no turns.jsonl); setting the spend tuple after the raising write fails it (no usd on the row).
- tests/test_session_events_ledger.py, TurnLedgerRow.test_the_result_branch_stays_one_try: the source pin. Moving the three resets after `self.inflight = 0` fails it and no driven test, since the order is not observable from outside the finally.
- Full pytest suite on this branch: 11719 passed, 45 skipped, 1746 subtests passed (pytest -n 6 over tests/ on the final tree, the served-page legs included).

Tier: fix

### DIFF
--- a/cli/restart_metrics.py
+++ b/cli/restart_metrics.py
@@ -260,12 +260,20 @@ def parse_events(rows: list[dict]) -> list[dict]:
 
 
 def parse_turns(rows: list[dict]) -> list[dict]:
-    """turns.jsonl rows with the derived seconds: feedToResultS, feedToFirstOutS (when both stamps exist)."""
+    """turns.jsonl rows with the derived seconds: feedToResultS, feedToFirstOutS (when both stamps exist and
+    the row says the turn was fed: a kernel before the 2026-09-10 writer fix stamped a turn the CLI opened
+    itself, fedTexts 0, with the previous fed turn's fedT and firstOutT, which read as hours of feed-to-result
+    and a duplicated first output; such a row is a turn and measures no feed latency). The gate is the count 0
+    alone: a row without a fedTexts key is read by its stamps, and a bool is not a count."""
     out = []
     for r in rows:
         if not isinstance(r.get("t"), (int, float)):
             continue
         rec = dict(r)
+        n = r.get("fedTexts")
+        if isinstance(n, int) and not isinstance(n, bool) and n == 0:
+            out.append(rec)          # nothing fed: the stamps, if any, are another turn's
+            continue
         fed, res, fo = r.get("fedT"), r.get("resultT"), r.get("firstOutT")
         if isinstance(fed, (int, float)) and fed > 0 and isinstance(res, (int, float)) and res >= fed:
             rec["feedToResultS"] = round(float(res) - float(fed), 3)
--- a/tests/test_restart_metrics.py
+++ b/tests/test_restart_metrics.py
@@ -83,6 +83,10 @@ def _fixture(state: Path):
         {"t": t1 + 100, "sid": SID2, "name": "api", "fedT": t1 + 90, "firstOutT": t1 + 91.0, "resultT": t1 + 100.0,
          "durationMs": 10000, "apiMs": 8000, "usd": 0.1, "opener": "human", "resumeNotice": False, "fedTexts": 1},
         {"t": t2 + 50, "sid": SID2, "name": "api", "fedT": 0, "resultT": t2 + 50.0, "opener": "", "resumeNotice": False},
+        # a turn the CLI opened itself at 02:01, written by a kernel before the 2026-09-10 writer fix: nothing fed
+        # (fedTexts 0), and the stamps still in memory were the 01:00 turn's (its fedT and firstOutT)
+        {"t": t2 + 60, "sid": SID, "name": "web", "fedT": t1 + 3, "firstOutT": t1 + 5.5, "resultT": t2 + 60.0,
+         "opener": "injected", "resumeNotice": False, "fedTexts": 0},
     ])
     _write(state, "states/%s.jsonl" % SID, [
         {"t": int(t1 - 100), "state": "working"}, {"t": int(t1 - 70), "state": "waiting"},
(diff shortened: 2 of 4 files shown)


ksarma:
Arming auto-merge: the PR lands when CI finishes green at head 82d52783. No changes requested.

Reviewed inline: the `parse_turns` gate keys on an int `fedTexts` equal to 0 and nothing else. A keyless row therefore still reads its stamps, and a bool is not a count. The gated row stays in the bucket's turn count. The summary and report frames do no arithmetic on it. Every romp writer since aeb44d30 has written `fedTexts` as `len(fed)`. That leaves no keyless row, and every pre-fix inherited-stamp row carries `fedTexts` 0. The pre-fix kernel did not raise `inflight` for a turn the CLI opened itself. A text fed during such a turn popped as fresh with its own stamps. No other legacy shape inherits them. The driven tests pass their doubles in as the handler's type arguments. The doubles carry the class names `msg_to_atom` keys on. The renamed source pin's two anchors each occur once inside `_on_message`, both in the `finally`. Moving the resets after `inflight = 0` fails only the pin, as the body says. `docs/reference.md` describes the current writer. The reader's notes already say a self-opened turn is not a latency sample. Privacy scan of the 171 added lines: no hits.

Red-first at base 9b0b1378 with the head's test files: 5 failed, 96 passed. The five failures are `Parsers.test_events_turns_and_state_log`, `test_parse_turns_gates_on_a_zero_count_alone`, `Document.test_week_buckets_carry_every_metric`, `Document.test_summary_text_names_the_numbers` and `Frames.test_frames_from_two_documents`, each for the reason the body gives. The three driven tests pass at the base and fail under each of the code mutations the body names. At head 82d52783 the five modules: 101 passed, 2 skipped (no plotting library).

CI at 23:28Z: Python 3.11, 3.12, 3.13, bats, gitleaks and the tier-label check are green. Python 3.10, vscode-extension and Tier policy are still running, so this arms auto-merge rather than merging now.

Review run by the maintainer automated review session; verified against head 82d52783.

ksarma:
Head note: the branch was rebased onto 7d72b752 to resolve the docstring conflict with #1362; the head is now a61f8138. Verified here: the PR's diff against its new base is identical line for line to the reviewed one (185 lines, four files), a merge-tree against current main is conflict-free, and both PRs' tests are present in tests/test_session_events_ledger.py (DrainRows from #1362; TurnLedgerDriven and the renamed pin from this PR). The review above carries over unchanged. Auto-merge is armed on this head, so it lands when CI is green.

Review run by the maintainer automated review session; verified against head a61f8138.

PR #1374, opened 2026-09-10
Tests: the CLI probe hands its child no ANTHROPIC_* name and runs the --bg refusal with no prompt, stdin at /dev/null, in a process group tearDown kills; the reference scopes the hook-timeout sentence to the tool hook it measured and counts three pinned facts

ksarma:
## Symptom

`tests/test_cli_control_protocol_probe.py`, the opt-in probe behind `ROMP_CLI_PROBE_LIVE=1`, hands the Claude Code CLI every `ANTHROPIC_*` variable the developer's shell exports, while its docstring says the child has no key. Its `--bg` refusal test runs a real prompt with the test's stdin inherited, no process group and nothing for tearDown to kill: had the CLI accepted `--bg --print` (the regression the test exists to catch), the prompt would have run under a daemon, on that credential, beyond the 60 s timeout's reach. In `docs/reference.md`, the paragraph on what the CLI does when its parent goes quiet says a hook callback's timeout ends with a tool-result error and the turn going on, which does not hold for the `UserPromptSubmit` hook romp registers; the same paragraph says the module pins two facts where it pins three.

## Cause

`hermetic_env` filtered `CLAUDECODE`, `CLAUDE_*`, `ROMP_*` and `XDG_*` only. Under pytest, `tests/conftest.py` pops `ANTHROPIC_API_KEY` and `ANTHROPIC_AUTH_TOKEN`, which hid the gap; a proxy's `ANTHROPIC_BASE_URL` and `ANTHROPIC_CUSTOM_HEADERS` still passed, and the module's own `__main__` loads no conftest. The `--bg` test used `subprocess.run` with a prompt argument no assertion reads. The hook-timeout sentence records a `PreToolUse` measurement (the only hook kind the module registers) without naming the kind; in the 2.1.266 CLI the hook dispatch converts a timed-out `UserPromptSubmit` or `UserPromptExpansion` callback into a blocking decision that suppresses the prompt, and hands every other event's timeout to that event's own handler. The count of two predates the module's third pinned fact, the idle exit on stdin end-of-file.

## Fix

- `hermetic_env` drops every `ANTHROPIC_*` name, by prefix as `kernel/judge.py` does for the codex judge: the CLI grows names, and the harmless ones may go too. The docstrings say what the scrub controls (the environment and the config directory; managed settings outside the config directory are not the test's to control) and that a group kill does not reach a daemon that left its group.
- Two module-level helpers. `kill_group(p)` is the group SIGKILL `StreamCli.kill` did, then a reap and a close of the pipes (never a drain: a daemon that left the group can hold the write ends). `run_probe(exe, argv, env, cwd, timeout, procs)` runs a command with stdin at `/dev/null` in a process group of its own, appends the Popen to `procs` before waiting, and on timeout kills the group and re-raises `TimeoutExpired`.
- `ControlProtocolPins.setUp` starts `self.procs`; `tearDown` kills every recorded group. The `--bg` test drops the prompt and runs through `run_probe`; its assertions are unchanged.
- `docs/reference.md`: the hook-timeout sentence names the measured kind, states the `UserPromptSubmit` outcome with its version (2.1.266) and source (read from the CLI's hook dispatch, not measured), notes that romp registers that hook and sets no matcher `timeout`, and marks the other hook kinds unmeasured; the probe sentence counts three facts. The rest of the paragraph is rewrapped at 80 columns with its words unchanged (`git diff --word-diff` shows the two sentences alone).

## Tests

Two new classes in the module, outside the live gate, so they run on CI and start no CLI:

- `HermeticEnvPins.test_drops_every_anthropic_claude_romp_and_xdg_name` plants four `ANTHROPIC_*` names (a key, a token, a base URL, a custom header; the values are three plain words and a loopback URL, all short, none credential-shaped) plus `CLAUDECODE`, a `CLAUDE_*`, a `ROMP_*` and `XDG_DATA_HOME`, calls `hermetic_env`, and asserts none reaches the child and that `CLAUDE_CONFIG_DIR` and `XDG_STATE_HOME` point under the test root; assertions render names, never the mapping. Before the change it fails listing the `ANTHROPIC_*` names kept.
- `ProbeRunPins.test_run_probe_takes_its_own_group_and_kill_group_reaches_an_orphan`: `/bin/sh -c` starts a background `sleep` with its descriptors redirected (so `communicate` sees end-of-file) and prints its pid; the sleeper's pgid is the sh's pid, and after `kill_group` the sleeper is gone (reaped, or a zombie per `/proc/<pid>/stat`, polled up to 5 s). Fails without `start_new_session` (the pgid is the test runner's).
- `ProbeRunPins.test_run_probe_kills_the_group_when_the_timeout_expires`: the sh records its sleeper's pid and waits on it; `run_probe` raises `TimeoutExpired` at 3 s, the sh has exited, its pipes are closed and the sleeper is gone. Fails when `kill_group` stops closing the pipes, and without `start_new_session`.
- `ProbeRunPins.test_teardown_kills_a_group_the_test_left_running`: the test leaves the sleeper running and asserts from a cleanup, which runs after tearDown, that it is gone. Fails when tearDown stops killing the recorded groups.
- `ProbeRunPins.test_run_probe_reads_stdin_from_dev_null`: for the duration of the run the test's own fd 0 is a pipe carrying bytes (pytest's capture otherwise puts `/dev/null` on fd 0, which would make an inherited stdin look the same as `/dev/null`); `cat; echo eof` prints `eof` alone at once. With `run_probe` inheriting stdin it fails with `inherited` echoed first, under pytest and under a bare unittest run alike.
- Before the change the four `ProbeRunPins` tests fail with `NameError` on `run_probe`. Module before: 5 failed, 2 skipped; after: 5 passed, 2 skipped (the live pair skips with the gate unset), the same under `-n 2`, under `-s` with a pipe on the runner's stdin, and as a bare `python3 tests/test_cli_control_protocol_probe.py` run (7 tests, 2 skipped).
- The live pair keeps its assertions and was not run here. Dropping the prompt changes the argv the CLI validates; the `--bg`/`--print` conflict is a flag check and `--input-format stream-json` takes its prompt from stdin, so the refusal text should not change, but one live run with a `claude` on PATH would confirm it.
- Full suite (`pytest -n 6 tests/`, extension dependencies installed so the served-page tests run): 11691 passed, 45 skipped, 1744 subtests passed, 1 failed. The failure is `tests/test_dashboard_reload_served.py::ServedAutoReload::test_build_drift_reloads_after_the_gesture_ends_and_a_restart_reloads_too`, a served-page browser test of the dashboard's reload that reads none of the two files this change touches; run alone right after, it passes (1 passed in 23 s).

Tier: docs

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -951,21 +951,30 @@ What the CLI itself does when its parent goes quiet was measured on Claude Code
 2.1.257 (2026-09-10, the restart-surviving sessions program's stage 3 probe, run
 against a throwaway config directory): a permission request (`can_use_tool`)
 waits for its answer with no expiry within ten minutes and the turn continues
-normally on a late answer; a hook callback waits 600 seconds by default, or the
+normally on a late answer; a tool hook callback (a `PreToolUse` hook on Bash,
+the kind the probe module registers) waits 600 seconds by default, or the
 matcher's `timeout` seconds when one is set, then the CLI cancels the request
 (`control_cancel_request`), records a hook-timeout error as the tool's result
-and goes on with the turn; a second `initialize` on the same stdin is accepted
-and its hook table replaces the first; stdin end-of-file ends an idle CLI at
-once (0.02 s) and a busy one after its turn (a 30 s tool call ran to completion
-first); an unread stdout does not stall the CLI (the pipe's 64 kilobytes fill,
-the rest buffers inside the process, the turn completes); `--resume` takes no
-lock, and two processes on one session id both append to the one transcript;
-`claude --bg` runs an interactive session on a pseudo-terminal under a daemon
-that stays in the launcher's cgroup, and refuses `--print`, so a background
-session has no stream-json channel. `tests/test_cli_control_protocol_probe.py`
-re-checks the two facts that need no model call (the second initialize, the
-`--bg` refusal) when run with `ROMP_CLI_PROBE_LIVE=1` and a `claude` on PATH; it
-skips otherwise, as every test that would reach the live CLI must.
+and goes on with the turn; the CLI instead treats a timed-out `UserPromptSubmit`
+callback as a blocking decision and suppresses the prompt (Claude Code 2.1.266,
+read from the CLI's hook dispatch rather than measured: that dispatch converts a
+timed-out prompt-hook callback into a block and hands every other event's
+timeout to that event's own handler; romp registers that hook and sets no
+`timeout` on any matcher); what a timed-out `Stop`, `SubagentStart`,
+`SubagentStop`, `PostToolUse` or `PostToolUseFailure` callback does is
+unmeasured; a second `initialize` on the same stdin is accepted and its hook
+table replaces the first; stdin end-of-file ends an idle CLI at once (0.02 s)
+and a busy one after its turn (a 30 s tool call ran to completion first); an
+unread stdout does not stall the CLI (the pipe's 64 kilobytes fill, the rest
+buffers inside the process, the turn completes); `--resume` takes no lock, and
+two processes on one session id both append to the one transcript; `claude --bg`
+runs an interactive session on a pseudo-terminal under a daemon that stays in
+the launcher's cgroup, and refuses `--print`, so a background session has no
+stream-json channel. `tests/test_cli_control_protocol_probe.py` re-checks the
+three facts that need no model call (the second initialize, the idle exit on
+stdin end-of-file, the `--bg` refusal) when run with `ROMP_CLI_PROBE_LIVE=1` and
+a `claude` on PATH; it skips otherwise, as every test that would reach the live
+CLI must.
 
 Who owns a running CLI is a lease, not its parent process. The kernel writes
 `leases/<sid>.json` under the state directory the moment the SDK connect hands
--- a/tests/test_cli_control_protocol_probe.py
+++ b/tests/test_cli_control_protocol_probe.py
@@ -11,13 +11,20 @@ with the probe described in docs/reference.md ("What survives a restart"):
   * `--bg` (a background session) and `--print` (the only mode that takes `--input-format
     stream-json`) are refused together, so a background session has no structured channel.
 
-Every test here runs the REAL CLI, so the suite's standing rule applies (tests/conftest.py floors
-ROMP_CLAUDE_BIN at /bin/false: no test reaches the live CLI unasked). These run only when the
-developer opts in with ROMP_CLI_PROBE_LIVE=1 and a `claude` is on PATH (or named in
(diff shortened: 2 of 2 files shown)


ksarma:
Merging when the Tier policy check reports. Nothing to fold in.

Reviewed inline: `hermetic_env` drops every `ANTHROPIC_*` name by prefix, the codex judge's idiom in `kernel/judge.py`. conftest pops only two of those names, so the gap was real under pytest and wider under a bare run. Every `os.killpg` target is the pid of a Popen the module started with `start_new_session`. `_gone` probes with signal 0 a pid the test's own sh printed, so nothing is killed by lookup. `run_probe` records the Popen before `communicate` and re-raises `TimeoutExpired` after the group kill. The `--bg` test's four assertions are unchanged and read rc and stderr only. The docs sentence holds by reading: the module registers PreToolUse on Bash alone. romp registers UserPromptSubmit and the five kinds marked unmeasured, with no timeout on any HookMatcher. The installed 2.1.266 CLI turns a timed-out UserPromptSubmit or UserPromptExpansion callback into a blocking decision that suppresses the prompt. It rethrows every other event's timeout. Its `--bg` check is an argv scan for `--print`, so the dropped prompt cannot change the refusal. The word-diff shows the two sentences alone. The live pair asserts three facts. The privacy scan of the 215 added lines is clean.

One follow-up:

- `tests/test_cli_control_protocol_probe.py:256`: `HermeticEnvPins` asserts only absences and two paths under the root. A filter that dropped everything would pass it, and the live pair that would notice skips on CI. One planted unrelated name, asserted present, closes that.

Ran here in a capped scope with every `ANTHROPIC_*` name removed. The module at the head: 5 passed, 2 skipped under pytest, and 7 tests with 2 skipped as a bare unittest run on two Pythons, also under `-n 2` and with a pipe on stdin. The base module with the new classes spliced in: 5 failed, 2 skipped, for the reasons the body gives. Four mutations (no `start_new_session`, no pipe close, no tearDown kill, inherited stdin) each turn the named tests red. No sleeper or sh survived. CI at 9de40748: CI and PR tier are green. The Tier policy runs have been queued since the PR opened, as has every Tier policy run in the repo since 23:26Z. A backlog on that workflow, not anything here.

Review run by the maintainer automated review session; verified against head 9de40748.

ksarma:
Head note: the branch was updated with main, so the head moved from 9de40748 to fcdd3233. The new head is a mechanical merge of 9de40748 and 1af162be: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head fcdd3233.

PR #1375, opened 2026-09-10
Lease heartbeat: the beat loop is stepped by a test through sleep and clock seams; a refused heartbeat thread leaves the session unleased instead of crashing its connect

ksarma:
Two changes in the lease heartbeat: the beat loop takes sleep and clock seams and is executed by a test, and a heartbeat thread the OS refuses no longer crashes the session's connect.

## Symptom

A session whose lease heartbeat thread the OS refuses to start dies at its connect. `Thread.start()` runs in `_lease_open`, inside the connect and after the handshake (`connected = True`), with no guard. CPython reports the refusal as `RuntimeError("can't start new thread")`; the connect's handler re-raises it, `_run` logs "sdk session X crashed", and `_on_session_gone` heals away a CLI that was up and answering. Rare (an OS thread limit), but the cost is a working session.

Separately, the heartbeat loop `_lease_beat_loop` was executed by no test. It called `time.sleep(LEASE_HEARTBEAT_S)` directly, so the three tests that reach it all patch it away, and its exit handshake (a census of 0, then the re-check of the held leases under the lock before the thread slot is cleared) had no pin. The post-merge review of #1342 named both as follow-ups.

## Cause

- `_lease_open` started the heartbeat thread under the lock with nothing around `start()`, and the connect path has no handler between that call and the one that crash-heals.
- `_lease_beat_loop` had no seams: the sleep and the clock were the module's `time.sleep` and `time.time` by name, so the loop could not be stepped without three real seconds per beat.

## Fix

`kernel/sdk_backend.py`, two hunks:

- `_lease_beat_loop(self, sleep=None, now=None)`: the seams resolve at call time (`sleep = sleep or time.sleep`, `now = now or time.time`), the shape `_end_cli_tree` uses, and the beat is `self._lease_beat_once(now=now())`. The thread target still passes nothing, so the kernel beats on the real clock; the exit handshake is unchanged.
- `_lease_open`: the Thread is created into a local and `start()` sits in a `try`. On `RuntimeError` or `OSError` this session's entry is popped from the held leases and the thread slot is set to None, both under the lock; after the lock is released a problem row says the heartbeat thread could not start and the session runs unleased and a boot judges it by parentage, and `_lease_open` returns before `write_lease`, so no file is written. That is the posture the missing-pid branch already takes, reused rather than a new `lease.*` anomaly kind (which would need a `LEASE_ANOMALIES` entry and a line in the reference's lease paragraph). On success `_lease_thread = t`, as before.

A choice to weigh: after a refusal only this session's entry is dropped. Other held leases stay in `_leases` with no heartbeat until the next `_lease_open` retries the start (the slot is None, so it does). Today nothing reads a beat while the kernel lives (`lease_state` runs from `_boot_reconcile`, when this kernel as holder is gone anyway; `find_session_cli` checks pid and start identity only), so this costs nothing now; a per-session host that reads beats in-life may prefer every held lease dropped, or the start retried on its own. Said here so the choice is visible.

## Tests

`tests/test_sdk_lifecycle_hardening.py::LeaseRules`, two tests. Neither sleeps for real, signals a process or spawns a CLI: the client is a `SimpleNamespace` whose pid is the test's own, and the only real threads are no-op recorders joined with a timeout.

- `test_the_heartbeat_loop_beats_on_its_cadence_exits_on_an_empty_census_and_restarts_at_the_next_open` runs the real loop on the test thread through the seams: a beat on the cadence stamped from the loop's clock (not `time.time()`), the exit on an empty census handing the thread slot back, an open landing between the beat's 0 and the lock's re-check keeping the loop alive and starting no thread, the restart at the next open under the thread's name, and the defaults resolving at the call to the module's `time.sleep` and `time.time` (a stand-in for the module's `time` name is patched, this module only, never the process-global functions; the recorder counts the test thread only and hands any other thread the real sleep). Whether the loop returns never rests on which sleep it called: the close that empties the census sits on the beat side, and every beat checks that one recorded sleep preceded it, so a loop that slept elsewhere (the seam bypassed, a default bound at the definition) fails at its first beat, one real beat later at worst, instead of hanging the run; a loop that never exits fails on the recorder's count. Before the change: `TypeError: _lease_beat_loop() got an unexpected keyword argument 'sleep'`.
- `test_a_heartbeat_thread_the_os_refuses_leaves_the_session_unleased_with_a_problem_and_no_raise` opens a second session under the no-op loop and joins its thread, so a lease is held and a finished thread sits in the slot, then patches `threading.Thread` for one `_lease_open` call with a subclass whose `start()` raises, once `RuntimeError("can't start new thread")` and once an `OSError`: no raise, no lease file for this session, the other session's entry and file untouched, the slot emptied, one problem row each naming the refusal and "runs unleased"; then the real Thread starts at the next open and the lease is written (the refusal is not latched). Before the change: the `RuntimeError` escapes `_lease_open` and the test errors.

Both fail before the change and pass after; the module (63 tests) is green alone and under pytest-xdist. Checked by mutation, each read in seconds: the slot reset or the pop dropped from the refusal branch (a `clear()` in its place), the guard or the early return removed, either default bound at the definition, either seam bypassed, the re-check under the lock dropped, and a loop that never exits. Full suite on this head (`pytest -n 6 tests/`, the served-page modules executing): 11672 passed, 45 skipped, 1744 subtests passed, 1 failed. The failure is `tests/test_paste_to_focus_browser.py::ServedPaste::test_a_trusted_paste_after_selecting_transcript_text_lands_in_the_box`, a served-page browser test of the chat pane's paste that reads none of the two files this change touches; run alone right after, it passes (1 passed in 9 s).

## Tier

Tier: fix

### DIFF
--- a/kernel/sdk_backend.py
+++ b/kernel/sdk_backend.py
@@ -8742,14 +8742,30 @@ class SdkBackend:
         lease = {"sid": str(sess.sid), "fsid": str(sess.resume_sid or sess.sid), "name": str(sess.name),
                  "pid": int(pid), "start": start, "holder": self._lease_holder(), "version": self.code_version,
                  "spawnedAt": int(now), "t": now}
+        refused = None
         with self._lock:
             self._leases[str(sess.sid)] = (sess, lease)
             if self._lease_thread is None or not self._lease_thread.is_alive():
                 # started UNDER the lock: an unstarted thread reads as not alive, so a second session
                 # connecting in the first's write window would otherwise adopt it too and both would call
                 # start() on one Thread (RuntimeError out of the connect; the review of T305, 2026-09-10)
-                self._lease_thread = threading.Thread(target=self._lease_beat_loop, name="sdk-lease-beat", daemon=True)
-                self._lease_thread.start()
+                t = threading.Thread(target=self._lease_beat_loop, name="sdk-lease-beat", daemon=True)
+                try:
+                    t.start()
+                except (RuntimeError, OSError) as e:   # the OS refused a thread ("can't start new thread")
+                    # Raised from here, inside the connect and after the handshake, the connect's handler
+                    # would crash-heal a CLI that is up and answering. This session runs unleased instead,
+                    # the missing-pid posture above: its entry goes, so no beat ever claims it, and the
+                    # slot stays empty so the next open tries the start again. Said outside the lock.
+                    self._leases.pop(str(sess.sid), None)
+                    self._lease_thread = None
+                    refused = e
+                else:
+                    self._lease_thread = t
+        if refused is not None:
+            self._log("lease (%s): the heartbeat thread could not start (%s); the session runs unleased and a boot "
+                      "judges it by parentage" % (sess.name, refused), problem=True)
+            return
         try:
             write_lease(self.state_dir, lease)
         except Exception as e:
@@ -8785,12 +8801,16 @@ class SdkBackend:
                 self._log("lease (%s): heartbeat write failed: %s" % (sess.name, err), problem=True)
         return len(items)
 
-    def _lease_beat_loop(self) -> None:
+    def _lease_beat_loop(self, sleep=None, now=None) -> None:
         """The heartbeat thread: one beat every LEASE_HEARTBEAT_S while any lease is held; exits when
-        none is (the next _lease_open starts a new one)."""
+        none is (the next _lease_open starts a new one). `sleep` and `now` are the test seams, resolved
+        at call time as _end_cli_tree's are, so a test steps a beat, the empty exit and the restart with
+        no second behind them; the thread target passes neither, so the kernel beats on the real clock."""
+        sleep = sleep or time.sleep
+        now = now or time.time
         while True:
-            time.sleep(LEASE_HEARTBEAT_S)
-            if self._lease_beat_once() == 0:
+            sleep(LEASE_HEARTBEAT_S)
+            if self._lease_beat_once(now=now()) == 0:
                 with self._lock:
                     if not self._leases:
                         self._lease_thread = None
--- a/tests/test_sdk_lifecycle_hardening.py
+++ b/tests/test_sdk_lifecycle_hardening.py
@@ -454,6 +454,141 @@ class LeaseRules(unittest.TestCase):
         self.assertIsNotNone(sb.read_lease(d, a.sid))
         self.assertEqual(be._lease_beat_once(), 1)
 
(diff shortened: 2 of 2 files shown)


ksarma:
Merging when CI is green. Nothing to fold.

Reviewed inline: both kernel hunks. The loop's sleep and clock seams resolve at call time. The beat passes `now()` into `_lease_beat_once`'s existing `now` keyword, which stamps the lease. The thread target still passes nothing, and the exit handshake is unchanged. In `_lease_open`, `start()` is guarded. On `RuntimeError` or `OSError`, this session's entry is popped and the slot set to `None` under the lock. The problem row is logged after the lock is released. The function returns before `write_lease`, matching the missing-pid branch's behavior and wording. The symptom holds: `_lease_open` runs inside the connect after the handshake, so an unguarded refusal reached the crash-heal path. No docs change is owed. The reference's lease paragraph lists no unleased causes and already covers a CLI without a lease under a live kernel. One choice for the owner: after a refusal, the other held leases stop receiving beats until the next open retries the start. Nothing reads a beat while the kernel lives (`lease_state` runs from `_boot_reconcile`; `find_session_cli` compares pid and start), so this costs nothing today. A host that reads beats while the kernel lives may want every held lease dropped, or the start retried on its own.

- `tests/test_sdk_lifecycle_hardening.py:572`, follow-up: the refusal test patches `sb.threading.Thread`, the process-global class, for one synchronous `_lease_open` call. The loop test in the same PR builds a module-local stand-in to avoid that. The window cannot hang; this is a consistency point, not a defect.

Red-first reproduced at base feb57836 with the head's test file, the two tests selected by `-k` after a collect-only gate (2 of 63). The loop test fails with a `TypeError` on the sleep keyword; the refusal test fails with the `RuntimeError` escaping `_lease_open`. At head: 2 passed, 61 deselected in 0.1 s. Each of ten mutations of `kernel/sdk_backend.py` was caught by the named test in under 4 s, and none hung. The mutations: pop, slot reset, guard or early return removed; a default bound at the definition; a seam bypassed; re-check dropped; loop never exits. Added lines carry no private strings, hostname or home paths. CI at af395d73: Python 3.10, 3.12, bats, gitleaks and the tier checks are green. Python 3.11, 3.13 and the vscode-extension job are still pending.

Review run by the maintainer automated review session; verified against head af395d73.

ksarma:
Head note: the branch was updated with main, so the head moved from af395d73 to 915092a6. The new head is a mechanical merge of af395d73 and 1af162be: its tree equals the merge-tree of its two parents, and its added and removed lines against the merge base with main are identical to the reviewed diff. The review above stands for this head.

Review run by the maintainer automated review session; verified against head 915092a6.

PR #1376, opened 2026-09-10
Chat: a copied mention chip pastes the @name as typed, and a remote chip's host prefix reads on its dark backing in the light theme

ksarma:
## Symptom

- Select a sent message that holds a mention chip and copy it, with Ctrl+C or the right-click menu's Copy: the paste reads "ask api" for a message that was sent as "ask @api". Pasted back into the composer, the word no longer names the session (the picker put the @ there so that it would).
- In the light theme, the "host:" prefix on a chip for a session on another machine, on the mention chip and on the Awaiting chip alike, is hard to read on the chips' dark backing.

## Cause

- The chip's text is the bare name, by design since the chip took the Awaiting chip's look; the typed @name rides `data-token`, and nothing reads that attribute on a copy. There was no copy listener, so the browser's own copy and the selection menu's `sel.toString()` both wrote the rendered text.
- The only colour reaching a chip's host prefix was the global `.host-prefix { color: var(--dim) }`. `--dim` is tuned for the page, and the light theme's #5D574E on the chips' theme-invariant 85% black backing reads 2.66:1 on the user bubble and 2.51:1 on the Awaiting chip (WCAG asks 4.5:1 for text and 3:1 for large text). The dark theme's #9a9a9a reads 6.7:1 and 6.4:1 on the same surfaces.

## Fix

- `ui/webview/render.ts`: `mentionCopyText(sel)` finds every chip the selection covers whole (Range.comparePoint on the chip's first and last text node, so a double-clicked chip word counts and a run of letters inside a chip does not), inserts "@" on each chip's first text node, reads `sel.toString()` and each range's `cloneContents()`, then deletes the "@" again. insertData and deleteData move a live range's offsets and move them back, so the DOM and the visible selection are unchanged after the copy. The rich flavour drops each chip's hover title (live status text) and its data attributes, so a paste keeps the chip's class and text and nothing about the session behind it. A document `copy` listener in the chip block writes text/plain and text/html from it, and returns before touching the event when no chip is whole, so every other copy stays the browser's in both flavours, including a copy inside the composer. `showSelectionMenu`'s Copy reads the same text. `transcriptSelection` (the Comment/Quote seed) is untouched: a thread's quoted passage anchors on the rendered text.
- `ui/webview/styles.css`: one rule beside the shared backing rule, `.chip-peer-name .host-prefix, .mention-chip .host-prefix { color: rgba(255, 255, 255, 0.7); }`. The backing is the same in both themes, so the colour on it is too: 9.5:1 on the user bubble and 9.1:1 on the Awaiting chip in both themes, still visibly quieter than the #fff peer name. Not `color: inherit` (the file viewer pill's idiom): on the mention chip that is the identity colour, and the palette's darker swatches are already the faintest text on this backing (a separate item, not changed here). No new token: the need is not theme-conditional. The global `.host-prefix` rule is unchanged, so its pins in host-prefix.test.ts, palette.test.ts and dense-chrome.test.ts keep matching; feed.css holds no chip rules, so no mirror is needed.
- `docs/guide.md`: the sent message still carries the @name you typed, and so does a copy of it.

## Tests

In `ui/webview/composer-mention-pane.test.ts` the file goes from 14 to 18 tests; one pin in `tests/test_spend_detail.py` is re-anchored.

- Chromium: copying a selection that holds a chip copies the @name as typed; a selection with no whole chip is left to the browser. A synthetic `ClipboardEvent("copy")` with its own `DataTransfer` over a bubble reading `ask @web and @api, not <code>@api</code>`: the whole bubble gives "ask @web and @api, not @api" as text, and as html each chip's text reads `>@web</span>` and `>@api</span>` with the code span as rendered, the chip's class kept and no `title=` or `data-` attribute; the DOM, the chips' text and token, and the selection are unchanged afterwards; the code span alone is not prevented and nothing is written; the api chip's text node 0..3 (a double-click) gives "@api" and the range is 0..3 again; 1..3 is left alone; a remote chip gives "ping @TESTHOST:web", with the @ on the host prefix span in the html; a copy inside the focused composer is left alone. Before the change it fails on the first assertion: `false !== true`, the copy is not taken. Two deliberate breaks of the fixed code were also checked against it: serializing the html before the @ insertion fails the html assertion, and dropping the attribute strip fails the `title=`/`data-` assertion.
- Chromium: a remote chip's host: prefix reads on the shared dark backing in both themes, on the user bubble and on the awaiting chip. Injects the real stylesheet (minus its `@import` line, which only the bundler can resolve), builds a mention chip through `markMentions` and an Awaiting peer name through `hostPartsNodes`, toggles `body.theme-light`, reads the computed colours and composites prefix over backing over surface. Before the change: `light theme, the mention chip: the host prefix rgb(93, 87, 78) on the backing over rgb(165, 90, 42) reads 2.66:1, below 4.5:1`.
- Pin, runs on CI without Chromium: the scoped rule exists after the shared backing rule, sets neither `inherit` nor `var(--dim)`, and its colour composited over the backing over each theme's `--bg` clears 4.5:1; the light theme's `--dim` on the same backing is under 4.5:1, recorded as the reason (if the light dim is ever lifted past it, the rule can retire). Before the change: the rule is absent.
- Pin: the copy listener lives in the sliced chip block (so the Chromium case runs the real one) and the selection menu's Copy reads `mentionCopyText(sel)?.text ?? sel.toString()`. Before the change: the listener is absent.
- `tests/test_spend_detail.py`: the landing page inlines a twin of the strip's host-prefix declarations, and the pin that compares the twin with the sheet found the first `.host-prefix {` text in `styles.css`. The new chip rule ends in the same selector and sits earlier in the sheet, so on this branch the pin read the chip rule's colour and failed (the full pytest suite's one failure before this change to the test). It now anchors on the rule that starts its line and reads the global rule again; its module is 23 passed.
- The Chromium cases skip by name on CI, whose test step runs before its Chromium install, as every browser case in this file does; both executed locally. `npm run typecheck` is clean. The twelve test files that pin the sheet's chip, prefix, colour and theme rules pass (168 tests). Full `npm test` on this head: 4319 tests, 4319 pass, 0 fail (4297 in the main run plus the 22 of `ui/timeline-tags-scale.test.ts`, run on its own). Full `pytest -n 6 tests/` on this head, the served-page modules executing: 11670 passed, 45 skipped, 1744 subtests passed, 1 failed. The failure is `tests/test_notification_tap_resume_browser.py::ServedTapLanding::test_a_click_the_worker_saw_lands_the_page_via_the_message_and_settles_the_row`, whose kernel-trail pin accepted only the `delivered` wording of the tap's reveal; main widened that pin to the parked-then-consumed road in #1363, after this branch's base, and the test reads none of the files this change touches. It fails the same way alone on this base and passes with main's pin.

Tier: fix

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -67,7 +67,8 @@ without inserting, and it stays closed for that `@` until you delete it: more le
 caret move away and back, do not reopen it. In the sent message, a name that matches a live
 session is shown as a chip: the name without its `@`, in that session's color on a dark
 backing, the way the Awaiting chip names the session it waits on. Hover it for how that
-session is doing; the message itself still carries the `@name` you typed.
+session is doing; the message itself still carries the `@name` you typed, and so does a
+copy of it.
 
 **A message that has not gone yet.** Send to a busy session and your message waits as a
 dashed bubble under an hourglass until the session takes it — while it compacts, while a
--- a/tests/test_spend_detail.py
+++ b/tests/test_spend_detail.py
@@ -658,7 +658,9 @@ class SpendDetail(unittest.TestCase):
         import re as _re
         css = open(os.path.join(os.path.dirname(HERE), "ui", "webview", "styles.css")).read()
         def decls(sel):
-            m = _re.search(_re.escape(sel) + r"\s*\{([^}]*)\}", css)
+            # the rule that STARTS a line: a descendant rule such as `.mention-chip .host-prefix` also ends in
+            # the selector and may sit earlier in the sheet than the global rule this twin mirrors
+            m = _re.search(r"(?m)^" + _re.escape(sel) + r"\s*\{([^}]*)\}", css)
             self.assertIsNotNone(m, sel + " missing from styles.css")
             return _re.sub(r"\s+", "", m.group(1)).rstrip(";")
         self.assertIn(".rsp-name .tab-label.colored{" + decls(".tab.colored .tab-label") + "}", js,
--- a/ui/webview/composer-mention-pane.test.ts
+++ b/ui/webview/composer-mention-pane.test.ts
@@ -5,7 +5,7 @@
 // stack. Skips by name when playwright or its browser is missing (md-sanitize-browser.test.ts's idiom; CI's
 // test step runs before its Chromium install, so the browser cases skip there). Source pins hold only the
 // wiring the slices cannot carry: the one clear helper every clear site calls, the keydown order, the roster
-// hook at the top of renderTabs, the transcript hook.
+// hook at the top of renderTabs, the transcript hook, the selection menu's Copy, the chips' stylesheet rules.
 // Synthetic names only (romp, rompdocs, web, api, TESTHOST; placeholder ids).
 import { test, after } from "node:test";
 import * as assert from "node:assert/strict";
@@ -16,6 +16,7 @@ import { createRequire } from "node:module";
 const PKG = process.cwd();                                   // vscode-extension, where npm test runs
 const UI = path.resolve(PKG, "..", "ui", "webview");
 const RENDER = fs.readFileSync(path.join(UI, "render.ts"), "utf8");
+const STYLES = fs.readFileSync(path.join(UI, "styles.css"), "utf8");
 const req = createRequire(path.join(PKG, "package.json"));   // runtime requires: esbuild must not bundle playwright
 
 let pw: any = null;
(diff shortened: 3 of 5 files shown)


ksarma:
Merging when the Tier policy check clears. Nothing to fold in; two follow-ups are noted below.

Reviewed inline: `mentionCopyText` counts a chip as whole only when `Range.comparePoint` returns 0 at its first text node's start and its last text node's end. So a double-clicked chip word qualifies, and a run of letters inside a chip does not. The "@" goes in with `insertData` and comes out in a `finally`. The live-range adjustment leaves the selection unchanged. The transcript's MutationObservers watch `childList` only, so the momentary edit triggers no rebuild. When no chip is whole, the copy listener returns before touching the event, so the composer and every other copy stay the browser's default. The selection menu's Copy reads the same text, and `transcriptSelection` is unchanged. The descendant rule has specificity 0,2,0 over the global `.host-prefix`. No `body.theme-light` rule targets `.host-prefix`. I recomputed the contrast. The new colour gives 9.5:1 on the user bubble and 9.15:1 on the Awaiting chip. The light dim gives 2.66:1 and 2.51:1. The spend-detail pin needed the re-anchoring, because the old regex on the head's sheet picks the chip rule. The docs sentence is true, and the label and the Tier line agree. The added lines carry no private strings, hostnames or home paths.

Neither follow-up is a regression: the browser's own copy carried both before this change.

- `ui/webview/render.ts:16867`: the html keeps the chip's inline `style` (`dressMentionChip` sets `--chip-bg` on it). So the line's comment and the body overstate what the strip removes. Drop the `style` attribute too, or narrow the wording.
- `ui/webview/render.ts:16868`: the `title` is removed from the chip element only. When the host is down, `hostNameNodes` sets `hostDownNote` as the `title` of the inner `.host-prefix` span (`host-prefix.ts:54`). That status text remains in the html.

Red-first at the base feb57836 with the head's `composer-mention-pane.test.ts`: 14 pass, 4 fail. The failures are the copy case at `false !== true`, the contrast case with the light theme's 2.66:1 message, and the two pins on the absent rule and listener. At the head, 18/18 pass with both Chromium cases executed. Three deliberate breaks of `render.ts` each fail one assertion. Also at the head, typecheck is clean. Across the 58 compiled files that reference the chip selectors or the copy listener, 832 tests pass. The re-anchored spend-detail pin passes alone. CI at 7b519620: every CI job is green. The `Tier policy: post verdict` job is still queued.

Review run by the maintainer automated review session; verified against head 7b519620.

PR #1377, opened 2026-09-10
Gear: a greyed Fast mode box fades the checkbox alone, so its hint stays legible

ksarma:
## Symptom
With a judge tier off Opus (the defaults: Sonnet for triage, Haiku for indexing), that tier's Fast mode box in the gear's Judges section greys, and hovering the box shows why. The hint is hard to read: it renders at 40% opacity, with the rows below it (the effort picker, the next model row) drawn across it.

## Cause
`gear.css` put `opacity: .4` on the whole `.rs-fastin.rs-off` label. The hint `<span class=rs-sub>` is a child of that label, so it faded with the checkbox. Opacity below 1 also makes the label a stacking context: the hint's `z-index: 10` became local to the label, and the following `.rs-row` siblings, painted later in tree order in the same step, drew over the translucent group.

## Fix
`ui/webview/gear.css`: the label no longer carries opacity. The checkbox alone fades (`#rsettings .rs-fastin.rs-off input { opacity: .4; }`, the `#rrefresh:disabled` look), and the word takes the section grey by token (`color: var(--text-faint, #6e7681)`; the fallback is the dark value, like every other fallback in this sheet, and the light theme resolves the token to its own). The hint keeps the colour the `.rs-row:hover .rs-sub` rule sets. No markup, id or gate changes: the hints stay inside the labels, so the one-hint hover rules and `fillMixedMarks`'s label scoping are untouched, and the three tiers' boxes share the rule.

Token choice: `--text-faint` is the token nearest the raw `#6f747c` that `.rs-sec` and `.rs-mixed` use for the section grey; `--text-muted` is lighter and reads as live text.

## Tests
- `ui/webview/gear.test.ts`, the Triage-row test: the label-opacity pin becomes three pins. The input rule carries `opacity: .4`; the label rule carries no `opacity`; the label rule colours by `var(--text-faint`. Before the change all three fail (node reports the first: the input rule carried no opacity; the label held the opacity and had no colour rule).
- `tests/test_kernel_settings_sections.py::SettingsSectionsTest::test_fast_mode_for_the_judges_sits_on_the_triage_model_row`: the same three pins beside the existing selector pin. Before the change all three fail (unittest reports the first: `'#rsettings .rs-fastin.rs-off input { opacity: .4;' not found`).
- `ui/webview/gear-judge-fast-browser.test.ts`, headless Chromium over the real `gear.js` and `gear.css`: `snap()` now reads the checkbox's own computed opacity instead of the label's (0.4 greyed, 1 live), the most faded ancestor of each hint up to `#rsettings` (`hintFade`, all 1), and each label's computed colour (the greyed word's differs from a live word's). After the first snapshot it hovers the greyed box and then a live one, and checks the greyed hint shows (display block) with its own opacity at 1, in a colour other than the greyed word's and equal to the live hint's. Before the change, with Playwright's Chromium installed: the checkbox opacity reads 1 on all three tiers (the assertion expecting 0.4 fails first), `hintFade` reads 0.4 on the greyed tier, and the three word colours are equal. The hover checks pass before and after the change; they are guards: the label's colour must not reach the hint, and no rule may fade the hint itself (opacity is not inherited, so the hint's own value reads 1 either way otherwise; a rule such as `.rs-fastin.rs-off .rs-sub { opacity: .4; }` passes every source pin and fails only this check). The test skips without a Playwright browser as before; on CI the extension job runs `npm test` before it installs Playwright's Chromium, so the leg skips there and the source pins carry the gate.
- Second commit, text only: six comments and docstrings in `tests/test_judge_auth_billing.py`, `tests/test_judge_settings_sync.py`, `tests/test_kernel_judge_model.py` and `tests/test_kernel_session_flags.py` still said Fast judging, the name the gear dropped; they say Fast mode now. No assertion, count or behaviour changes.
- Full runs on this head, both with the served-page modules executing: `pytest tests/` (`-n 6`) 11675 passed, 45 skipped, 1744 subtests passed, 0 failed; `npm test` 4315 passed, 0 failed (4293 in the main run plus the 22 of `ui/timeline-tags-scale.test.ts`, run on its own); `npm run typecheck` clean.

Tier: fix

### DIFF
--- a/tests/test_judge_auth_billing.py
+++ b/tests/test_judge_auth_billing.py
@@ -264,7 +264,7 @@ class JudgeArgvBilling(_JudgeAuthBase):
         self.assertNotIn("--settings", jd._judge_cmd("sonnet", "SYS", None), "the default: no auth argument")
 
     def test_a_fast_login_billed_call_rides_one_overlay_with_both_keys(self):
-        # `--settings` takes ONE value. With Fast judging on (STATE/judge-fast "on"), an Opus login-billed call
+        # `--settings` takes ONE value. With the triage tier's Fast mode on (STATE/judge-fast "on"), an Opus login-billed call
         # carries the fastMode opt-in and the helper suppression in one JSON overlay; a login-billed call on a
         # model that cannot run fast keeps HELPER_OFF byte for byte; a key-billed or unpicked Opus call carries
         # the opt-in alone, and the helper key never appears in it.
@@ -490,7 +490,7 @@ class JudgeRunBilling(_JudgeAuthBase):
         self.assertEqual(seen["cmd"][i:i + 2], HELPER_OFF)
 
     def test_a_fast_login_pick_on_opus_launches_with_one_overlay_and_logs_the_readback(self):
-        # end to end through _judge_run: Fast judging on, a login pick, an Opus model. The child gets ONE
+        # end to end through _judge_run: the triage tier's Fast mode on, a login pick, an Opus model. The child gets ONE
         # --settings overlay carrying both keys, the login tokens, no key; the usage row keeps the envelope's
         # fast_mode_state, the CLI's own word on whether fast engaged.
         jd._LOGIN_AUTH_ENV_FN = lambda: {"CLAUDE_CODE_OAUTH_TOKEN": "synthetic-login-token"}
--- a/tests/test_judge_settings_sync.py
+++ b/tests/test_judge_settings_sync.py
@@ -74,7 +74,7 @@ class ApplySettings(unittest.TestCase):
                          "garbage never reaches the files or `claude --model`")
 
     def test_judge_fast_applies_and_reports_raw(self):
-        # Fast judging rides the same cross-kernel door as the judge tiers: "on" | "off", off by default,
+        # the judges' Fast mode rides the same cross-kernel door as the judge tiers: "on" | "off", off by default,
         # answered RAW, a value outside the pair ignored and visible in the ack
         self.assertEqual(km._apply_judge_settings({})["judgeFast"], "off")
         res = km._apply_judge_settings({"judgeFast": "on"})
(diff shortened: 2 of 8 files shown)


ksarma:
Merging once CI is green. Nothing to fold.

Reviewed inline: the fix is two rules in `ui/webview/gear.css`. The greyed label carried `opacity: .4`. That faded the hint span inside it and made the label a stacking context. The hint's `z-index: 10` was then local to the label, so the later `.rs-row` siblings painted over it. The head fades the checkbox alone, with the `#rrefresh:disabled` values, and greys the word with `color: var(--text-faint, #6e7681)`. The panel `#rsettings` is position fixed with z-index 60, and neither `.rs-row` nor `.rs-card` opens a stacking context. The hint now paints above the following rows. `--text-faint` is defined for both themes in `styles.css` and `feed.css`. Its fallback is the dark value, like every other fallback in the sheet. Markup, ids, the gate, the hover rule and gear.js scoping are unchanged. The second commit changes six comment and docstring lines in four Python test modules and no code. The fix label and the Tier: fix body line agree. A privacy scan of the 58 added lines and the body is clean.

Test evidence:
- At base 9b0b1378 with the head's test files, `gear.test.ts` fails on the input-opacity pin. `gear-judge-fast-browser.test.ts` fails with the checkbox opacity reading 1 (Chromium executed). `test_kernel_settings_sections.py` fails with the input rule not found (1 failed, 9 passed).
- At head, node on the six compiled UI tests that reference these selectors passes 44 of 44 with the browser leg run. Pytest on the five touched modules passes 130. Typecheck is clean.
- Three mutations (opacity back on the label, a rule fading the hint, no colour rule) each fail at least one test. Only the browser hover check catches the hint-fading one, as the body says.
- CI at e8f0f852: bats, the extension job and Python 3.13 green; Python 3.10 to 3.12, gitleaks and Tier policy still running.

Review run by the maintainer automated review session; verified against head e8f0f852.

PR #1380, opened 2026-09-11
Mention chips: a copied chip drops the title on its host prefix too, so a down host's reconnect note never reaches the clipboard

henrypinkard:
Follows #1376 (merged). Its copy sanitizer stripped the hover title and data attributes from the mention chip itself, but not from the span.host-prefix inside it, which hostNameNodes gives its own title (the live reconnect note) when the mentioned session's host is unreachable; that note survived into the copied text/html flavour, against the guarantee that a copy keeps nothing about the session behind it.

The strip now covers the chip's descendants too. A Chromium test stubs the federation state so a remote chip's host reads as down, copies a selection over the bubble, and asserts the copied html carries no title or data attribute while the live chip keeps its mark and note.

Tier: fix

Review find from the triage of #1376 (2026-09-10), filed as a follow-up because that PR did not allow maintainer edits.


ksarma:
Advisory review (the owner merges): no objection.

Reviewed inline: the product change is one selector in `mentionCopyText` (`ui/webview/render.ts:16981`), widened from `.mention-chip` to `.mention-chip, .mention-chip *`, plus the reworded banner. Nothing else in `render.ts` moves. The strip runs over the cloned range contents, never the live DOM; the test's after-equals-before check pins that. `hostNameNodes` (`ui/webview/host-prefix.ts:54`) is the only other title writer inside a chip. It reads `globalThis.__rompFed`, so the test's stub drives the production branch. No chip descendant carries a data attribute, and both paste listeners read `text/plain`. The wider strip therefore removes nothing a consumer needs. `text/plain` never carried the note, so the fix is html-only, as the body says. Head a3b1af4b is two merges of main over the single content commit 6b41de6a and brings no other content. Privacy scan of the 42 added lines: no hits. The `fix` label fits.

Two nits, neither a hold:
- `ui/webview/composer-mention-pane.test.ts:659` (follow-up): at the base, this class-travel match fails before the title check on 660. The surviving title sits between the class and the closing bracket, so a regression reports the wrong reason. Swapping the two asserts fixes the message.
- `ui/webview/render.ts:16952` (follow-up): the copied html still carries the chip's inline `--chip-bg` colour (set at 16867). The banner's claim that a paste keeps nothing about the session therefore overstates slightly. Pre-existing from #1376; noted because this PR rewrites the sentence.

Test evidence: typecheck is clean at head. `composer-mention-pane.test.ts` in Chromium is 19/19 at head and 18/19 with the head's test file over merge base 04c42b6d. The one failure is the new case, whose copied html still shows the prefix's reconnect-note title. CI runs `npm test` before it installs Chromium, so the browser cases skip there; the local run is the execution evidence. CI at a3b1af4b when written: eight checks green, Python 3.10 and 3.11 still running, none failed.

Review run by the maintainer automated review session; verified against head a3b1af4b.

PR #1392, opened 2026-09-11
Billing is reachable from a terminal: POST /auth and `romp billing` read or switch what a session bills, and a session never picked says so

lczh:
Tier: feature

## What this adds
A way to set which account a live session bills, login or API key, from outside the dashboard, and an honest word for sessions that never picked.

## Why
The per-session billing pick exists today only as a WebSocket message the tab menu sends. A terminal, a script, or another machine's plugin has no way to set it, and no way to ask. Long-running sessions created before the pick existed carry no pick at all; their tab menu shows the machine's default checked as if someone had chosen it. On a box with both a login and an API key, that means several managers have been billing the key with no control to move them and no sign that nobody decided.

## Design points
- **One route, the gear pick's semantics exactly.** `POST /auth` with `{"id"|"name", "value": "login"|"key"}` sits beside `/fork` under the same token gate and goes through the same helper the tab menu uses. A switch that cannot apply yet is parked in the same queue as a model or effort change and answered as queued, truthfully, from the park's own decision: the session is mid-turn, compacting, being moved, held back by a usage limit, or has changes queued ahead. Before anything is parked, the route asks whether the switch can apply at all, and refuses at once with the backend's own sentence when it cannot: no login signed in, a managed helper in the way, no API key configured, not an SDK session, or a session that has ended. Only a switch that could apply is parked when the session is mid-turn, compacting, held, or has changes queued ahead, and the answer says queued. A dormant session, alive but not running, takes the pick into its record and applies it on its next resume. A remote session is forwarded the way `/send` is, and a dead tunnel answers with the host named rather than a false ok. Without a value, or as `GET /auth?id=`, the route only reads. Nothing in any answer, log line or CLI line carries a fragment of a key.
- **`romp billing <session> [login|key]`.** Bare, it prints one line: which account the session bills, whether a switch is still applying, which login account, or that the session is unpicked and bills the machine's default. With a value, it posts the switch and echoes the kernel's answer. A refusal prints the reason and exits 1, an unknown billing state exits 1 rather than pretending, and a kernel that predates the route is named as such, the way `romp api-health` does it.
- **Unpicked is said, not implied.** Status gains one field, whether the session's record carries an explicit pick. The tab menu's Billing submenu shows "unpicked (bills the API key)" with no current mark for such a session, or "unpicked (the CLI reports the login)" when the CLI's own report differs, and the hover row says the same in one line. An unpicked session is one whose record carries no pick: created before the picker existed, or when no remembered default stood. Clicking either side of the flyout on such a session makes it an explicit pick, including the side it already bills; a picked session's current option still only dismisses. What a pick does is unchanged; only how it is reached and how its absence is described.

## Deliberately left out
A remote session addressed by name still resolves locally, as the send and end routes do today. The sessions listing is unchanged. No row was added to the docs' command table, which open PRs are editing; the billing section itself gains the route, the command and the unpicked case.

## Tests
Twenty-two route cases drive the real request handler over loopback with only the host lookup and the remote forward stubbed: a quiet switch, a parked one through the real queue, a read with no side effect, a bad value, an unknown name, a malformed body, a missing id, a token-less request, a busy tmux target and a busy Codex-shaped backend refused with nothing parked, an unbillable side refused before parking, an ended session refused for both switch and read with its record and the defaults untouched, a backend refusal, a remote forward and a dead tunnel. Thirteen bats cases drive the real `romp billing` against a stub kernel: bare status in each state, a switch, usage, the token refusal, the stale-kernel line and the unknown-state exit. The menu case renders the unpicked sub-line through the real builder. On `main` the route answers 404, the command is unknown, and the sub-line is absent. Three existing pins were updated where their text moved. The modules pass 22, 8, 73, 7, 13 and 150; the extension suite passes except one layout case that fails identically on `main`.

Screenshot: the tab menu's Billing submenu for an unpicked session, posted alongside.

### DIFF
--- a/bin/romp
+++ b/bin/romp
@@ -677,6 +677,7 @@ EOF
     _romp_cmd "romp perf client [--minutes <n>|--json]" - "Browser-side performance of the open dashboards (client-diag.jsonl, surface perf): handler ms/min by frame type with window p50/p90/p99 and max, main-thread-free p90, long animation frames and their attributed callbacks, the worst minute, heap and DOM, the slowest frames — per dashboard and pane over the last <n> minutes (default 10)"
     _romp_cmd "romp api-health"            -            "The API-health signal across your sessions (GET /api-health as JSON: per-key/family retry rates and a thrash state)"
     _romp_cmd "romp fork <parent> <name>"  -            "Branch a session into a new parallel one (parent untouched; --at <uuid> picks the cut)"
+    _romp_cmd "romp billing <session> [login|key]" -    "Which account an SDK session bills (login or API key): bare reads it, a value switches it (mid-turn it waits for the turn to end)"
     _romp_cmd "romp rename <session> <name>" -          "Rename a session (uuid-keyed: mailboxes, goals and history all follow)"
     _romp_cmd "romp move <session> <dir>"   -           "Move an SDK session's working directory (its conversation follows; mid-turn it waits for the turn to end)"
     _romp_cmd "romp color <session> [<#hex|slot>]" -    "Recolor a session's identity swatch (palette hex or slot number; no color arg reads bg/fg)"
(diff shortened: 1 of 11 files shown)


lczh:
Closing: the need this served, switching billing for sessions that never picked, is met by the tab menu on current main; the route, CLI and unpicked label are on offer later if wanted.

PR #1426, opened 2026-09-11
Chat: a hot key per tab, hot keys that cycle the split's columns, and the session bell as a command

amith-ananthram:
## What this adds

Four keyboard and tab-strip capabilities inside the chat's existing model, on top of the split (#1126):

- **A hot key per tab.** Right-click a tab, pick **Hot key…**, press a combination: the shell's shortcuts recorder opens on that one row, the chord binds, and the tab shows it minified after its name. Pressing it switches to that session, in the column that holds it. Once bound, the row reads **Update hot key…**: the recorder re-records, or its **Remove** button (Backspace too) drops the command, the set entry and the binding together; Escape keeps what was there. A chord the shell owns (the Alt+Arrow pane focus) is refused on the row, and says so. The set of sessions with a hot key persists per browser with their names; their commands (`session.hotkey.<sid>`) register from it at boot, a departed tab leaves the set, a renamed one re-titles.
- **Hot keys that cycle the split's columns.** **Focus the next chat column** and **Focus the previous chat column** are bindable commands; they wrap.
- ~~Pinned tabs~~ — **dropped 2026-09-13**: the tab lock (T395, in the strip's gear menu) covers the wish of keeping tabs from shifting, all at once rather than per tab, and the maintainer prefers one mechanism. Nothing of the pin store, its menu row or its slot-holding remains in this branch.
- **Toggle notifications for this session** is a palette command, bindable like the rest: it flips the active session's bell (the tab menu's **Notify me**) in the column last worked in and says which way it went. The served story found that a bell flipped in one window reached a caught-up client only with its next full frame, so the per-session flags now ride the tail frame beside the status; every window, and another browser, sees the flip at once.

## Design points

- Hot keys are ordinary commands in the shared bindings store, recorded in the same dialog (opened on a single row); nothing ships bound, the user records their own.
- The tab badge is the chord at its shortest (glyph modifiers, no separators), with the full spelling in the tooltip and for a screen reader.
- The per-tab commands are per browser, mirroring the tab arrangement's scope.

## Rebased onto main of 2026-09-13

Main moved under this branch (the chat columns' partition, the T379 tab widgets, the T395 tab lock, the uuid-anchored chat wire), so the rebase is one squashed commit re-derived against it:

- **The keycap is T379's.** Main's hot-key widget already reads the same `romp:tabkeys` set and `session.hotkey.<sid>` override this branch writes, so the tab's keycap is the widget's: the pane's own badge, its `.tab-key` rule and its chord helpers are gone, and the tab menu's row spells the chord with the widget's `miniChord`.
- **One column per session.** A hot key switches to the column that holds the session (`__rompChatTarget`).
- **The flags ride both wires.** A flag-only change (the bell, the feed mute, the mail mute) is an empty-suffix `chatTail` on the index client's wire *and* on the uuid-anchored one (`_send_chat_proto2`); the served story's second window found that only the first carried them.
- **Menu placement.** The partition retired the "Open in new split" row, so **Hot key…** sits with Rename, ahead of the colour swatches.
- **The served story** (`tests/test_chat_split_served.py`) re-derives the hot-key and bell steps for the partition, including a three-column cycle and another browser window learning the bell from the kernel; `tests/test_kernel_delta_send.py` pins the proto-2 tail's flags.

- **A short window.** The tab menu ran past a 300px-tall window's edge in the Billing lab (main's own Billing and Browse files rows already sat below it there; the hot-key row adds one more). The menu now clamps to the window and scrolls, mirrored in `feed.css`, and a scroll inside the menu no longer counts as the dismissal that closes it.
- **CI.** The served split story waits for a new column's tab to be draggable before dragging it back and records the tab's state if the drop zone never appears (a CI-only timeout left no trace); its time budget allows for the added steps.

## Review round (2026-09-14, an independent pass over the diff)

- **A stale frame could flip a just-toggled bell back.** The tab menu's rows and the bell command set the flag optimistically and tell the kernel; a frame the pusher built before the kernel applied the click could land after it with the old value, and since this branch every tail carries the flags. `flag-pending.ts` holds the click's expectation against such frames: the echo clears it, three disagreeing frames yield to the kernel, a `settingRefused` drops it at once (tab-meta.ts's label/colour guard is the precedent). Executed tests plus source pins on the three render.ts sites.
- **An unbind in the full Keyboard shortcuts dialog left a chordless "Switch to …" command registered.** The prune that keeps the hot-key set honest stands down while the dialog is up, and the regular close never ran it. The dialog gains `onClose`, run after every exit, and the shell prunes there.
- Stale comments and anchors: the keycap widget's header no longer claims nothing writes its store; a test whose block anchor no longer existed (it spanned to end-of-file) is re-anchored; the duplicate `.tab-key` CSS pins and the guide's glued paragraph are fixed.

**CI note.** Two served suites unrelated to this branch went red on the PR's runs: `test_dashboard_reload_served` (also red on other PRs' runs that day) and `test_rail_relative_served`, a lab that landed on main after this branch's base. The rail lab fails on a pure `main` checkout as well when run late in the UTC day (2026-09-14 ~00:30 UTC: `'5 min ago' != '2 hours\nago'`, the same assertion the PR's runs hit at 23:50 and 00:30 UTC, while main's own runs earlier in the evening passed), so it looks time-of-day dependent rather than anything here. The branch's own suites are green locally and on CI.

## Tests

- UI: `tab-keys.test.ts`, `tab-menu-scroll.test.ts`, `shortcuts-modal.test.ts`, `keybindings.test.ts`, `commands.test.ts`, `palette.test.ts` (the bell command, the handler on the live session only, the tail's flags applied); `flag-pending.test.ts` (the view flags' pending guard, executed); pins in `tab-strip-skip`, `composer-placeholder`, `reload-notices`, `awaiting-box-sync` and the extension's `tab-menu-flags` follow.
- Python: `tests/test_kernel_delta_send.py` pins the flag-only tail (an empty suffix carrying the flags; the flip alone is a new frame). `tests/test_chat_split_served.py` runs the split story in a real browser against a hermetic kernel: the keycap; the chord going to the column that holds the session; the cycle both ways across two and three columns; the recorder opening on one row from the tab menu, refusing a built-in chord, binding the pressed one and handing the keyboard back; Remove and Escape; the bell command from inside a column (Ctrl+P, type, Enter): the toast, the menu row flipping in that column at once and in the other from the kernel's push, both ways, and the kernel's flags file.
- `docs/guide.md`: the keyboard paragraph.

Features:
- allow attaching hot keys to individual session tabs
- add hotkey support for turning notifications on for a particular session
- add hotkey support for swapping between split panes

Screenshot 2026-09-11 at 3 13 34 PM
Screenshot 2026-09-11 at 3 13 47 PM
Screenshot 2026-09-11 at 3 14 02 PM
Screenshot 2026-09-11 at 3 14 28 PM


amith-ananthram:
Third commit, c9d4a1ac, from review: a pinned tab now wears the tab menu's pushpin (one drawing for the menu row and the tab, after the hot-key badge) instead of the folded corner; and the menu's "Change hot key…" / "Remove hot key" rows are one "Update hot key…" row, whose recorder offers a Remove button beside the Backspace hint. The served story removes a hot key through that door; the pins and the guide follow.

amith-ananthram:
Fourth commit, a4f2e8fc: the bell command's flash reads "Notifications enabled for <session>" / "disabled", and once the command has a key, the tab menu's Notify me row shows it after its sub-line.

henrypinkard (approved):
Approved on the maintainer's word; the two review nits (a quiet confirmation for the bell command, a Remove button visible without hover) were landed on the branch by the maintainer's session.

PR #1499, opened 2026-09-12
Chat: a path in a fenced code block links too, on the kernel's verdict and under the viewer's code-aware gate

amith-ananthram:
## What this adds

A path a session prints inside a fenced ``` block now links in the chat, the way the same path in prose or in inline backticks does. Until now the chat's walk skipped fenced blocks outright, a rule that was never written down and that predates two safeguards: the kernel's check that a mentioned path is a real file (`pathLinks`), and the file viewer's code-aware shape rules for paths inside code.

## Design points

- **The kernel's word or nothing inside a fence.** Prose keeps its shape-only linking under an old kernel; a fenced token links only when the kernel's verdict names it. Code is verbatim material, so nothing there links on shape alone.
- **The viewer's code gate applies to fenced tokens.** An `import`'s or `require()`'s package, a glob's tail, a substitution, a site name stay text, exactly as inside a shown file. Prose keeps the chat's rules.
- **A fence's rows are the walk's units**, as the viewer's rows are, so a highlight's spans never cut a line's path into pieces the scan cannot see.
- **A fenced hit opens the file and previews on hover like any link, and renders no figure** under a code sample.
- The shared walk (`path-links.ts`) grows two small things: `preVerified`, and `inPre` on the gate's context and on each hit. The viewer passes neither and is unchanged.

## Tests

- `path-links.test.ts` executes the fenced walk against the fake DOM: a verified fenced path links (quotes stay text), a package the kernel could verify is refused by the code gate, prose beside it links as before, no verdict leaves the fence dead while prose still links on shape, an empty verdict links nothing, and the walk without the options is the old walk.
- Source pins on the chat's call, the options, the hit shape and the CSS follow (`file-uri-link`, `chat-path-links`, `chat-inline-preview`, `file-view-links` tests); `tests/test_path_links.py` (kernel–client token parity) is unchanged and green.
- `docs/guide.md` names the rule beside the viewer's.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -147,7 +147,8 @@ same document scrolls to it when the document has a heading or an anchor by that
 click, since a section of the shown file has no tab of its own. One click does one thing: a plain
 click acts in the dashboard, and a Cmd-click (Ctrl on Windows and Linux) or a middle-click opens
 the link in a browser tab of its own. Inside a file the test for a path is stricter than the one
-a chat message gets: a path links only when it has a slash and a file extension, starts on its
+a chat message's prose gets, and a fenced code block in a chat message follows the file's test
+too, linking a path only once the kernel has verified it is a file: a path links only when it has a slash and a file extension, starts on its
 own, at the start of a line or after a space, a quote, a bracket, a comma, a semicolon, an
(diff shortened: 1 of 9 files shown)


henrypinkard (approved):
Approved on the maintainer's word.

henrypinkard (approved):
Approved on the maintainer's word (the branch was brought up to date with main).

PR #1500, opened 2026-09-12
Chat: the whole pane takes a file drop, each split column its own; the shell and the other panes refuse one instead of navigating

amith-ananthram:
## What this adds

Dropping an image on the dashboard anywhere but the composer box used to navigate the pane, or the whole page, to the image: the browser's default for an unhandled drop. Now:

- **The whole chat pane is the drop area.** A file dropped anywhere in a chat column attaches to the box of the session that column shows, through the same accept the box uses (the remote rule and the shipped-bytes path unchanged). In a split, each column is its own drop area.
- **The pane says so while you drag**: a dashed ring over the pane and the box's drop-target outline as the landing spot. Enter and leave are counted, since a dragleave fires at every element edge, so the ring doesn't flicker and clears when the pointer leaves the pane or the drop lands.
- **Everywhere else refuses the drag** rather than opening the file in place of the page: the shell's own chrome and every non-chat pane show the not-allowed cursor and swallow the drop.

Only OS file drags are taken (the transfer's types hold "Files"); a tab drag, a text selection and the strip's own drags keep their handlers.

## Design points

- One accept for the box and the pane (`acceptDroppedTransfer`), lifted from the box's handler unchanged.
- The refusals live in the two shell scripts every pane and the shell already run (`_shim`, `_LANDING_FOCUS_JS`); the chat's document is left to its bundle.

## Tests

- `ui/webview/drop-anywhere.test.ts`: source pins on the shared accept, the counted document listeners, the ring's CSS and the two refusals.
- `tests/test_drop_anywhere_served.py`: a real browser against a hermetic kernel with synthetic files. A drop on the transcript rings the pane while over, is taken, and the kernel's saved copy lands as the box's attachment chip; a drop on the shell's document and on the feed pane is prevented and neither navigates; a second column opened on the other session takes its own drop into its own box and leaves the first column's alone.
- `docs/guide.md`: a paragraph in the chat section.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -23,6 +23,11 @@ complementary views of what the agents are doing:
 
 ![Tool calls fold into runs; each expands to one line per call](assets/guide/chat-detail.png){ width="100%" }
 
+**Dropping a file.** Drop an image or any file anywhere on the chat pane and it attaches to
+the message box of the session you are looking at; a dashed ring shows the pane is the target
+while you drag, and in a split each column takes its own drops. Dropped anywhere else on the
+dashboard, a file is refused (the cursor says so) rather than opened in place of the page.
+
 **Reviewing a document.** Select any passage in the file viewer and it lands in the
 composer as a quote chip, labeled with the file and the line the passage lives on. Type
 what should change and press **⌘⏎** to set the note aside; keep reading, select the next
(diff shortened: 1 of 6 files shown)


henrypinkard (approved):
Approved on the maintainer's word.

henrypinkard (approved):
Approved on the maintainer's word (the branch was brought up to date with main).

henrypinkard (approved):
Approved on the maintainer's word; the served test's second-column wait adapted to the column model.

PR #1538, opened 2026-09-12
The shared text readers are registered by code object, one holdable-echo predicate and one echo floor serve the comments frame and the echo landing set, the tool-use pin covers both assistant shapes (T384 follow-up)

henrypinkard:
Fix tier, two commits: the three lows from the review's third round of the reader change, and a fourth fix the first deploy boot exposed.

- **Shared text readers by code object.** The hydration attribution matched the shared readers (the judges' `_unit_text`, `_prompt_text`, `_atom_text`; the kernel's `_atom_user_texts`) by NAME, so a local function named like one was consumed as the reader. They are registered by code object where each is defined (`register_hydrate_text_reader`), as the whole-read passthrough already was; the attribution tests' stand-ins register themselves.
- **One holdable-echo predicate.** The comments frame spelled the four clauses (a text key, not a command, not dropped, not landed) twice, in its landing floor and its held filter; `_echo_holdable` serves both, the held filter adding its two dependent clauses.
- **One echo floor.** The first deploy boot hydrated 198 MB through the echo landing set: its caller took the oldest send among every live echo with a text, so a dropped send older than the compaction, never popped by the backend's prune, sank the floor and every pre-cut user text above it was read. `_echo_floor`, the min over the holdable echoes, is the floor for the landing set's caller and the frame's walk alike.
- **The tool-use pin.** It ran over a scenario with no tool call at all and compared two empty sets; it runs over a tree with a prose-only marker beside a marker carrying calls before the cut, and a mutation probe (a reader that resolves nothing) fails it.

Red first at the base for the code-object match (a same-named local attributed as the reader), the predicate and the floor (absent); the rebuilt tool-use pin passes at the base, since the reader's rule was already right, and is coverage. The sixteen modules that pin the readers or their callers: 914 passed. Full Python suite green locally (12,518 passed).

### DIFF
--- a/kernel/event_model.py
+++ b/kernel/event_model.py
@@ -5757,8 +5757,25 @@ def _hydrate_one(a, rec):
     a.pop("lazy", None)
 
 
-_HYDRATE_TEXT_READERS = ("_unit_text", "_prompt_text", "_atom_text", "_atom_user_texts")   # the shared text readers (the judges'
-#                                   three and the kernel's user-texts reader): attributed with their caller
+_HYDRATE_TEXT_READERS = set()     # the CODE objects of the shared text readers (the judges' _unit_text, _prompt_text and _atom_text,
+#                                   the kernel's _atom_user_texts), each registered where it is defined: a hydration through one is
+#                                   attributed with the reader's caller. Matched by code object, never by name (T384 round three: a
+#                                   local function named like a reader was consumed as one)
+
+
+def register_hydrate_text_reader(*fns):
+    """Register shared text readers the hydration attribution names together with their caller."""
+    for fn in fns:
+        _HYDRATE_TEXT_READERS.add(fn.__code__)
+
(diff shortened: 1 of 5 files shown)


henrypinkard:
Round one fold (90fe3dce): the landing set's floor in `_merge_live_atoms` is over every live echo with a text again, dropped, landed and command echoes included. The text sets it floors are also `prune_live`'s by-text retirement input and the chat's echo dedupe input, and a dropped echo whose text landed in a pre-cut turn must still be hidden and retired; the holdable-only floor (`_echo_floor`) stays the comments frame's alone. The pre-cut markers carry no exact text key, so the wide floor's read stays until a scalar road exists. The test now drives `_merge_live_atoms` with a stubbed backend and a dropped echo straddling the cut, in both shapes, red at the previous head and green at the base and here.

PR #1544, opened 2026-09-12
One command sets up a second Claude login: scratch sign-in, masked setup-token, 1Password template store, romp login add (T393)

henrypinkard:
One command sets up a SECOND Claude login for romp (the user, 2026-09-12): `bin/romp-login-setup <1password-vault> <label>` signs in to the other account under a scratch configuration (the machine's own login is untouched), mints a `claude setup-token`, stores it in 1Password as an API Credential item titled with the label through a template file (never a command argument), and registers the login with the running romp as `romp login add <label> --op op://<vault>/<label>/credential`. The script is the user's own dozen lines, kept readable end to end: no comments past the usage line, no prose output, each step failing loudly and stopping the ones after it. The token goes to exactly two places, a private file under a temporary directory removed on exit and the 1Password item; the one CLI line that carries it reads `<token captured>` on screen.

**Design points**
- The token grep carries a pipefail guard (`|| true`): without it a sign-in that prints no token ends the script on the grep's own exit status, silently, before the loud line. The bats module is red without the guard.
- `--check`, off by default, proves the stored token is what Claude Code accepts: one request through an `apiKeyHelper` that reads the item, with every other credential removed from the environment, must be accepted; the same request through a helper that returns junk must be refused, which proves the helper, and nothing else on the machine, authenticated.
- `romp login add` with `--op` exists on the several-logins pull request only, so the script checks for it BEFORE any sign-in (a bare `romp login add` prints that branch's usage line; main's unknown-command line stops the script with a plain sentence). This pull request waits for that change to land; the reference says the command needs it.
- The label is validated before the sign-in (letters, digits, spaces, dashes and underscores): a quote or backslash would break the JSON template after the token was minted, and a slash would read as 1Password's section form, a reference to a different item.
- The mask covers the CLI's own redactor shape (`sk-ant-` then letters, digits, `_ . -`) and the capture its detector shape (`sk-ant-(oat|ort)` two digits, then twenty or more of those characters), so an `oat02` prefix or a dot is masked and stored whole.
- `--check` clears every credential variable the CLI reads (the three token variables, the token file and file-descriptor variables, the custom headers); the machine's own settings files and keychain still apply, which is what the junk-helper control is for. It is accepted anywhere on the line; an unknown flag or a third positional is refused with the usage.

**Tests**
- `tests/romp-login-setup.bats` (twelve tests): stubbed `claude`, `op` and `romp` on PATH, a synthetic token assembled at run time (no credential-shaped literal in the repo), the script run directly (it reads no terminal device: the CLI's stdin is the terminal it runs from). Pinned: the token masked on screen and absent from every command's arguments and every left-behind file, whole in the template (both token shapes, the second with an `oat02` prefix and a dot), romp told the reference after the store and its `login add` checked before the sign-in; a romp without the command, and a label with a quote, a backslash or a slash, each stopped before any sign-in; no token, a refusing 1Password and a refusing romp each a loud stop with nothing after; the check's acceptance with the environment cleared, its two failure modes; usage, `--check` anywhere, an unknown flag refused.
- `docs/reference.md` gains "Several Claude logins" (the one command for colleagues, `--check`); `bin/README.md` gains the row.

Tier: feature


henrypinkard:
Closed on the user's word (2026-09-13): the second-login setup, which mints a Claude setup token and stores it in 1Password, is specific to one machine's tooling and does not belong inside romp. romp keeps the generic mechanism alone: a stored login is a label and a token command (`romp login add <label> --cmd '<shell line that prints the token>'`, pull request 1447, with the 1Password shorthand removed there too). The setup becomes a script in the user's own dotfiles that ends by calling that command.

PR #1611, opened 2026-09-13
A dashed yellow ring on the tab of a session with something waiting on you, in every live state

alowet:
A tab wears a dashed **yellow ring** whenever the feed files a card of its session under needs-you, in every live state the red ring does not already own: idle, waiting on background work, and **still working**. The red ring stays what it is (a live permission/picker prompt, or an API stop only you can clear) and outranks the yellow; the yellow outranks the amber auto-retry ring. A folded tag group's pip goes yellow for a hidden member with an ask, ranked between red and gold, so a fold cannot hide it. The phone's session picker marks the same sessions (a yellow bar on the row, a dashed yellow border on the current-session chip).

Why: the sessions that need you should grab attention in the strip without clicking through each of them or going to the feed. Today the only in-strip cue is the red ring for a live prompt, so a session that asked something in prose and went idle, or asked and went on working in the background, looks like every other tab. Notifications are unchanged: the card entering needs-you is already the event that notifies when the bells are on; the ring is that card shown in the strip, and it stays as long as the card does (across a kernel restart too, which announces nothing).

**Design points**

- **One verdict, three surfaces.** The kernel already computed the per-session needs-you boolean (`_feed_needs_input_of`, from the last feed build) for the ledger's `needsInput`, which the section-at-a-glance rows read. This puts the same value on the session **status** as `needsYou` (tri-state like the ledger's), because the tab rule reads one object and a skeleton tab gets status frames alone. No re-derivation, no heuristic: the tab, the section row and the feed's Blocked list cannot disagree.
- **A second class, not a state.** `tabAskClass` (tab-state.ts) returns `tab-ask` beside the state class, so a working tab wears the gold dot inside the yellow ring rather than losing its state. Red (`tab-awaiting`/`tab-blocked`) and closed tabs never get it; the CSS `:not` chain enforces the red exclusion by construction too.
- **Repaints on new information only, and promptly.** The strip's signature reads `tabAskClass` for loaded and skeleton rows, so the ring appears or goes exactly when the card enters or leaves the column, never on an inference flap. A push builds the chat sessions before the feed, so the verdict ships on the next cycle; a feed build that moves the needs-you set now wakes the pusher (the `_mark_views_dirty` pattern), so the ring and the section row trail the card by one build rather than a backstop tick. `build_session` reads the verdict once for both the ledger and the status, so a row and its tab cannot disagree within a frame.
- **Tokens, both themes.** `--st-ask-bg`/`--st-ask-fg`: a clear lemon (`#f5d33f`, 11.3:1 on the page) apart from Working's mustard gold and the retrying amber; light `#7a6400` (4.82:1). theme-parity pins both pairs and the ring-on-page contrast.

**Screenshots** (synthetic fixture `tools/ui-verify/fixtures/tabs-ask-ring.html`, notes-api demo sessions): left to right, a working session that asked (gold dot inside the yellow ring), an idle one that asked, a live prompt (red outranks), a plain working tab, an API retry (amber), and a folded group whose hidden member asked (yellow pip).

Dark theme:

the chat tab strip, dark theme: a working tab with a gold dot inside a dashed yellow ring, an idle tab with the yellow ring, a red-ringed live prompt, a plain working tab, an amber retrying tab, and a folded group header with a yellow pip

Light theme:

the same strip in the light theme

**Tests**: `tab-state.test.ts` (executed rule and pip ranking), `tab-strip-skip-exec.test.ts` (executed: a working tab wears both classes; the verdict flip alone rebuilds the strip once; red never pairs with yellow), new `tab-ask-ring.test.ts` (wiring, cascade order, tokens, guide, phone picker), `theme-parity.test.ts`, `tab-groups.test.ts`, `tab-strip-skip.test.ts`; `tests/test_kernel.py` (status `needsYou` moves with the ledger's `needsInput` through None/True/False/muted), `tests/test_kernel_mobile.py`. Full `npm test` (4696 pass), `npm run typecheck`, `npm run build`, `tests/test_kernel.py` (460 pass), `test_kernel_mobile`, `test_chat_build_sig_inputs`, `test_awaiting_count` all green locally.

**Known limit**: the ring inherits the needs-you column's coverage. An ask the judges do not file as a block draws no ring; improving that is a judge change, not a strip change.


henrypinkard (approved):
Approved on the maintainer's word. The maintainer's session made the rings tab widgets on this branch (a ring slot in the registry, three ring widgets with a switch each in the settings, the light-theme colour re-inked with the contrast validator, the attribution comments reworded).

PR #1652, opened 2026-09-14
A summary click that switches the focused session scrolls the feed to the top, so the section's new cards are in view (T414)

henrypinkard:
T414: with the feed's focused-session section on, a click on a card's summary line that switches the chat to another session now scrolls the feed's box to the top, so the section, which shows that session's cards from that moment, is in view. Tier `fix`: the node pins and the served lab fail before it.

## What the user asked for

Paraphrased from their words (2026-09-15): with the current-session section on, a click on a card's Summary that jumps into that session and changes the chat's active tab must also scroll the feed to the top, so the section, now showing that session's cards, is in view; today the feed stays where the reader was and the cards it just switched to sit above the fold.

## Design points

- **The trigger is the switch, not the click.** The summary line's click records the session it jumps to; the kernel's `activeChat` frame for that session (the event the feed's focused-session seam already runs on) settles it, after the section has repainted with that session's cards, with one plain scroll to the top of the feed's scroll box. No animated chase: the scroll is the reader's own gesture's consequence. A frame for some other session drops the pending scroll.
- **A click on the session already focused scrolls at once.** No tab changes, so no frame comes; the section already holds that session's cards and the reader asked to look at them, so the top comes at the click. The named choice.
- **A later user scroll wins.** The feed's own scrollTop writes (this scroll, and the render's scroll restore) are guarded, so the list's scroll listener can tell the reader's scroll from ours: a reader's scroll between the click and the frame cancels the pending jump scroll, and once ours has landed it stands until they scroll (the standing scroll rule).
- **The hover-freeze gate stands.** The click lands with the pointer on the card, which holds the payload gate; the frame is deferred as every frame is until the pointer leaves, and the scroll settles with that deferred repaint. The board never moves under the pointer.
- **With the section off, nothing changes**; the same behaviour in the row and the single-column layouts.

## Tests

- `ui/webview/feed-focus-scroll.test.ts` (3 tests): the summary click notes the target under the section switch and scrolls at once for the focused session; the frame settles one plain scroll after the repaint and a frame for another session drops it; a user scroll cancels and the feed's own writes are guarded (the render's restore too).
- `tests/test_feed_focus_scroll_served.py` (new served lab, one browser run): a feed tall enough to scroll 600 px, the section on, web focused; a click on api's summary line posts the jump and moves nothing by itself, the frame for api brings the list to the top with the section's block head in the list's viewport (row layout, with screenshots in both themes); a reader's scroll between the click and the frame stands at 300 px; a summary click on the focused session's own card scrolls to the top at once; with the section off the click and the frame move the list by nothing; the single-column layout repeats the jump (screenshots).

## Verification

All on the branch worktree under the load cap, every pytest run pinned to Python 3.13. Typecheck clean and the whole extension suite green (4,823 tests, 4,818 passed, 5 pre-existing skips); the served lab green in both layouts. Red first at main f6b2b4cf: all 3 node pins red (no such mechanism), and the served lab red on the jump road (the frame left the list at 199 px, the browser's clamp of the reader's 600 px in the shorter document, with the section's block head out of view: the user's complaint). The full pytest 13,334 passed, 56 skipped, 2,183 subtests; bats 658 passed. Screenshots of the list at the top with the switched section in view, row and single-column layouts in both themes, are in the maintainer's drops folder.


henrypinkard:
Folded into #1658: the jump scroll now rides the section's local switch on the click itself, so no record waits on a kernel frame (the unbounded-record finding) and every jump, the split summary's paragraphs included, passes one note (the second finding). This branch's commit was carried into that PR and reworked there.

PR #1761, opened 2026-09-15
Chat columns can stack: drag a tab to the bottom edge for a second row, any arrangement of four columns over two rows

alowet:
## What

The chat's columns can stack. Drag a tab to the chat's **bottom edge** and a row opens below with that session as its column; drag a tab to a row's **right edge** to add a column to that row; drag the gutter between the rows to resize them. Any arrangement of the existing cap of four columns over two rows (1+1, 2+1, 1+2, 2+2, 3+1, 1+3). An emptied bottom row folds away and the top row takes the full height. The first pane stays in the top row as the overflow for every session not placed elsewhere. Keyboard: **Move this session to a new column below** in the palette, bindable like the other column commands; next/previous column walk the top row then the bottom.

## Screenshots

Synthetic sessions on a hermetic kernel (Playwright, 1440×900); no real session data.

| 2×2 (dark) | 2×2 (light) |
|---|---|
| 2x2 dark | 2x2 light |

| 1+1 stacked (dark) | 1+1 stacked (light) |
|---|---|
| 1+1 dark | 1+1 light |

## Design points

- **A row is a property of a column, not a new object.** Each later column's store entry gains `row: 2` when it sits in the bottom row (absent = top), and `rowSplit` (the top row's share) exists only while a bottom row does — so a dashboard with no bottom row writes exactly today's `{"v":2,"cols":[…]}` bytes (asserted byte-for-byte). `read()` accepts entries without `row`; garbage `row`/`rowSplit` fall to top / 0.5. Two dashboards reconcile rows through the same storage-event path.
- **A static wrapper, never a re-parent.** `#chat-area` (a flex column of `#chat-row-1`, `#gv-rows`, `#chat-row-2`) is rendered at page load, so `#f-chat` is never moved in the DOM (a moved iframe reloads); the served test stamps a nonce on its document and reads it back after every step. The outer row grows `#chat-area` by the weight `#chat-pane` had (`--g-chat`, so the outline/feed gutter code is untouched); inside a row the first pane takes `--g-chat1` and columns keep `--g-chatN`; a pre-rows store is upgraded once, pixel-exactly: `chat1` is seeded from `chat`, a restored column with no saved weight gets what the pre-rows shell gave it (the mean of the panes on screen), and the columns' weights are converted so every pane keeps the width the old shell would have drawn (within a pixel, asserted against the old helper run verbatim over a 192-case matrix). If the Chat pane is hidden at boot the upgrade waits for the pane's own show event and finalises from live state at that instant (live outer weights, the live column roster), and a store another dashboard has already upgraded is taken in the moment it lands (a `storage` listener) and before any local change, never overwritten. A gutter drag is one transaction: a peer's write during the drag is noted and applied at release, with the dragged pair rebased so the divider lands where the landing line was; an abandoned drag (pointer cancel, window blur, tab hidden, a second press) restores the press-time widths and writes nothing, and every structural change (a column made, moved or closed, a pane toggled, another dashboard's write) ends a live drag first, so a release can never write the width of a pane that is gone. A pane id that is not registered maps to no weight (main's fallback to `files` is gone). Halving/hand-back normalises only a pane's container siblings, so opening or closing a column never moves the chat area against the feed. Rows share height as weights out of 100 (a lone row on a 0.5 grow factor fills only half the area — found by the served test).
- **Zones.** During a tab drag: a column zone on every pane but the source, an edge zone at each row's rightmost pane, and a bottom zone along `#chat-area`'s bottom edge (open the row with this session, or append to it). The ghost shows the target row's rightmost pane's right half, or the area's bottom half. A session alone in a later column gets no making zone in its own row (it would only move itself). The cap counts across rows: a refused ghost and today's notice.
- **Fold.** Closing the bottom row's last column (×, `colEmpty`, the session moved away, or another dashboard's write) hides the row and its gutter, drops `rowSplit`, and the top row takes the full height; gutters are tidied so none is orphaned. Mobile is unchanged: one pane, the wrapper is `display: contents`, the bottom row hidden.

## Tests

- Node-run split JS (`tests/test_chat_split.py`, `RowsExecute`): bottom drop → row 2 with `rowSplit`; bottom row's edge → 2×2; 3+1 and 1+3; cap refused; the fold restores today's bytes; a store without `row` → all top; reconcile places rows; the walk order; halving within the right row.
- Served (`tests/test_chat_rows_served.py`, 17 tests, hermetic kernel + synthetic sessions + Playwright at 1440×900): real pointer drags build 1+1 then 2×2 then 3+1; the fourth column is refused; the row gutter drag changes `rowSplit` and survives a reload; `#f-chat`'s nonce survives every step; emptying the bottom row folds it; close-all returns to one pane.
- Pre-rows upgrade (`tests/test_pane_gutters.py` 12–15): legacy widths kept within 1 px with one, two or no saved column weights; the 192-case matrix against the `ab112d49` helper (the embedded helper is diffed against `git show`); Chat hidden at boot → upgrade deferred, store shape unchanged, a reload while hidden still pending; then reveal Fleet / drag an outer gutter / close the column / a peer publishes an upgraded store → show Chat → the old shell's widths, no phantom column, the peer's bytes untouched. Served step 8 repeats the hidden-Chat sequences with real rail clicks and the palette's close command, and the peer cases with a second page in the same browser context (its first rail reveal equals a page booted current; a peer write between two pointer moves; an abandoned drag via a dispatched blur writes nothing; a peer closing a column under a live drag ends it, and the store never returns to the pre-rows shape).
- A session alone in a column over a create in flight stays put across rows (the busy guard runs before the bottom drop mutates anything); a reopened bottom row is the half its ghost promised (`rowSplit` resets on fold); the bottom ghost accounts for the 7 px row gutter.
- Gutter, rail, files-pane and mobile pins updated; `chat-split.test.ts` pins the zones, the palette command, and the byte-identical no-bottom-row shape.
- Gates: `npm run typecheck`, `npm test` (4897 pass), `npm run build`; `test_chat_split.py`, `test_pane_gutters.py`, `test_pane_gutter_drag.py`, `test_kernel_pane_rail.py`, `test_landing_bundles_built.py`, `test_kernel_mobile.py`, `test_files_pane.py`, `test_inline_js_parses.py`, `test_chat_split_served.py`, `test_chat_split_host_served.py`, `test_chat_rows_served.py`.

## Known edges (deliberate)

- **Fixed on main since this branch was cut:** another dashboard's write closing a column whose page has a create in flight (`close(n, keep)` skipping the busy check) is fixed by #1774, which defers that close to the page's idle signal. This branch merges that main; the rows add no new way around it (a row change goes through the same `reconcile`).
- **Mixed builds, last writer wins.** A dashboard still on the pre-rows build open beside one on this build writes the pre-rows store shape; the rows dashboard ignores that write and its next action (or a reload, which upgrades once more, pixel-exactly) converts the store back. Two truly simultaneous cross-tab writes between one action's read and its write are resolved by the write guard (a pre-rows shape is never written over an upgraded store), not merged.
- **Move this session to the next column** from the bottom row's last column still opens a new top-row column, as on main; Alt+Down spatial navigation still targets the timeline band rather than the bottom row; reconcile from another dashboard does not reorder columns within a row (as main).


henrypinkard:
Closing this draft; the facts behind the decision:

- The vertical split is being built on a within-column model: the design is `plans/chat-vertical-split.md` on main (merged the morning of September 15, refined that evening), its first pull request is #1769 (CI green), and a second is in progress.
- This draft's store shape (a `row` number per column entry plus one shared `rowSplit`) and the plan's (a `place:'below'` entry with a `parent` column and a `ratio`) both extend `romp-chat-cols`, so only one of the two can land.
- The plan's model can express every layout the strip can (a full-width second row is every column split at the same ratio), but not the reverse: it splits one pane while the others stay whole, the first column included. It also rewrites no browser's `romp-pane-grow` store.
- The decision is to proceed with the plan, so this draft will not be merged. The pane-width store conversion, the cancellable gutter transactions and the served test in this branch were read in full.
- The edge named in the body is confirmed real on main: a peer dashboard's `romp-chat-cols` write reaches this dashboard's storage listener, whose reconcile closes a column missing from the peer's store with `keep` set, and `keep` skips the busy gate, so a column whose page has a session creation in flight is torn down with its queued text. A fix is being prepared from this side; the deferral will wait for the column's own signal that the creation has ended rather than a timer.

Thank you for the work on this.

PR #1763, opened 2026-09-15
GET /perf carries a heap block and a gc block: where the kernel's resident memory sits and what the collector's pauses cost each cycle, read live

lczh:
Tier: feature

Design points (a self-contained, read-only addition to the existing diagnostics route):
- Two new blocks, heap and gc, beside process in GET /perf; each split row on the pusher and jobs rings gains a `gc` delta and the kernel-samples row two cumulative fields (the second commit, below); no client reads any of it yet.
- Gauges only (the occupancy at the read); gc.stats is the collector's cumulative tally and is named as such.
- Each read is O(1) or O(entries) over a copied value list; no object-graph walk, no gc.collect, no eviction, no cache fill, no file read, no build, no setting changed.
- Owner locks only where the container's own readers take one; at most one container lock at a time; never under the perf lock (snapshot assembles outside it), and no holder of those locks reaches the perf stats, so no lock order with it exists.
- A gauge the process cannot read is null, said once per key on stderr; the snapshot still serves.
- Measured overhead in a lab process over populated caches: 166 microseconds median (min 152, max 270) over 20 calls.

Sample block from a lab snapshot (synthetic entries: one built tab with 40 cached events in the shipped wire's shape, one 300-byte image data URL):

```json
{
 "allocatedBlocks": 129625,
 "gc": {
  "enabled": true,
  "counts": [
   140,
   0,
   4
  ],
  "thresholds": [
   700,
   10,
   10
  ],
  "stats": [
   {
    "collections": 44,
    "collected": 371,
    "uncollectable": 0
   },
   {
    "collections": 4,
    "collected": 122,
    "uncollectable": 0
   },
   {
    "collections": 0,
    "collected": 0,
    "uncollectable": 0
   }
  ]
 },
 "tracing": false,
 "hydrated": {
  "entries": 0,
  "bytes": 0,
  "capBytes": 1073741824
 },
 "assemblyEntries": 0,
 "parseSlots": 0,
 "lazyIndexes": 0,
 "materializedLruSlots": 0,
 "judgeUsageRows": 0,
 "builtChat": {
  "tabs": 1,
  "events": 40,
  "serializedBytes": 0
 },
 "imgCache": {
  "entries": 1,
  "bytes": 322
 }
}
```

The snapshot reported VmRSS and cumulative counters (bytes read, atoms built, bodies hydrated since boot) and no
occupancy at all, so the lag investigation (2026-09-15) had to attribute a 5-6 GiB resident size from source-side
byte counters and lab tracemalloc runs; the questions it left open (a resident-budget factor, the judge-usage rows)
need live gauges to settle. The new `heap` block beside `process` answers them from the running process.

What it reports, every value an int, bool, float, None or a dict of those, all gauges (the occupancy at the read)
except gc.stats, which is the collector's cumulative tally by nature: allocatedBlocks (sys.getallocatedblocks);
gc (enabled; counts, the young generation's allocations since its last collection and then how many times each
younger generation was collected since the older's last; thresholds; stats per generation: collections,
collected, uncollectable); tracing (whether a tracemalloc tracer runs); hydrated (entries, bytes, capBytes of the lazy-body
memo, read under _ASM_CKPT_LOCK as the reader's own path does); assemblyEntries (len of the assembly cache under
_ASM_LOCK); parseSlots (len of the judge's one parse store under _PARSE_CACHE_LOCK; the store is an OrderedDict,
so the count is O(1)); lazyIndexes (len of the WeakSet of live indexes, O(1), never iterated);
materializedLruSlots (len of the materialized-atom LRU under _MAT_LOCK: named SLOTS and documented as an upper
bound on the live atoms, since a kernel whose LRU holds its lists weakly keeps a collected list's slots until
they expire; where the lists are held strongly the two are equal); judgeUsageRows (len of the usage reader's
rows); builtChat (tabs; events, the cached payloads' event counts summed over a list() snapshot of the values,
the resident measure of that cache; and serializedBytes over the cached JSON strings, which only the index wire,
a proto-1 client, stores, so under the shipped wire it reads 0); imgCache (entries and bytes over a list()
snapshot of the data URLs; the cache has no cap, so this gauge is O(entries) over whatever it holds, a refused
file counting as an entry of zero bytes). hydrated.bytes is the records' length on disk, what capBytes bounds, a
floor on the memo's resident share, not its size; materializedLruSlots is the same read asmIndex.resident already
publishes, repeated so the holders sit together. A dict's value list is one C-level step under the GIL, so the pusher mutating
the built-chat and image caches beside the read cannot raise.

What it never does: walk an object graph, gc.collect, evict, fill a cache, change a setting, read a file or
build anything. Each gauge is O(1) or O(entries) over a copied value list (the built tabs, bounded by the open
tabs; the image entries, whose cache has no cap). Every read is wrapped:
a container the source has not got (an older event model beside this kernel, a test's stub module) or an
accessor that raises yields None for that key alone, the snapshot still serves, and the failure is said once
per key on stderr (_HEAP_SAID), never once per snapshot. Lock discipline, checked: snapshot() assembles its
return outside the perf lock, so the heap reads never run under it; at most one container lock is held at a
time; and every holder of the four container locks does dict work and thread-local stage marks only (the
_read_stage provider reads a thread-local), none reaches the perf stats, so no lock order with the perf lock
exists to get wrong.

Overhead, measured in the lab over 1,000 image entries, 50 built tabs of 100 KB, 100,000 hydrated entries,
10,000 LRU slots and 50,000 usage rows: 166 microseconds median over 20 calls (min 152, max 270), against the
5 ms bound the test asserts.

Tests (tests/test_perf_heap_block.py, hermetic, synthetic only): the gauges move by exactly the added amounts
and fall back on removal (the built-chat tabs planted in both wire shapes), an entry removed under the reader's
lock leaves both the hydrated count and bytes, two snapshots move no cumulative
counter (recordCache.inserts, asmIndex.materialized, builds.chat.built, the hydration totals); json.dumps
succeeds and every leaf is a plain scalar; with the event model swapped for a stub lacking _HYDRATED, _MAT_LRU
and _LIVE_INDEXES and gc.get_stats raising, those keys are None, the rest answer, three snapshots say each
failure once, and the gauges return with the containers; the same for kernel-side containers; two snapshots
read no bytes, insert no record, evict nothing, build nothing and, with the collector disabled so only an
explicit collect could run, leave every generation's collection tally and the young count untouched; the
overhead bound; and docs/reference.md names the block, every gauge, the gauge-against-counter caveat and the
slots-against-atoms caveat. tests/test_perf_stats.py's key-set pin and its doc pin gain `heap`.

Run green, one module at a time: test_perf_heap_block 10, test_perf_stats 62, test_perf_bench 45,
test_kernel_samples 5, test_asm_index 26 (3 subtests), test_asm_checkpoint 90 (85 subtests),
test_kernel_jsonl_cache 16. The assertIn pin scan over the tests reports 0 lost pins on the branch.

Built for the 2026-09-15 lag investigation, which had to attribute a 5-6 GiB resident kernel from cumulative counters and lab runs; reviewed before filing by three adversarial lenses each refuted by three skeptics, with their finds folded (an events gauge for the built-chat cache, since the shipped wire stores no serialized string; hydrated bytes named as on-disk; the LRU slots named as the same read as asmIndex.resident; the uncapped image cache said; allocator and collector wording). Filed for the maintainer's review, separate from the ready performance fixes, so those can deploy while this is considered.

## Added 2026-09-16: the collector's pauses, per generation and per cycle

Pusher cycles stalled for 9 to 33 seconds and a native profile caught a 9.2 s full garbage collection charged to whichever stage was running; nothing in the kernel counted collections or their pauses. The second commit adds a top-level `gc` block on `/perf`: per generation, the number of collections and their total, longest and last pause in milliseconds (wall time on the collecting thread), from a `gc.callbacks` hook main installs once, beside the collector's thresholds, counts and frozen count. Each split row on the pusher and jobs rings carries the cycle's own delta as `gc` (collections per generation and the full-collection milliseconds), so a slow row names the full collection it paid for instead of the stage the collector interrupted. The kernel-samples row gains the cumulative full-collection count and milliseconds.

The hook never takes the stats lock: the interpreter runs a collection on the allocating thread at its next check, which can be inside one of the stats class's own locked regions, and a callback that waits there deadlocks that thread and, because the collector's in-progress flag stays set, stops every later collection in the process. The tallies are immutable tuples replaced in one store, read without the lock. A regression test drives an automatic collection inside the held lock on a helper thread and asserts it finishes.


lczh:
## Why the second commit, and where to look

**The problem it measures.** On a kernel serving 22 sessions, pusher cycles have a median around 0.3 s but a tail of 15 to 22 cycles over ten seconds an hour (worst 27 to 33 s), charged to whichever stage happened to be running. A native profile caught one 9.2 s generation-2 collection on the pusher thread. Nothing on `/perf` counts collections or their pauses, so every attribution so far has been by inference.

**What the labs say.** On the deployment machine a full collection costs about a microsecond per tracked container (0.8 to 1.05 s per million, linear from 0.26 M to 2 M), so a 9 s pause means roughly ten million tracked containers. CPython 3.12 runs a full collection when the objects promoted since the last one exceed a quarter of the long-lived total, which makes the seconds lost per hour proportional to the promotion rate and independent of the retained heap's size. The pusher promotes tens of millions of containers an hour (view-delta pairs, feed card tints). #1766 removes the largest of those streams; this block is how the next measured hour tells whether it worked, cycle by cycle.

**What the commit adds.**
- A top-level `gc` block beside `heap`: per generation, collections and their total, longest and last pause (wall time on the collecting thread), the last collected count, plus thresholds, counts, frozen count, callback errors and whether the hook is installed.
- Every split row on `pusher.stageRing` and `jobs.stageRing` carries the cycle's own delta as `gc` (`n0`, `n1`, `n2`, `ms2`), so a slow row names the full collection it paid for.
- The kernel-samples row gains cumulative `gcGen2Collections` and `gcGen2MsSum`.

**Where a reviewer should look.**
- `gc_event` never takes `_PerfStats.lock` or any other lock. The interpreter runs a collection on the allocating thread at its next check, which can be inside one of the stats class's own locked regions; a first draft that took the lock deadlocked that thread and, because the collector's in-progress flag stays set through the callbacks, silenced every later collection in the process. `tests/test_perf_gc_block.py` drives an automatic collection inside the held lock on a helper thread, with a witness callback proving it ran there, and asserts the thread finishes.
- The tallies are immutable tuples replaced in one store; readers difference them without the lock.
- The per-row delta is process-wide over the cycle's wall window (a collection on any thread stalls the cycle, since it holds the interpreter lock); a collection inside overlapping pusher and jobs windows shows in both rings, so neither ring sums to `gen.collections`. `docs/reference.md` says so.
- Collector accessors go through the same per-key guard as the heap block, so a runtime lacking one answers `null` for that key alone.

Tests, one module at a time: `test_perf_gc_block` 16, `test_perf_stats` 62, `test_perf_heap_block` 11; the assertIn pin scan reports 0 lost pins. CI is green on every job; the tier check waits on approval.

henrypinkard (approved):
Approved on the maintainer's decision of 2026-09-16, with the review round (the boot-health row's collector delta, a said-once line on a hook failure, the record-cache pointer in the heap paragraph, a loud built-chat gauge on an unknown shape) carried onto the branch.

PR #1778, opened 2026-09-16
Nine tests read the platform, not Linux: the macOS CI cells green again (tests only)

henrypinkard:
Nine tests red on the macOS CI cells (run only by manual dispatch), every one an environment reading of the test itself, not of the product; seven test modules change and no product file. Causes per test from the macOS run's own log and the release triage; fixes:

- `tests/test_env_value_redaction.py` (2, HookEndToEnd): the nested pytest was handed the scratch dir unresolved; on macOS that dir sits under `/var`, a symlink to `/private/var`, and the child rendered the collected file's node id as a long relative path, so the short summary line was cut at the terminal width before the message the tests read. The dir is resolved once for `cwd` and `--rootdir`. The redaction itself held on both cells.
- `tests/test_tempdir_hygiene.py` (2): macOS `mktemp(1)` with no template names its directory from the per-user temp root, not from `TMPDIR`. The shell half of the inherit check is a stated skip on Darwin (its Python half stands), and the deliberately leaky inner module skips its one shell `mktemp -d` there with the reason in place. No product script relies on the behaviour: both shell users pass an explicit template.
- `tests/test_cli_scope.py` (1): the live cgroup probe needs procfs; without it `cut` names the missing file on stderr. The live assertions run only where `/proc/self/cgroup` exists, the shape the same test already uses for the adjustment probe; the static command pins stay unconditional. The feature is off on macOS anyway (it needs `systemd-run`).
- `tests/test_cut_turn_tree_kill.py` (1): the leftover-scope sweep proves a live child's parentage from `/proc/<pid>/stat`; with no procfs it read None and stopped the child's scope too. The parent is patched for the child's pid alone, as the file's sibling tests patch liveness. The sweep returns 0 on macOS before any read, as soon as `systemctl` is missing.
- `tests/test_sdk_lifecycle_hardening.py` (1): the fake orphan pid read as alive with no procfs, since the liveness fallback is `os.kill(pid, 0)`, which the test had replaced with its recorder; the grace then expired into a SIGKILL. Liveness is patched to False as the two sibling tests in the file already do, with the reason.
- `tests/test_first_cycle_stage_split.py` (1): the ring override is clamped to the memory fraction by design, and a 7 GB runner reads 28 where the test expected the literal 40. The assertion reads the clamp.
- `tests/test_login_records.py` (1): CoreFoundation adds `__CF_USER_TEXT_ENCODING` to every child's environment on macOS; the whitelist assertion allows that one name on Darwin. romp passed nothing extra.

Linux stays green: the seven modules, the isolation pin and the whole-suite collection on the head. The macOS legs run by manual dispatch of CI on `stack/romp_timeline-macos-tests` at the same head; the run id and its conclusions are in the review thread.

Tier: fix


henrypinkard:
Manual CI dispatch on `stack/romp_timeline-macos-tests` at this head: run [35115585082](https://github.com/romp-on/romp/actions/runs/35115585082).

- Every Linux leg success (Python 3.10 to 3.13, bats, gitleaks, the extension, the tier policy).
- Python 3.10 and 3.13 on macos-latest: 20 failed, 13530 passed each. None of the nine tests this PR fixes appears in the FAILED or SUBFAIL lines. Nineteen are the resolved-path group carried by the sibling PR (`test_asm_checkpoint`, `test_codex_runtime`, `test_file_slice`, `test_github_repo`, `test_glossary`, `test_perf_bench`); the twentieth is `tests/test_git_pointer_file_bytes.py` (a torn pointer file's fault keyed under the handed path and cleared under git's resolved toplevel), in neither PR.
- Shell (bats, macos-latest): cancelled at the job's 35-minute ceiling inside the bats step, as on the previous dispatch from main; no shell file changes here.

The combined proof is a dispatch on main after both PRs merge.

PR #1779, opened 2026-09-16
A column another dashboard closes while a session is being created in it stays open until that create settles, so nothing typed into it is lost

alowet:
## What

A column that another dashboard closes while a session is being created in it now stays open until that create settles, so nothing typed into it is lost.

## The bug

The chat split refuses to close a column whose page has a create in flight: `close()` reads `busy()` and toasts. But `reconcile()`, which applies another dashboard's write of `romp-chat-cols`, called `close(n, true)`, and `keep` skipped that check. A peer moving the column's last listed member elsewhere (or closing the column) removed the pane at once, and the create's queued text and draft died with the document. `reconcile` was the only `close(…, true)` caller; every other close leaves `keep` unset and is refused as before. Named as a pre-existing edge in #1761.

## The fix

- **Hold, don't close.** `reconcile` keeps such a column mounted and unlisted (its members went where the peer put them; `sets()` gives it none, so its page shows the create alone) and writes nothing back. `save()` never writes an empty entry, matching `read()`'s shape. No toast: the write is a peer's act, not this user's, and nothing is refused.
- **Close on the exact event.** The chat page posts `{romp:'colBusy', busy}` to the shell from every write of the two facts `__rompColumnBusy` reads (`openProvisional`, `dropProvisional`, `failProvisional`, a failed tab's discard), only when the answer flips. On `busy:false` from a held column the shell re-reads the store and reconciles against it as it stands then. If the create landed a session meanwhile, the page's claim (`__rompClaimSession`, the path every create takes) already listed the column and wrote the store, so that reconcile keeps it.
- **Two facts, one reader.** A column is busy while a create is in flight or a failed create is still kept; `columnBusy()` is the one reader behind the shell's `__rompColumnBusy`, the flip, and the column-emptiness guard (which used to read the facts on its own and could post `colEmpty` mid-question). When a column's last listed session vanishes while its document is busy, the shell now holds the column exactly as it does for a peer's write, instead of refusing with a toast, and the flip completes the close.
- **The folder question keeps the tab.** The provisional tab now lives through the folder question (create it, or edit the path): nothing is carried in document-local state, so no close can destroy the text and it never lands in another conversation's draft. Retrying re-sends on the same tab; dismissing the question (Escape, backdrop, opening the picker over it) turns the tab into the existing failed-create state, where the text stays visible and only the user's own ✕ discards it. Every create attempt carries a request id that the kernel echoes on each of its direct replies (the folder question, a warning, the focus that ends a create), and the page routes every tagged reply by that id alone: to the create still pending, to the failed tab it belongs to (refining its reason), or as a plain notice for a create that already settled; a reply from a create the user replaced is dropped for the rest of the page's life; nothing tagged ever lands on a different create. Untagged replies from an older kernel behave as before. A prompt's buttons are bound to that request. A failed create's tab, its reason and whatever it holds (typed, staged, cited or attached) survive a reload, and a create still pending when the page reloads comes back as a failed tab with a truthful reason and everything it held (its outcome is unknown to the page; the session appears in the first pane if the kernel did create it). The page announces its busy state once booted, so a held column still closes only on the ✕ the tab promises. Anything left under a create id with no record is swept from the store. A dialog that replaces the folder prompt settles that create quietly into its failed tab (the reason shown there, never as a second modal), so there is only ever one confirm on screen. A create superseded by a second one hands its text into the new tab instead of dropping it. When a create settles (a new session, or an existing one it resolves to), everything its tab held travels to the real session: the typed text (joined after any draft already there, never over it), the staged messages with their citations (a citation staged on its own included), unsent citation chips (merged without mixing a goal with quoted context) and attached files, through the same hand-off the close uses for any session shown elsewhere. Uploads never cross documents: a session with an upload still on its way cannot be moved until it lands, a column with such an upload stays busy until the kernel answers, and only completed files travel; a message held for its uploads is sent from where it was staged, only when every upload completed, otherwise it is kept as a draft and the user is told. An upload's answer is routed to the composer or the comment box that started it, never by whichever is open (a comment's file arriving after its comment closed is announced, never attached to the composer; a comment upload failing never cancels a message the composer is holding for its own uploads, and the composer's count names only its own); a host detaching fails its uploads instead of holding the column forever; a column held only by an unshown session's upload asks before it closes; a session's upload guard asks every mounted column, not just the one that currently lists the session.
- **The flip is the last act of every settling path**, after that path's final write of the text (`adoptProvisional`, `resolveProvisionalToExisting`, `cancelProvisional`, `onCreateDirMissing`; `failProvisional` stays busy). A create that resolves to an already-running session, shown in another pane, hands its typed text to that pane before the flip: `close()` now also carries the state a closing page holds for sessions it does not show (`__rompOrphanStateSids`) to their owner's pane, through the existing orphan/adopt mechanism.
- **Ordinary store bytes are unchanged.** `save()` omits a column only because it is held, never for being momentarily empty; every write of the existing move and prune paths is pinned byte-for-byte against main's sequence.
- A dashboard that reloads while a hold is pending loses nothing main did not already lose: the document, and the create with it, goes with the page.

## Tests

- `tests/test_chat_split.py` runs the hold under node against the stub (50 tests; the hold, the orphan hand-off, the per-write bytes, and the busy `colEmpty` hold fail before their fixes): held under a peer's write with no toast and no store write; a still-busy flip, a flip from another column and one from no chat frame change nothing; a user's write meanwhile lists no empty entry; the store moving again before the flip is reconciled with the hold standing; the flip closes pane and gutter with the store untouched; a create landing is claimed and the column kept; a peer re-listing the column ends the hold; the user's own cross is still refused. Also: the state a closing page holds for a session it does not show travels to that session's owner; per-write byte pins for the move and prune paths equal main's sequence. **Fails at `ea91a228`** (the storage event closes the column at once).
- `tests/test_chat_split_served.py` step 13 runs the hold in the served shell with the page's busy answer and its flip set by hand; step 14 runs a real provisional create whose typed text and a staged message must reach the already-running session's pane after a peer's removal (the peer's write as a synthetic storage event, the resolution driven through the shell's own focus hop since the lab's sessions are not live). Step 14 fails on the first commit of this branch. Step 15 runs a real provisional whose folder is missing but creatable: the kernel's real folder question arrives while the column is held; the column stays with the prompt over the live tab, a dismissed question leaves the failed tab with its text, and the tab's ✕ closes the held column. `ui/webview/dir-question-busy.test.ts` (53 tests, booting the page's composer state in production order) pins the tab staying through the question, one tab per retry, the dismissed question becoming a failed tab with its text, a stale reply from a superseded create being ignored, the no-request-id fallback for an older kernel, that there is never a second confirm, and that no other session's draft ever changes. `tests/test_new_session_dir.py` and `tests/test_kernel_create_session_ack.py` pin that every direct reply to a create, including the focus that ends a successful create and the malformed-flag refusal, echoes the page's request id (fail before with `KeyError: 'rid'` and an unexpected-keyword error). The reload tests rebuild a failed tab from the persisted store whatever it held (typed, staged, cited, attached, or only the reason), check the orphan sweep reached the disk (staged entries included), check the one busy announcement at boot, route each tagged reply to the create it names with another create pending, rebuild a create still pending at the reload as a failed tab holding its staged text, check the lost-upload notice precedes the replayed notices, and check that a settling create's staged messages (a context-only citation included), unsent chips of either flavour and attached files all reach the real session on both settlement paths and across panes, joining onto an existing draft rather than over it, with nothing left under the retired id for the next boot's sweep; that a column with an upload of its own in flight stays busy until the kernel's answer and never re-sends the file, that a message held for two uploads is sent by session id only once both completed and is kept as a draft with a notice if either failed, that a session with an upload in flight refuses to move from every entry point, that an upload's answer goes to the box that shipped it, that a detached host fails its uploads, that a column held only by an unshown session's upload asks before closing, that an emptied column closes once its upload lands, that a comment's late file never reaches the composer, and that a session moved back with nothing to carry can still send a message held for its upload; `tests/test_new_session_dir.py` pins the request-id echo on every direct create reply; `ui/webview/staged-messages.test.ts` pins the stack's whole-item move, that a context-only staged citation survives a reload (it used to vanish at every reload — a silent loss this branch removes, called out in the commit), and that a persisted row with no words and no valid citation is still dropped; the stale rule's truth table and the late-warning toast are executed, not regex-pinned. `tests/test_session_env.py` and `tests/test_session_auth.py` follow the create helper's new signature.
- Source pins in `ui/webview/chat-split.test.ts`; `provisional.test.ts` widened for the discard site.
- Gates: `npm run typecheck`, `npm test` (4952 pass, the extension's own tests included), `npm run build`; `test_chat_split.py`, `test_inline_js_parses.py`, `test_new_session_dir.py`, `test_kernel_create_session_ack.py`, `test_session_env.py`, `test_session_auth.py`, `test_kernel.py`, `test_sdk_backend.py` and every other test file naming a changed symbol, `test_chat_split_served.py`, `test_ship_reship_served.py` (its driver now attaches through the composer's own file input, marked `data-for="composer"`; the old selector picked the comment clip's hidden input), `test_drop_anywhere_served.py`, `test_federated_split_cutfloor_served.py`, `test_local_split_cutfloor_served.py`, `test_local_redial_reask_served.py`, `test_chat_split_host_served.py`, `test_landing_bundles_built.py`.

## Review

Twenty-six rounds of adversarial review (GPT-5.6 Sol) on this branch. Every finding about the bug itself, about creates, request ids, settlement, hand-off, holds, and current kernels' tagged uploads is closed and pinned by tests. The one class the reviewer still flags, and that this PR accepts and documents rather than fixes, is below.

- **Accepted residual, stated plainly.** A kernel older than 2026-09-01 answers an upload without naming it, so its file is attributed by a guess. This branch makes such an answer never fire a waiting send itself and cancel the waiting send of the composer it lands on (main fired that send unconditionally). What it does not do: a waiting send that was armed *before* the answer and then moves with its session (a settling create resolving onto the session, a peer moving the session, a pane still running an older bundle during an in-place rebuild) can still fire with the guessed file attached. A durable per-attachment provenance mark was built to close that (four rounds) and withdrawn: each round found a new way for the mark to drift from the file list, and the cost was out of proportion to pre-September kernels. Updating the kernel removes the class entirely.

## Left as on main

- A local connection that never returns has no terminal event, so an upload waiting on it stays pending until the user closes the column (the close asks first) or reloads the page; from the first pane, which has no close control, only a reload.
- A reconnect can save a second copy of an upload on the kernel, and there is no cleanup of the drops folder.
- An owed dashboard reload after a column closes waits for the reload core's 60-second backstop, since the closing page's ending notice dies with its document.
- A file picked in VS Code lands on whichever session is active when the picker answers, as on main.
- Before 2026-09-01 a kernel's upload answer named no upload, so with such a host a file is attributed by a guess and can land on the wrong session (as on main); see the Review section for what this branch does and does not do about the send that waits on it. Users of such a host lose the automatic wait-and-send for uploads and send by hand after the notice. In the F5/watch development workflow an extension host older than the webview sends an unmarked pick, which is read as such a legacy answer (installed builds ship the two together).
- A file picked in VS Code (or the kernel's own dialog) is classified as a pick where it arrives, so it lands on the active composer and never settles an upload; an answer with no upload id is therefore an older kernel's (before 2026-09-01): it attaches the file to the best-guess upload (the same saved name on the answering host, in the order the uploads were sent; the active composer when nothing matches) and retires that chip, but never sends a message itself — whichever composer receives such a file has its waiting send cancelled with a notice, so the user checks the attachment and sends by hand — including the pane that later adopts the file when the answering document does not show the session; the file then stays as an ordinary attachment: no automatic send may be armed on it, moved onto it from a settling create, or fired from it meanwhile; kernels from 2026-09-01 on name every answer and are unaffected; with no match it lands on the active composer, and with no active tab the user is told rather than nothing happening.

Tier: fix


alowet:
Superseded by #1774, which fixes the reported bug (a peer dashboard's write closing a column whose page has a create in flight) with the same shape this branch converged on: hold the dropped column, close it on the page's idle signal, re-read the store first. Closing this draft rather than re-basing 27 commits onto that model.

For the record, the review of this branch surfaced further edges in the create and upload flow that #1774 does not touch and that may be worth separate, smaller fixes; each has a synthetic reproduction in this branch's tests:

- A late reply to a create the user has since replaced (folder question, warning, focus) is read against whichever create is pending; a request id echoed on the kernel's direct replies fixes it.
- A create that resolves to an already-running session shown in another pane strands its typed text, staged messages, unsent citation chips and attached files in the closing document.
- A failed create's tab, and a create still pending at a reload, are lost with the page; persisting them with the drafts brings them back.
- The folder question ("that folder isn't there") carries the text in document-local state that a close destroys; keeping the provisional tab through the question avoids that.
- A session with an upload in flight can be moved out from under its upload; an upload's answer is routed to the comment box whenever one is open, whichever composer started it.
- Kernels older than 2026-09-01 answer uploads without naming them; the client's guess can fire a waiting send with the wrong file.

Branch kept at `alowet:chat-close-busy` for reference.

PR #1787, opened 2026-09-16
The kernel's and the bus's servers bind without a reverse lookup, and CI's bats step carries a per-test bound: the macOS bats leg completes

henrypinkard:
The bats macOS CI cell (a manual dispatch) was cancelled at the job's 35-minute ceiling on every run, and the last dispatch died mid-way through the `romp tag` tests. The log says why, and it is not a retry loop: of 572 tests recorded before the cut, the median took 0.21 s, but 49 took 35.4 to 36.7 s each and the sum was the whole budget. Every slow test starts a Python HTTP server, a stub kernel in the test body or the real postal bus in setup, and the one stub test that ran fast installs the mock curl and never reaches its stub. The same tests take 0.2 to 2 s on the Linux cell of the same run.

**The cause is Python's stock server bind.** `HTTPServer.server_bind` runs `socket.getfqdn(host)` after `bind()` and before `listen()`; on GitHub's macOS 15 and 16 images the reverse lookup of `127.0.0.1` blocks about 36 seconds (reported against `actions/setup-python` and measured the same by others), so every server paid it once before it could answer, and the CLI or client waited on the port file or the ping behind it. The tag verb itself has no retry road: `_tag_curl` is one bounded curl and prints the kernel's words on a refusal.

**The product fix.** The kernel and the postal bus construct `_LoopbackServer`, a `ThreadingHTTPServer` whose `server_bind` binds through `socketserver.TCPServer.server_bind` and sets `server_name` to the bind address without asking the resolver; `server_name` feeds nothing either program reads. A Mac with a stale resolver would otherwise keep the kernel or the bus from answering for as long as the lookup takes. The twelve bats-inline Python stubs bind the same way.

**The harness bound.** CI's bats step now carries `BATS_TEST_TIMEOUT=180`, bats' own per-test bound (since bats-core 1.8, in the pinned 1.11.1 and in Homebrew's 1.14.0 alike, implemented with a background sleep and pkill or ps), so a hung test fails in minutes and is named in its TAP line instead of the job dying nameless at the ceiling. No coreutils timeout is involved.

**Pins.** `tests/test_server_bind_skips_reverse_dns.py`: the stock class does call `getfqdn` at construction (the positive control), the kernel's and the bus's classes do not and name their address, both serve sites construct them (source pins), and no bats-inline stub constructs the stock class. `tests/test_ci_bats_bound.py`: the step's env carries the bound within a sane range and the step still runs every bats file. Red first at the base by copy-aside of the touched files.

**Proof.** The whole bats suite on Linux under the pinned bats with the same bound; the full Python suite and the whole extension suite on the head; and a manual dispatch of CI on `stack/romp_timeline-bats-retry`, judged by the bats macOS leg completing with the tag tests green (the run id and its legs in the review thread). The branch carries main as a pure merge, so the dispatch also runs main's own macOS fixes for the resolved-path tests and the re-delivery stamp flake. Two source pins in other modules that anchored on the renamed constructor line are re-anchored on the new one.

Tier: fix


henrypinkard:
Manual CI dispatch on `stack/romp_timeline-bats-retry` at this head: run [35129966730](https://github.com/romp-on/romp/actions/runs/35129966730).

- Shell (bats, macos-latest): completed in 6 min 57 s, 669 tests, where every earlier dispatch was cancelled at the 35-minute ceiling after 572. The tag tests are green (572, 573); only the two romp-serve probe tests exceed 20 s, at 60 s by design. Nine tests the leg never reached before fail, all macOS portability and none touching this change: `tests/romp-cli-scope.bats` 206 to 209 and 219 (the direct-path exec-in-place assertions; the ignored line's fake-call count), `tests/romp.bats` 590, 591, 593 (GNU `sed -i 's/...'` on BSD sed), `tests/romp-serve.bats` 425 (a blocking interpreter is not refused within the version probe's bound on macOS: the coreutils-timeout hazard).
- Python 3.10 and 3.13 (macos-latest): 1 failed, 13553 passed each, the one being `tests/test_git_pointer_file_bytes.py` (a torn pointer file's fault keyed under the handed path), in no PR yet.
- Every Linux leg success. The extension job failed on one served lab of 430, `ServedMobilePickerOrder`'s fold test (`t:cache/-` against `t:cache/`); main's own CI at the merged parent passed that job, and run alone it passed 7 of 7 twice at this head and twice at the base.

On this head locally: the whole bats suite on Linux 669 ok under the same bound, the full Python suite 13969 passed with none failed, the extension suite 4967 with none failed.

PR #1796, opened 2026-09-16
The version probe bounds itself without coreutils timeout, and the bats tests read macOS: the nine reds the macOS cell reached for the first time

henrypinkard:
The bats macOS cell completed for the first time on the dispatch for PR 1787 and read nine failures it had never reached. All nine are macOS behaviours in the tests or in one product road, none in the tag verb:

- `tests/romp.bats` 590, 591, 593 (three tag tests): `sed -i 's/.../' file` is GNU's shape; BSD sed takes the word after `-i` as the backup suffix and then reads the expression as the file ("invalid command code f"). The four edits of the curl mock go through `_sed_inplace`, a temp file written back into the file itself so the mock keeps its mode and inode. No product script uses `sed -i`.
- `tests/romp-cli-scope.bats` 206 to 209 and 219: the direct-path tests compare the fake CLI's parent pid with this shell's, and on the macOS leg that parent was an intermediate process; the ignored-line test counts a fake systemd-run's calls, and the count differs from the Linux shape. The scope is a systemd feature, off wherever `systemd-run` is absent, so the wrapper is never on macOS's launch path; the five skip on Darwin with the reason stated, never a Linux shape asserted there.
- `tests/romp-serve.bats` 425, the product road: `bin/romp-serve`'s version probe was bounded by coreutils `timeout -k 1 5` and documented as unbounded where there is none (a stock Mac); the leg ran a blocking interpreter for its whole life and never refused it. The probe now bounds itself there: the probe runs in the background with a watchdog beside it that sends TERM at five seconds and KILL a second later, and the watchdog is put down when the probe answers on its own. The bound's codes read 143 and 137 as they read 124 and 137 from `timeout`, and the refusal is the same line. The watchdog is a plain background child acting on no trap of this script's (round five's watchdog acted from a TERM trap under `set -e`); a TERM to the script mid-probe still leaves no probe file behind, as the sibling test pins. A PATH without `sleep` cannot time the bound, so the watchdog stands down and the probe runs unbounded there as it did before, never killed at once (the bare-PATH probe test, which the first cut of this failed).

**Red first.** Two new tests in `tests/romp-serve.bats` hide `timeout` and `gtimeout` from PATH through a symlink farm of every other tool, then drive a blocking interpreter and a TERM-deaf one: the refusal must come within the bound. At the base they run the unbounded road and never see it.

**Proof.** The three touched bats files and the whole bats suite on Linux under the pinned bats with CI's per-test bound; the Python modules that read `bin/romp-serve`; a manual dispatch of CI on `stack/romp_timeline-macos-bats`, judged by the bats macOS leg completing with zero `not ok` (the run id and its legs in the review thread).

**A low for a later sweep, not this PR:** 72 Python test modules construct the stock `HTTPServer` in their own stubs and so reverse-resolve their bind address at startup; the macOS Python cells pass in 16 to 21 minutes today, so they do not pay the lookup on that interpreter, and the sweep is its own change.

Cut from PR 1787's head, which is in main's history since 1787 merged; the diff against main is this change alone.

Tier: fix


henrypinkard:
Manual CI dispatch on `stack/romp_timeline-macos-bats` at this head: run [35140483558](https://github.com/romp-on/romp/actions/runs/35140483558).

- Shell (bats, macos-latest): success in 6 min 32 s, plan 1..671, zero `not ok`, zero timeouts. The five CLI-scope direct-path tests skip there with their reason; the blocking-interpreter probe test and the tag tests pass.
- Python 3.13 (macos-latest): 1 failed, 13553 passed, the one being `tests/test_git_pointer_file_bytes.py` (a torn pointer file's fault keyed under the handed path), in no PR yet. Python 3.10 (macos-latest) was cancelled at the python job's 25-minute ceiling after 25 min 27 s on a slow runner; the same cell completed in 24 min 43 s on the previous dispatch, and nothing in this change is Python.
- Every Linux leg success, and the extension job success.

On this head locally: the whole bats suite on Linux 671 ok under CI's per-test bound, the two new probe tests red at the base and green here, every Python module that reads `bin/romp-serve` green.

PR #1802, opened 2026-09-16
The version probe's watchdog signals the process group, stands down with its sleep and owns the 143 code on its road; the CLI-scope tests read their pid a way bash 3.2 has

henrypinkard:
The four lows from PR 1796's read, on `bin/romp-serve`'s version-probe watchdog and the CLI-scope tests, filed after the v0.16.0 tag so the probe did not move under the release's combined proof.

- **The 143 refusal is the watchdog road's alone.** The bound's own exit codes are the road's: `timeout` gives 124 at the bound and 137 when the KILL reaches it; the shell's watchdog gives 143 for its TERM and 137 for its KILL. An interpreter that exits 143 of its own accord on the timeout road, where no watchdog runs, is no longer refused as unresponsive: it is the no-version leg, as any exit without a version line. The refusal message names the road's codes.
- **The watchdog stands down with its sleep.** The probe runs under job control in a subshell of the function's own, so every background job there is its own process group: the stand-down signals the watchdog's group and takes its `sleep 5` with it, where a pid-only TERM left a five-second sleep orphan per probe of a good interpreter.
- **The watchdog signals the probe's process group**, as coreutils `timeout` does, so a helper the interpreter started dies at the bound with it, where a pid-only signal let it outlive the bound. The subshell's stderr is dropped, so bash's job-status lines never reach the log.
- **The CLI-scope direct-path tests run on Darwin again.** Their expected parent pid was `$BASHPID`, a bash-4 variable the repo's own portability pin lists; under a stock Mac's bash 3.2 it expanded to nothing and the four tests read as failures on the first completed macOS leg. The helper reads this shell's pid a way bash 3.2 has, `${BASHPID:-$(exec sh -c 'echo $PPID')}`. The ignored-line test's fake `systemd-run` switched on `wc -l` output, which BSD wc pads with spaces, so no arm matched on macOS; the count is stripped. The five Darwin skips are gone.

**Red first.** Three new tests in `tests/romp-serve.bats`: an interpreter exiting 143 of its own accord (accepted on the timeout road, refused on the watchdog road with the message naming 143), a good interpreter on a hidden-timeout PATH whose `sleep` records its pid (no orphan outlives the probe), and a blocking interpreter with a child of its own on that PATH (the child dies at the bound). At the base the first is refused, the orphan lives and the child survives. The CLI-scope changes have no Linux red: their red is the macOS leg's own log, and the dispatch is their proof.

**Proof.** The two touched bats files and the whole bats suite on Linux under the pinned bats with CI's per-test bound; the Python modules that read `bin/romp-serve`; a manual dispatch of CI on `stack/romp_timeline-probe-lows`, judged by the bats macOS leg completing with zero `not ok` and the five CLI-scope tests running rather than skipping (the run id and its legs in the review thread).

Tier: fix


henrypinkard:
Manual CI dispatch on `stack/romp_timeline-probe-lows` at this head: run [35150661614](https://github.com/romp-on/romp/actions/runs/35150661614), success on every leg.

- Shell (bats, macos-latest): success in 6 min 55 s, zero `not ok`, zero timeouts. The five CLI-scope tests (206 to 209, 219) run and pass there instead of skipping; the probe's watchdog tests pass on the hidden-timeout road (the 143 test skips on macOS by its own guard, since its timeout-road half needs coreutils timeout).
- Python 3.10 and 3.13 (macos-latest): 13599 passed, 489 skipped, none failed, the first fully green macOS Python cells on any dispatch.
- Every Linux leg success, gitleaks and the extension job included.

On this head locally: the three new probe tests red at the base by copy-aside of `bin/romp-serve` and green here; the whole bats suite on Linux 674 ok under CI's per-test bound; every Python module that reads `bin/romp-serve` green.

PR #1807, opened 2026-09-16
The portability pin reads its surfaces from git, bounds its list and its scan and names the stalled half, and scans with one grep per file

henrypinkard:
The confirming macOS run on the v0.16.0 tip (35157245689) had its bats leg red on one test: `tests/shell-portability.bats`'s pin of the shell surfaces sat the job's whole 180-second per-test bound (the bound named it: "failed due to timeout" at the file list) where the proof run at 781f687e ran it in 3.3 s, and no shell file changed between them. Three changes make the pin stall-proof, bounded and light:

- **The surfaces come from git's tracked list**, not a glob over the working tree: `git ls-files` over the same paths, filtered by shebang, gives the same 21 files, and a stray file another test leaves under `bin/` or `tools/` on a shared runner cannot enter the set.
- **The list and the scan run under bounds of their own** (`_bounded_fn`: coreutils `timeout` where it exists, else perl's `alarm`, the child dying to SIGALRM), 30 s for the list and 120 s for the scan, and the pin says which one stalled instead of leaving it to the job's bound to name the test alone.
- **One grep per file.** The scan ran two greps per construct family per file, 924 processes for 22 families over 21 files, minutes on a slow macOS runner; it now runs one grep per file with every family's pattern joined and names the family in the shell (bash's `=~` is the same POSIX ERE grep reads), 42 processes. A line grep saw that no family claims is still reported, as an unclassified construct, so a hit is never lost.

**Red first.** Two new tests: a `head` on PATH that blocks is cut at the list's own bound in seconds (at the base the bounded runner does not exist), and an untracked shell file planted under `bin/` is not a surface (at the base the glob lists it). Both run against the base's helpers with the new tests grafted on.

**Proof.** The pin file twice with `--timing`, the whole bats suite on Linux under the pinned bats with CI's per-test bound, and a manual dispatch of CI on `stack/romp_timeline-portability-pin`, judged by the bats macOS leg completing green with the pin's own time in its log.

Tier: fix


henrypinkard:
Manual CI dispatch on `stack/romp_timeline-portability-pin` at this head: run [35160826898](https://github.com/romp-on/romp/actions/runs/35160826898), success on every leg.

- Shell (bats, macos-latest): success, zero `not ok`, zero timeouts. The pin itself ran in 0.4 s there, where the v0.16.0 tip's confirming run sat the whole 180-second bound; the blocking-head test is cut at its own 3-second bound (3.2 s), the stray-file test 0.3 s.
- Python 3.10 and 3.13 (macos-latest): 13633 passed, 494 skipped, none failed.
- Every Linux leg success, gitleaks and the extension job included.

On this head locally: the two new tests red at the base by grafting onto the base's file and green here; the pin file green twice with the pin at 220 ms; the whole bats suite on Linux 673 ok under CI's per-test bound.

PR #1810, opened 2026-09-17
Settings, Automation, Model: Always fast, and Retry upgrades after downgrades

amith-ananthram:
## What this adds

Two switches under **Settings, Chat, Model**, both off by default, both kernel-side (stored on the kernel like the judges' Fast mode boxes, stamped, following to every connected machine's kernel):

- **Always fast.** Every session runs Claude Code's fast mode whenever its model allows it (Opus-only, billed at a premium). The kernel arms the CLI's fast-mode opt-in at each connect for a session whose model is Opus; a session that lands on Opus later, by a pick or by an automatic fallback, reconnects at the end of its turn to arm it then. A session you set to **Slow** from its statusline stays slow until you set it to **Fast** again.
- **Retry upgrades after downgrades.** When a session's model changes to a lower tier without a pick (the automatic fallback the Completed card reports), the kernel asks for the picked model again every ten minutes by reconnecting the session, now if idle, at the end of the turn if busy, so nothing is cut. When a turn is served on the picked tier again, a second Completed card says the session is back and the retry ends.

The user's case (2026-09-17): sessions get moved from Fable to Opus mid-work by a trigger in the task's context, and asking for Fable again later works once that context has aged out. The cause is outside romp's view, so the retry is a fixed cadence; each attempt keys on a turn boundary.

## Design points

- The switches ride the judge-knob machinery end to end: `_set_judge_state` stores (`always-fast`, `retry-upgrade`), `_JUDGE_SETTING_FIELDS`, `_GT_STORES`, `/version` top level and the cross-machine `settings` dict, a sibling socket arm (`setAlwaysFast`, `setRetryUpgrade`) with `_as_bool` and the stale-gesture answer, `KERNEL_SETTING` in federation. The SDK backend reads the stores by path at use time (no `jd` import there; `thinking_summaries_on` is the precedent).
- **Always fast** is one expression, `SdkSession.fast_effective()`, read by both `_options` (the flag file) and `_amain` (the per-connection snapshot), so they cannot disagree. It arms only the `fastMode` flag-settings key, never the literal `/fast on`, which on a non-Opus session makes the CLI switch model. The model tested is the CLI's reported id, else its name, else the pick; the Opus family only.
- Two things beat the switch: an explicit Slow, remembered as the reg's new `fastOff` by `set_fast` (a Slow pick that flipped back at the next reconnect would be a switch nobody asked for), and a refusal reason the CLI persisted (`liveFastReason`), so a machine-wide rule cannot loop refuse → reconnect → refuse. The rule's refusal is logged once per reason; the user's own ask keeps its toast-and-revert path unchanged.
- **Retry upgrades** arms in `_learn_model`'s downgrade branch beside the fallback card, fires from the kernel's jobs loop (`_retry_upgrade_tick`, every 30 s looking; attempts ten minutes apart) through `SdkBackend.retry_model_upgrades` → `request_reconnect`, which re-asserts `--model <pick>` on a fresh CLI and waits for the turn's end. It ends on a PARENT turn served on the pick's tier (`_learn_model(..., served=True)` from the AssistantMessage; the init's report of the configured model and a context refresh are not evidence), on a pick of the user's own (`set_model`), with the session, or with the switch. The happy end mints a card through `on_model_restored` → `jd.mint_restored_card` (the fallback card's twin, same dedupe). A fallback that happens again while armed is logged, not carded again.
- Nothing is injected into any session; the two hooks and the toast paths speak to the dashboard only.

## Review round (an independent three-lens pass over the diff, 2026-09-17)

Twenty findings held up; several were one cause seen from several sides. All fixed here:

- **A pick to Opus left the session slow.** The control channel's context refresh ran the model-change hook before storing the reported id, so the rule read the old id and asked for nothing, and every later learn of the same model took the early return. The id is stored first now, with an executed test driving the refresh.
- **The rule's refusal was charged to the user's own ask, and forgotten at the next flagless connect.** The rule's refusal is now decided from the pre-clear ask and only under the switch, and remembered in its own key (`fastRuleRefused`), because a flagless connect reports the opt-in reason that blanks `liveFastReason`; the memory lifts when the CLI reports fast on or the user toggles the session.
- **A reconnect flap when the served model differs from the pick.** An Opus pick served a Sonnet fallback reconnected at every turn's end. `fast_effective()` now arms on any known face of the model (the pick, the reported id, the name), and the hook asks for one reconnect per connection.
- **The switches reach running sessions.** Flipping Always fast reconnects every live session whose flag should change (deferred to the turn's end); flipping Retry upgrades arms every session already sitting below its pick. Off ends a standing retry everywhere, so no "back on" card can credit a retry that never ran. The gear rows and the reference say so.
- **Cards and logs.** The re-fallback while armed asks the store as before (its existence-keyed dedupe decides: nothing while the swap's card stands, a fresh card once you cleared it) and is logged once per attempt with the real wait. The fork dialog's explicit off, and a parent on Slow, count as a Slow pick.
- **Census.** The new jobs stage joined `_PerfStats.JOBS` and the two tests that pin the census; the browser recut pins the Model section head; the boolean-refusal table has both ops; the tick has executed tests; two vacuous or half-stated pins are real now.

## Tests

- `tests/test_always_fast.py`: the store reader, the Opus rule (any known face), the session's ask / a Slow pick / a refusal each having their say, the connect and the snapshot reading one expression, the reconnect on a model that can run fast arriving on a locked connection (and not when the flag is there, the switch is off, or the session is on Slow), the refresh path executed, the rule's refusal said once and remembered across a flagless connect, no flap across a served fallback, the switch reaching running sessions both ways, the fork's Slow.
- `tests/test_retry_upgrade.py`: arming only with the switch on, a re-fallback logged not carded, the tick asking once per cadence by a deferred reconnect and standing down when the switch is off or the session ended, the way back (a served turn, a higher tier too, the hook's arguments), a pick of the user's own ending it, the kernel wiring.
- `tests/test_model_switches_kernel.py`: the stores' validation and defaults, `/version` and the stamps, the socket ops (apply, propagate, stale stand-down, malformed refusal), the back-on card and its dedupe on a private synthetic sid.
- `ui/webview/gear-model-switches.test.ts`: the Model section between Thinking and Chat history, the rows' grammar and words, the stamped emitters, the stale maps, the fill, the mixed marks, the broadcast set, the docs.
- Extended: `test_setting_gesture_order.py` (22 stores, both tables), `test_kernel_settings_sections.py`, `gear-tabs.test.ts`, `test_judge_settings_sync.py`, and the two exact-string pins on the fast flag's source lines.

Tier: feature

## Quiet gate (live finding, 2026-09-17)

The first live run showed the rule's deferred reconnect firing at a turn's result while three subagents and a background task still ran inside the CLI; the reconnect force-killed them. A switch is nobody's gesture on a session, so it now cuts nothing: its ask stands until the session is quiet (no turn in flight or queued, no live subagent, no background task) and is carried at the exact event that ends the last of it (the settle, a subagent's stop, a task's end), with the 30 s tick as backstop, as an immediate reconnect whose loop-side re-check stands. The retry's attempts follow the same road; no attempt stacks on a standing ask. The CLI's `sdk_opt_in_required` is also read back as the truth that a connection has no flag, which an attach to a surviving CLI after a kernel restart got wrong.

## Quiet gate, round two (live finding, 2026-09-17)

Four switch reconnects still ended a CLI by force two minutes after "the session is quiet", and one dropped a background task. `inflight` counts fed turns only; a turn the CLI opens itself from a background task's notification never passes through the feeder, so the session read as quiet while the CLI ran dozens of tool calls a minute. quiet() now reads the stream's own busy signal (`_cli_working`), the waker looks once more the instant before a switch's reconnect leaves the client and stands down if the CLI has started work, and a switch's reconnect gives the host a twenty-minute end grace so a turn that starts in the last instant is finished rather than killed. A user's own reconnect is untouched.

## Fast restart (2026-09-17)

The transcripts' per-message speed showed why a downgraded session stayed slow: the CLI settles fast mode when a turn begins, for the model it is configured on then, so a turn that began on the pick and fell back to Opus at its first request ran the whole turn at normal speed. Under Always fast, such a turn is now stopped at its first reply, before it has streamed, and re-asked at its settle when the result proves the stop took effect; the re-asked turn begins on Opus in fast mode. The stop is the polite rung only, never the user's escalation ladder; it stands down when a stop already stands, when a message was forwarded into the turn, or when a previous re-ask on the connection still came back slow; a reply that completed first, or a turn the user stopped too, is left alone. The cut is stamped as romp's and the re-ask carries a cut signature, so it is never filed as the user's stop. Reviewed by five adversarial reviewers with per-finding verification before landing.

## Requested-model mark in the picker (2026-09-17)

While an automatic fallback answers instead of the picked model, the session's model picker marks the requested model with a yellow tick beside the blue one on the model that answers. Its tooltip says why ("Requested model blocked due to safety classifiers." once the CLI has named them) and what romp is doing: the retry cadence and next attempt when a retry is armed, that auto-retry is on and applies at the next start when only the switch is, or where to turn it on. The mark keys on the model the API last served, not the configured one, so it neither vanishes at a turn's start nor flaps; a pending pick shows the badge's dots instead. Found and fixed on the way: the retry's arm was lost when a kernel restart re-attached to a surviving CLI, so a standing fallback was never retried; the attach now re-arms it.

Follow-up (2026-09-17, evening): the tooltip names the refusal category, a fallback that predates the kernel has its cause read off the transcript at the attach, and the unknown case says "not answering" rather than "unavailable".

Follow-up (2026-09-17, night): a served fallback that merely re-observes the standing one keeps the cause on record (the attach seed no longer races the replay), and the timeline's lane picker wears the requested-model tick with the same tooltip.

Follow-up (2026-09-17, later): the requested-model row wears a permanent sub-line with the tooltip's gist in both pickers.

Indication that you're being served a downgraded model (with text describing auto-retry upgrade behavior, if enabled):
Screenshot 2026-09-17 at 1 42 55 PM

Screenshot 2026-09-17 at 1 43 04 PM

Follow-up (2026-09-17, audit): a permission or picker ask waiting on the user counts as busy for the switch reconnects.


henrypinkard (approved):
Approved on the maintainer's decision of 2026-09-17: both switches wanted, relocated to Settings, Automation (a Model section beside the Nudges, the tab's own row shape) on this branch by the review session; behaviour unchanged.

amith-ananthram:
@henrypinkard just a heads up, I don't think this is working correctly just yet; was dogfooding it last night and the -> fast post downgrade seems to be broken.  will keep fiddling with it today but might want to hold off on merging!

amith-ananthram:
ah okay, it's not a bug so much as a design issue -- basically, upgrade retries (every 10 min) typically fail and when they do the model falls back to opus slow for that turn which plays poorly with the "always fast" config.  have a fix in the works, basically when an upgrade retry fails, if "always fast" is enabled, eagerly interrupt the response, toggle back to "fast" and then resume.  explains some of the weirdness i was seeing overnight, should be in ~an hour after i test it a little.

amith-ananthram:
@henrypinkard okay issue described above is fixed now

henrypinkard:
Closing this pull request; the facts behind the decision, and what happens next.

- The two switches, Always fast and Retry upgrades after downgrades, were approved on 2026-09-17 at head 1be4e717, which also moved their rows to Settings, Automation (a Model section beside the Nudges, in that tab's row shape).
- After that approval the branch received nine more commits in about eight hours, each resetting the approval and adding behaviour the review had not seen: an interrupt-and-re-ask of a fallen turn, and a requested-model tick in the model pickers with fallback-cause bookkeeping. A pull request that keeps changing what it does cannot carry one approval, so the maintainer decided to split it.
- The finished pieces will be re-filed from the maintainers' side as separate pull requests with these commits cherry-picked as they are, authorship preserved: (1) the two switches with the Automation relocation and the later quiet-gate hardening (the CLI's own busy signal, the last look before the teardown, the pending-ask hold); (2) the requested-model tick and its bookkeeping, stacked on the first.
- The interrupt-and-re-ask commit is left out for now, with two defects to fix before it can land on its own: the re-ask line carries only the `<!-- romp-injected -->` marker while the chat seam's stamper (`_stamp_interrupt_causes`) keys on `<!-- romp-system -->`, so the cut is drawn as the user's own stop instead of "restarted in fast mode"; and `_do_interrupt`'s failure path sends SIGINT and raises the ladder, so a kernel-initiated stop can become a signal filed as the user's. A separate pull request for it, with those two fixed, would be welcome.
- Noted for the tick's pull request, not blockers: the tooltip says "auto-retry" where the switch is named "Retry upgrades after downgrades"; a re-fallback for a different cause keeps a stale safeguards cause; the working-state yellow is reused for the requested mark; the attach-time transcript read reports no misses.

Thank you for the work on this; it is landing, in pieces.

PR #1813, opened 2026-09-17
fix: keep lazy hydration sources with their restored views

lczh:
Tier: fix

Restoring a second transcript under the same stable session can strand a lazy body held by the first restored view. The second restore replaces `_LAZY_FILES[session]`, so hydration either cannot find the first transcript or follows another document's path. A feed build that reads that body then raises and leaves the last successful feed on screen. The transcript itself can still exist and its checkpoint can be valid.

Bind each restored lazy body's source path from its verified document, privately on the body sentinel, and retain that document's file map in `LazyIndex` for later materialization. Atoms-only restores bind the source too. Shallow atom copies keep the binding; a removed source still raises instead of redirecting to another document. Legacy unbound descriptors retain their existing lookup. There is no wire or checkpoint format change and no accumulation of historical paths in a shared map.

## Review round

The review of the first cut found a race in the new source resolution and two claims without a test.

- The race. The first loop of `hydrate` checked the memo under the lock, released it, and only then read the atom's body for its bound source. A peer thread finishing the same atom in that window leaves a plain dict, which carries no source, so the loop fell back to the per-session map; after another leaf of the same session had restored, the map named the newer document and the loop raised "no file known" for an atom whose body was already in place, failing the whole call. The loop now reads the body once: a body that is no longer a `_LazyBody` counts as filled, its bookkeeping finished from the memo when the entry still stands, the tolerance the second loop already had. Two tests pin it: a memo double whose first miss performs the peer's put and fill, and two real threads with the second held by events between its memo miss and its body read while the first runs a complete call. Both are red on the first cut with "no file known"; on main, whose first loop never reads the body, the threaded test's worker raises the same "no file known" before the paused read is reached, and the test surfaces that error rather than a timeout.
- The pin beside `_restore_prefix_atoms`. A bound body whose fsid is absent from its document's files map stays a loud failure and never borrows a path from another snapshot. One test: two documents under one session sharing a fsid, the api one restored last, then a `LazyIndex` over the web document with that fsid removed from its files map; hydration raises "no file known". Under the mutation that falls through the sentinel, the test reads the api copy and fails with "the record at its offset is user-api".
- The legacy unbound road. Two tests: a `LazyIndex` over a document with no files key builds an unbound body that hydrates through the per-session map the restore's own write primed, and the same with no entry is refused with "no file known". The primed test goes red when the write is a no-op or the fallback is None; the no-entry test pins the refusal alone.

Also folded: the three copies of the files-map comprehension are one helper, and the test module scrubs the event model's state it touches in tearDown, since every test module shares that one module object.

Validation:

- The six original regressions fail with the kernel at the base: four with "no file known", two with "the record at its offset is user-api" (a lazy body borrowing the other document's path). The five review-round tests are red as described above.
- `test_lazy_source_lifetime`: 11 passed (the six original regressions and the five above), the same over ten repeated runs.
- `test_asm_checkpoint`: 91 passed, 85 subtests passed. `test_asm_index`: 26 passed, 3 subtests passed. `test_asm_mat_lru_release`: 9 passed. `test_chat_pages`: 43 passed, 27 subtests passed. `test_kernel`: 460 passed, 12 subtests passed.
- `git diff --check` clean; no pins on the changed lines in the other test modules or the extension.

Separate from #1812's containment of individual session derivation failures. This fixes the source lookup that caused the observed failure.


ksarma:
This is a review of b5be57efe, executed against that head, unchanged since the review ran. The first item is a race in the new source resolution that a small change to the first loop of `hydrate` closes; the other two are tests to pin claims the PR body and the new comments make.

### At this head

- kernel/event_model.py:7087. The first loop of `hydrate` checks the memo under the lock at line 7076, releases it, and only then reads the atom's message for its bound source path. A second thread that finishes the same atom in that window has replaced the message with a plain dict, which has no source attribute, so line 7087 yields the unbound sentinel and line 7090 consults `_LAZY_FILES`. That is the per-session map the PR stops depending on: once another leaf of the same session has been restored, the map names the newer document, the fsid is missing, and line 7092 raises for an atom whose body is already in place. The raise leaves the loop before any group is read, so the whole call fails, unrelated atoms included. The interleaving needs two consumers hydrating the same atom of one retained view at once, after a second restore under that session; the stripe comment at line 7095 names two threads hydrating the same atoms as a case the code expects. The consequence is the PR body's symptom made rarer, not removed.

  Executed against b5be57efe, this reproduces two ways: a memo double that misses once while doing the peer's put and fill made a one-atom call raise "no file known" with that atom already filled and its lazy marker gone; and with two real threads, the second paused at line 7087 while the atom was still lazy and the first ran a complete call that filled it, the second resumed into the same raise. Fix: read the message once in the first loop; when it is not a `_LazyBody`, count the atom as filled and continue, the tolerance the second loop already has for a missing marker; otherwise take its source path. It is safe: the only writer of an atom's lazy marker is `_restore_prefix_atoms`, which always pairs it with a sentinel body. Legacy unbound descriptors stay `_LazyBody` instances and keep their lookup. A refinement covers the instant between the peer's put and the end of its bookkeeping (a kind-u atom's tool result, the popped marker): re-read the memo under the lock and call `_hydrate_one` with the hit as the memo-hit branch does, counting the atom filled when the entry is already evicted. With the first form applied, both probes pass and the PR body's five modules stay at 166 passed, 115 subtests passed. Retiring the per-session fallback is separate cleanup, not needed here, since no production caller mints an unbound body.

### Also at this head

- kernel/event_model.py:6661. The comment above line 7087 says a missing bound source stays a loud failure and never borrows a path from another snapshot; nothing pins that. Changing line 6661 so a bound body whose fsid is absent from its document's files map falls back through the sentinel (`source_files.get(fsid, _UNBOUND_LAZY_SOURCE)`) leaves 322 tests across fifteen modules, named in the tests note, passing. Reaching that state on a written document takes a corrupt or hand-edited files map, since the writer keys files and fsids from the same ingested file. One test pins it: two documents under one session sharing a fsid, the api one restored last so the per-session map names it, then a `LazyIndex` over the web document with a files map that omits that fsid, asserting hydration raises "no file known". The fsid must be present in the per-session map and absent from the bound map: a document with no files key takes the unbound road by design, and a None fsid raises "no file known" on both roads, so neither separates the two. That test gives 2 passed at this head and 2 failed under the mutation, the failure reading "the record at its offset is user-api", the borrowed read.

### Smaller points

- kernel/event_model.py:7088. The PR body says legacy unbound descriptors retain their existing lookup, and no test exercises that road. With every loaded event model's `_LAZY_FILES` replaced by a recording dict, the 37 test modules that can mint lazy atoms made zero reads of it (683 passed), and replacing the write at line 6983 with a no-op and the fallback at line 7088 with None leaves those 37 modules as green. The one unbound construction left is the writer's identity check at line 6329, whose atoms feed `_pre_tree_identity` and are discarded. The write at line 6983 does run on every successful restore; what is unpinned is its effect. Two tests cover it: a `LazyIndex` over a document with no files key (the record location sits at the row level, not inside the lazy scalars), the per-session map primed with the leaf path under a real fsid, not None, asserting the built body is unbound and hydration fills it; and the same with no entry, asserting "no file known". The primed test passes at this head and fails under the mutation; the no-entry test passes under both, since the mutated fallback raises the same message, and pins only the refusal. Removing the fallback and the write instead is open too, since no production caller mints an unbound body, but hydration would then have to turn the sentinel into a loud `LazyBodyRead` (else open receives an object), the two harness resets that clear the map (tests/test_asm_checkpoint.py:108, tests/test_chat_pages.py:94) would need the name kept or a change of their own, and the PR body sentence would go; for a fix, the two tests are the smaller change.

The modules run for each item and their counts are in the tests note that follows.

ksarma:
Test runs behind the review of b5be57efe on PR 1813, by module and count; every run executed against that head in capped scopes.

- tests/test_lazy_source_lifetime.py: 6 passed at this head; the threaded test 10 of 10 in isolated runs. With kernel/event_model.py restored to the base sha 3b3b7c6 in a scratch copy: 6 failed, four with "no file known" and two with "the record at its offset is user-api". The fails-before claim is behavioural, not an import error.
- The PR body's five modules (test_lazy_source_lifetime, test_asm_checkpoint, test_asm_index, test_asm_mat_lru_release, test_chat_pages): 166 passed, 115 subtests passed, the body's count exactly. With test_intr_marks_memo, test_stage_marks, test_merge_tx_sets_light and test_lazy_body_readers added: 233 passed, 199 subtests passed.
- Hunk pinning by partial revert: dropping the files map from the index's build alone fails five of the six new tests (the atoms-only test passes); dropping it from the atoms-only restore alone fails only test_atoms_only_checkpoint_keeps_its_source_too.
- First item (line 7087): two probe tests derived from the new module's harness, both red at this head with "no file known", both green with the first-loop fix (14 passed with the inherited cases); the five modules above stay at 166 passed, 115 subtests passed with the fix applied.
- Second item (line 6661): the mutation passes 322 tests, 327 subtests across test_lazy_source_lifetime, test_asm_checkpoint, test_asm_index, test_asm_mat_lru_release, test_chat_pages, test_intr_marks_memo, test_lazy_body_readers, test_merge_tx_sets_light, test_stage_marks, test_asm_write_over_grown_entry, test_assembly_road_counters, test_card_summary_anchor, test_card_summary_anchor_e2e, test_rewind_memo and test_settled_cut_oracle; the nine served-page modules with test_restart_phases give 17 passed, 31 skipped. The proposed pin test: 2 passed at this head, 2 failed under the mutation.
- Third item (line 7088): with the per-session map replaced by a recording dict, the 37 modules that can mint lazy atoms give 683 passed, 471 subtests passed with zero reads of it; the two reds are ordering-dependent cases in test_assembly_road_counters that pass alone (31 passed, 102 subtests passed), with and without the instrument. With the write at line 6983 replaced by a no-op and the fallback at line 7088 by None, the same 37 modules stay as green (683 passed on one run; a repeat's one extra red was a render wait in test_chat_vsplit_served that passes alone). The proposed primed test passes at this head and fails under that mutation; the no-entry test passes under both.

lczh:
Taking this draft over. The first cut was written by another session that has since closed; the review above stands and every item in it is folded at e13ecc05, on top of a merge of main.

What the two new commits do. The first loop of `hydrate` reads the body once: a body that is no longer a `_LazyBody` counts as filled, with its bookkeeping finished from the memo when the entry still stands, so a peer thread finishing the same atom between the memo miss and the body read can no longer send a filled atom to the per-session map. Two tests pin the race, a memo double whose first miss performs the peer's put and fill, and two threads with the second held by events at the body read; both are red on the first cut with "no file known". The pin beside `_restore_prefix_atoms` is there: two documents sharing a fsid, the api one restored last, a `LazyIndex` over the web document with that fsid absent from its files map, hydration refusing with "no file known"; under the mutation that falls through the sentinel it reads the api copy instead. The legacy unbound road has its two tests: a document with no files key hydrates through the per-session map the restore primed, and the same with no entry is refused; the primed test goes red when the write is a no-op or the fallback is None. Smaller: the three copies of the files-map comprehension are one helper, the test module scrubs the shared event-model state in tearDown, the two module-level comments carry their dates, a script run of the module can import its harness, and the threaded test surfaces the worker's error instead of a timeout when the read is never reached, which is how it reads red on main.

Runs at e13ecc05, one module at a time: test_lazy_source_lifetime 11 passed (ten repeated runs); test_asm_checkpoint 91 passed, 85 subtests; test_asm_index 26 passed, 3 subtests; test_asm_mat_lru_release 9 passed; test_chat_pages 43 passed, 27 subtests; test_kernel 460 passed, 12 subtests. The assertIn pin scan against the merge base finds nothing lost. The body carries the review round and the red-first split for the six original regressions.

ksarma:
The evidence here was executed against e13ecc050, the PR's head, which merged as 1bcf0f89f by a plain merge; both changed files read the same at the merge commit, and every file and line reference is on e13ecc050. The two earlier comments (https://github.com/romp-on/romp/pull/1813#issuecomment-5740388253 and https://github.com/romp-on/romp/pull/1813#issuecomment-5740388324) reviewed b5be57efe: the race in the first loop of `hydrate` is closed at kernel/event_model.py:7163-7170 in the shape they asked for, both race tests are red with the first cut's kernel, the bound-body pin at tests/test_lazy_source_lifetime.py:201 is red under the mutation at 6740, and of the legacy-road tests at lines 215 and 229 the primed one is red when the write at 7062 is a no-op and when the fallback at 7177 is None, the no-entry one green under both. The author's comment (https://github.com/romp-on/romp/pull/1813#issuecomment-5751627745) gives the per-module counts, which reproduce one module at a time. What is left is one window the new branch opens, one shape change, and two test gaps.

- kernel/event_model.py:7169, the entry-gone arm. When the branch at 7164-7170 meets a body that is already a plain dict, it re-reads the memo under the lock and finishes the fill from the hit, or counts the atom filled when the entry is gone. The peer's fill in `_hydrate_one` sets the message at 7085, a kind-u atom's tool result at 7086-7087 and pops the marker at 7103, after its memo put at 7213; an eviction between that put and this thread's re-read makes the arm return while those last statements are in flight, so the caller sees the marker still present and, on a kind-u atom with a tool result, no `toolUseResult` until the peer's next statements run. The second loop's tolerance at 7187-7188 keys on the marker, popped last, so it never counted an in-flight atom; this is the residual of the shape the earlier comment asked for, not a departure from it.

  Executed with the cap patched to 0 and the peer paused inside `_hydrate_one` after its message set, a call from another thread returned 1 at once with the marker present, the tool result None and no read of its own; the marker was gone and the tool result set once the peer finished. Reaching it needs that eviction: with the default cap (the larger of 1 GiB and memory divided by 32) it takes a cap's worth of puts by other threads on other files inside a few dict writes, since the peer's stripe blocks same-file puts; `ROMP_HYDRATED_CAP_MB` set to 0 gives a cap of 0 through the knob reader at 859, so every put evicts itself, and a whole-MB value evicts every record over that size; outside the module that knob is set only in tests/test_asm_index.py. In the window kernel/kernel.py, which hydrates on the marker, reads the tool result at 39133 for diff rows and answers and finds it absent for that one build; `is_lazy` at kernel/event_model.py:4099 keys on the body type and says hydrated. The state is transient except on a shallow copy taken in the window, which the memo-hit arm heals unless the entry is evicted.

  Fixes, smallest first. Setting the kind-u tool result before the message in `_hydrate_one`, as kind k sets its skill text before its message at 7097-7098, leaves only the marker pop in flight and misleads no reader. Bounding the LRU loop at 7214 with `len(_HYDRATED) > 1`, the document memo's rule at 5593, stops only the peer's own put from evicting itself; executed, a put from a second file during the fill left the memo with that key alone and this thread returned with the marker present, so that bound alone reopens the window. A complete close runs `_hydrate_one` inside the memo-lock block at its call sites 7160, 7168, 7193 and 7217, not inside the function, since the lock at 5558 is a plain Lock and the test module's memo double calls the function under the caller's lock, which would deadlock; a plain dict seen at 7163 then means the pop completed before the re-read took the lock, at the cost of the fill (dict assignments, message normalization, a regex for kind o) inside the memo lock per atom. Treating a present marker with no entry as in flight needs a spin, and the arm cannot be gated on the marker alone, since a plain dict with the marker present and no entry gives the consumer nothing to finish from; without the lock shape, the arm's comment is the place for the residual. The pin for either shape is a cap-0 probe where the peer's message set lands and then waits, with the caller's call in a second thread, asserting the tool result present at return and, under the lock shape, the marker gone; a wait placed before the set proves nothing, since that thread sees a `_LazyBody`, takes the bound path, blocks on the stripe, and returns clean on every version.

## Smaller points

- kernel/event_model.py:7164: the class test `isinstance(msg, _LazyBody)` fails for a sentinel built before a re-execution of the module (kernel/loadsource.py:37-55 re-executes an existing name into the same module object, rebinding the class), so the branch counts the atom filled, reads nothing and leaves the marker and the sentinel body in place, and the caller's next body read raises "message read before hydration" for an atom the call was made on; the event model is re-executed only at import time (kernel/kernel.py:52 and kernel/judge.py:96) and inside tests that keep no lazy atom across the reload, so reachability in the product is nil, and the fix is `if not hasattr(msg, "source_path"):` at 7164, the shape the first cut tested, with the race tests passing unchanged and a None message counted filled as now.

## Tests

- kernel/event_model.py:7167: neither arm of the new branch is pinned, since both race tests reach it with the peer's fill complete and its entry standing, and disabling the re-read at 7167 (`if False:`) or breaking the entry-gone arm (`if True:`, so a gone entry is subscripted) each leaves the module at 11 passed. Two tests beside the race tests close that, both green at the head: a memo double whose first miss puts the record and sets a plain-dict message but leaves the marker, on a kind-u atom whose record's top-level tool result carries one of the keys in `TUR_CONSUMED_KEYS` at line 53 (otherwise the parsed atom carries no tool result and the assertion is vacuous), asserting the call returns 1, the marker is gone and the tool result equals the record's, red with the re-read disabled; and the evicted case, a first miss that fills the atom and puts nothing under a memo double, or `_HYDRATED_CAP` patched to 0 with two real threads so the complete call's put self-evicts, asserting the memo empty before the resume, either red with the entry-gone arm broken, a TypeError on None.

- tests/test_lazy_source_lifetime.py:225 and 235: the Validation bullet says the five review tests are red as described above, in the sentence after the base-red split for the six originals and naming no baseline of its own, but the two legacy-road tests' base red is an AttributeError on `source_path`, the slot this PR adds, not a hydration outcome (with the two assertIs lines removed both pass at the base), so the sentence is ambiguous rather than false, and a bullet giving the two faces, the race and borrow tests red at the base on hydration and the legacy tests red on the attribute with the two mutations as their behavioural pin, removes the ambiguity; optional.

PR #1817, opened 2026-09-18
The chat-build signature's row component drops a task row's lastTool and the row's raw ctxTokens, which no chat reader reads, so a progress event that only renamed a task's last tool no longer rebuilds a background tab, and a usage report's raw count no longer counts as a row change

lczh:
Tier: fix

## The defect

Every tab's chat payload is served from a cache while its signature holds, and the signature's `row` component was the whole liveness row minus two fields. Two fields no chat reader reads therefore moved the key of every background tab that carried them.

The first is each background-task row's `lastTool`, which the SDK backend rewrites on every task_progress event that names a tool and which no kernel function reads (the task normalizer, the task-row reader, the agent-liveness check and the feed's id signature read toolUseId, taskId, type, desc and since). A progress event that changed only lastTool does not wake the pusher (the backend pokes only when the task set changed or a subagent retired), so the byte-identical rebuild lands at the next pusher cycle: one rebuild per cycle in which lastTool moved. In practice an agent's tool call usually also grows a recorded output file in the same cycle, which moves the `taskout` component on its own, so the lastTool-only saving is the residue where a progress event lands in a cycle with no recorded output growth.

The second is the row's raw `ctxTokens`, and for it this change corrects the attribution only: the same context refresh writes the token count into the SDK registry as well, and the signature's `reg` component keys registry content by design (every field but the host offsets), so a refresh that moved only the raw count still rebuilds the tab after this change, now counted under `reg` alone. That registry field is a sibling defect with its own census and its own design decision in the registry signature, not closed here. The payload renders only the clamped percent and its overflow flag; the raw count's one reader is the compaction-suggestion tick, not a build. The context refresh is an asynchronous control request fired at turn end (also at model and effort switches and on connect), its `changed` flag covers the raw count, and its poke lands a cycle after the transcript's result write. So every turn end whose percent did not step produced one byte-identical background rebuild, and /perf's `builds.chat.bg_miss.row` counted it under `row` beside `reg` as if a rendered row fact had changed.

One defect, then: the row component keyed two fields no chat reader reads. In production the merged row never carries `snapT` or `interrupting` (the merge copies an explicit key list, and interrupting is a feed-entry field), so the two existing exclusions were inert outside hand-built test rows and these two fields were the operative movers.

The gain is an estimate, not a bound. Live /perf at the last sample of this boot: bg_built 7615, row 2708, bg 2362, live 2876, transcript 1950, postal 2360, taskout 499. Every SDK task start or end moves bg and row together, so if every bg move was also a row move the row-without-bg builds were about 350, under five percent of background builds; bg has movers of its own, and the aggregate counters cannot separate row-only builds (a count of the perf log's chat-build lines whose miss is exactly row would). The rebuilds this change removes are the lastTool residue: a progress event landing in a cycle with no recorded output growth. The certain gain is attribution: after this change `row` in the bg_miss counters means a rendered row fact changed, so the next diagnosis of the chat push reads the counters straight.

## The fix

Two module constants directly above the signature name the row fields the chat build never reads: `snapT`, `interrupting` and `ctxTokens` on the row, and `lastTool` on each background-task row. A small helper, `_chat_row_sig`, projects the row through them: the row's items without the first set, each task row reduced to its sorted items without the second (a tuple, compared by value as the dict was). The signature's row append calls the helper; its comment says what is dropped and why. Nothing else changes: no new label, no new /perf key, no census-table change (the task SET still gates the transcript scan, so the census notes for the task-row and liveness-map readers under `row` stay accurate). The feed's own memo key folds the same whole row with its own literal and has its own census; that is the same defect in a different memo and a sibling change, not folded in here.

The constants are a denylist rather than an allowlist of read fields on purpose: an unread field left in costs one byte-identical rebuild per cycle it moved in, which the /perf counters show; a read field left out of an allowlist would serve a stale payload silently, which nothing shows. The census test pins that the two dropped fields stay unread.

## Tests

Red first, in the chat signature's input census module. A new differential test hands the signature a row carrying one background task and a raw token count, then moves one field at a time: a task_progress that rewrote lastTool must move nothing, and a usage report that moved the raw count and not the percent must move nothing. On the stock kernel both moves report `('row',)`, and the test fails at the first with "a task_progress that rewrote lastTool: nothing the build reads changed" (the refuter's standalone probe, run against the stock clone, shows the same for both: two red, one green). The same test pins what still moves and under which labels: a second task started, the task ended, its description or type landed (bg and row); the percent stepped, the overflow flag flipped, a subagent started, the state changed (row); and the snapshot stamp still moves nothing. A new census guard pins both constant sets by value, checks by word that none of the twelve named chat readers mention either dropped field, and, as the transitive backstop a reader added inside a helper they call would slip past, scans the whole kernel source: every mention of lastTool lies in the projection block or the key itself, and every mention of ctxTokens there, in the compaction-suggestion tick or in the merge that writes it. On the stock kernel the guard errors on the missing constant.

After the change the census module is green: 61 passed, 10 subtests (59 stock tests plus the two new ones). The other modules that pin the signature, the row slot, the pusher's source lines, the feed's census and the /perf key set were run one at a time:

- tests/test_stage0_log_readers.py: 36 passed (state, subagents, retryCount, modelPending, connected, spawning, model and auth each still move row; snapT alone moves nothing)
- tests/test_chat_fixed_cost_memos.py: 40 passed (no row and an empty map keep the (None, False) slot; a row handed over misses under row alone)
- tests/test_chat_delta_resync.py: 12 passed (a context-percent step still busts the chat build signature)
- tests/test_perf_stats.py: 62 passed (the bg_miss keys are the label tuple plus cold and nosig; no new key)
- tests/test_kernel_rewind.py: 43 passed and tests/test_cold_tab_gate.py: 22 passed, 4 subtests (the pinned source lines in the signature and the push are untouched)
- tests/test_kernel_timeline_split.py: 27 passed (the view signature's own lastTool exclusion, the precedent)
- tests/test_feed_memo_inputs.py: 15 passed (the feed's census, untouched)
- tests/test_kernel_bg_tasks.py: 48 passed (task-row fixtures carrying lastTool)
- tests/test_compact_suggest.py: 22 passed (the raw count's one reader)
- tests/test_awaiting_rows.py: 28 passed, tests/test_kernel_awaiting_merge.py: 6 passed, tests/test_sessions_listing.py: 10 passed (fixtures naming the two fields)
- tests/test_kernel.py: 460 passed, 12 subtests

The static pin scan against main found 0 lost literals (8310 checked). The extension suite: 5001 tests, 4995 passed, 1 failed, and the one failure is the KaTeX layout pixel probe that fails on main in this environment; no extension test pins the chat signature, its label tuple or the row fields.


ksarma:
A post-merge note; evidence executed against a650504d8, main after this and its three sibling PRs merged. The four PRs' key narrowing was checked for staleness through the real build functions and held: a row differing only in a dropped field leaves the signature unchanged, and a field a card reads still moves it.

One finding: tests/test_chat_build_sig_inputs.py:409-421 permits no word-regex match for `lastTool` in raw source outside the projection and the key, so a mention counts as a read. A comment naming the field, inserted before kernel/kernel.py:5028, fails at line 416 with four mentions against three permitted and a message naming a reader; a real read fails with the same message. `ctxTokens` (line 418) behaves alike; both are live row fields, so this recurs: #1817 and #1819 were green alone and red together; #1825 reworded two kernel docstrings, not the test. Any comment or string naming either field, case-sensitive and whole-token, elsewhere in kernel/kernel.py fails the test. Counting read shapes instead fixes it: get calls, subscripts and membership tests with a constant key, on any receiver (the shapes `_row_fields` at tests/test_feed_memo_inputs.py:448 derives, without its receiver binding), count zero for prose and one per real read. Stripping string literals would not: the key of a real read is itself a string literal, so the test would go blind to the reader it hunts.

PR #1820, opened 2026-09-18
The feed re-reads a warm session's parse in place when its files moved, instead of deriving it cold and warm again a build later

lczh:
Tier: fix

## The defect

The feed's per-session card memo keys each living session on every input its derivation reads, and one of those inputs is the transcript parse. The feed reads that parse cache-only, on purpose: a cold kernel paints its cards at once and lets the dots and anchors fill in when someone else parses. But the same cache-only read runs on every build, and a streaming session moves its transcript between the chat's build and the feed's every cycle. On that build the read misses, the key's parse component falls to "no parse", and the session is derived cold: no working dot, no open-turn narration, no deep-link anchors, its state reads unknown, its background-task set reads empty and the closer's swirl is off. On the next build, after the chat has parsed the same file, the same session is derived warm again. Two derivations and a visible blink per append, and a card move on no new information, which the design rules forbid. The same flip happens when the states log moves with the transcript still.

Live figures on the stock kernel at about 5074 s uptime: 1195 feed builds and 4087 derivations; the miss labels read parse 2061 against transcript 1297, with bg 1488 and closer 288 riding the same flip, and cold 23. Parse flips explained by neither a transcript move nor the boot's own cold-to-warm come to roughly 2061 - 1297 - 23 = 741 warm re-derivations of sessions derived cold the build before, about 0.6 per build and about 18 percent of all derivations, on top of their cold twins being wrong rather than merely wasted. A later read at 1763 builds (parse 2963, transcript 1861, cold 23) gives the same rate.

## The fix

The key still reads the parse cache-only, and when that read misses for a session whose memoized entry was derived over a WARM parse, it now re-reads through the kernel's parse in place instead of deriving the session cold. The gate is the previous key's own parse component, handed into the key through the build context by the loop that already hands over the previous entry, so the key builder's signature is unchanged. A session never derived warm still reads cache-only, so a cold kernel's first paint parses nothing, as before.

The parse is the shared store's, the same slot the chat's build hits next cycle, so a tab the chat builds every cycle pays the parse once either way, in the feed's build instead of the chat's. A session the chat skips pays one assembly fold per appended version, on whichever thread builds the feed. The parse bit never flips back to False for a warm entry, and the background-task and closer components stop flipping with it.

Two counters join the memo's block on GET /perf under builds.feed.memo: coldLive, every living session whose cache-only read missed, per session per build (a session nothing has parsed rides it every build by design, so a standing count is not a fault), and coldFlip, those the memo held warm and the key re-read in place, which is exactly the count of in-place parses and also rides parses.kernel. A coldFlip climbing every build for one session with no appends would mean its parse never stores (a row still naming a leaf discover retired); the counter makes that visible.

The perf bench (tools/perf-bench.py) tripped on the same mechanism in CI, and its cold rows change with this PR: its steady-state feed row memoizes every entry under a warm key, and its build_feed_noparse row then emptied only the parse store before each sample, so the kernel now re-reads every entry in place there, derives nothing cold and asks for no background warm, which the bench's tripwire read as a lost cold branch. A freshly started kernel has no feed memo, so the bench's cold-kernel cache set now includes it: the cold rows and the noparse row drop the memo through the kernel's own forget (its counters stay right) beside the parse store, the report names it among the emptied caches, and the tripwire holds again for the reason it was written. Two consequences worth knowing when comparing bench runs across this change: build_feed_noparse now measures a cold derivation per sample (before, its kept samples were memo hits, since the first cold sample had memoized cold keys that the next samples matched), and push_cold_cycle's feed is built over an empty memo, as a restart's first cycle is. The kernel's warm request is untouched: a cold kernel's first paint still asks for the background warm for every warm-wanted session, and a session the memo never held warm still does on every build.

One completeness caveat the change introduces, written beside the re-read: a parse whose content changed while the transcript's identity, both states identities, the cut, the registry and the last turn's end all stand now hits, where the cold-and-warm pair used to re-derive it. The only parse input outside the key is the anchor candidate transcript, which the store treats as immutable after the fork; were that contract to break, the anchor's identity belongs in the transcript component.

## Tests

Red before, in tests/test_feed_session_memo.py, class AWarmEntryIsNeverDerivedCold over the module's hermetic three-session board: the chat's parse warms one session, its transcript then grows by an ended exchange with no parse between, and the next build derives it. On the stock kernel that build's working card reads sessState unknown where quiet is asserted (the exact red line: 'unknown' != 'quiet'), the stored key's parse component reads cold, the miss attribution carries bg and closer beside transcript and parse, and the kernel parse count does not move; the build after the chat's next parse derives the session once more where none is asserted. A third test, added after CI's bench failure, pins the pair the bench tripped on with a recorder in place of the background warm: a cold kernel's first paint asks once; with every session parsed and every entry memoized warm, emptying the parse store alone re-reads all three in place (coldFlip rises by three, the working card stays quiet) and asks for nothing more; emptying the memo with it derives all three cold and asks again. The second test pins the gate the fix must not widen: a cold kernel's first paint parses nothing and counts three cold reads and no flip; on stock it raises KeyError on coldLive, since the counter does not exist there. The two exact key-set pins on the memo's /perf block (tests/test_perf_stats.py and tests/test_feed_session_memo.py) are widened with the two counters; both are red on stock for exactly those two names, the module's own pin only after a review round (2026-09-18) guarded the module's memo reset so it zeroes the keys the kernel defines instead of planting them: the first stock run wrote the two counters into the stats dict from the test's own setUp, the kernel's report copied them out, and the widened pin passed on a kernel without them, so that run's red was the two new tests alone. tests/test_kernel_feed_warm.py gains two source pins on the gate and the in-place parse, red on stock as not found.

Green after, one module at a time: test_feed_session_memo 47 passed, test_perf_stats 62, test_perf_bench 45 (the bench run end to end against this checkout, its warm-request tripwire and its cold-cache pins included), test_kernel_feed_warm 4, test_feed_memo_inputs 15 (the census's label list and CTX table unchanged; the ps note reworded), test_chat_build_sig_inputs 59, test_boot_parse_gating 28, test_shared_parse 17, test_kernel_rewind 43, test_kernel_working_narration 5, test_feed_first_cold_push 7, test_kernel 460. tests/test_boot_parses_served.py 2 passed on a real hermetic kernel (its served lab needs the extension's node_modules to bundle; it skips without them): the boot with no client still parses nothing, since with no client the only feed build is the /feed.json GET and no warm-keyed entry exists before it, and the lab's key files never move after boot, so no warm entry ever misses; its bound on the tabs' parses gains a note saying so. The static pin scan over the branch lost none of the 8311 assertIn literals it checked. The extension suite: 5001 tests, 4995 passed, 5 skipped (the Firefox Playwright legs, no such browser here), 1 failed (the pre-existing KaTeX pixel probe that fails on main in this environment); not re-run after the review round, whose edits touch tests only.

The branch carries a merge of main after the narrowed feed row key (#1819) landed ahead of it: three hunks, each two additions to one literal, each resolved as the union (the memo's counters carry coldLive and coldFlip beside row_by; both exact key-set pins name all three). On the merged tree, one module at a time: test_feed_session_memo 52 passed (main's five row-key tests included), test_feed_memo_inputs 23, test_perf_stats 62, test_kernel_feed_warm 4, test_feed_first_cold_push 7, test_perf_bench 45, test_kernel 460; the pin scan lost none of 8315 literals.

Measurement to run live before calling the flap ended: parses.kernel per cycle should rise by no more than coldFlip predicts; miss_by parse should fall toward transcript plus the boot flips, with bg and closer falling toward zero; push.feed lengthens by the parses push.chat no longer pays.


ksarma:
A post-merge note, with evidence executed against a650504d8, main after this and its sibling feed PRs merged. The read paths of this PR and its two siblings were checked by execution for stale or missing rows (a shared read-only view missing a write, a postal-log reduction skipping or double-counting a row, a warm re-read over moved files) and held.

- A warm memo entry whose transcript stat fails re-parses every build: `parse_cached` returns None (kernel/judge.py:3174), the re-read at kernel/kernel.py:41040 parses the missing leaf under a None key and stores nothing (kernel/judge.py:3928), so the next build misses again. In tests/test_feed_session_memo.py's board, one session warmed, its transcript removed: build 1 derives it, builds 2 to 6 hit, yet each build moves parses.kernel +1, coldFlip +1, coldLive +3, and the card reads `sessState` quiet where the PR's base read unknown and parsed nothing. Renamed and chmod 0 match; only an already-warm entry is affected, while its transcript stays unresolvable. A transcript the kernel cannot read is thus painted as a settled quiet session, where the repository's rule on authoritative sources asks for the failure to surface rather than a value derived from nothing. Fix: gate the re-read at 41040 on `os.stat(path)` succeeding, counting coldFlip only when it runs; all three cases match the base, four test modules passing (151).

- This PR's two new GET /perf counters (kernel/kernel.py:40480) have no line in docs/reference.md, whose builds.feed.memo paragraph (2325-2369) names nine of thirteen keys (failed and failing also absent), so a by-design standing `coldLive` count reads as a fault. Fix: one sentence after the row_by sentence (mid-line 2355) naming all four, plus a pin like tests/test_perf_stats.py:1204-1208 that every `_feed_memo_report()` key is backticked there.

PR #1826, opened 2026-09-18
Two feed docstrings name a task row's last tool in words, so the chat census that counts the field's mentions holds again

henrypinkard:
Tier: fix

Main went red on `tests/test_chat_build_sig_inputs.py::Census::test_the_row_projection_drops_only_fields_no_chat_reader_reads` once #1819 landed beside #1817: the census from #1817 counts every mention of the task row's field name in `kernel/kernel.py` against the mentions inside the chat key block, and #1819's two feed docstrings named the field in prose, so the whole-file count read 5 against the block's 3. Each PR was green on its own CI merge; the two together broke the count.

**Fix:** the two feed sentences say "last tool" in words. No code changes; tests unchanged (the census is the test that fails before and passes after).


henrypinkard:
Superseded by #1825, the author's own fix for the same census red; closing.

PR #1833, opened 2026-09-18
The served drag test re-aims at the drop zone's current place before releasing, so the timeline band's late fit cannot move the zone out from under the pointer, and each wait names itself

lczh:
Tier: fix

## The defect

The vertical-split drag lab (`VSplitDrag` in `tests/test_chat_vsplit_served.py`) went red on CI twice more after #1824 made the drag start on the runner: all four tests died with a bare "page.waitForFunction: Timeout 20000ms exceeded", and passed on rerun. This is the second half of that fix. The first half (the atomic in-page box read) got the drag started; this one gets the drop to land.

Which wait died is in the failing lines themselves: each carries the pane, the bottom zone and the ghost, and no `afterDrop`. The only 20 s wait between the ghost read and the `afterDrop` write is the store wait right after the release, for the `place:'below'` entry in `romp-chat-cols`, and the module's slowest-test durations (21.7 s and 32.1 s for the first test) say the 20 s elapsed. That store write is synchronous inside the shell's `moveTab`, reached only from the bottom zone's drop listener, so the wait was already keyed on the exact event: the drop never reached the zone.

The mechanism, in the failing lines' own numbers: `bottomZone.paneHeight` 533 against `ghost.paneHeight` 686 in every one, the ghost's top at 267, half of 533. Between measuring the zone and reading the ghost the chat pane grew by 153 px. The shell's timeline band auto-fits its content height (a load listener and a `ResizeObserver` on the timeline iframe write the band's height variable), from its first paint of 250 px to its settled 97 px, about two seconds after the browser opens, and the drag runs at about that moment; on the runner the refit fell inside the drag. The bottom zone is absolutely positioned and pinned to the pane's bottom, so it moved from 426..533 to 579..686 while the pointer stayed at the zone's old centre. Under Playwright's intercepted drags a `dragover` fires only on a pointer move and the `drop` is dispatched at the last pointer position, with no `dragover` or `dragleave` in between, so the ghost stayed on over its stale rectangle (one run) or, when the refit landed before the pointer arrived, never showed at all (the other run, its 10 s ghost wait swallowed as designed). Either way the release landed on the pane's iframe, which has no drop listener; `moveTab` never ran, the page's `dragend` unmounted the zones, and nothing wrote the store. Not the fixed sleep, not a kernel round trip, not a busy frame.

## The fix

Test-only, two changes to the pointer driver.

First, the re-aim. After the ghost wait and before the ghost is read, the driver measures the bottom zone again in one in-page step. If the pointer is outside the zone's current rectangle, it moves onto the current centre (two steps) and waits for the ghost again; the `dragover` that move produces re-cues the ghost against the current pane. The release point is kept in a variable; the ghost is read after the re-aim, so it and the pane it is compared with in the second test come from the same layout; and the release happens where the pointer actually is.

Second, a small `named(label, promise)` helper wraps the three 20 s waits (the zone's mount, the store write, the bottom pane's iframe mount) and rethrows with the label. The store wait's label carries the release point and the class of the element under it, so the next failure says where it released and onto what. The 20 s bounds are unchanged: where the drop has not happened, no bound helps. What remains is the instant between the re-aim move and the release, and a failure there is now attributable rather than a bare timeout. A dated comment in the driver carries the why. What the four tests prove is untouched: the real pointer press, the page's own `dragstart`, the shell's zones, the drop and the bottom pane's fill to turn 0.

## Tests

The red-first demonstration is a harness lever kept out of the repo. It runs the lab with either the old driver text or the new one and forces the runner's layout change deterministically: it hooks the shell's writes of the timeline band's height variable, holds the band at its first fit (250 px, the chat pane 533 px) through the drag start and the zone measurement, then releases it to its settled fit (97 px, the pane 686 px) between the ghost wait and the ghost read, where the runner's numbers put it. Against the old driver it produces the CI failure in its exact shape: the zone measured against a 533 px pane, the ghost read against 686 with its stale top at 267, the store wait out at 20 s, four tests red. Against the new driver the same lever passes four of four: the re-measure finds the pointer outside the zone's current rectangle, the pointer moves onto its centre, the ghost re-cues at top 343 against the 686 px pane, the `place:'below'` entry is written and the bottom pane fills to turn 0. The lever stays out of the repo because it intercepts the shell's own style writes: it is a test of the test, not of romp, and would pin a private implementation detail of the shell's layout; the re-aim with its dated comment is the pin.

Locally, `tests/test_chat_vsplit_served.py` passes 12 of 12 with chromium (`ROMP_SERVED_TESTS_ENGINES=chromium`, 32 s). Nothing outside the module pins the replaced driver text (a grep of `tests/`, `vscode-extension/`, `scripts/` and `docs/` for the old wait lines and the driver's name finds only the module itself), and the static pin scan reports 0 lost.


ksarma:
The evidence was executed against df5356224. The current head 03134a080 is that commit with main merged into the branch. Main's copy of `tests/test_chat_vsplit_served.py` has not changed since #1832 merged, so the merge brought only #1832's edits to this module: the stepped drag-start loop with its own "no drop zone" message, the event wait for `.chat-sub iframe` in place of the 500 ms sleep after the store wait, and the removal of the 600 ms sleep after the iframe-mount wait.

Against that, the change's own added lines differ from the reviewed ones in two places, both where #1832's edits meet the change: the `named()` wrap on the zone-mount wait is gone, and the iframe-mount line no longer carries `waitForTimeout(600)`. The helper, the re-aim block, the `under` read, the named store wait and the named iframe-mount wait are the reviewed text. Each point below is stated for 03134a080.

## At this head

The one finding at the reviewed head was the conflict with #1832. At 03134a080 it is resolved, and by the same text the evidence executed.

- `tests/test_chat_vsplit_served.py` lines 226, 256 to 257 and 265 on df5356224. #1832 merged first (e175242e2) and rewrote the same three regions, so the PR reported CONFLICTING against main. Executed against df5356224, this reproduces: `git merge-file` over the head, base and main copies exits 3 with three hunks, at merged lines 229, 288 and 306, and `git merge-tree --write-tree` reports the same three. They are the `named()` zone-mount wait against #1832's stepped drag-start loop; the named store wait plus `waitForTimeout(500)` against #1832's `.chat-sub iframe` wait; and the named iframe-mount wait plus `waitForTimeout(600)` against #1832 with that sleep removed. Nothing else in the change conflicted: the helper, the re-aim block and the `under` read merge clean.
- At 03134a080 GitHub reports MERGEABLE, and the module is byte-identical to the resolution that was executed: #1832's drag-start block with no wrap on the zone-mount wait (its try/catch already throws "no drop zone"), the named store wait and the `under` read followed by #1832's iframe wait with no 500 ms sleep, and `named("the bottom pane's iframe mount", ...)` with no 600 ms sleep. Neither wrong resolution is present: a keep-both take at the drag-start hunk would have put two labels on one wait, and a wholesale take of the branch side there would have dropped #1832's stepped drag start and its message.

## Also at this head

PR #1882 edits the same pointer driver in this module (the zone read, the ghost wait and the release, lines 247 to 254 on main), so whichever of the two lands second re-reads the other.

## Smaller points

The helper's comment and two lines of the PR body still describe the reviewed head rather than the merged text.

- `tests/test_chat_vsplit_served.py` line 197 on 03134a080 (193 on df5356224): the helper's comment says the label names "the zone's mount, the store write, the bottom pane's iframe mount". At this head the zone-mount wait is not wrapped; the wrapped waits are the store write, the iframe mount and the 10 s ghost wait after the re-aim. Fix: name the three that are wrapped.
- The PR body says the helper "wraps the three 20 s waits (the zone's mount, ...)" and that the module "passes 12 of 12 with chromium" in 32 s. Both predate the merge: the count is now two 20 s waits plus the 10 s re-aim wait, and the 12 of 12 line could say which head it ran on.

## Tests

`tests/test_chat_vsplit_served.py` with `ROMP_SERVED_TESTS_ENGINES=chromium`: 12 of 12 at df5356224, and 12 of 12 on the merged text (the file at 03134a080). The red-first claim was reproduced by execution with a layout lever that moves the pane between the zone measurement and the ghost read: `VSplitDrag` failed 4 of 4 on the base's driver text (zone measured against the pre-growth pane, ghost read against the grown one, store wait out at 20 s), and passed 4 of 4 on the reviewed text and 4 of 4 on the merged text. The runs, their numbers and the mechanism trace in `kernel/kernel.py` follow in a second comment.

ksarma:
The test record behind the comment above, executed against df5356224 and against the merged text (the file at 03134a080).

- `tests/test_chat_vsplit_served.py` at df5356224, `ROMP_SERVED_TESTS_ENGINES=chromium`: 12 passed in 24.5 s (`VSplitLocal` 8, `VSplitDrag` 4). The same module on the merged text, over an archive of main at e175242e2: 12 passed in 21.1 s.
- The red-first claim, reproduced by execution. A lever intercepts `setProperty` for `--tl` in the top frame, defers every write, and releases the settled write (under 150 px) between the ghost wait and the ghost read. On the base's driver text, `VSplitDrag` failed 4 of 4 in 24.4 s: `TimeoutError` on `page.waitForFunction` at 20000 ms, `bottomZone.paneHeight` 583 against `ghost.paneHeight` 686 with the stale ghost top at 292, no `afterDrop`. On the reviewed driver text, 4 passed in 7.3 s: the re-aim fired (pointer y 524.7 to 627.7, the zone now 569 to 686), the ghost re-cued at top 343 against 686, `place:'below'` written, the bottom pane filled to turn 0. On the merged text, 4 passed in 4.6 s with the same numbers (pane 583 to 686, ghost top 343), the re-aim firing after #1832's stepped drag start.
- The local geometry is 583 to 686 where the runner's was 533 to 686. A first lever variant that released a 250 px write instead (pane 583 to 533, the pointer still inside the moved zone) left only `test_2` red on the old driver; that is a different geometry from the runner's, not a counterexample to the mechanism.
- Mechanism, traced in `kernel/kernel.py`: `.col-drop` is absolute with inset 0, and `.col-drop-bottom` has top auto with an inline height set once at mount, so the zone moves with the pane and keeps its height; `showGhost` positions the ghost synchronously inside the zone's dragover handler against the current parent rect, so a ghost read after the re-aim move sees the current layout; autosize writes `--tl` from the timeline iframe's load listener and a `ResizeObserver`, with a CSS default of 200 px; the bottom zone's drop listener is the drag's only path to the `moveTab` branch that writes the `place:'below'` entry, and the keyboard split entry point, the other caller of that branch, does not run in this test.
- The re-aim's branches: `rel` inside `zr` means no move and a release at the measured centre, the old behaviour; `zr` null falls through to the labelled store wait; the re-aim's ghost wait is not swallowed where the first one is, and since the dragover that reaches the zone is the same handler that turns the ghost on, a ghost-less re-aim is a drop that would miss anyway, now named instead of a bare 20 s timeout. `named()` keeps 200 characters of the original error inside the 500-character `died` slice; `under` reads the class name, id or tag of the top-document element at the release point, the zone itself at its centre.
- Regression: a grep of `tests/`, the source and webview directories under `vscode-extension/`, `scripts/` and `docs/` for the driver's name, the old wait lines and the 20000 ms message finds only this module; the `col-drop-bottom` mentions in `test_chat_split.py` and `test_chat_split_served.py` pin the shell's CSS and class names, which this change does not touch; `VSplitLocal` is untouched and its 8 tests pass.

lczh:
Superseded by #1882, which lands the same diagnosis (the bottom drop zone rides down as the pane grows mid-drag) with a deterministic growth step and a red stale-drop case. Closing this one.

PR #1843, opened 2026-09-18
Timeline: a relayed message ends at the recipient's lane, named with its host; the two kernels' copies fold into one connector

henrypinkard:
Tier: fix

## What the user saw (2026-09-18, three screenshots)

A local session and a remote session of the same name (spelled here as the lab spells them: `web` and `TESTHOST:web`) on the merged timeline: a message marker on the local lane read "web → peer:TESTHOST" (the HOST named as the recipient) with no connector to the recipient's lane, a hook (an outgoing hidden-counterpart stub) dipping off the local lane, and the remote twin's lane reported empty with a stub at its far right.

## What was wrong

- **The sender's kernel addressed the relay, not the recipient.** A relayed send's row in `timeline/messages.jsonl` carries `to_id: "peer:<host>"` and, since 2026-09-08, the recipient's stable id (`to_sid`) and display name (`toName`, `<host>:<name>`). `_postal_messages` kept the relay address as the connector's far end and looked the display name up by it (empty), so the tooltip fell back to the raw id ("peer:<host>"), the merge could match no lane, and the view drew the hook.
- **The recipient's kernel drew a second copy that named the recipient bare.** The delivered copy lives under its own delivery mid on the recipient's bus. The merge stitched its foreign end onto the local lane and drew a connector, but the tooltip read "web → web": the merge filled a display name only when one was missing, and a kernel names its own sessions without a host.
- **Once the sender's row ends at the recipient's lane, both copies would draw.** Two connectors for one message unless the merge knows they are one.

## The fix

- `kernel/kernel.py` `_postal_messages`: a relayed row's far end is `to_sid`, its display name `toName`; the relay address keeps one job, the pending flag's cross-host leg (the far end's liveness is not knowable locally). A delivered copy hands the sender's id on as `originMid`. Rows older than the fields keep the relay address.
- `ui/webview/federation.ts` `stitchMessages`: the two kernels' rows for one message are keyed under the sender's id, by the read receipt's `dmid` or, before the receipt lands, by the delivered copy's `originMid`; an upgrade from the recipient's copy keeps the sender's ids and ends the pending state. A remote lane's endpoints take the lane's display name (`TESTHOST:web`) as the board labels the lane; a thread-anchored end keeps the thread's name; local lanes keep the kernel's own names.
- `ui/romp-timeline-view.js`: the hidden-counterpart stub path carries `data-tl-stub` (the lane-less end) for the served pins.

## Proof

**Lab** `tests/test_timeline_same_name_served.py`: two hermetic kernels, a hub owning `web` and a checked-in `TESTHOST` owning `web` and `api` (the notes-api world), with the rows both buses write for one relayed send from the hub's `web` to `TESTHOST`'s `web` (the hub's sent row with `to_sid`/`toName` and its receipt's exec row with `dmid`; `TESTHOST`'s delivered copy with `originMid` and its own exec). Chromium on the hub's `/timeline`; merged frames captured through the manager's frame registration; lanes, bars and connector paths read from the SVG.

- The remote twin's lane draws its own five bars, the local twin its own six, `TESTHOST:api` its four; merged turns keyed by prefixed sid. **The report's empty remote lane did not reproduce** with two kernels at one version; the counts are pinned as the guard (green at the base too, stated as such).
- One row for the message on the merged board (the two copies folded), from the local twin to the lane `TESTHOST:<sid>`, named "web" → "TESTHOST:web"; exactly one connector path between those lanes; no stub.
- The reverse leg (right by construction, proven): the pane's own compact and command hooks addressed to `TESTHOST:web` park the ops on `TESTHOST`'s kernel alone (`[["compact"], ["effort", "high"]]` under its `web`'s sid), the control compact to the bare `web` parks the hub's one op, nothing refused on either side.

**Pins**: `tests/test_timeline_pending.py` class `RelayedRowNamesTheRecipient` (the far end and its host-prefixed name, pending honest on the relay leg until the receipt, the delivered copy's `originMid`, an older row keeps the relay address); `ui/webview/multi-kernel-merge.test.ts` (dedupe by `dmid` and by `originMid`, remote lane naming, thread-anchored ends, local lanes untouched, the upgrade's ids and pending).

**Red first, executed at the base 713dc9ef** by copy-aside of the three product files (the head's tests):
- `tests/test_timeline_pending.py` `RelayedRowNamesTheRecipient`: 2 failed, 1 passed (the older-row guard, green at both ends by design). The far-end test: `('peer:TESTHOST', '') != ('<recipient sid>', 'TESTHOST:web')`; the delivered-copy test: `originMid` `None != 'm1'`.
- `ui/webview/multi-kernel-merge.test.ts`: 2 fail, 57 pass. The dedupe test: `expected 1, actual 2` (two connectors for one message); the naming test: bare names where the board labels `TESTHOST:web` / `TESTHOST:api`.
- `tests/test_timeline_same_name_served.py` under `ROMP_SERVED_TESTS_REQUIRE=1`: 1 failed, 3 passed. The connector test: `2 != 1 : the two kernels' copies of the one message are ONE row on the merged board` (the hub's row ending at `peer:TESTHOST` with an empty name beside the recipient's copy). The three that pass at the base are guards and the reverse leg: the lanes and frames arrive, each lane draws its own bars (6 / 5 / 4), the compact and command park on the remote kernel alone.

**At the head**:
- `tests/test_timeline_pending.py`: 11 passed. `ui/webview/multi-kernel-merge.test.ts`: 59 pass, 0 fail. The lab under REQUIRE (chromium): 4 passed.
- `tsc --noEmit`: exit 0.
- The whole extension suite (CI's glob, `out-tests/**/*.test.js`): 5033 tests, 5028 pass, 0 fail, 5 skipped.
- The whole Python suite (`uv run --python 3.13`): 14180 passed, 59 skipped, 2329 subtests passed, 0 failed (33 min).


ksarma:
A note after the merge: evidence executed against 821143d8a, the merged head.

- ui/webview/federation.ts:758. The recipient-copy upgrade runs `Object.assign` over the sender's row, so sent and summary become the copy's; the copy stamps sent at delivery (postal/postal_service.py:746). Executed: sent 36000, delivered 39600; merged sent is 39600 until the receipt's exec row lands (postal/postal_service.py:4281), then 36000; exec 39610 in both. On the board a 10:00 message jumps to 11:00 when read, then back; the jump needs the sender local (ui/webview/federation.ts:930 orders local first; a local recipient leaves the origin at delivery time), the peer down at send time (else the gap is seconds), the copy read, and its receipt unapplied (seconds; longer on a stalled return leg; permanent if never applied). Only the recipient's kernel captions relayed mail (the caption join, kernel/kernel.py:44682, keys on the marker's delivery mid, postal/postal_service.py:1113): a caption shows only in the lag. Fix: keep the sender's sent; in both branches an empty summary takes the copy's; m4 and m6 (ui/webview/multi-kernel-merge.test.ts:451) use equal sent, no summary.

- ui/webview/federation.ts:750. The name rule re-labels only remote-lane ends, so a remote row to a local lane keeps its bus's host-prefixed `toName` (postal/postal_service.py:2969). Executed alone: to stays "HUB:api", so the tooltip (ui/romp-timeline-view.js:6633) reads "TESTHOST:web → HUB:api" under a lane labelled api. The row is alone while the local kernel lacks a copy: a directed peer's mail is quarantined until approved (postal/postal_service.py:4684; default tier, kernel/kernel.py:21123); isolated mail is dropped, refused mail re-relayed; a trusted peer's copy lands one exchange later and wins the key. Also executed: a remote row carrying the receipt's exec upgrades an unread local copy to "HUB:api" until the local kernel's next frame. Fix: also re-label when the row's original id has a host prefix; local ids are bare, so m5 (ui/webview/multi-kernel-merge.test.ts:474) stands.

PR #1859, opened 2026-09-19
Settings across machines, phase one A: proposals, pins and the broadcast's origin

henrypinkard:
Settings across machines, phase one A (plans/settings-across-machines.md): a remote machine's newer setting is a PROPOSAL the user answers, never a silent apply; a per-machine pin; the gear's second line under the row; a sync notice as the interim alert. The manager's dispatch of 2026-09-18 on the user's ask that this move now; the first commit records the split into one A (this head: every kernel-side rule and the gear surface) and one B (the owner-less notice card with its three actions, after pull request 1831 and the notice action-kind API land). Cut from main at e7e68f0c. Feature tier, unarmed.

**Premise, checked in code.** The four mesh-synchronized booleans (Auto Nudge, Suggest /compact, file editing, task tracking) converged by the newest gesture stamp: the tunnel supervisor's poll handed a peer's `/version` to `_adopt_peer_settings`, which applied any value under a stamp newer than the local store's through the setting's own gt-gated setter, and a hub's push over `/mesh-settings` applied on the receiver the same way. That is how the user's task tracking turned off when a new machine attached with it off (2026-09-18). Two checks the manager asked for before building: the pin gate sits BEFORE the stale check in each setter and applies only to a gesture whose origin is not this machine's own dashboard (pinned at the source by a test); a proposal never writes the local store (the store's bytes are compared before and after in a test), the one exception being the pre-existing stamp-only lift for a SAME value under a newer stamp, which the plan keeps.

**The change.** kernel/kernel.py: the adopter is gone; `_propose_peer_settings` is the poll's inbound leg: a newer peer stamp with a DIFFERENT value writes or refreshes a record in `settings-proposals.json` (host, value, the peer's stamp, `seen`, the local value at the time) and applies nothing; the same value lifts the stamp only; a pinned store raises none; a stamp the user kept against (`answered`) is not proposed again; the record drops when the peer's current news agrees with us or is older, or we hold the proposed value. `/mesh-settings` proposes on the receiving kernel through the same function (the push body carries the hub's host name for the record) and answers the snapshot with the pins and the proposals; a peer that reports a store pinned is not pushed our value for it. The pin store `settings-pins.json` with `_set_setting_pin` (gt-gated, per machine, set by the local dashboard's `setSettingPin` message or a proposal's Pin answer; never a KERNEL_SETTING); `/version` gains `settingsPinned` and `settingsProposals` (both additive; the poll lifts `settingsPinned`). The four setters take `origin`; `_pinned_stand_down` refuses a remote-origin (or origin-less) gesture on a pinned store with a `settingStale` frame whose `why` is "pinned on this machine" and the kept value named. `POST /setting-proposal` (`{"store", "gt", "answer": "apply" | "keep" | "pin"}`, behind the serve token) acts only on this kernel's own record for the store and only when `gt` names the record's stamp: apply runs the store's setter under the peer's stamp as a local gesture and drops the record; keep drops it and remembers the stamp as answered; pin sets the pin and drops it; a stale stamp, an unknown store, a bad answer or a missing record is refused with the reason. The interim alert: one sync notice per record raised or refreshed, in the user's terms (which machine, which setting, from what to what, "Apply or Keep mine in Settings"), until one B's card replaces it. ui/webview/federation.ts stamps every broadcast KERNEL_SETTING copy with `origin` ("local" to this dashboard's kernel, "remote" to every attached host), the message otherwise unchanged. ui/webview/gear.js draws the pending proposal under the affected row from `/version` ("TESTHOST proposes off; this machine is on." with Apply and Keep mine, both posting to `/setting-proposal` with the proposal's stamp; a refusal is toasted and the panel re-fills) and a pinned store's note; gear.css dresses the line at the row's line size with the toast's action dress, through the tokens. The reference's mesh-settings paragraph is rewritten; the plan's status and phases record the split. The switch `STATE/judges-process` is untouched. The mixed mesh: an older kernel without this change still adopts the value a one-A kernel pushes to it and still applies our poll's value on its side, so the two converge one way until it updates. No clock input to any verdict.

**Tests, red first at e7e68f0c** (a detached base worktree, the committed test files copied then restored) under `--python 3.13` and under `3.14` the same: `38 failed, 118 passed` over the four Python modules this head moves (the mesh module, the task-tracking switch, the wiring pins, the seed-token harness). The behavioural reds, each executed at the base: the poll and the push both wrote the peer's value into the local store, `Tuples differ: (True, 9000) != (False, 5000)` twice, and left no record, `Lists differ: ['compact-suggest'] != []` twice; the store moved where the head leaves it, `b'{"enabled": true, "gt": 2000}' != b'{"enabled": false, "gt": 1000}'` byte for byte and `(True, 2000) != (False, 1000)` on the switch (the head's harness falls back to the base's adopter where the new function is absent, so these red on the state, not on a name); `/version` without `settingsPinned` and the proposals map without the store (`KeyError`). The name reds, the honest state of a base with no pin, answer or proposal road: `_set_setting_pin` seven times, `_answer_setting_proposal` five, and the seed-token harness's stub list naming the seam by its new name, nine, incidental to what those tests cover. The three wiring pins red on the setter calls without the origin argument. The extension test red on all three at the base (`# tests 3 # pass 0`): no origin on the copies, no proposal line, no rules in the sheet. Green at the head under `--python 3.13`: the six touched and neighbouring Python modules (`test_mesh_settings_adopt`, `test_task_tracking_switch`, `test_bus_seed_token`, `test_kernel_settings_sections`, `test_settings_page`, `test_judge_settings_sync`) `122 passed`; after the harness tweak the three modules that hold the mesh road and the wiring pins `146 passed`; the served settings labs under `ROMP_SERVED_TESTS_REQUIRE=1` (the task-tracking switch, the tabs recut, the section labels) `26 passed`. The extension: `ui/webview/settings-proposals.test.ts` executes `routeOutbound` (a kernel setting's copies carry local and remote origins, the message otherwise unchanged; a per-install setting carries none) and pins the gear's line, its route call and the sheet; the eight federation queue expectations carry the origin; the routing pin in `multi-kernel-merge.test.ts` (each copy's origin per host, the single-kernel copy local), the toast pin in `setting-stale.test.ts` (the refused value read beside type AND origin) and the gear, stale-fold and settings pins `93 tests, 93 pass`. Both test trees were grepped for every changed literal (the adopter's name, the setters' signatures, the WS calls, the `/version` keys, the fan-out line): only the re-pointed pins named them. The whole-suite rule applies (kernel.py shared helpers, the `/version` shape, the gear's messages): the whole Python suite under `--python 3.13`, devbox, capped, no xdist, `4 failed, 14235 passed, 56 skipped, 2329 subtests passed in 1961.95s (0:32:41)`, the four being pins this head moves and then moved (three wiring pins on the setter calls in `test_setting_gesture_order.py`, now carrying the origin argument; the served task-tracking lab's toast, which read the copy's new origin key as the refused value until `staleRefused` excludes it), each rerun green after and the whole extension suite `5048 tests, 5043 pass, 0 fail, 5 skipped` after the two pin re-points (the first whole pass at this head read one fail each time: the routing pin, then the toast pin, both this head's own literals moved), both BEFORE the push. The added lines carry the ruled-out word zero times and no em-dashes; the only host name is the synthetic TESTHOST.

**Round two (the verifier's finds on 4709b31e, folded at 1e38ce0d).** What changed, by finding:
- HIGH, one record slot per store: records are per (store, machine) in `settings-proposals.json` (`{"hosts": {machine: record}, "answered": {machine: stamp}}`), the sync notice fires per record revision only, Keep mine remembers the stamp per machine, and `/version` serves `settingsProposals` as a list per store, newest stamp first; the gear draws one line per proposing machine and its answer names the machine (`host` in the `/setting-proposal` body; with one record pending the body need not). One name per peer across both legs: the token-gated `/version` now carries the peer's own name (`host`), the poll learns "self-name to remotes-row key" from it (`_learn_peer_name`, memory only, re-taught on the first poll after a boot), and a push naming itself by `_self_host()` lands on the poll's record through `_peer_key`. Driven in `RoundTwo`: A on polling B and C (both off) over four passes holds two records and says two notices, not eight; Keep against B sticks over three more passes; a poll under an alias followed by a push under the self-name is one record, said once.
- MEDIUM, the pin's stand-down read as a file fault: the refusal frame carries `pinned: true` and no `why` (kernel `_tell_stale_gesture`, from `_note_refused_gesture(..., pinned=True)`), files no error-center row, and the gear renders the plan's words: "Task tracking: off was not applied on this machine. Keeping on. Kept: this machine's value is pinned." (a remote kernel's refusal names that machine). Pinned in `setting-stale-fold.test.ts` through the gear's lifted toast block, byte for byte, and in `RoundTwo` on the frame.
- Low 1: Apply against a local click that already outranks the record is refused ("your own newer choice stands; the proposal is dropped") and drops it. Low 2: `current` in the map is the live store value; a record whose value the local store now holds, or whose stamp a later local click outranks, drops at the next pass whoever raised it. Low 3: `_set_setting_pin` takes `origin` and stands down unless it is local; the WS arm hands the message's origin over and federation.ts routes `setSettingPin` to the local kernel alone stamped local. Low 4: a stale answer is refused with "<machine> changed its mind since this line was drawn; the line is refreshed". Low 5: a stale pin gesture is said on stderr and files no `settingStale` frame (`_stale_seen` cleared, no `pin:<store>` in the gear's vocabulary). Low 6: `/version` serves `settingsProposals` and `host` only to a caller that presented the token (`_version_info(authed=self._authorize(q)[0])`); the pins ride both (no names in them). Chosen over stripping the host: the proposals' whole content is another machine's pick, and the gear and the peer's poll both carry the token.

Red first for round two at 4709b31e (the round-one head, the module copied then restored) under `--python 3.13` and under `3.14` the same: `19 failed, 23 passed` over the mesh module (the round-one tests moved by the shape change included). The finds, executed: the flap `8 != 2 : one notice per RECORD, two machines, four passes`; a second machine's record `'name the machine' not found in 'the proposal moved: TESTHOSTC now proposes under gesture 3000'` (the first cut held one slot and the answer met C's record where B's had been); Apply over a newer local click `True is not false` (answered ok, nothing applied); Keep's bookkeeping `{'answered': {'gt': 2000}} != {'hosts': {}, 'answered': {'TESTHOST': 2000}}`; the pin's frame `KeyError: 'pinned'` on the refusal record, twice; the stale-answer words `'TESTHOST changed its mind since this line was drawn' not found in 'the proposal moved: ...'`; the version route `_version_info() got an unexpected keyword argument 'authed'` and the pin's origin `_set_setting_pin() got an unexpected keyword argument 'origin'` (the two names the first cut had no gate for, said as such); the one-record shape `{'compact-suggest': {'host': ...}} != {'compact-suggest': [{'host': ...}]}` across the poll, the push and the route. The gear's pinned toast in `setting-stale-fold.test.ts`, executed against the round-one gear (4709b31e, the test file copied into a detached worktree then restored): `expected: "Task tracking: off was not applied on this machine. Keeping on. Kept: this machine's value is pinned." actual: 'Task tracking: off was not applied on this machine. Keeping on.'` (the round-one gear knew no `pinned` field; with the kernel's `why` it would have read the file-fault clause).. Green at the head: the four moved Python modules (mesh, task-tracking switch, gesture order, seed token) `163 passed`; the extension pins for the gear line, the fold toast, the routing and the send queue `105 pass`; the whole Python suite under `--python 3.13`, devbox, capped, no xdist, `4 failed, 14242 passed, 56 skipped, 2329 subtests passed in 1974.96s (0:32:54)`, the four being `test_chat_vsplit_served.py` VSplitDrag's driver timeouts under the suite's load (`page.waitForFunction: Timeout 20000ms exceeded`, the lab that was hardened against runner load on 2026-09-18), `12 passed` alone at the same head; the whole extension suite `5049 tests, 5043 pass, 1 fail, 5 skipped`, the one being `setting-stale.test.ts`'s size bound on the toast block's source slice, which the pin clause's comment lines pushed past 6000 characters, re-pointed to 8000 and the three toast modules green after.

**Round three (the verifier's finds on 1e38ce0d, folded at 01945b26).** The self-name mapping now lives ON THE REMOTES ROW (`self_name`, learned by the poll from the token-gated `/version` in `_converge_peer_settings`, saved with the row by the supervisor's signature-gated save, gone with the row on detach) and a push's self-name resolves through the rows (`_peer_key`). Chosen over the tunnel because a push arrives on loopback through the pusher's own forward, which names no peer on this side; and over a map in the proposals file because the name is a fact about a row, so it is keyed per row and pruned with it (low c: a row attached under another machine's own name no longer captures that machine's records; the poll's host is always the row key, only a push's self-name is resolved). Both halves of the MEDIUM: when a row first learns its peer's name, every record and kept stamp filed under the self-name is RE-KEYED onto the row key (`_rekey_peer_proposals`, merging into an existing row record by the newer stamp), so a push that beat the first poll leaves one record; and the persisted name closes the restart window, so a Keep answered before a restart holds against the same stamp pushed after it. `detach_remote` forgets the row's records and kept stamps (`_forget_peer_proposals`). Low a: the first cut's answered-only shape migrates to the one recorded machine's kept stamp, else to nothing (no machine named gt). Low b: the gear appends each line after the last drawn one, so the panel reads in `/version`'s order, newest stamp first.

Red first for round three at 1e38ce0d (the round-two head, the module copied then restored) under `--python 3.13` and under `3.14` the same: `6 failed, 41 passed` over the mesh module. Executed: the push-first drive `Lists differ: ['hostb'] != ['TESTHOSTB']` (one machine, two records: the push's stays under the self-name); the persisted name `None != 'TESTHOSTB' : the name persists with the row` (round two's row carried none, so the restart reopened the window); the collision `KeyError: 'TESTHOSTB'` (C's poll under the alias landed under hostb); the round-two one-way test's new row assertion `KeyError: 'self_name'`; the migration `{'hosts': {}, 'answered': {'gt': 2000}} != {'hosts': {}, 'answered': {}}` (a machine named gt); and the detach test on the name `_forget_peer_proposals`, the honest state of a head with no such road, said as such. The gear's order pin in `settings-proposals.test.ts` moved from the `row.nextSibling` insertion to the append-after-last one.. Green at the head: the mesh and task-tracking modules `85 passed`; the three gear pin modules `25 pass`; the whole Python suite under `--python 3.13`, devbox, capped, no xdist, `14251 passed, 56 skipped, 2329 subtests passed in 1944.81s (0:32:24)`; the whole extension suite `5049 tests, 5044 pass, 0 fail, 5 skipped`; both test trees grepped for every changed literal (the global map, the insertion line, the learn call's shape).

Tier: feature


ksarma:
A note after the merge. The evidence was executed against c3695c152, the head that merged; the evidence was executed against it and against the two heads before it (1e38ce0d0 and 4709b31e9) as each was pushed, and every item below was re-read on c3695c152.

The two alias mechanisms come first, since they decide which row a record lands on.

## At this head

- kernel/kernel.py:51207. `_rekey_peer_proposals` re-keys every record under the bare self-name without checking that the name is already another row's key, so when another row's key is the name a poll later learns for a different machine, the records under that key move onto the learning row and are overwritten with the learning machine's stamp. The body's claim is true only in the B-first polling order the new test uses; a C-first drive leaves the other row's record gone (a `KeyError` on its lookup) and a wrong-machine notice. Fix: in `_learn_peer_name`, store the name but skip the re-key when it is already a row key; a stamp-equality clause on top reopens the duplicate the fourth commit closed. A push before the learning row's first poll still reaches the wrong row through `_peer_key`.

- kernel/kernel.py:51216. `_peer_key` returns the first row whose learned name matches, so two peers with one short hostname collapse their pushes onto one row's record, which flaps between their stamps with a "proposal moved" notice on every push and poll. Fix: learn a name two rows report on neither, un-learn the first, persist an ambiguous marker so a restart does not relearn it, and resolve the shared-name push by matching its stamp to the row whose lifted version stamp equals it, else file it under no row, said once; under the pusher's own name it still flaps at the same value.

- kernel/kernel.py:22729 and 22082. The absorb in `attach_remote` and the check-in rename pop a row with neither re-key nor forget and carry no learned name to the survivor; only the detach at 22959 forgets. Through the real attach and check-in code one machine ends with two records, two lines and two notices, and a Keep on the live line leaves the dead-key line standing. Fix: re-key old to new at both points after the remotes lock is released (a forget drops a Keep) and carry the name.

- kernel/kernel.py:51380. The "already answered" check is exact equality and `answered` is popped on every raise: a Keep at 2500, then a late push at 2000, rewrites the record down with a notice; Apply at 2500 in between is refused as "changed its mind"; the next poll rewrites it back up with another notice. Fix: the kept stamp as a floor (`pgt <= answered.get(host) or 0`), a push stamp older than the standing record ignored without rewrite or notice, the kept stamp never popped.

- kernel/kernel.py:67715. The drop-on-equal rule runs only in `_propose_peer_settings`, the poll path, so a click on this machine matching a standing proposal leaves the record up while the proposing peer is down, Apply refused as "your own newer choice stands", on all four settings. Fix: after a local apply in each socket arm and in the Apply answer, drop records whose value equals the new one or whose stamp it outranks, under `_SETTINGS_LOCK` in the arm callers, never in the setters (the nudge lock inverts the order); omitting equal-value rows from the served map is not enough, since a newer differing click leaves a differing record.

- kernel/kernel.py:51120. `_json_store` maps every read fault to an empty map: with mode 000 on the pins file a remote gesture overrides a local pin silently, and the next pin write replaces the unreadable file, dropping the earlier pin for good with no quarantine copy. Fix: the four-state read the auto-nudge store already uses here (absent, parsed, quarantined, or UNPROVED with the last proved snapshot) for both stores, cached per store under the settings lock; a non-local gesture and a pin write refused over UNPROVED; no raise or drop over an UNPROVED proposals read, so a Keep survives; the tag kept through both readers' dict filter and out of the pinned map served to peers.

- kernel/kernel.py:51329. The notice a hub's push raises on the receiving peer reads "(this machine has it ...); Apply or Keep mine in Settings"; host-prefixed into the hub's bell through the built `mergeHostFeeds`, it names the wrong machine and points at the hub's Settings, which draws no line for that record since the gear fills from the local kernel only. Fix: `_self_host()` naming the machine in both clauses, the existing host-name environment override as the seam. That an answer lives on one machine's Settings is a separate, deferred limit; this is a bell row, not a session injection.

- kernel/kernel.py:51092. A push with no host (an older hub) is filed under the literal "another machine" beside the poll's record for the same machine: two records and two notices while that peer stays older, in both the push-then-poll and poll-then-push orders, the fallback wording uncapitalized. Fix: resolve a host-less push to a standing record with the same value and stamp, or a kept stamp at that value, scoped to host-less reports so two named machines never merge; have the poll's inbound leg re-key such a fallback record onto the row key once a name is learned, carrying `answered` across, since the push-first order needs the poll side to re-key; capitalize the fallback.

## Also at this head

Five code paths have no executed test of the behaviour they carry; the first four are shown by mutation at this head, the fifth by browser probes.

- kernel/kernel.py:23164. The poll's lift of the peer's own name, the only production path teaching `_learn_peer_name`: the three-machine test hand-builds the polled dict, the real-poll class never sends a host, and a lift mutated to always yield None turns no test red. Fix: a real-poll test that sends a host.

- ui/webview/settings-proposals.test.ts:45. The lines, the pinned note, the host lookup, the answer POST and the newest-first order are pinned by source regexes alone (the fourth commit replaced one regex with two). A lift of the proposal code over a mini DOM fails under order, host, handler-swap and pin-branch mutants the regexes pass (a literal revert is caught); the existing browser test reaches `fillProposals` only on the empty version path. Fix: an executed leg; the node browser leg skips in CI because npm test runs before the browser install, so the leg CI runs is a served-lab Python test.

- tests/test_mesh_settings_adopt.py:331-336. The pin-before-stale order and the gate in three setters are pinned by `inspect.getsource`: a textual reorder is caught, a gate whose result is ignored with call text and order intact is not, and under that mutation auto-nudge, compact-suggest and file-editing stay green while task-tracking alone reds. Fix: the executed loop over the four stores, whose older-stamp remote leg catches a gate moved under the stale check.

- tests/test_task_tracking_switch.py:196. No dispatch through `_dispatch_ws` in the tests carries an origin or sends the pin message; mutants forcing the task-tracking arm's origin to local at the setter call, wrapping the pin arm the same way, or reading the wrong body key leave the suite green, and no source pin covers the setter line. Fix: one executed dispatch per arm, remote on a pinned store then local; a message with no origin key reads as remote.

- ui/webview/gear.js:1850. `fillProposals` removes every proposal and pinned node and rebuilds them on each fill (every stale frame while the panel is open), handlers bound on the rebuilt nodes, no in-flight guard; the fourth commit moved only the insertion anchor. In Chromium a stale frame between mousedown and mouseup detaches the pressed node and no click fires (0 POSTs; the control press posts once); a double click under a delayed answer posts twice, the second refused and toasted as an error; answering one machine's line drops a concurrent press on another machine's line for the same store. Fix: reconcile in place keyed by store and host (delegation alone cannot recover a mid-press detach); disable and relabel the pressed button under an in-flight key of store, host and stamp, re-applied on every fill; check the anchor's `nextSibling` before inserting, since re-inserting the same node is itself a detach; an executed browser leg.

## Smaller points

- kernel/kernel.py:51235, 51206, 51022. The re-key moves a self-name record whose stamp equals the row's kept stamp rather than dropping it, so kept stamp and standing line coexist once the name is lost and relearned; the peer-supplied name is persisted with no shape check, unlike `kernel_sha`; and it is written onto the live row outside the remotes lock, against the call-site comment. Fix: drop at or under the kept stamp; gate on the id rule, said once; write under the lock, re-looking the row up by key.
- kernel/kernel.py:51274. The migration maps a first-cut kept stamp onto the one recorded machine only when exactly one host record stands, a shape a first-cut Keep never produced, so the shape it did leave is re-proposed after upgrade; bounded by the first cut never reaching main. Fix: a wildcard kept stamp the answered lookup consults, or remove the branch together with its docstring sentence.
- kernel/kernel.py:51450 and 67713. The Apply-on-pin answer ignores the stand-down of `_set_setting_pin`: with the pin store stamped ahead of the clock it answers ok, pins nothing, drops every record, and the next poll re-raises with a notice; the socket pin arm hands the same mint a stampless gesture; reachable today only through the route. Fix: mint max(now, previous plus one) in the setter when no stamp is given, answer false on a stand-down, keep the record.
- kernel/kernel.py:51086. A stampless push body is ignored yet acknowledged ok; no in-tree pusher omits the stamp. Fix: answer false naming it, no fresh mint, pinned in `TrustGate` for absent and zero.
- kernel/kernel.py:51432. Apply is refused by stamp alone, so a click on this machine already on the proposed value gets a refusal toast. Fix: compare values, equal drops and answers ok; the refusal wording covers a click and a same-value peer lift alike, which the kernel cannot tell apart.
- kernel/kernel.py:51051. `_older_peer_settings` lists a store the peer kept or has standing as older, so the hub re-POSTs it every supervisor pass after a Keep (five passes, five POSTs). Fix: kept and standing stamps in the version payload, matched by stamp, since the peer's key for the hub may be an alias the hub does not know.
- kernel/kernel.py:51329. The notice lives only in the process-memory ring, so a standing record is not said again after a restart until Settings is opened. Fix: a stable signature on store, host and stamp, persisted with the record and served every frame while it stands; said once after boot instead, it re-logs in every browser, since the bell's seen set is rewritten from the served set each frame.
- kernel/kernel.py:51378. Keep is keyed on the exact stamp, and a peer's stamp lifts with no click when it polls a third machine at the same value under a newer stamp, so a kept proposal returns as a fresh record and notice. Fix: stamp and value per machine in `answered`, nothing raised while the value equals the kept one, the entry spent once the value equals this kernel's again so a later re-divergence still proposes.
- ui/webview/gear.js:1078, 1093, 1095, 1097. The four kernel-setting posts carry no origin, so a dashboard without the federation layer (VS Code, a served page whose federation script failed) is read as remote and refused on a store its own machine pinned; the pin half is latent, with no sender. Fix: origin local when `window.__rompFed` is absent, or unconditionally, since the federation router overrides it per copy; re-point the regex in ui/webview/task-tracking.test.ts.
- ui/webview/gear.js:1843. `answerProposal` parses JSON then catches to null, so a 404, 403, 500, a refused connection or a non-JSON 200 shows nothing. Fix: check ok after the kernel's JSON refusal path (kernel refusals are 200 JSON) and toast, keeping the re-fill, with an executed leg.
- ui/webview/gear.js:1658. The pin clause rides the merged toast's latest-frame why slot: a pinned refusal joining a live toast opened by a plain stale refusal keeps "Apply anyway" against a pinned value; in the other order the pin clause is erased. Fix: a per-toast pinned set in `staleOpen`, the action removed on a pinned join, the clause rendered from the set; the file-fault clause has the same shape and predates this PR.
- kernel/kernel.py:51440. The Apply answer calls the setter bare and skips the arm's post-apply work (views dirty, pure-feed drop, producer wake, push, tracking echo); the gear re-fills at once and the panes catch up within about 2.5 s through the feed rebuild floor, while the shell rail waits for its next version event. Fix: one helper for arm and answer, the echo by broadcast or the gear's fill.
- kernel/kernel.py:67713. A stale, remote or origin-less pin gesture clears `_stale_seen` and files no frame, the floor the docstring and the body state; no dashboard sends it today, and a store-vocabulary frame would climb the value store's gesture clock. Fix: a note at the arm and in the plan's phase-two section.
- tests/test_mesh_settings_adopt.py:753 and 796, kernel/kernel.py:51364. The dead `_version_payload` branch stands; the version handler's token gate is pinned by source text, and the pin on its authed argument is the only guard against always-authed wiring, so it is replaced by an executed GET pair, not deleted; every pushed assertion is the empty list, so the push's applied branch runs in no test; the stamp-outranks drop has no test, and its second poll needs a different machine or the mutation stays undetected. Fix: delete the dead branch; an executed GET pair through the auth harness in place of the text pin; one stub whose call applies the payload, asserting the store in `pushed`; the two-machine stamp-outranks test.
- kernel/kernel.py:51181, 27371, 27380, 51082; plans/settings-across-machines.md:55, 67, 85; docs/reference.md:4313. The stand-down docstring says the frame carries why when pinned (it carries pinned true, no why), and the supervisor comment, the "adopt-settings" log key and the push handler's docstring describe adoption through the setters, a path that no longer exists; the plan still says one proposal per store, the one-slot shape, the first cut's answered shape and a route body without host, while the code is per store and machine and refuses a host-less answer when two records stand; the reference says the pin is set from this dashboard by `setSettingPin` while no control sends it and the plan puts the pin glyph in phase two. Fix: rewrite the four kernel sites; update the plan's model bullets (the per-machine record and file shape, host in the route body, the pinned field on the frame, the row's name) or date a note; one clause in the reference, the body and the kernel docstring that both pin doors are API-only until then.
- ui/webview/gear.css:220. `.rs-proposal` and the pinned rule are 11px while the comment calls that the row's line size; the row's line is 0.92em (11.96px) and the hint 12px, so the drift shows under Auto Nudge and Suggest /compact, whose line is visible. Fix: 0.92em on both, the comment corrected, the two regexes re-pointed.

The tests run per item, their results, and the shapes of the tests asked for are in the next comment.

ksarma:
Tests run for the review of c3695c152, by item, with the shapes of the tests asked for.

## Tests

Alias mechanisms (kernel/kernel.py:51207, 51216, 22729, 22082):
- A three-machine drive at c3695c152 in the C-first polling order (the alias row polled before the machine whose name it equals) left the alias row's record gone (a `KeyError` on its lookup) and a wrong-machine notice; the B-first order in the PR's own test passed. On a scratch copy with the one clause "skip the re-key when the self-name is already a row key", the C-first drive passed and the mesh module stayed green; a scratch copy that also required the old record's stamp to equal the peer's version stamp reopened the duplicate record the fourth commit fixed. Asked for: the C-first attach and poll ordering added to the three-machine test.
- Two rows whose peers both report one short hostname, pushed and polled repeatedly at c3695c152: one row's record alternated between the two stamps with a "the proposal moved" notice each time. A scratch copy that un-learned both rows and persisted an ambiguous marker stopped the alternation; a scratch copy that filed the push under the pusher's own name still alternated at the same value. Asked for: a two-peers-one-name test.
- The real `attach_remote` absorb and the real check-in rename (token fetch and tunnel stubbed) at c3695c152 each left one machine with two records, two gear lines and two notices; a Keep on the live line left the dead-key line standing. A scratch patch re-keying at both points and carrying the popped row's name fixed every road with the PR's 47 mesh tests green.

Standing checks, local click, pins, notice, host-less push (kernel/kernel.py:51380, 67715, 51120, 51329, 51092):
- Keep at 2500, late push at 2000, poll: the record went to 2000 with a notice, Apply at 2500 was refused as "changed its mind", the poll restored 2500 with a second notice. A patched copy with the floor, the ignore-older and the never-pop clauses raised no spurious notice or rewrite and kept the mesh, task-tracking and bus-seed modules green.
- The real `_dispatch_ws` arms, all four settings: a local click matching a standing proposal left the record standing with Apply refused. A scratch patch with a caller-side settle after apply under the settings lock fixed all four arms with the mesh module green; at the previous head one patched probe passed 5 legs with the PR's 163 tests green. The test at tests/test_mesh_settings_adopt.py:458 pins the next-pass drop and needs rewriting through the real arm rather than the harness's direct set.
- Mode 000 on settings-pins.json at c3695c152: a remote gesture overrode a local pin with no notice; a following local pin write replaced the file through `os.replace` and the earlier pin was gone with no quarantine copy. At the previous head the same ran with a garbled file. The four-state reader the auto-nudge store uses exists in the file and is not wired to these two stores.
- The mesh harness plus the built `mergeHostFeeds` at c3695c152: the hub's bell showed a line naming the wrong machine with no matching line in the hub's Settings. The text pin at tests/test_mesh_settings_adopt.py:157 needs rewriting with the new wording; the seam is the existing host-name environment override or reassigning the kernel module's self-host function.
- Host-less push then poll, and poll then host-less push, at c3695c152: two records and two notices for one machine in both orders. A scratch copy resolving host-less pushes against standing records and kept stamps, with the poll side re-keying the fallback record, collapsed both orders to one record and one notice across restarts and several older-hub scenarios. At the previous head the older push and version shapes were checked from the base's code: zero host keys.

Coverage gaps (kernel/kernel.py:23164; ui/webview/settings-proposals.test.ts:45; tests/test_mesh_settings_adopt.py:331-336; tests/test_task_tracking_switch.py:196; ui/webview/gear.js:1850):
- The host lift at kernel/kernel.py:23164 mutated to always yield None: zero reds in the full existing suite (the one observed failure reproduced identically on the unmutated head). The test asked for, written and run: red under the mutation, green at c3695c152. Shape: a `ThroughTheRealPoll` test whose fake peer answers the version route with host TESTHOSTB, asserting the polled dict carries that host, then a converge on the row and an assertion that the peer-key lookup of TESTHOSTB resolves to the row key, and that a payload without host polls as None; the seam is the host-name environment override or reassigning the self-host function (the old name map no longer exists).
- The block from the proposal rows through `fillProposals` lifted from gear.js over a mini DOM shim, fed synthetic task-tracking records plus a pinned store: green at c3695c152, red under order, host, handler-swap and pin-branch mutants that the regex-only test passes unmutated; a literal no-op revert is caught by the newest regex, text-preserving behavioural mutants are not. Shape: the shim's `querySelectorAll` answers the exact two-selector query for proposal and pinned nodes (else the clear step clears nothing and a re-fill duplicates lines); assert the two lines in newest-first order, the POST bodies with host and stamp, the pinned note absent while a line stands and present once records are gone; drive the click through the stored listener and await a microtask before reading the body. The leg CI runs needs to be a served-lab Python test under ROMP_SERVED_TESTS_REQUIRE=1, since the node browser leg skips before the browser is installed; the existing browser test reaches the fill only on the empty version path.
- Mutation over the four setters with the gate's result ignored and its call text and order kept: auto-nudge, compact-suggest and file-editing green, task-tracking red. The loop asked for, written and run: green at c3695c152, red under every single-setter mutation and a stale-check reorder. Shape: one fresh kernel per store (the class fixture pins task-tracking only), set at 1000, pin at 1100; a remote gesture with a newer stamp returns None with the pinned refusal record, an unchanged store and no stale notice; a remote gesture with an older stamp refuses pinned with no stale notice (the leg that catches a gate moved under the stale check); a local gesture applies. The `assertLess` source pin can stay as a secondary.
- Mutants on scratch archives of c3695c152 over tests/test_task_tracking_switch.py and related modules: the setTaskTracking arm's origin forced to local at the setter call, the setSettingPin arm wrapped the same way, and the arm reading the wrong body key all left 158 to 168 tests green; a class written to the shape asked for failed under each and passed unmutated. No dispatch in tests/*.py sends the pin message or an origin through `_dispatch_ws`. Shape: on the module's base class, dispatch the task-tracking setter with a remote origin on a pinned store and assert exactly one frame with type settingStale, the setting, pinned True, the kept value and the gesture stamp, no why key, the gesture echoed minus its stamp with origin kept; then a local origin, asserting the echo frame and the store; a message with no origin key reads as remote. One pin test: local applies with the store in the pinned map; remote or absent files no frame and leaves the map unchanged; a stale local pin gesture also files no frame; the non-bool leg needs a client with a callable send, since the flag refusal replies only then. Wrap dispatches to capture stderr. The `assertIn` at tests/test_mesh_settings_adopt.py:355 may stay as a secondary.
- Chromium over the bundled gear.js at c3695c152 and the previous head: a settingStale frame between mousedown and mouseup detached the pressed node and fired no click (0 POSTs), a control press posted once; a delayed answer left the button unacknowledged; a double click posted twice with the second refused and toasted as an error; answering one machine's line dropped a concurrent press on another machine's line for the same store; focus moved to the body after a keyboard answer. Asked for: an executed leg in the gear browser harness, re-pointing the two regex pins the reconcile changes.

Smaller points (executed at the previous head, 1e38ce0d0, and re-read at c3695c152, whose lines are cited):
- kernel/kernel.py:51450: an ahead-stamped pin store answered ok with an empty pinned map and the next poll re-raised with a fresh notice; a scratch fix passed 46 tests including the stamp-gate test in one run and 80 in another, with the write-failure leg refusing and keeping the record.
- kernel/kernel.py:51086: absent, zero, negative, string and None stamps all acknowledged ok, where the base had applied a stampless body; a scratch fix passed 88 tests; no in-tree pusher omits the stamp.
- kernel/kernel.py:51432: a same-value newer-stamped click was refused with an error toast; a patched copy answered ok with 42 mesh tests green.
- kernel/kernel.py:51051: five supervisor passes after a Keep, five POSTs, byte-identical records.
- kernel/kernel.py:51329 (restart): with the ring cleared, a poll of a standing record produced no notice; a say-once-after-boot variant re-logged in every browser.
- kernel/kernel.py:51378: a three-machine mesh, a Keep undone by a same-value stamp lift on a peer.
- ui/webview/gear.js:1078: an origin-less setter call refused with pinned True and a stderr stand-down line; origin local applied. VS Code's extension and the served shim forward the message verbatim.
- ui/webview/gear.js:1843: 404, 403, 500, a refused connection and a non-JSON 200 each gave 0 toasts; the 200 JSON refusal control toasted once.
- ui/webview/gear.js:1658: a node probe over gear.js reproduced both orders (stale then pinned kept the action; pinned then stale erased the clause).
- kernel/kernel.py:51440: zero post-apply hooks fired on Apply; the gear re-filled at once; the panes caught up within about 2.5 s.
- kernel/kernel.py:51274: a true upgrade simulation (the first cut's Keep write loaded, then a poll with that head's kernel) forgot the Keep and re-raised; a scratch wildcard fix passed 42 PR tests plus 4 probe legs.
- tests/test_mesh_settings_adopt.py:753 and 796, kernel/kernel.py:51364: the handler's authed argument mutated to always true left 91 tests green; a dead-branch and an IndexError mutation of the push's applied branch survived 163 passed; the executed pairs asked for passed there and failed under each mutation; the stamp-outranks leg's mutation stays undetected unless the second poll comes from a different machine.
- ui/webview/gear.css:220: headless Chromium computed `.rs-proposal` at 11px against 11.96px for the row's line and 12px for the hint; the 0.92em fix landed both proposal divs at 11.96px with the action buttons at 11px.
- kernel/kernel.py:51181 and the plan and reference items: checked by grep and blame, no test owed.

PR #1860, opened 2026-09-19
Chat history: refuse a frame the page cannot place onto its held regions, never dropping scrolled-back history

henrypinkard:
## The bug

On a chat read long and scrolled back (reported on remote sessions viewed through the federation relay, the user's top priority), the turns between the reader's read point and the live tail VANISH on a single frame: an assistant reply from an hour earlier ends up directly above the running turn, the hour between gone from the render though present on disk. A message sent from the page could also stay a "sending…" bubble after its reply landed.

The defect is entirely page-side. The relay only amplifies it (a remote session is the one a reader keeps open long and scrolls back into); the loss reproduces on a local session too, and `federation.ts` copies each frame whole. A chat read long holds several regions: the loaded history runs above a gap and the always-resident tail run. Several branches trusted one index from the kernel without checking it against what the page holds, and applied a frame they could not place instead of refusing it.

## Reconciled onto PR 1877

PR 1877 rewrote this same `upsert` path and now owns two of the shapes this PR first guarded. Its invariant: **a full frame is authoritative for `[tailLo, end)`; a frame whose tail start the kernel could not name leaves no regions, files one `regions-dropped` row, and a landing takes the older wire.** So the no-`tailLo` and non-proto-2 cases are 1877's by design, and this PR drops its own keep-regions and not-proto-2 branches (the landing lab's road 13 returns to 1877's regions-dropped shape). What remains is the one loss 1877's merge-by-position does not catch, below.

## The fix (page-side guards, plus one kernel bound)

1. **`chatTail`** resolves the delta's `afterUuid` inside the **tail run only**; an anchor in a history run asks for the full frame. A numeric-`from` delta that would truncate into the loaded history is refused **before** truncating (keeps the parked read point).
2. **`upsert` guard 3 is the merge's own loss rule, not a `tailLo` comparison.** For every held run the frame claims to cover (`r.lo >= tailLo`, the **tail run included**), it refuses when `splitHeldAgainstFrame.before` holds a transcript row the frame's later-starting content does not carry, and asks for a frame the merge can place. This catches the two shapes 1877's merge drops whole and the earlier `tl <= r.lo` test missed: a `tailLo` that lies **inside** a held read run (the held tail run went 503 to 258) and a `tailLo` at the **sole tail run's own lo** (536 to 291). A frame that re-carries a run from `tailLo` shares its first key, so its `split.before` is empty and it applies; rows dropped at or after the frame's content are legit retractions 1877 keeps.
3. **The kernel answers EVERY needFull** (the per-client dedup is removed). The page bounds a refused-frame loop itself: the `refusedFrameLatch` asks once, keeps the held state after a second identical refusal, and stops asking. A kernel dedup that dropped a genuine second gap ask left the page's `awaitingFull` set and froze the tab until a reconnect.
4. **`regionsAbsorbTail`** returns a bool and the caller re-bases on false rather than relabel a short tail over eaten history (a backstop for the `chatTail` anchor branch).

The refused-frame loop diag carries the real accumulated count when the latch clears (the in-flight row only says a loop has started) and re-arms per clear, so a recurrence after a clear is filed again.

## What this fixes, and what it does not

Fixed: **the page never drops rows it holds, and never storms needFull.** These guards are containment with a diag trail. **Not fixed:** the root cause of a wrong or missing `tailLo` on a relayed full frame is still open (see the hunt's report); when it is found, these guards keep the symptom from reaching the reader in the meantime.

## Red first

`tests/test_relay_tail_vanish_served.py` boots two kernels and a relay, parks a reader mid-thread, and injects each frame through the page's own inbound door. Nine scenarios: `taillo_low`/`straddle_lying`/`tailrun_drop` (the loss, refused), `anchor_in_history` (guard 1), `equal_hi_legit`/`straddle_legit` (legit larger windows apply with no ask), `loop` (five identical refused frames give at most two asks), `from_into_history` (the read point survives), and `not_proto2_rebase` (a non-proto-2 frame re-bases with a `regions-dropped` row, no throw, 1877's regions-less landing). With guard 3 neutered at the base the three loss shapes red (store 503 to 24, 503 to 258, 536 to 291); at the head all nine pass.

`tests/test_needfull_no_dedup.py` drives the real WS dispatch: three gap asks for one sid are each answered with the full frame (`[[sid],[sid],[sid]]`). Red with a per-client kernel dedup, where the second and third drop.

Node extension suite green (5101 pass, 6 skip); the 1877 client-merge-guard lab and the landing lab green (19); `tsc` clean. Source pins re-pointed on both test trees.


ksarma:
Review of 84ed73dfe, the current head; everything below was executed against it, with earlier heads named by sha where a finding moved.

Two invariants cover most of what follows:

1. Every `needFull` gets an answer. The page arms `awaitingFull` when it asks and clears it only when a full frame lands or the socket reopens (render.ts:17433, 17141, 17440). Deduping the kernel's work is fine; deduping the answer is not: an unanswered `needFull` leaves the latch armed for the socket's life, and every later gap ask for that sid is refused. The 2 s window at kernel/kernel.py:65761 approximates an event, the previous answer landing on the page; key any dedupe on that, or answer a repeat by re-sending the last full frame unchanged. A dedupe keyed on the request's content would not help, since the repeat's answer is often byte-identical.

2. A full frame supersedes a held run when it carries the head identity plus the tail bound: a `headKnown` frame with `tailLo` 0 is the kernel's whole list and supersedes every held run by construction; a windowed frame supersedes a run only if it re-carries it from `tailLo` on. Guard 3 at render.ts:17173 tests neither: it refuses the whole frame (/clear, rewind) and admits a windowed frame that lies.

## At this head

- kernel/kernel.py:65759-65762, the `needFull` dedupe. On a hermetic kernel load, two asks for one sid inside 2 s collapse to one push; the refused frame is the synchronous answer to the previous ask, so the re-ask always lands inside the window and `awaitingFull` stays armed. In a two-kernel relay run under Chromium with a rewritten `tailLo`, the store never advances at this head, where the merge-base 713dc9ef8 applied the frame and bebdc3193 looped; the swallow path files no diag row and writes no stderr line (if any early return stays, file a client-diag row naming the client and sid). Fix: drop the window (`refusedFrameLatch` already bounds identical refusals to one ask), or answer a repeat with the last full frame so `awaitingFull` clears. tests/test_chat_delta_resync.py:135 is a text pin of the window and goes red under this; replace it with an executed case driving `Handler._dispatch_ws` twice (age `client['needFullAt'][sid]`, no sleep) beside tests/test_chat_skeleton_reconnect.py test_04, which catches a flipped comparison but not an inert gate; add a served pin that the tail moves after the re-ask, not only that the re-ask was answered. Necessary, not sufficient: a persistently lying `tailLo` still freezes the tail's content. Applying the identical answer once through the pre-PR merge converges but re-drops history in the PR's motivating relay case, so scope it to a frame answering the page's own pending ask (`pendingFullWhy.has(msg.id)`, read before the delete at 17180) or make the kept state visibly stale; nothing consumes the diag row or the latch into view state today.

- render.ts:17173, a whole frame refused. A `headKnown` frame with `tailLo` 0 sharing any uuid with a held run is refused as `taillo-below-held`. With the real `sharesAnyUuid` over a reader holding a head run [0,3) plus a tail run, a /clear (constant uuid `system:head`) and a rewind cutting inside the head run both reproduce: the merge-base applied both; this head keeps the pre-clear or pre-rewind conversation, asks once, latches on the identical re-send, and after the first later turn the dedupe swallows the re-ask, so the tab freezes (both shapes looped at bebdc3193, the whole-frame rewind at bcf7dd3e4 too). A live relay run widened the trigger to any held tail ending in an overlay card (`apiError` there); it needs the cut to move backward (a rewind, a chat delete, a /clear, an overlay-card shift), never the ordinary append path. Fix: `!(msg.headKnown && tl === 0) && ...` in guard 3; with that clause every /clear and whole-frame rewind shape applied as at the merge-base, the lying `taillo_low` shape stayed refused, and the node suite and served lab passed. A windowed rewind, a cut below a held run that abandons that run's tail, stays refused under the rules in the next item and converges only through the own-ask bound above. Excluding the synthetic uuids inside `sharesAnyUuid` itself is unsafe, since it also decides same-transcript versus new episode; apply any exclusion at the guards' entry only.

- render.ts:17173, a lying frame admitted. `tl <= r.lo` re-admits the target defect for a same-transcript full frame whose `tailLo` lies strictly above a held history run's lo and below the tail run's lo without re-carrying the tail run. Over a reader holding run [137,263) plus a tail run, `tailLo` 138, 200, 263 and 300 each applied with 0 asks and 0 diag rows: the merge superseded the tail run and dropped its uncarried rows, and inside (137,263] coalesced the history run onto the frame with the gap gone. At the merge-base and at bebdc3193 each of these was refused or handled correctly. Fix: refuse by the merge's own drop rules: (a) a run with `tl <= r.lo` holding an event the frame does not carry; (b) a run straddling `tl` whose last event the frame does not carry; (c) the frame's first key held in a run whose lo exceeds `tl` (a held event's turn index never falls), guarded by `first != null` and searched over the tail run too. With all three, every lying `tailLo` was refused with one ask, the `equal_hi_legit` and `straddle_legit` windows and an honest re-send still applied, and the served lab passed; a `tailLo` equal to the tail run's lo with a short frame is correctly accepted.

## Also at this head

- render.ts:17199, the no-`tailLo` keep. A proto-2 full frame with no numeric `tailLo` for a regions session is kept with its events discarded and no ask; a new turn it carries never reaches the store, and since the kernel advances the client's base on send, every later `chatTail` anchors on an event the page never applied. Bounded: the held tail must already be behind when the frame lands (missed turns, a rewind, a forced probe); the stated real trigger, a session idle past 48 h reconnecting, gave in-sync frames with no source patch. The comment's analogy to the empty status frame does not hold at the kernel: that frame leaves the client base empty, this one leaves it at the frame's last event. Fix, kernel side first: `_tail_run_start` (kernel.py:49904) resolves the sid only through `_sessions(now)`; `_session_row` (kernel.py:4739), the single-sid lookup with the same wide-walk fallbacks `_alive_sessions` uses, returned a numeric tailLo where that lookup returned None. An ask alone does not converge, because the kernel answers with another such frame. Page side, splice a frame sharing a key with the tail run onto the tail from the first shared key (past the last key keeps rewound-away events; never into a history run), or show the kept page as stale; nothing sets such a state today. The `not-proto2` face is mechanically identical but unreachable for a regions session on the shipped wire (a page's proto is fixed and every kernel forwards it).

- render.ts:17185-17190, the latch key. `coordKey` includes the frame's last key, so a refused re-send of a live session (one more turn per frame) never latches: five such frames gave 5 asks against 1 with the last key dropped, and left `awaitingFull` armed, so later chatTail gaps for the sid were dropped until reconnect. Bounded: dropping the last key cuts a wire-speed burst from 248 asks to 11 and halves the slow face but leaves the per-change ask, and against this PR's kernel the choice is moot, since both variants freeze after two asks through the dedupe above. Key on (why, `tailLo`, first key), or on whether the frame answers the page's own pending ask, and add a lab scenario whose refused frames each carry one more turn, with no applying delta between them, asserting one ask.

- The `clearRefusedLatch` resume paths (17206, 17557) run as no-ops in existing tests and are observed by none: deleting both left the webview suite, tsc and the Python pins green, and a session refused twice never asks again after progressing. Add a lab scenario that refuses twice, progresses by a delta or a full frame, and refuses again, asserting one ask per cycle; executed, it passes at this head (11 passed) and fails with the calls deleted (2 failed, 9 passed).

## Smaller points

- render.ts:17202: the `full-frame-desync-loop` row is filed on the second identical refusal for every why, but `no-taillo` asked nothing at the first, so the row's claim is false there; gate it on the same why list as the ask.

- render.ts:17503-17504, `tailStart` by key lookup. Bounded: the numeric-`from` guard at 17551 now catches a duplicate landing anywhere but the last event of the last history run; the residual, a duplicate that is also the tail run's first key landing there, truncated the tail run silently (7 events to 4) with no ask. Compute `tailStart` by count of the history runs' non-optimistic, non-held events and resolve the anchor in `kernelEvents.slice(tailStart)`.

What the suite pins and does not pin at this head follows in a second comment, so this one stays about the merge.

ksarma:
What the suite pins and does not pin at 84ed73dfe, as a second comment to the review above.

- tests/test_relay_tail_vanish_served.py: `_assert_scenario` (300) reads no `needFull` value, so gating both `requestFullSession` calls (17199, 17551) off left all 8 scenarios and the node tier (5031) green. Add an `expect_asks` argument (taillo_low 1, no_taillo 0, not_proto2 1, anchor_in_history 1, from_into_history 1). Make the loop test assert exactly one ask: the mechanism asks once, so the adjacent comment and the PR body's "at most two asks" should say exactly one. `equal_hi_legit` and `straddle_legit` pass at the merge-base (6 failed, 2 passed there) and are red only at bebdc3193, so "each verified red at its base" and the docstring's "Red at main" overclaim (the equal_hi_legit comment already names bebdc319; straddle_legit's does not). Four names are dead (import re at 28, BIN at 38, expect_read_drop at 300, the midGap snapshot field at 83). The docstring still says four shapes, a guard at a run's hi, and every site asks; reword it to five refusal shapes across eight scenarios, the lo rule, and the ask split (taillo_low and not_proto2 ask once, no_taillo asks nothing, a repeat latches). The skip text "CI installs none" is stale, since CI's vscode-extension job ran this lab at this head (run 35423033810).

- Node tier: no harness executes guards 1 to 4, either diag row, or the latch; mutants killing each pass the node suite. chat-proto2-exec's `liftUpsert` stubs the same-transcript check to a no-op and resolves the refusal latch to a stub that throws once the real check is in scope. Give the guard cases their own lift scope with the real `sharesAnyUuid`, a real latch map, the latch clear, and recording stubs for the full-frame ask and postMessage; a probe built that way goes red under every mutant. Pin the head's contract: not-proto2 asks, no-taillo does not, a second identical refusal posts the loop row with count 2 and asks nothing, new coordinates ask again. Guard 4's false branch is unreachable through `chatTail` at this head (`from < hist` precedes it), so pin `regionsAbsorbTail`'s body directly. The dupkey case for tailStart is red at this head until the count-based tailStart lands; use the shape that is red, the repeated anchor as the last event of the history run, and record [id, why], since the harness's fulls records ids only. The lab already pins the latch's upper bound and the `equal_hi`, `straddle` and `loop` asks, so the gap is the ask gate and the latch clear.

- tests/test_landing_notice_browser.py:692, road 13. Replacing the re-point branch at render.ts:11449 with a busy-refusal stub leaves the road green with an identical payload, where the merge-base's road went red. The click issues a fresh `loadAround`, and nothing in the landing branch reads `gapLoading`, so a held `loadTurns` is never re-pointed and the comment describes a mechanism no code path connects. Keep the road but reword it to what it pins: a kept-regions page with a loadTurns fetch held still resolves a click with no busy refusal. Assert `pointer-fetch-window` from the click's locateDiag row instead of the either-or at 708, narrow heldGap13 to loadTurns, and drop the roads 1/9/12 claim (they assert no trail token). Restore the re-point pin on a proto-2 page whose first frame carries `tailLo` null (an init-script rewrite before the boot frame; a post-boot re-post keeps regions under guard 2), with loadOlder parked, asserting `pointer-fetch-older` from the click's locateDiag row, which is what 11449 pushes; that road must go red with pointer-fetch-busy under the same busy-refusal mutation.

- Also: docs/reference.md:2842 still says an anchor the page does not hold is a gap; the head's rule is an anchor it does not hold in its tail run.

ksarma:
A note after the merge. Everything below was executed against 203bfe57c, the merged head, with the base and the earlier head named by sha where a comparison is drawn; every file and line reference is on that commit. It follows the two review comments posted on this PR at 84ed73dfe: from those, the kernel answering every `needFull` (with its test) and guard 3 as the merge's own loss rule landed, and the items below did not.

- Guard 3 refuses a frame that legitimately supersedes a held run, and the page then keeps the old conversation with no visible sign. The loop at ui/webview/render.ts:17209-17211 reads every held run with `r.lo >= tailLo`, tail run included, and `splitHeldAgainstFrame` returns a run sharing no key with the frame whole as `before`, so a retracted run reads as uncarried history: a windowed rewind cut below the held tail run's lo is refused, and so is a whole frame (`headKnown`, `tailLo` 0, every whole list the kernel sends, kernel/kernel.py:50596) sharing only the constant `system:head` or an overlay card's kind-keyed uuid with a held run. Executed, a windowed rewind, a whole rewind and a clear-shaped whole frame are each refused once and latched on the identical re-send; the base e2fffe994 applied all three, and the head the comments reviewed applied both of these rewinds and refused only the clear-shaped frame, so widening guard 3 to the tail run introduced the rewind refusal (and the refusal of a held tail ending in an overlay card the frame carries). Whether a real /clear reaches this depends on how the session runs: a session the SDK backend runs keeps its id across a /clear, the registry's `lastSid` naming the new transcript (kernel.py:4776), so its whole frame arrives against the held state and is refused; a terminal-launched session's /clear mints a new transcript id (kernel.py:5484), so its frame arrives with no held state and guard 3 cannot fire. After the latch the tail freezes under a live status chip, a composer send stays a sending bubble, and the one trace is a client-diag row. The latch bounds upsert's re-ask only: the kernel's post-full empty-suffix `chatTail` anchors at a key the page never applied, so the gap ask at 17613 fires and is answered with a full every cycle, unbounded.

  The fix has two halves, and the order matters. First, show a refused and latched frame in the pane rather than the journal: the exemption below fixes the enumerated shapes, the signal covers the shapes the enumeration missed, and a guard that refuses visibly survives a wrong predicate where one that refuses silently does not. Key it on the `n === 2` event (the first ask is routine), say the view is behind rather than catching up (nothing is in flight), show it at the tail's end and on the tab and composer (a parked reader may lack the tail in the DOM), clear it in `clearRefusedLatch` and `dismissSession`, and keep it across a reconnect. Second, exempt a whole frame, `!(msg.headKnown && msg.tailLo === 0)` at 17196, on `headKnown` rather than `tailLo` alone; executed, both whole-frame shapes apply as at the base with the three loss controls still refused and both legitimate windows applying. The windowed rewind also needs the per-run rule (refuse a claimed run only when the frame shares a transcript-row key with it and held transcript rows sit before that key), both sides filtered to `txRow`, or an overlay card's constant kind key makes a rewind carrying one still refuse; that rule alone re-admits a `tailLo` below the held tail whose frame re-carries the tail from its first key, so pair it with the next item's clause. A clause treating a held run past the frame's claimed end as a legitimate drop must not land: it also applies `straddle_lying`, since a lying short frame and a windowed rewind have identical coordinates.

  The guard is wrong in both directions, too wide for honest frames and too narrow for dishonest ones (next item), so tuning the predicate alone does not close it; a visible refusal makes a wrong predicate survivable.

- A lying `tailLo` whose frame re-carries the whole held tail run passes guard 3, and the merge drops the mid gap silently. Over [gap, run 137-263, gap, tail from 475], a frame with `tailLo` 200 carrying the tail run's own events: guard 3 skips the read run (lo below `tailLo`), the tail run's `before` is empty, and the merge coalesces the read run onto the relabelled tail with the gap gone and the store count unchanged, so the lab's shrink invariant misses it. A real full frame carries the held tail's window whenever no turn was appended, and the base and the reviewed head do the same. Fix: refuse when the frame's first key is held in a run whose lo exceeds `tailLo` (a held event's turn index never falls), guarded by `first != null`, the tail run searched too; executed, lying same-window frames are refused once while the legitimate windows, an honest re-send and a slid frame apply; the clause leaves the rewind shapes above as they are at this head, so it composes with the exemption rather than replacing it.

- The kernel half of the no-tailLo item did not land. `_tail_run_start` at kernel.py:50648 resolves the sid through `_sessions(now)` alone, the 48 h discover horizon, and returns `tailLo` None for a sid `_session_row` resolves; the frame leaves with `tailLo` null (trimmed frames only) and the page takes 1877's regions-dropped road, every held run gone, one `regions-dropped` row on the first such frame and none after. The page road is the base's, so this is unlanded rather than a regression. Reaching it needs history loaded before the session crossed the horizon, since `_chat_history_reply` (50222) and `_chat_history_page` (49974) use the same lookup and fault past it; discover's fingerprint ignores appends, so a resumed session stays unresolved until a `names/` change. Fix: `sess = _session_row(sid, now)`, a superset of the lookup, memoized per pusher cycle, the `try/except` keeping the None fallback; over a real 72 h idle transcript it gives `tailLo` 75, the in-window answer. Route the two sibling lookups through it too, or gap paging still faults. One caveat: `_session_row` follows the registry's `lastSid` and `build_session`'s fallback the names anchor, so a cleared session outside the window could resolve to different files; a mismatch fails closed to None.

- The latch key still ends in the frame's last key (17221-17223), so a refused frame that grows one turn per re-send never latches: five growing frames, five asks; five identical, one. The item is narrower now that the kernel answers: a wrongly refused rewind costs two asks and two fulls per transcript change, none landing, rather than a freeze. A key on (why, `tailLo`, first key) is necessary and not sufficient, since the kernel's cut is `WIRE_TAIL` moved to a turn edge, so past 250 events the first key slides per change too and that key still costs 2+2. Also count a frame as a repeat when it answers the page's own ask (read the sid's own-ask record before the delete at 17216); alone, that gate treats an unsolicited re-send as new (three asks of five on an identical stream); combined, one ask and one full per change in every shape. No key lands content, so this is a cost bound, not a heal, and it bounds upsert's ask only, not `chatTail`'s; the `n === 2` loop row also refiles on every change.

## Smaller points

- render.ts:17605-17606 locates `tailStart` by a first-match key lookup. A key repeated as the last event of the last history run and the tail run's first lands `tailStart` in history; a delta anchored on it passes the numeric-`from` guard exactly, truncates to the history runs, and `regionsAbsorbTail` relabels: the tail run went from 250 events to 2 with no ask. Rare, since the kernel keys events uniquely within a build. Fix: `tailStart` as the count of the history runs' non-optimistic, non-held events, the anchor resolved in `kernelEvents.slice(tailStart)`; pin the repeated key placed last in the history run (a mid-run copy is already caught) with the real `regionsAbsorbTail` lifted.

- docs/reference.md:2956-2957 says an anchor the page lacks is a gap; guard 1 (render.ts:17604-17614) also refuses an anchor held only in a history run, which the base applied by truncation. Amend to: an anchor not held in the page's tail run is a gap (`needFull`).

ksarma:
What the suite pins and does not pin at 203bfe57c, the merged head, as a second note to the one above; line references are on that commit.

## Tests

- tests/test_relay_tail_vanish_served.py:314, `_assert_scenario` reads no `needFull` value: with guard 3 changed to refuse but never ask, the lab passes 9 of 9 and the node tier stays green (a plain deletion of the ask call is caught by a census; the escaping mutant keeps the text and gates the call). Add an `expect_asks` argument asserting `after.needFull - before.needFull`: 1 for the three loss scenarios, `anchor_in_history` and `from_into_history`; 0 for `equal_hi_legit`, `straddle_legit` and `not_proto2_rebase`. With the loop bound below tightened as well, the mutant reds exactly the three guard-3 scenarios and `loop`. Drop the dead `expect_read_drop` with the signature change.

- Line 377, `assertLessEqual(asks, 2)` is slack: only the `n === 1` branch asks, so a mutant asking again at `n === 2` passes; `assertEqual(asks, 1)` reds it, and the PR body's "at most two asks" should read exactly one.

- render.ts:17498, `clearRefusedLatch`: the cleared-count row, the desync row's re-arm and both call sites (17240, 17659) are observed by no test; with its body a no-op, every node test and lab scenario passes (it runs in the lab when a frame applies, but over an empty latch). Add one scenario inside one page life (`park()` reloads per scenario): refuse twice, progress by an honest full re-carrying the tail run from its first key (else guard 3 refuses it) or a delta anchored at the tail run's last uuid, refuse again; assert asks 1 then 2, a loop row with `cleared` true and the accumulated count, two desync rows; the WS hook must keep `data.count` and `data.cleared`. As a control it passed at this head and failed under the mutation. The delta call site also needs ui/webview/chat-exact-tail-exec.test.ts:57 to hand `chatTail` a real latch Map and the lifted function.

- ui/webview/chat-proto2-exec.test.ts:62, `liftUpsert` names no `sharesAnyUuid`, which `liftWith` resolves to a function returning undefined, so guard 3, the latch and the desync rows execute in no node test (an inert guard or latch passes every one). CI's served lab under chromium does catch an inert guard 3, so the gap is the node tier plus the rows, which no tier pins. Add a separate `liftUpsertGuards` scope lifting `sharesAnyUuid`, `upsert` and `clearRefusedLatch` together, with a real latch Map, a Set for `emptyFrameDiagSent`, and recording `requestFullSession(id, why)` and `vscodeApi`; pin the three loss shapes (one ask and a row each), the identical repeat (no ask, loop row count 2, a third silent), new coordinates asking again, the legitimate windows (no ask), and the clear (the accumulated count, 3 after three identical refusals, `cleared` true, the desync row re-armed). Such a harness passes at this head and goes red under an inert guard, an inert latch and a deleted ask.

- Nothing pins that a /clear or rewind whole frame passes guard 3, and over a reader holding a head run with `system:head`, a gap and a tail run those frames are refused at this head, so the cases are red until the exemption lands; a single-run reader applies them. Write them over both shapes, assert no desync row rather than no rows (a legitimate rewind files a frame-behind row), and keep the three loss controls refused beside them.

- The lab's docstring (lines 8 to 23) describes a deleted guard 2 (`no_taillo`, `not_proto2`), "Four such shapes" against nine in `SCENARIOS`, guard 3 keyed on a run's hi (`equal_hi_legit` applies at `tailLo` equal to the hi), "Red at main" (7 of 9 red at e2fffe994; `equal_hi_legit` and `straddle_legit` pass there) and "asks for one it can" (`not_proto2_rebase` re-bases with a `regions-dropped` row and no ask, store 536 to 0, so the shrink invariant must exempt it). `import re` (28), `BIN` (38), `expect_read_drop` (314) and the `midGap` field (87) are dead. Both skip texts (176, 284) say CI installs no browser; the vscode-extension job installs chromium and runs this lab under `ROMP_SERVED_TESTS_REQUIRE=1`.

PR #1861, opened 2026-09-19
romp card names its board: -b/-c, the resolver last in post_notice, the row and the card carrying board and category (card boards phase three, the producer's half)

henrypinkard:
Tier: feature. Filed without auto-merge for the maintainers' read. Card boards phase three, the producer's half (plans/card-boards.md; the first half, the store and the pure resolver, landed in #1845): `romp card` names the board a card files on.

**The design first (the first commit).** plans/notice-cards.md, "The card command names its board": the two flags ride the `/notice` body; `post_notice` resolves where the card files through the pure resolver AFTER every other check, so a refused post never leaves a board behind, and writes a first-use board (or a data board with a category appended) right before the row appends; the row stores `board` and `category`; the card builder copies them and keeps the phase-two mapping for a row without them; `--needs-you` is the board's needs-you category or the resolver's refusal; the posted line names the board and says when the post created it.

**What lands.**
- Command (bin/romp): `-b <id>` / `--board <id>` and `-c <id>` / `--category <id>`, each on the body as `board` and `category`, both absent when not given (an older kernel files the card on the feed as today); `-c` alone names a feed category. The posted line reads `romp card: posted (key fig, rev 1) on board figures/new (board figures created with category new)`, `(category kept added to board figures)` when the post appended one, and nothing extra for a known pair, so a typo in `-b` is a new board seen at once (`romp board list` names it, `romp board remove` takes it away while no card stands on it). The usage line, the help line and docs/reference.md's entry name the flags.
- Kernel: `post_notice(..., board="", category="")` calls `_board_resolve_post(board, category, needs_you=..., producer=..., key=...)` after the key, the caps, the producer, the session or the owner-less road, the expiry, the actions and the attachment have passed; the resolver's error is the post's refusal, word for word; a `pending` definition goes through `define_board` right before the append (a write fault refuses the post); the row carries `board` and `category` as resolved; the answer carries `created` (`"board"` or `"category"`) for the command's line and is never stored. `POST /notice` hands the two members through. `_notice_cards` copies a row's fields and reads a fieldless row the phase-two way (`feed`, `needs_input` when needsYou else `completed`), the same read `_notice_standing_count` makes; a card on a data board keeps the needsYou mapping as its feed `column` while the pane files it under the feed's default column (Working) by its category, until phase four's view switch.

**Tests, red first at the design commit 3c59195b with the test files copied in.**
- bats (tests/romp.bats, one card test): the flags ride the body and their long forms; the posted line names the board and the created word; `-c` alone rides `category` and no `board`; neither flag rides nothing; a flag with no value or another flag as its value is a usage error. Red at the base: the flags were unknown (`[ "$status" -eq 0 ]' failed`, usage exit 2).
- Python (tests/test_notice_card_store.py, the Boards class, five tests): a first-use board created with the named category, the row and the card carrying the fields, the standing count seeing them; a category appended to a data board in neutral dress, the rules, the badge category, the default; every refusal in the resolver's words and no board file after a post refused by the resolver or by an earlier check (a bad key, a refused attachment, an unknown session); a row without the fields read the phase-two way; the route handing the members through, the owner-less road on a board too. Red at the base: `TypeError: post_notice() got an unexpected keyword argument 'board'` (three), `KeyError: 'board'` (two).
- The served lab (tests/test_notice_cards_browser.py, one road): a card posted with a board answers with the resolved pair and the created word, and the pane names the board beside the producer (`via cli · on Figures`) under the feed's default column. Red at the base: `KeyError: 'board'` (the answer carried no board).

**Round two (the read of 968f41c3; head b0dccfb1).** MEDIUM, one rule for "needs you": this branch let a notice card's `category` and `column` disagree for the first time, and the kernel's readers were split between the two fields. `romp card -t x -c needs_input` (no `--needs-you`) left `needsYou` false and the column `completed`, so the pane filed the card under Blocked (askColumn reads the category) and the app badge counted it (`_needs_you_count` reads the category) while the session's yellow ask ring stayed off (`_needs_input_sids`, and through it `build_session`'s `needsYou`, read the column); the mirror on a data board with a hot needs-you category (`-b urgent --needs-you`) lit the ring and the badge while the pane showed the card under Working and a tag lens dropped it (the lens read `isNeedsYou` against the feed board). Now ONE predicate, `_card_needs_you(card)`: the card sits in its OWN board's needs-you category (`_board_needs_you`; `category`, an older frame's `column`), read by the badge and by the needs-input sids (and so by `build_session`); the pane's lens reads `isNeedsYou(boardOf(card), category)`. The post makes the flag and the category agree by construction: a card resolved into its board's needs-you category IS a needs-you card (`-c needs_input` alone files one), and `--needs-you` beside a category that is not that one is refused by name (drop one of the two). A feed card's `column` IS its `category`, always; a card on another board keeps the needs-you mapping as its feed column, the column an older pane files it under, until phase four's view switch. plans/notice-cards.md states the rule. LOW: the posted line for an owner-less card on a board says both facts, `on board figures/new, with no session`.

Red first at 968f41c3 with the files copied in: the store (the feed's `-c needs_input` post carries `needsYou` true and its card's column `needs_input`; the hot board's `--needs-you` post files under `hot`; the badge count and the needs-input sids agree with the predicate across four cards; the two contradictions refused by name; a fieldless row read the same on every side): red on `(False, 'feed', 'needs_input') != (True, 'feed', 'needs_input')` (the feed post's `needsYou` false); the harness (a data board's needs-you card stays on the board under a tag lens its session is outside of): red on `the needs-you card of the same session stays: its OWN board's badge category passes every lens` (the card was dropped); bats (the owner-less board line): red on `[[ "$output" == *"on board figures/new, with no session: it shows until"* ]]' failed` (the note missing).

**Verified locally (through capped) at b0dccfb1.** The whole Python suite (uv run --python 3.13, no xdist, REQUIRE=1): 14248 passed, 62 skipped, 8 failed, every one this box's and none reading notices, boards or the command (tests/test_kernel_webpush.py, three, and tests/test_notification_tap_resume_browser.py, three, on the python cryptography package this box lacks, which REQUIRE turns from skips into reds; tests/test_pane_hidden_word_browser.py's firefox and webkit legs on browsers this box lacks and CI installs none of); the whole node glob: 5055 tests, 5049 passed, 0 failed, 6 skipped, tsc clean; before the commit, the feed, federation and board-def pane files 675 passed, the bats card and board tests 6 ok under the CI-pinned bats, the store, marker, card-boards, served-lab and kernel-trust modules 112 passed; merge-tree against main a29219fa clean.

**Verified locally (through capped) at 968f41c3.** The whole Python suite (uv run --python 3.13, no xdist, REQUIRE=1): 14247 passed, 62 skipped, 8 failed, every one this box's and none reading notices, boards or the command (tests/test_kernel_webpush.py, three, and tests/test_notification_tap_resume_browser.py, three, on the python cryptography package this box lacks, which REQUIRE turns from skips into reds; tests/test_pane_hidden_word_browser.py's firefox and webkit legs on browsers this box lacks and CI installs none of); the whole node glob: 5054 tests, 5048 passed, 0 failed, 6 skipped, tsc clean (this branch touches no TypeScript); before the commit, the bats card and board tests 6 ok under the CI-pinned bats, the store, marker, card-boards modules and the served notice lab 75 passed; merge-tree against main 6dfd6df1 clean.


ksarma:
A note after the merge, executed against 968f41c38. b0dccfb13 adds a needs-you check above `created = None` and a `column` line in `_notice_cards`; the listed lines are unchanged (define and append seven lines later, the cap check twelve). Standing as executed:

- `kernel/kernel.py` 24086 to 24105: the resolve, `define_board(pending)` and the append run outside `_boards_lock`. Threaded probes of 300: a `romp board define` between resolve and write lost (up to 299); two posts appending categories leaving a row under a category the file dropped (30 to 53); a `remove_board` before the append leaving a row on an undefined board (104). Fix: `_boards_lock` from `_board_resolve_post` through `_notice_append`, `define_board` split over an unlocked variant; `define_board` already takes boards then notice via `_notice_standing_count`: no deadlock. Stopping at the write left 4 to 53 orphans; through the append, 0.

- 24091: a board file `_boards_data` skips (unknown member, id mismatch, truncated JSON) leaves `prior_def` None, so the post writes the defaults over it, leaving standing cards under categories the new file lacks. Fix: on the first-use branch only, refuse when `_board_path(board_id).exists()`, naming the file; `romp board define` stays the repair road.

- 24093: the define runs before the `_notice_lock` block, so a post the archive read or append refuses keeps the board or appended category (chmod on `notices-archive` or `notices`, both `created` shapes); `tests/test_notice_card_store.py` 877 has no `derr` case (dropping the check left 63 green; a regular file at `km._board_dir()` catches it). Moving it inside `_notice_lock` deadlocks; an unlink misses the appended shape. Fix: save `prior_def`; after the lock block, any refusal removes the board or restores what was saved.

- 54226: a ninth category on a data board is refused with the schema's cap sentence, naming neither board nor category, `bin/romp card` included. Fix: in the append branch, wrap the checker's error with both names, sharing `_board_check`'s cap constant.

- `docs/reference.md` 129: the `POST /notice` shape omits `board` and `category` (`actions`, `dismissOnAction` and `t` were already absent: an existing gap widened). Add both.

Changed by b0dccfb13: a feed card's `column` came from the needsYou mapping, not its category, splitting `_needs_input_sids` from `_needs_you_count`; it sets `column = category_id` on the feed and refuses `--needs-you` beside another category by name.

Scratch copies at 968f41c38 verified the widened lock, the `exists()` refusal and the column fix (`tests/test_notice_card_store.py`, `tests/test_card_boards.py` green), not the rollback or the wrapped error; nothing ran against b0dccfb13.

PR #1867, opened 2026-09-19
Codex sessions mail through the six postal tools as Codex dynamic tools the kernel services, so no credential enters the sandbox and Sandboxed and Auto mail alike

lczh:
Tier: feature

## The defect

A Codex session has no working way to reach the postal bus, and the one way it ever did reach it was the widest possible one. The bus and the kernel accept exactly one credential, the kernel's serve token, a 0600 file under the romp state directory, and the `romp_workspace` sandbox profile leaves that directory unmounted on purpose (docs/codex.md, Sandboxing). So `romp mail send` from inside a Codex sandbox fails three independent ways: no token (every kernel route answers 403, and the service's own loader either faults at import or mints a token the bus never issued), no identity (the Codex registry the shell would read to learn its own session id lives in the same unmounted directory), and no reachability (the `romp` command resolves into a checkout under the user's home, outside both the workspace and the minimal runtime set). What let mail out of some threads was escalation: in Auto mode the command ran outside the sandbox as the kernel's user with full access once Codex's automatic reviewer allowed it, and the reviewer ruled differently for a fresh thread. In Sandboxed mode mail was simply impossible. Mail from Codex was therefore per-thread luck, silent when refused, and when it worked it traded the whole sandbox for one send.

## The fix

Every Codex thread now registers the six postal tools (`send_message`, `check_inbox`, `list_agents`, `set_working`, `check_sent`, `recall_message`) as Codex dynamic tools on `thread/start` and again on `thread/resume`, together with the postal instructions as the thread's developer instructions. The app-server routes each call back to the backend as the `item/tool/call` server request; the backend answers it before its declined-request path (so no warn toast), binds the call to the session by the request's thread id only (a thread no live session holds is refused with a retry sentence, never resolved by name or by the only live session), forwards only the arguments the tool's schema declares (a `from_id` or `id` a model puts in the arguments is dropped), copies the session's id and name out under its locks and releases every lock before calling the kernel's callable, and never raises (a fault is a failed result carrying the exception text: the handler runs inline on the SDK's single reader thread, where a raise would end the transport for every Codex session). The kernel's callable dials the bus over loopback with the serve token as that session, bounded at 3 seconds, replicating the client-side checks the bus's own MCP tool makes (to and body required, a kind outside delegate/coordinate/question refused rather than downgraded, tracked a JSON boolean and a delegate's only), echoing the bus's note when present, and returning the bus's delivered sentences; a written-but-unanswered send is reported as possibly delivered with a pointer at `check_sent`, so a delegate is not sent twice. `set_working` writes the kernel's own store directly. The normalizer maps `dynamicToolCall` items to the `tool_use`/`tool_result` pair an `mcpToolCall` gets, named `mcp__romp-postal-service__<tool>` (the item's namespace when set), so a Codex session's send shows in chat, the timeline and the judges' view exactly as a Claude session's does. The bus's push banner names the `send_message` tool in its reply hint for a Codex recipient (the bus's agent rows now carry the kernel row's backend field, which is what the banner keys on), and `romp mail send` from a Codex shell now points at the tool in its refusal. The registration sits behind one setting, default on: the bare value file `codex-postal-tools` in the state directory, `off` to disable, read where the backend is built, so it applies at the next kernel start; the backend logs the off decision once. No credential enters the sandbox, nothing is written under `~/.codex`, no new bus authentication exists, and Sandboxed and Auto behave identically.

Live-probed before building, on runtime 0.153.3 through the pinned 0.144.4 client: `thread/start` accepts `dynamicTools` (a spec without a description is refused with -32600), the call arrives as `item/tool/call` with threadId/turnId/callId/namespace/tool/arguments, the `{success, contentItems}` reply is accepted, `item/started` and `item/completed` of type `dynamicToolCall` bracket it on the turn's stream (3 ms before the request, 6 ms after the reply; the whole probe turn took 4.2 s), no approval or reviewer step touches it in either mode, `dynamicTools` on `turn/start` is accepted and ignored, and the registration persists server-side across a resume. `developerInstructions` is honored at `thread/start` and persists as the thread's first developer message (replayed on every resume and across kernel restarts); a value passed at `thread/resume` is accepted and ignored, so an existing thread's instructions cannot be revised through resume, and a changed text reaches only threads started after it. The wire-shape half of the probe is now in `tests/smoke_codex_live.py` (the tools registered, the call bound to the calling session, the reply accepted, no warn, the transcript pair, the registration surviving a resume), runnable under the danger-full-access override because the call never enters the sandbox; the sandbox negative (a command cannot read a file outside the workspace) runs only under the shipped profile and prints SKIPPED otherwise.

## Design points

Four ways were on the table. The one built, (c), registers the tools with Codex and services the calls in the kernel: the kernel already knows the session's id and holds the token, the app-server routes the call to it, and the model sees a tool, not a door. Exposure equals a Claude session's today.

Option (a), adding the serve token (or the state directory) to the profile's runtime reads, was rejected because the token is full control as the user: kernel routes to inject text into any session, spawn or end sessions, restart the kernel, push to attached machines, switch stored logins, and stream every transcript; on the bus, mail as any session, read and clear any inbox, rewrite any working note. Every command the model ran in any Codex thread would hold that, which erases the sandbox's stated purpose and contradicts the profile's own docstring and the documented design. It was also the smallest change, which is why it is named here rather than left implicit.

Option (b), a per-session postal token minted at thread start and accepted by the bus for postal routes only, was rejected as the wrong tier and the wrong shape for this defect: a new credential kind (mint, store hashed, rotate on session end), a second authentication branch on a daemon that has had one credential its whole life (a mistake there is a silent widening), CLI import-order surgery so the service stops faulting when the serve token is unreadable, delivery through an unverified Codex environment knob, and it would still need the checkout's `bin/` and `postal/` inside the sandbox. That is a new contract on the bus, a major feature, and it would only make the shell path work, which the tool path makes unnecessary. If a maintainer wants the sandboxed shell path to work, that is this option and should be filed and tiered separately.

Within (c), the MCP-server variants were passed over: a server in `~/.codex/config.toml` cannot tell which thread it serves (Codex hands no thread id to MCP children), so it could not sign mail as the right session, and it would rewrite the user's global Codex config; a per-thread server through thread config would cost one Python process per thread and a new bus identity variable, and is the fallback had the probe failed. The tool table is a KEEP-IN-SYNC copy of the bus's, not a shared module: the bus is its own process and neither side imports the other; a test loads the bus by path and pins the copy equal, name for name and schema for schema, and pins the kernel's copied delivered sentences equal to the bus tool's.

## Review round

Folded from the maintainer's review as one commit on top of the reviewed head, in the review's order. (1) The create branch of `_prepare_thread` (a pending- or failed- row turned real on its first turn) has its own registration pin beside the registration test: the last thread/start carries the spawn's own tool list and the instructions; red with that branch's merge removed (no `dynamicTools` at all). (2) The callback census in `tests/test_stage_marks.py` descends a conditional hand-off (an IfExp's body and orelse, nested ones too, a None branch no row), proven on a SNIPPET, and `_codex_postal_call` is listed in `CALLBACK_ALLOW` as loopback HTTP to the bus plus a working-note write; with the allow entry and the SNIPPET but not the descent, the stale-entry pin and the SNIPPET test were both red, as the review predicted. (3) A failed call leaves one line on the kernel log naming the tool, the session and the cause: the two bus faults (unreachable, written but unanswered) once per fault spell and re-armed by the bus's next answer of any status, after the interrupt-marks memo's latch; a store raise every time; the write wrapped so a stderr that cannot be written leaves the per-fault sentence intact; nothing re-raises. (4) `_bounce_oversize` takes the banner `_push` judged the message with, so an oversize bounce to a Codex recipient reports the size as delivered to it; red before on a message sitting exactly on the cap as a shell body (the note named a size under the very limit it had crossed). (5) The fault branches of `check_inbox` (the written-but-unanswered warning that mail may now count as read), `list_agents`, `check_sent` and `recall_message` are pinned, a written request told from an unwritten one; the review's collapse of those branches turns them red. (6) `_codex_recall_text` is pinned equal to the bus tool's recall branch over six shapes, the empty pair and a carried kept row without its own why included, every removed row carrying both `to` and `body`. (7) The sandbox refusal's reason in `docs/codex.md` reads as suggested: the identity lookup and the serve token both live outside the mounts, and a workspace without a checkout has no command at all. (8) The docstring and the test comment agree with the doc now: a resume cannot add a registration (`ThreadResumeParams` has no dynamic_tools at rust-v0.153.3, read for the review), so a thread started without the tools stays without them for its life; the doc gained the off half (an existing thread keeps its persisted tools and instructions until it is ended, each call answered that mail is not available).

### Second round

Folded from the maintainer's second review as one more commit on top, in its order. (1) The two kernel-log tests were order-dependent: `TheKernelLogNamesAFault` captures all of stderr, and the bus-port dial's census line (`_bus_port_census`) lands there on a process's first dial, so each test alone was red (the count read 2 lines, not 1; with stderr dead, the census write raised before the request and the session got the generic sentence) while the module in file order was green. One `_bus_port()` call in setUp says the census before any capture; the class alone was 1 failed, 1 passed before and is green after. (2) The same unwrapped census write reached production: a port or source change after boot makes a Codex call's dial say the census, and with stderr unwritable the raise reached `_codex_postal_call`'s catch-all, so the session got the generic sentence, not the per-fault one, and the bus was never dialed. The write is wrapped as the log's is; the dial stays where it was (inside the request's try it would misfile the cause). Pinned by resetting the census's memory before the dead-stderr capture: red before on the generic sentence, green wrapped. (3) The shell recipient's bounce measurement was unasserted: the test's on-the-cap message rode the wake and never reached `_bounce_oversize`, so measuring with the tool banner unconditionally stayed green. The test now pushes to the shell row a message from the Codex session (a sender equal to the recipient gets no note), one byte over as a shell body, and asserts the count in the sender's note equals `_deliver_body_bytes` for that recipient; under the unconditional tool-banner measurement it reads 2 more (921603 != 921601). (4) One latch slot served two causes: `_CODEX_POSTAL_SAID` was one boolean for both bus faults, re-armed only by an answer, so a spell whose cause changed with no answer between logged only the first (refuse then hang gave one line). It is a set keyed by cause now (`unreachable`, `no-answer`), cleared on an answer of any status, so such a spell logs one line per cause; the refuse-then-hang case expecting two lines was red before (1 != 2).

### Third round

Folded from the maintainer's third review as one more commit on top. (1) The per-cause test's clause that an answer ends the spell for both causes pinned only the unreachable cause's re-arm, so a kernel that clears only that cause on an answer passed the module. After the answering call the test now runs a refusal and then a hang under the capture and asserts one no-answer line beside the re-said refusal; under a kernel mutated to discard only the unreachable cause on an answer it is red (1 != 2) while the class's other three tests stay green, which is the gap the review named. (2) The log line's suffix said the fault was said once per fault spell while the latch is per cause, so a refuse-then-hang spell wrote two lines each claiming to be the spell's one; it now says once per cause per fault spell, the latch's comment and the docstring say the same, and the per-cause test asserts the phrase on every captured line (red on the stock suffix before the change).

## Tests

Red before, each against the stock tree, on its own assertion. In `tests/test_codex_backend.py`: the registration test fails on the constructor refusing the `postal` keyword (the one test that needs the keyword); the four handler tests assign the callable after construction, a plain attribute write the stock backend accepts, and fail on their asserted behavior: the call bound to its thread fails on the fake having seen nothing (`[] != [('send_message', ...)]`), the lock-order test and the never-raises test on the stock handler's `{}` carrying no `success` key, the unknown-tool test on the answer's key set not being `{success, contentItems}`; the no-hook test fails on the said-once log line (`0 != 1`). In `tests/test_codex_events_golden.py` the normalizer emitted nothing for a `dynamicToolCall` item (`[] != ['assistant', 'user']`). In the new `tests/test_codex_postal_tools.py` the kernel had no `_codex_postal_call`, the backend had no spec table or instructions to pin against the bus, `format_push` took no reply-tool flag, and the Codex refusal of `romp mail send` did not name the tool; with only the bus's agent-row line reverted (the field the banner keys on not carried), the push test fails on the banner's last line naming the shell command, which is the production defect the first cut's hand-built row hid.

Green after (counts at the review-round head, with main merged in): `test_codex_backend` 126 passed (the create-branch registration pin added) (the profile test additionally pins that the serve token and the registry stay out of the profile and that the config does not override the SDK's experimental-API default); `test_codex_events_golden` 40 (the pair rendered and chained, the failed and namespaced variants; 31 of them at the reviewed head, the rest main's); `test_codex_postal_tools` 26 (the send payload and sentences, client-side refusals without a request, a bus refusal echoed, the unanswered send said as maybe, the inbox naming the tool, agents, receipts, recall, `set_working` in the kernel's store without the bus, the wiring into the backend, the setting file turning the registration off, the push banner driven through the bus's real row producer, the CLI refusal, and the parity class: the tool specs and instructions against the bus module loaded by path, the send sentences and the `/send` wire payload against the bus tool's own HTTP stub, one synthetic receipts list through every branch of both renderers with the refusal gate copied, the agents renderer including the empty sentence, the inbox renderer equal on every line but the reply hint, and the recall renderer equal on every shape; plus, from the review round, the kernel log line and its once-per-spell latch, the fault branches of the four read tools, and the oversize bounce measured with the recipient's banner; and, from the second round, the census line said before any capture, the census write's wrap pinned through a re-created first dial on a dead stderr, the per-cause latch on a refuse-then-hang spell, and the shell recipient's own bounce measured with its banner). `test_stage_marks` 32 (the conditional hand-off SNIPPET and the postal row listed). Unchanged and green: `test_codex_runtime` 11, `test_backend_call_parity` 4, `test_postal_live_only` 26, `test_postal_delivery` 22, `test_postal_peers` 97, `test_stat_rule_readers` 15 (4 subtests; its source-pin counts over `_push` and `_bounce_oversize` unmoved), `test_injected_voice` 10 (222 subtests), `test_kernel` 460, `test_kernel_trust` 36 (it pins the census memory and the census line's text, both unchanged by the wrap), `test_state_dir_override` 7 (it runs the smoke by path), `test_postal_agents_render` 8, `test_postal_freshness` 6, `test_postal_self_send` 28, `test_feed_session_started` 21, `test_perf_stats` 62, `test_feed_memo_inputs` 31, `test_chat_build_sig_inputs` 65, `test_postal_deferred_push_identity` 7, `test_postal_stranded_putback` 16, `test_postal_sender_identity` 6, `test_state_isolation_order` 8; pin-scan 0 lost of 8498 literals.

## Open decisions for the maintainer

- Tools, not shell: the contract built is that a Codex session mails only through kernel-serviced tool calls, and `romp mail` from a sandboxed shell stays refused with its reason. Confirm; the alternative is option (b) as its own major feature.
- The setting's surface: the registration sits behind `codex-postal-tools` (`off`; default on), a bare value file in the state directory read at kernel start, the shape the judge-tier settings use. Whether it should also have a gear row, live without a restart, or a different key is yours; the seam is one constructor argument.
- Transcript naming: rendered as `mcp__romp-postal-service__<tool>` for exact parity with a Claude session's postal calls in chat, timeline and judges (the item's namespace wins when a spec sets one). A distinct prefix is a one-constant change if you would rather tell the backends apart.
- Developer instructions: the postal instructions ride `developerInstructions`, the bus's text minus its Claude-Code paragraph. Live-probed: honored at `thread/start` and persisted with the thread; accepted and ignored at `thread/resume`, so a thread started before a wording change keeps the old text for its life. If you would rather rely on the tool descriptions alone, dropping it is one key.
- The reader-thread budget: a 3 s bounded bus call inside the SDK's reader loop stalls every Codex session's notifications and interrupt replies for that long on a wedged bus. Accept, or ask for a worker hand-off design the pinned synchronous handler does not directly support.
- `mcp_status` is unchanged (the postal tools are not MCP servers); whether the dashboard should list the six tools there is yours.


ksarma:
The evidence was executed against 27828243d. The current head 2af116491 is that commit with main merged in and the change unchanged (same 10 files, +1335/-34, and the same added and removed lines in kernel/codex_events.py, the merge's one conflict), so the findings apply to it as written. Lines are on 27828243d; the kernel/kernel.py anchors sit lower at 2af116491 (the hand-off at 18815, the postal helpers 3 lines down).

## At this head

**The resume create path merges the postal parameters with no assertion.** The registration test (tests/test_codex_backend.py:681) pins two of the three sites that merge `_postal_params()` into thread params; the create branch of `_prepare_thread`, a thread/start for a pending- or failed- row (kernel/codex_backend.py:1699), has none. Executed against 27828243d, this reproduces: with that merge removed from the create branch only, the three changed test modules stay green (176 passed, 5 subtests), and a probe forcing the create path fails on the mutant with `AssertionError: None != [...]` and passes on the original. Fix: a case beside the registration test that sets `s.tid = "pending-%s" % sid[:8]` and `s.loaded = False` (a failed- row takes the same branch, so one pending- case covers both), calls `be._prepare_thread(s, fake)` directly (the pattern at lines 2674, 2705, 2759 and 2809), and asserts the last thread_start call's params carry `dynamicTools` equal to the spawn's own list (not a literal, so the parity test in tests/test_codex_postal_tools.py stays the spec text's only pin) and `developerInstructions` equal to `cb.POSTAL_INSTRUCTIONS`.

**The postal hand-off is invisible to the callback census.** kernel/kernel.py:18773 passes `postal=(_codex_postal_call if _codex_postal_tools_on() else None)`, and `callback_census` in tests/test_stage_marks.py skips any value that is not a Name, Attribute or Lambda, so the IfExp yields no row. Executed against 27828243d, this reproduces: the census reports 14 rows and no postal row, so the marked-or-listed test (tests/test_stage_marks.py:395) passes without seeing `_codex_postal_call`, a kernel callable the backend runs inline on the SDK's reader thread (kernel/codex_backend.py:975-1017), the case the `CALLBACK_ALLOW` comment names; written census-readably, the row appears and fails as unmarked. No conditional hand-off at the base commit named a module-level def, so the gap had no consequence until now. Fix, keeping the PR's design: teach `callback_census` to descend an IfExp's body and orelse (ignoring a None constant), as the Thread census's `_names_in` does for conditional targets, with a SNIPPET test, and add `("CodexBackend", "postal", "_codex_postal_call")` to `CALLBACK_ALLOW` with the reason "loopback HTTP to the bus and a working-note file write; no session parse, build or hydrate". The other shape, passing the callable unconditionally, has two variants: a separate on/off boolean the backend reads works but needs a backend change and the test at tests/test_codex_postal_tools.py:270 re-aimed; a callable that refuses when off conflicts with the design, since `_postal_tools(None)` registers no tools on purpose rather than a tool that answers nothing. Whichever shape lands, the census change and the allow entry go together; either alone turns one of the two pins red.

## Smaller points

- **Bus faults leave no kernel log line** (kernel/kernel.py:4184 and the two fault returns of `_codex_postal_http`, the except clauses at 4103 and 4108). Executed against 27828243d: a refused connection, a timeout and a working-note store raising OSError(28) each came back from `_codex_postal_call` as `(False, <sentence>)` with empty stderr; the backend's log at kernel/codex_backend.py:1020 fires only on a raise, which this callable never does. The text does reach that session's transcript as an is_error tool_result, visible per call in its chat; the kernel log, the one cross-session surface, carries nothing. Fix: one stderr line in the kernel's voice naming the tool, sid and cause at those three sites, the two bus cases latched once per fault spell and re-armed on the next success, after the kernel's `_INTR_MARKS_WRITE_SAID` and `_RELAY_SAID` precedents, the store-raise case logged every time (it is not a spell), the write wrapped so an ENOSPC cannot swallow the sentence, with a test over the refuse and hang fakes and a patched `_set_working_note` asserting the line and the once-per-spell latch. Not a re-raise: the backend's generic sentence would replace the per-fault sentences the parity tests pin and log an unlatched traceback for the whole outage.

- **An oversize bounce to a Codex recipient reports the shell banner's size** (postal/postal_service.py:2445). `_bounce_oversize` measures without `reply_tool` while `_push` judged the message with the tool banner through `_push_chunks(sid, msgs, reply_tool)`; the two bodies differ by 2 wire bytes (json.dumps escapes the shell hint's two double quotes, absorbing 2 of the raw 4), so every oversize bounce to a Codex recipient under-reports the delivered size by 2 bytes. Executed against 27828243d with the real `_agent_rows` and a Codex row: a message 1 byte over the cap as delivered was bounced with a note reading 921599 bytes, under the 921600 limit; no existing test drives the path. Fix: a `reply_tool` parameter on `_bounce_oversize` (default False, so the two-argument calls at tests/test_postal_peers.py:1179 and 1185 keep the shell banner), passed from `_push`, and a case in `test_push_picks_the_hint_from_the_recipients_backend` sized so only the tool-banner body crosses `_PUSH_MAX_BYTES`, asserting no /deliver post for the Codex row, a note count equal to `_deliver_body_bytes(SID, [m], reply_tool=True)` and over the cap (red before the fix), and the shell row still measured with the shell banner, with `_name_for_id` stubbed to capture the note since the test's stubs leave the real deliver path live.

- **Fault branches of four tools are unpinned** (tests/test_codex_postal_tools.py:157). The `check_inbox` written-but-unanswered branch (kernel/kernel.py:4149-4151), whose sentence warns that mail may now count as read without having been shown, has no test, nor do the timeout and unreachable branches of `list_agents`, `check_sent` and `recall_message`. Executed against 27828243d: collapsing status -1 into 0 for check_inbox, or hang and refuse into 200 and `{}` for the other three, leaves the module at 20 passed. Fix: `bus({'hang': True})` in the inbox test asserting ok False and "count as read" present, and `bus({'refuse': True})` asserting it absent (redundant, since the parity class already pins that sentence, but harmless); for the other three, a hang case asserting ok False, the request written and "No answer", and a refuse case asserting `log == []` and "could not be reached", pinning the written-vs-unwritten distinction `_codex_postal_http` draws.

- **`_codex_recall_text` has no parity pin** (tests/test_codex_postal_tools.py:282). ParityWithTheBus pins the other renderers and sentences against the bus module; the kernel's KEEP-IN-SYNC copy of the bus tool's recall branch (postal/postal_service.py:5600-5612) has only substring asserts at lines 228-241. Executed against 27828243d: the copy is equal today (0 diffs over five shapes), but with the bus's three recall sentences mutated the module stays at 20 passed; with the pin added the same mutation gives 1 failed. Fix: for each synthetic `{removed, kept}` shape, stub `pm._http` and assert `km._codex_recall_text(removed, kept)` equals `pm._mcp_call('recall_message', {'to': 'web'})[0]`, including the empty pair's "Nothing to recall" sentence; include a carried kept row without its own `why` (the only path through the kernel's inlined WHY_CARRIED literal, since the bus server fills why itself) and give every removed row both `to` and `body` (the bus tool indexes them directly where the kernel uses `.get`, so a missing key raises on the bus side instead of yielding a comparable string).

- **The sandbox refusal's stated reason** (docs/codex.md:141-143). The refusal sentence is reached from a sandboxed shell when the workspace contains a romp checkout: executed against 27828243d under the sandbox's bwrap filesystem and uid conditions (Codex 0.153.3), the CLI minted a throwaway token on the tmpfs, `/ping` passed token-exempt, the identity lookup failed on the unmounted registry, and the printed refusal and pointer appeared. The code and the CLI parenthetical (postal/postal_service.py:5730-5731) stand; the doc's "no token" does not: the shell fails on the identity lookup (registry and serve-token state live outside the mounts; the minted token is harmless and discarded with the sandbox), and in a workspace without a checkout the command is absent rather than refused. Suggested: "refused inside the sandbox (its identity lookup and the serve token both live outside the mounts) or, in a workspace without a romp checkout, not there at all; either way a Codex session mails through the tools."

- **Doc and docstring disagree on what a resume adds** (docs/codex.md:151; kernel/codex_backend.py:195; tests/test_codex_backend.py:696). The doc says a change reaches only threads started after it; the docstring and the test comment say a thread first started without the tools gets them from a later resume. Read against openai/codex at rust-v0.153.3: `ThreadResumeParams` has no dynamic_tools field, resume_thread_with_history builds StartThreadOptions with dynamic_tools empty, and session/mod.rs falls back to the persisted rollout session_meta, so a resume cannot add a registration; the smoke test's resume leg shows persistence, not addition. Fix: the doc is right; rewrite the docstring and the test comment to say a thread started without the tools stays without them for its life, naming the runtime tag read; add one clause to the doc for the off half (an existing thread keeps its persisted tools and instructions until it is ended, each call answered that mail is not available).

## Tests

The modules executed at 27828243d, their counts, the fails-before and mutation runs behind the points above, and the live claims not executed are in a second comment.

ksarma:
Modules executed at 27828243d, with results.

- The three changed modules: test_codex_backend 125 passed, test_codex_events_golden 31, test_codex_postal_tools 20 (176 passed, 5 subtests), matching the PR body's counts.

- Neighbouring modules at the same head: 526 passed, 245 subtests, 0 failed across those three plus test_stage_marks, test_postal_sender_identity, test_postal_stranded_putback, test_postal_cards_served, test_comment_outline_served, test_postal_delivery, test_state_isolation_order, test_state_dir_override, test_postal_sdk_agents, test_postal_freshness, test_postal_agents_render, test_codex_echo_merge, test_codex_models_route, test_injected_voice, test_backend_call_parity, test_codex_runtime, test_codex_discovery, test_postal_multi_marker, test_postal_hydrate_scope, test_perf_stats, test_kernel_names, test_postal_undelivered, test_postal_walked_ask, test_postal_live_only and test_kernel_awaiting_merge. In separate runs: test_kernel 460 passed with 12 subtests, test_injected_voice 10 passed with 222 subtests, test_backend_call_parity 4, test_state_isolation_order 8, test_codex_runtime 11, test_tempdir_hygiene 23 passed with 64 subtests.

- The body's unchanged-and-green list (test_postal_live_only, test_postal_delivery, test_postal_self_send, test_postal_freshness, test_postal_deferred_push_identity, test_postal_stranded_putback, test_postal_sender_identity, test_codex_runtime, test_feed_session_started, test_state_dir_override): 149 passed, 1 failed. The one red, `HealOlderStores.test_a_blocked_machine_top_keeps_its_needs_you_card_and_wears_the_face` in test_feed_session_started, fails identically over the base sources with `sdk-backend: claude_agent_sdk not found` on stderr: a condition of the test environment used, not of this change.

- Red before: in a scratch tree at the head with kernel/codex_backend.py, kernel/codex_events.py, kernel/kernel.py and postal/postal_service.py restored from the base ea3c82725, test_codex_postal_tools, test_codex_events_golden and test_codex_backend give 28 failed, 148 passed; the 28 are all six PostalTools tests, both dynamicToolCall golden tests and all 20 of the new module, each on the assertion the body names. With only the `_agent_rows` backend line reverted, `test_push_picks_the_hint_from_the_recipients_backend` is red on the banner's last line naming the shell command and its two siblings stay green, as the body says.

- The lock-order test asserts its mechanism: moving the postal call under `_sessions_lock`, and separately under `s.lock`, each turns `test_the_postal_callable_runs_with_no_backend_lock_held` red while the other five PostalTools tests stay green.

- Randomized parity, kernel and bus both loaded by path, seed 1867: recall 400 cases, send_message 1500 (kind and tracked varied over type and case, comparing the result and the wire payload), receipts 600, `list_agents` 600, inbox 600 on every line but the hint, `check_sent` and `list_agents` 200 through the dispatcher: all equal.

- test_stage_marks at the head: `test_every_callable_the_kernel_hands_a_backend_is_marked_at_the_hand_off_or_listed_with_a_reason` and `test_every_listed_callback_still_has_a_hand_off_behind_it` 2 passed, with no postal row (the census point in the main comment).

- Mutation and probe runs behind the points in the main comment. Create-branch merge dropped at kernel/codex_backend.py:1699: the changed test modules green in two independent runs (176 passed, 5 subtests over the three changed modules; 191 passed, 11 subtests over those plus test_codex_runtime and test_backend_call_parity), the "not registered" line at kernel/codex_backend.py:967 never reached because it sits inside `_postal_params`, and a probe test forcing the create path (a pending- tid, `s.loaded = False`, `_prepare_thread` called directly) red on the mutant, green on the original, 2 passed after the source was restored byte-identical. Kernel log line: `HTTPConnection` patched to raise ConnectionRefusedError, then TimeoutError, and `_set_working_note` patched to raise OSError(28), each called directly and through `CodexBackend._postal_tool_call`; a control callable raising RuntimeError produced the backend's `codex tool call ... failed` traceback line, the production callable produced nothing on stderr; the bus logs the equivalent fault for its own tool at postal/postal_service.py:5567. `check_inbox` status -1 collapsed into 0, and hang and refuse collapsed into 200 and `{}` for `list_agents`, `check_sent` and `recall_message`: test_codex_postal_tools 20 passed each time, while the unmutated fake shows the hang and refuse texts differ (one carries "count as read") and the other three tools' texts and write behavior are distinct. The bus's three recall sentences mutated: 20 passed without the proposed pin, 1 failed with it (on the "Recalled" wording), 21 passed with the bus restored and the pin kept; five hand-built shapes and 400 randomized ones showed 0 diffs between `_codex_recall_text` and the bus tool today. Oversize bounce: driven through the real `_agent_rows` with a Codex row, the bounce note read 921599 against the 921600 limit for a message whose delivered body was 1 byte over; no test in test_postal_delivery, test_postal_peers, test_stat_rule_readers or test_codex_postal_tools drives `_push` with an oversize message for a Codex row, and the source-pin counts in test_stat_rule_readers.py are unaffected by the proposed change. Sandbox refusal: a probe reproducing the bwrap filesystem and uid conditions from Codex 0.153.3's bwrap.rs and bubblewrap.c (tmpfs root owned by the sandbox uid, no --clearenv or --remount-ro), with a fake loopback `/ping`, CODEX_THREAD_ID set and an empty HOME, printed the doc's refusal and pointer with a token minted on the tmpfs; a read-only HOME variant raised the import-time RuntimeError instead, a condition the real sandbox does not have. Resume registration: read, not run, against openai/codex at rust-v0.153.3 (`ThreadResumeParams` versus `ThreadStartParams`, thread_processor.rs, thread_manager.rs `resume_thread_with_history`, session/mod.rs).

- Not executed: tests/smoke_codex_live.py (no Codex runtime available); its module loads under test_state_dir_override. The live wire-shape claims (dynamicTools accepted at thread/start and thread/resume, the call arriving as item/tool/call, the reply accepted, no reviewer step, developerInstructions ignored at resume) rest on the PR's own probe. The pinned openai-codex 0.144.4 wheel was read instead: every server request reaches the approval handler, the handler's return is written verbatim as the JSON-RPC result, a raise there runs `fail_all`, and `_params_dict` passes a plain dict through unchanged, so `dynamicTools` reaches the wire although the generated `ThreadStartParams` has no such field.

lczh:
Review round folded as one commit on top (70ae36dd): the create-path registration test, the callback census descending an IfExp plus the allow entry, a once-per-spell kernel log line for bus faults, the oversize bounce measured with the tool banner, the fault-branch and recall-parity pins, and the two doc corrections (the sandbox refusal's reason; a resume cannot add tools on rust-v0.153.3). Red evidence and counts are in the body's Review round section.

ksarma:
Executed against 70ae36dd8, where every item from the earlier comments landed as proposed. Shown by execution: the create branch's registration pinned (red with that merge dropped); four tools' fault branches pinned; the recall sentences equal to the bus tool's over six shapes; the bounce measured with the recipient's banner; the census reading a conditional hand-off; the kernel log line latched once per spell. The doc, docstring and comment read as suggested.

## At this head

**The two kernel-log tests are order-dependent.** `TheKernelLogNamesAFault` captures all of stderr; the dial at kernel/kernel.py:4124 sits outside the request's try and the census write at 21185 is unwrapped, so a process's first dial puts the census line in the capture: the count at tests/test_codex_postal_tools.py:457 reads 2 != 1, and in the dead-stderr test the write raises before the request, so line 496 gets the generic sentence. Each test alone is red, the class alone 1 failed, 1 passed, a worker collecting it first red; in file order the module is 24 passed. Fix: one `km._bus_port()` call in setUp before any capture (class alone 2 passed, module 24); a line filter repairs the first test only, the second failing on the raise.

## Smaller points

- **The same unwrapped census write reaches production** (kernel/kernel.py:4124, 21185). A port or source change after boot makes a Codex call's dial say the census; with stderr unwritable the raise reaches the catch-all, the session gets its generic sentence (4223), not the per-fault one, and the bus is not dialed. Fix: wrap that write as the log's is (the dial inside the try would misfile the cause), pinned by a `km._BUS_PORT_SAID[0] = None` reset before the dead-stderr capture: red here, green wrapped.

- **The shell recipient's bounce measurement is unasserted** (tests/test_codex_postal_tools.py:566). That message sits on the cap as a shell body, rides the wake and never reaches `_bounce_oversize`; with postal/postal_service.py:2448 measuring with the tool banner unconditionally, nothing goes red. Fix: push to the sdk row a message from the Codex session (a sender equal to the recipient gets no note), one byte over as a shell body, and assert the count in the sender's note equals `pm._deliver_body_bytes(WEB, [m])`, 2 under the tool-banner measurement: 921603 != 921601 on that edit.

- **One slot for two causes** (kernel/kernel.py:4093). `_CODEX_POSTAL_SAID` is one boolean for both bus faults, re-armed only by an answer (4149), so a spell whose cause changes with no answer between logs only the first: refuse then hang gives one line. The earlier comment asked for one latch; a set keyed by cause (`once='unreachable'` and 'no-answer', cleared on an answer) gives one line per cause per spell; a refuse-then-hang case expecting two lines is red here.

lczh:
Second round folded as one commit on top (5a3d226c): the kernel-log tests dial the bus port in setUp before capturing; the first dial's census write is wrapped like the log's, pinned by a dead-stderr case with the memory reset; the shell recipient's bounce is measured against its own banner in the hint test; the fault latch is a set keyed by cause, cleared on an answer, with the refuse-then-hang case expecting two lines. Counts in the body's Second round subsection.

ksarma:
Executed against 5a3d226ca. Everything from the earlier comment landed at this head: `TheKernelLogNamesAFault` dials the bus once in setUp before any capture (each test alone, the class alone, the class first on a worker, all green); the census write is wrapped in `_bus_port_census` with the dial left outside the request's try, pinned by a test that resets the census memory before a dead-stderr capture (red without the wrap); the shell recipient's bounce is asserted from a message the Codex session sends to the shell row, one byte over, the count read from the sender's note (red with the measurement hardcoded to the tool banner); and the said slot is a set keyed by cause, cleared on any answer.

Two small points at this head:

- tests/test_codex_postal_tools.py:543: the per-cause test's clause that an answer ends the spell for both causes pins only the unreachable cause's re-arm; a kernel that clears only that cause on an answer passes the module (executed). After the answering call, run a hang under the capture and assert one no-answer line beside the re-said refusal.
- kernel/kernel.py:4115: the log line's suffix still says the fault was said once per fault spell, while the latch is now per cause, so a refuse-then-hang spell writes two lines that each claim to be the spell's one; say once per cause per fault spell, and assert the phrase on each captured line in the per-cause test.

lczh:
Third round folded as one commit on top (90b17716): the per-cause test now runs a refusal and a hang after the answering call and asserts both lines, so the answer's re-arm of both causes is pinned (red with only the unreachable cause cleared); the log suffix says once per cause per fault spell, asserted on every captured line.

ksarma:
Executed against 90b17716c. Both points from the earlier comment landed at this head: the per-cause test now runs a hang under the capture after the answering call and asserts one no-answer line beside the re-said refusal, and a kernel that clears only one cause on an answer is red there (1 != 2, either cause, with the class's other three tests green); the log line's suffix reads once per cause per fault spell (kernel/kernel.py:4117), with the latch comment and the docstring at the same grain and the phrase asserted on every captured line, red on the earlier suffix. The module is 26 passed; the census, the trust module and the backend module are unmoved (194 passed together). Nothing further at this head.

ksarma:
Executed against 3d1138b7d, the merge of main into the branch. The resolution in kernel/codex_backend.py is main's try around the resume with the branch's one keyword spread placed inside it (lines 1855 to 1859): the except still swallows only a reply the pinned SDK cannot read and re-raises everything else, the postal parameters reach the server on the resumed road as on the two fresh-thread roads, and the backend and postal-tools modules are 157 passed at this head. One thing the merge left behind: docs/codex.md:186 keeps the earlier list fragment "and postal delivery into Codex sessions." right after the sentence that closes the list at 185, so the paragraph reads "...a Claude session's do. and postal delivery into Codex sessions. A resume whose reply..."; the branch had replaced that line with its three-line sentence at 183 to 185, and keeping both sides whole brought it back. Dropping the fragment so 186 starts at "A resume whose reply" closes it. Smaller: no test drives a resume whose reply fails to parse while the postal tools are configured; the fake at tests/test_codex_backend.py:1418 records the parameters before it raises, so asserting the recorded dynamicTools inside the parse-failure test would pin the composition.

henrypinkard (approved):
Approved on the maintainer's decision of 2026-09-20.

PR #1868, opened 2026-09-19
The proto-2 chat base never ends on an input echo or the command chip, so the landing that replaces one is a delta and not a whole session frame; every full is counted by reason on /perf and filed when the client held a base

lczh:
Tier: fix

## The defect

The uuid-anchored chat wire (proto 2) keeps, per client, the first and last keys of the tail run the client holds, and sends a delta only when the next built list still carries both. The last key was taken from the last event that is not an overlay card, and after a send into an idle session that event is the kernel's own input echo: the bubble the SDK backend paints the instant a message is sent, under a synthetic key of the form `echo:<hex>` (the Codex backend's echo is `echo-<hex>`; the acknowledgement chip for a `/model`, `/effort` or `/auth` pick is `cmd:<t>:<name>`). The moment the CLI writes its own record for that message, the echo leaves the built list (its text is now in the transcript, so the live merge drops it) and its key is in no list at all. The next push therefore could not map the base and fell through to the full branch: a whole session frame, the last 250 events, to every caught-up client of that session. This happened on every user send, every nudge, every postal delivery and every manager dispatch into an idle session, and the page treats a whole frame for a session it already holds as a reconnect repair, rebuilding its window and re-deriving the reader's place. The 2026-09-13 change that taught the base to skip overlay cards left these keys out, and no test covered an echo at the tail.

Measured on a serving kernel on 2026-09-18, about twenty hours of uptime, from `/perf` alone: 1022 full chat frames (644 MB) against 70,138 deltas (581 MB); a re-read an hour later gave 1034 (650 MB) against 71,018 (594 MB), with 450 targeted builds and, on the kernel's side of the client diagnostics, only socket-open and history-reply rows (87 and 6), while the page had filed 51 slow-frame rows naming a whole session frame. The kernel had no way to say which producer made any given full.

## The fix

The base's last key now skips the kernel's transient live-tail keys the way it skips the overlay cards: a small tuple of key prefixes beside the overlay kinds, read by one predicate that is safe on a missing key and matches a `uuid#n` key by its prefix. With that, the base ends on the record before the echo, and the landing is a delta after that record carrying the CLI's record and whatever followed, the road proto-1 clients already take; the client's tail handler truncates the echo away, appends, and re-runs its pending-send reconcile, asking for no full. The prefixes are not folded into the overlay kinds, which is a list of kinds twinned with the client's and pinned equal; the events keep their shape on the wire. The known residue, named in the docstring: a list of nothing but such events (a transcript-less session whose only event is its first echo) still anchors on its last event and sends one full at that first landing, once per session creation; and the first edge can still sit on a transient key when the tail-run cut lands on a placed dropped echo, which takes the existing road for a first edge the list no longer holds.

This PR carries the fix and its recurrence meter. The prefix rule is a convention, not a stamp on the event, so a fourth live-tail key minted outside the tuple would re-open the defect silently; two things make that loud. A source pin holds the four mint sites to the tuple: the SDK echo minted at a send, the same echo re-minted at a kernel boot for a persisted mirror entry that carried no uuid (an older kernel's mirror), the command chip, and the Codex echo; with them the gate that admits a client's own id, the pattern the handler checks it by, since the client's echo rule is a consumer of the form and not what lets it in. Two behavioural arms drive the boot reseed and an admitted id through the real backend to the built list's tail and assert the base skips both. And every proto-2 full is now counted by the reason the sender had for it under `/perf` as `pusher.chatFullWhy`: no base held, an empty list, both edges gone (a fork, a rewind), the last edge gone with the first held (this defect's shape, labeled by the vanished key's family so a recurrence names its prefix), an inverted base, a change at the list's first event (the targeted push's shape), or a change before the held first. The label vocabulary is bounded (a vanished record uuid is one label, never one per uuid) and the map is capped like the sends map, counted through a locked collector method on an attribute of its own that the snapshot copies. The count follows the frame that left, not the decision to send one: the per-client sender dedupes a whole frame the client already holds byte for byte within its repost window (a targeted push repeated on one build, the connect handshake's second push) and hands the client nothing, and the first cut of this meter fired before that check, so it counted those fulls and filed a row for a frame nobody received, its total outrunning the sends count, which has never counted a deduped frame. The sender now reports whether the frame went, and the meter and the row run only when it did; every other caller of the sender ignores the answer. A full sent to a client that holds a base also files one `chatFull` client-diag row (the client's id and kind, the session, the reason, the change index, the list's length, and which edges the list still held; no text, no names), under the client's slot lock, which is safe because the diag append takes only its own leaf lock and no other inside it. The reference doc names the new pusher key, the row, and the base rule, and says of the key and the row that a frame the dedup swallowed is neither counted nor filed.

What is inferred rather than proven here: the client-side effect (no whole-window rebuild at a landing) follows from the frame type, and the node test on the lifted tail handler is the cheap proof that fits this tier; a served proof that a sender's tab shows one bubble through echo then landing, with no DOM-clearing rebuild, belongs to a client sibling. Two shapes worth stating from the probe: an echo now rides the suffix once more than before when an overlay card arrives after it (the delta after the record carries the echoes and the card), which the client strips, re-applies and re-hides through its pending reconcile with no stale mark, so no rebuild; and the client's echo rule still names the `echo:` form alone, so a Codex echo is never hidden behind the sender's bubble there, a sibling's to close. The targeted push's fulls to caught-up clients show under the change-at-zero label and are a sibling's design.

## Tests

Red first, on the stock kernel. In the chat pages module's proto-2 class: the echo landing test fails at its base assertion (the base ends on the echo key, not the record before it), and its landing send is a whole session frame; the Codex-form, dropped-echo dismissal, command-chip and two-echoes-then-overlay arms fail at the same assertion; the rule-and-label test and the counter test fail on the missing predicate and the missing `/perf` key. In the queued-copy identity module, the real road (a send through the SDK backend into a thread-parked session, the echo made the visible copy, a build, the landing record written, a second build) fails at the base assertion the same way, with the key vanishing through the merge's own filter rather than by a test omitting it. The perf-stats module fails on the missing key, the missing method and the thread-safety hammer; the overlay-kinds module fails for want of the prefix tuple. A second red, taken on the first cut of this change after its review: the counter test's identical-frame arm (the same whole frame sent again inside the dedup window, to a client holding a base) read the change-at-zero label at two against the expected one, while the frames that left the client's send and the sends count were unchanged; green once the count follows the send. The reseed and admitted-id arms and the two new source pins cover ground the first cut left unpinned and are green on it and after.

Green after the fold, the same modules one at a time (chat pages 41 with 27 subtests; chat pages warm 13; chat window spans 8; chat skeleton reconnect 27; perf stats 63; queued copy identity 28; send-pending overlay kinds 3; kernel 460 with 12 subtests; feed memo inputs 31; chat build-signature inputs 65; perf bench 49; chat split 55; tail-lo leading note 9; backend call parity 4; browser first draw 4; history reply diag 4; and the socket-drop module, 8, which reads the per-client sender's source). The static pin scan reports no lost pins. The extension suite, which pins kernel source lines and bundles the webview tests, passed 5048 of 5049 on the first cut with the pre-existing KaTeX pixel probe the one failure, and the new node test on the lifted tail handler (a resident list ending in a hidden kernel echo receiving the delta after the record before it: the echo truncated, the record and reply appended, the pending reconcile re-run, no full asked) passes; the fold changes no kernel line a node test reads (the only sender lines they pin are the comments and glossary call sites, untouched), so it was not re-run.


ksarma:
The evidence below was executed against fa92e8a46. The current head 333a308cf is that commit with main merged into the branch three times; the change is byte-identical (7 files, +549/-18, the same added and removed lines), so every finding carries over. Line numbers are as at fa92e8a46; at 333a308cf the kernel/kernel.py reason function sits at lines 50304 to 50314, the row call and the base write at 50396 and 50397, the two echat reads at 49000 and 49073, and the docs/reference.md bullet at line 2021; the test-file lines are unchanged. #1870 carries this commit as its first, so it lands there with whatever this PR has.

### At this head

One mechanism is worth changing before this merges. For a session whose built list is empty, kernel/kernel.py line 50125 records a base with both edges None. A just-created SDK session is that case (registered before any first send, built with zero events, not skeletoned for want of a transcript), so every proto-2 client receives the empty full, and the pusher loop, which reaches every client every cycle, re-sends the identical frame once per 60 s repost window, the dedup in `_send_client` being its only gate. Each repost is counted under `empty`, and since the base is a dict, line 50124 files a chatFull client-diag row per repost per client with no event behind it. The row's stated meaning, a client owed deltas that took a whole-window rebuild, cannot apply: the delta branch is gated on `total`, so an empty list has no delta form.

Executed against fa92e8a46, this reproduces. One empty send and three re-sends past the repost window give noBase 1 and empty 3 under `chatFullWhy` and three rows (reason empty, total 0, both edge flags False). Through the real `_push` loop with three proto-2 clients: 3, 6 and 9 cumulative rows on successive reposts, all labeled empty. The idle case measures at 180 rows for one fresh session left an hour with three tabs open, about 41 KB an hour at 230 bytes a row, far short of the 8 MB rotation.

The same base has a second effect at line 50036: the first content frame finds neither edge, so it is counted `baseGone` and files a row with both flags False. Over the creation sequence (empty frame, first echo alone, landing record and reply) the meter reads noBase 1, baseGone 2 and two rows per client, against noBase 1, baseGone 1 and one row for a client connecting after the echo: one spurious count and row per client per creation, and the doc's sentence that a first send files nothing fails for that client. The full predates this PR; the label and the row are new.

The fix is to record no base for an empty list: write `st[sid]` only when the total is non-zero, else pop the entry. In a scratch copy the reposts and the first content frame become noBase fulls with no row, the frame after that is a delta, the labels still sum to the full chat send count, and the counter test's empty arm still reads empty 1. Two consequences to document: the entry's presence is read at lines 48766 (the cold-tab gate, for a client whose reconnect is unresolved) and 48834 (the reconnect resolve), so a session with a stat-able but empty transcript can be dieted as a skeleton for a reconnecting client instead of re-sent whole (a transcript-less one is already excluded by `_skeleton_for`); and the idle reposts still leave once per client per window, now under noBase, whose gloss (a first send or a reset) should add that repost. The other shape keeps the entry, so those two reads are untouched: relabel to noBase in `_chat_full_reason` for a base with neither edge and also skip the row, since relabeling alone still files one. Pin both: a re-sent empty list past the dedup slot files no row and no empty label, and the first content frame after it is noBase without a row.

### Smaller points

- kernel/kernel.py line 50124: the row is appended to disk under the client's slot lock, on the pusher's road and the targeted push's (handler and backend threads), where at the merge base nothing writes to the state disk. Executed: a 0.6 s stall in the append made a handler thread calling `_client_reset_chat_sid` wait 0.551 s; normal cost is p50 20.7 microseconds, and only a full that left to a base-carrying client files a row. The two lock sites are `_send_chat_or_status` and the targeted push's sender (the one-client push takes no client lock): hand the row out, append it after the lock block with the decision kept under it, and drive the Proto2Wire row tests through one of those two.

- tests/test_send_pending_overlay_kinds.py line 59: the Codex echo mint is pinned by source text alone; executed, a key rewritten to a codex- prefix behind an intact mint line leaves the pin and every Codex test module green, while a reformatted line with the key unchanged turns the pin red. The chip differs: test_sdk_backend already fires `_ack_cmd_chip` through set_model and asserts the cmd: prefix, so only the link to the kernel's tuple is missing there, and its literal duplicates tests/test_kernel_cmd_gesture.py line 78. One executed line closes the Codex gap, in tests/test_codex_echo_merge.py after the send: assert `km._transient_key` of the first live atom's event key; keep the tuple-equality assertion.

- kernel/kernel.py lines 50042 and 50046: the `inverted` and changeBelowFirst labels are asserted by no test; swapping the two comparisons is caught by the counter test's changeAt0 arm, but dropping either branch or loosening its comparison leaves the three modules that observe the meter green while the full lands under other. Add two direct assertions in the rule-and-label test, or two sender-driven arms in the counter test, each with a fresh `_client()` since a byte-identical repost within the window is deduped and counts nothing, and extend the closing sum check rather than weaken it. The other arm is unreachable through the sender.

- tests/test_chat_pages.py lines 1000 to 1001: the row's client identity and edge flags are asserted only for a bare client (cid None, kind None, wid unchecked) with both edges present, and the lastGone arms at lines 1015 to 1022 check the reason string alone; executed, swapped or constant flags, swapped identity fields and a wid taken from the cid field each leave the module and the other three changed modules green (a dropped key is caught: the whole dict is compared). Give c3 a synthetic cid, a non-relay kind such as page and a wid, and assert the lastGone:echo row's wid and data (changeFrom and total both the list length, `firstHeld` True, lastHeld False): that turns all four red and stays green on the head.

- docs/reference.md line 1962: no test reads the doc for the new key or the row, where other /perf additions are pinned (test_browser_first_draw asserts connectPush); executed, all 21 doc-reading tests pass with the doc replaced by the merge base's copy. No project rule requires such a pin and the behavioural coverage is real; two asserts where `GoalIoCounters` already reads the doc in tests/test_perf_stats.py follow the precedent, and asserting every label the reason function returns appears in the parenthetical, as test_perf_heap_block's Documented test does, also catches a renamed label.

The tests run against this head, with their results, follow in a second comment.

ksarma:
Tests run against fa92e8a46 and their results; every run was memory-capped and hermetic (the manager, kernel and serve ports pinned to 1, no session variables in the environment), never against a serving kernel.

- Fails-before, executed: kernel/kernel.py restored to the merge base ea3c827 in a scratch archive with the tests at the head gives 15 failed, 120 passed across test_chat_pages, test_perf_stats, test_queued_copy_identity and test_send_pending_overlay_kinds (7 in Proto2Wire, each landing arm red at its base assertion with the echo or chip key as the base's last edge; 3 in test_perf_stats on the missing key, method and thread hammer; 3 in TheEchoLandingKeepsTheProto2Base on the missing predicate; 2 in TransientKeysAgree). At the head the same four modules give 135 passed, 27 subtests (41, 63, 28, 3, matching the body).

- Every module named in the body's Tests section run standalone under pytest -n 6: every count matched (chat pages 41 with 27 subtests, chat pages warm 13, chat window spans 8, chat skeleton reconnect 27, perf stats 63, queued copy identity 28, overlay kinds 3, kernel 460 with 12 subtests, feed memo inputs 31, chat build-signature inputs 65, perf bench 49, chat split 55, tail-lo leading note 9, backend call parity 4, browser first draw 4, history reply diag 4, ws drop loud 8).

- Neighbouring pin and consumer modules at the head: test_ws_drop_loud, test_chat_skeleton_reconnect, test_history_reply_diag, test_chat_pages_warm, test_chat_window_spans, test_tail_lo_leading_note, test_chat_split, test_backend_call_parity, test_feed_memo_inputs, test_chat_build_sig_inputs, test_perf_bench: 273 passed, 21 subtests; test_ws_open_row and test_pane_shim_return: 54 passed; test_view_deltas, test_chat_payload_clock_invariant, test_payload_dedup_invariant, test_kernel_tabs_first, test_timeline_skeleton_dedup, test_client_diag_rotation, test_kernel_pusher_snapshot, test_pusher_change_gates, test_first_cycle_stage_split, test_perf_gc_block, test_state_isolation_order and test_browser_first_draw green in three overlapping sets (289 passed with 31 subtests, 279 passed with 24 subtests, 277 passed with 27 subtests).

- The modules that read docs/reference.md: 310 passed, 3 skipped, 1 failed in test_host_transport EndToEnd under a -n 6 fan-out (a session host exited before serving its socket); rerun alone it passed in 7 s, and the PR touches no host code.

- Node: node esbuild.js --tests, then node --test over out-tests: chat-exact-tail-exec 13 pass with the new landing case, and 13 pass with the same test copied onto the merge base (the PR changes no client code: its one file under ui/webview is the test itself); chat-exact-tail, chat-delta-resync, css-vocab, menu-theme-tokens, dense-chrome, click-safe, feed-delegation, send-pending, send-pending-overlay, chat-window and chat-windowing: 145 pass, 0 fail in one set and 103 pass, 0 fail in another; tsc --noEmit exits 0 with the new chat-window import and the optional hook fields.

- The empty-list finding: sender-level probes and a probe driving the real _push loop with three proto-2 clients over the test_chat_pages harness (2 passed); with the proposed fix applied in a scratch copy, test_chat_pages, test_perf_stats, test_queued_copy_identity, test_send_pending_overlay_kinds, test_chat_skeleton_reconnect, test_chat_split, test_chat_empty_build_guard, test_ws_drop_loud and test_drop_anywhere_served: 230 passed, 4 skipped, 27 subtests; and test_chat_pages, test_chat_empty_build_guard, test_queued_copy_identity: 74 passed, 27 subtests.

- The lock finding: a probe over Proto2Wire with the diag append replaced by a spy recording lock ownership and thread (50 passed): the row filed under the slot lock on both roads at the head, an empty log on both roads at the merge base, the 0.551 s wait behind a 0.6 s stall, and the append timing over 5000 rows of 238 bytes (p50 20.7 microseconds, p99 41.7, max 132.9).

- Mutations for the pin findings, each in a scratch copy with the kernel restored byte-identical after: the Codex mint reformatted (pin 1 failed, 2 passed); the Codex key rewritten (pin 3 passed; test_codex_backend 119 passed; test_codex_echo_merge, test_chat_pages, test_queued_copy_identity 73 passed); the chip key rewritten (pin 3 passed; test_kernel_cmd_gesture, test_chat_pages, test_kernel_timeline_awaiting, test_sdk_echo_durability 121 passed; test_sdk_backend 1 failed on the acknowledgment-chip test); the reason function's two comparisons swapped (1 failed, 106 passed) and either branch dropped (107 passed, 27 subtests, twice); the row's flags swapped, constant, identity fields swapped, wid from the cid field (test_chat_pages 41 passed each; the other three changed modules 94 passed) and a dropped key (1 failed, 40 passed); the doc reverted to the merge base's copy (17 doc-reading ids, 21 passed).

lczh:
Follow-up for this review: #1895 (an empty chat list records no proto-2 base, so its reposts and the first content frame are noBase with no row; the chatFull row is appended after the client lock; the mint, label, flag, doc, led_changed and road-reset pins; the changeAt0 prose; the _send_chat docstring; the reference's stage attribution).

PR #1870, opened 2026-09-19
The targeted one-session push serves each client from the base it holds, so a caught-up page gets a tail and a skeleton tab its status, not a whole session frame

lczh:
Tier: fix

Stacked on #1868 (the proto-2 base anchor fix): both change the chat send path, so this branch carries that commit until it merges; the diff that is this PR's own is the second commit.

## The defect

The kernel's targeted one-session push, the one the SDK backend runs at every connect handshake and the Codex backend at every stream event, and the one a create, a fork and a promote run to paint the new tab at once, ended by handing every connected chat client a whole session frame. It called the plain chat sender directly with the change index forced to zero, bypassing the per-client road the pusher's own loop takes. That one bypass had two visible consequences. A page that already held the tab whole received a full session frame it treats as a reconnect repair: it marked its view stale, re-rendered the whole window and re-derived the reader's place, a repaint for no new content. And a tab the page held as a skeleton after a redial was force-loaded by the same frame, undoing the redial diet the skeleton exists for.

The docstring justified the forced zero with a design rule: a full always, so no client's tail state could be left ahead of the shared delta baseline. That rule was already obsolete. The ready arm's connect push sends from the untouched baseline and leaves clients ahead of it as a matter of course, and the next pusher cycle re-sends the overlap idempotently (a proto-2 tail starts at the change or right after what the client holds, whichever is earlier; the index wire truncates at the frame's from). The 2026-07-28 hazard the rule remembers was advancing the baseline from a push that did not reach every client, which this push never did and still never does.

Measured on the live kernel over about 20 hours: 426+ targeted builds across 22 sessions, and 973 full chat frames against 68,377 deltas with only 133 of the fulls being connect pushes. The client's diagnostics over the same window show slow-frame rows for session frames with no socket open or redial anywhere near them.

## The fix

The targeted push now serves each client through the same per-client road the pusher's loop uses, with the same inputs: the diff of this build against the shared baseline, and whether the ledger changed against the baseline's ledger. The baseline itself is still left alone by this push, because only a push that reaches every client may advance it; that half of the old rule stands and its runtime pin is unchanged. Per client, all through existing code: a caught-up client gets a chatTail whose suffix begins at the change against the baseline or right after what it holds, empty when nothing moved (the status flip the handshake exists for rides it; an identical tail is deduped on its slot within the sixty-second repost window and re-sent as a no-op after it); a client holding the tab as a skeleton gets its status on the status slot and keeps the skeleton until the click or the prefetch releases it, as with the pusher; a client with no base (a fresh socket, a redial's watched tab, a needFull reset) gets the full as before; and a client armed for the ready arm's connect push gets nothing here, which is the pusher's existing rule applied to this sender too, not a new withholding. The tail needs a baseline, and a based client is on the full road without one: a session the pusher has not built yet (a create, then its handshake seconds later; a new Codex session's first-turn notifications before the pusher's first cycle), one whose only cycle build was transcript-less (an empty baseline reads as none), or one that left and re-entered the strip (the pusher pops the baseline of a tab no longer shown) still gets the full from every targeted push until a cycle has built it, exactly as today.

Reading the baseline from another thread needed two ordering rules the review found missing; both are part of the fix. First, the pusher now advances the shared baseline only after its cycle has delivered to every client, not before its send loop as it did. The targeted push reads that baseline on the backend's thread with nothing ordering the two, and a push landing in the old write-to-send gap diffed its build against an equal baseline and anchored an empty tail at the list's last event, one the racing client had not received yet; the page reads a tail past what it holds as a gap and asks for a full, the very frame this change removes. Read after delivery, an older baseline only re-sends the overlap, which the client applies idempotently; a newer one was the hazard, and the pusher's own sends never depended on the write (they use the cycle's local diff). Second, the proto-2 no-change tail anchors at the client's held last when that lies before the list's last transcript event, so a client behind the build for any reason receives the events after its own anchor instead of a tail past what it holds; a caught-up client's tail stays the empty suffix it was, anchored at the list's last, so a trailing overlay card (an api-error notice, a queued marker) does not ride every status-only tail. The index wire is covered by the first rule; its gap ask for a build shorter than the baseline is unchanged.

To be plain about what this reverses: the change removes a documented design sentence (the forced-zero justification) and flips two tests that encoded it, the one asserting the push's full releases a skeleton tab and the one asserting the push's own full follows a redial's first strip. Both now assert the routed behavior: status only, skeleton kept. The docstring carries the positive safety argument in the old sentence's place: for proto 2 the next tail starts at the earlier of the change and the held last plus one, the no-change tail included, so a client ahead of the baseline is never skipped, only re-sent the overlap since the last pusher cycle, and a client behind it is served from its own anchor; the baseline is advanced only after delivery to every client, so a mid-cycle read sees the older one; an index-wire client left behind the baseline asks for the gap itself, the road every connect push already relies on; and the baseline is never advanced here. One residual is named rather than hidden, shared with the connect push and not new: a client ahead of the baseline whose held event at one index differs from both the baseline and the next build while the next build equals the baseline there is re-sent by neither sender until a full. No builder produces such a flap today.

Nothing else in the handshake needed an explicit refresh. A chatTail carries status, the notify and hide flags and the postal flag, and the ledger when it changed, but not the full frame's top-level fields (the mail-off reason, the branch, the worktree, the cwd, the repository, the background tasks, the name and color); the pusher's delta has the same shape, and no caller of this push depended on the accidental full for them: the SDK handshake and the Codex mode and effort changes flip the status, and a create, fork or promote is a new session with no baseline and no held base, so it still gets the full.

For the post-deploy read, every frame this push sends, its tab strip and the cold-tab gate's status included, is counted on /perf under its own slot, a `.targeted` suffix on the send slot (`chat.targeted`, `status.targeted`, `taborder.targeted`), through a thread-local road name the push sets for its whole body and the per-client sender hands to the counter. That instrument is included deliberately as the post-deploy verification for this fix; the spec had listed an attribution row as a separate item, and this is the narrower slice of it (which road sent a frame, not why a full went to a client holding a base). After a deploy, `sends.full.chat.targeted` should stay near the number of fresh sockets plus the first-cycle fulls described below, `sends.delta.chat.targeted` should carry the handshakes, and the client's slow-frame rows for session frames with no socket open nearby should stop. Three things to know when reading it. The tail is computed against the baseline, so every targeted push re-sends the suffix since the last pusher cycle rather than since the last targeted push, and for a Codex session streaming many notifications inside one long cycle (the live cycle was 96.8 s at its worst) that suffix grows through the cycle, bounded by the cycle and never worse than the 250-event full it replaces unless the suffix itself exceeds the wire tail. The first unchanged targeted push after a full now sends one empty-suffix chatTail where the old code sent nothing (the slot held the full's signature), harmless on the client but one more delta on the counter. And a session the pusher has not built since it entered the strip is a full to every client on every targeted push until the first cycle builds it (a create and its handshake seconds apart; a new Codex session's first turn), so `sends.full.chat.targeted` runs above the fresh-socket count by those, with no client left holding a stale view.

Open, as separate items. The attribution of the 426+ targeted builds: the lane's premise was frequent SDK reconnects, but the read-only check of the live kernel found the SDK handshake ran only at boot (twenty-one attaches, no reconnect lines, no crash heal) and the Codex backend's per-notification push calls are the likely producer; confirming that split, and moving the Codex stream roads to the plain push wake so the build itself stops, is a separate change that needs the Codex maintainer's view. The order inversion between a targeted push and an in-flight pusher cycle building from different lists is not closed here and is not made worse. When the targeted build is the fresher one and the client's anchor is absent from the older list, the cycle sends a full carrying the older list, as today. When the targeted build is the older one and the client's anchor is in both lists, the push now sends a tail carrying the older rendering of an event below the anchor, which the client applies in place, and the next cycle, finding the baseline equal to its build, never re-sends it; today a full of the same older list left the same stale content after a visible rebuild. The closer for both forms, a per-session build generation the sender checks before moving a client backwards, is its own fix. The echo base-miss, which forces a full on the push after every send into an idle session, is the sibling fix; this push now shares the pusher's behavior there rather than adding its own full.

## Tests

Red before, in the kernel unit harness that drives the real push over fake clients with the builder stubbed to synthetic payloads. A caught-up proto-2 client that took the cycle's full and is pushed again by the targeted push: the no-full assertion is green on the unchanged kernel too, its identical full deduped on the slot; the empty-suffix tail asserted right after it is the first red one (the unchanged kernel has no tail road for this push); and the status flip is the user-visible symptom, a whole session frame as soon as anything differed from the frame on the slot (the flip, an appended event, an edited event). A client holding the tab as a skeleton received a whole session frame and lost the skeleton; a redialing page whose first strip was this push received the push's own full for a skeleton tab; a client armed for the ready arm's connect push received a full it could not hear. Green after: the caught-up client receives an empty-suffix tail for an unchanged push, a tail carrying the new status for the flip, a one-event tail for an append, and a tail from the changed event for an edit, while a client with no base still receives the full and the shared baseline is unchanged throughout; the skeleton holder receives its status only and keeps the skeleton while another page holding the same tab whole receives an empty-suffix tail with the new status; the redialing page receives the strip, its parked focus, and the skeleton tab's status, no full; the armed client receives the strip and nothing else.

The two ordering rules have their own red-first tests. The race: a stub on the pusher's per-client sender runs the targeted push at the cycle's first send for the session, with the session grown by one event since the baseline, over a proto-2 client and an index-wire client at once. With the baseline still written before the loop, the index-wire client received a tail from past what it held (from 6 against five held events), and without the anchor rule the proto-2 client's tail was anchored at the new event it did not hold. Green with both rules: every tail is anchored at an event the client holds and carries the missing event. The anchor rule alone: a baseline set one event ahead of a proto-2 client by hand; the old code sent an empty tail anchored at the event the client lacked, the new code a tail from its held last carrying it. The cold-tab gate's handshake test pins that the gate's status and the strip land under the `.targeted` slots, and the /perf collector's road suffix is pinned in its own unit test (frames with the road land under the suffixed slot, the pusher's under the plain one).

No served browser test is included: the change is wire-side only, and the lab kernel has no trigger for a targeted push on a held session. The follow-up that would pin the client-visible half is a WindowLab road that fires the push through a test-only hook and asserts no DOM-clearing rebuild and an unchanged top visible row; that is distinct from the history labs' bottom-reachability assertion, which this fix does not address.

Modules run one at a time under the devbox lock after the review folds, each green: tests/test_chat_skeleton_reconnect.py (32 passed; against the unchanged kernel it fails the seven tests named above, and the two ordering tests fail against the first cut of this branch), chat_pages (34 passed, 27 subtests passed), perf_stats (63 passed), perf_bench (49 passed), chat_build_sig_inputs (65 passed, 10 subtests passed), feed_memo_inputs (31 passed), kernel (460 passed, 12 subtests passed), kernel_pusher_snapshot (31 passed), chat_window_spans (8 passed, 4 subtests passed), tail_lo_leading_note (9 passed, 7 subtests passed), cold_tab_gate (22 passed, 4 subtests passed), chat_delta_resync (12 passed), kernel_opening (7 passed), chat_empty_build_guard (5 passed), ws_drop_loud (8 passed), view_deltas (47 passed), chat_fold (46 passed, 17 subtests passed), chat_split (55 passed), kernel_tabs_first (7 passed), kernel_push_responsiveness (15 passed), single_flight_builds (11 passed), stage_marks (31 passed), kernel_fork (19 passed). Run on the first cut and not re-run, the folds touching no backend surface: sdk_backend (265 passed, 36 skipped, 7 subtests passed), sdk_kernel (38 passed), codex_backend (119 passed, 5 subtests passed). The static pin scan over the branch found 0 lost assertIn literals (8370 checked). The extension's node tests pin three lines of the targeted push (its def line, the mark-then-push adjacency at the create sites, the backend wiring line) and the index wire's tail literal; none of them changed and no node test names a line this change removed, so the extension suite was not run for it.


ksarma:
A review of 5d68aa6c5, executed against it. One item decides ordering; the rest are prose, test pins and two counts.

## At this head

This branch's first commit is #1868's fa92e8a46, unchanged (#1868's head is that commit plus merges of main), so a merge of this PR lands #1868's change as it stands. #1868 has an open item at that commit: for a session whose built list is empty, `_send_chat_proto2` records a `{first: None, last: None}` base (kernel/kernel.py:50149 here) and re-sends the identical empty full to every proto-2 client once per dedup window; each repost counts under `chatFullWhy.empty` and files a `chatFull` client-diag row (kernel/kernel.py:50148) on the repost clock with no event behind it, and the next non-empty push reads both edges as gone and files the first content frame as `baseGone`. Executed against fa92e8a46, three proto-2 clients over an empty session through the real `km._push` loop file 3, 6, 9 cumulative `empty` rows across three reposts, and an idle empty session filed 180 rows in an hour across three tabs. Both lines are byte-identical at 5d68aa6c5.

The fix on #1868 records no base for an empty list (`st[sid]` written only when `total`, else popped), so a repost and the first content frame are `noBase` with no row, though `noBase` still climbs once per client per window (the reference's `noBase` gloss then needs the empty-frame repost beside first sends and resets). Popping the entry also changes two reconnect-only reads of the echat key set, a cold-tab gate and a reconnect resolve: a stat-able but empty transcript can then be dieted as a skeleton, worth a line in the reference. The alternative that keeps the entry reads a both-`None` base as no base in `_chat_full_reason` and gates `_note_chat_full` on a base with at least one edge. Ordering: #1868 first with the fix, then this branch after a merge of main; or the fix taken here.

## Smaller points

Stale `changeAt0` prose, one item in three places.
- docs/reference.md:1968 and the `_chat_full_reason` docstring at kernel/kernel.py:50042 (and a comment at tests/test_chat_pages.py:993), all from the carried commit, still describe `changeAt0` as the targeted push's shape. This PR's own change removes that shape: `_push_session_now` diffs against the shared baseline (kernel/kernel.py:53469) and reaches `change_from` 0 only when the session has no baseline or an empty one (`_chat_diff` returns 0 for `if not prev`); executed, the base ea3c827 produces `changeAt0` on a targeted push to a built session and this head an empty delta. Nor is that shape the targeted push's alone: the pusher's cycle (kernel/kernel.py:52943) and the connect push compute `change_from` the same way, so the pusher's first cycle over a client that took a targeted full lands under the same label. Reworded: a change at the list's first event against a held base, a genuine first-event change or a no-baseline send, by whichever sender; `sends.full.chat.targeted` splits the senders. Prose only; it can equally land on #1868.

Two unpinned inputs on the targeted road.
- `led_changed` at kernel/kernel.py:53470: forced to `False`, the four modules that exercise the push stay green (159 passed). A pin: in tests/test_chat_skeleton_reconnect.py `test_13_a_caught_up_client_gets_no_full_from_the_per_session_push`, change the field `led_changed` compares on `self.SESS[S1]` on the status-flip push (an existing push, since the closing send count counts every targeted delta) and assert the tail carries it, with the key-absent assertion on the first empty-suffix tail or after a `km._push([c])` cycle: the targeted push never advances the shared baseline, so unchanged means unchanged against the map read at kernel/kernel.py:52944, not against the previous targeted push.
- The road reset, `finally: _SEND_ROAD.name = None` at kernel/kernel.py:53477: with the two lines removed the same four modules stay green (159 passed), and the leak is real, since the create, session-branch and promote paths and the ready handler's connect push share the dispatch thread. `assertIsNone(getattr(km._SEND_ROAD, "name", None))` after `km._push_session_now` returns pins it, in test_13 (the send-loop exit) and in `test_10_the_handshake_push_resolves_the_set_before_it_decides` in tests/test_cold_tab_gate.py (the cold-skip early return).

Two counts in the body's Tests paragraph.
- tests/test_chat_pages.py gives 41 passed, 27 subtests passed at this head, not 34 (the figure before #1868's commit, whose own message reports 41); tests/test_perf_stats.py gives 64 passed, not 63, since the second commit adds a test the carried figure omits.

The wrapper without a caller.
- `_send_chat` at kernel/kernel.py:49430 has no production caller at this head (the base had one, in `_push_session_now`), and its docstring describes a road nothing in the kernel runs. Routing `_send_chat_or_status` through it (the client lock is an RLock) fails two source pins, tests/test_chat_skeleton_reconnect.py:516-520 and tests/test_chat_split.py:116, so the fix is one docstring line naming it the test-facing locked entry, with `_send_chat_or_status` as the pusher's and the targeted push's road.

The stage attribution in the sends paragraph.
- docs/reference.md:2422-2424 puts every targeted push under the `push.session` stage above. That mark is only the default wrapped onto the two backend hand-offs (kernel/kernel.py:18377 and 18501); the in-kernel callers the sentence names call `_push_session_now` bare from a request handler and run under the request's own `http.<METHOD>.<route>` mark, as the reference's definition at 2266-2270 and tests/test_stage_marks.py pin. The `.targeted` suffix is independent of the mark, so only the cross-reference is off: split the parenthetical into the two backend callers under `push.session` and the rest under the request's route mark.

The carried pin.
- tests/test_send_pending_overlay_kinds.py:58-59 pins two of the four transient-key mint sites, the command chip in `_ack_cmd_chip` and the Codex echo in `CodexBackend.send`, by source text alone: a key rewritten before it reaches the event leaves the module green (3 passed) while executed probes through the existing harnesses catch it. Twins that fit, each driving the mint in its harness and asserting `km._transient_key(km._event_key(...))` on the built event: the chip in tests/test_queued_copy_identity.py `TheEchoLandingKeepsTheProto2Base` (its `_World` harness); the Codex one in tests/test_codex_backend.py's `build()` harness, with the kernel source loaded for `_transient_key` or the assertion placed in a module that has it. tests/test_sdk_backend.py already executes the chip's `cmd:` prefix through `set_model`, so the Codex twin is the one adding coverage. The literal lines can stay as the census guard; the chip literal duplicates tests/test_kernel_cmd_gesture.py:78. Also on #1868.

Not listed: the base-holding road at kernel/kernel.py:53469 (an index-wire client whose base came from the push's own full, no pusher baseline) is pinned by tests/test_kernel_opening.py `PushSessionNow.test_idempotent_with_the_periodic_pusher`, present at the base and untouched here; the mutation at that line fails it with 467 others green across the 16 modules that reference `_push_session_now`.

## Tests

The modules run for each item above, with their results, are in the comment that follows.

ksarma:
The test runs behind the review of 5d68aa6c5 above, by item; every run was in a scratch archive of that head or in its worktree, with the kernel's ports and session variables unset, unless the base or #1868's commit is named.

- Body counts: `pytest -p no:cacheprovider tests/test_chat_pages.py -q -n 6` at 5d68aa6c5: 41 passed, 27 subtests passed. tests/test_perf_stats.py at 5d68aa6c5: 64 passed in 3.40s; `git diff fa92e8a4..5d68aa6c5 -- tests/test_perf_stats.py` shows the added test.
- `led_changed` (kernel/kernel.py:53470 forced to `False`): tests/test_chat_skeleton_reconnect.py, tests/test_cold_tab_gate.py, tests/test_perf_stats.py, tests/test_chat_pages.py together, `-n 6`: 159 passed, 31 subtests passed (the same four unmutated: 159 passed, 31 subtests passed); the 13 other modules that call `_push_session_now`: 350 passed. The proposed assertions in tests/test_chat_skeleton_reconnect.py test_13: pass at 5d68aa6c5, fail with `led_changed = True` and with `led_changed = False`.
- Road reset (the `finally` at kernel/kernel.py:53476-53477 removed, `ast.parse` clean): the same four modules, `-n 6` and serially: 159 passed, 31 subtests passed both ways. `assertIsNone(getattr(km._SEND_ROAD, "name", None))` after `km._push_session_now` in test_13: fails on the mutant, passes at 5d68aa6c5.
- Base-holding road (the finding not listed): `change_from` at kernel/kernel.py:53469 mutated, `pytest -n 6` over the 16 modules that reference `_push_session_now`: 1 failed, 467 passed; the red is tests/test_kernel_opening.py::PushSessionNow::test_idempotent_with_the_periodic_pusher (AssertionError: 3 != 2, unchanged payloads dedup, no re-send storm), a test present at ea3c827 in a module this PR does not touch.
- `changeAt0`: scratch archives of 5d68aa6c5 and of ea3c827, real `_push` then `_push_session_now` over a based proto-2 client under a capped scope; the base yields `changeAt0` on the targeted push to a built session, the head an empty delta, and the head yields `changeAt0` again once `_prev_chat_events` has no entry for the session.
- `_send_chat`: an AST walk over kernel/*.py at 5d68aa6c5 (no Load-context reference outside tests) and at ea3c827 (one, in `_push_session_now`); `_send_chat_or_status` rewritten to end in `_send_chat(...)`: tests/test_chat_skeleton_reconnect.py (the pin at 516-520) and tests/test_chat_split.py (the pin at 116) go red.
- Stage attribution: `grep -n push.session kernel/kernel.py` at 5d68aa6c5: 18377 and 18501 only, both `_stage_default` wrappers; a scratch pytest calling the real `_push_session_now` under a set mark and under none: the frames carry the caller thread's mark or none.
- Transient-key mints: tests/test_send_pending_overlay_kinds.py with `_ack_cmd_chip`'s key rewritten to `"chip-" + uid` and `CodexBackend.send`'s to `echo_uuid.replace("echo-", "cx-")`, the pinned lines byte-identical: 3 passed. Executed probes, the `_World` harness in tests/test_queued_copy_identity.py plus `_ack_cmd_chip`, and tests/test_codex_backend.py's `build()` plus `send`: 2 passed each at 5d68aa6c5, red under the mutations.
- The #1868 item, executed against fa92e8a46: the real `km._push` loop with three proto-2 clients over an empty session, the dedup slot back-dated between pushes: 3, 6, 9 cumulative `empty`-labeled `chatFull` rows; a sender-level probe of an idle empty session over an hour: 180 rows across 3 tabs. With `st[sid]` written only when `total` (else popped): reposts are `noBase` with no row, and `sum(chatFullWhy) == sends.full.chat.count` holds; the existing counter test's `empty` arm, which sends to a client already holding a content base, still reads `empty == 1`.

lczh:
Follow-up for this review: #1895 (an empty chat list records no proto-2 base, so its reposts and the first content frame are noBase with no row; the chatFull row is appended after the client lock; the mint, label, flag, doc, led_changed and road-reset pins; the changeAt0 prose; the _send_chat docstring; the reference's stage attribution).

lczh:
A second follow-up for this review's no-baseline shape (a targeted push before the pusher has built the sid sends a full to every based client; measured live at 42 such frames in the three minutes after a kernel restart): #1909 seeds the shared baseline at the first whole frame a sender delivers, with a race detector and a per-session raced mark so no single-client sender can strand another holder.

PR #1872, opened 2026-09-19
POST /new echoes a refused effort level as refused, never as applied, and romp new says so with the kernel's reason

ksarma:
Stacked on the parked-effort drain PR (#1865) and lands after it; that PR is in turn stacked on #1863, the settingRefused frame. This PR's own change is the one commit on top of that head: the POST /new leg of the effort setter's verdict and the `romp new` reader of it.

## Symptom

`romp new --effort <level> <name>` against a session whose backend refuses the level (a Codex model whose catalog does not list it, an SDK level outside its list) prints `romp new: applied model ..., effort <level>` and exits 0, while nothing changed: the session keeps its previous level and no line says so.

## Cause

`_set_effort_or_park` returns `(took, parked)` so a caller can be loud about a refusal (`took` is False when the backend refused the value), and the setEffort op in `_drive` and the typed `/effort` route in `_route_meta_command` both read it since #1814. `_apply_new_session_prefs`, the leg POST /new runs for its optional `effort`, was the one caller that still discarded the verdict: it called the setter and echoed the asked level as `effort` whatever the backend said, on the fresh create and on the idempotent existing-open arm alike.

On the CLI side, the `romp new` reader in `bin/romp` reports the echo's `model`, `effort`, `env` and `tags` and keys its "did not acknowledge ... (older kernel?)" warning on the effort echo's PRESENCE. It reads no refusal, so a kernel that answered one would have drawn that older-kernel warning in place of its answer: a false protocol gap.

## Fix

`kernel/kernel.py`, `_apply_new_session_prefs`: the effort leg unpacks `took, _parked = _set_effort_or_park(...)`. When `took`, `out["effort"] = e` as before. Otherwise `out["refused"] = _effort_refusal(be, e)`, the same sentence the setEffort op and the typed route answer with (the Codex catalog wording when the session's backend is the Codex one, the backend wording otherwise), and one stderr line in the typed route's shape with the door named: `effort %r for %s refused by %s (POST /new)`. The response stays 200 `ok: true`: the open itself stands (the session runs or was just created), one preference was refused, and a headless script must not read its create as failed. The `refused` key rides `extra` like `model`, `effort` and `env`, so the handler's three arms (fresh create, existing open, comment thread) need no change. The function's docstring records the echo's meaning and that the verdict used to be dropped here; `_effort_refusal`'s docstring now names every door that answers with its sentence (the setEffort op, the typed route, the drain of a parked pick, this prefs pass). No new setter, no new state, no gate.

`bin/romp`, the `romp new` echo reader: after the applied line, a `refused` string in the echo prints one stderr line, `romp new: effort <asked> refused: <the kernel's reason>`. The asked level is then left out of the unacknowledged-ask check (`_effask`, empty when a refusal arrived, replaces `$effort_arg` in that check's arguments), because a refusal is the kernel's ANSWER to `--effort`, not a dropped key. The applied line keeps reporting only what the echo names, on stdout, and the exit status is unchanged because the session exists.

Additive on the wire. A kernel with this change against the previous CLI prints the older-kernel warning for a refusal (misnamed but loud, in place of a false applied line); this CLI against a previous kernel behaves as before, since no `refused` key arrives and the dropped check is unchanged. Only the effort leg changes: `_set_model_or_park` returns `parked` alone, so the model echo is out of scope here.

## Tests

- `tests/test_new_route_prefs.py::NewRoutePrefs::test_a_refused_effort_is_not_echoed_as_applied_and_the_echo_carries_the_refusal` drives the real Handler over HTTP on the existing-open arm, with the effort setter stubbed to answer `(False, False)` while recording the call and one fake object installed as both the session's backend and the Codex singleton (so `_effort_refusal` picks the catalog wording). It asserts the 200 is `ok` and `existing`, the model is echoed, no `effort` key is present, `refused` equals `_effort_refusal(fake, "bogus-level")` and carries the catalog wording, the setter was asked once after the model setter, and that exactly one captured stderr line names the refusal and reads `effort 'bogus-level' for <sid> refused by object (POST /new)` (the fake backend is a bare `object`), which pins the line's shape and the door suffix. Before the change the response carries `effort: "bogus-level"` and no `refused` key and nothing reaches stderr; `assertNotIn("effort", r)` fails first. With the leg as it was before this change and these tests: 1 failed, 34 passed over the module. Green after: 35 passed. Dropping the ` (POST /new)` suffix alone or the `refused` write alone turns the case red.
- The class's setUp stub for the effort setter now answers `(True, False)`, the setter's shape. This is part of the fix's test: the old stub returned None, and once the leg unpacks a tuple the two existing cases that post an effort (the existing-open reassert and the fresh create) answer 500 through `do_POST`'s outer `except` and fail on `urlopen` (measured: 2 failed, 33 passed with the old stub against the new leg); the two cases that post no effort never reach the leg.
- `tests/romp.bats` "new --effort: a kernel that REFUSES the level says so with its reason, never as an unacknowledged ask": a fake kernel answers `model` and `refused` with no `effort` key. The case reads stdout and stderr apart (stderr to a file; `run_romp` merges them) and asserts exit 0, the started line on stdout, the request carried the level, exactly one stderr line containing `refused`, which reads `romp new: effort turbo refused: the catalog for this model does not offer it`, nothing about the refusal on stdout, neither `did not acknowledge` nor `older kernel` on either stream, and an applied line that names the model alone. Before the change `bin/romp` prints the WARNING that the kernel did not acknowledge `--effort` (older kernel?) and no refused line; the one-refused-line count fails first. With `bin/romp` as it was before this change and this test: not ok 3 of 3 under the filter. Green after: 3 of 3 ok. Printing the refused line on stdout instead of stderr, or leaving `--effort` in the unacknowledged-ask check, turns the case red.
- The acking case gains `[[ "$output" != *"refused"* ]]`: a level the kernel TOOK is reported as applied and never as refused (green before and after).
- Full `tests/romp.bats`: 116 of 116 ok, 0 skipped, on the changed tree.
- Full pytest on this head (`tests/`, `-n 10`, `-p no:warnings`, `tests/test_cut_turn_tree_kill.py` excluded because it ends the process tree it runs in on this machine, the extension's node_modules present so the served-page legs run, short `/tmp` TMPDIR): 14303 passed, 11 skipped, 13 failed, none in a file this change touches. Two are SDK-gated cases in `tests/test_sdk_backend.py` red at the base too (`OptionsAssembly.test_ultracode_is_a_choice_everywhere_but_never_the_seeded_default`, `ReconnectReconcilesInflight.test_reconnect_settles_a_turn_stranded_by_the_teardown`). The other eleven are timing-sensitive under a ten-worker run on this machine and passed alone on the same head straight after (11 of 11): the seven `tests/test_tab_lock_browser.py::ServedTabLock` cases (one served page's worker under load), `tests/test_judges_process.py::ChildRoad::test_one_pass_through_the_child_feeds_perf_and_keeps_the_kernels_bookkeeping_in_order`, `tests/test_kernel_remote_identity.py::RemoteIdentity::test_second_alias_absorbs_the_old_row`, `tests/test_postal_via_dedupe.py::PeerRoutePrefersDirect::test_resolve_recipient_lands_direct_with_no_ambiguity_error` and `tests/test_session_host_restart.py::ServedRestart::test_with_hosts_on_the_turn_survives_the_restart_and_with_hosts_off_it_is_cut`.

Tier: fix


ksarma:

This branch was cut from the original commit of #1865 and does not carry that PR's two later review-fix commits (39f59db74 and 994347a39), which route the fast-toggle cases in `tests/test_kernel_meta_command_gate.py` through a session no Codex backend vouches for. Since #1864 landed, main refuses a vouched `/fast` on a Codex session before the gate parks it, so merging main into this branch while #1865 is still open turns three cases in that module red: `test_a_setter_answering_none_at_the_drain_is_not_reported_as_a_refusal`, `test_the_command_route_answers_a_refused_fast_toggle_on_the_same_frame` and `test_a_parked_fast_toggle_the_backend_refuses_at_the_drain_is_said_the_same_way`, each failing at its `state["queued"]` assertion (lines 235 and 272) with `KeyError: 'queued'` and the refusal line on stderr.

Merging #1865 first, or merging its current head (4dff63fdc) into this branch, avoids it. This branch's own commit does not touch that test module, so either order leaves the file as #1865 has it.

Checked in a scratch tree against this head: with main merged in alone, `pytest tests/test_kernel_meta_command_gate.py tests/test_sdk_registry_blind.py tests/test_codex_slash_guard.py tests/test_model_fallback_flag.py` gives 3 failed, 78 passed; with #1865's head merged in as well, 81 passed.

PR #1873, opened 2026-09-19
The model pickers mark the requested model with a yellow tick while an automatic fallback answers, with the cause and the retry state under it

henrypinkard:
Tier: feature

The model pickers (the chat statusline's badge, the comment popover's picker, and the timeline's lane picker) mark the model the user asked for with a second, yellow check while an automatic fallback answers instead, with a permanent one-line sub-line under that row saying why and whether the kernel is retrying. Re-filed from #1810, which was closed so that its later additions could be reviewed on their own; the author's six commits are cherry-picked as they were (authorship preserved). Stacked on the two model switches (Always fast, Retry upgrades after downgrades), which it depends on for the retry state it reports.

**What it adds.** A requested-model row in each picker (`.meta-item.requested`), its sub-line reading, for example, "requested, blocked by safety classifiers (bio), retrying every 10 min", or the retry switch's state when nothing is armed; the refusal's category from the CLI's refusal frame; a standing fallback's cause read off the CLI's transcript tail at a host attach when the kernel has none on record (bounded, off the loop thread, persisted, said once); a re-observed fallback (a replay at attach, a reconnect's init) is not a new episode and keeps the cause; the ten-minute retry re-armed at every host attach for a fallen session with the switch on. Registry fields: `servedModel`, `fallbackCause`, `fallbackCategory`. Nothing is written into any session.

**Noted, not blockers**, for a later round: the sub-line and tooltip say "auto-retry" where the switch is named "Retry upgrades after downgrades"; a re-fallback for a different cause keeps a stale safeguards cause; the requested mark reuses the working-state yellow for a meaning that is not "working"; the attach-time transcript read reports no misses; the tooltip builder is duplicated into the timeline view (the timeline inlines its twins by design).

Tests: `tests/test_model_fallback_flag.py`, `ui/webview/model-requested-tick.test.ts`, `ui/timeline-requested-tick.test.ts`, plus pin updates.


ksarma:
The evidence below was executed against 01b1ebee8; 98864e389 is the same six patches rebased onto main (identical per-commit patch-ids), so every finding holds there, and line numbers are on 98864e389.

Two things to change first.

**The chat and popover pickers' requested family row has no sub-line.** ui/webview/render.ts 15757-15770 appends the sub-line, then the label chain's else branch at 15789 assigns `item.textContent`, replacing every child. Executed against the head with `toggleMetaMenu` lifted over a DOM: the family row has no sub-line child in the chat statusline or the comment popover, for a version pick and for an alias pick, while the version submenu row keeps its one. The PR's ui/webview/model-requested-tick.test.ts passes 5/5 over this because its pins (lines 50-52, 63-73) are source regexes. Fix: move the whole requested block to just after the label chain, keeping `const fb` with it since the submenu's closure reads it; the caret is absolutely positioned (ui/webview/styles.css 1714), so the move is layout-safe, where guarding the else branch instead would double-append if a model row ever gains a real sub-line. With the block moved both pickers gain the sub-line and the existing tests stay green. Replace the line-50 pin with an executed assertion through `liftMenu` in ui/webview/codex-meta-choices.test.ts that the requested family row has a sub-line child, in the statusline and in the popover.

**The retry state carries a per-second countdown.** kernel/sdk_backend.py 5094 computes `nextIn` from the clock at every build. Measured scope: while an upgrade retry is armed (switch on) and a chat client is connected, the chat status frame is byte-distinct every second and re-sent to every client each pusher cycle for the arm's whole life, and the chat payload is rebuilt each cycle because `_chat_row_sig` (kernel/kernel.py 33501) keys the field; signatures one second apart differ while the displayed minute count does not. The comments frame does the same for an open thread's own live session only; the timeline skeleton is not re-sent per second, since the lanes' view signature does not key `modelFallback` and rides the 5-second bucket. Fix: carry the arm's deadline as `nextAt`, drop the countdown, and compute the minutes from the client clock at render time in the tooltip, the sub-line and the timeline twin. Update the `ModelFallback` interface and the three pinning test files; add a differential case to tests/test_chat_build_sig_inputs.py (unchanged signature when only time advances under a fixed arm) and a two-rows-one-second-apart case to tests/test_payload_dedup_invariant.py. Adding the field to `_CHAT_ROW_UNKEYED` or stripping it from the dedup signature is not enough: the build still reads it and serves a stale value.

Also at this head:

- The mark vanishes at every init. All three pickers gate it on the current row, `status.model`, which an init or reconnect learn resets to the configured pick: executed kernel-side and in the menu, the blue tick moves to the pick and the yellow mark and sub-line disappear until the next fallen reply, at each turn start, each reconnect and the ten-minute retry's own init. Reporting the served model in the status while a fallback stands beats reading `fb.live` in the pickers (render.ts 15757 and the timeline twin), since the badge text, popover model, lane label and registry listing read the same field. Check `_do_refresh_context` too: it runs after every turn's result, and if the context-usage report carries the configured pick the mark is gone for the whole idle stretch between turns. Land it with an executed menu test and a kernel test.

- The timeline row's caret drops to a third line. ui/romp-timeline-view.js 3363 makes the requested row a column flex, so the caret at 3443 becomes a further flex item: two headless-Chromium measurements of the real `_openMetaMenu` give about 55px against 25px for a plain row, the caret alone at the right edge. A column wrapper around label and sub-line with the item left a flex row (preferred: it matches the chat's head div and leaves plain rows untouched), or an absolute caret as in styles.css 1714, each give a 39px row with the caret beside the label. The regex pin at ui/timeline-requested-tick.test.ts 40 passes with the defect present; re-aim it and add a layout assertion.

- A sentence row under nowrap. `requestedModelSub` (render.ts 15580) returns a sentence, the menu row is nowrap with no width cap (styles.css 1695), the empty row's wrap-and-cap treatment at 1701 is not applied to the requested row, and the timeline inlines the same nowrap although its own empty row (3472) wraps. Measured in Chromium with the bundled font: the family row is 321 to 361px wide, its left edge 9 to 11px off-screen in a 360px pane, and the Latest submenu is off-screen below roughly 870 to 900px of pane width, since it opens to the left of the widened menu. The wrap rule fixes the main menu at every width and the submenu from about 600px up; below that the submenu collapses to its 96px min-width (placed by right alone, it shrink-to-fits), so also pin the submenu's measured width and clamp its right offset to the pane width minus 8px minus that width, and give the timeline's requested row the wrap and cap its own empty row already has. Latest could also keep its own line rather than being joined onto the sub-line (render.ts 15819, timeline 3393).

Follow-ups:

- The lanes' view signature (kernel/kernel.py 53752, row tuple at 53808) does not key `modelFallback`, so a lane's mark lifts on the next transcript-mtime move or the 5-second bucket rather than the serving cycle; the chat is fine, since the message's tail forward wakes the pusher. Key a reduced form (never the deadline) with a differential test.
- `_arm_if_below_pick` (sdk_backend.py 6603) runs at every host attach. It arms through `_arm_upgrade_retry` (6619), whose log line at 6632 says "fell back to", so after a restart that line names a fallback that did not happen then, and the attempt count resets (three sessions built from one persisted registry row each logged the identical line). Give the attach its own wording, pinned via captured logs; persisting `since` and the attempts is not worth a write, as neither is rendered.
- `_reconcile_stranded` (5985) zeroes the in-flight count without clearing the refusal latch (cleared only at the settle, 8513), so the first unsettled turn after an abandoned one, and only that one, carries the old episode's safeguards cause and category on a new downgrade (reproduced: an abandoned turn, then a served downgrade, returned the old cause in memory and in the registry). Clear `_refusal_this_turn` there (and the swap-cards list), with a test.
- `set_model` clears the fallback fields in a second locked write (14864) after the one setting the model, and the served-model learn writes alone at 7324. A wrapped `write_reg` counted two writes per pick and per served downgrade learn, with a stale mark readable between a dormant pick's pair. Merge the clears into the first write; in `_learn_model`, merge the served model into the early branch's write and into the fields dict otherwise, since the served-on-the-pick branch at 7331 returns (7340) before that dict is built (one write everywhere in a copy, 19 passed).
- tests/test_model_fallback_flag.py: lines 100-101 and 180 pin behaviour by source text and stay green under mutations that executed probes catch (a real session fed a result-message double; a wrapped switch reader counting one call per listing; 21 passed unmutated, and under the mutations the pins still pass at 19 while the probes fail). The kernel test at 306 names three sites and asserts four literals (rename it to match); a probe feeding the builders a live row catches a duplicated key and a replaced else-branch in `_comments_frame` that the literals miss (it needs a real kernel load, so a sibling module). That else-branch is a coverage gap only, not a likely raise, since `fallback_row_for_reg` guards a non-dict registry row. The hedge at 235 masks only the missing-row shape; use the direct form of lines 81-82.
- docs/reference.md 331-338 describes only the tooltip and names the statusline and the timeline lane; say the row carries a one-line explanation with the tooltip as the long form, and add the comment popover's picker in a short clause.

Landing order, a fact rather than a finding: #1863 merged on 2026-09-19 with the hunk on ui/webview/codex-meta-choices.test.ts that adds five stub parameters to `liftMenu`, so GitHub now shows this PR conflicting with main in that file (lines 311 and 315-317); #1865 and #1872 carry the same hunk. The resolution that passes (22/22 on the merged tree) keeps main's `opts.vscodeApi ?? null`, the five trailing stubs and the widened return object. Its parameter list reads

```
"setTip", "pruneTip", "liveSession", "pendingFlags", ...
```

with a `() => {}` argument between the setTip stub and the liveSession lookup; dropping the argument gives 11 failures of "liveSession is not a function", dropping the parameter 12 of "pruneTip is not defined".

ksarma:
A note after the merge. The evidence was executed against c980a40ed, the merged head, following the review comment above (its lines were on 98864e389); every line here is on c980a40ed; between the two heads the diffs on the files named are empty or leave these lines unchanged.

**The requested family row has no sub-line in the chat and popover pickers.** ui/webview/render.ts 15764-15776 appends the sub-line, then the label chain's bare-text branch at 15795 assigns `item.textContent`, replacing every child. Executed over a DOM, both pickers' family row has label and caret only; the version submenu row keeps its sub-line. Moving the requested block to just after the label chain gives one sub-line child in both (135/135 and 46/46 green); `const fb` moves with it, since the submenu opener's closure reads it. The regex pin in ui/webview/model-requested-tick.test.ts passes over this; an executed `liftMenu` assertion on both family rows would not.

**The retry state is a per-second countdown.** kernel/sdk_backend.py 5098 emits `nextIn`, a countdown from the clock at every call; snapshots 1.2s apart give 599 then 598: unequal rows, equal rendered minutes. The chat row signature (kernel/kernel.py 33800) keys the field and the dedup signature (48730) strips only the clock and build-id fields, so while an upgrade retry is armed the chat status frame differs every second and is re-sent to every client each cycle, as is the comments frame for an open thread's own live session; the timeline skeleton rides the 5-second bucket. Carry the arm's epoch as `nextAt` and compute the minutes at render time in the four retry-state readers (render.ts 15577 and 15588, ui/romp-timeline-view.js 1150 and 1157); rebuild the sub-line on hover or at the next status frame, or an open menu shows a stale count.

**The mark vanishes at every init.** All three pickers gate the mark on the current row (render.ts 15764, ui/romp-timeline-view.js 3350), read from `status.model`, which the init and reconnect learn set to the configured pick while the fallback stands. Executed through `_learn_model` and the real init branch, all three lose the yellow tick and sub-line at each init, reconnect and turn-end refresh until the next fallen reply. Reporting the served model in the status while `modelFallback` is set fixes badge text, popover, lane label and dormant registry branch together. A third site needs it: `_comments_frame` (kernel/kernel.py 17571) builds the popover's model from the registry's live model, not the snapshot, and has to prefer the served model while a fallback row stands.

**The timeline row's caret lands on a third line.** ui/romp-timeline-view.js 3363 makes the requested row a column flex, so the caret at 3443 becomes a further flex child. The real `_openMetaMenu` in headless Chromium, two-version family: 55.4px for the requested row, 24.8px for a plain one. A column div around label and sub-line, the item left a flex row, gives 38.6px, caret beside the label. The label is the item's own text node, so the wrapper takes it in; wrap only the requested branch (ui/webview/codex-meta-choices.test.ts reads plain rows' children); re-aim the pin at ui/timeline-requested-tick.test.ts 40 and add a layout assertion.

**A sentence row under nowrap.** `requestedModelSub` (render.ts 15586) returns a sentence into a nowrap row with no width cap (ui/webview/styles.css 1695; the empty row's wrap-and-cap at 1701 is not applied); the timeline inlines the same nowrap at 3353; its menus are placed by left, so wrap and cap suffice there. Measured in Chromium: the version submenu is off-screen today under 870 to 900px of pane width; the family row, once its sub-line renders, is 321 to 361px wide, 9 to 11px off-screen in a 360px pane. The wrap rule alone is not enough in the chat picker: placed by `right` (render.ts 15871), the submenu shrink-to-fits the room left of the menu and collapses to its 96px minimum; measure and pin its width after the wrap rule applies, then clamp the right offset to pane width minus 8 minus that width.

### Smaller points

- The lanes' view signature (kernel/kernel.py 54268) does not key `modelFallback`, so a lane's mark lifts at the next transcript mtime move or 5-second bucket, not the serving cycle (the chat is fine). Key a reduced form, never the deadline, with a differential test.
- `_arm_if_below_pick` (kernel/sdk_backend.py 6603) runs at every host attach and re-arms through the fallback path, whose log line at 6632 says a model fell back: three sessions from one persisted registry row each logged it, attempts reset to 0. Give the attach its own wording, pinned through captured logs.
- `_reconcile_stranded` (kernel/sdk_backend.py 6021) zeroes the in-flight count without clearing the refusal latch, so the first served downgrade after an abandoned turn, and only that one, keeps that turn's cause and category in memory and the registry. Clear the latch and the swap-cards list there, with a test.
- `set_model` clears the fallback fields in a second locked write (kernel/sdk_backend.py 14864 and 14872) after the one setting the model, and the served-model learn writes alone at 7324: two registry writes per pick and per served downgrade learn, a stale mark readable between a dormant pick's pair. Merge each into the one write; in the learn, into the early branch's own write, which returns before the fields dict is built.
- docs/reference.md 331-338 describes only the tooltip and names the statusline and timeline lane; the row's one-line explanation and the comment popover's picker are missing.

ksarma:
Test points, continuing the note above; lines are on c980a40ed, and tests/test_model_fallback_flag.py is byte-identical to 98864e389.

### Tests

- Lines 180 and 100-101 pin the settle's latch clear and the once-per-listing switch read by source text. Under mutations that move the clear or re-read the switch per row, the pins stay green (19 passed) while executed probes fail: a real session fed a result-message double, and a counting wrapper on `retry_upgrade_on` over one listing with two rows.
- The kernel test at 306 names three sites and asserts four literals (rename it to the count). A mutation that duplicates a key after each literal and replaces the comments frame's else branch passes all 19 tests, while a probe feeding a fallback row into `build_session` and the timeline builder fails 3 of 4; that probe needs a real kernel load, so a sibling module. The else branch is a coverage gap only, since `fallback_row_for_reg` guards a non-dict registry row.
- The hedge at 235 compares the literal to itself when the listing lacks the row; inverting the alive filter leaves it passing, while the direct form at 81-82 fails. Only the missing-row shape is masked: a listed row without the field raises before the comparison.

PR #1875, opened 2026-09-19
Design: action kinds and the held-mail card on the notice model (plans/notice-cards.md)

henrypinkard:
The design section for the user's ask (2026-09-18, through the manager): a notice card's actions are of a KIND the backend defines (its route and body shape named by the kernel; the allowlist admits kinds, never bare routes), the held-mail kind first with Approve and Deny only and no Edit through the kernel's existing quarantine act road; the held-mail card a notice card the kernel produces when a message is held, replacing the hand-built quarantine family; the card's modal showing the message text and never a struck-through name; a chat-pane approval box above the background-agents box listing the session's held messages and vanishing on the decision; a held message putting the session in the blocked state (the dashed yellow ring) until decided.

The section (plans/notice-cards.md, 'Action kinds and the held-mail card') checks each premise in code at upstream/main 48c313a5, corrects one (the struck name in the screenshot is the recipient's, struck because every notice card is stamped live False), and decides the stored action shape and the kernel's kind table (send, quarantine, the settings plan's kind), validation at the post and exact matching at the click with the one click-time input a kind may name, the held-mail card's fields against a census of today's card, who posts it and when (the kernel's backfill from the bus's held files at the notice attach, idempotent over the live rows and the revision index), how it moves (the acted and expire rows on a decision, Clear as today), the liveness fix, the removal census with its pins, the chat frame's per-session notices slice and the box above #bg-tasks with the same stored actions over the chat's socket, the ring's pin, phasing and tiers (this docs PR; the kind API with the card, the liveness fix and the removal as a feature PR; the chat box and the ring's lab as a feature PR, or one if the size stays readable), what stays out, and privacy. Two deferred bullets of the plan are updated to point here. Synthetic examples only.

Tier: docs


ksarma:
A note after the merge of a5be93a98, checked against that head's code. The first item shapes #1885, whose description keys the card on the mid.

Lines 797 and 801 key the card on the mid; line 197's rule admits 64 characters, and a cross-host px- mid (postal_service.py 2936) is 48 fixed characters plus the sender's pid digits and host name: 65 for a 7-digit pid and 10-character host. `post_notice` accepted 64 and refused 65 through 69; a held mid is never resent, so without today's family nothing shows it, and a 63-character TESTHOST mid misses it. Fix: key on a hex hash of the mid (a counter breaks the backfill's restart check), the mid in a field and the action body; test a 65-character px- mid.

- Line 814 is true of the kernel (`test_cleared_card_is_hidden` passes), false of the pane's corner and session Clears, withheld on held-mail cards and granted by the notice kind's CLEAR, so a Clear hides the message everywhere while its file stays. Fix: a kind-level noClear, the kernel's cleared-id skip kept, pinned.
- Line 807's reason is false for deny: deny for a gone recipient succeeds on the bus and approve's refusal advises deny; an owner-less no-actions card drops deny, but only for a toId with no names-registry row, since `_notice_session_known` counts ended sessions and renames keep sids. Fix: post under any registry-known toId (Approve draws the bus's refusal), owner-less otherwise; test both.
- Lines 795-796: `fillNoticeFace` appends every body node and .fask-nbody has no clamp (5690 bytes rendered whole); with the route line first, the first ninety characters are the route. Fix: a face line clamp, modal unclamped, both sentences agreeing, pinned.
- Line 825's census misses quarantine-route.test.ts, feed-awaiting-swirl.test.ts's append pin, feed.css's blocks, docs/reference.md 2435 and docs/guide.md 923; card-notify.test.ts pins none of this; tests/test_notice_card_store.py 922 does.
- Line 734 credits `_notice_cards`. Today's struck name comes from `_quarantine_cards` (also live False), on face and modal.
- Line 686 calls the chat pane's list shipped. Line 719's `quarantineRefused` lacks its text member. Line 803's publish is a `Path.rename`, not os.replace.

PR #1876, opened 2026-09-19
The parked-op drain says once per hold when a stale open-turn count holds a queue while the transcript at rest shows the turn closed

henrypinkard:
The parked-op drain says once per hold when a stale open-turn count holds a session's queue while the transcript at rest shows the turn closed, and retracts when a later version of the transcript shows a turn open. The belt pulled out of #1838, as its own fix. Round two: the belt reads the count alone (never the composite busy signal), its transcript read runs on the jobs thread, a cancelled chip clears it, and its row agrees in number. Round three: the decision and the write run under the queue lock against the live hold, so a hold that lifts during the parse files nothing and leaves nothing behind.

**The defect.** #1838's shape: a host counted a message folded into a running turn as its own open turn, every kernel adopted the count at its attach, `busy()` read it, and the drain parked every send in silence while the page read Ready. #1838 fixed the count; #1849 lets a host re-exec into fixed code. What was still missing is the belt: when the working gate holds a queue on evidence the transcript contradicts, nothing said so anywhere.

**The fix** (`_mark_held_working` on the drain, `_held_working_pass` on the jobs thread, kernel/kernel.py; `SdkBackend.count_says_open`, kernel/sdk_backend.py). When `_working_now(sid)` holds a sid with parked ops, the drain records the hold and nothing more. The jobs pass, once per pass, reads for each recorded sid the one source a stale count lives in: the backend's open-turn count with nothing queued to start (`count_says_open`: inflight above zero and an empty queue; never the composite `busy()`, which also holds for a queued turn about to run and for a feeder waiting with the count at zero, holds that are correct: a parked deploy restart or a quiesce with the lease refreshed, an armed reconnect after an effort, model or auth switch, a pending rewind before it arms, the gap between `send()` enqueuing and the feeder's pop). When the count says open, the belt reads the session's transcript, the backend-agnostic evidence, and files a `pending-ops.held-working` problem row once per hold (the ledger, the kernel log, the error center's ring: a decision-shaped fault, the remedy the user's) when the transcript shows its last turn closed; the row names the session, how many items wait (in agreeing number) and the remedy (ending and reviving the session replaces the count; a kernel restart does not, since the attach adopts the count). A later version of the transcript that shows a turn open files `pending-ops.held-working-retracted`. The constraints from the review of the note this replaces, each in the code: no verdict from absence (no transcript, a parse with no turns or a parse that raises say nothing); the transcript is parsed only AT REST and once per file version (the stat unchanged from the previous pass: a just-started turn's record lands within a pass, a streaming turn's file never rests, so neither is read as closed; the parse is the kernel's shared `_parse`, memoized on the same stat); never on the pusher's thread (the 2026-09-05 hoist emptied its cycle of transcript work; `heldWorking` is a job of the jobs pass after `clearDoneNotes`, in the JOBS census and the partition pin); no `busy()` read of the belt's own (the gate's one read stays the pin in tests/test_drain_hoists.py); a compacting sid is the compacting gate's hold, not this one; the state clears when the hold ends (the gate passes, the queue empties, or the user cancels the chip in `_cancel_parked`), so the next hold says again; the belt never delivers. The drain's two gates stand on their own lines (the compacting gate first, the working gate with the marker behind it); the drain-hoists order pin and the extension's queued-indicator pin read that shape. docs/reference.md's per-session hosts section gains the paragraph.

**Round three** (the review's item on afec6fee). The pass read the count at the top of a sid, parsed with no lock for tens of milliseconds, and filed with no re-check, so a hold that lifted mid-parse (the turn settling and the drain delivering and popping both dicts, or the user cancelling the chip) still got a row saying that zero items wait, never retracted, with a said mark on an orphaned entry. Now the count is re-read after the parse, and the decision and the write run under `_pending_ops_lock` against the live queue and the live entry (the entry still present and the same object, the queue non-empty, `queued` read from the live queue); a lifted hold files nothing and leaves no orphan. Two lows: the test module and its backend test write `off` into their minted state roots' `session-hosts` (the repo rule), and the red-first line below counts what the three test files onto the previous head actually give.

**Round two** (the review's items on 7023c74f). Medium one, the premise: the belt fired on the composite `busy()` and called the feeder's correct holds a stale count; it reads `count_says_open` alone now, and the four pending states are pinned silent. Medium two: `_cancel_parked` left the belt's state behind, so a cancelled chip and a new send into the same stuck session said nothing and parsed nothing; the state goes with the queue. Promoted to medium: the belt put a transcript parse back on the pusher's cycle (40 to 96 ms per version on a 2.8 MB transcript at every tool boundary that rested one cycle, per held open-turn session); the read is the jobs thread's, and the pusher's cycle is pinned to parse nothing. Low: the row's singular verb.

**Tests.** tests/test_drain_held_working.py on the drain-hoists harness (synthetic sids, a fake backend with the count reader, `_parse` and the path stubbed, the working gate stubbed to hold; each cycle a drain cycle then a jobs pass): the belt records the file's stat on the first pass and parses on the second, files one row with the sid, the name, the queued count and "1 parked item waits", says nothing more over three further cycles with one parse in all, leaves the queue held and reads no `busy()`; an open turn at rest, a parse with no turns, an absent transcript and a raising parse each say nothing; a transcript that changes every cycle is never read, and once at rest is read once; a later version showing an open turn retracts with a row after one more parse and nothing more until the file changes; the state clears when the hold ends and a new hold says again; a compacting sid leaves no state and no row; a hold the count does not carry alone (the working gate holding with the count at zero) is neither read nor said and drains when the feeder catches up, with nothing to retract; `SdkBackend.count_says_open` against a running turn (true), a queued turn with the count at zero (false), both (false), neither (false) and a sid it does not run (none), with `busy()` unchanged beside it; a cancelled chip clears the state and a new send is read and said again; the pusher's cycles (run on a thread named for it) parse nothing across a held sid at rest while the jobs pass parses once, by count and by source, and the job sits in the jobs pass. tests/test_jobs_thread_split.py's partition pin gains the job; the drain-hoists pin is re-cut to the two gates' lines.

Red first at afec6fee (round two's head) by copy-aside: the lifted-hold drive, both variants, `Lists differ: ['pending-ops.held-working'] != []` (the row for a hold that no longer stood). Round two's red first at 7023c74f, the three test files copied aside: 21 failed in all, six of them on assertions and fifteen `AttributeError` in the jobs-split module's `_LabCycles.setUp` (its QUIET stub tuple names `_held_working_pass`, absent at that head); the six: the count-alone drive `(1, ['pending-ops.held-working']) != (0, [])` (the base said on the composite signal); the backend reader `unexpectedly None`; the jobs pass `unexpectedly None`; the cancelled chip's state `unexpectedly found` in the belt's dict; the pusher's parse `['pusher'] != []`; the partition pin's tuple lacking `heldWorking`. Round one's five reds at main 8f9d986e, all on assertions: the belt's presence, `(0, []) != (1, ['pending-ops.held-working'])` on the say-once drive, `(0, []) != (1, [])` on the no-verdict drive, `[] != ['pending-ops.held-working']` on the hold-end drive, and the pin `-1 not greater than 0`.

**Suites at the head** (054c1238): whole Python suite 14400 passed, 71 skipped, 3 failed, the three web push failures identical at the base (the pinned runner has no cryptography package); the belt, drain-hoists, jobs-split, stage-marks and perf-stats modules 129 passed; extension suite 5052 passed, 0 failed, 6 skipped at the tree of e4d90109 (nothing under ui/ changes since).

**Not here.** The count itself (#1838) and the host's code refresh (#1849) are merged; this PR only says when the count and the transcript disagree. The body makes no claim that a streamed row re-opens the kernel's adopted count: only the feeder increments it and the hello adopts it.


ksarma:
Executed against 32e2c5c3. The head 7023c74f3 adds one commit on top of it touching only `ui/webview/queued-indicator.test.ts` (a test pin on the drain's quiet gate), none of the files below, so everything here holds for the current head.

Two conditions need to hold before the belt speaks, and neither holds at this head: the count it reads must be one adopted at the hello and not yet refreshed (provenance, rather than the transcript resting across two drain cycles that may be milliseconds apart), and the closed verdict must not fire on a turn whose last atom is an auto-compaction boundary or a wrapper-only command record. Everything below was executed against 32e2c5c3; the head 7023c74f3 changes only ui/webview/queued-indicator.test.ts, so every cited line is unchanged.

**At this head**

- kernel/kernel.py:35012. The at-rest test is stat equality across two consecutive drain cycles, which the pusher runs on every wake, so they can be milliseconds apart. Between a turn's end_turn record and its ResultMessage the transcript rests closed while `inflight` is still above zero: over 896 measured turns that window is 0.215 s at the median and 0.339 s at the 90th percentile, and for 53% of end_turn records a second assistant record of the same reply lands about 5 ms later and wakes a second drain inside it. Executed with the real parse over a synthetic transcript ending in an end_turn record: two cycles file a pending-ops.held-working row, the settle drains the queue, and the gate pass at 36020 pops the state without a retraction. One bound: on the SDK backend a plain follow-up is forwarded mid-turn, so the parked op here is a slash command, a model, effort, fast, env, cwd or compact pick, or a send typed while compacting or behind an existing queue. Fix: mark the count adopted at the hello (kernel/sdk_backend.py:10418, when open_turns is above zero), clear the mark on the feeder's increment at 6556, on the settle at 8021 and on the first live streamed atom after a hello (else a legitimately open attach gets a row and a retraction), and have `_note_held_working` speak only for a marked count; and file the retracted row when the working gate passes while said is true, since the only retraction path today is a later open version, which never comes after the settle. The at-rest and no-verdict-from-absence rules are not the defect and stay. Three cases pin it on the harness: a feeder-minted count says nothing, an adopted count says once, a gate pass after a said row files the retraction.

- kernel/kernel.py:35025. The closed verdict is `_session_working`, which reads the event model's ended flag, and the event model marks a turn led by a command record or a compact_boundary with no assistant atom as ended, so with busy() true and an op parked (the same parked shapes as above) the belt files the row two cycles later and retracts it when the first assistant record lands. Executed with the real event model and drain, a compact_boundary-led turn with no continuation and a wrapper-only command turn each file the row. A census of real transcripts bounds the two shapes. Auto-compaction is universal: 37 of 37 auto compact_boundary instants parse as ended with no assistant atom, the continuation's first record lands at least 2 s later in every case (7.2 s at the median) and the file itself rests at least 2 s in 36 of them, against the 0.5 s drain backstop, and `_compacting()` reads false for an auto trigger. The command shape is narrower than every skill invocation: a typed skill whose body record lands parses open and files no row, as did both typed-command instants in the census; the row fires only for a wrapper-only turn (a custom command with a command-contents record, a built-in prompt command, a skill body written as plain isMeta text), a shape the census did not show. Fix: after the turns check in `_note_held_working`, return when the last turn has no assistant-typed atom (that ended flag is not a reply's stop_reason); retype the harness atoms from text to assistant (three of ten tests fail under the guard until then; 10 passed after) and add two legs, a lone user command atom and a lone compact_boundary system atom, three cycles each, no row and the queue still held. The docstring and docs/reference.md should then say the belt speaks only when the closed turn ends on a reply's stop_reason, which also names the cost: a turn interrupted before any reply stays silent.

**Also at this head**

- kernel/kernel.py:35281. Emptying a queue in `_cancel_parked` pops the queue and the drain hold but not the belt's state, and the drain walks only sids with a queue, so said stays true and the next hold on the same stale count files nothing until the gate passes once or the kernel restarts. Executed: cancel the only parked op after the first row, re-park, two cycles, one row where two are expected. The dead-session drop at 36107 is not a second site (36020 has already popped the sid on that iteration). Fix: `_held_working.pop(sid, None)` beside the drain-hold pop under the lock, plus that leg in tests/test_drain_held_working.py; on a copy of the harness carrying the cancel legs, 2 failed at the head and 7 passed with the line, and the two PR modules stay at 10 passed.

- tests/test_drain_held_working.py:156. The empty-queue clear at kernel.py:36003 has no test: deleting it leaves 10 passed. That line covers only a queue emptied between the drain's sid snapshot and that sid's turn in the walk (a cancel landing mid-cycle); the cancel road above never reaches it. Fix: after the first row, clear the queue without delivering, one cycle, assert the sid has left `_held_working`, re-park, two cycles, a second row; this pins 36003 but not the cancel road above.

- tests/test_drain_held_working.py:167. The compacting leg clears the belt's state at line 165 first, so its absence assertion holds without the pop at kernel.py:36015; deleting 36015 leaves 10 passed. Fix: drop the clear at 165, assert the sid is present before the compacting cycle and absent after it, then set `_compacting_now` false and expect a third pending-ops.held-working row after two cycles; the corrected leg fails on both assertions with 36015 deleted.

**Smaller points**

- kernel/kernel.py:35025 via the working signal's last clause at 1448 (`not _suspended_after(...)`): an open turn whose last activity predates a recorded host suspension reads as closed and gets the row, then a retraction. Bounded: each resting version is parsed once, so this fires only when a version's first parse lands after the suspend is recorded (producer tick 3 s, drain 0.5 s): a follow-up parked on wake or within about a second of the lid closing, or a kernel restart after the wake. Fix: judge closure from the turn itself (ended, or an idle atom at the tail after trailing command atoms), ideally one `_turn_closed` helper shared with the working signal; pin with an open turn, the suspension check stubbed true, no row.

- kernel/kernel.py:35011. The version key is the transcript's stat alone, while the memo behind `_parse` also keys on the states file, so a states-only record that flips the verdict is never re-read (executed: a tool_use tail open on file, a waiting state appended, three cycles, no row); a silence, never a wrong row. Fix: add `_stat_key_ns` of the sid's states file to the key, as the settle checkpoint at 12105 does, with a test that proves the re-read rather than the stub; each states append then resets the at-rest wait, one extra memoized parse per write, the cost the judges already pay. The body's "memoized on the same stat" is imprecise rather than false (the memo's key is a superset) and wants a wording fix in the docstring and body.

- kernel/kernel.py:34981. `_held_working` is in-memory while the queue reloads from pending-ops.json and the count is re-adopted at every attach, so a restart during a hold files the row again and a retraction owed to a pre-restart row is never filed (executed by clearing the state and reloading the queue). The re-filed row keeps the error center's per-kernel-life ring from reading empty, so state in the docstring, docs/reference.md and the body that once per hold means once per hold per kernel life and that a retraction is filed only for a row this kernel life filed, pinned by a test that clears the state with the queue intact and expects the second row.

- tests/test_drain_held_working.py:105. The queued count is asserted only for one item: replacing `queued` at kernel.py:35028 with the constant 1, deleting the plural branch, or summing every sid's queue each leaves 10 passed. Fix: park two ops and assert a count of 2 and "items" in the text, plus one op under a second sid whose path is None, which catches the summing mutation.

- kernel/kernel.py:35033. The row reads "1 parked item wait" for one item; "%d parked item%s waiting" fixes it, and the count is at least 1 here.
- tests/test_drain_held_working.py:12 and 29. `import time` and the second sid constant are unused; drop both, or give the constant the second-sid leg above.

Tests run against 32e2c5c3 with the session environment scrubbed: tests/test_drain_held_working.py and tests/test_drain_hoists.py, 10 passed unmutated, 10 passed with kernel.py:36003 deleted, 10 passed with 36015 deleted, 10 passed with the cancel pop added; a copy of the `HeldWorking` harness with the cancel legs, 2 failed at the head and 7 passed with the pop; tests/test_state_isolation_order.py, 8 passed.

ksarma:
A note after the merge, executed against 054c1238b, the merged head; every file and line is on that commit. It follows the review comment above, whose cancel-road and singular-verb items are fixed and pinned (tests/test_drain_held_working.py:224 and 124).

**Still standing after the merge**

- kernel/kernel.py:35205. The closed verdict is `_session_working` alone. Executed with the real parse and gates, an auto-compaction boundary at rest mid-turn, or a wrapper-only command turn with no reply yet, parses as ended while the count reads one open, so a parked op behind either (on the SDK backend a slash command, a settings pick or a send behind a queue or a compaction; a plain send is forwarded mid-turn) files a pending-ops.held-working row with the end-and-revive remedy, retracted once the continuation's first assistant record lands. The command shape fires for a wrapper-only turn, not for a skill whose body record lands. Fix: after the empty-turns check, treat a last turn with no assistant-typed atom as no verdict; the fixtures then need `{"type": "assistant"}` atoms plus two negative legs, a lone compact_boundary atom and a command-led turn, three cycles, no row. The guard also silences a turn interrupted before any reply, a cost the docstring and docs/reference.md should state. Placed before the open-turn read, it delays a retraction until the first assistant record; gate only the closed arm if that matters.

- kernel/kernel.py:36316. A row filed in the end-of-turn window stands after the settle drains the queue: the belt keys on the transcript resting across two passes (the 0.5 s jobs cadence), not on the count's provenance, and the gate-pass pop is a bare `_held_working.pop` with no retraction; on the harness a gate pass after a said row leaves it. Over 1254 measured turns the settle lagged 0.5 s or more in 3.7%, so about 2.6% of turn ends that have an op parked get the row, down from nearly all. Fix: when the said flag is true at this pop, file a retraction first, saying what is known there (the count cleared and the items drained; the open-turn wording would be false), and only here: the empty-queue pop at 36299 and `_cancel_parked` disprove nothing about the count. Then mark the count adopted at the hello (kernel/sdk_backend.py:10760, open_turns above zero), clear it on the feeder's increment (6868), the settle (8358) and the first live streamed atom after the hello (else a live attach files a row and its retraction), and let the belt speak only for a marked count.

These two compound: the row is filed from a state the belt cannot observe, is permanent because no retraction follows the gate-pass pop, and tells a person to end and revive a working session, advice they cannot check. The two-pass rest at the jobs cadence is a time window standing in for the event it approximates, the count's refresh; the repository's design rule prefers the event. Order: the gate-pass retraction is the smaller change and makes the row transient; keying the verdict on the count's provenance is the fuller fix.

- kernel/kernel.py:35206. The post-parse count re-read runs outside the queue lock, and the locked block re-checks the entry and the queue but never the count, so a settle between them files a row the next drain cycle orphans; the re-read is unpinned. The window is microseconds (the row lookup between them is a memo hit), and a settle after the locked write orphans the same row, so only the retraction above closes the failure; alone it changes nothing observable. Fix: read the count inside the locked block after the identity and queue checks, `_session_row` moved above, pinned by a leg whose row stub clears the count; the first backend call under the queue lock, safe as a fast lock read, which the comment should say.

- kernel/kernel.py:36311 and 36299. Neither pop of the belt's state is pinned by execution: the module stays green with the compacting gate's pop deleted (ui/webview/queued-indicator.test.ts:215 pins the text only), and tests/test_drain_held_working.py, test_drain_hoists.py and test_jobs_thread_split.py stay green with the empty-queue pop deleted, which no production writer reaches today. Legs under Tests.

**Smaller points**

- kernel/kernel.py:35191. The version key is the transcript stat alone while `_parse` also keys on the states file, so a states record closing the verdict with the transcript unchanged is never re-read: a silence, never a wrong row. Fix: add that file's stat via `_stat_key_ns`, pinned by a states-only append expecting the row.
- kernel/kernel.py:1473, read at 35205. The verdict's last clause reads a turn stranded before a detected suspension as closed with ended false, only when a version's first parse lands after the suspension is recorded. Fix: judge closure from the turn itself in a shared helper without `_suspended_after`; pin with an open turn, suspension stubbed true, no row.
- kernel/kernel.py:35145. The say-once state is a module dict while the queue reloads from disk, so a restart mid-wait re-files the row, and the docstring, docs/reference.md:1646 and the body leave that once-per-kernel-life scope unstated. Fix: state it in all three, or persist the said flag beside the queue.
- kernel/kernel.py:35145, 35156 and 35232. The comment names `_note_held_working`, which no longer exists; the since field is never read; the stderr line names no sid.
- The body's red-first count for the earlier head (21 failed, six on assertions) is one short: the merged head's test files give 22 and seven, the mid-parse lift test the seventh.

ksarma:
**Tests** (tests/test_drain_held_working.py at 054c1238b; the first two legs were run red against the named deletion in a scratch copy and green at the head)

- Line 184. The compacting leg clears `_held_working` before the compacting flag flips, so its absence assertion holds without the pop at kernel/kernel.py:36311. Drop that clear so the compacting cycles start from the said state, assert the sid is absent after the compacting cycle, then set the compacting stub false and expect a third pending-ops.held-working row after two cycles (the two the gate-pass leg already filed, plus one); an opening assertion on the said flag keeps the precondition explicit. With 36311 deleted the corrected leg fails on the membership assertion and then on the third row.
- Line 172. The empty-queue pop at kernel/kernel.py:36299 has no leg: after the first row, empty the sid's queue in place with `clear()`, one cycle, assert the sid has left the belt, re-park, two cycles, expect a second row. With 36299 deleted it fails on the membership assertion (said stays true) and then on the row count. If appended after the compacting block, reset the compacting stub first and count from the rows already filed.
- Line 123. The queued count is asserted for a one-item queue only, so `len(ops)` replaced by the constant 1 leaves the module green. Park two ops and assert a count of 2 and "2 parked items wait" in the text.
- Lines 18 and 39. `import time` and the second sid constant are unused; drop both, or write the second-sid case the setUp stub anticipates (a sid whose path is None while another is held).

PR #1877, opened 2026-09-19
The chat merges a full frame by position, never by key absence, refuses a window or a page for a session with no regions, and lands a regions-less deep link on the older wire, so an older frame or a null tail start no longer puts older history at the bottom of the view

lczh:
Tier: fix

## The defect

Two client defects in how the chat merges a proto-2 full frame, both in the dashboard's own code and both reproduced by executing it (the scratch proofs the design lane wrote are red on the same main this branch is cut from).

The first is a missing recency reading. A full frame's tail run is open-ended: it claims every event from its first turn to the transcript's end. When the frame overlapped a run the page already held, the merge decided where each held event went by key absence alone: the frame lacks it, so it must be before the frame. That is true only for held events positioned before the frame's first shared key. A held event positioned at or after that key and absent from the frame is covered by the frame's span and missing from it, either retracted (a rewind, a canceled queued message, an echo replaced by its record, a bubble the client injected) or because the frame is behind, a full built from an older list. The merge filed both kinds into a held run above the frame's tail, and the pure run merge then concatenated held-then-frame, so the client's newest rows sat above older ones: the bottom of the view showed older content, the next delta re-appended the row and duplicated it, and a later delta whose anchor resolved to the first copy truncated the frame's tail away. With the same tail start the held run was dropped wholesale, which is the right answer, but silently: nothing said the kernel had sent a frame behind what the page showed. The everyday shape of this is the echo landing after a send: whenever the kernel's cut crossed a turn edge since the echo's delta, the echo bubble was filed above the frame's tail as a phantom that rendered as a duplicated user message at the seam and persisted across later deltas.

The second is the confirmed mis-order. A proto-2 frame whose tail start the kernel could not name (tailLo null, headKnown false) leaves the page with no regions. Any page or window that arrived afterwards had a tail run invented for it at turn 0; a non-overlapping older window touched that run, the run merge found no shared key and no lower start, and the older window landed after the newest events. The bottom of the view was history, the landing put the reader on an old message at the bottom of the transcript, the head was marked known so no gap ever asked again, and the shape was permanent for an idle session.

The reload that started this investigation: the client-diag journal on the box this was diagnosed on recorded two DOM-clearing rebuilds within one second and a 7,280 px drop in the view's height fifteen seconds before the person reloaded the page. Which producer made those frames, and whether the kernel's list was shorter or the window only re-estimated its spacers, the journal cannot tell; the two mechanisms above are the ones the investigation reproduced by execution, and they are the motivation for this change. The rows it adds are what would make such an incident attributable next time.

## The fix

The merge now keeps four stated invariants, written above the branch that keeps them: the runs' events concatenated are the session's events after every mutation; a full frame is authoritative for its span, so a held event whose key the frame lacks is placed by position, never by key absence (before the frame's first shared key it is history the frame did not carry and stays above; at or after it, it leaves the model); the tail run's last event is the kernel's last as of the newest frame applied, so the resident newest row is displaced only by a frame that carries the span it sits in; and no run is ever minted with a start the kernel did not name.

A new pure helper in the regions module reads a held run against a frame and returns the prefix to keep, the rows that go, and whether the frame is behind: the frame's last transcript key (the kernel's own anchor rule, the live overlay cards skipped, mirrored from the kernel's constant with a pin that the two sets agree) is resident and a transcript row the page holds is dropped after it. The client's own optimistic and held groups never ride a held run, so the model no longer resurrects a phantom bubble mid-list. A behind frame is applied as the kernel sent it and files one frame-behind client-diag row through a named helper (with the tail start, the held run's start, the two last keys, the dropped counts and whether a client-initiated rewind is pending, so a legitimate shorter list is separable in the journal). A tailLo-null frame for a session holding regions files a regions-dropped row, as does a numeric-tailLo full onto a regions-less session that held prepended history, so the kernel's null frames become countable.

The insert of a page or a window returns whether it placed the run and refuses, touching nothing, for a session with no regions. Its callers rebuild nothing on a refusal: a page reply frees its ask and files a trail word and a locate row; a window reply puts the pre-jump origin back, consumes its ask record and files the same, with no toast, because the landing's next attempt takes the older wire. Who makes that next attempt depends on what armed the landing. A seek (the one a tab switch with an anchor arms) is durable and re-arms on every pass, so a seek-backed landing is left to it. A notch's or a comment tick's jump and the reload restore arm no seek, and the pass that asked the window nulls the armed anchor at its end, so a refusal there used to end the click silently with the pre-jump undone; such a landing is now re-armed from the ask's record, the way a placed navigation's is, and the pass runs at once, while the older wire can answer (a dead end with no wire left keeps its row and is not re-asked, so the client and the kernel never ping-pong a window). A landing into history on a regions-less session now asks the older wire, which lands by key, and never a window; the branch sits after the existing in-flight re-point and before the window ask, and does not alter the pinned index-wire branch. No kernel line and no wire field changes. The merge of a tailLo-null frame into a key-sharing held tail is left as a parked decision, as the design's refuter asked: it would reverse a documented, pinned rule and forces a re-cut of the landing-notice lab's null-frame road.

## Residuals

The older-wire landing on a regions-less session is exact but not durable. Such a session holds no runs, so any later full frame replaces its whole list: a landing that needs several older chunks is reset when a full lands between two of them (the next chunk's anchor no longer matches the first resident row, the chunk is ignored, and the walk restarts from the landing's next attempt), and history a reader paged in is dropped under them. The regions-dropped row makes the numeric-tailLo case of that drop countable; it does not prevent it. The no-shared-key straddle after an outage longer than the wire tail (the page's newest key older than the frame's oldest) still places the held run above the frame with a silent hole between them that no gap can name; today's behavior, neither fixed nor counted here. The frame-behind row files on every occurrence, which is the point of it, so a redial storm of full frames could file many of them in a minute. The cure for the regions-less shape is kernel-side: a tail run start that is never absent for a non-empty list, so every frame names its tail start and every session holds regions; that companion is the direction, not this change.

## Tests

Red first with the fix reverted and the new helpers and exports in place: main's regions module lacks the split helper and the mirrored overlay set the new node cases import, so the extension's test bundle does not build against main as is, and the served driver's bottom check lives in the window lab's driver head, which this change adds; the reds below are the committed test files against the stock render.ts (and, for the served lab, the stock regions module too), which is the property a fix needs.

Node, built with the extension's test bundle and run singly. The executed harness over the frame merge gained nine cases, every one red before the change: a behind frame whose tail start moved down left the session's events reading newest-above-older (t3, t1, t2) and filed no row; the same frame with the same tail start filed no row; the echo landing kept the echo bubble in the run; an optimistic bubble in the held tail landed mid-run inside the merged tail's array where the strip could not reach it; a tailLo-null frame dropped the held regions without a row; a numeric full onto a regions-less session holding prepended history dropped it without a row; the insert placed a window into a session with no regions after its newest events (t250, t251, t252, w96, w97, asserted before the return value so the case's red is the misplacement); and a page reply and a window reply each rebuilt the view on a refusal. The window case was then extended to run the landing pass's nulling before the reply and was red again for the re-arm (the pass never ran) until the refusal re-armed it. After the change the module is twelve for twelve, its five earlier cases untouched. The windowing pins gained one test (red on the old insert signature) and are twenty-five for twenty-five. The regions module's tests gained the helper's cases and a pin that the mirrored overlay set equals the kernel's constant member for member, fifteen for fifteen.

Served, in Chromium on the hermetic window-lab kernel over a synthetic transcript longer than the wire tail. A new lab drives three roads on fresh pages; the committed module, run once against the stock render.ts and regions module, was red on every one: re-posting the boot frame with a null tail start filed no regions-dropped row, and the focus landing on a deep row asked a window (parked at the socket) and no older chunk; a full frame two rows behind the page and eight turns down kept the tail run at 251 events instead of 249, rendered the dropped newest row above older rows at the seam, and filed no frame-behind row; the echo landing left the echo in the run (253 instead of 252) and rendered one phantom bubble at the seam. After the change all three pass, on two consecutive runs: the regions-less landing sends one loadOlder and no loadAround, arms on the older wire (its one older ask is parked at the socket for the trail read, as the landing-notice lab's null-frame road does, because the kernel answers it within milliseconds and the landing pass that follows starts a fresh trail) and lands the row; the behind frame drops its two rows, files one frame-behind row naming the held run's start, the frame's start, both last keys and the two dropped rows, the seam reads in transcript order, and a real turn appended to the transcript afterwards lands as the newest row at the bottom; the echo leaves with the frame, no bubble renders, no row is filed. The shared bottom assertion (scroll to the bottom until the height settles; the last rendered transcript row is the newest, the view is within two pixels of the bottom, the rendered rows stand in transcript order, the runs are ordered with the open-ended run last) is now the last measurement of the history-regions lab (six tests), the landing-settles lab (thirteen) and the reload-restore lab's three roads, all green; the landing-notice lab is unchanged and green (sixteen).

The wire modules and the input censuses are green and unchanged: the proto-2 served lab, the window-spans, pages, leading-note and skeleton-reconnect modules, the served-labs census, the kernel module (460), perf stats (62), the feed memo census (31) and the chat build signature census (65). The static pin scan reports 8,370 literals checked and none lost. The full extension suite runs 5,068 tests with one failure, the pre-existing KaTeX pixel probe that fails on main in this environment; the type check is clean.


ksarma:
A note after the merge (head eb7d19dfc). Evidence was executed against d430fedae, the content commit; merging main atop it touched no line named (numbered at eb7d19dfc).

The item touching an open PR: the null-tailLo branch at `ui/webview/render.ts` 17231 drops a regions-holding session's regions (a `regions-dropped` row); #1860's guard keeps them when the frame shares a uuid with the held session (a `full-frame-desync` row). The two heads merge clean, and there #1860's guard runs first and sets `keepResident`, gating the block enclosing this branch: dead code for that frame whatever lands second, #1860's pins green. On that tree, `tests/test_client_merge_guard_browser.py` (Chromium) gives 1 failed, 2 passed (the road at line 196) against 3 passed on this branch alone; `chat-proto2-exec.test.ts` stays 12 for 12 only because `liftUpsert`'s Proxy scope answers `sharesAnyUuid` with a no-op, and the case at line 174 goes red with the real function in scope. Whichever rule wins, the losing side's pins need re-cutting and `liftUpsert` should carry the real `sharesAnyUuid`; under #1860's, the `insertRegionRun` refusal and the older-wire landing still matter, since a first frame with `tailLo` null (no `prev`) leaves the page regions-less.

Follow-ups on the older-wire branch (`render.ts` 11456), each reproduced: no wait state for a navigation that arms no seek (the reply chip, the reload restore, the unplaced-window re-arm; the notch and comment tick resolve on the resident path, barring a stale paint), where the window road showed the landing notice; a navigation on that wire lost silently on a socket death, since `onWireDown` (17481) counts a live landing only by `landingNoticeSid` or a gapped window ask, neither set here (served probe: no `wsdown-lost` row, no toast; before this change the same sequence toasted); and the pointer-exact `locateDiag` row filed without `anchorT` and `kind`: `chatHead` re-arms with both null (17930), older code now carrying every regions-less navigation; the attempt row and the window road carry both.

Smaller: `ui/webview/chat-regions.ts` 57 duplicates `send-pending.ts` 234's `OVERLAY_KINDS`, already pinned by `tests/test_send_pending_overlay_kinds.py` and imported by `queued-held.ts`; a re-export keeps the node tests green; a kernel-constant mutation broke both pins, two copies for one change.

lczh:
Follow-up for this note: #1896 (the exec harness carries the real sharesAnyUuid and executes #1860's guard; the older wire's landing shows the wait, comes down on every path that ends it, survives a socket death with its word, and files the cancel row with the click's anchor, time and kind; OVERLAY_KINDS re-exported from send-pending). Main after #1860's reconcile was already green on the served lab.

PR #1880, opened 2026-09-19
A typed /clear on a Codex session starts a fresh conversation for the same session: a new thread under the same id, swapped under the turn lock, with every surface following the new file

lczh:
Tier: feature

Stacked on the Codex slash-command guard (the #1864 branch, whose commit sits directly below this one): that change put the route in front of the setter body and left an empty handler table between the slash predicate and the refusal. This change registers the first heads in it. Filed without auto-merge: a maintainer approves it.

## The defect

On a Codex session the composer's slash list offered `/clear`, the confirm asked "Clear this conversation?", and nothing cleared. Before the guard, the word reached the model as a prompt on the same thread: idle it was queued and started a turn (the model answered "Cleared."), mid-turn it was parked and replayed alone as text. No romp surface saw anything: the Codex backend had no clearing bracket, so the chip never read clearing and the chat showed no "Clearing conversation…" element; the thread id and the transcript file did not change, so the episode boundary saw no new root head, no cards settled, the bell did not ring, no "Conversation cleared" card appeared; and the turn counted as a real reply for the judges. The guard then made the same press a refusal, which was honest but still not a clear. The same held for `/new`, Codex's own word for the operation. Nothing in the codebase handed an existing session a fresh thread: the backend's resume re-resumed the same thread id, and only the pending-placeholder branch of thread preparation ever minted one.

## The fix

A `clear` verb on the session contract, a concrete default in the same idiom as `move`: an empty string on success, `busy` when a turn is in flight (the kernel parks and retries at the turn's end), any other string the reason, shown verbatim. The default is a refusal worded for no backend in particular, and without the word "backend", which reaches a toast. The Claude Code backend does not implement it: a typed `/clear` there still goes to the CLI as text, which executes it.

The Codex backend implements it as a new app-server thread under the same session: `thread/start` tagged with Codex's own `sessionStartSource: "clear"` (the runtime validates the enum; live probe) and the row's working directory, mode and picked model; the row's thread id swaps to it under the worker's turn lock, which the worker holds across thread preparation and the whole turn, so the swap can never race a worker between preparing a thread and starting a turn. Behind that lock sits a belt that follows the backend's own busy rule: a send that is queued but whose worker has not yet taken the lock answers `busy` too, because a clear that took the lock first would land a message typed before it on the fresh conversation; a queue parked on a permanent rejection of the old thread rides into the fresh one instead, and the swap is the explicit change that re-arms it. The bracket the chip and the chat read is latched before `thread/start` and dropped when the new thread id is durable, or when the attempt raises. The fresh transcript file is touched before the registry names it (discovery signs the registry's modification time and skips a row whose file does not exist), and a raising registry write publishes nothing: the in-memory swap rolls back and the touched file leaves with it. Then, still under the turn lock, the normalizer is reset on the empty file (its own non-reentrant lock is released before the rebuild, which takes it): a prompt typed while `thread/start` is in flight is handed over (nothing reads busy through the bracket) and queues a worker on that same turn lock, and the worker holds whatever normalizer it reads for its whole turn, so a reset after the release let a scheduler switch hand it the old normalizer with the new thread id and chain the fresh file's first record to the old conversation, a file that would never earn a boundary. After the release, an acknowledging chip carrying the command as typed (`/clear`, `/new`, `/clear now`: the composer retires its optimistic bubble only by that exact text) is left in the live tail with its durable twin on disk (the same record the Claude backend writes for its setter picks, so the chat's existing gesture rendering and dedup apply), the Codex-side thread name follows, and a parked queue is kicked; everything after the swap is bookkeeping on a clear that already happened, so a raise there (the twin's write on a full disk) is logged and the verb still answers success, with the kick and the pushes running regardless. Name, mail, tags, color, note, mode, model and effort survive because they live on the row or in stores keyed by the session id, never by the thread id. The old thread is left in place: romp reads only its own materialized file, which stays where the "Conversation cleared" card's lazy body looks for it.

The read side needs nothing new. The new file's first record is a root, so the episode boundary tick records the boundary beside the seed row the old file earned at its own first record, settles the open cards, rings the bell once, and the chat inserts the head card, whose body resolves the previous file as a sibling in the same projects directory. That lands on the first prompt into the fresh conversation, not at the clear itself (an empty file has no head); the doc says so.

In the kernel, the two heads register in the guard's table and take the drive-op FIFO exactly as the effort and fast arms do, when the command is the whole message (the slash predicate's trailing whitespace admits a newline, so a pasted `/clear` with a second line under it is slash-shaped, would have cleared with the rest of the text dropped, and had passed the composer's confirm, whose gate wants a space after the head; it is refused instead, in words that say the command must be the whole message, on the typed route and for a command op parked before this change alike): one gate evaluation; mid-turn, behind a queue, compacting or under an account hold the press parks as a visible chip showing the words typed, in press order; `busy` from the backend parks the same way and is never shown; any other answer is said on the delivering socket with the session and the press named, so the chat retires the optimistic bubble it drew and puts the words back into an empty composer, filed under its own key so POST /send and `romp send` answer with the backend's words, kept on the bell, logged once. The drain continues past a clear (it opens no turn, so a message typed after the `/clear` lands on the fresh conversation in press order); on `busy` it leaves the head in place with nothing recorded as in flight (the chip's cancel still works) and with no clock hold; and it routes a command op a Codex session parked before this change to the same verb rather than the guard's refusal. The Codex backend's turn-end poke now fires after the turn lock is released rather than inside it: the kernel's drain runs on that poke, and a clear it fires takes that lock, so poked from inside every clear parked mid-turn met `busy` on the very poke that announced the turn's end and waited for the pusher's backstop instead. Released first, the poke is the retry event.

The live probe showed one more thing the design had to absorb: a thread with no turn has no rollout, so a fresh app-server refuses to resume it with `no rollout found`, an invalid-request error the permanent-rejection rule would have parked the queue on for good. A `/clear` mints exactly such a thread and saves its id (a spawned thread is the same until its first turn). Thread preparation now handles that refusal, and only that refusal, for a thread romp's own record says never ran a turn (its materialized file is still empty): it re-creates the thread once with the row's parameters through the create branch's own transaction and rollback, logs it loudly, and removes the empty file of the thread that never ran. The same refusal for a thread whose file holds records is a real inconsistency and parks as before. Of the two options, this one keeps the spec's rollback discipline: the clear still saves a real thread id in one transaction and every surface follows it at once; recording a placeholder instead would have changed the transcript path twice, lost the start-source tag, and put the whole design behind the first prompt.

The composer gives a typed `/new` the same open-cards confirm as `/clear`, on Codex sessions only (a Claude session never sees a confirm for a command its CLI refuses as unknown), reading the backend from the session's status; the composer's slash list for a Codex session gains `/clear` and `/new`; the bell's explanation of the refused kind no longer says a Codex session has no `/clear`.

## Screenshots

Synthetic data throughout (the notes-api demo sessions under `/TESTDIR`, placeholder ids, invented prompts) over a hermetic lab kernel built from this branch.

A Codex session mid-turn: the /clear queued behind a message typed during the turn, in press order `chip-midturn.png`

Right after /clear on an idle Codex session: the fresh, empty conversation with the system card and the /clear gesture chip `after-clear.png`

One prompt into the fresh conversation: the "Conversation cleared" head card expanded over the old Codex conversation, the dropped card named `head-card.png`

## Design points

- The verb is a concrete default on the contract, not abstract: the unowned route and the test fakes need no change, and a backend without one refuses by omission in words.
- Idle-only through the backend's turn lock, taken non-blocking, plus the busy-rule belt for a queued-but-untaken send; `busy` is the kernel's word to park on and is never shown.
- The bracket latches before the request and drops on the deciding event, the durable thread id, never on a timer.
- The fresh file exists before the row names it; a raising registry write publishes nothing.
- The normalizer reset rides the same turn lock as the swap, and everything after the durable swap is bookkeeping that can only log: the verb answers what happened to the thread.
- A registered head takes the message only when it IS the message: a second line under `/clear` is refused in words, never cleared with the rest dropped.
- The chip lives in the backend, the Claude backend's own precedent: the composer's optimistic bubble ends only on a landed user event with the text, and the kernel has no live-atom store of its own.
- The drain's `busy` branch keeps the head, records nothing in flight, and arms no clock; the poke moved outside the lock makes the turn's end the exact retry event.
- The restart fallback keys on two facts at once: the server's exact refusal and romp's own record of no turn.
- The old thread is left, not archived; no prior-thread list on the row; both are recorded below as the maintainer's call.

Open decisions the spec left to the maintainer, with the recommendation built here:

1. Archive or leave the old Codex thread. Built: leave. Romp never touches the app-server thread again, its effect on the rollout is unverified live, it is one more call that can raise after the swap committed, and a thread with no turn cannot be archived at all. If the Codex CLI's own list should stay tidy, a tolerant best-effort archive after the swap is a small addition.
2. A list of prior thread ids on the registry row, so the pre-clear stretch stays on the timeline, the feed, the judges and the spend readers. Built: not in this change. Parity with Claude today is the current file only on those surfaces; giving Codex more would re-judge dead conversations as lanes. A feature for both backends if wanted.
3. Where the acknowledging chip lives. Built: in the Codex backend, with a small twin of the Claude backend's durable-gesture writer (the Codex module cannot import the SDK-gated one). The alternative, moving the writer into the contract module and importing it from both, is cleaner but touches the SDK backend, which two webview tests pin.

Known residuals, none widened here: a refused clear that was parked (a `("clear", text)` op carries no copy id) leaves the composer's optimistic bubble until its cancel, the same gap every refused parked setting has today; a `/clear` typed on the first press of a session whose registry write then fails is said, not retried; two rapid clears append two ops, and the second clears an empty thread.

## Review round

The maintainer's review of the first cut, folded in one commit on top of it, in the review's order. The composer's bubble: the kernel road was taken, closing both `/new` and `/clear now`. The command as typed, whitespace-trimmed, rides the parked op (`("clear", text)`) and the contract's verb (`clear(sid, text="/clear")`) onto the echo, its uuid and the durable twin; the parked-op renderer returns it, the drain's arm passes it (a parked `/new` no longer lands as a `/clear` chip), readers of the longer tuple are length-guarded because the parked-ops file survives a restart, the fold that hides the queued chip under the live "Clearing conversation…" element keys on the op's kind rather than the literal `/clear`, and the signature pin in the contract test moved. A `/model` pick landing inside the clear bracket: the locked swap now takes the row's current model, as thread creation does, and the pre-request model rides the start params only; the review's probe (a scripted client whose second thread start blocks, the pick made during the block) is a test. The cycle that drains a parked clear: the targeted push, which is the backend's hook and runs inside the pusher's cycle on that thread, resets the cycle's discover-row memo to a fresh dict and drops the sid's path memo at entry, before any read and whether or not a client is connected, so the hook's own frame during the verb and every build after it in the cycle resolve the fresh file; the test records what the hook resolved. The two in-code claims about the twin are reworded: it renders in the fresh conversation only until the boundary tick, then closes the cleared conversation as the last gesture row of its episode, and a test on the episode harness pins that (a twin stamped between the two files' records, the tick, the session build leading with the head card and no `/clear` chip, the episode build ending with the `/clear` row). Smaller: the shared-root scrub also removes both twins and the registry lock when set-up found none; the drain's whole-message rule for a pre-upgrade command op is pinned for `/clear` and `/new` (the verb never runs, the send behind it delivers, one warn with the press id and the whole-message wording, nothing parked); the rollback test's save patch records that the fresh file existed at the write; thread creation touches the fresh file before the save that names it, drops the trailing touch and removes the file on both rollback branches, pinned at the save itself; the drain's lock docstring names `thread_start` beside the model listing and the handshake as the waits it takes with no bound of its own; and the doc's "queued messages carry over" is now the three cases.

The maintainer's second review, folded in one more commit on top. A Codex `("command", "/new")` op can reach the drain by roads that skip the route into `_send_or_park` (a pending-ops.json written before the heads registered, a notice card's plain action, a follow-up with no item id); the drain ran it as a clear while the fold that hides the queued chip under the live "Clearing conversation…" element keyed on the op's kind `clear` or the literal `/clear`, so that op's chip drew beside the element. One predicate now, `_parked_clear_op` beside `_CODEX_CLEAR_HEADS`: true for a `("clear", …)` op of any length, or a one-line `("command", …)` op whose head is one of those heads on the Codex backend; the drain's arm reads it in place of its own copy of the rule, and `build_session` reads it where it collects the parked chips the fold removes, so the drain and the chat agree by construction. The fold's text match stays for one thing the predicate is not about: the SDK's own queued `/clear`, a text in that backend's queue under which its clearing bracket is already lit. The fold test gains the command op (red on the previous head: the build carried both the clearing and the queued events), a drain test pins the predicate's truth table and that the drain consults it for its head, and the chat-build census classifies the helper.

## Merges from main

Main's refusal rule changed under this branch: a Codex session is refused only the slash commands the kernel knows it cannot take (`_CODEX_REFUSED_HEADS`, one predicate for the route and the drain), and a slash-shaped text outside that set reaches the model as prose. The two designs hold together at this head: a registered head takes the text on the route when the command is the whole message, a parked one runs at the drain ahead of the refusal, and the head stays in the known set, which is what refuses it with more lines under it. The guard's known-heads census skips a registered head and pins that the skipped set is the handler table, its seam test registers `/compact` in place of `/clear`, the route test reads `/clearx` as prose, and the Codex doc's paragraph names the two heads as taken ahead of the refused list.

Main landed the six postal tools as Codex dynamic tools between this branch's heads. A thread is registered with them at thread/start alone (a resume cannot add a registration), and this branch has two thread/start sites main did not know about: `_create_thread`, the helper the placeholder branch and the restart fallback share, and clear()'s own start. Both carry the postal params now. Without the second, a Codex session would have lost its mail at its first /clear, with nothing said. Pinned red first in the PostalTools tests: a clear must reuse the spawn's own registration beside the clear start source.

## Tests

Red first on the stock tree (the #1864 guard branch), one module at a time.

`tests/test_codex_backend.py`, new class NativeClear (16 tests at this head; 13 in the first cut, the review round added three): of the first cut's thirteen, twelve fail on stock. Nine with the missing verb or bracket (`AttributeError: 'CodexBackend' object has no attribute 'clear'`, or `'clearing'` for the bracket test); one because the stock module has no durable-gesture writer to patch; and two by a wait that times out rather than by the raise itself: the turn-end-poke test, whose worker raises the missing verb inside the lock (the raise is in its captured stderr) so the poke's answer never comes, and the bracket-race test, whose clear thread raises the missing verb before any `thread/start`. The thirteenth (a `no rollout found` refusal for a thread that did run stays a parked rejection) pins today's behavior and is green on both. Two of the twelve, the bracket-race and the post-swap-raise tests, were red against the first cut of this change as well: with the reset after the release, a forced handoff chained the fresh file's first record to the old leaf under the old thread id; a raising gesture write escaped the verb. Green after: the clear mints a second thread with the exact create parameters (cwd, auto-mode approval policy and reviewer, the workspace permission and root, the picked model, the clear start source), the row's thread id becomes the new one while name, cwd, model, effort, mode, dead, queue, note and color are unchanged and the launch error is cleared, the fresh file exists empty before the row names it and the old file keeps its records, the Codex-side name follows, the live context resets while the model stands, the bracket reads false afterwards, the next turn runs on the new thread with no resume, and the new file's first record is a root stamped with the new thread id; the bracket is true while `thread/start` blocks and false the instant the registry holds the new id; a raising registry write returns the reason, drops the bracket, restores the thread id and model, removes the touched file, leaves the registry unchanged and the next turn on the old thread; an open turn answers `busy` with no thread minted; a queued send whose worker has not taken the lock answers `busy` and then runs on the thread it was typed into; the chip carries the command flag under a `cmd:` id in the fresh thread with its durable twin on disk, survives a human record in its own second, retires on the next one, and a plain echo under the same floor stays; a queue parked on a permanent rejection rides into the fresh thread and the launch error clears; a clear pressed on the turn-end poke runs rather than answering `busy`; a prompt typed during the bracket runs on the fresh thread and its first record is a root stamped with the new thread id even when the worker takes the lock the instant the clear releases it; a raise after the swap (the twin's write on a full disk) is logged, the verb still answers success, the live chip stands, and a queue parked on the old thread's rejection still runs on the fresh one; an ended or unknown session refuses in words; and, across a simulated kernel restart with a fresh scripted client that refuses the resume with `no rollout found`, one re-create with the row's cwd, mode and model, the row naming the re-created thread, the prompt running on it, the empty file of the never-run thread gone, the new file's first record still a root, and the log line naming the thread.

`tests/test_codex_clear_route.py`, new module (19 tests at this head; 12 in the first cut, the review round added six and the second review one): of the first cut's twelve, eleven fail on stock (the route answers the guard's refusal and never calls the verb, so `refused` stands where `queued` was expected and nothing parks; the palette lists only the setter heads; the drain's clear op takes no arm and the message behind it is delivered alone; over the real backend the typed `/clear` is refused and the parked one never fires), the episode-chain test is green on both as the spec predicted. Green after: `/clear`, `/new` and `/clear now` take the route to the verb with one gate read, an empty socket and `queued: false`, while `/clearx` (a head outside the known set: since main's change of 2026-09-19 the guard refuses only the commands it knows) and prose both fall through to the model, a `/clear` or `/new` with a second line under it is refused in words that name the command and what the message must be (one warn on the delivering socket with the press named, nothing parked, the verb never called), and Claude's `/clear` keeps its path; a gated press parks a `("clear",)` op and a `busy` answer parks after the call, never shown; a refusal is one warn frame naming the session and the press, filed under its own key, on the bell, answered verbatim by POST /send, and broadcast to the chat panes when no socket carried the op; the palette lists the two heads and the table maps both to one handler; the drain fires a parked clear then delivers the message behind it with nothing held or recorded, leaves a `busy` head with no clock and no in-flight record so its cancel still takes it, warns and pops a refused clear while the setting behind it delivers, runs a `("command", "/new")` op parked before the upgrade as a clear and carries the parked copy's id on a refusal, and hands a backend without the verb the contract's refusal; the real episode tick over two synthetic Codex files under one projects directory seeds on the old file's head, records the boundary on the new file's root head, settles the open top into the clears log and the goal store, leads the chat with the head card naming it, renders the old file as the cleared episode with no error, and rings the bell once; and, over the real Codex backend with the codex-backend module's own scripted client installed as the kernel's singleton, a typed `/clear` swaps the thread with the start source, drops the bracket, leaves the chip and nothing else in the fresh conversation, and the chat merge's own verdict (the chip formula over the live-merged turns, not the backend's row) reads that conversation's turn as ended and the chip as ready, never working, with the backend's row agreeing; and a `/clear` parked mid-turn fires on the turn-end poke itself with one thread minted and no clock armed.

`tests/test_session_api.py` gains the contract test in `move`'s style: `clear` is not abstract, the default is a reason, not empty, not `busy`, without the word "backend", the Codex backend defines it and the SDK backend does not (red on stock: the ABC has no such attribute).

`tests/test_codex_slash_guard.py` (the #1864 module) re-pointed from `/clear` to `/compact` as its refused example wherever a head now taken was used, kept a two-line `/clear` in its refused set, gained `clear` on its fake, and its palette pins carry the two new rows; `ui/webview/clear-confirm.test.ts` gains the `/new` truth table and a source pin on the Codex-only gate; `ui/webview/codex-slash-guard.test.ts` follows the bell explanation's new wording.

Review round, red first on the first cut: the chip test raises (the verb took no text), the route tests see `/clear` where `/new` was typed and a `("clear",)` op where `("clear", "/new")` was expected, the model-pick test reads the pre-clear model back from memory, the cycle test's hook resolves the old file twice, and the create-branch test records the file missing at the save. Green after, one module at a time on the review-round commit: test_codex_backend 135 (+5 subtests); test_codex_clear_route 18; test_session_api 13; test_codex_slash_guard 15; test_codex_events_golden 29; test_kernel_send_park 51; test_kernel_parked_ops_lock 19; test_kernel_pusher_snapshot 31; test_kernel_headless_ops 20; test_kernel 460 (+12); test_chat_pages 34 (+27); test_injected_voice 10 (+222); test_chat_build_sig_inputs 65 (+10); test_feed_memo_inputs 31; test_perf_bench 49; and the route module with the pusher-snapshot module and the discover-walk class in one run, 52, for the shared-root leak. Static pin scan on the final commit: 0 lost. The extension suite was not rerun this round: no ui file changed, and the kernel lines the webview tests pin (the targeted push's definition, its wiring and callers, the clearing fold's `if`) are untouched by it. The first cut's runs follow.

Second review, red first on the review-round commit: the fold test's new stretch fails with the queued event present beside the clearing one (`'queued' unexpectedly found in ['system', 'user', 'clearing', 'queued']`), and the predicate test fails on the missing helper. Green after, one module at a time on the final commit: test_codex_clear_route 19; test_codex_backend 135 (+5 subtests); test_kernel_send_park 51; test_kernel_parked_ops_lock 19; test_kernel_headless_ops 20; test_kernel 460 (+12); test_chat_pages 34 (+27); test_chat_build_sig_inputs 65 (+10); test_feed_memo_inputs 31; test_codex_slash_guard 15. Static pin scan on the final commit: 0 lost. The extension suite was not rerun: no ui file changed, and none of the kernel lines the webview tests pin (the clearing fold's `if`, the chip formula, the targeted push) is among the lines this commit touches, checked by matching every pinned regex and string literal in the 161 webview and extension test files that read kernel source against the removed lines.

First cut, one module at a time: test_codex_backend 132 (+5 subtests); test_codex_clear_route 12; test_codex_slash_guard 15; test_codex_events_golden 29; test_codex_echo_merge 4; test_kernel 460 (+12); test_chat_pages 34 (+27); test_injected_voice 10 (+222); test_chat_build_sig_inputs 65 (+10); test_feed_memo_inputs 31: all green. On the cut before the review's folds (the same change less the normalizer-reset order, the guarded tail, the whole-message rule and the two added assertions), the wider set: test_backend_call_parity 4; test_session_api 13; test_clear_chat_view 14; test_episode_boundary 17; test_kernel_send_park 51; test_kernel_meta_command_gate 7; test_kernel_parked_ops_lock 19; test_drain_hoists 6; test_kernel_parked_ops_liveness 9; test_session_move_kernel 19; test_codex_discovery 6; test_send_pending_overlay_kinds 1; test_kernel_cmd_gesture 5; test_kernel_model_queue 9; test_scanner_folds 12; test_session_env 43; test_stage0_log_readers 36; test_codex_launch_error_card 4; test_kernel_compact_route 14; test_kernel_headless_ops 20; test_sdk_registry_blind 31 (+12); test_model_live_midturn 17; test_kernel_cmd_cache 6; test_perf_stats 62; test_kernel_push_responsiveness 15; test_perf_bench 49: all green. Static pin scan over every assertIn literal, on the final commit: 0 lost. Extension suite (vscode-extension npm test, which pins kernel and render source lines), on the cut before the folds: 5062 tests, 5055 pass, 1 fail, the pre-existing KaTeX pixel probe that fails on main in this environment; the folds change no kernel line a webview test pins (grepped for every literal touched).


ksarma:
The evidence was executed against d658ea87b and re-checked against the current head e514a39cd, which amends the same commit with a test-only change (the new test module builds its backend over the shared state root and scrubs after itself); every item below was re-read on e514a39cd and one item is about that amendment.

Two items are worth doing before this merges.

### At this head

**A typed /new, or a /clear with words after it, never retires the composer's bubble.** The route at kernel/kernel.py:35910 parks the text-less `("clear",)` and calls the verb without the typed text or the press id; the chip at kernel/codex_backend.py:1416 is the literal /clear under a cmd uuid with no copy id, the twin at 1417 the same, and the parked op renders as /clear (kernel.py:34979). The composer retires a pending entry only by exact text or copy id (send-pending.ts:179-191), so a typed /new or "/clear now" leaves the sending bubble standing (landed 0, keep 1) and a /new parked mid-turn draws the queued /clear chip beside it. Kernel side, closing both: carry `text.strip()` on the parked op and through the contract's clear (default /clear) onto the echo, the twin and the uuid, return it from the parked-op renderer, pass it from the drain's arm at 36172 (else a parked /new lands as a /clear chip), length-guard readers of the longer tuple (the parked-ops file survives a restart), key the clearing indicator on op kind, and move the signature pin at tests/test_session_api.py:83 (an inner double space in the typed text still misses, the same residual the Claude route has). A press id as the echo's uuid does not retire the bubble; an id would have to travel in the event's qid field, which `_op_qid` (kernel.py:35427) does not read for a clear. Webview side, closing /new only: rewrite a Codex /new to /clear inside `deliver()` (ui/webview/render.ts:19245, where the text bound at 19213 feeds the optimistic register, the last-sent record and the staged flush), after the confirm at 19336, since a rewrite before it makes the /clear detail swallow the text and leaves the /new gate dead, and not only in the message router, which is too late for the bubble and the last-sent record. Either road alone closes it for /new; saying which one is taken lets the tests pin one shape, and both a route test typing /new and asserting the chip's text (the kernel road) and a send-pending test that a /new press lands are red here.

**A /model pick landing inside the clear bracket is accepted, saved, then reverted.** kernel/codex_backend.py:1354 carries the model read before the thread start, and the swap writes it onto the row at 1365 (`s.model = new_model`); nothing reads busy through the bracket, and the model setter never takes the mode lock the bracket holds. A scripted client whose second thread start blocked showed the pick saved during the bracket, then memory, registry, the live listing and the next turn's params all back on the pre-clear model. In the locked swap take the row's current value as `_create_thread` does, keep the pre-request model for the start params only, and drop the dead local; the rollback tuple already covers a pick landed in the bracket. That one line fixed the probe on a scratch copy with the backend class still green; the probe is worth landing as a test.

### Also at this head

**The cycle that drains a parked clear pushes the old transcript with the chip appended.** The arm at kernel/kernel.py:36172 runs the verb inside an open pusher scope whose builds share the memos `_live_scope.paths` and sessions filled before the swap, and the verb's own push tail runs there before the arm regains control. Both that tail push and the post-drain build resolved the old file with the old prompts plus the chip (0 builds on the new path); the fresh conversation appeared a cycle later. Inside `_push_session_now` (53396), under an open scope, reset the sessions memo to a fresh dict and drop the sid from the paths memo before the build; the sessions reset carries the weight, and a reset in the drain's arm alone is too late because the hook's frame during the verb is already stale. The discover cache needs no reset (its fingerprint follows the registry's mtime, which the swap rewrote); emulated at hook entry, the reset turned both frames to the new path, and the pin should assert what the hook resolved during the verb itself.

**The twin interleaves only until the boundary tick, then closes the cleared conversation.** tests/test_codex_backend.py:3151 and the `_append_cmd_gesture` docstring at kernel/codex_backend.py:252-256 say the kernel interleaves the durable chip once the live one retires. With the chip and twin stamped between two turns' record times (a real session's order; the module's scripted clock sits under the real one, so its tests never meet the floor), the builds before the tick carry the row, the post-tick session build drops it at the floor, and the twin renders only as the last gesture row of the old file's episode. The body's own sentence on rendering is accurate, so only the two in-code claims need rewording, to say the twin renders in the fresh conversation only until the boundary tick and then closes the cleared conversation; the states write stays, and the floor rule is not this change's to bend. Add a test on the `CodexClearEpisode` harness (its own root, so the shared-root scrub does not widen) that prompts after the clear, ticks the boundary, and asserts the session build leads with the head card and no /clear chip while the episode build ends with the /clear row.

### Smaller points

- tests/test_codex_clear_route.py:455, the amendment: `_scrub` removes the new project files, the names entries and the registry, but the real clear also writes states/<sid>.jsonl (kernel/codex_backend.py:1417) and the registry save creates codex/registry.lock (399), so both twins and the lock stay in the shared root. Unlink the two states files with the same guard, unlink the lock when set-up found none, and name the twin in the docstring.
- tests/test_codex_clear_route.py:272: the drain's whole-message rule for a pre-upgrade command op (`_slash_alone` at kernel/kernel.py:36147) is unpinned; with the guard removed the suite stayed green and a parked two-line /clear ran with its second line dropped. Park a two-line command op with a send behind it and assert the verb never ran, the send delivered, one warn with the press id and the whole-message wording, and nothing parked; cover /new too.
- tests/test_codex_backend.py:2917: the file-before-registry order is asserted only as a post-state; the touch moved after the save left five modules green. Have the rollback test's save patch append `new_file.exists()` to a list before raising (the verb catches Exception, so an assert inside the patch would surface only indirectly), and assert [True] after the clear returns.
- kernel/codex_backend.py:1707: `_create_thread` saves the tid before the file exists (the touch follows at 1719), the order the clear avoids; discovery skips a row whose file does not stat, so a discovery in the gap (the restart fallback included) drops the session until the next save. Touch before the save and drop the trailing touch, unlink on both rollback branches (the raise and the dead return at 1696), and pin that the file exists whenever tid is among the saved fields.
- kernel/kernel.py:36025: the clear arm runs the thread start on the pusher thread with no timeout, but that predates the change: at the base a parked effort pick blocks in the model listing and a parked send in the handshake (that one bounded at 30s), as the ONE LOCK docstring says. One sentence there naming `thread_start` beside them fits; a watchdog is a separate change.
- docs/codex.md:162 says queued messages carry over; by execution a message typed after the /clear lands on the fresh conversation, a message queued before it runs first on the conversation it was typed into, and only a queue stuck behind a failed start of the old conversation rides into the fresh one. Reword to those three cases.

One claimed gap was checked and found already pinned, so it is not listed: widening the "only that refusal" check at kernel/codex_backend.py:1782 fails two existing lifecycle tests.

The modules run, their counts, the mutations and the probes are in a second comment below.

ksarma:
Test runs behind the comment above, on e514a39cd and on d658ea87b (every kernel, webview and docs file cited is byte-identical between the two; they differ only in the new test module's set-up, the subject of the amendment item).

### Tests

On e514a39cd, each in a capped scope with the session environment unset: tests/test_codex_clear_route.py alone, 12 passed; the same with tests/test_codex_backend.py, tests/test_codex_slash_guard.py and tests/test_session_api.py, 172 passed and 5 subtests, run twice with identical results. The new module's own pins were read as evidence for the first item: a typed /new calls the verb with no text, parks `("clear",)`, and renders it as /clear (lines 152-156, 195, 205, 208).

On d658ea87b: tests/test_codex_backend.py 132 passed and 5 subtests; tests/test_kernel.py 460 passed and 12 subtests; tests/test_codex_clear_route.py 12, tests/test_codex_slash_guard.py 15, tests/test_session_api.py 13, tests/test_backend_call_parity.py 4 and tests/test_state_isolation_order.py 8, in one run of 52; 24 further modules from the body's lists, 428 passed and 271 subtests; 35 neighbouring modules in one isolated run, 757 passed and 276 subtests; tests/test_kernel.py with tests/test_perf_stats.py and tests/test_kernel_push_responsiveness.py, 537 passed and 12 subtests in an isolated copy. On the base 594b7bc43 with this change's four test files copied in: 24 failed, 14 passed, matching the body's red-first counts (11 of 12 in the route module, 12 of 13 in the backend class, and the contract test).

Mutations, each on a scratch copy of the head: the whole-message guard removed at kernel/kernel.py:36147 left the suite green, and a parked two-line command op then ran as a clear with its second line dropped, where the head warns and keeps the queued send; the fresh file's touch moved after the registry save in the clear gave 178 passed and 5 subtests across the five codex modules, so the order is unpinned; the refusal check at kernel/codex_backend.py:1782 widened gave 2 failed and 142 passed across tests/test_codex_backend.py and tests/test_codex_clear_route.py, the two failures being the existing lifecycle retry tests, which is why that gap is not listed.

Probes, each on a scratch copy in a capped scope: the composer's decisions came from an esbuild bundle of ui/webview/send-pending.ts driven under node, reconciling a pending entry against the backend's chip (typed /clear landed 1 and keep 0; typed /new and "/clear now" landed 0 and keep 1; parked mid-turn the queued /clear chip is drawn beside the /new bubble); the real Codex backend with the module's scripted client was driven through the kernel's route, and through a real pusher cycle for the same-cycle build; a subclass of the scripted client whose second thread start blocked on an event carried the model pick during the bracket, on the head and on a copy with the one-line change; a subclass of the module's real-backend class stamped the chip and twin between two turns' record times and ran the boundary tick for the twin's rendering; and a wrapper around the registry save recorded whether the transcript existed at every save that named a tid, calling the real discovery in the gap.

lczh:
Review round folded as one commit on top (8052c459), kernel road for the composer bubble: the typed text rides the parked op, the echo, its uuid and the twin, so a typed /new or a /clear with words retires its bubble; the locked swap keeps a /model pick that lands inside the bracket; the targeted push resets the scope's session and path memos so the drain's cycle builds the fresh file; the twin's rendering claim reworded and pinned on the episode harness; the scrub covers the states twins and the registry lock; the drain's whole-message rule, the file-before-registry order (clear and _create_thread), the ONE LOCK docstring and the docs' three queue cases. Red evidence and counts are in the body's Review round section.

ksarma:
Evidence executed against 8052c459d. Every item from the earlier comments landed at this head, and execution shows that the typed words reach the chip, the drain's whole-message rule is pinned, and a model pick inside the bracket survives the swap.

### At this head

A Codex ("command", "/new") op's queued chip draws beside the live clearing element: the fold keys on op kind `clear` or the literal /clear (kernel/kernel.py:38839, 38876), while the drain (36157-36159) runs it as a clear. With that op parked after a turn and the bracket patched true, the build carried both clearing and queued events; the clear runs and pops the op (display only). The op comes from a pending-ops.json written before the heads were registered, or here from a notice card's plain action (24739) and a follow-up with no item id (19432), both of which skip the route into `_send_or_park`, which parked it. One predicate closes it: a helper beside `_CODEX_CLEAR_HEADS`, true for a clear op or a one-line Codex command op with one of those heads, read at both sites; the fold test gains that op.

### Smaller points

The body's Tests section counts `NativeClear` at 13 and the route module at 12, against 16 and 18 here; its final run line says 135 and 18.

lczh:
Second round folded as one commit on top (e95bbbbc): one predicate beside the clear heads, read by the drain and by both chip-fold sites, so a parked one-line Codex /new command op is a clear everywhere; the fold test gains that op (red on the previous head with both the clearing and the queued events); the body's counts corrected (NativeClear 16, route module 19). The previous head's one red check was the held-mail served lab (the deny-note buttons not latched), which this branch does not touch and which also went red once on main locally.

ksarma:
Executed against e95bbbbc6. Both items from the earlier comment landed at this head: `_parked_clear_op` sits beside the clear heads and is read by both the drain's clear test (kernel/kernel.py:36181) and the chat's chip collection (38861), so a Codex ("command", "/new") op parked from a notice card's plain action or a bare follow-up folds under the live clearing element with no queued chip beside it, and the fold test carries the pre-upgrade op (red on 8052c459d at the new stretch, green here). The body's Tests counts now read the head's (NativeClear 16, the route module 19, both counted). Also executed: the predicate's edges (a two-line command op is not folded; a second clear-shaped op parked behind the first is not; a bare or non-string text is not), the effort op behind a folded clear still delivers, and no source pin in the webview or Python tests matches the old text and not the new. Nothing further at this head.

ksarma:
Executed against 8d759c260, the merge of main into the branch. The merge resolves one file, kernel/codex_backend.py, and equals the automatic merge everywhere else: the branch's hunks in the six other files main also changed are byte-identical against the old base and the new one. The resolution (lines 1902 to 1935) keeps both arms verbatim as an if/else on the exception: a reply the pinned SDK cannot read logs once and falls through to the loaded-generation bookkeeping; anything else takes the branch's arm, which re-raises unless the server refused a thread with no rollout that never turned, and then re-creates once. The order is the right one: a probe drove a pydantic-shaped validation error whose text also carries "no rollout found" against a never-turned thread and a turned one, and both took the first arm (no thread_start, turn_start on the same thread, one log line), which is correct because the server's own refusal arrives as an InvalidRequestError, never as a validation error; the same refusal on the empty file re-creates once, on a thread with records it raises, and a RuntimeError on the empty file raises. Both sides' modules are green at this head (the backend module 140 passed, holding the five resume-tolerance tests and the sixteen native-clear tests; the clear route 19; the slash guard 15; the session API 13; the build-signature inputs 65) and the webview set is 5144 of 5144; the earlier comments' items stay landed. Nothing further at this head.

lczh:
Merged main at 7596bf7d. The branch's own diff is unchanged except for one integration with #1867, which landed between the two heads.

#1867 registers the six postal tools on thread/start (and again on thread/resume, where the registration is accepted but cannot be added: a thread started without the tools stays without them for its life). This branch had extracted `_prepare_thread`'s create body into `_create_thread`, and clear() starts a thread of its own, so both were thread/start sites main did not know about. The merge carries `**self._postal_params()` into `_create_thread` and into clear()'s params. Without the second, a Codex session would have lost its mail at its first /clear, with nothing said: the fresh thread would answer every send_message with an unknown tool, and no resume could repair it.

Pinned red first in the PostalTools tests: `test_a_clear_mints_the_new_thread_with_the_tools_too` spawns with a postal callable, runs one turn, clears, and asserts the second thread/start carries the spawn's own `dynamicTools` list and the developer instructions beside `sessionStartSource: "clear"`. It fails on the merge without the clear() change (`None != [...]`) and passes with it.

Runs at 7596bf7d, one module at a time: test_codex_backend 148 passed, test_codex_postal_tools 26, test_codex_clear_route 19, test_codex_slash_guard 15, test_session_api 13, test_chat_build_sig_inputs 65, test_kernel 460; the extension suite 5156 passed, 1 failed (the KaTeX layout probe that fails on main in this environment), 6 skipped; the assertIn pin scan against the merge base finds nothing lost.

henrypinkard (approved):
Approved on the maintainer's decision of 2026-09-20 (the four Codex and feeder features merge), on the head rebased by the review session onto main after 1867, 1891 and 1908: the handler owns a whole-message /clear or /new, the drain still refuses a multi-line one, and the postal tool registration rides the clear road.

ksarma:
Executed against fbe3d8c34, the branch rebased onto main after PRs 1867, 1891 and 1908. Against an automatic replay of the three commits onto that main the resolution changes six files, and every line it replaces is replaced by a superset: the postal parameters spread into the clear's thread/start (kernel/codex_backend.py:1632) and into the re-create (1980), the route consulting the handler table first for a whole-message head and then the setter arms before the refusal (kernel/kernel.py:36831 to 36841), the docstrings and the comment over the known set reconciled, the slash paragraph of docs/codex.md rewritten with the native clear ahead of the refused list, and the tests moved with them (the known-heads census skips the registered heads and pins the skipped set equal to the handler table; a typed /clearx reaches the model as prose). The new pin for the integration is red without it: with the spread removed from the clear's parameters, test_a_clear_mints_the_new_thread_with_the_tools_too (tests/test_codex_backend.py:804) fails at the dynamicTools assertion, and a probe of the restart fallback shows the one re-create carrying the spawn's own tool list and instructions with no session-start tag. The earlier resolution in _prepare_thread keeps its truth table at 2082 to 2115 (an unreadable reply logs once and falls through; a no-rollout refusal on a never-turned thread re-creates once; anything else raises). At the drain the clear arm keyed on _parked_clear_op (37184) precedes the command arm, so a parked one-line /clear or /new runs the clear while a two-line copy is refused once with the whole-message wording, and the chat collects the parked command chips by the same predicate (39889); every earlier item stands unchanged under the rebase. Modules at this head: the backend 148, the clear route 19, the slash guard 21, the postal tools 26, the session API 13, the build-signature inputs 65; the webview set 5163 of 5163 with tsc clean; each intermediate commit of the rebased series is green in the same four modules, so a merge leaves no red commit behind. Nothing further at this head.

ksarma:
Merged 2026-09-20T23:25:08Z as 4c528d0f3, its second parent fbe3d8c34, the head the earlier rounds last read. The merge is the automatic one: its tree equals git merge-tree of its two parents, and the PR's hunks against the first parent 55d8e8f0c are the same fourteen files with the same line counts as against the merge base 8dbd3908a. Eleven of fourteen files are byte-identical to fbe3d8c34; kernel/kernel.py, kernel/session_backend.py and ui/webview/render.ts differ only by main's hunks from PRs 1925 and 1928.

Executed at 4c528d0f3, one pytest run over the PR's five test modules: 266 passed of 266 (the backend 148, the clear route 19, the slash guard 21, the session API 13, the build-signature inputs 65), 15 subtests passed of 15. The webview set was not run.

At this head the route consults the handler table first (kernel/kernel.py:36835), the whole-message refusal keeps its wording (36975), the drain and the chat's chip collection call the same _parked_clear_op (36945; 37191 and 39896), a refused clear reaches the delivering socket, the bell and POST /send (36905 to 36916, 37278 to 37286, 25477), and clear() spreads the postal parameters into thread/start (kernel/codex_backend.py:1632). Nothing from the earlier rounds stands at this head.

PR #1881, opened 2026-09-19
The in-place converge rebuild builds the production profile unless ROMP_EXT_DEV_BUILD is set, the same profile and knob as the boot build and the installer

ksarma:
Stacked on #1872 (POST /new echoes a refused effort level as refused, never as applied, and romp new says so with the kernel's reason) and lands after it: this branch's earlier commits are that PR's chain, which stands on #1865 and #1863, and this PR is the one commit on top. It carries a one-line comment correction in the same kernel file as a rider, described at the end of Fix.

## Symptom

After a fast-forward that converged in place (a UI-only change, no kernel-code change), the dashboard served unminified bundles with sourcemaps, and kept serving them across kernel restarts until an unrelated UI source change happened to coincide with a boot. An update that restarted the kernel had the same result: its pre-restart rebuild wrote the same bundles, and the fresh kernel's boot build judged them current. Measured on this tree, `node esbuild.js` against `node esbuild.js --production`: render.js, feed.js and files.js are about a fifth larger gzipped in the dev profile, a quarter for files.js (20, 20 and 26 percent with gzip at its default level; 20, 19 and 25 percent at level 9), and 46, 48 and 61 percent larger uncompressed, each with a sourcemap beside it.

## Cause

`_rebuild_dist` in `kernel/kernel.py` is the converge's builder: `_in_place_converge`, `_dist_converge_check` from the update-check loop, and the build before a restart in `_run_main_update` all call it. It ran a bare `node esbuild.js`. The other two builders take the production profile with one opt-out: `_ensure_bundles` builds `["node", "esbuild.js", "--production"]` unless `ROMP_EXT_DEV_BUILD` is set, and vscode-extension/install.sh sets `build_flags="--production"` and clears it on the same variable. `_ensure_bundles` never repaired the profile because its staleness test is mtime against dist/render.js, not profile: the bundles a converge or a pre-restart rebuild had just written were current to the next boot whatever their profile.

## Fix

`kernel/kernel.py`, `_rebuild_dist`: the argv is the one `_ensure_bundles` builds, `["node", "esbuild.js"] + ([] if os.environ.get("ROMP_EXT_DEV_BUILD") else ["--production"])`, read the same way (the empty string counts as unset, as install.sh's `-n` test reads it), with a comment inside the function naming the profile, the knob and why the boot build never repaired it. No new knob, no new argument; the return shape, the 300-character stderr tail the failure notice shows (pinned by FailureLineTail against esbuild.js's fit and the node tests' `KERNEL_TAIL`) and the 180 s timeout are unchanged. At the next in-place converge the rebuilt bundles are minified and the notice asks for a reload, the converge's ordinary path. A UI dev loop that wants readable bundles from the converge sets `ROMP_EXT_DEV_BUILD`, as it already must for the boot build and the installer.

`vscode-extension/src/esbuild-build.test.ts`: the comment that described the in-place rebuild as the run without `--production` now names the two profiles by what selects them: the dev-knob build (`ROMP_EXT_DEV_BUILD` set, or a hand-run `node esbuild.js`) and the production build (every kernel rebuild and install.sh with the knob unset, and the release build). Comment only; the test and its `KERNEL_TAIL` constant are unchanged.

One-line rider in `kernel/kernel.py`, `_lane_segments`: the wire-bar comment said `_BAR_WIRE below`; `_BAR_WIRE` is defined once, about 600 lines above the comment, so the word now reads `above`. The word was wrong from the comment's first commit (T278c, #1154), when the block sat in `build_timeline` with the table already above it; the block's move into `_lane_segments` (#1241) carried it unchanged. Comment only, no behaviour; the one test reading the region (tests/test_timeline_bars_wire.py, BarLiteral) anchors on the bar literal after the comment and is unaffected.

## Tests

`tests/test_bundle_build_mode.py` gains an executed class, `TheBuildersRunTheProductionProfile`. It loads bin/romp-kernel under a private module name, swaps the loaded module's `subprocess` attribute for a recorder whose every `run` returns code 0 (the kernel module's attribute, never `subprocess.run` itself, which every module in the process shares), puts `ROMP_EXT_DEV_BUILD` in one of three states inside `mock.patch.dict(os.environ)` (unset, the empty string, "1"), and asserts the recorded argv whole.

- `test_the_converge_rebuild_passes_production_unless_the_dev_knob_is_set` calls `_rebuild_dist()`, asserts it reports ok, and asserts the single recorded argv is `["node", "esbuild.js", "--production"]` for unset and for the empty string and `["node", "esbuild.js"]` for "1". With the argv line as it was before this change (the bare run, the explanatory comment kept) the module alone reads 2 failed, 10 passed, 4 subtests passed: the unset and empty-string subtests fail with `[['node', 'esbuild.js']] != [['node', 'esbuild.js', '--production']]`. Two other single-line variants also go red: reading the empty string as set fails the empty-string subtest, and ignoring the knob fails the "1" subtest.
- `test_the_boot_build_passes_production_unless_the_dev_knob_is_set` runs `_ensure_bundles` with `ROOT` and `DIST` repointed at a synthetic checkout (a `vscode-extension/node_modules` directory, no dist, no sources: stale, one build; the recorder never raises, so the npm-install retry is never entered) and asserts the same three argvs. Green before and after: it pins the agreement the fix restores and executes what the module's existing source-read pin on `_ensure_bundles` only greps.

An executed test rather than a text pin, because the explanatory comment now inside `_rebuild_dist` names both `--production` and the knob and would satisfy a grep of the body with the argv line reverted. The module gets the state preamble a kernel load needs, above the load and below the `romp_load` import, as tests/test_state_isolation_order.py holds.

Green on this head: the module (10 passed, 6 subtests passed), the state-isolation ratchet, tests/test_timeline_bars_wire.py and the four modules that stub `_rebuild_dist` (test_dist_converge, test_converge_declined, test_main_drift_notice, test_restart_classifier): 134 passed, 6 subtests passed. `npm run typecheck` clean; the esbuild-build test file alone: 10 pass, 0 fail.

Full pytest on this head (`-n 10`, short /tmp TMPDIR, no worker lost): 14313 passed, 11 skipped, 2335 subtests passed, 5 failed in 297 s. None of the five is in a file this change touches, and none calls `_rebuild_dist`:

- `tests/test_judges_process.py::ChildRoad::test_one_pass_through_the_child_feeds_perf_and_keeps_the_kernels_bookkeeping_in_order` (a cpu figure, 7.04 against 7.0, under load): passes alone; the same failure appears in 15 other full-suite runs on this machine at other heads.
- `tests/test_session_host_restart.py::ServedRestart::test_with_hosts_on_the_turn_survives_the_restart_and_with_hosts_off_it_is_cut` (timed out waiting for the turn to start): passes alone; in 22 other full-suite runs on this machine.
- `tests/test_sdk_backend.py::OptionsAssembly::test_ultracode_is_a_choice_everywhere_but_never_the_seeded_default` and `::ReconnectReconcilesInflight::test_reconnect_settles_a_turn_stranded_by_the_teardown`: alone they skip (`claude_agent_sdk not installed`, the module's guard); under xdist they ran, which that import-time guard permits only when a sibling module's fake `claude_agent_sdk` was in `sys.modules` on that worker. Each appears in 28 other full-suite runs on this machine.
- `tests/test_kernel_awaiting_stamp.py::AwaitingWake::test_dormant_session_is_not_woken_but_converts_to_the_dead_wait_block`: fails alone on this head, and fails the same way at this branch's base commit without this change (the dead-wait road stands down with `the SDK registry directory cannot be read`, its `_sdk_records_blind` check reading the test's fresh state root, which has no sdk/ directory, as blind). It appears in 15 other full-suite runs on this machine and passes in others at nearby heads, so it passes only when an earlier test in the same process happens to create that directory, and it predates this change; nothing in it reaches `_rebuild_dist`.

Full `npm test` on this head, two legs in one 16 GB unit: 5037 pass, 0 fail, and the timeline-tags-scale file alone 22 pass, 0 fail; both legs exit 0.

Tier: fix


ksarma:

This branch is stacked on #1872 and inherits the ordering note posted there: the stack was cut from the original commit of #1865 and does not carry that PR's two later review-fix commits, so merging main into this branch while #1865 is still open turns three cases in `tests/test_kernel_meta_command_gate.py` red (a vouched `/fast` on a Codex session is refused before the gate parks it, since #1864 landed; the fixes route those cases through a session no Codex backend vouches for). Merging #1865 first, or merging its current head (4dff63fdc) into the stack, avoids it; no commit in this stack touches that test module.

PR #1882, opened 2026-09-19
Chat vsplit drag lab: track the live drop zone through the mid-drag pane growth

henrypinkard:
## The flake

The vertical-split drag lab dies about one in ten extension-job runs on PR CI (three times on 2026-09-18), always the same signature: `page.waitForFunction` times out 20s after `mouse.up`, the ghost still on, the same geometry. Every contributor pays a rerun.

Cause: the bottom drop zone is anchored to the pane's bottom (`.col-drop-bottom` is `position:absolute; bottom:0`, its height set inline), and `#chat-pane` (`flex:60 1 0`) **grows as the split's sessions load** (533 → 686 px in CI). So the zone rides **down** mid-drag. The driver read the zone rect once, right after it mounted, then hovered and released at that pre-growth point; once the pane grew, the point sat **above** the shifted zone, over the column zone, so the ghost never lit and the drop landed as a column move — no `place:'below'` entry, and the 20s wait died.

**This is the lab's drive, not the product.** A real user releases where they see the ghost, on the live zone; the product hit-tests the live pointer against the current zones, so a release on the live zone is correct. The lab released at a stale computed point. (Were the product to mis-hit under the same shift, the fix would be the product's — it does not.)

## The fix (drive)

Track the **live** zone the way a user does: read it, move the pointer onto it, and repeat until the pane height stops growing **and** the ghost is lit, then release there (a final re-read guards a last reflow). rAF-paced and bounded — no fixed sleep.

## Red first, deterministic (the shape: a layout shift between the zone read and the release)

The flake is timing-dependent (rare; it did not reproduce reliably under capped CPU pressure), so the driver now **forces** the shift: it pins `#chat-pane` short (the bottom zone rides to the short bottom), records where a single-read drive would aim, then releases the pin so the pane springs to its real height and the zone drops well below that point. `test_5` asserts the forced growth rode the zone down past the read point (not vacuous).

- `VSPLIT_STALE_DROP=1` drives the old single-read behaviour (release at the pre-growth point): **5 failed** (the `place:'below'` wait times out — the exact death), quoted:
  `AssertionError: 'TimeoutError: page.waitForFunction: Timeout 20000ms exceeded.'`
- Default (track the live zone): **5 passed**.
- Whole file: **13 passed**. Counts, not timings; no sleeps in the drive.


ksarma:
A review of 4f5d95547, executed against that head.

## At this head

**The growth is the timeline band collapsing; the loop exits before it.** Line 248 of `tests/test_chat_vsplit_served.py` attributes the 533 to 686 px growth of `#chat-pane` to the split's sessions loading. Every unpinned growth measured is the timeline band's autosize collapse from its 250 px loader to its two lanes (`--tl` 250px to 97px); `flex:60 1 0` is a width share and the pane's iframe is absolutely positioned, so a loading session cannot change the pane's height.

The exit at line 281 is a proxy for that event: two equal `paneHeight` reads about 10 ms apart plus a lit ghost, met on the first iteration whether or not the band has collapsed. In 11 of 12 instrumented drives (3 of 26 in a second series, those 3 run alone) the loop exited at `i=0` with the loader still rendered and the collapse landed 90 to 600 ms after the drop. Not every run races: the band settled before the drag in 2 of 5 natural runs, before the first zone read in 4 of 8, and inside the loop in 1 of 6, where the loop followed it. With the band at its loader height at zone mount, every measurement predates the settle and the window from line 286's re-read to line 288's `mouse.up` stays open (3 to 61 ms measured; the base drive's own is 10 to 36 ms). A collapse forced into it gave `5 failed` with the 20 s `TimeoutError` in every run (4 of 4, 2 of 2, 2 of 2), as did one unforced run.

Fix: before `mouse.down` at line 230, wait for the event the loop approximates: the `.col` element's `--tl` equals `Math.min(body.scrollHeight + 2, Math.round(innerHeight * 0.7)) + 'px'` for the timeline body and no `.rl-in` loader has a client rect, bounded with a named failure. With that gate added, 7 of 7 and 14 of 14 runs passed, the band at 97 px before `mouse.down` and no pane growth after any drop. Details: the `--tl` equality already holds at the loader (250 = 248 + 2), so the loader clause carries the gate; `.tl-loader` stays in the DOM `display:none` after the bars land, so test client rects, not existence; short-circuit when `document.body` lacks `po-timeline`. The loop can stay as a second guard; the final re-read at line 286 does not guard a last reflow, since the collapse can land after it (+126 ms and +129 ms in one run).

**`test_5` does not exercise the loop.** The pin is released at line 263 and `shiftTo` read at line 265, so the loop's first read at line 276 already sees the grown pane: one pass in 14 of 14 instrumented runs. A single read after the unpin in place of lines 276 to 288 passes all five cases; `if (true) break` as the exit leaves the file at `13 passed`. The `VSPLIT_STALE_DROP=1` red separates read-before-unpin from read-after-unpin, and nothing sets the switch. Fix: release the pin inside the loop after its first `mouse.move`, read `shiftTo` there, record the iteration count, and have `test_5` assert more than one iteration and `bottomZone.paneHeight` above the first read's. Executed: `5 passed` with two iterations (360 to 533); with no flag, the loop-deleted mutant `5 failed` on the timeout and the broken-exit mutant `2 failed`. Cautions: the final re-read lands the drop even with the loop broken, so the count and `bottomZone` assertions, not `place:'below'`, detect it; the ghost keeps its pinned rect across the release, so the `paneHeight` term carries the check; a `requestAnimationFrame` release can fire before the loop's first read, and one on the zone's dragover fires inside `mouse.up`; the `staleDrop` branch keeps its own release; once in this longer drive a natural collapse between the loop exit and `ghostRead` failed `test_2`, which the gate above removes.

Lines 249 to 250 and 267 to 268 misplace the stale release: `document.elementFromPoint` there returns the pane's own `IFRAME`, not a column zone (`mountZones` mounts none on the source pane), and the page's dragend cancels, leaving `cols` null; the PR body says the same.

## Smaller points

- Line 287: the zone unmounting mid-drive leaves `drop` null and the line throws a bare `TypeError` (reproduced by posting `{romp:'tabDrag', on:false}` mid-loop; not reached unforced); a guard after line 286 naming the missing zone would match the driver's other named failures.
- Line 407: `bool(os.environ.get("VSPLIT_STALE_DROP"))` treats `0` as on: `VSPLIT_STALE_DROP=0` gave `5 failed` with the timeout signature. `== "1"` matches the body's usage, though the sibling `RT_INJECT_REBUILD` uses the same form; moot if the switch goes.
- Line 277: the loop records no iteration count or settle marker, so an exhausted bound with the ghost lit reads in `RESULT` like a settle at `i=0` (reachable only by mutation; a real exhaustion leaves `ghost.cls` empty and fails `test_2`). Recording iterations, settled, last `paneHeight` and `ghostOn` in `out` is one line; a throw would discard the `afterDrop` and `botFill` evidence.

#1833 touches the same driver, so whichever lands second re-reads the other.

Runs and results are in the next comment.

ksarma:
The runs behind the review of 4f5d95547; unless stated, each is `pytest -p no:cacheprovider tests/test_chat_vsplit_served.py -k VSplitDrag` on a scratch copy of the head, in a memory-capped scope with every `ROMP_` variable cleared and the three `ROMP_*_PORT` values set to 1.

## Tests

Head as shipped:

- Unmodified, 40 runs: 40 x `5 passed`. In the last 30, `bottomZone.paneHeight` and `ghost.paneHeight` read 533 in 21 runs (measured before the band settled) and 686 in 9.
- `VSPLIT_STALE_DROP=1`: `5 failed`, every case `TimeoutError: page.waitForFunction: Timeout 20000ms exceeded.`, `bottomZone.paneHeight` 360 (the body's red reproduces).
- `VSPLIT_STALE_DROP=0`: `5 failed`, the same signature.

Drive unchanged, passive recorders added (ResizeObserver on `#chat-pane` and `#tl-pane`, `--tl` and timestamps in `readZone` and `ghostRead`, capture listeners for drag events):

- 12 runs, all `5 passed`: 11 of 12 exited the loop at i=0 with the `.rl-in` loader rendered and the first 686 read 90 to 600 ms after the drop; 1 of 12 settled before the zone mounted.
- 26 runs, all `5 passed`: in 3 (the runs made alone) the band was at 250 px at zone mount, the loop exited after one pass at 533 and the collapse landed 168 to 178 ms after `mouse.up`.
- 7 runs: exit at i=0 in all 7, 5 to 34 ms after the loop began; final read to `mouse.up` 25, 13, 11, 13 and 10 ms; the band was at its loader at zone mount in 3 of 5 natural runs and settled before the drag in 2 of 5.
- 6 runs: `loopIters` 1 and `loopBreakAt` 0 in 6 of 6; band at loader at mount in 5 of 6; one run hit the race unforced (collapse between the final read and `mouse.up`), `5 failed` with the 20 s `TimeoutError`.
- 6 natural runs: drop landed before the bars arrived in 3, collapse before the drag in 2, collapse inside the loop in 1 (iteration 0 read 533 to 686 with the ghost off, iteration 1 settled at 686, drop landed). Final read to `mouse.up` 11 to 23 ms; the base drive's zone read to `mouse.up` 20 to 22 ms in 3 runs.
- 8 natural runs: collapse before the first zone read in 4 (316 to 355 ms after goto, or during the pin) and 73, 87, 140 and 298 ms after `mouse.up` in 4; turn count unchanged at the instant of every collapse.
- 3 default runs: collapse before the first zone read in 2, inside the loop in 1 (the loop followed it).

Collapse forced into the drive's window:

- Unpin moved after the release move, pointer stationary, ghost on: 4 of 4 and 2 of 2 runs `5 failed`, the 20 s `TimeoutError`; the drag log shows dragleave on the bottom zone and no drop.
- Unpin after the re-read, before the release move: 2 of 2 `5 failed`, same error.
- Timeline body kept at its loader height by min-height and released after the release move: `5 failed` in the run where the lanes had arrived by the release.
- The `/timeline` response delayed and released between the final read and `mouse.up`: `5 failed`; released after the 2-step move, before `mouse.up`: `5 failed`.
- Released between the loop exit and `ghostRead`: `1 failed, 4 passed`, `test_2` with `280 != 343 within 2 delta` (another copy: `267 != 343`), the drop landed through the re-read.

Gate on the band's settle:

- Gate before `mouse.down` (`--tl` equal to the autosize formula, no `.rl-in` with a client rect), drive otherwise unchanged: 4 of 4 `5 passed`, and 3 of 3 with one extra zone read before `mouse.up`, band 97 px at the gate, loop one pass at 686, no growth after the drop; 14 of 14 `5 passed`, gate wait 39 to 390 ms in the 6 sequential runs with the collapse inside it every time, pane 686 through the drop in all 14. A first attempt testing `.rl-in` existence ran to the 40 s bound.
- Settle wait before the drag, or after the zone mount and before the first zone read: 8 of 8 runs `5 passed` with the band at 97 px before every measurement; four further runs across three copies `5 passed` with every read at pane 686, two of them with the `/timeline` route held until the wait; 3 of 3 single-case runs passed; with `VSPLIT_STALE_DROP=1` under the wait: `5 failed`, the red stays red. Every band log across 22 instrumented runs is a single step 250 to 97 with no later change.

`test_5` and the loop:

- Instrumented head: one pass with the first read equal to `shiftTo` (479.7, or 617.4 when the pane was already 686) in every run across seven instrumented copies (3, 5, 3, 3, 1, 1 and 14 runs).
- Lines 276 to 288 replaced by a single read after the unpin (move, ghost wait, `ghostRead`, `mouse.up`): `5 passed` in eleven runs across seven copies, `test_5` green each time.
- Exit condition replaced by `if (true) break`, whole file: `13 passed`.
- Loop deleted with no move before `ghostRead`, whole file: `1 failed, 12 passed` (`test_2`, `'on' not found in []`); `test_5` passed.
- Base file with the PR's `test_5` appended, three concurrent labs: `5 failed` on the real 20 s timeout with the ghost on at the 533 geometry, not on the missing `shiftFrom` key; with the collapse outside the base drive's window, the base fails on the missing key instead, so its deterministic fails-before is that key, not the missed drop.
- Fix (pin released inside the loop at i === 0 after the first move, `shiftTo` read there, iteration count recorded, `test_5` asserting the count and `bottomZone.paneHeight` above the first read's): `5 passed` in eight runs across seven copies, iterations 360 to 533 then 533 equal; in one further run `test_2` failed when the natural collapse landed between the loop exit and `ghostRead`.
- Fixed copy with the loop deleted and no flag: `5 failed`, the 20 s `TimeoutError`, `bottomZone.paneHeight` 360 (five copies).
- Fixed copy with `if (true) break` or the loop capped at one pass: `2 failed, 3 passed` (`test_5` on the count, `test_2` on ghost top 180 against midline 343); the drop still landed through the re-read (three copies).
- Fixed copy under `VSPLIT_STALE_DROP=1`: `5 failed` (two copies).
- Pin released from the zone's dragover: `5 failed`, `pinReleased` false at the pre-up read.

Smaller points:

- Line 287: `{romp:'tabDrag', on:false}` posted at i === 0: `died` `TypeError: Cannot read properties of null (reading 'x')`, `1 failed`; with the guard after line 286: the named `Error`, `1 failed`; the same injection at i === 1 never fired (the loop settled at i=0), `1 passed`.
- Line 250: `document.elementFromPoint` at the stale point (417, 291.4) is `IFRAME` `f-chat`; zones present: edge and bottom only, neither hit; after the timeout `cols` null, `.col-drop` count 0, ghost class empty.
- Line 277: pointer aimed off any zone: `iters` 60, `settled` false, drop landed through the re-read, `1 failed, 4 passed` (`test_2`, ghost `cls` ''); exit forced never-true with the pointer on the zone: `iters` 60, `5 passed`, `RESULT` key set identical to a settled run.
- Line 407: `bool('0')` is `True`; `not in ('', '0')` is `False` for `'0'` and `True` for `'false'`; `== '1'` is `True` only for `'1'`. `VSPLIT_STALE_DROP` appears nowhere else in the tree or `.github`.

PR #1883, opened 2026-09-19
The dropped-sends card is posted at a kernel restart: the boot echo reseed parks the post until the kernel has wired the notice door

ksarma:
Stacked on the production-profile converge rebuild PR (#1881) and lands after it; this PR is the one commit on top of that PR's head. No protocol change, no new card or row field, no knob, no timing constant.

## Symptom

After a kernel restart, a message the user typed before the restart and older than the re-delivery age line (thirty minutes by default) at that restart shows in the chat as never delivered, but the card that lets them send it again with one click never appears. The error centre row for the restart says the card could not be posted because no notice door is wired. Later restarts never post it either: the send stays flagged and silent. The card appears only when a fresh CLI is spawned for the session while the kernel is up, which is not the road a restart takes.

## Cause

The kernel builds the backend in `_sdk_locked` and wires the notice door on the class (`type(_sdk_backend).on_notice = staticmethod(post_notice)`) some forty lines after the constructor returns. The constructor's echo reseed (`SdkBackend.__init__` calling `_reseed_echoes`) runs `_mark_dropped_echoes` synchronously for every alive session whose host lease does not survive the restart; the age line puts an old human send on the `stale` list, and the `if stale:` block posts the card through `SdkBackend.post_notice`, which resolves the door with `getattr(type(self), "on_notice", None)` and, finding none, answers (None, "no notice door is wired on this backend"). The problem row records that; the `dropped` and `stale` flags reach the registry mirror through `_persist_echoes`; and at every later boot the reseed re-creates the echo with `dropped` set, which the selection's `not a.get("dropped")` term keeps out of `newly`, so the card is never posted for that send.

Hoisting the wiring above the constructor is not a fix: kernel `post_notice` checks `_notice_session_known(sid)`, which falls to `Sessions.live()`, which calls `_sdk()`, which takes `_sdk_lock`, the plain (non-reentrant) lock the boot thread holds through the constructor. A synchronous post from inside `__init__` would block the boot on its own lock for any session the names registry does not list.

## Fix

kernel/sdk_backend.py:
- `SdkBackend.__init__` creates `self._boot_notices = []` immediately before the reseed, with a comment saying why the door is not a constructor argument (the lock re-entry above).
- `_reseed_echoes` passes `park=getattr(self, "_boot_notices", None)` to `_mark_dropped_echoes`; getattr like the `forget_fed` guard a few lines above it: a stand-in that never ran `__init__` has no slot and keeps the synchronous post.
- `_mark_dropped_echoes` gains a trailing `park: list | None = None`. In the `if stale:` block the post and the problem row that names its outcome move, unchanged in text apart from `count` for `len(stale)`, into a local `post_card()` closure; with `park` a list the closure is appended to it, else it runs there as before. The flag writes, the queue re-add, `_persist_echoes` and `_wake_push` stay synchronous, so the flags reach the mirror inside the constructor exactly as today. The spawn-road caller and the resumable-reconnect caller (`refeed=False`) pass no `park` and are unchanged.
- New `post_boot_notices()` right after `post_notice`: takes the parked list whole and sets the slot to None (a second call posts nothing; nothing parked starts no thread) and posts on a daemon thread named `sdk-boot-notices`, each callable under its own try/except that files one problem row on an exception and goes on to the next. A thread because the kernel calls it under `_sdk_lock` and the door's session check re-enters that lock; the thread waits the lock out, the shape of `_push_session`'s `sdk-push-session` thread around a kernel callback. Returns the thread, or None.

kernel/kernel.py:
- `_sdk_locked` calls `post_boot_notices` directly after wiring `on_notice`, through `getattr(_sdk_backend, "post_boot_notices", None)` and a guard, like the `pending_cut` and `owns` probes below it, so a stand-in backend without the method still constructs.

The card now moves on the event it was promised for, the first restart after a send crossed the age line; its row lands on the posting thread a few milliseconds after the flags reach the mirror, where before nothing was posted on this road at all.

## Tests

- tests/test_restart_redelivery_stale.py `TheBootRoad` (new): a real `SdkBackend` over its own temp state root (removed at cleanup) with `session-hosts` off, a registry row for the module's private synthetic sid carrying a three-day-old human echo, built with reconcile off (no thread, no connect); the door is the kernel's class-level assignment, captured, wired after construction the way the kernel does; the posting thread is joined. The helper records the live atom and the mirror row at the moment the constructor returns, before the door exists.
  - `test_the_first_boot_posts_the_card_once_and_the_next_boot_posts_nothing`: the constructor posts nothing; at its return, before any door is wired, the echo and the mirror already carry `dropped` and `stale` (the mirror write is the constructor's, not the posting thread's), and the same reads hold after the join with the queue empty; exactly one post with sid, key `dropped-sends`, producer `dropped-sends`, needs_you and dismiss_on_action true, the title, the one `Send again` action and the text in the body; one `age line` problem row carrying `offered back on a card (key=dropped-sends rev=1)`; a second `post_boot_notices()` returns None and posts nothing; a second boot over the same root parks nothing and posts nothing. Before the change: `AttributeError` at `post_boot_notices`; with that call stood in, 0 posts (0 != 1). With `_persist_echoes` moved onto the posting thread instead, it fails on the at-return mirror read.
  - `test_the_parked_post_runs_on_its_own_thread_never_the_constructing_one`: a door recording `threading.current_thread()` sees exactly the returned thread, which is not the caller's. Before: `AttributeError`; stood in, the door never ran.
  - `test_a_refused_boot_post_is_said_in_the_problem_row_and_the_flags_stand`: a door answering (None, 'no session answers to "<sid>"') leaves the flags standing and the one `age line` row says `could not be posted: no session answers to`. Before: `AttributeError`; stood in, the row reads `could not be posted: no notice door is wired on this backend (the kernel wires on_notice at boot)`, the defect itself.
  - `test_a_raising_parked_post_is_one_problem_row_and_the_next_parked_post_still_runs`: two sessions park a card each; the door's first answer is a row that is no mapping, so that parked post raises `AttributeError` at its row read (the one raise a parked post can make from the door's side, since `post_notice` turns a door that raises into a refusal); asserts both sessions' posts reached the door, exactly one card posted (the second), exactly one `the boot post failed` problem row naming `AttributeError`, one `age line` row for the posted card, and the flags standing on both sends. Before the change: `AttributeError` at `post_boot_notices`. With the per-post try/except in the drainer removed, it fails on the door being reached once: the thread dies at the raise and the second session's card is never posted.
- tests/test_api_health_hover.py `TheBootRoadCall::test_the_kernel_posts_the_parked_boot_notices_once_after_wiring_the_door` (new): the real `_sdk_locked` with the loaded backend module swapped for a recorder whose `post_boot_notices` records `getattr(type(self), "on_notice", None)` at call time; asserts exactly one call, made with `km.post_notice` already on the class, and the door still wired after. Before the change: 0 calls (0 != 1). The stub-and-restore harness the existing boot_at test carried inline is lifted into a module-level `_sdk_locked_for_real(recorder)` both tests share; it also stubs `_sdk_import_notice` (and keeps the `_ensure_sdk_on_path` stub that sits behind it) and restores all five judge wires (`_LOGIN_AUTH_ENV_FN`, `_USAGE_REFRESH_FN`, `_DEFAULT_AUTH_FN`, `_DEFAULT_LOGIN_FN`, `_API_HEALTH_NOTE_FN`) to their saved values, never to None.
- tests/test_notice_card_store.py `TheDoors::test_the_backend_helper_resolves_the_hook_defensively_and_the_kernel_wires_it_at_boot`: beside the existing text pin of the `on_notice` wiring, a regex pin that the getattr, the guard and the call appear in the kernel source, in that order, after that wiring. Before the change: no match.

Targeted: the three modules 98 passed; fourteen neighbouring modules that read `_mark_dropped_echoes`, `_reseed_echoes` or `_sdk_locked`'s source 438 passed.

Full pytest (`-n 10`, short /tmp TMPDIR, extension node_modules installed, no worker lost): 6 failed, 14317 passed, 11 skipped, 2335 subtests passed in 4:57. The six, each rerun alone: `tests/test_judges_process.py::ChildRoad::test_one_pass_through_the_child_feeds_perf_and_keeps_the_kernels_bookkeeping_in_order` (a CPU-time compare) passes alone; `tests/test_kernel_remote_identity.py::RemoteIdentity::test_second_alias_absorbs_the_old_row` passes alone; `tests/test_session_host_restart.py::ServedRestart::test_with_hosts_on_the_turn_survives_the_restart_and_with_hosts_off_it_is_cut` passes alone; `tests/test_sdk_backend.py::ReconnectReconcilesInflight::test_reconnect_settles_a_turn_stranded_by_the_teardown` and `tests/test_sdk_backend.py::OptionsAssembly::test_ultracode_is_a_choice_everywhere_but_never_the_seeded_default` are `skipUnless(_HAVE_SDK)` cases that skip alone (the test interpreter has no `claude_agent_sdk`) and ran under the suite only because other modules assign an SDK stub into `sys.modules` at module level, so in a shared worker they exercised another module's stub; `tests/test_kernel_awaiting_stamp.py::AwaitingWake::test_dormant_session_is_not_woken_but_converts_to_the_dead_wait_block` fails alone the same way (the dead-wait conversion stands down because its state root has no SDK registry directory when the test runs alone) and fails identically on an unchanged copy of the base commit, so it predates this change; it reads none of the changed code, and its failure appears in other recent full runs of unrelated changes on the same machine, as do the other five. Reported here, not called green.

Tier: fix


ksarma:

This branch is stacked on #1872 and inherits the ordering note posted there: the stack was cut from the original commit of #1865 and does not carry that PR's two later review-fix commits, so merging main into this branch while #1865 is still open turns three cases in `tests/test_kernel_meta_command_gate.py` red (a vouched `/fast` on a Codex session is refused before the gate parks it, since #1864 landed; the fixes route those cases through a session no Codex backend vouches for). Merging #1865 first, or merging its current head (4dff63fdc) into the stack, avoids it; no commit in this stack touches that test module.

PR #1884, opened 2026-09-19
Pane docking phase two: the engine (grab a pane by its empty space, the live accent outline, four half-zones per pane, default off)

henrypinkard:
Tier: feature

Phase two of the pane docking kit, the engine, on the design of record `plans/pane-docking.md` (sections 2 to 6 and 11) over the merged phase-one tree module. The user's direction (2026-09-18, three asks): no title bars; grab a pane by its EMPTY SPACE; an open-hand cursor over a grabbable region and a closed hand while held; a live blue outline that follows the drag and shows where the pane will land; DEFAULT OFF behind a switch. Filed unarmed.

## The split at the plan's own seam

This pull request ships **the engine and the zones**. **Tabs as drop payloads** (a tab dropped into a zone becomes a pane there; a chat strip as the join zone) are the plan's second seam (section 4's "one drop protocol, two payloads") and follow in the next pull request: they need the chat columns to be tree leaves the engine makes on a drop, which is phase three's fold of `moveTab` into the tree. Until then, with the kit on, the shipped tab-drag zones keep working and a column they open or close is mirrored into the tree as a leaf right of the last chat leaf. A pane dragged over a chat strip already shows the refused outline ("a pane is not a tab"), so the zone set is the plan's from day one.

One plan item is deferred inside this phase and said here: the plan names the GUTTERS as a move handle too ("a plain drag resizes an edge as now; a drag that crosses into another pane's zone docks"). A gutter borders two panes and the plan does not say which one moves, so the engine's dividers resize only; the ring, the top-row run and Option-drag are the move handles. Queued as a low for the plan's owner to decide.

## What the engine does (default off)

- **Switch** (from the skeleton commit, romp_chat's): `paneDocking`, a fresh per-browser boolean in the gear's store, default false, only the literal true turns it on (`denseChrome`'s pattern). The bundle `dist/panedock-main.js` loads by `<script src>`, so the inline script count pin stays at 21. The gear gains a **Pane docking** row in the Panes section (a `rs-panes-row`, so the off-dashboard hide takes it); its checkbox writes the key through the one gear save, which the engine hears in this document and from another tab.
- **Geometry, never re-parenting** (section 2): with the switch on the engine injects its stylesheet, the `.row` stays the flex kid above the rail but lays out nothing, every `.pane` and the band become absolute rectangles inside `.col` from `layout(tree, box, 7)`, and the shipped gutters and `#gh` hide. No iframe moves in the DOM. Switch off: the stylesheet, the outline node and the dividers go, the inline geometry is cleared, and the shipped flex row is back.
- **The band keeps a fixed height via a fixed-px kid** (section 11, decided for phase two): `pane-tree.ts` gains `fixed` on a split (a kid's px instead of a ratio share; `layout` allots fixed kids first, `splitAt` beside a fixed kid mints an equal ratio share, `detach` carries a collapsed fixed kid's px up, `resize` leaves a fixed edge alone), plus `setFixed`, `edges`, `dockRoot`, `seedRowOverFixedBand`. The band's px follows the shipped `--tl` (a style observer on `.col`), so the inline autosize keeps working and the band's own divider writes `--tl`.
- **The store** (section 6): `romp-layout` `{v:1, tree, parked}`, seeded ONCE from `romp-panes` (which leaves), `romp-pane-grow` (the row's ratios) and the chat columns in row order; the three old keys are read, never written (pinned in the lab). The rail's toggles keep their meaning: off PARKS (the iframe mounted and hidden, `po-*` exactly as today), on opens at the pane's default dock.
- **Grab surfaces** (section 3): the 3 px ring around a pane's iframe (its padding, a press on the pane element itself), the empty run of an existing top row (the chat's `#tabbar`/`#tabs`/`.tab-strip-end`, the Files `.fileview-bar`; the tabs and group heads keep their own cursor), and Option/Alt held anywhere over a pane. A press on a control, a link, a field or a tab never grabs. The drag arms after a 4 px slop with document listeners (the settings-grip precedent, section 11), never native HTML5 drag; the pointer is captured on the pressed element so the moves keep arriving over the iframes (a served find: without capture the shell heard nothing until the pointer crossed a gap). Escape cancels and restores nothing, because nothing moved.
- **Cursors**: `grab` over a ring and over a top-row run, `grab` on every pane document's root while Option is held (`body.pd-alt`), `grabbing` on the shell and the panes while a pane is held (`body.pd-drag`, the iframes pointer-transparent).
- **The live outline** (`#pd-outline`, the `#col-ghost` dress: the accent wash inside a 2 px accent ring, above the focus ring, never a hit target): over a zone it is the rectangle the pane will land in, the target's half on the nearest edge (four half-zones per pane, chosen by the nearest edge as a fraction of the pane's own dimension); over no zone a thin ring following the pointer with the pane's title; over a chat strip the refused ring.
- **The drop**: `move(tree, pane, target, edge)` from the zone read against rects re-read at the release, persisted, one recompute. A refusal (the only pane, a target gone) goes through the shell's notice and changes nothing.
- **Dividers**: one per internal edge from `edges()`; a horizontal edge moves a deferred landing line and commits once at release (the row's cost rule: a grow write reflows every pane document), a vertical edge writes live, the band's edge writes `--tl`; the shipped clamp (min 120 px or a quarter of the pair).
- **The phone**: the engine stands down under the mobile layout (`__rompMobileOn`), section 8.

## Proof

**Node** (pure): `ui/webview/pane-dock.test.ts` (half-zones by fraction, the strip join zone for a tab payload and as a zone a pane is refused on, landing rects, the grab gate and slop, grow keys, the seed's parity with today's flex row over a fixed band, default docks, reconcile: park and re-open, the band's px following `--tl`, a column mirrored and parked, an all-parked store restarting from the shown set, idempotence), `ui/webview/pane-tree-fixed.test.ts` (fixed kids in layout, the fixed-band seed, setFixed, splitAt beside and onto a fixed kid, detach carrying px up, move, resize's fixed no-op, edges with path and index, dockRoot, the store round-trip and its refusals), `ui/webview/panedock-main.test.ts` (the skeleton's switch read).

**The served lab** `tests/test_pane_docking_served.py`, one hermetic kernel, Chromium on the shell page, REAL pointer events (`mouse.down`/`move`/`up`, `keyboard.down("Alt")`), the engine's state read only after two animation frames (an event wait, not a timer), the rects re-read right before every comparison and at the release:
- ON: the engine positions chat, outline, feed and the band; the band's height equals the shell's `--tl` and sits at the bottom of the pane area; the store's root is a column with the band as its fixed kid.
- Cursors by computed cursor: `grab` on the feed's ring and on the chat strip's empty run, a tab and the content root not `grab`; Option down turns the chat root to `grab` and the body wears `pd-alt`, Option up restores both; a press alone and 3 px of travel lift nothing, 14 px lifts, and the body and the pane read `grabbing`.
- The outline over the outline pane's top quarter is the pane's top half (full width, half the height less the gutter's share), over its bottom quarter its bottom half, and it moved with the pointer.
- Four drops, one per half-zone: the feed below the outline (a column inside the row), the feed left of the chat, the outline above the chat, the outline right of the feed; after each the rects and the store agree, and the band stayed fixed at the bottom.
- Escape mid-drag drops the pane where it was, the outline goes, the store is byte-identical; 2 px of travel is a click.
- `romp-pane-grow` and `romp-panes` are byte-identical after four drops.
- OFF (nothing seeded): the bundle stays inert, no engine class, stylesheet, outline node or divider, no `romp-layout`, no open hand; the same two gestures leave the pane row's outerHTML and the band's byte-identical.

**Red first at the base 48c313a5** (upstream/main; the changed product files checked out from the base, the new modules moved aside):
- `ui/webview/pane-tree-fixed.test.ts` at the base (the module exists there without the fixed kid): 10 of 10 red, the layout test by a wrong rectangle (a `fixed` array the base ignores divides the band by ratio), the other nine by the missing exports (`seedRowOverFixedBand`, `dockRoot` are not functions); the 22 phase-one tests in `pane-tree.test.ts` stay green at the base.
- `ui/webview/pane-dock.test.ts` and `ui/webview/panedock-main.test.ts`: red by absence (their modules do not exist at the base; moved aside for the base build, since an import of a missing module fails the whole test build).
- `tests/test_pane_docking_served.py` at the base under REQUIRE: 6 failed, 1 passed. The ON tests fail on the measured absence, test 1 `False is not true : the engine came on and the panes have rects` and tests 2 to 6 `the engine never came on (no bundle, no switch, or no rects): ready False, rects0 {}`; the OFF test passes at the base as it must (inert by construction, a guard).
- `tests/test_settings_tabs_recut_browser.py` `five_panes_rows` at the base: 1 failed: `covered false, count 5` (the base gear has four Panes rows and no Pane docking row; the head's count is six with the row).

**At the head**:
- `tsc --noEmit`: exit 0. The four node test files (`pane-dock`, `pane-tree-fixed`, `pane-tree`, `panedock-main`): 46 tests, 46 pass.
- The lab under `ROMP_SERVED_TESTS_REQUIRE=1` (chromium): 7 passed. `tests/test_settings_tabs_recut_browser.py` with it: 20 passed for the two modules.
- The whole extension suite (CI's glob, `out-tests/**/*.test.js`): 5086 tests, 5080 pass, 0 fail, 6 skipped.
- The whole Python suite (`uv run --python 3.13`): 14336 passed, 60 skipped, 2329 subtests passed, 0 failed (33 min 39 s) at the second commit, after the off-dashboard rows pin moved to five; the first lane at the engine commit had exactly that pin red (`6 != 5`) and everything else green (14335 passed).

## Where the plan and the code disagree

Nothing found that contradicts the plan's decisions; two places the plan's text is silent are named above (which pane a gutter drag moves; tabs as payloads before phase three) and decided the plan's way (defer, split at the seam).

## Round two (the read of 0a567790)

- **HIGH, the iframe collapse.** The ring rule sized the iframe `width:auto;height:auto`; an absolutely positioned REPLACED element then takes its intrinsic 300 by 150 px and ignores its far offsets (CSS 2.1 10.3.8 and 10.6.5), so every pane's content sat in a 300 by 150 box at the pane's top-left while the pane box itself was right. The rule now sets `width:calc(100% - 6px);height:calc(100% - 6px)` beside `inset:3px`; the vertical split's top iframe keeps `height:auto` (its flex height). A plain press inside a pane's content (not the ring, no Option) is re-checked: the content's cursor at rest (`auto`), 14 px of travel arms nothing, no outline, the store and the rects untouched.
- **MEDIUM, the lab's geometry source.** `rects()` now carries each pane's iframe rectangle, and the lab pins for every pane and the band, after the seed and after each of the four drops: iframe left and top are the pane's plus 3, width and height the pane's less 6 (measured at the head: a 461 by 766 pane holds a 455 by 760 iframe at +3, +3; the band 1397 by 97 holds 1391 by 91). Red at 0a567790 by the collapse.
- **Low (a)**: the shown set names the panes whose elements exist and the reconcile prunes a parked pane whose element is gone (a park keeps a MOUNTED iframe; a closed chat column has none). Node pin: the outline hidden by the rail stays parked, the closed column's park is pruned; without a present list an older caller keeps every park.
- **Low (b)**: while the kit is on the shipped grow writers (`__rompSplitGrow`, `__rompSplitShrink`, `__rompGrowFair`, `__rompGrowFairIfNew`, `__rompUnregisterPane`) answer as no-ops and the originals return when the kit goes off, so the kit is the only geometry writer and the old keys stay the OFF path's source of truth (section 10). Lab pin: a chat column opened with `__rompMoveTab(sid, "new")` and closed with `__rompCloseSplit(2)` under the kit is mirrored right of the last chat leaf, its iframe fills its pane, it leaves the tree and the parked set, and `romp-pane-grow` and `romp-panes` are byte-identical across the open and close.
- **Low (c)**: the driver's five fixed sleeps are bounded waits on the conditions they stood for: the engine's settle (the store written, a divider mounted, the stylesheet applied), the Option class on and off, the pane documents complete on the OFF page, and two animation frames for gestures that must do nothing.
- Screenshots in drops: `romp_timeline-panedock2-on.png` (the kit on, every pane filled, feed | outline | chat over the band after the four drops) and `romp_timeline-panedock2-drag.png` (the feed held, the outline over the outline pane's bottom half).

**Red first at the previous head 0a567790** (the two engine modules checked out from it, the head's tests):
- `ui/webview/pane-dock.test.ts`: the prune pin red (`reconcileShown: a parked pane whose element is gone ... is pruned`), 12 pass.
- `tests/test_pane_docking_served.py` under REQUIRE: 5 failed, 4 passed. The iframe pins: `chat-pane iframe width fills the pane less the ring (after the seed): 300 vs 455`, `(after drop 1): 300 vs 689`, the column's `300 vs 338`; the grow store: `romp-pane-grow untouched: '{"chat":695,"feed":344,...}' != <the seed>` (the shipped writers wrote the kit's geometry across the column open and close); the content press: `14 px of travel from a press over content lifts nothing: dragging True, outline True, bodyCursor grabbing` (the collapsed iframe left the pane element exposed under the pointer, so a plain press over content armed a move). The four that pass at the previous head (cursors, the outline across two zones, Escape, the OFF page) measure what round one already held.

**At the head**:
- `tsc --noEmit`: exit 0. The four node test files: 47 tests, 47 pass. The lab under REQUIRE (chromium): 9 passed (with `romp-pane-grow` seeded to a real value so the byte-identity pin is not of two absences); `tests/test_settings_tabs_recut_browser.py` with it: 22 passed for the two modules.
- The whole extension suite (CI's glob): 5087 tests, 5081 pass, 0 fail, 6 skipped.
- The whole Python suite stands from the previous head (14336 passed, 60 skipped, 0 failed): this round changed two engine modules (covered by the extension suite and the lab) and the lab itself.


ksarma:
A note after the merge; every file and line below is on 0a567790d, where the evidence was executed. Commits 135be696d and 032afaf31 edit no line named here. The kit is behind the `paneDocking` switch, off by default.

- panedock-main.ts:386. The frame-side press gate tests the target with `instanceof Element` from the shell realm, false for every iframe node, so the top-row and control reads are always false: the strip's empty run never arms a drag and Option-drag ignores controls. In served probes the run press armed only with a duck-typed `closest` check, under which an Option press on a control yields; no test arms a drag from a frame; every lab drag starts on a ring.

- pane-dock.ts:195 with pane-tree.ts:274 and panedock-main.ts:370. The reconcile writes the band's height through `setFixed`, blind to the split's direction. Once the Sessions band shares a row split, the next height write pins its width to that value and the divider beside it writes `--tl` from a vertical pointer; both roads were executed, a side-zone drop and a band drag by ring or Option press in its iframe. A column-only write leaves the height stale and a target-only refusal misses the band as source; retarget the px onto the nearest column-ancestor slot, refuse the band as source and target of a side zone, and key the divider's fixed branch on the edge's direction. The band-only tree's fallback at pane-dock.ts:189 docks a returning pane to the right, the same way; docking to the top restores the seed's column.

- panedock-main.ts:187 with 581. The only resize listener lives inside `start` and goes at stop; the boot block hears settings and storage only: a narrow boot never starts the kit on widening, and a desktop page loses it after narrowing and re-widening; a boot-block listener (replacing the in-start one) restored both.

- panedock-main.ts:520. On a live divider `clampDelta` clamps a cumulative offset against a pair measured from already moved ratios, so past the moving limit the value alternates and the divider stalls near two-thirds of the pair, short of the 120 px floor; clamping against the press-time pair reached it.

- test_pane_docking_served.py:130 and 200, gear.js:449. No divider is dragged, both contexts at scale 1: a mutant with `onDivMove` returning early passed every test; the band's edge also writes an absolute px from the pointer, not a press-point delta. The switch is never toggled live, so `stop` and the boot listeners run untested; a toggle leg caught a gutted mutant; a landing leg drives both write paths (same-document event, settings-frame storage event). Nothing pins the gear write, its reflect or settings.ts:87.

- panedock-main.ts:105. The injected reset `.tab,.tab-group-head,.tab-row-line{cursor:auto}` overrides the strip's own pointer on tabs and group heads; the lab's inequality against grab passes on it; drop the first two selectors.

- panedock-main.ts:217. The frame list excludes bottom-split iframes, so a bottom frame wired through the columns event at 186 keeps the injected stylesheet after the switch goes off, and one restored at load is never wired; widening `allFrames` covers both.

- panedock-main.ts:94 and 86. The shell stylesheet copies only the dark half of the shipped dress: under the light theme the outline wash, label and #333 divider line stay dark; light twins after the dark rules, both directions, close it. The dividers also lack the gutter's hover cue; the shipped classes cannot supply it, the kit hiding `.gv` and outranking the hover rule.

## Changed by the later commits

- test_pane_docking_served.py:430. A rail re-show under the kit wrote the grow key through the shipped fair-grow writer, byte-identical to the off context's, so the gap was only the body's never-written claim; no-op writers were argued against, since a rail toggle then leaves the off path different. 135be696d stands those writers down until stop; nothing was executed against it.

- pane-dock.ts:134 and panedock-main.ts:230. The seed weights the row from the grow key alone, which only a gutter drag writes, so a browser that never dragged one seeds equal shares, not the shipped ratios; the row's live `--g-<key>` properties can stand in. 032afaf31 seeds that key in the lab, which no longer represents that browser.

## Smaller points

- panedock-main.ts:476. The refusal notice is the tree module's raw error string; only "pane not in the tree" is reachable (a reconcile mid-drag), so one plain sentence for every throw from `move` suffices; neither it nor the phone stand-down at 149 is tested.
- panedock-main.ts:540. The band divider lacks the shipped content-height ceiling; a drag past the content is undone by the next autosize.
- pane-dock.ts:186. A rail hide then show re-opens through the generic halving of `splitAt`, not the off path's equal share, which belongs on re-open only.
- test_pane_docking_served.py:440 and 361. `engineOn` is false with the bundle absent or inert; the tab inequality at 361 passes on None when the 161 lookup misses.
- plans/pane-docking.md:3 and 288. The Status line reads as proposed with no landing commits; section 6 says Chat/Layout while the switch sits under General, Panes (for the dashboard-only hide), named in the body, unreconciled in the plan.
- gear.js:449 and gear-tabs.test.ts:48. The T317 comment on the Files control moved from 448 onto the pane-docking handler; the node sections pin lacks rs-panedock, unlike the Python twin.
- docs/guide.md:388. No guide text for the switch; three of four Panes rows have none and the sub-line states the surfaces; optional.

PR #1885, opened 2026-09-19
Held mail is approved on the notice model: kernel action kinds, the held-mail notice card, the pane's quarantine family retired

henrypinkard:
Held mail is approved on the notice model: the kernel defines action KINDS, the held message becomes a notice card, and the pane's hand-built quarantine family leaves (`plans/notice-cards.md`, "Action kinds and the held-mail card", merged as PR 1875). Feature tier: a capability inside the existing notice model, no new contract.

## What the user sees

- A message held from a DIRECTED peer is a notice card under the recipient session, in Blocked: title `New message from <sender>`, producer `postal`, body = the route line (`from <host>:<sender> to <recipient>, held because peer <host> is DIRECTED`) and then the message text, with **Approve** and **Deny**. No Edit: nobody edits mail.
- Approve delivers, Deny drops, both through the kernel's existing quarantine act road (`_bus_quarantine_act`, which carries the recipient sid and dials the bus by its port record since PR 1848). Deny asks for the optional note back to the sender first, in a small prompt on the document body (two choices, no Cancel; the backdrop closes without deciding). Both buttons latch on a click and let go on the kernel's answer; a refusal toasts its reason and the card stays; a success takes the card off at once.
- The card modal shows the message text under the recipient's name. The name is struck through only when the recipient is dead: every notice card said `live: false` before, so the older card's modal struck the recipient through as a dead session's. `_notice_cards` now takes the build's alive roster.
- Clear dismisses the card as any card's Clear does; the held file stays for the bus.

## Design points (the merged section, followed)

- **Kinds, never bare routes.** `NOTICE_ACTION_KINDS = ("send", "quarantine")`. A stored action is `{label, kind, body}`; `_notice_action_body_check(kind, body)` holds each kind's shape (send: non-empty text, no target, no leading slash; quarantine: exactly `{mid, verdict}`, `verdict` approve or deny, the mid path-safe). A kind the table lacks or a bare route is refused by name. The older `{label, route: "/send", body}` shape (rows written before this PR, the older pane's wire) reads as the send kind; `NOTICE_ACTION_ROUTES` stays as a name for it.
- **The wire.** The pane posts `noticeAction {itemId, sid, kind, body, input?}`; `route` from an older pane is accepted and mapped. `input` may carry only what the kind names: `{note}` on a quarantine deny; anything else is refused ("the action takes no 'text' from the click"). The kernel answers `noticeActionDone` by the card's id as before. `quarantineDecision` and `quarantineRefused` are gone (an unknown op now).
- **The card is the kernel's.** `_held_mail_backfill()` runs at the notice attach in `build_feed`: for every `STATE/postal/quarantine/<mid>.json` with no live post row and no archived revision for (recipient, mid), it posts the card through `post_notice` (key = mid, owner = `toId`, needs-you, `dismissOnAction`, `t` = the hold's time). Memoized on the directory's stat; a refusal is said once per (mid, reason) on stderr. An unknown recipient files owner-less, informational, without actions. On a decision the runner appends an `expire` row (the card retires whatever the ledger later says) and then the `acted` row and the dismissal.
- **Removal census.** Kernel: `_quarantine_cards`, the `quarantineDecision` op branch, the `quarantine` kind in `_CODE_BOARDS` and `_BOARD_KINDS` (the `quarantine:` prefix stays in `_CLEARED_NO_SESSION`: older clear logs hold those ids). Pane: the `blocked` quarantine fields, the latch kind, `qApprove`/`qDeny`/`qbody`, the `isQuar` block, `showQuarantineDialog` (replaced by `showDenyNoteDialog`), `quarWho`, the `showBlk` exclusion, `clearable`'s exclusion, the `quarantineRefused` handler, the render counter (`renderSeq`, the gate's `seq` and its quarantine term, the gate's `blocked` slice), the sender-repo helpers only the held gist used. `board-def.ts`: the `quarantine` kind, its descriptor and `kindOf` branch. `plans/card-boards.md`: the family's lines.

## Tests

- `tests/test_kernel_trust.py`: `HeldMailCards` (the card's fields and actions; `live` from the roster; idempotence and the directory memo; a decided or dismissed message never re-posted, before and after the archive pass; an unknown recipient owner-less; a bad id skipped; no directory) and `HeldMailDecision` (the op: a refusal answers by the card's id and dirties nothing; a success retires the card and rebuilds; the click's input limited to the kind's; the older ops unknown); `BusPortRecord`'s decision test re-aimed to the action road.
- `tests/test_notice_card_store.py` `ActionKinds`: the stored shape and the route compatibility; every refusal message; the quarantine runner with the bus stubbed (owner as recipient, the note as feedback, a bus refusal and a fault as the answer, whole-body matching).
- `tests/test_held_mail_served.py` (new served lab): a held file written before boot becomes the card at the first build; the modal shows the message text under the recipient's unstruck name; Approve with no bus is refused as unreachable, re-arms and toasts; Deny's prompt, its backdrop, the bare deny; Clear leaves the held file.
- `ui/webview/feed-render-incremental.test.ts`: the held-mail card's Approve wire, both-latch and re-arm; Deny's prompt on the body, the bare verdict, the note as `input`, the backdrop; the send wire carries the kind. `feed-quarantine.test.ts` rewritten as the removal's pins and the new shapes. Pins re-aimed in `feed-card-gate`, `board-def`, `feed-sess-clear`, `feed-judge-auth`, `feed-apierror-working`, `feed-flip`, `feed-awaiting-swirl`, `pr-links` tests and `tests/test_card_boards.py` (six hand-built families).

## Red-first (the head's tests copied into a detached worktree at the base 9bcfdc41, run there)

- Extension: 16 red of 159 in the ten changed files: `board-def` (kinds table, kindOf), `feed-apierror-working` (the action row), `feed-awaiting-swirl` (the body order), `feed-card-gate` (no seq), `feed-judge-auth` (showBlk), all four `feed-quarantine` pins, the two held-mail harness tests and the two send-wire expectations in `feed-render-incremental`, `feed-sess-clear` (clearable), `pr-links` (the import). The eleven pins the whole suite found later (`feed-card-layout` x2, `feed-checkstatus`, `feed-parked-handoff`, `federation-notice`, the six `quarantine-route` tests) failed at the head before their re-aim, so their red is the head's own run.
- Python: 12 red in `test_kernel_trust` (every HeldMailCards and HeldMailDecision test, the BusPortRecord decision test), 7 in `test_notice_card_store` (the six ActionKinds tests, TheDoors' build_feed pin), 1 in `test_card_boards` (the six-family census), all 5 of `test_held_mail_served` (no card at the base), and the 5 of `test_notice_cards_browser` (the driver's kind wire is refused at the base's door, so its second card never appears). Everything else in those files stayed green at the base.

## Counts

- 26 files changed, 1131 insertions(+), 606 deletions(-) (against upstream/main after the carry). Production files: `kernel/kernel.py`, `ui/webview/feed.ts`, `ui/webview/feed.css`, `ui/webview/feed-card-gate.ts`, `ui/webview/board-def.ts`, `plans/card-boards.md`. One test file added (`tests/test_held_mail_served.py`), one retired (`ui/webview/quarantine-route.test.ts`, its one surviving pin moved into `feed-quarantine.test.ts`), fourteen test files re-aimed.

## Rounds

- Before the head mail: CI's Python 3.12 run failed one test in a module this PR does not touch (`test_rename_route`, the poisoning guard). Cause: the held-mail tests wrote a names entry under the shared placeholder sid into the run's one state root, and the rename route read the rename of that sid to `web` as its own-name no-op. Fixed in the tests only: a private synthetic sid and a tearDown that removes what they wrote.
- Round two (the manager's review at 178a61af): one medium, four lows.
  - **Medium.** The runner never read the held file, and the bus's recipient check reaches only a hold it does not have, so a producer's card under one session's name and colour decided another session's mail (executed: a card under `api` naming a message held for `web`, its Approve delivered to `web`). Now the held file's `toId` must equal the card's owner: refused at the post when the file exists (`_notice_actions_check(actions, owner)`), and at the click before the bus is asked (`_held_mail_record`). Pinned in `test_kernel_trust` (the post, the click with the bus stubbed: never called, the card standing, the file still held) and `test_notice_card_store` (the post in its own world).
  - **Low 1.** A hold posted owner-less while its recipient was unknown was posted again under the recipient once the names entry existed (three cards for two holds). The backfill reads the key's standing in every home it could have been posted to: a live owner-less card for a recipient known now is re-homed (the recipient's card posted, the owner-less one expired); an archived owner-less card (dismissed, decided, expired) is left alone. Pinned.
  - **Low 2.** `_bus_quarantine_act` returned `bus HTTP 200` as the error of a success whose JSON carried no error member, and the runner passed it out on `noticeActionDone`; an empty error on ok. The three fake-bus pins now read the empty error.
  - **Low 3.** The held-mail tests wrote and deleted in the run's shared state root wholesale; they now restore the shared files they append to (the owner-less home, its archive and index, the cleared ledger) to their prior bytes and unlink only their own sids' files and their held files.
  - **Low 4.** Removal residue: the mangled record comment, the stale `clearable` comment, the unused `hostDownNote` import, the gate comment naming `quarWho`.
  - Red-first for the round (the head's tests copied into a detached worktree at 178a61af): the re-home test, both post-time refusal tests, the click refusal (the stubbed bus was called) and the two empty-error pins red; every other test in the two files green there.

## Sibling and main

- PR 1861 (`romp_cards-boards3`) merged while this was in verification and is carried: one conflict, in `tests/test_card_boards.py` (its builder census versus my removal of the quarantine builder), resolved to five literal feed stamps plus the notice family's row pair. Its `post_notice` gains board and category; my backfill passes neither. Its one needs-you rule (`_card_needs_you`) reads the held-mail card by its category, so the badge and the ring see it. PR 1863 carried too.

## Verification

- Red-first as above. On the head as pushed (after the carry): tsc, the changed kernel modules (`test_kernel_trust`, `test_notice_card_store`, `test_kernel_goal_compaction`, `test_card_boards`), the fourteen changed webview test files, the two notice-family served labs (`test_notice_cards_browser`, `test_held_mail_served`), and the whole extension suite.
- On the head before the carry (b101376c, main at aab092e1): the whole extension suite (5065 tests, 0 failures) and the whole Python suite in three capped shards; the two unit shards green but for `test_kernel_judge_model::JudgeSettings::test_judge_run_applies_tier_effort`, which passes alone at that head and at main (a shard-order flake, not this change); the served-labs shard (103 files, serial) green.

## Out of this PR

- The chat pane's approval box above the background-agents box and the blocked ring's lab: the section's next phase, a second feature PR on this one.
- The settings plan's kind, a Clear that decides, an Edit on the card, moving the bus's hold into the notice store.

Synthetic fixtures only (the notes-api world, `TESTHOST`, placeholder ids); no personal identifiers; no em-dashes in added lines; the ruled-out word appears in no added line.


ksarma:
A note after the merge: evidence executed against 764d19dd2, the merged head, and the earlier head 0c21cca73 as pushed; lines are on 764d19dd2.

The one mechanism that loses mail is kernel/kernel.py:24434, in `_held_mail_backfill`: the card is keyed on the bus message id, and every held file whose id fails the 64-character key rule is skipped with a bare continue before the homes loop. Bus ids of the shape `px-<epoch>.<pid>_<32 hex>.<host>` exceed 64 characters once the sending host name reaches 12 characters under a 5-digit pid or 10 under a 7-digit pid (bus-minted ids of 65 to 69 characters posted nothing; a 64-character control posted). Such a message gets no card and no stderr line, and with the hand-built quarantine family retired it has no surface at all, where the base surfaced it through that card. No test uses an id of the bus's shape. Fix: the mid as the key when it fits, else "held-" plus 40 hex characters of its sha256, the same derived key in the homes loop and archive checks, the real mid in the action body, the stderr line in place of the continue; patched, Approve and Deny reach the bus with the real mid, the runner unchanged. Add a HeldMailCards test with a bus-shaped, 12-character-host id.

## After the commit titled Round two (90faba8ed)

- kernel/kernel.py:24421: the re-home of an owner-less card runs only when the quarantine directory's stat moves. The trigger is a recipient with no names entry yet (not a session ending: a names entry is never deleted), and the card stays informational, without Approve or Deny, until another held message or a decision moves the directory or the process restarts, against the body's unconditional sentence; include the names directory's stat in the memo key. At 24472 a failed expire append is unreported and never retried, since the recipient's live row skips the mid first; read and say the error.
- The recipient check landed kernel-side (kernel/kernel.py:24603, verified); the bus side did not. postal/postal_service.py:4563 compares the sid with its roster only when no record exists, so an approve or deny with another session's sid, from an older kernel or a direct POST, still decides (executed); refuse when a record and a sid are both present and the toId differs.
- kernel/kernel.py:24452: a dismissed owner-less card whose recipient becomes known is re-homed as an actionable needs-you card when `_compact_notices` has not run, and left dismissed once it has (both executed). Treat a live owner-less row whose item id is in `_cleared_ids()` as archived.

- kernel/kernel.py:24411: nothing retires a card whose held file left the directory, leaving Approve and Deny answering "no held message"; the triggers are a restart between the bus's answer and the expire row, a direct POST or a removed file, not a client timeout (the approve arm is a fast local write). Expire live rows whose quarantine-kind action names a mid with no file, an owner-less informational card included, a missing directory read as empty.
- kernel/kernel.py:24388: `_held_mail_body` posts the whole text, so a message over the 64 KiB body cap (cutoff 65470 bytes with the route line) is refused and gets only a stderr line, where the base built the card; clip by bytes with a marker, since the runner never reads the card text.
- kernel/kernel.py:24527 and 24571: the two reads of the older action row (label, route "/send", no kind) reach no test, since every route in the tests is a `post_notice` argument the door rewrites before storing; they matter for rows written before this PR and any writer bypassing the door.
- ui/webview/feed.css:2125: `.fask-nbody` has no clamp, so the card face renders the whole text against the merged plan's first-lines-then-modal split. A max-height on `.fitem .fask-nbody` (not line-clamp: the body has block children) reaches every notice card face, the class being shared. A backstop under `.feed-modal-notice` lifts it; the modal mounts outside any card, beyond the card rule.
- ui/webview/feed-sess-clear.test.ts:51: the title says the session Clear never takes a held message while the body pins `return !it.provisional`, under which a held message's card is cleared. Not new (the board-wide Clear all took a held card at the base): retitle, with no guard in `clearable`.
- kernel/kernel.py:24450: an unreadable revision index in any home visited drops the message silently, and the loop now reads the no-owner home's index for every message with a known recipient: corrupting that index alone gave 0 cards, against 2 at 0c21cca73. Say it on stderr rather than drop the check, which re-posts a dismissed owner-less card as actionable.

## Smaller points

- kernel/kernel.py:24446: check and post are two locked steps; two builds double-posted (up to 261 of 300 barrier trials), one card whose id moves between revisions. The trigger is a GET /feed.json racing the pusher's first build or two bare `build_feed` calls, not Clear all. One dedicated lock, not `_notice_lock`, gave 0 of 300.
- kernel/kernel.py:24359: 50 notices posted after the card (not before) drop it from the live projection, the archive pass archives it as capped, and the archive check refuses a re-post; exempt held-mail rows carrying a quarantine-kind action in both the projection and the archive pass.
- kernel/kernel.py:24436: a message dismissed under the older id `quarantine:<mid>` returns as a needs-you card, where the base hid it. Skip a mid whose older id is in `_cleared_ids()`.
- Stale text, unpinned: kernel/kernel.py:24412 (docstring, no re-home), 43015, 67529 to 67530, docs/guide.md:923 (Edit), docs/reference.md:2444, feed.ts:1065 to 1066, 1222 to 1224, 2717 to 2718, 2763 to 2765, feed.css:1107 to 1108 (`.fdismiss.fq-ok` has no element; its siblings stay).
- feed.ts:2070: `renderNoticeModalBody` rebuilds Approve and Deny on every render (a rebuild older than this PR), so a push between mousedown and mouseup drops the click (no engine dispatches a click on a node detached before mouseup, so delegation cannot recover it); gate on the card's nkey.
- feed.ts:2378: an older peer kernel still sends the hand-built card (blocked.state "quarantine") over federation, rendered here as a picker chip with Clear and no Approve or Deny; update mode defaults to ask, so releases mix. Keep a read of that state: no chip, a line naming the holding host's dashboard.
- feed.ts:6646: `showDenyNoteDialog` has no Escape close, dialog role or aria-modal, as the base's deny step had none; the feed's other dialogs share the ARIA gap.
- kernel/sdk_backend.py:3343, 3347, 3349: `dropped_sends_card` still writes route "/send". Only `_NOTICE_ROUTE_KINDS` admits it; write kind "send".

Test pins and fixture scoping are in the next comment.

ksarma:
Tests, continued from the note above: pins that are proxies, absent, or on dead code, and fixtures that read or write outside their own state, executed at 764d19dd2, the pins under the named mutation.

- tests/test_notice_card_store.py:324: the older action row (label, route "/send", no kind) is exercised only through `post_notice`, which rewrites it to a kind before storing, so the two raw-row reads at kernel/kernel.py:24527 and 24571 are unpinned: both mutated, 479 to 486 tests pass (one preexisting unrelated failure unchanged). Fix: seed a card through `post_notice` with kind "send", not a route. Append a raw row carrying route and no kind through the store's append under its lock, with its own key and rev 1. Clear `_NOTICE_MEMO` inside the same lock after the append (the memo is stat-keyed, so clearing before it can leave the seed-only read cached). Assert the on-disk row still carries route with no kind, then that the card's action kind is "send" and that `_notice_action` succeeds for both "send" and "/send" with delivery stubbed and two deliveries recorded. The probe passes at the head and fails distinctly under each mutant ("" is not "send"; "no such action on that card").
- tests/test_kernel_trust.py:281: the assertion labelled as the memo passes with the memo check deleted, since the live-row check also yields 0 (43 tests in the file and 77 across the notice-card suites stay green under the mutant). Fix: assert the slot of `_HELD_MAIL_MEMO` equals the directory path plus its stat key. Then patch the module's listdir through mock.patch.object with a stub raising a non-OSError (an OSError stub is caught by the listing's except and stays green under the mutant) and assert the second call returns 0 without reaching it; applied to a copy, 10 pass unmutated and 1 fails under the mutant.
- tests/test_kernel_trust.py:334: the re-home test proves the re-home only by writing a second held file, which moves the directory; with the names entry alone the count stays 0 until the memo slot is cleared, and dropping the second write reproduces the failure. Once the names directory's stat is in the memo key (it moves on entry creation or removal, not on a rewrite), drop that write and assert on the names entry alone, keeping a leg where nothing changes and nothing posts.
- kernel/kernel.py:24466: the stderr write can be replaced by a no-op with 177 held-mail tests green, and the line reads "[notice] held message" with no sid where the module's eight other notice lines read "notice: " and name the sid's first eight characters (every expire path in the module is silent, so the re-home's silence is no deviation). Fix: reword the line to that voice, then stub `post_notice` to refuse (preferred over an over-cap body, since clipping the body would make that path stop refusing) and capture stderr. Assert one line naming the mid and "no card" on the first backfill and none on a second after resetting `_HELD_MAIL_MEMO`, not the said-set.
- kernel/kernel.py:67533: the `or msg.get("route")` fallback is unpinned; dropping it leaves 151 tests green across the held-mail and card suites. Fix: dispatch a noticeAction carrying route "/send" and a text body through `Handler._dispatch_ws` against a send-kind card with delivery stubbed, asserting noticeActionDone ok true, beside the Handler pins in tests/test_notice_card_store.py (passes at the head, fails under the mutation).
- kernel/kernel.py:24388: no test pins the over-cap refusal path; the clip needs a red-first test with a body over `NOTICE_BODY_MAX`.
- ui/webview/feed-sess-clear.test.ts:51: the header Clear all taking a held message's card has no executed test; mutating `clearable` to exclude postal notices breaks nothing. feed-sess-clear.test.ts has no harness and `sessionCards` is module-private, so the executed test goes in feed-render-incremental.test.ts beside the header Clear-all test: dispatch a held-mail notice card (flavour notice, producer postal) and assert askClearMany includes its id, keeping the regex here as the body pin.
- ui/webview/feed.ts:731: `sessionColors` is written on every frame (702, 6357 and 6359) and read by nothing, since its one consumer left with this PR, while feed-color-echo.test.ts:32 and feed-quarantine.test.ts:59 to 61 pin the dead writes: deleting the map, the writers and the index loop leaves tsc clean and fails exactly those two tests. Fix: remove the map, its writers and the loop; trim only the colour sentence of the preamble at 726 to 729; drop the feed-quarantine test that pins the map (the whole test); re-aim feed-color-echo.test.ts:32 at the colour echo; reword the negative references at feed.ts:6834 and file-view.test.ts:828 and 832.
- tests/test_kernel_trust.py:253: after the commit titled Round two (90faba8ed) dropped setUp's wipe of the quarantine, notices and archive directories, HeldMailCards and HeldMailDecision assert over the whole shared state root: a module seeding one notice card or one held file ahead of the file turned 9 tests red (35 or 36 passed, 43 alone); a copy of the earlier head's file was immune. Fix: scope `_cards()` to the fixture's sids, and to its mids in the no-owner home. Extend the filter to HeldMailDecision's two direct `_notice_cards` calls. Scope every backfill return assertion, or drain once in `begin()`.
- tests/test_kernel_trust.py:317: the no-key test unlinks every file in the shared quarantine directory and removes the directory itself (a planted peer file was deleted), against the body's sentence that the tests unlink only their own files; latent today, since no other module writes a held file there. Fix: unlink only "bad id!.json" and run the empty and absent-directory legs against a private temporary directory by patching `_held_mail_dir` with mock.patch.object (both call sites read it through the module global); with that, the peer file survives and 43 pass.
- tests/test_kernel_trust.py:347: `_compact_notices(now=10**9)` archives every session's dismissed rows in the shared state root (a planted foreign dismissed row was archived: new archive file, new index, live row cleared). `_HeldMailFixture` restores only the notes files it names, and the decided-or-dismissed test's archive call with now=5000 does the same. Fix: snapshot and restore the whole notices and notices-archive directories in the fixture (verified, 10 of 10 pass with no file changes); rebinding `jd.STATE` for the archive call alone fails, since the notice directory under a fresh root lists nothing and the greater-than assertion fails, so a rebind must wrap the whole test or class. Reword the body's sentence to what the fixture restores.
- kernel/sdk_backend.py:3343: no test runs `dropped_sends_card`'s actions through the kernel's check; the redelivery test's on_notice stub validates nothing. Fix: re-aim tests/test_restart_redelivery_stale.py:128, 157 and 200 to kind "send" (red at the base) and add one test through `_notice_actions_check(actions, owner=None)`.

PR #1889, opened 2026-09-19
Queued sends reach the agent one message each: the SDK feeder holds the next text until the CLI has taken the last one

ksarma:
Stacked on #1883 (https://github.com/romp-on/romp/pull/1883, the dropped-sends card's boot-road post) and lands after it: this PR is one commit on top of that PR's change (its head before it merged main into its branch, 2378fa77a; this head merges onto the PR's current tip and onto main with no conflict). That chain stands on #1881; #1872, #1865 and #1863 have merged.

## Symptom

Two messages sent to a session during one open turn (a composer message, then a reply) reach the agent as ONE message: the first text followed by the second with no boundary. The chat shows one bubble wearing both texts, and the first message's optimistic bubble, matched by its own text, never finds its landing and stays up until it is dismissed.

## Cause

`SdkSession`'s `inputs()` forwards every queued text to the CLI's stdin the moment it is queued, mid-turn included. The CLI drains every prompt it holds into one user record when it next reads its queue (at a turn's start, or when a running turn ends). Nothing records that the last fed text is still in the CLI's own queue, so a second text queued before that drain is fed at once and the two are drained together. The settle then wakes the feeder in the window the CLI drains, so a text queued behind a mid-turn send lands in the same record as the send.

## Fix

All in `kernel/sdk_backend.py`; no new op, no protocol change, no knob, no timer. Every hold lifts on a streamed frame, a transcript record, the CLI's exit or a move's own result.

- **One fed text at a time.** `inputs()` feeds one text and holds the rest (`_untaken`) until the CLI demonstrably took it. The take is witnessed by exact events only (`_turn_frame`, `_untaken_taken`): a turn frame after a feed from idle (the rename ping's existing rule, generalised to every send); a turn frame after the result of the turn a mid-turn feed went into (the CLI drains its queue at a turn's end, so the next turn's first frame proves the drain); or the fed text's `queued_command` attachment landing in the transcript (a mid-turn splice at a tool boundary, which lets the next text follow it into the same turn). Only the init, assistant, user and result frames of the main conversation count: task, hook, status, progress and rate-limit frames stream independently of the queue and release nothing, and a subagent's frames (an assistant or user message tagged `parent_tool_use_id`, the sidechain `msg_to_atom` already drops) are not turn frames either, since a backgrounded Task streams on after the main turn's result and says nothing about the main conversation's prompt queue.
- **The CLI's own turns count.** A turn frame at `inflight` 0 raises it and starts the turn's clock (`since`, which `snapshot()` reads for a turn in flight, as a fresh feed's pop does; the stop reading is cleared the same way), so a text fed into a drained turn or a notification-started turn is mid-turn and waits for its own take instead of clearing on the running turn's next frame. `busy()`, `queue_recallable` and the reconnect gate (`_do_request_reconnect`) read the hold too: in the gap after a mid-turn feed's result the counters said idle, so a drive op drained into the drained turn as text, a queued message lost its cancel, and a teardown ended the CLI with a message still in its queue. The settle marks a live hold settled and leaves a deferred reconnect armed until the drained turn's result. A subagent's frames at idle count nothing (above).
- **The turn-opener stamp reads the fed-turn twin.** `_forward` marks a turn the CLI opened itself (a stamped user atom: a background task's notification, a scheduled prompt, a peer's channel message) as `injected`, and the kernel's turn-finished push skips the buzz for such turns. That rule read `inflight == 0`, and the count above runs ahead of `_forward` on the same frame, so it never fired again: the opener kept the previous turn's value, the Stop hook stamped `human`, and the phone buzzed for every notification turn. It now reads `_inflight_texts`, what romp fed into the turn in flight: a stamped atom with the twin empty opens the CLI's own turn, one arriving with a fed text in the twin is a splice into that turn (which keeps its opener), and a drained mid-turn text rejoins the twin at the count, so the turn that drains it stays the person's.
- **Every exit of the CLI's queue has a release.** The drain's first frame; the splice record; the CLI's exit (`_release_hold_at_exit`, from `_run`'s finally: a held text whose record the final transcript does not hold goes back to the head of the queue for the next client, a landed one is left alone, an unreadable transcript flags the echo on a resumable conversation and re-heads when no conversation ever materialised); and the loop top, where a settled hold's text takes the stranded reconcile's two branches. A landing scan that raises is a logged fault (once per hold, on the problem ring) the hold escapes from at the next turn frame, at once when that frame is the result. The exit is the one event that loses the CLI's queue: the interrupt control request is not (the CLI keeps queued prompts across it and drains them into a new turn right after the interrupted result; a SIGINT makes it emit the interrupted result and exit without running them; a kill emits nothing), verified on the installed CLI over stream-json.
- **A resumable landing scan.** `_records_from_mark` and `_text_landed` take a `cursor`, so the per-frame take check reads only what the transcript grew by, never consumes a line still being written, and does not reopen a file that has not grown. Callers without a cursor are unchanged.
- **Why move code is in a queued-sends fix.** An accepted live move makes the CLI emit an init and a turn-less result with no query behind it; counted as a turn, that init left an idle session busy until its next real turn. So the count skips the init while a move's settle is expected, the move's result zeroes a count that fired anyway, and a real turn's result while the arm stands proves the arm stale and drops it (a standing stale arm would switch the count off for the session's life and re-open the fuse). The arm also holds the queue while the CLI relocates: a text fed into that window could run a whole turn before the move's turn-less result, whose late arrival the settle would take for a turn end. Every way the arm is lowered wakes the feeder (`_disarm_move_settle`, which tolerates a closed loop); `move()` arms under the session lock with its busy reading, so a text queued between the idle reading and the arm answers "busy" with no request out; its standing-down exits lower the arm before they drop the claim (`_stand_down_move`), since a second mover claims and arms the instant the claim is released; the lost-reply exit answers from the heal's own outcome (`_heal_cwd_pending` gains a `release` hook and returns the outcome; the boot path is unchanged) and announces the hold it leaves on the problem ring.
- **One persist at a time.** `_persist_queue` snapshots and writes under one lock (`_persist_lock`), closing a lost update between the feeder's post-pop persist and an enqueue's persist that the feeder tests exposed under load.

### Consequences

- **A behaviour change beside the fix.** A drained run of parked sends (a, b, c parked while the session compacted or sat under a usage hold) now reaches an SDK session one message each, in order, where the 2026-07-17 preference had the run merge into one turn. `SessionBackend.forwards_sends`, `SdkBackend.forwards_sends` and the send-park tests say so. The alternative, one prompt carrying one text block per send (a new backend contract), was designed and not built.
- **Latency.** A message sent mid-turn now waits for the last fed text's take (its splice record, or the drained turn's first frame) instead of the next tool boundary: two texts sent within one tool step reach the agent one step apart. The splice record bounds the wait.
- **A CLI that dies during a turn it opened itself is healed as a cut.** `_on_session_gone` reads `inflight > 0` for such a turn now, so the death takes the crash-heal road (one bounded resume with the crash notice at the head of the queue, `_heal_cut_session`) where before it settled `waiting` and the cut went unreported until the next kernel restart.
- **The card reads a CLI-opened turn off the counter.** `snapshot()` reports `working` from `inflight` with the clock the count set at the turn's first frame; before, it read the state log's `working` mark, written at the turn's first work atom.

### Overlap with #1827

#1827 (https://github.com/romp-on/romp/pull/1827) is on main; this stack's base predates it. It added `SdkSession.quiet()` and the switches' reconnect re-check, both reading `_cli_working` for a turn the CLI opened itself; this PR counts that same state in `inflight` from the turn's first frame, so a CLI-opened turn reads busy by either signal, and the two agree. `git merge-tree` of this head onto current main reports no conflict (the adjacent hunks in `_do_request_reconnect` and at the loop top merge cleanly). One sentence goes stale at the merge: `quiet()`'s docstring says `inflight` stays 0 while the CLI works, which stops being true with this PR. It lives in code this base does not carry, so it is a one-line follow-up after the merge rather than a change here.

`docs/guide.md` gains one sentence in the queued-message passage, after the recall paragraph: several messages sent during one turn reach the session one at a time, in order, the next shown as queued until the session has taken the one before it. `kernel/session_backend.py`'s `forwards_sends` docstring describes the same.

## Tests One further case of that module is held back while a load-sensitivity question of ours is settled, and is offered separately after that.

`tests/test_queued_sends_not_fused.py` (new, 36 cases in `OneFedTextAtATime`) runs the real `_amain` (its `inputs()` closure, `_on_message`, the loop top, `_run`'s finally) against a stand-in `claude_agent_sdk` module whose client records each write with the test's phase label and yields the frames the test pushes, plus an EOF sentinel for the CLI's exit and a gated control channel for the move cases. Each case is an executing behavioural test over the real code paths; the ones below are red before the change:

- `test_two_texts_sent_mid_turn_reach_the_client_one_at_a_time_and_in_order` fails before the change with `3 != 2`: the third text is written the moment it is enqueued, all three writes in the first turn's phase. After: two writes until the next turn's init, then the third, in queue order and into that turn.
- `test_a_reply_after_a_composer_message_is_fed_as_its_own_message`, `test_a_mid_turn_splice_releases_the_next_text_into_the_same_turn`, `test_an_interrupt_keeps_the_held_text_in_the_clis_queue_and_its_drain_releases_the_hold`: the same fuse from three angles (a reply, the splice accelerator, an interrupted turn).
- `test_a_text_fed_from_idle_is_taken_at_the_turns_first_frame_without_a_scan`: a text fed from idle is taken at the turn's init with no transcript read (the first turn's transcript does not exist yet); before the change both texts are written at once, so the wait for a single write times out. It also fails when the fresh rule is removed from `_untaken_taken`: the init's scan faults on the missing file and the second text waits for a later frame.
- `test_the_take_scan_starts_at_the_feed_time_mark_so_an_earlier_record_wearing_the_words_is_not_the_take`: a record wearing the fed words that was on disk before the feed, stamped ahead of the send-time floor, does not release the text behind; only the text's own record past the mark does. Red before the change (no hold), and red when the feed-time mark is not stored on the hold (the scan from the file's start takes the earlier record for the splice).
- `test_a_text_fed_into_the_drained_turn_is_mid_turn_and_waits_for_its_own_take`, `test_a_turn_the_cli_started_itself_is_counted_and_a_send_into_it_waits_for_its_take`, `test_only_turn_frames_witness_a_take_never_a_task_hook_or_progress_frame`: the CLI-owned-turn count and the turn-frame rule (`inflight` stays 0 at an init before the change; `_turn_frame` is absent).
- `test_a_subagents_frames_count_no_turn_and_release_no_hold`: a backgrounded Task's assistant and user frames at idle leave `inflight` 0, `busy()` False and the snapshot `waiting`, and one arriving after a feed from idle does not release the fed text's hold; the fed turn's init does. Red before the change (no hold), and red when `_turn_frame` accepts tagged frames (`1 != 0` at idle).
- `test_a_turn_the_cli_opens_itself_starts_the_working_clock_at_its_first_frame`: with stale `since` and stop readings left from a settled turn, the CLI's own init starts the clock and clears them, and `snapshot()` reads `working` with that clock; before the change the turn is never counted, and without the clock in the count the snapshot wears the old `since`.
- `test_a_turn_the_cli_opens_itself_is_stamped_injected_through_the_stream`: through the real stream, a stamped user atom while nothing is fed marks the turn `injected`; one spliced into the person's fed turn keeps `human`; one arriving in the drained turn of a mid-turn text keeps `human`. Before the change the CLI's own turn is never counted; with the count and the opener rule still judged on the counter, the middle assertion fails with `'human' != 'injected'` (the mutation the fix answers). The existing pin `tests/test_kernel_notify_popover.py::test_a_task_notification_streamed_while_idle_opens_an_injected_turn` calls `_forward` directly with `inflight` 0 and stays green either way; it is unchanged.
- `test_a_deferred_reconnect_waits_until_the_cli_has_taken_the_text_it_still_holds`, `test_a_reconnect_asked_for_while_the_cli_still_holds_a_text_is_deferred_not_immediate`, `test_busy_reads_a_settled_hold_as_in_flight_like_the_reconnect_gate`: the reconnect gate fires the deferral and builds a second client at the mid-turn feed's result before the change; `busy()` answers False in the settled gap.
- `test_a_teardown_that_finds_a_settled_hold_hands_its_text_to_the_stranded_reconcile`: the loop top's backstop (the flagged echo and the 'never reached its CLI' line).
- `test_the_clis_exit_puts_a_held_text_back_at_the_head_of_the_queue_for_the_next_client`, `test_the_clis_exit_leaves_alone_a_held_text_the_cli_did_take`, `test_the_clis_exit_flags_a_held_text_it_cannot_check_rather_than_re_feed_on_doubt`, `test_the_clis_exit_before_any_conversation_re_heads_the_first_text_rather_than_flag_it`, `test_the_clis_exit_still_flags_when_a_resumable_conversations_transcript_is_missing`: the exit's verdicts (no exit release exists before the change).
- `test_a_take_scan_that_raises_is_logged_once_and_the_hold_escapes_on_the_next_turn_frame`, `test_a_faulted_hold_escapes_at_the_result_when_nothing_later_streams`: the faulted scan.
- `test_the_scan_resumes_where_it_stopped`: `_text_landed` with a cursor (a `TypeError` before the change: no `cursor` parameter). The bytes behind the cursor are rewritten to carry the text after the first miss: the resumed scan answers False without re-reading them, the same call without a cursor answers True from the mark, so a scan that restarted at the mark fails here.
- `test_the_queue_mirror_never_loses_an_entry_to_an_interleaved_persist`: the reg's queue reads `['plain']` after a forced snapshot-A, snapshot-B, write-B, write-A; before the change the mirror ends empty.
- `test_the_moves_turn_less_result_zeroes_a_count_that_fired_anyway`, `test_a_stale_move_arm_is_dropped_at_a_real_turns_result_so_the_drain_still_counts`, `test_a_reconnect_drops_a_stale_move_arm_so_the_new_clients_turns_count`: the stale-arm drops (`_move_settle_expected` stays True before the change).
- `test_a_text_sent_while_the_cli_relocates_waits_for_the_moves_turn_less_result`, `test_a_refused_move_lowers_the_arm_and_the_held_text_is_fed_then`, `test_a_standing_down_exit_releases_the_claim_after_the_arm_so_a_second_movers_arm_survives` (four subtests), `test_a_lost_reply_after_the_cli_relocated_announces_the_hold_it_leaves`: the hold during a relocation, the refusal's release, the exit order with a second mover, the problem line. The relocation case is also the pin for the count's skip of the move's init (`inflight` 0 after the init while the arm stands).
- `test_a_move_that_never_happened_lowers_the_arm_through_the_heals_release_so_a_queued_text_is_fed`: the uncertain-outcome exit with no second mover: the heal's release lowers the arm and the text queued during the hung move is fed when `move()` returns. Before the change the text is fed during the move (no hold on the arm); with `_heal_cwd_pending` ignoring its `release` hook the arm stands and the text is never fed. `tests/test_session_move.py::MoveRefusals::test_a_control_error_leaves_the_session_where_it_was` pins the same hook from the control-error side.
- `test_an_accepted_live_move_leaves_an_idle_session_idle` asserts `inflight` 0 after the move's init, before its result, so the count's skip while the arm stands is pinned there as well as in the relocation case (the move's result zeroes a stray count, which masked the skip when the case looked only after it). Green on the base, where nothing counts; red when the count ignores the arm.
- Green both before and after, as fixtures and controls: `test_the_first_turns_record_lands_under_the_clis_sid_however_late_the_reg_write` (the helper's own race), `test_a_move_whose_cli_exited_after_the_claim_names_the_error_and_drops_the_claim` (the bare flag clear could not raise before the change), `test_teardown_opens_a_replaced_control_channels_gate_so_its_worker_does_not_outlive_the_test` (the fixture's own hygiene).

Pins re-aimed in existing modules, each red before the change: `tests/test_sdk_rewind.py::WiringPins::test_reconnect_never_defers_on_rewind_held_turns` (the gate's two-line quiet shape); `tests/test_session_move.py::SpuriousSettle::test_the_guard_only_fires_on_an_armed_zero_turn_result` (a real turn's result drops a stale arm, logged); `tests/test_session_move.py::MoveRefusals::test_a_text_queued_after_the_idle_reading_is_busy_before_any_request` (a send landing inside the claim makes `move()` answer "busy" with no request out); `tests/test_session_move.py::LostReply::test_a_lost_reply_after_the_cli_moved_stands` (the lost-reply line is on the problem ring). Green both ways: `tests/test_sdk_rewind.py::DeleteWhileBusy::test_busy_answers_for_the_double_with_its_three_attributes` (`busy()` on the module's double, guarding the duck-typed call), and the wording in `tests/test_kernel_send_park.py` (`test_sdk_delivers_a_run_as_separate_sends_one_message_each` and two assertion messages) and `tests/test_sdk_backend.py::LiveTail::test_forwards_sends_is_true_for_the_sdk`. `tests/permission_ask_feeder_drive.py` stands in for the take by hand before its later enqueues (it pushes its stream through `be._forward` and never reaches `_on_message`); without those two lines `tests/test_permission_ask_state_survives_the_stream.py` goes red on `control_pop_marks_working`.

Two fixtures modelled a CLI that takes a fed text without ever streaming a frame for it, which the CLI does not do (its init opens a turn, and a fed text streams a turn frame or lands a record); each is brought to the CLI's own shape rather than the hold relaxed for it:

- `tests/test_host_transport.py`: `FakeHost` gains `echo_turns`, a CLI that answers each fed user text with a one-step turn (its own user record back, then the turn's result), and `AttachStandDown::test_four_unanswered_attaches_stand_the_session_down_once_until_a_send` serves with it. That case expects the three messages held across the stand-down to reach the host in order; with the hold, the second and third wait for the first's take, which a host that echoes nothing never signals, so before this fixture change the case timed out with one message fed. This file is the one change beyond the queued-sends surface itself, and it is a fixture, not an assertion change.
- `tests/test_sdk_backend.py::ReconnectReconcilesInflight`: the stall client streams its init with the first dequeued turn instead of at connect (the CLI's init opens a turn; none streams on a turn-less connect). The connect-time init already failed the case at the base under the open SDK gate, on its last read (the state log never reached `waiting`); with the count it failed on `inflight` held at 1 for the stall's whole life; the fixture fixes both.

One case fails under the full suite on this head and on the base alike, for a reason outside this change: `tests/test_sdk_backend.py::OptionsAssembly::test_ultracode_is_a_choice_everywhere_but_never_the_seeded_default` (`sdk-defaults` reads `effort: ultracode`). It is gated on the SDK being importable; `tests/test_host_transport.py`'s import block puts the SDK venv on `sys.path`, so every module the same xdist worker collects after it sees the SDK, the gate opens, and the case runs and fails; run alone it skips. Reproduced on the base tree by running the same six modules together. It stays in the red count below and is listed here for the record.

Red run (the new module's first 31 cases and the re-aimed pins on the unchanged kernel): 34 failed, 135 passed; the six cases added since are red on the base as described above. Targeted green run (the new module plus every module that pins the touched code or reads `_forward` and `_on_message` by source: `test_sdk_rewind`, `test_session_move`, `test_kernel_send_park`, `test_permission_ask_state_survives_the_stream`, `test_sdk_backend`, `test_sdk_stream_survives_handler_errors`, `test_sdk_rename_ping`, `test_deploy_drain`, `test_state_isolation_order`, `test_queued_copy_identity`, `test_queued_copy_press_id`, `test_injected_voice`, `test_kernel_notify_popover`, `test_retry_gaveup`, `test_host_transport`, `test_api_health`, `test_comment_threads`, `test_spend_rebill`, `test_injected_origin` and the turns.jsonl row tests, which read `_on_message` by source): 1040 passed, 37 skipped, 257 subtests passed.

Full suite on the final commit (the full pytest under xdist with 10 workers, a short `/tmp` TMPDIR, the extension's `node_modules` installed so the served legs ran): 4 failed, 14357 passed, 11 skipped, 2339 subtests passed in 4 min 53 s, no worker lost. Every test of the new module and of the re-aimed modules passed under the suite. None of the four reads what this change touches, and each was rerun alone on this commit:

- `tests/test_sdk_backend.py::OptionsAssembly::test_ultracode_is_a_choice_everywhere_but_never_the_seeded_default`: the SDK-gated case described above (it skips alone: the test interpreter has no `claude_agent_sdk`; under the suite a sibling module's import block makes the SDK importable on the same worker, and the case fails on the base tree the same way).
- `tests/test_judges_process.py::ChildRoad::test_one_pass_through_the_child_feeds_perf_and_keeps_the_kernels_bookkeeping_in_order` (a CPU-time compare) and `tests/test_kernel_remote_identity.py::RemoteIdentity::test_second_alias_absorbs_the_old_row` (an empty bus list): both pass alone on this commit (0.22 s and 0.19 s) and fail under load in most full runs of unrelated changes on this machine.
- `tests/test_session_host_restart.py::ServedRestart::test_with_hosts_on_the_turn_survives_the_restart_and_with_hosts_off_it_is_cut`: a real session host exited before serving its socket under load; passes alone on this commit (22 s).

Not called green: reported as the run produced it.

The run before this one, on the same tree with two comment words changed since, had 5 failures as well: the first three above, the judges figure replaced by `tests/test_kernel_remote_identity.py::RemoteIdentity::test_second_alias_absorbs_the_old_row` and the postal case by `tests/test_host_transport.py::AttachStandDown::test_the_retry_counter_counts_consecutive_incomplete_attaches_only` (a host attach that did not complete under load; it passes alone three times and with its class on this head, and fails in other full runs on this machine without this change). Two earlier runs, the first before the two fixture changes above and the second after them, had 6 and 3 failures: the SDK-gated case and the session-host restart case both times, the two fixture cases (`AttachStandDown::test_four_unanswered_attaches_stand_the_session_down_once_until_a_send` and `ReconnectReconcilesInflight`) and two load-sensitive cases that passed alone (the judges figure and the remote-identity case) in the first, and the awaiting-stamp case in the second.

Tier: fix


lczh:
Evidence executed against 07d65589. One item is a silent regression on the default hosted road and is worth fixing before this merges; the other two are low.

## At this head

**A kernel restart under session hosts feeds a held text a second time.** `SdkSession._release_hold_at_exit` (kernel/sdk_backend.py:5945) treats the session thread's exit as the CLI's exit: on a scan that finds no record it puts the held text back at the head of `_pending` and persists it (5984-5990). Under a session host, the default configuration, that premise fails. `SdkBackend.drain` (11440) latches `s.detached = True` and `s._host.detach_mode = True` (11468-11470) before `shutdown()` (11475); `HostTransport.close` (kernel/host_transport.py:318) sends `detach`, not `end`, and the host's `_detach` (kernel/session_host.py:873) leaves the CLI and its stdin queue running. The loop breaks on `ended` (6969) without passing the loop top, the only other place that clears `_untaken` (6779), so `_run`'s finally (6628) runs the release with the hold set, and the function checks only `u is None` (5975): the one exit path that does not read `detached`, where `_on_session_gone` (16236) and the drain's reaper (11488) do. The next kernel seeds `_pending` from `reg["queue"]` (5449), re-attaches to the same CLI, and `inputs()` feeds the copy.

Executed over the real connect loop with a fake host and the SDK present: the CLI received "first" and "mid-turn note"; `drain` latched `detached` and the transport sent `detach` and no `end`; `_release_hold_at_exit` still logged "back at the head of the queue" and the reg queue read `["mid-turn note"]`; a second backend seeded `_pending` from it and fed it, and the CLI had received `["first", "mid-turn note", "mid-turn note"]`. A hold fed from idle duplicated the same way. The base at 2378fa77 has no exit release, so this is a regression on the hosted road, and a silent one: the re-head line carries no problem flag, and the agent gets the message twice, fused into one record, the shape this PR exists to prevent. Fix: return early from `_release_hold_at_exit` when `self.detached` is set, leaving the text to the re-attach road where held sends stand, with a dated comment (a detached CLI keeps its queue, so the thread's exit is not the queue's loss event); add a case to tests/test_queued_sends_not_fused.py that feeds a text, leaves it untaken, sets `detached` as `drain` does, shuts down, and asserts the reg queue does not gain the text (the module's exit cases all model a CLI that died); and name the detach exception in the docstring and the body.

**The `fresh` rule still fuses two texts when the CLI opens its own turn at the instant of a feed from idle.** `_untaken_taken` (7741) returns True on any turn frame while the hold is `fresh` (7769), and `_on_message`'s count block is gated on `inflight == 0` (7783), which the pop already raised (6711). When the CLI's own init is on the wire as A pops at idle, it arrives with `inflight` 1, the count is skipped, A's hold clears with A still in the CLI's queue, B is fed mid-turn, and the CLI drains both at that turn's end. Through the PR's own harness, a foreign init after A's pop released the hold and B was written at once with no record of A; in the other order B was held. With replay-user-messages off the frame says nothing about whose turn it opened, so there is no event to key on: a residual to state, not a defect to fix. docs/guide.md:85-86 says "two messages are never joined into one". Soften it to what holds (the next waits until the session has taken the one before it), add the residual to Consequences, and note it in `_untaken_taken`'s docstring.

**A nudge drained into its own turn keeps the previous turn's opener, so the phone buzzes for it.** `_note_turn_opener` (7212) writes only when `fresh` or the opener is `human` (7220), and `inputs()` passes `fresh` False for a mid-turn feed (6737), so an injected text fed mid-turn writes nothing; the settle clears the twin but not the opener (8311); when the CLI drains it into its own turn the count block (7805-7812) puts the item back into `_inflight_texts` without noting an opener, and `_forward`'s stamp (15744-15745) sees a non-empty twin and stands aside. On the module's harness, a human turn, a marked text fed mid-turn, the result, then the drain's init left `_turn_opener` at `human`; `_stop_hook` stamps `lastTurnOpener` from it (8706) and `_turn_notify_tick` (kernel/kernel.py:55820) buzzes for a turn romp itself started. Not a regression, but the count block has the item in hand: call `self._note_turn_opener(fed_text_opener(item), True)` where it rejoins the twin (7812), and add a case (an injected text fed mid-turn, the result, the drain's init, the opener reads `injected`); the existing case at 731 drains only a human text.

## Smaller points

- kernel/session_backend.py:174 says `inputs()` holds the next text "since 2026-09-08", and tests/test_kernel_send_park.py:253-254 says that hand-off "superseded" the one-turn merge as of that date. Both commits are authored 2026-09-19 and `_untaken` is absent at 2378fa77, so the mechanism is dated eleven days before it exists. Of the eight added lines dated 2026-09-08, those naming the sighting and the CLI verification can stay; date the "since" clause and "superseded" at the landing.
- tests/test_sdk_backend.py: `ReconnectReconcilesInflight`'s `StallClient` now streams its init with the first dequeued turn (4492), but `PendingQueueLoop`'s `GatedClient` (4315), `InterruptWithQueue`'s `StallClient` (4401) and `InterruptSettlesStall`'s (4098) still yield it at connect, the shape the body says the CLI never produces. An init counted before A's pop makes A non-fresh, and the fake streams nothing until release, so `test_second_turn_forwarded_immediately_mid_flight` (4345) fails; forced by one scheduling delay at the first enqueue it did, `received=['A']`. The natural ordering passed 12 of 12 runs, and CI's SDK gate is closed. Bring all three to the first-dequeued-turn shape.

The modules run, the probes and their controls are in a second comment below.

lczh:
The test runs behind the comment above, on 07d65589 (merge-base with main 2378fa77).

### Tests

- tests/test_queued_sends_not_fused.py alone in a worktree at 07d65589: 36 passed, 4 subtests passed, 12.9 s, the body's count for `OneFedTextAtATime`. Its five exit cases (1050, 1098, 1116, 1155, 1200) each push the stream's EOF sentinel, a CLI that died; none sets `detached`.
- tests/test_sdk_backend.py's SDK-gated classes (`PendingQueueLoop`, `InterruptWithQueue`, `InterruptSettlesStall`, `ReconnectReconcilesInflight`) with the SDK gate opened the way tests/test_host_transport.py opens it: 4 of 4 passed in one run, and `test_second_turn_forwarded_immediately_mid_flight` passed 12 of 12 natural runs. CI has no SDK venv, so these run only on a machine with one.

### Probes and controls

- The duplicate: a fake host over the real connect loop with the SDK present. "first" fed and taken, "mid-turn note" fed during the turn and held; `SdkBackend.drain` latched `detached`, the transport sent `detach` and no `end`, no cut turns; `_release_hold_at_exit` logged the re-head line and the reg queue read `["mid-turn note"]`; a second `SdkBackend` over the same state directory seeded `_pending` from it and fed it; the CLI had received `["first", "mid-turn note", "mid-turn note"]`. A text fed from idle with `detached` latched before its init duplicated the same way. `git grep` for `_untaken` at 2378fa77 finds nothing.
- The `fresh` residual: the module's harness driving the real `_amain`. A popped at `inflight` 0; a foreign turn's first frame (an init in one run, a stamped task-notification user record in another) reached `_on_message`; `_untaken_taken` returned True on `fresh` and B was written with no record of A and no result for the turn. Control: the same two events in the other order held B until the fed turn's init.
- The opener: the module's harness; a human turn, an injected-marker text fed mid-turn, the result, the drain's init: `_turn_opener` read `human` through the drained turn's assistant frame. Controls: the same text fed from idle read `injected`; one drained after a CLI-opened turn read `injected`, so the stamp is inherited, not derived.
- The fixture race: `PendingQueueLoop`'s exact `GatedClient` with the init-first ordering forced by one scheduling delay at the first enqueue: the connect-time init counted, A popped non-fresh, B sat in `pending_queued` for the full 6 s, and the case failed with `received=['A']`; the base tree and a `GatedClient` yielding its init with the first dequeued turn both forwarded B at once. The enqueue lands 1-3 ms after the thread start, the connect preamble 160-930 ms later.
- The dates: `git log` gives 2026-09-19 for both commits; the diff adds eight lines dated 2026-09-08 and none later.

henrypinkard (approved):
Approved on the maintainer's decision of 2026-09-20 (one message each for a drained pile accepted), with the exit release standing down for a CLI a host keeps, and the kernel docstrings brought in line, carried onto the branch by the review session.

henrypinkard (approved):
Re-approved on the current head (the earlier approval landed on the pre-push head).

lczh:
Round two, executed against f1d4a5c0.

The round-one high item landed: `_release_hold_at_exit` stands down before its landing scan when the session is detached or the sid's lease reads 'attach', three new cases run, and the drain road holds when driven through `SdkBackend.drain` itself. The four low items are unchanged. Two sentences in the new docstring do not match the code; one of them is where the re-attach road's residual belongs.

## Landed

**The duplicate under session hosts.** kernel/sdk_backend.py:6093 clears `_untaken`, 6094 reads `self.detached or self.backend._lease_survives(self.sid)`, and the scan at 6099 runs only past it; `_run`'s finally calls it at 7072, before `_on_session_gone` at 7077. The three roads:

- The drain. `SdkBackend.drain` latches `detached` for every session with a host or a host intent (12110-12113) before `shutdown()` (12117, which sets `ended`, 5978-5987); the loop breaks at 7431 without the loop top's clear (7230). Executed through `drain()` on the module's harness (`_host_intent` set, a settled hold): stopped 1, reaped 0, no cut turns; `detached` True, the reg queue still `["behind the note"]`, the echo unflagged, the new line logged once. The lease reading covers the gap between an iteration's teardown finally (7413-7430) and the loop top: a kernel child drops its own lease at 7419-7420 before this runs, a host's is left (7421-7425). `host_lease_state` (kernel/host_transport.py:163-172) answers 'attach' only for a lease whose holder kind is `host` and which `lease_state` (kernel/sdk_backend.py:4033-4059) calls valid: CLI and holder each alive with a matching start time, beat within 12 s (`LEASE_TTL_S`, 3947). The attach stand-down sets `detached` at 7461 and reaches the finally with no hold. Reading the code, the only exit where a host keeps the CLI and neither reading fires is the stale beat below.

- The crash. A host that dies with its CLI leaves a lease whose pids are gone: `lease_state` answers 'no-live-process' or 'holder-gone', `_lease_survives` False, the plain re-head runs as before. A host that dies while its CLI outlives it briefly reads 'holder-gone' too, and `_host_orphan_recover` waits for that CLI to exit (11193) before spawning, so a duplicate needs the orphan to drain the text first, as before this PR. A live host whose beat is older than 12 s reads 'orphan' too, so a thread exit at that moment re-heads with the CLI still holding the text. Reading the code; no probe.

- The re-attach. The next kernel does not learn of the held text. The hello carries `inflight`, `parked`, `journal.next` and `exited` (kernel/session_host.py:1254-1262), never the CLI's stdin queue (`_queue_in`, 1298-1313); `__init__` seeds `_pending` from the reg queue (5544) and sets `_untaken` None (5570; "RUNTIME-ONLY" at 5564). Executed: a second backend over the same state with a live host lease written, a second session from the reg: `pending()` read `["behind the note"]`, `_untaken` None at construction, "behind the note" fed 0.000 s after the connect, the only hold afterward the new kernel's own on that text, and the reseeded echo for "mid-turn note" not dropped (14105 skips the marking when the lease survives). The person's view holds: the held text shows as sent and in flight, as before the restart. The residual: while the CLI still holds the fed text (a restart during the turn it went into), the text queued behind it joins the CLI's queue at once and the two share the drain, the shape this PR fixes, bounded to that window. Without a surviving lease the second boot instead re-delivers the held text to the queue's tail (14254; executed: `["behind the note", "mid-turn note"]`), which under a wedged host that later recovers is the duplicate again.

## Still open

Unchanged at this head:
- (2) `_untaken_taken` returns True on any turn frame while the hold is `fresh` (8280); docs/guide.md:88-90 still says "so two messages are never joined into one".
- (3) The count block puts a settled hold's item back into the twin at 8323 without noting an opener; `_note_turn_opener` (7723-7734) writes only for `fresh` or `human`.
- (4) kernel/session_backend.py:174 still dates the hold "since 2026-09-08" (the new clause at 175-177 puts 2026-09-20 beside it), and tests/test_kernel_send_park.py:253-254 still says that date "superseded" the one-turn merge; the commits that build the hold are dated 2026-09-19 and 2026-09-20.
- (5) `PendingQueueLoop`'s `GatedClient` (tests/test_sdk_backend.py:4314-4317), `InterruptWithQueue`'s `StallClient` (4400-4401) and `InterruptSettlesStall`'s (4097-4098) still yield the init at connect; only `ReconnectReconcilesInflight` (4484-4492) has the first-dequeued-turn shape.

## New at this head

- The docstring at 6082 and the commit message say `drain_and_reap` latches `detached`. No such name exists in the tree (`git grep` finds that line and the test comment at tests/test_queued_sends_not_fused.py:1248); the latch is `SdkBackend.drain`, 12110-12113.
- 6088 says "the next kernel's attach reads the hold from the host's replay". Nothing rebuilds a hold at the attach; the replay carries the CLI's output records, where the held text appears only once drained. State the residual here instead; to close it, the reseeded echo carries `off` and `fsid` (14069 onward), enough to seed `_untaken` at the attach with `fresh` False and let `_text_landed` witness the take.
- The detached case (1234-1258) sets `detached` by hand at 1248 and pushes the EOF; no case drives `drain()`, and the suite has no end-to-end drain with a held text (tests/test_host_transport.py:1147 pins the `detach_mode` line by source; tests/test_session_host_restart.py:195 restarts a real host mid-turn with nothing queued). The lease case (1260-1282) writes this process's pid as both CLI and host, a live host with a live CLI; the predicate case (1284-1308) runs the plain exit with no lease file; the dead-host reading is pinned only one level down (tests/test_host_transport.py:178-180). Nothing shows the re-head with a dead host's lease on disk, the crash road the commit calls unchanged.
- The log line (6095-6097) is fine on the voice rules; "the queue the host kept" claims more than the attach does. Minor.

## Runs

- tests/test_queued_sends_not_fused.py alone at f1d4a5c0, serialized behind the machine's test lock: 39 passed, 4 subtests passed, 13.5 s (36 at the first head).
- Three probes on the module's own harness (fake SDK client, hosts-off file: no host transport, no real detach frame; the round-one fake-host harness was not rebuilt): the drain road through `SdkBackend.drain`, the re-attach with a live host lease written, the same boot without a lease. 3 passed; their readings are quoted above.

PR #1890, opened 2026-09-19
The chat pane's approval box lists a session's held messages with Approve and Deny above the background box, and a held message rings the tab until decided

henrypinkard:
The chat pane's approval box and the ask ring for a held message: the second code phase of `plans/notice-cards.md`, "Action kinds and the held-mail card" (design PR 1875; the kind API, the held-mail card and the removal landed as PR 1885). Feature tier: a surface inside the notice model, no new contract.

## What the user sees

- Above the background-agents box at the bottom of the chat, a box lists this session's held messages: the sender line as its title (`New message from <sender>`), the message text, and **Approve** and **Deny**. It sits nearest the composer, above the agents, so a decision only the user can make is where the eye rests.
- Approve delivers; Deny first opens the optional note back to the sender inline, with `Deny & send note`, `Deny without note` and a way back. Both are the same notice actions the feed card carries, over the chat socket; every button of the row latches on a click and lets go on the kernel's answer. A refusal says why in the row; a success drops the row at once.
- A held message puts the session in the blocked state: its tab wears the dashed yellow ask ring until the decision, and the ring leaves with the row.

## Design points

- **The rows are the kernel's.** `build_session` ships `status.notices`: this session's standing needs-you notices that carry actions (`_chat_notices`: the feed card's projection with the cleared ledger applied, each action with its kind; today a held peer message's Approve and Deny). On the STATUS, so a status-only delta carries a decision's disappearance the way the ring's `needsYou` travels. The chat signature carries the rows' ids as one more component, labelled `notices` in `_CHAT_SIG_LABELS` (right after `needs`, the builder's order) and classified in the signature census (`tests/test_chat_build_sig_inputs.py`), so a hold posted or a decision taken brings a frame forward and a miss is attributed to it by name.
- **One wire.** The box posts `noticeAction {itemId, sid, kind, body, input?}` like the feed card and handles `noticeActionDone` by the row's id. Nothing new in the kernel's op table.
- **The ring needs no new code.** The chat is a feed audience, so the feed builds with the feed pane closed; the held-mail card files under needs-you and `status.needsYou` turns the ring on. The lab pins it end to end.
- **The box renders through the background box's road.** `awaitKey` carries the rows, so every status-carrying handler repaints it through `awaitChanged`; `#notices` sits between `#content` and `#bg-tasks` in the chat page's markup; the chrome is the background box's with the ask ring's yellow edge, dense-chrome aware.

## Tests

- `tests/test_chat_notices.py`: the slice (needs-you with actions only; an older row's route read as its kind; a decision, a Clear and an expiry drop the row; the owner-less run and a fault give an empty box), the source pins on the status field, the signature line and its label's position, and the markup order. Every write keyed by a private synthetic sid and removed in tearDown.
- `ui/webview/chat-notices.test.ts`: the renderer's shape (the row keyed by id, the message text as the body, the kind on the wire, the deny step's three buttons, the whole-row latch, no Edit), the handler, the CSS.
- `tests/test_held_mail_chat_served.py` (served lab, feed pane closed): a hold before boot shows the row above the background box and the tab's ring; Approve is refused with no bus and re-arms with the reason in the row; Deny's inline note step, Back, the bare deny; the decision recorded as the runner records it drops the row and the ring.

## Red-first (the head's tests copied into a detached worktree at the first PR's head b101376c, run there)

- `ui/webview/chat-notices.test.ts`: all 4 red. `tests/test_chat_notices.py`: all 5 red (the helper, the status field, the signature and the markup absent). `tests/test_held_mail_chat_served.py`: all 4 red (no box on the page).

## Counts

- 15 files changed, 803 insertions(+), 40 deletions(-) against main after the carry. Production files: `kernel/kernel.py`, `ui/webview/render.ts`, `ui/webview/feed.ts`, `ui/webview/federation.ts`, `ui/webview/board-def.ts`, `ui/webview/styles.css`, `ui/webview/notice-face.ts` (new, the shared face); three test files added, the signature census, two chat harnesses' stub lists, the bottom-box and federation pins widened.

## Verification

- Round one's head (e4b82fff, before its carry): the whole extension suite (5071 tests, 0 failures) and the whole Python suite in three capped shards, green but for `test_kernel_judge_model::JudgeSettings::test_judge_run_applies_tier_effort`, the shard-order flake seen on 1885 (passes alone).
- Round two's head as pushed (main carried through 892d7c0a): tsc, the kernel modules (`test_chat_notices`, the signature census and label tests, `test_kernel_bg_tasks`, with `test_rename_route` behind them), the pane pin files, the three served labs that read the notice face (the chat box's, the feed's held-mail card, the notice cards lab) and the whole extension suite. CI is the whole-suite gate on the pushed head.

## Rounds

- Round two (the manager's review at ad96e193): two mediums, five lows.
  - **Medium 1.** The box replaced every row on every status change, so a second hold landing while Approve was pressed destroyed the pressed button before its mouseup (no click, nothing posted, nothing latched), and the next status change wiped a row's refusal line. Now the rows are keyed by the notice id and reconciled in place (title, body and attachment updated when they change; the buttons and the refusal line rebuilt only when the row's own actions change; a row that left leaves; a row moves only when out of order), and the clicks ride one delegate on the stable `#notices` container, reading the notice from the active frame at click time. The lab presses Approve, lands a second hold, checks the pressed button is the same element, releases: latched, posted, refused, re-armed; a third hold then lands and the first row's refusal line stays.
  - **Medium 2.** A remote host's `noticeActionDone` arrived host-prefixed while the row's id was bare. `prefixInbound` prefixes `status.notices[].itemId` like the feed's notice cards (a copy, the local host the identity); pinned with a federated session frame beside the answer.
  - **Low a.** `#notices` joined the bottom-box observer list; the lab's transcript is sixty turns long, and the at-bottom reader stays at the bottom when the box appears, when rows are added and when the note step opens.
  - **Low b.** Main carried; the `Status` interface keeps both `notices` and main's `modelFallback`.
  - **Low c.** The unit test restores the shared files it appends to and unlinks only its own sid's files.
  - **Low d.** The slice is `None` before the first feed build since start, as `needsYou` is; the signature reads it as empty. Pinned.
  - **Low e.** `noticeBodyNodes` and the attachment renderer moved to `ui/webview/notice-face.ts`, one module the feed card, its modal and the chat box import; the row carries the stored attachment and shows the pinned picture or the file's name as the card does.
  - Red-first for the round (the head's tests copied into a detached worktree at ad96e193): the box-below observer pin and the federation prefix pin red; the chat box's pin file cannot load there (notice-face.ts does not exist at that head); the three new unit tests (the attachment on the row, the first-feed-build gate, the signature's empty read) red; all four lab tests red (the reader off the bottom when the box appears, the pressed button replaced by the frame, the note step's growth, the rows after the decision).

## Sibling and main

- Built on PR 1885 (merged e2fffe99) and carried through it: the branch's diff against main is the chat box alone. No sibling shares these hunks.

## Out of this PR

- The settings plan's kind; a Clear from the chat box (the feed card keeps Clear); the feed card's prompt and the chat's inline note are the two faces of the one optional input.

Synthetic fixtures only (the notes-api world, `TESTHOST`, placeholder ids); no personal identifiers; no em-dashes in added lines; the ruled-out word appears in no added line.


ksarma:
A note after the merge: evidence executed against 5dd220073, the merged head; lines on 5dd220073.

The one mechanism that loses mail: `_chat_notices` (kernel/kernel.py:24406) reads only the notice projection, never the quarantine directory, so the box inherits the silent skip at 24494 of a held file whose id fails the 64-character key pattern at 24023. A relayed id (`px-<epoch>.<pid>_<32 hex>.<host>`) runs 48 characters plus pid digits plus host name, crossing 64 at a 10-character sender host under a 7-digit pid. Executed: bus-shaped ids of 65 to 83 characters, a shape no test uses, gave no row, ring or stderr line; a 36-character control showed. The body's sentence that a held message blocks the session until decided is false for them. The skip is #1885's; this PR adds the surface. The fix is in the note on #1885: the mid as the key when it fits, else `held-` plus 40 hex characters of its sha256, that key in the post, idempotence checks and re-home rows, the real mid in the action body, a stderr line for the continue. Patched, an Approve with a 67-character mid reached the bus call: `_notice_action_run` reads the mid from the body, never the key. Widening the pattern to 128 changes a pinned grammar (tests/test_notice_card_store.py:104).

- kernel/kernel.py:24424: the bare except silently returns an empty list, so one malformed row (a non-int field or a non-dict action) blanks the whole slice and hides the box while the tab keeps ringing; only a bypassing writer produces one. The feed's builder raises on the same store, older than the PR. Fix: say the fault once per sid and drop only the bad row, by a shape predicate shared by every reader.
- ui/webview/render.ts:18836: the handler removes the row on every ok answer, and the row carries no `dismissOnAction`, so a needs-you action notice posted without it vanishes on a delivered click and returns re-armed at the next box repaint, with nothing against a second delivery (executed). Only the `/notice` door posts one, the flag defaulting false. Fix: ship the flag as `_notice_cards` does at 24588 and remove the row only when ok and dismissing.
- ui/webview/render.ts:14674: a null `status.notices` is read as empty, so the first frame after a restart tears the box down and the next, 40 to 50 ms later, rebuilds every row, losing a deny note in progress. Fix (executed): treat null as unknown, leaving rows the host shows for this session; clear on an explicit empty list or a session change. The delegate reads the same slice; latch it in the frame handler for both.
- ui/webview/render.ts:13845: the scroll listener calls followReader before classifyScroll, so a write's own echo re-derives `stick` from geometry a box below has changed. At boot with a held message the box appears between the land-bottom write and its echo, which records `stick` false; the observer at 13959 stands down and the box covers the transcript's last 198 px. Executed: 4 of 20 lab runs red on the three at-bottom assertions, 3 of 3 with the frame released before a scroll echo was dispatched. Seeding `lastH` (already 0) changes nothing; a follow on show fixes only this case. The background box shares the loop; the race predates the PR. Fix: an echo keeps the writer's follow mode, or the observer judges from pre-growth geometry.
- ui/webview/styles.css:1068: the body is clamped to four lines with no way to read the rest, and the comment at 1061 promises a hover nothing implements. Fix: a click on `.ntc-body` through the existing delegate lifting the clamp, the open state keyed per item id, no toggle inside a link. A title alone gives no tooltip on touch.
- kernel/kernel.py:24412: with Task tracking off the ring's input is empty while `_chat_notices`, gated on None alone, still lists rows, so the box shows without the ring. A held message written while off reaches no surface until the switch returns; the backfill runs only inside `build_feed`. Fix: return an empty list when `_feed_needs_input_of(sid)` is False (4 of 7 unit fixtures change), or run the backfill outside the feed build; either way with a tracking-off test.

## Smaller points

- kernel/kernel.py:24417: actions-only, where plans/notice-cards.md says every standing needs-you notice; an actionless one lit the ring and hid the box. Amend the plan; the unit test already pins the exclusion.
- ui/webview/styles.css:1063: the left edge uses `var(--st-working-bg)`. The ring is drawn with `var(--st-ask-bg)`, a different token in both themes; use it here.
- ui/webview/styles.css:1072: the Approve border hardcodes the dark accent's RGB against accent-coloured ink, orange text in a blue border in the light theme. `color-mix(in srgb, var(--accent) 60%, transparent)` keeps the alpha.
- ui/webview/styles.css:1070: Approve and Deny stay 26 px on a coarse pointer where the composer's buttons and the chips grow to 32 to 40 px; a 32 px minimum fits one row at 390 px. Other in-flow buttons never grow either.
- ui/webview/styles.css:1074: `#e5484d` four times where the blocked-state token carries the same value; the feed's Deny face at feed.css:1110 is bare too, so change both or neither.
- ui/webview/styles.css:1062: no bottom margin, so it sits flush on the footer when the background box is hidden, which keeps 6 px. A bare `#notices + #bg-tasks` rule matches while hidden.
- ui/webview/render.ts:4: the sanitizer comment trails the notice-face import; move it back.

Tests follow in the next comment.

ksarma:
Tests, continued from the note above; lines on 5dd220073.

- tests/test_held_mail_chat_served.py:36 and tests/test_chat_notices.py: every id is 36 characters or a two-letter key; the 64-character skip has no red test. Add a held file with a bus-shaped id of 65 or more characters and assert its row, not the backfill's count, with the lab's row selector built from the derived key.
- tests/test_held_mail_chat_served.py:241, 256, 273: the three at-bottom assertions rest on a precondition the test never establishes. Open with no held message, settle at the bottom and assert it, then write the file; keep the pre-boot held message for the row and ring legs, and pin the frame ordering.
- tests/test_held_mail_chat_served.py:121: the decision is a hand-written expire row; no test executes the pane's ok branch; tests/test_kernel_trust.py already executes the runner's bookkeeping and the bus body, and the next status frame masks a neutered branch 43 ms later. A stub bus cannot answer ok from boot: arm it for the final step, and read the row inside the `noticeActionDone` dispatch, not after it vanishes.
- tests/test_held_mail_chat_served.py:114: the deny latch is read 1.1 to 1.5 ms after the click while the refusal re-arms at 1.5 to 2.2 ms (1 of 25 runs red at 275); return the flags from the evaluate that clicks. The Approve read at 95 follows a real mouse release: snapshot from a once-only click listener on `document`.
- tests/test_held_mail_chat_served.py:69: the driver writes the held file in place where the bus renames a `.tmp`; the backfill memoizes the directory's stat after an empty read, which a content write does not move, so a run's last held message can go unposted and one leg reads a row short. Rename in the driver only.
- tests/test_chat_notices.py:88, 151, 153: text pins on the signature component and the status field, both executed by the served lab. An optional Differential case asserting `moved(a, sig()) == ("notices",)` needs the session in the needs-input set first.
- ui/webview/box-below.test.ts:31: the title names two boxes where the regex enforces three, as does the header comment at line 8.
- Pins that move: ui/webview/chat-notices.test.ts:82, 84 and 85, the null-as-empty lines of renderNotices, the `ChatNotice` regex, the comment at styles.css:4568. Cases owed: a non-dismissing notice surviving a stubbed ok, a null frame keeping a row and note, a bad row leaving healthy rows, a tracking-off case, a long body growing on click.

lczh:
A flake report on this lab: tests/test_held_mail_chat_served.py HeldMailChatServed.test_deny_opens_the_note_inline_with_two_choices_and_a_way_back_and_the_bare_deny_posts_the_verdict fails intermittently on 'latched on the decision' with all three buttons still enabled (Deny & send note, Deny without note, Back). Seen today on PR CI for two unrelated branches (#1880 at 8052c459, #1896 at 06fba3b2; neither touches this lab or the dialog) and once locally against main 09c853ed (the module 1 failed / 4 passed, then the single test alone passed; the same module 5 passed on the other branch). Looks like a race between the deny click and the latch read.

PR #1891, opened 2026-09-19
A Codex session refuses only the slash commands the kernel knows it cannot take; a message that merely begins with a slash reaches the model as text

ksarma:
## Symptom

Typing a message that begins with a slash-word or a path at a Codex session is refused as a command it never was. `/tmp is nearly out of space`, `/workspace is mounted read-only` and `/s to skip` each get the toast "This session runs in Codex, which has no /tmp; nothing was sent" and a bell row: from the composer, from POST /send and `romp send`, and for a copy parked behind an open turn when the drain reaches it. Before the slash guard these reached the model as prose, as they still do on a Claude Code session.

## Cause

The Codex slash guard keys its refusal on `_is_slash_command`, the shape predicate that decides when a text may fire as a fresh prompt (its own comment: matched by shape, not against any command list). The route refuses every text that matches the shape unless its head is `/model` or `/effort`, and the parked-op drain refuses every parked `("command", ...)` op for a Codex session. A leading slash token is not a decidable command marker, and for a Codex session there is no CLI to ask what executes, so the shape test turns ordinary prose into a refused command.

## Fix

A checked-in constant, `_CODEX_REFUSED_HEADS`, names the slash commands the kernel knows a Codex session cannot take: `/clear`, `/compact`, `/new`, `/fast`, `/autocompact`, `/help`, `/mcp`. The census behind it (literal slash heads over `kernel/`, `ui/webview`, the timeline view, `cli/`, `bin/` and the extension source, tests excluded) is in the constant's comment: `/compact` (the battery, `_compact_or_park`, the judge and event model), `/clear` (the clear confirm, `sdk_backend`), `/fast` (the settings menu and the setter body), `/mcp` (the composer's palette), and the Claude CLI commands the docs and comments name as things a session types, `/new` (`docs/codex.md`, the `_CODEX_COMMANDS` comment) and `/autocompact` (the drain's command arm). `/help` is named by no surface: it is a Claude CLI built-in the guard has refused since it landed, kept so a typed `/help` never lands in a Codex conversation as prose. The per-cwd Claude CLI command cache is deliberately not a source: its warmth would make the same text refused one minute and delivered the next.

One predicate, `_codex_refuses(text)` (a head in that set, or a setter head), is read at both arms so they cannot drift:

- The route keeps `_is_slash_command` as its cheap first test, so plain prose still never calls `_codex()`. For a Codex session a registered handler takes the text; a setter head goes to the setter body and a wrong shape is refused with the hint; a head in the set is refused through `_refuse_codex_slash`; any other head falls through to the same `_route_setter_command` call the other backends take, which answers False for it, so the caller sends it idle or parks it as a command op busy.
- The drain refuses a parked command op only when `_codex_refuses` says so. Otherwise it takes the `_send_with_id` road every other backend takes, so slash-shaped prose parked behind an open turn reaches the model with its echo stamped at fire time.

The refusal's frame, sentence, bell row and `state["refused"]` are unchanged; `_CODEX_COMMANDS`, `_commands_for_sid`, the compact arm and the UI are untouched; the identity test stays `be is _codex()`.

`docs/codex.md` names every refused head in backticks and says that any other message beginning with a slash (a path, a word such as `/s`, a skill's name, a Claude Code built-in such as `/init`) reaches the model as text; that is the documented trade-off. The sentence claiming a command queued before the guard still reaches the model is gone (the drain refuses a queued known head). The follow-up path is described as it is: a bare slash body reaches the model when the session is idle (`askFollowUp` calls `_send_or_park` directly), and is refused at the drain when the session is busy. `/mcp` is listed as refused, with the note that a bare `/mcp` typed into the composer opens the MCP panel instead of reaching romp.

## Tests

All in `tests/test_codex_slash_guard.py`, on the module's existing Codex fake (its `send` records `("send", text)`). The final test file run against the unchanged tree: 6 failed, 15 passed.

- `CodexSlashRefused.test_slash_shaped_prose_reaches_the_model`: for `/tmp is nearly out of space`, `/workspace is mounted read-only` and `/s to skip` (with the path `/etc/hosts has the entry` as a control that never matched the shape) the route returns False with `state` empty, the composer's press (`_drive` sendMessage) records the send with no warn frame, `_deliver_text` answers `(True, "", False)`, nothing is parked and no bell row is filed. Fails before: the route returns True for `/tmp is nearly out of space`.
- `ParkedBeforeTheGuard.test_parked_slash_shaped_prose_drains_to_the_model`: a parked `("command", "/tmp is nearly out of space", "human", qid, True)` op drains as a send: no warn broadcast, no bell row, no compacting cue, op popped. Fails before: `be.calls` is `[]` (refused with a broadcast warn).
- `CodexSlashRefused.test_every_known_head_is_refused_at_the_route_and_the_drain`: every member of `_CODEX_REFUSED_HEADS`, bare and with an argument, is refused at the route (`state["refused"]` names the head, nothing sent) and a parked copy drains through the refusal (one broadcast warn naming the head, carrying the press id; nothing sent; op popped). Fails before: `AttributeError` on the missing constant.
- `CodexSlashRefused.test_the_refused_set_is_the_known_set` (replaces `test_the_refused_set_is_the_parked_set`): the refused examples are unchanged; the three prose texts are asserted slash-shaped (`_is_slash_command` True) and answered False by the route, since shape no longer decides; the not-slash-shaped controls `/tmp/x is a path`, `hello /clear` and `prose` stay, and `/etc/hosts has the entry` joins them (a second slash never matched the shape). Fails before: the route returns True for `/tmp is nearly out of space`.
- `CodexSlashRefused.test_the_doc_names_every_refused_head`: reads `docs/codex.md` and checks the paragraph's two claims apart: before "Any other message that begins with a slash", every head of `_CODEX_REFUSED_HEADS` appears in backticks; after it, every backticked slash word is an example of prose, so none may be a member of the set, and the heads of the three prose fixtures are not listed as refused. Fails before: the pivot phrase is absent from the old paragraph.
- `CodexSlashRefused.test_a_registered_handler_takes_the_head_ahead_of_the_refusal`: with `/clear` registered in `_CODEX_SLASH_HANDLERS` (a `mock.patch.dict`, restored), `/clear` and `/clear the thread` reach the handler with the press id and nothing is refused: no warn frame, no bell row, no send; unregistered again, `/clear` is refused. Fails before on the missing constant; its behavioural claim, that a registered handler is consulted before any refusal, holds on the unchanged tree too, and the test is here to pin that order (moving the refusal above the handler lookup fails it).
- `ParkedBeforeTheGuard.test_a_parked_setter_head_drains_through_the_refusal_with_the_hint`: parked `("command", "/model", ...)` and `("command", "/effort one two", ...)` ops drain through the refusal with the setter hint ("takes one Codex value here" and the example) on the chat panes, each with its copy id, nothing sent, one bell row each. Passes before the change (the unchanged drain refuses every parked command op on a Codex session) and pins the predicate's setter clause at the drain: replacing the drain's predicate with membership in the set alone fails it.

The drain tests call a fixture helper that runs one pusher pass and asserts the pass consumed the queue before any backend call is read, so a pass the walk skipped (its non-blocking lock acquire, a hold on the sid) reads as that rather than as a wrong list of calls; the fixture's teardown also clears `_inflight_ops` for the sid. The fixture's bell check reads the notice sequence (`_sync_notice_count()`) instead of the ring's length: the ring is capped at 40 rows (`SYNC_RING`), the added refusals fill it, and a later test's length comparison stopped moving.

Mutation checks run by hand, each caught by the module: `/init` appended to `_CODEX_REFUSED_HEADS` fails the doc pin; the drain's predicate replaced by membership in the set alone fails the parked-setter-head test; the refusal moved above the handler lookup fails the handler test; the drain's predicate forced False fails the known-head drain test, the parked-command test and the parked-setter-head test; the route's predicate forced True fails the two prose tests.

Targeted: `tests/test_codex_slash_guard.py tests/test_kernel_send_park.py tests/test_injected_voice.py tests/test_stage_marks.py tests/test_codex_backend.py tests/test_kernel_meta_command_gate.py`: 243 passed, 227 subtests passed. `ui/webview/codex-slash-guard.test.ts`, which pins the bell legend in `kernel.py`: 5 passed.

Full pytest on the final tree (Python 3.12, `-n 10`): 5 failed, 14411 passed, 11 skipped, 2329 subtests passed in 4:43, no worker lost. This is not a green run. None of the five is in a module this change touches, and none of the five modules references the route, the drain, the refusal, the constants, the handler table or `docs/codex.md`. Run alone at this head: `test_kernel_remote_identity.py::RemoteIdentity::test_second_alias_absorbs_the_old_row` and `test_session_host_restart.py::ServedRestart::test_with_hosts_on_the_turn_survives_the_restart_and_with_hosts_off_it_is_cut` pass; the two `test_sdk_backend.py` legs (`OptionsAssembly::test_ultracode_is_a_choice_everywhere_but_never_the_seeded_default`, `ReconnectReconcilesInflight::test_reconnect_settles_a_turn_stranded_by_the_teardown`) skip with `claude_agent_sdk not installed`, so in the suite they ran only under a peer module's stubbed import; `test_kernel_opening_chip.py::OpeningChipDecidingEvent::test_the_live_merge_threads_spawning_through` fails alone at this head and fails alone, with the same `AssertionError: unexpectedly None`, on main without this change, so it is red on this machine independent of the change.

Tier: fix


lczh:
Evidence executed against 4fcaeaa3. The mechanism holds: `_codex_refuses` (kernel/kernel.py:35999) is the one predicate the route (35962) and the drain (36229) read, slash-shaped prose reaches the fake as sends through all three doors and at the drain, every member of `_CODEX_REFUSED_HEADS` (35992) is refused on both roads, and the module's six reds on the base are the six the body names. The items below are small; none needs to hold the merge.

## At this head

**A well-formed setter pick parked as a command op is refused at the drain with a hint whose example is the refused text.** `_codex_refuses` answers True for any setter head regardless of shape, so the drain's arm (36229) hands a parked `("command", "/model gpt-5", ...)` to `_refuse_codex_slash` (36019), and `_codex_slash_refusal` (36010) words it as "This session runs in Codex: /model takes one Codex value here (for example /model gpt-5); nothing was sent." On the module's fake, `_route_meta_command` applies `/model gpt-5` on the idle session (`set_model` called, `queued: False`); the parked copy drains to that sentence and `set_model` is never called. Reachable, narrowly: the askFollowUp arm calls `_send_or_park` directly (19472), `_followup_body` leads with the raw text when the card yields no quote (3749), and a busy session parks that as a command op. Pre-existing in effect (the base drain refused every parked command op with the same sentence); this change codifies the clause in the predicate's docstring and pins the wording, so a later fix moves that test too. Either hand a parked setter head to `_route_setter_command` at the drain (a well-formed pick applies, a wrong shape gets the hint), or word the drain-side sentence as a pick that arrived as a message and did not apply. The pinned test (tests/test_codex_slash_guard.py:339) uses only wrong shapes, so the first road keeps it green as written.

**`_Guarded._drain` cannot tell a consumed queue from one the walk's failure arm dropped.** The helper (tests/test_codex_slash_guard.py:138-144) asserts only that `_pending_ops[SID]` is empty after one pass. The except arm of `_apply_pending_ops` (36294) writes a `pending ops apply:` traceback and pops the whole sid queue (36297), and `_refuse_codex_slash` broadcasts the warn and files the bell row before its own stderr line, so a raise between the refusal or the backend call and the identity pop (36258) reads to the helper like a delivered pass. On a scratch copy: `_refuse_codex_slash` raising right after wording the drain-side refusal left `test_every_known_head_is_refused_at_the_route_and_the_drain` green while 14 `pending ops apply:` tracebacks hit stderr; `_after_turn_opening` raising after a delivered send left `test_parked_slash_shaped_prose_drains_to_the_model` green with one traceback; a raise on the last op's tail left all four drain tests green. The two multi-op tests catch a raise on a non-final op (1 != 2 warns; the missing effort call) through their own counts, not through the helper. No production line raises there today: a fidelity gap, not a red test. Capture stderr around `km._apply_pending_ops()` in `_drain` and assert no line starting `pending ops apply:`; that prefix belongs to the two failure arms alone (36280 and 36294), and `tests/test_kernel_parked_ops_lock.py` already asserts it that way.

## Also at this head

**Merge order against #1880.** That branch registers native `/clear` and `/new` handlers in `_CODEX_SLASH_HANDLERS` (a follow-up adds `/compact`), which this change anticipates: the dict stays empty here, and the comment at 35993-35994 says those registrations stop the refusal. The two merge with one textual conflict, on the `_CODEX_SLASH_HANDLERS` definition line and its comment. #1880 also adds a whole-message rule (`_slash_alone`) a registered handler must satisfy, while the handler test here (253) pins that a handler takes `/clear the thread` as well as `/clear`. Whichever lands second re-reads the other: the registrations, the with-words pin, and the doc sentence "until the native clear lands" (docs/codex.md:169-172). Not a defect of either.

## Smaller points

- The body's last line still says `Tier: fix`; the owner relabeled to `feature` at 12:16Z (the author's `fix` was applied at 11:37Z), and the Tier policy check on 4fcaeaa3 is red with "body declares `fix`, but the `feature` label stands" and no admin approval. Changing the line to `Tier: feature` drops only the disagreement sentence; the approval is what lands it. A filing detail; the Fix section already carries a feature's design points.

The modules run, their counts, the mutations and the probes are in a second comment below.

lczh:
The test runs behind the comment above, on 4fcaeaa3 and on a read-only copy of the base bf87bc36.

- tests/test_codex_slash_guard.py at 4fcaeaa3, alone under the machine's shared pytest lock: 21 passed in 0.98 s; a second run gave the same count.
- The body's red-first claim, by execution: a `git archive` of bf87bc36 with the head's test module overlaid gives exactly 6 failed, 15 passed, the six ids the body names at the loci it describes: two `AttributeError` on `_CODEX_REFUSED_HEADS`, two `True is not false` from `_route_meta_command`, the doc pivot absent, the parked prose drained to `[]`.
- Probes for the first item, on the module's fake: `_codex_refuses("/model gpt-5")` is True and `_codex_slash_refusal("/model")` puts `/model gpt-5` in its example; `_route_meta_command` applies `/model gpt-5` on the idle session (`set_model` called, `queued: False`); a parked `("command", "/model gpt-5", "human", qid, True)` drains to the one-value sentence with `set_model` never called.
- Mutations for the second item, each on a scratch copy of the head, control 4 green and 0 tracebacks: `_refuse_codex_slash` raising after wording the drain-side refusal, known-head test green, 14 `pending ops apply:` tracebacks; `_after_turn_opening` raising after a delivered send, parked-prose test green, 1 traceback; a raise on the last op's tail after the refusal, the send, or `set_effort`, all four drain tests green; a raise on a non-final op, the parked-setter test red at 1 != 2 warns and the command-and-compact test red at the missing effort call.
- The tier line: the label events (the author's `fix` at 11:37Z, the owner's `feature` replacing it at 12:16Z), the check summary on 4fcaeaa3, and a synthetic run through `scripts/ci/tier_policy.py` `evaluate`, which reproduces the check's text and drops only the disagreement sentence when the body reads `Tier: feature`.
- Pin sweep over tests/, ui/webview *.test.ts, vscode-extension/src, docs/, cli/ and bin/: nothing pins the removed docstring and comment phrases or the old test name; `ui/webview/codex-slash-guard.test.ts` and `tests/test_kernel_send_park.py` pin lines this diff leaves unchanged. Census comment spot-checked: `/new` in the `_CODEX_COMMANDS` comment, `/autocompact` in the drain's command arm, `/fast` in the setter body, `/mcp` in the palette, `/help` in no non-test surface.

henrypinkard (approved):
Approved on the maintainer's decision of 2026-09-20 (the seven-head list as filed).

henrypinkard (approved):
Re-approved on the maintainer's decision of 2026-09-20 after the branch update from main (no change to the pull request's own diff).

PR #1892, opened 2026-09-19
Settings across machines, phase one B: the proposal is a needs-you notice card with Apply, Keep mine and Keep mine and pin

henrypinkard:
Phase one B of plans/settings-across-machines.md: a settings proposal is a needs-you NOTICE CARD, so the user is told where they work, with the three answers the design names. Phase one A (PR 1859) left the proposal as a record and a gear line with a sync notice as the interim alert; this PR gives it the card and retires the notice. The first commit is the plan section that reconciles this document with the notice-cards plan; the second is the code.

**The contradiction the plan section resolves.** The settings plan said the card takes the owner-less notice road; the notice-cards plan's kind table said the owner-less home takes no actions, and the kernel refused them twice (at the post and at the click). Both refusals exist for `send` and `quarantine`, whose runners act on the card's own session. The settings kind names no session, no text and no target: it acts on this kernel's own proposal record through `/setting-proposal`'s checks. So the kind table now says whether a kind NEEDS AN OWNER (`NOTICE_ACTION_KIND_OWNER`: `send` and `quarantine` true, `setting-proposal` false); the owner-less refusals key on that fact, word for word for the two kinds that need one, and admit the settings kind. Not taken: a reserved kernel-owned producer sid (the plan's own model rules out a second reserved owner), a stand-in session (the misattribution the model forbids), an unconditional relaxation (a `send` on the owner-less home has nowhere to send). The notice-cards plan's sentence is restated as "no action that needs an owner".

**The kind is the kernel's alone.** `post_notice(..., internal=True)` is the one door for `NOTICE_ACTION_KINDS_INTERNAL`; the route and `romp card` never set it and are refused the kind by name ("posted by the kernel alone"): a producer outside the kernel could otherwise put a label on a button whose body answers a pending proposal the other way. The body shape is checked at the post (`{store, host, gt, answer}`, the store synchronized, the answer one of three, no click input); the runner hands the stored body to `_answer_setting_proposal` unchanged, so every guard of the route (the machine's record present, the stamp current, a local click that outranks the record refused and dropped, the pin from the local origin) holds on the card exactly as in the gear, and the refusal's words are the route's.

**The card follows the record.** `_proposals_write`, the one writer of the records, diffs the file before and after each write: a (store, machine) record that is new or moved (value or stamp) posts its card (a new revision under the same key `proposal.<store>.<machine>`), a record gone for any reason (an answer from the card or the gear, the values coming to agree, a local click that outranks it, a pin, a re-key onto the row key, a detached row) expires it. Producer `settings`, `needsYou`, `dismissOnAction`, the title "Task tracking: TESTHOST proposes off; this machine is on", a body in the user's terms saying nothing changed here and what each answer does, three actions Apply, Keep mine, Keep mine and pin this machine. A card the user clears without answering leaves the record standing (the gear's line shows it; the poll finds it standing and posts nothing until the stamp moves). The sync notice of one A retires: two alerts for one event is noise.

**Tests, red first at 892d7c0a** (main after 1859; the three test files copied into a detached worktree then restored) under `--python 3.13` and under `3.14` the same: `17 failed, 92 passed` over the mesh module, the notice store module and the new served lab (the require flag on). Executed: no card posts with a record, `{} != {'proposal.compact-suggest.TESTHOST': (1, ...)}` and `0 != 2 : the moved stamp is a new revision of the card`; the owner-less click on the settings kind refused at the base, `(False, 'an owner-less card has no actions') != (True, '')`, twice; two machines, no cards, `[] != ['proposal.task-tracking.TESTHOSTB', 'proposal.task-tracking.TESTHOSTC']`; the re-key with no card to move, `[] != ['proposal.task-tracking.hostb']`; the kind table `('send', 'quarantine') != ('send', 'quarantine', 'setting-proposal')` and the three refusal texts naming two kinds; the served lab's three drives, `unexpectedly None : the Task tracking card drew` and `the click ran` twice (the push landed as records at the base, and no card came of them). The round-two and round-three tests that counted the sync notice now count card revisions and red the same way (`0 != 2 : one notice per RECORD`). Green at the head: the mesh and notice-store modules `105 passed`; the served lab `4 passed` under the require flag; the whole Python suite under `--python 3.13`, devbox, capped, no xdist, `14514 passed, 56 skipped, 2337 subtests passed in 2060.59s (0:34:20)`; the whole extension suite `5125 tests, 5119 pass, 0 fail, 6 skipped` (no TypeScript or JavaScript changed; the extension's pins on kernel text read the new kind table); both test trees grepped for every changed literal (the kind tuple, the two owner-less refusal texts, the retired notice's name).

Mixed mesh: unchanged from one A (an older kernel adopts our push and applies our poll's value; a one-B kernel proposes and now draws the card). `/version`'s shape is unchanged.

Tier: feature


ksarma:
A note after the merge. The evidence below was executed against 613944f3d, the merged head; file and line references are on that commit. A is the kernel with the record, B the kernel that pushes.

- **A local click that ends the divergence leaves the card standing with a false title.** The setters never touch a record; the rule that drops one the local value has matched or outranked runs only in `_propose_peer_settings` (kernel/kernel.py:51701), from a poll or a received push. After B's push, a click on A to the proposed value (or a remote-origin write of it) leaves record and card standing, the title saying this machine is on, until a button is pressed, since B pushes nothing more once the stamps agree; Apply then refuses, though that drop clears it. Fix: apply that two-part rule after any successful write, whatever its origin, at the click's door or with a re-read after the lift, not inside the setters (the stamp lift at 51706 calls them under a records copy the pass writes at 51728, which re-posted a card dropped there) and not under `_NUDGE_LOCK`.

- **The card key is not injective over machine names.** `_proposal_card_key` (51592) maps characters outside the key grammar to "-" and cuts at 64, so "a:b" and "a-b" share one key, as do two long aliases alike through their first 40 or so characters. One card shows for two records; Keep on it strands the other record with no card, no stderr line and no re-post until its stamp moves (Keep alone does this, since Apply and the pin drop the store's every record). Ordinary attach rows admit "@", ":" and long names; a space needs a hand-made push. Fix: when the sanitizer or the cut changed the name, append 8 hex of a sha1 of it inside the 64, in the key function alone so post and expire agree; refusing instead is the same silent absence.

- **A late push with an older stamp rewrites the record downward and re-posts the card.** The kept-stamp check at 51717 and the standing-record check at 51719 test equality. A same-value push below the record's stamp lowers it and posts a revision, the next poll lifts it and posts another, re-showing a cleared card; after a Keep the same push raises the record and erases the kept stamp, and the next poll re-posts under it. The push's stamp is fixed before its thread starts, so the window is one HTTP exchange; the ordinary pass does not flap. Fix: widen both checks to "not newer", the record branch still adding the store to the poll's returned list; patched, each scenario posts once and the Keep survives.

- **A kernel updated with a record already standing never posts its card.** Cards post only from the writer's diff in `_proposals_write` (51589); the ordinary pass finds the record unchanged and writes nothing. Executed: a record raised under the base kernel, this head loaded over the same state root, three passes, no card until B's stamp moved. Fix: at boot or notice attach, post rev 1 for every standing record whose key has no post row under Notes and no archived revision, as `_held_mail_backfill` (24475) does; "no standing post" alone would re-post a card cleared without an answer, which keeps its post row.

- **Kernels from before 1859 share one record under the name "another machine".** Such pushes name no host, so `_apply_mesh_settings` (51390) keys them alike (every kernel since the settings push began). Two of them rewrote the record on every push, a revision each; a Clear was undone within a pass, Keep survived one. A card-layer fix is the wrong layer: `_answer_setting_proposal` refuses a body whose stamp is not the record's, so a frozen card over a swinging record refused every button; keying by the arriving tunnel is not available either, since a push arrives on loopback through the pusher's own forward. Fix at the record layer: the "not newer" skip above, plus a Keep under this name compared by value, since two pushers' stamps are not comparable against one kept entry (a later flip back to the kept value is then silent too). The body's mixed-mesh sentence omits that such pushers share one card and that Apply or the pin ends it.

- **The plan names a runner that does not exist.** plans/settings-across-machines.md:127 names `_notice_run_setting_proposal`, which kernel.py does not define. The dispatch is the inline branch at 24681 of `_notice_action_run`, calling the answer function directly at 24686. plans/notice-cards.md:817 names two such runners for send and quarantine, neither existing either; the new text extends that.

## Smaller points

- **The card says "this machine" and names only the proposer** (`_proposal_card_post`, 51620). On B's merged dashboard A's card sits under an "A:Notes" header naming B as proposer and A as this machine, with no host on the card face when ungrouped, and its pin label pins A, where the click routes. Fix: name the kernel with `_self_host()` in title, body and labels, the pin label inside 60 characters or the card is refused; a viewer's prefix for a host can still differ from that kernel's own name.

- **The record and its card are kept in step only by the diff** (`_proposal_cards_follow`, 51597). Not permanent, but nothing on A repairs it: a failed post left a record with no card through three passes and a restart, a failed expire left a card whose buttons all refused, and only B's next moved stamp, or a Clear on the stale card, ends either. Fix: reconcile on the pass, at boot and on the push-receiver leg (a receiver that polls nobody never polls), matching cleared revisions by stored body stamp rather than key alone, and expire a stale card when the answer route refuses it.

- **The title embeds the alias unbounded.** `_safe_ssh_host` admits 255 characters; with the longest label a 149-character alias posts and 150 is refused (one fewer when both values read off), leaving the record standing with one stderr line and no card. Fix: compose the title to fit `NOTICE_TITLE_MAX` and keep the full name in the body; the action bodies have no length cap.

- **The retired rule still stands in prose.** docs/reference.md:129 and plans/notice-cards.md:508 say an owner-less card carries no actions (508 contradicts 821 in the same document); bin/romp:1218 repeats it, and ui/webview/feed.ts:135 and kernel/kernel.py:51268 ("for the sync notice") name the two old kinds and the retired notice. All five lines predate this PR; no test pins them. Fix: mirror notice-cards.md:821, say in the reference that the command and the route are refused the kind by name, point feed.ts at the kind table, write "for the proposal card's title" at 51268, qualify bin/romp's closing clause.

- **The kind table's row is still a stub.** plans/notice-cards.md:814 reads "defined in that plan and added to the table by its phase one", no row carries an owner value, and the header at 803 names `_NOTICE_ACTION_KINDS`, a name in neither base nor head. The settings plan's line 125 says each entry carries an owner value, which describes neither table; the code keeps that fact in a separate map. Fix: give the row its body shape, no click input, owner false, kernel-posted; correct the header's name; and either add owner true to send and quarantine or reword the settings plan's sentence to name `NOTICE_ACTION_KIND_OWNER`.

ksarma:
## Tests

Three gaps at 613944f3d; each fix above needs a test red at this head.

- **Key collisions.** No test calls `_proposal_card_key`; every test host is short and inside the key grammar. Add in TheCard two names that sanitize to one key and two that cut to one, asserting one standing card per record with distinct keys right after the second push, the other card still standing after Keep on one, and a direct pin of the key function on a pair colliding under compact-suggest and not auto-nudge, since the cut is per store.

- **The no-poll click.** The one local-click test (tests/test_mesh_settings_adopt.py:716) polls right after the click; deleting that line turns it red at 717. Add: set locally with no poll, assert `cards() == {}` and the store's hosts empty, for the take-the-value and re-assert shapes, on the poll and push-receiver paths, one with origin remote. Two existing tests pin today's behaviour (the Apply refusal over a newer local click, the gear line after a click) and move with the fix.

- **The older stamp.** Nothing pins the downward rewrite: the PR's test modules pass unchanged with the "not newer" patch, and the served lab drives newer stamps only. Extend the Keep test with a push at a stamp between the local and the kept one, asserting one post and the kept stamp intact; add a same-value push below the record's stamp that posts no revision and leaves the stamp in place.

PR #1894, opened 2026-09-19
Design: a full frame honors the client's held tail (chat history regions)

henrypinkard:
## What this is

A design paragraph in `plans/chat-history-regions.md`: the kernel-side complement to PR #1860's page-side containment (guard 3), so the owning kernel stops emitting a full frame whose `tailLo` lands below a run the page holds. **Docs only.** The code waits for the user's word, since the containment already shipped.

## Why the kernel emits the bad frame (diagnosed by execution)

`tailLo` is re-derived on every full frame as the turn at `head_from = max(0, total - WIRE_TAIL)` over the kernel's *current* event list, consulting nothing the client holds (`_send_chat_proto2`, `_tail_run_start`). So when the sid's current transcript is shorter in turns than when the page minted its resident tail run, the cut maps to a lower turn and the frame carries only the new short tail. Driven on a hermetic two-kernel relay:

- **Producers:** a live shrink 600 → 250 turns re-derived `tailLo` 475 → 125 on one kernel; a fork or resume to a shorter transcript is the same shape (older turns still on disk in the pre-fork file, absent from the current one).
- **Non-producers (driven, so named from execution not analogy):** a compaction only moves the render floor (`cutTurn`), it does not renumber turns, so `tailLo` tracked the growth up to 481, never below the held run; a plain floor rise keeps the cut a fixed `WIRE_TAIL` of events from the unchanged end.
- **Vehicle:** `baseGone`/`noBase` to a *still-connected* page (the shrink makes the held base uuids stop mapping), not a re-dial. A healthy relay holds one socket and the keepalive traverses the splice every 10s (measured: 1 open, 0 closes, `ka` every 10.0s over 97s); a genuine reconnect is a secondary vehicle.

## The design

- **Wire field:** the page sends `heldTailFirst` (its resident tail run's first event key) on `ready` and `needFull`.
- **Kernel read:** serve from `min(head_from, index of heldTailFirst)` so the frame is a superset of the held run (guard 3's `split.before` empty → applies, no ask); when `heldTailFirst` no longer maps, carry `rebased: true` so the page drops the stale run under the one landing notice rather than holding it across a refused-frame latch.
- **Fallback:** a page that sends no `heldTailFirst` gets today's behavior; the field is additive, so old and new pages and kernels interoperate.
- **Test shapes (from execution):** the superset case, the shrink 600→250 (`rebased`), the fork/resume (same), the compaction control (still applies, no rebase), and the discover-window cold-boot `tailLo` null (1877's regions-less landing).


ksarma:
Read at head 13c516240 (lines are in `plans/chat-history-regions.md` there); code claims executed at base 797850260, the design read against PR 1860 as executed at its merged head. The paragraph is the right complement to look for: a kernel serving what the page keeps.

### At this head

**Floored lists** (line 223). A held first key not mapping in the current list is read as proof the held turns are gone, and `rebased: true` drops the run. With the pusher's list floored (render floor above 0), a held first under the floor does not map there either, though the parse resolves it; executed twice independently through `_send_chat_proto2` and the page's merge: today that page keeps its straddling run above a frame from index 0; here it drops a legitimately held run under a signal the drop was right. A silent drop leaves the reader checking, a visible refusal leaves them knowing the state is unresolved, a confident false signal leaves them not checking: this lands past where the guard was, on the road meant to fix it. The change: reserve `rebased` for a first resolving against neither list nor parse; a first under the floor serves from index 0, tailLo at the floor turn, matching the delta path.

**The carrier** (line 216) cannot deliver the aim at line 199 or the no-ask at line 222. `ready` posts once at bundle evaluation and never again on redial, so it cannot carry a held key; the primary vehicle, a baseGone or noBase push to a still-connected page, has none (a noBase frame answering a needFull does have one). Today's frame goes first and the superset or rebased frame second, after guard 3 refuses and `needFull` carries the key, a frame added, not removed; a frame sharing no uuid with the held run bypasses guard 3, so nothing fires and the merge drops the run. The base's recorded first (`pc['first']`), already the held older edge, could store the key for the pushed frame, though written on send, not on apply, so a page-sent key wins where both exist; a redial's first frame still needs one on the dial or the page's first message.

**The superset half** (line 220). A held first mapping under the cut never trips guard 3 today, so `min(head_from, idx)` buys no containment there, while making the full frame's size client-controlled, up to the floored list: a tail run that absorbed the reader's history fills re-sends the whole transcript, a cost the server cannot refuse. One shape benefits: a reconnect after missing more than WIRE_TAIL events, where today's no-shared-key merge glues the held run onto a new frame over an unmarked hole; so condition rather than drop it: serve the superset only when the held first maps and the last does not (the page sending both keys), with the `/perf` sends counter as the cost reader.

**Test shape (5)** (line 236). A live session idle 72 h, past the 48 h horizon, resolves through `_session_row` but not the tail-run start's narrow roster; its full frame carries tailLo null and the page drops every held region. Resolving it there the same way, the one-line change named on 1860, is what the superset path needs to run, and is absent.

**The rebased outcome** (line 225). What the reader sees is unspecified, and the notice named is Part B's cancel affordance for an unapplied landing, meaningless for a replacement that is not a landing. Say what is shown, when, and how it is dismissed; who caused the rebase is not on the wire, so carry a reason or pick one treatment (silent, as today, is legitimate).

**The page-only half** (line 199). Showing a refused and latched frame in the pane rather than the journal needs no kernel or wire change nor version agreement, is doable today, and makes a wrong predicate survivable on either side. It stays open, with 1860's other open items (latch key, headKnown whole-frame exemption and rewind discriminator, chatTail gap-ask loop, per-scenario ask pins), each applying where `rebased` is absent: an old kernel, the first frame on the primary vehicle, the lying tailLo.

### Smaller points

- Line 216: `heldTailFirst` is the run's first event key, not excluding constant-kind overlay cards; a run from a floor-0 list's head keys on system:head, which maps everywhere, so the flag never fires and the index minimum picks a card's index; define it over transcript rows.
- Line 207: the re-pointing producer is listed as executed though only the shrink was driven; line 210 counts WIRE_TAIL in turns, not events.
- Line 221: a frame re-carrying the whole held tail run under a lying tailLo passes guard 3 and the merge drops the mid gap; comparing `msg.tailLo` with the run's lo when first keys match goes unsaid.

PR #1896, opened 2026-09-19
The older wire's landing shows the wait, survives a socket death with its word, and lands with the click's time; the exec harness runs guard 3 for real; one overlay-kinds constant

lczh:
Tier: fix

## The defect

The follow-ups the maintainer's post-merge note on #1877 (the client merge guard) named, reconciled with #1860 (the dropped-history fix), which merged after it and already settled the rule the note was most concerned about. The five items below are that note's set, landed together because the note asked for them together; the fifth is a dedupe with no behavior change and no red-first test, and the Tests section says so.

What #1860 settled on main before this change: its reconcile commit dropped its own keep-regions (no tailLo) and not-proto-2 branches, so #1877's null-tailLo branch is live again for a frame that shares a uuid with the held session (a `regions-dropped` row with `no-tail-lo`), and its guard 3 became the merge's own loss rule (`would-drop-held`: only a proto-2 frame with a numeric tailLo whose content starts later than the run it claims to cover, refused with a `full-frame-desync` row and one gap ask). It re-cut the landing-notice lab's null-frame road back to #1877's shape and added the not-proto-2 `regions-dropped` row beside #1877's block. On that main the client-merge-guard lab is three for three, so the "1 failed, 2 passed" the note measured on the two heads' plain merge is gone. What remained open, and is fixed here:

1. The executed harness over `upsert` (the proto-2 exec tests) answered `sharesAnyUuid` with the Proxy's no-op, so #1860's guard 3, which reads it, never ran there and its refusal could not be tested in that file. The harness was green for a dishonest reason.
2. The older wire, which every regions-less proto-2 navigation takes since #1877 (a window has no regions to join), showed no wait state. The window road shows the one landing notice for a navigation's ask; on this wire a reply chip's, the reload restore's or an unplaced window's re-armed jump waited with nothing on screen (a notch or a comment tick usually lands on resident content, so they showed the gap least).
3. A socket death with such a landing in flight lost the jump silently. `onWireDown` counts a live landing by the notice's session or a gapped window ask, and this wire set neither, so no `wsdown-lost` word was filed and no toast said the jump was lost, where the same sequence on the window road toasts.
4. `chatHead` re-armed the landing with `anchorT` and `kind` both null, so the pointer-exact `locateDiag` row of every regions-less navigation filed without the time and kind the attempt row and the window ask's record both carry. On main, a focus landing through the older wire files its exact row with `anchorT: null` although the focus carried a time.
5. `chat-regions.ts` carried its own copy of `OVERLAY_KINDS`, already defined in `send-pending.ts` (pinned equal to the kernel's `_OVERLAY_KINDS` by `tests/test_send_pending_overlay_kinds.py` and imported by `queued-held.ts`): two copies for one constant, and a kernel-constant change broke two pins for one change.

## The fix

One arm for the older wire, `waitOnOlderWire`, which both of `scrollToAnchor`'s older-wire branches (the re-point of an in-flight fetch and the fresh fetch for a regions-less session) end in. It arms the landing for `chatHead` and writes the trail's word as before, and adds what the window road has: for a navigation it shows the landing notice with the click's time (the keep-offset re-land of the reader's own row, `relandAsk`, is no navigation and shows nothing, as its window ask would not), and it records the click's time and kind in a mark keyed to the anchor (`pendingOlderMark`, beside the older-anchor maps). The first attempt's mark stands across a durable seek's re-attempt passes on the same anchor (its re-arm carries no time; a window ask's record stands the same way across "pointer-fetch-waiting" passes), and a click on another anchor replaces it.

The notice comes down where the wait ENDS, never at the reply itself. A reply the pass re-arms on (another chunk is needed) leaves it standing and `waitOnOlderWire` re-shows it in place; the landed or honest-failed attempt brings it down inside the pass (`landActive`, after the attempt, when no window ask's record owns the notice and no deep-link fetch is in flight for the tab). That decision sits inside the pass on purpose: in compact mode, the default, `showActive` defers the pass a frame past `chatHead`'s reply, and a hide at the reply left that frame with no landing for `onWireDown` to count, so a socket death in it lost the jump with no word (the review's second round). The paths where a reply lands nothing end the wait with the notice and a word: a fault (`head-fault`, as before), a stale reply whose `beforeUuid` no longer heads the resident list (`head-stale`) and a missing anchor (`head-missing`); for a navigation with no seek behind it (a reply chip's, the reload restore's) the stale and missing replies toast the fault's "the history could not be loaded just now", where before this the notice went down and nothing was said; a seek-backed landing re-arms from its own record on the next pass and its backstop speaks for it, so it files the word without the toast. A reply for a tab the reader left drops the landing with its claim and the notice with it.

Every release of the older wire's claim brings the notice down too. The seek's backstop (`failSeek`), the seek note's cancel (`cancelSeek`) and a time-only navigation all release through `releaseSeekFetch`, and none of them hid it: on this wire the notice stood with no ask behind it, beside the backstop's own "couldn't locate" toast, until a click, a later landing's reply or the socket's death (the chunk's `chatHead` reads the released claim as a scroll-back's and hides nothing). `releaseSeekFetch` hides the notice when it is this session's and no window ask's record owns it; a window landing's notice stands through the release, as its record does.

`chatHead`'s re-arm reads the mark for the anchor it names and carries the time and kind through, so the pointer-exact row files with both; a mark another click left, or a scroll-back's arrival, carries neither, as before. The mark goes with the fetch's other records (`forget`, `releaseSeekFetch`).

The notice's click is an honest cancel on this wire too: `cancelLanding` with no ask record releases the fetch's claim (`releaseSeekFetch`, the seek's own release), so the chunk arrives as a pure prepend re-anchored on the reader's row and `chatHead` re-arms nothing but that row. Its `locateDiag` row names the landing it canceled from the fetch's mark, read before the release deletes it: at the click the pending anchor, time and kind are null (`landActive` nulls them at the end of the pass that armed the wait), so the row filed with none of them where the window road's cancel names its anchor from the record. Because the notice is up for the wait, `onWireDown`'s existing rule counts the landing: a socket death mid-walk files `wsdown-lost` and toasts, with no change to that pinned line.

The exec harness's `liftUpsert` carries render.ts's real `sharesAnyUuid` (lifted whole), a real refusal latch, a recorder for the full-frame asks and a `vscodeApi` sink, and one case executes guard 3: a frame whose numeric tailLo claims the held tail run while its content starts later is refused (held rows and regions stand, one `full-frame-desync` row, one gap ask), and the frame that re-carries the run from tailLo applies. `chat-regions.ts` re-exports `OVERLAY_KINDS` from `send-pending.ts`: a dedupe with no behavior change, the same seven members before and after, pinned member-for-member against the kernel's constant by the existing regions test and imported unchanged by the exec harness.

## Tests

Red first, the committed tests against the stock `render.ts` and `chat-regions.ts` (main). The exec module: the `waitOnOlderWire` case fails on the lift (no such function), the `chatHead` case fails with `pendingAnchorT` and `pendingAnchorKind` re-armed null, and the `cancelLanding` case fails with nothing released. The windowing module's re-cut source pins are red. The served client-merge-guard lab (Chromium, the window lab's hermetic kernel) fails its older-wire roads: road 1 reads no notice while the older ask is parked and the exact row with `anchorT: null`, and road 4 (a fresh page, last: a focus landing waiting on the older wire with its ask parked, then the socket closed) records the death with `trail: ["pointer-fetch-older"]`, no `wsdown-lost`, and `toast: null`. The guard-3 exec case, run once with the harness's old no-op `sharesAnyUuid`, is red at "the held rows stand" (the frame dropped the two earlier rows): the scope change bites. Item 5, the re-export, has no red-first test by design: a constant moved is not a behavior, and the member-for-member pin is green on both sides.

The review's second round added three red-first cases on top of the first cut. Against the first cut's code, the exec module is 3 of 17 red: the `chatHead` case at `hides` `[1]` where `[]` is expected (the notice hidden at the reply and re-shown a pass later), the cancel case at the row's anchor, time and kind all `undefined` where `u96`, the time and `user` are expected (the real `releaseSeekFetch` in the scope deletes the mark before the row is built, and the pending fields are null at the click), and the release case at `hides` `[]` where `[1]` is expected (`failSeek` and `cancelSeek` left the notice standing). The windowing module's re-cut pins are 1 of 25 red. The served lab gains a fifth road (a fresh page, before the socket death): with the older ask parked and the notice up, the notice is clicked, then the parked ask alone is released; against the first cut it is red at the cancel row's `(None, None)` for the anchor and time. The same road catches the reviewer's failure scenario: with the cancel's release removed, the chunk re-arms the canceled target and it lands (`pointer-exact` filed, no keep-offset row), which the road reports as red.

Green after. Node, built singly with the extension's test build: the exec module 17 for 17, regions 15, windowing 25, exact-tail exec 13, delta-resync 9, needfull-latch-hygiene 5, reveal-progress 21 (its explicit world names `liveWindowAsk` beside the hides it already counted), chat-older-restore 8, glow-history 8, locate-audit 4, render-scroll 19, seek-indicator 4, send-pending-overlay 6. The served client-merge-guard lab is five for five on two consecutive runs: road 1 shows the notice ("click to stay here") while the older ask is parked, hides it at the landing, and files the exact row with the focus's time; road 4 files `wsdown-lost` and toasts "the connection dropped before the jump"; road 5 files the cancel row with the anchor and the click's time, the released chunk lands as a keep-offset restore on the reader's own row, no `pointer-exact` row is filed for the canceled target, no second older ask goes out, and the notice stays down. The sibling served labs are green on the changed tree: the landing-notice lab (16) and its anchor sibling (5), the landing-settles lab (13), the reload-restore lab (3), #1860's relay-tail lab (9), the history-regions lab (6), the proto-2 served lab (2), the served-labs census (4), the overlay-kinds pin (3) and the kernel module (460). The extension suite runs once with 5,136 tests, 5,129 passing, 6 skipped and one failure, the pre-existing KaTeX pixel probe that fails on main in this environment; the type check is clean; the static pin scan reports 8,526 literals checked and none lost.


lczh:
CI's one red was tests/test_held_mail_chat_served.py HeldMailChatServed (the deny-note buttons not latched), a lab this change does not touch: run locally against main the same module gave 1 failed / 4 passed once and the single test passed; against this branch 5 passed. Reopened to re-run the checks.

PR #1897, opened 2026-09-19
The chat approval box lab reads the latch with the click's request held at the socket (a transient asserted by timing)

henrypinkard:
`tests/test_held_mail_chat_served.py` failed twice in hours on CI's extension job, on commits that cannot touch it (main at 09c853ed, a docs-only merge; PR 1894's run): the same assertion both times, line 275, "latched on the decision" reading the three deny-step buttons re-armed. The latch is a transient: the kernel's refusal (no bus in this lab) re-arms the row and says why by design, and on the runner the answer landed before the read. A test may not assert a transient by timing.

## The fix

- The driver parks the click's `noticeAction` frame at the socket (the merge-guard lab's shim: `WebSocket.prototype.send` holds frames whose type is in `window.__hold`; `window.__release` sends them) while it reads the latched buttons, then releases it and reads the re-arm and the refusal line. Both decisions the lab takes, the mid-press Approve and the bare Deny, hold and release exactly one request, and the test asserts that count.
- The "postal bus unreachable: Connection refused" text in those runs is the lab's designed refusal inside its recorded result (`HELDCHAT: {...}` on stderr): the lab kernel runs client-only with a free port and no bus (`kernel_env` in `tests/test_ship_reship_served.py`), so the refusal is the expected answer to every decision and there is no bus startup to wait on. No change there.

## Red-first

- The shape, executed at the base d0f1ab02 (a detached worktree): the base lab with the kernel's answer forced ahead of the read (a three-second pause before each latch read stands for CI's runner) fails three tests, the two latch reads and the mid-press test that reads Approve's latch, with CI's exact message. The fixed lab passes with the same pause in place, and as it is.

## Verification

- The lab as it is and with the early-answer pause, green on the branch head; tests only, no production file touched. CI is the whole-suite gate.

Synthetic fixtures only; no personal identifiers; no em-dashes in added lines.


ksarma:
A note after the merge; evidence executed against 1bbf9a3eb, the merged head.

The latch read is deterministic under the hold: the page's `send` (kernel/kernel.py:58924) writes the click's frame synchronously while the socket is open, so the parked `noticeAction` frame cannot be answered before the read, and the only re-arm paths (a `noticeActionDone` matched by item id, a rebuild on an actions-signature change) are unreachable while it is parked. Executed: a fourth held file landing during the hold left all three buttons disabled through a three-second pause (5 of 5); releasing before the read re-armed them (the shipped order is load-bearing); the head's latch reads passed 60 of 60 where the base's deny read failed 2 of 60 under the same load; the red-first pause fails 3 of 5 on the base file and 0 of 5 on this one. The four CI failures named in the body all sat at the deny read, which this closes.

What it leaves: the three at-bottom assertions (lines 241, 256, 273) are still the scroll listener's echo race (the listener at ui/webview/render.ts:13852 at this head, described on #1890), red at the same rate at base and head under a paired 70-run drive (5 and 8 of 70, 4 and 6 of 64), so the extension job can still go red there for a reason this change does not touch.

PR #1898, opened 2026-09-19
Pane docking: the empty space inside a pane is a grab surface (the inner pages detect, the shell drags)

henrypinkard:
Tier: feature

The empty space INSIDE a pane is a grab surface, on the plan's seam (the user switched the docking kit on, reached for the empty area of the feed below its cards and of the sessions band, and got no hand and no drag: the hand showed only on the ring). The first commit is the design paragraph, `plans/pane-docking.md` section 3 extended; the second is the mechanism and its pins. Filed unarmed.

## The design (section 3, added)

The shell can see only its own chrome and what it reads into the same-origin pane documents; a pane's content is an iframe whose pointer events never reach the parent. So each inner page owns the detection and the shell owns the drag:

- While the kit is on, the shell marks every pane document's body `pane-docking` and injects the grab detector (`dist/pane-grab.js`, a new bundle, the same road as the cursor stylesheet it already injects: no kernel change, no new inline script; the `/dist/` route serves it by name). With the kit off the pages carry no class, no detector and no rule.
- The detector knows each page's EMPTY BACKGROUND by app (`window.__rompApp`): the feed's list, its columns and their card lists where no card, chip or control is under the pointer; the sessions band's SVG and its wrap outside every lane, bar and mark (`svg *` yields); the outline's list below its rows and its footer; the Files pane's empty state and its body. The press must land on the container ITSELF (`matches`), never a descendant. The chat is the deliberate exception: a press-drag over the transcript is a text selection (section 3's own rule), so its grab surface stays the strip's empty run.
- The open hand shows over exactly those targets: the detector toggles a body class (`pd-grab-hover`) on pointer moves by the target under the pointer, and the rule `body.pane-docking.pd-grab-hover{cursor:grab}` is inherited by the hovered container; a card's text never wears a hand it cannot honour (the class drops the moment the pointer is over a card).
- On a primary-button press with no modifier (Option is the shell's own path) on an empty target, the detector CAPTURES the pointer on the pressed element, so the moves keep flowing to that document once the pointer leaves the iframe, and posts `{romp:"paneGrab", app, clientX, clientY, pointerId}` to the shell. The shell (`onGrabMessage`) resolves the pane from the message's source frame and arms exactly the press it arms on the ring, `beginPress` with the frame's window: the 4 px slop, the closed hand, the live outline, Escape, the drop as a tree move, hearing the frame's pointermove and pointerup through the frame's window as it already does for Option-drag. No transparent capture layer over the iframes: capture inside the child document is what keeps the pressed pointer's events addressable at all.

## Proof

**Node** `ui/webview/pane-grab.test.ts` (fake targets: `matches` and `closest` are all the decision reads): the feed's list, columns and card lists are empty and a card, text inside a card, a button or a role'd region is not; the band's SVG root and wrap are empty and every SVG child is not; the outline's list and footer, the Files empty state; the chat has no empty background and unknown apps, a null or a throwing target never grab; the primary button with no modifier.

**The served lab** `tests/test_pane_docking_served.py` (real pointer events; the spot is found by the page's own detector through `elementFromPoint` over the list, so the press lands where the page says its background is):
- The feed: a spot on `#feed-list` below the cards; the hovered element's computed cursor is `grab` and the body wears `pd-grab-hover`; over a card the cursor is not `grab` and the class is gone; a press there, 14 px of travel, arms the shell's drag (the body reads `grabbing`), the outline shows the chat's left half, and the release lands the feed right before the chat in the row with every iframe still filling its pane.
- The sessions band: a spot on the SVG root outside the lanes; the open hand; a press there lifts the band (the outline on, the closed hand); Escape drops it where it was.
- OFF: the feed document carries no `pane-docking` class, no detector script and no detector global; the shell-side byte-identity checks stand.

**Red first at the base 387a40ee** (the engine module and the build entry checked out from the base, the detector moved aside):
- `tests/test_pane_docking_served.py` under REQUIRE: 2 failed, 9 passed. The two empty-space legs fail on the measured absence, `an empty spot in the feed below its cards (the page's own detector says so): {'error': 'no detector'}` and the same for the band's SVG (no detector is injected at the base, so no spot can be found and no hand shows); the nine that pass are the merged engine's own pins, present at the base.
- `ui/webview/pane-grab.test.ts`: red by absence (the detector module does not exist at the base; moved aside for the base build).

**At the head**:
- `tsc --noEmit`: exit 0. The node test files (`pane-grab`, `pane-dock`, `pane-tree-fixed`, `pane-tree`, `panedock-main`): 53 tests, 53 pass. The lab under REQUIRE (chromium) with `tests/test_settings_tabs_recut_browser.py`: 24 passed.
- The whole extension suite (CI's glob): 5142 tests, 5136 pass, 0 fail, 6 skipped.
- The whole Python suite stands from PR 1884's second round (14336 passed, 0 failed): this pull request changes no Python beyond the lab, which ran under REQUIRE above.

## Round two (the verifier's read of 3c17aa02)

- **M1, a press that outlives its release.** The detector posts in `pointerdown`; the shell's message task can run after the `pointerup`, before the engine's own listeners existed, so a press stood with no button held: the next move past the slop armed a drag, the next click dropped the pane, and every grab was refused meanwhile. The detector now posts the release too (`{romp:"paneGrabEnd", pointerId}` on `pointerup`, `pointercancel` and `lostpointercapture` of the captured pointer) and the shell cancels an unarmed press on it; defensively, any `pointermove` with `buttons === 0` cancels the press. Pinned with the verifier's drive: a `paneGrab` message from the feed frame with no button, then a 60 px move: nothing arms, the press is cancelled, no `pd-drag`, no outline; a press message followed by its release message: no press stands.
- **M2, the prevented press.** `preventDefault()` in the detector's `pointerdown` suppressed the compatibility `mousedown` that moves focus. Removed; a selection is stopped by `user-select:none` on the hovered empty background (the same rule as the open hand, so it holds exactly where a press is a grab) and the shell stops the rest once the slop is crossed. Pinned: a press on the band's SVG focuses `.romp-tl-wrap` (its keyboard road) as at the base; the outline's search field, focused, blurs on a click on the empty list as at the base, and the click leaves no press; a grab that lifts anchors no selection.
- **M3, the base red.** The two legs went red at the base on the detector's missing global. The spot is found by GEOMETRY now (the feed's list below its last card; the band's SVG root outside every lane; the outline's list; the files pane's empty state), the element under the point checked to be the page's own container and never a card, a row, a control or an SVG child, so the base measures the behaviour: no hand and no drag after 14 px there; the hand, the drag, the outline and the drop at the head.
- **L1** (folded): the kit going off removes the detector's tag, its `pd-grab-css` style and its window global from every pane document, so the off page's DOM is what it was (the listeners stay bound once and answer only to the body class; re-injection never binds them twice, a wired flag). Pinned: on, then off through the gear's own event, then on: one tag, style and global; none; one again.
- **L2** (folded): a third context with the Files pane on (`showFilesControl` and `files` seeded): its empty state wears the hand and a press there lifts the pane; the zoom flyout concern is moot with the prevent gone.

**Red first for the round.** At the previous head 3c17aa02 (the engine module and the detector checked out from it): 4 failed, 11 passed. `'romp-tl-wrap' not found in 'BODY pane-docking pd-grab-hover'` (the band press left focus on the body), the outline's search field id equal before and after the click (the field stayed focused through it), `a move with no button held after a stale forwarded press arms nothing: dragging True, pd-drag, outline True`, and the off path `(1, True, True) != (0, False, False)` (the tag, style and global stayed). The feed leg and the files leg pass there (their behaviour existed at 3c17aa02; the files leg is a guard the shipped lab lacked). At the base 387a40ee (the reworked legs, on behaviour): 4 failed (the feed, the band, the outline and the files legs), each on the hand, `'auto' != 'grab'`; no detector and no hand there, the drag never arms.

**At the head**:
- `tsc --noEmit`: exit 0. The node test files: 53 tests, 53 pass. The lab under REQUIRE (chromium) with `tests/test_settings_tabs_recut_browser.py`: 28 passed (15 and 13); the lab again after the driver's hook read was made defensive: 15 passed.
- The whole extension suite (CI's glob): 5142 tests, 5136 pass, 0 fail, 6 skipped.
- The whole Python suite stands from PR 1884's second round: this round changes the detector, the engine and the lab.


ksarma:
A note after the merge. The evidence below was executed against f3c8a2ee2, the merged head, under Chromium (the two plan points and the docstring point are read from that tree); file and line references are on that commit.

- **A selection vanishes while the pointer rests on empty background** (ui/webview/pane-grab.ts:62). The hover rule puts `user-select:none` on the pane document's body whenever the pointer is over an empty target, and every card's text inherits it, so a standing selection loses its highlight and copies nothing while the pointer rests there. Dropping the declaration is sufficient (the shell clears the ranges and makes the frame root unselectable at the arm, panedock-main.ts:503 and :504); skip the hover toggle too while a button is held, or the hand shows mid-selection. A selectstart cancel during the press is optional and costs something: a click on the empty spot then keeps a selection the kit-off page clears. The replacement must not be another toggled inherited body property (see the recalculation point).

- **The closed hand never shows during an empty-space drag** (ui/webview/panedock-main.ts:502). The arm adds `pd-drag` to the shell body only, while the pointer stays captured in the pane document whose body still wears the hover class, so the open hand shows for the whole drag. Fix: write the closed hand on the pane document's body at the arm, an inline cursor cleared in `cancelPress` or a shell-owned body class whose rule follows the hover rule in the detector's stylesheet (source order decides at equal specificity); removing the hover class does nothing, since the captured pointermove re-adds it, and skipping that toggle while a button is held only freezes the class as it was at the press. The chat pane gets no detector stylesheet, so only the inline write reaches an Option-drag from chat content.

- **A touch press is forwarded** (ui/webview/pane-grab.ts:54 and :82). `forwardable` and the pointerdown gate read only the button and the modifiers, so a one-finger horizontal pan on the band's empty SVG is forwarded and re-docks the band on release as an 11 px column. Only the band reproduces this (its wrap is touch-action pan-y, so no pointercancel fires); a vertical pan there, or any pan on the feed or outline lists, is green with or without a gate. Fix: refuse in the detector's pointerdown before `setPointerCapture` and the post; a shell-side refusal alone leaves the page's capture on the touch pointer. Refusing all but the mouse also refuses the pen, which arms under a touch-only gate; the plan speaks only to touch, so the choice needs a sentence. A lab leg must record that the frame saw a touch pointerdown, so a green proves a refusal, not an absent event.

- **The Escape legs cannot tell Escape from the release** (tests/test_pane_docking_served.py:264). The band leg's only cancel read comes after the release, which lands inside the band where `zoneAt` yields no zone, so the press is cancelled without a drop: with the Escape branch gated off or the key wiring removed, only test_5 goes red. The files leg (:359) and the outline leg (:697) copy the shape. Fix: read dragging and the outline's on class between the Escape press and the release, asserting both false in the three tests; a store comparison alone does not discriminate, and a keydown-only regression stays invisible even then, since `onKey` also cancels on keyup.

- **The release message cancels any unarmed press** (ui/webview/panedock-main.ts:410 to 413). The `paneGrabEnd` branch cancels any standing unarmed press without reading the message's source window or pointer id, so a release message from any frame ends a ring or Option press: a click on the feed's empty spot, then a ring press before the click's release message lands, about 35 ms later on an idle page. The consequence is a lost grab, never a wrong drop. Fix: cancel only when `e.source` equals the press's window and the message's pointer id equals one the press carries (numeric, or an Option press with none matches undefined). The mouse's id is always 1, so a stale release can still end an Option press begun in the same frame within the message latency; a per-press token echoed in both messages closes that.

- **The no-button guard reads the wrong pointer and the wrong bit** (ui/webview/panedock-main.ts:497). The guard reads `e.buttons === 0` of whatever pointer's move reaches the wired windows, and the press records no pointer id. A second pointer with no button, usually a hovering pen, cancels a mouse drag; and a chorded release (primary up, a secondary still down) arrives as a pointermove with buttons 2, so the press stands, the next move arms, and the right button's release drops the pane. Fix: record the press's pointer id, ignore other pointers in the move and up handlers and the pointercancel listener at :478, and test the primary bit with `(e.buttons & 1) === 0` on the press's own pointer. A pen hover can revoke the mouse's capture on an unarmed ring press, so the under-slop case after a capture loss needs its own decision; the filter is safe once armed. The stale leg then posts the mouse's id, not 77.

- **The release message runs under no test** (tests/test_pane_docking_served.py:299). The stale leg posts both messages itself and no leg records what a real press and release send: with the detector's post deleted, the 15 lab tests stay green. Fix: a shell-window message listener around one real press and release on an empty spot, asserting exactly grab then end with one pointer id, filtered to values beginning with `paneGrab`; the node test cannot carry this (`install` never runs there).

## Smaller points

- **Scrollbars wear the hand** (ui/webview/pane-grab.ts:76 to 84). The hover toggle and the press gate read only the event target, and a scroll container is the target on its own scrollbar, so over the feed and outline lists' scrollbars the body wears the hover class and a press posts a grab while the browser scrolls, though the shell never arms; fix in both listeners: treat a point outside the target's client area as not empty (`e.offsetX >= t.clientWidth || e.offsetY >= t.clientHeight`, left-to-right pages only).

- **Each card-to-gap crossing recalculates the whole pane document** (ui/webview/pane-grab.ts:77 and :62). The hand rides a body-class toggle of an inherited property per pointermove, and the `user-select` declaration made each flip costlier (35 to 190 ms median at 1500 cards); dropping the declaration restores the earlier cost, and a static rule while the kit is on removes the per-move cost, provided every selector is prefixed with `body.pane-docking` and the containers' non-card children (the session and column headers) get a counter-rule.

- **The plan still says the off page carries no detector** (plans/pane-docking.md:209). After a live switch-off the detector's three document listeners stay bound and `__rompPaneGrabWired` stays true, so either reword, or give the detector an uninstall that `unwire` calls before deleting the global; test_17 should read the flag plus a listener count.

- **The plan's protocol sentence names one message** (plans/pane-docking.md:218). The shell now cancels on the release message (panedock-main.ts:410 to 415), yet the sentence names the grab message alone and omits the `app` field (pane-grab.ts:95).

- **The lab docstring says two contexts** (tests/test_pane_docking_served.py:4). Later commits added a third, the Files pane on, that test_18 reads; the new paragraph should say what that test asserts.

## Tests

The stale leg's store clause (tests/test_pane_docking_served.py:707) compares the layout with itself, and the leg never asserts the stale press stood, so a refused grab message leaves it green; snapshot the store inside the leg before its first post, compare after with a truthiness guard, and assert `afterMessage.pressed` is True. The post-drag selection read (:657) passes in every build because the arm removes all ranges, so it belongs right after the press, under the slop, reading the range, not its text (a caret's text is empty in every build): count 0 under the present rule, a collapsed range once the declaration goes; and no leg reads `hasPointerCapture` after a forwarded press. Neither the outline leg (:692) nor the files leg (:727) asserts the spot's id; one-line id pins turn an accidental red into a stated one; and test_17 (:720) counts nothing that changes when the listeners bind twice: with `__rompPaneGrabWired` removed, 15 of 15 pass while one press after the cycle posts two grab and two release messages.

PR #1899, opened 2026-09-19
Pane docking: the grab ring is six px and the hand closes on the press itself

henrypinkard:
Tier: feature

Item (2) of the user's three: on the docking kit's ring the hand showed but the first press did nothing until later attempts. Reproduced on the built page and pinned; its own pull request on the plan's seam. Filed unarmed.

## What the built page showed

- **A press without travel does nothing, by design and correctly.** The 4 px slop keeps a click a click (section 3: a cancelled or too-short press leaves the layout untouched, with no flash of zones). The lab's own probe: 2 px of travel arms nothing and the store is untouched; 14 px lifts. What was missing is any answer to the press itself: the hand stayed open while the pointer was down, so a press that had registered looked exactly like one that had not.
- **The 3 px ring is a target a pointer lands on only by aiming.** At a normal pointer speed the press lands a hair inside the iframe (content, no grab) or on the divider (resize). The hand showed the ring existed; it did not make it easy to land on.
- **The first press after a fresh load arms.** The engine binds its listeners when its bundle runs at the end of the body, the same moment the ring's hand appears (the stylesheet is injected by the same `start()`); there is no late binding. The lab's first pointer action after the engine reports on is a press on the feed's ring, and it lifts at 14 px.

## The two changes

- **The ring is six px** (`pane-dock.ts` `RING`, the one constant the stylesheet's inset and iframe size, the strip offset and the lab read): a target a pointer lands on without aiming, still the pane's margin rather than chrome. The alternative the read named, making the whole gap between panes the surface, is the divider's space (resize), and the plan leaves open which of a gutter's two panes a gutter drag would move, so the ring it is.
- **The hand closes on the press itself**: `body.pane-docking .pane:active{cursor:grabbing}`, before any travel, so the press is answered the moment it lands. The lift still waits for the slop; a release without travel still changes nothing.

## Proof

Served lab (`tests/test_pane_docking_served.py`, real pointer events): the pressed pane's computed cursor is `grabbing` before the slop while nothing has lifted; 2 px of travel is a click and the store is untouched; every iframe is its pane inset by the six px ring after the seed and after each of the four drops (the ring is a named constant in the lab, never a literal); the first press after a fresh load lifts at 14 px. Node: `pane-dock.test.ts` pins the ring at six.

**Red first at the base 387a40ee** (the two engine modules checked out from the base, the head's tests):
- `ui/webview/pane-dock.test.ts`: the ring pin red (`expected 6, actual 3`), 12 pass.
- `tests/test_pane_docking_served.py` under REQUIRE: 4 failed, 5 passed. The iframe pins read the three px ring (`chat-pane iframe left (after the seed): 3 vs 6`, after drop 1, with the column open); the press pin reads the open hand while pressed (`'grab' != 'grabbing' : the pressed pane wears the closed hand before the slop`).

**At the head**:
- `tsc --noEmit`: exit 0. The four node test files: 47 tests, 47 pass. The lab under REQUIRE (chromium) with `tests/test_settings_tabs_recut_browser.py`: 22 passed.
- The whole extension suite (CI's glob): 5136 tests, 5130 pass, 0 fail, 6 skipped.
- The whole Python suite stands from PR 1884's second round (14336 passed, 0 failed): this pull request changes no Python beyond the lab.


ksarma:
A note after the merge. Evidence executed against 9e42c9468, the merged head, on Chromium unless another engine is named; file and line references are on that commit.

- **The press cue never fires on the plan's Option grab road.** The new rule `body.pane-docking .pane:active{cursor:grabbing}` (ui/webview/panedock-main.ts:93) and the later Option rule at line 96 share specificity (0,3,1), so source order decides: with Alt down a ring press reads the open hand. The drag rule at line 94 loses to line 96 the same way once armed, a half that predates this change; reordering line 93 alone closes the hand on press and reopens it at the slop.

- **`:active` is a proxy for the accepted press, and the two disagree both ways.** On Firefox 153 a pressed ring never matches it, since the preventDefault on pointerdown (line 402) stops Gecko activating, so the hand closes only once the pane lifts, the symptom this change targets. On Chromium 151 and WebKit 26.5 a right or middle ring press reads the closed hand while `grabbable` (ui/webview/pane-dock.ts:118, button 0 only) arms nothing; not on Firefox, and only while the pointer is still, the state clearing at the first move. Chromium's active chain climbs through the iframe: a content or strip press computes the closed hand on the shell pane though no pointer event reaches the shell; whether it paints is unmeasured, the iframe under the pointer selecting the cursor, and no test pins the cursor under a refused press.

- **After Escape cancels a lifted drag, the pressed pane keeps the closed hand until release.** `cancelPress` (line 490) removes only the drag class, so the press rule outlives it: the pane reads the closed hand with nothing armed, where the base reads open; cosmetic, the store unchanged. A pin on the pressed pane's cursor before release in the lab's Escape branch (tests/test_pane_docking_served.py:131) is red at the head.

One fix covers all three: a class on the pressed pane, set in `beginPress` (line 426, after the gate, on both roads) and removed at the cancel, its rule replacing line 93 and sitting, with the line 94 drag rule, after line 96, where both tie. A sheet reorder covers only the first item. A press inside a pane document shows that document's cursor until the iframes go pointer-transparent, beyond any shell rule: it needs a rule in the injected `PANE_CSS` and, under Option, the pressed frame's root cursor set at the press and restored to the Alt state, not blank, at the cancel.

## Smaller points

- Two lab docstrings lag: `_frames_fill`'s (line 366) still says the iframe is inset by the 3 px ring while its assertions read the constant (6, line 41; naming it cannot drift, a spelled-out six can), and the module docstring (lines 6 to 8) omits the pre-slop closed hand test 5 now pins.

- The body says the lab's first pointer action is a ring press, but the content press (`o.contentPress`, line 182) precedes it (line 192), so nothing pins a first press after a fresh load arming. The order predates this change; a fresh context whose first action is a ring press armed at 14 px, and that pin or a reword closes it, while reordering breaks test 8's seed comparison.

- The hand reopening on release is unpinned: the click probe (line 210) and the drag record (line 133) read no cursor after release, and a mutated engine leaving a stuck closed-hand class passed every test; the cue being stateless at the head, this guards the class fix, not a present defect.

- The six px ring and the press cue reach the Sessions band (`#tl-pane` carries the pane class, kernel/kernel.py:64687): its ring doubled and a ring press now closes the hand. The band item in the note on 1884, the fixed-slot write, is byte-for-byte unchanged; what changed, unsaid in the body, is the size and cue of one road into it; a body sentence is the fix; refusing the band in the shell handler alone is incomplete: its iframe also accepts an Option press, and an id cursor rule would outrank the drag rule.

PR #1909, opened 2026-09-19
A session's first whole chat frame seeds the shared delta baseline, so the pushes after it send tails instead of whole-session repaints

lczh:
Tier: fix

## The defect

The chat wire diffs every send against one shared baseline per session, `_prev_chat_events`, and that baseline had exactly one writer: the pusher's non-connect cycle, after the cycle's sends for the session (the write-after-deliver rule #1870 settled). The connect push, which the ready arm and every needFull repair run, and the targeted push (`_push_session_now`: the SDK connect handshake, the Codex backend's stream events, a create, a fork, a promote) never wrote it. So a session whose first whole frame since the boot came from either of those roads left its clients holding a base and the kernel holding no baseline. `_chat_diff` reads 0 with no baseline, and in the uuid-anchored sender a change of 0 against a held base is the whole-session road: counted `changeAt0`, filed as a `chatFull` client-diag row with both edges held, and treated by the page as a reconnect repair (the window rebuilt, the reader's place re-derived). Every targeted push for the session did this until a pusher cycle had built it, and the cycle's own first build did it once more before writing the baseline.

Measured live on the merge base with 22 sessions and a dashboard connected: 42 whole-session frames to caught-up pages in the three minutes after a kernel restart, each with `changeFrom` 0 and `firstHeld` and `lastHeld` true, and zero after minute three. Every session's handshake push and the pusher's first cold build of every tab the page's idle prefetch released land in that window. The same shape recurs off-boot on every tab released from a skeleton on a kernel that has not built it, and on a created session once it has content (until then its frames are `noBase`, since an empty list records no base).

This follows #1868 (the reason meter and the `chatFull` row that made the shape countable) and #1870 (the targeted push routed through the pusher's per-client road, and write-after-deliver for the cycle's baseline write); the review of #1870 named the residual this closes, a change of 0 reached only when the session has no baseline.

## The fix

One helper, `_seed_chat_baseline(sid, m, seen)`, called by the connect push (an `else` branch on the cycle's existing `if not connect:` write, whose two lines are unchanged) and by `_push_session_now` after its send loop. `seen` is the baseline the sender diffed against before its sends, read once. If `seen` was absent or empty and the map is still so, the list the sender handed out whole becomes the baseline, along with its ledger; an empty list never does, since an empty frame records no base on any client. If `seen` was present, nothing happens.

Seeding is not advancing. The 2026-07-28 stranded-delta rule says only a push that reaches every client may ADVANCE the baseline, because a connect push that moved it left every other client behind the next diff's change point. The seed never moves a present baseline: it fires only when none exists, and the deciding event is the send itself. A build was handed to clients whole while the kernel held no baseline, so from that instant that list is a lower bound on every base holder's state; no client can hold a base for a session with no baseline except one it was handed in that send. `test_19` pins that a connect push over a grown list, with a baseline present, leaves it byte for byte, and the `test_chat_delta_resync` source pin on the gated write still holds.

The race detector, and the mark it leaves. Two whole-frame senders can both read the baseline absent before either loop runs (a targeted push and a connect push for a second page; a create's push and its handshake's), and per-client delivery order is whichever thread reached each client's lock first, so one client can end holding the OLDER build. A seed that only wrote when absent kept the newer list as the baseline, and the next cycle diffed equal lists and never re-sent the stale card. The seed reads the map back against `seen`: absent then and a DIFFERENT non-empty list now (by length, or a diff that lands before the end) means a racing sender wrote, no one list describes every base holder, and the entry is popped, so the next cycle's whole frame repairs every client as it did before the seed. `test_23` and `test_24` drive the two orderings through the real senders. The pop alone was not enough (the review): a popped entry leaves clients holding bases with no baseline, the state a never-seeded session is in too, and the next single-client connect push (a needFull, or the idle prefetch releasing any tab on any page: one such push per released tab at a boot, while a cold cycle takes 30-84 s) read the map absent, repaired its own client with a `changeAt0` full and seeded from its build, so the cycle diffed equal lists and handed the other holder of the older build an empty-suffix tail: a card that filled between the two racing builds stayed stale until a reconnect, with no row, where the merge base self-healed through the cycle's full. So the pop also marks the session (`_chat_baseline_raced`); no seed writes while the mark stands, and the cycle's write-after-deliver clears it with the write, as one step (the strip-exit eviction clears it too, since no client holds a base once it has run). The contract: a raced pop is repaired by the next sender that reaches every client, the cycle, and is never re-seeded by a single-client push in between; until that cycle every base holder is served the full, as before the seed. The targeted push's loop reaches every alive chat client too, but its seed declines while the mark stands all the same: a single-client sender that read the map absent may still be mid-flight, its older full landing on some client after the targeted push's and its own seed already declined, so only the cycle's unconditional write re-establishes the baseline (the residual the cycle's write carries today, not widened). `test_25` drives the race, one connect push for one client, then the cycle. The seed's read-back and write are one step under a small module-level lock (`_chat_baseline_lock`, nothing else held inside; the detector's pop, the cycle's write and the eviction take it too): unlocked, two seeds that both read the map absent both wrote, the last writer won and the detector never fired (`test_26`, two seeds over the real function with the first parked between its read and its write).

The eviction pop keeps the invariant true after a tab leaves the strip. The pusher popped the baseline of a tab the strip stopped listing, but only for a session its build cache held, and left every client's `echat` base in place; with the seed, a re-entering tab's first whole frame would have established a baseline over clients holding an older base, the 2026-07-28 shape. The page tears a tab down when the strip stops listing it (`applyTabOrder`), so the kernel's belief that a long-lived socket still holds it is false from that moment. The eviction now walks the union of the build cache and the baseline map, and for every gone session pops each alive chat client's base and `("chat", sid)` dedup slot under that client's lock, after snapshotting the client list under `_clients_lock` and releasing it. The eviction walks the detector's marks too and clears them, so a tab that left is a never-seeded one again (no client's base, no baseline, no mark), and a re-entering tab is a `noBase` full for everyone, filed nowhere.

Prose: `_chat_full_reason`'s `changeAt0` gloss and the `chatFullWhy` gloss in `docs/reference.md` now name the seed and the racing residual instead of a no-baseline send by whichever sender, `noBase` gains the re-entered tab, and the `_push_session_now` docstring is rewritten around the seed, the detector and its mark. No wire change: frames, labels and the delta contract are as before; the seed changes only what the NEXT sender diffs against.

## Tests

Red first, on the merge base, in `tests/test_chat_skeleton_reconnect.py` (the harness that drives the real `_push` and `_push_session_now` over synthetic builds): `test_18`, the boot shape, with a status flip before each step so no frame dedups; nine of its twelve assertions are red alone on the stock kernel (the connect push writes no baseline; the handshake and the cycle's first build each send a whole frame, `chatFullWhy` reads `changeAt0: 2` with two `chatFull` rows); the other three are guards that hold on both. `test_20`, content replaces an empty baseline (red: the baseline stays `[]` and the targeted push sends a full). `test_21`, a tab that left the strip loses every client's base and the baseline, with no cache entry (red on all three counts). `test_22`, an empty build after a seeded full takes the stand-in road (red: the empty frame goes, counted `empty`). `test_19` is the never-advances guard, green before and after. `test_23` and `test_24`, the two racing orderings, are green on the stock kernel (which never seeds and repairs at the next cycle) and red under the plain seed without the detector (the newer list kept while one client holds the older card, and the cycle sends no repair); the detector makes them green. `test_25`, the race, then one connect push for one client, then the cycle: red on the detector without the mark (the connect push re-seeded the popped baseline with the filled list, and the cycle then left the other holder's last u3 card at the pending text, with no repair frame and no row), green with it. `test_26`, two seeds over the real function with the first parked between its read-back and its write: red on the unlocked seed (the second did not wait, both wrote, nothing popped, no mark) and green under the lock. In `tests/test_kernel_opening.py`, `test_the_shared_delta_baseline_is_never_advanced_and_seeded_when_absent` (renamed from the left-alone test, red by design: with no baseline the list sent whole becomes it; with one set by hand it is left exactly where it was), and `test_idempotent_with_the_periodic_pusher` now pins the seeded shape: a repeated targeted push on an unchanged build sends one empty-suffix tail (the status-only frame every cycle sends after a full) and the third dedups, where the unseeded kernel re-sent the whole session and relied on the dedup to swallow it; the re-pin fails on the merge base at the `n + 1` assertion (`2 != 3`: with no baseline the second push's diff reads 0, the index wire's tail needs a change past 0, and the repeated full dedups).

Green after, one module at a time: `test_chat_skeleton_reconnect` 48, `test_kernel_opening` 7, `test_chat_delta_resync` 12, `test_chat_pages` 43 (27 subtests), `test_perf_stats` 65, `test_chat_empty_build_guard` 5, `test_needfull_no_dedup` 1, `test_kernel_delta_send` 13, `test_cold_tab_gate` 22 (4 subtests), `test_chat_build_sig_inputs` 65 (10 subtests), `test_feed_memo_inputs` 31, `test_chat_fixed_cost_memos` 44, `test_git_pointer_file_bytes` 9, `test_kernel_goal_cache_wiring` 17, `test_outline_provisional_rows` 10, `test_pusher_change_gates` 13, `test_single_flight_builds` 11, `test_codex_models_frame` 8, `test_comment_create_idempotent` 14, `test_comment_name_refusal_log` 4, `test_comment_threads` 115, `test_kernel_create_session_ack` 4, `test_kernel_fork` 19, `test_kernel_names` 55, `test_kernel_tabs_first` 7, `test_per_viewer_focus` 16, `test_stage_marks` 31, `test_kernel` 460 (12 subtests), `test_perf_bench` 49: 1198 passed across the 25 modules that name either `_prev_chat_events` or `_push_session_now`, plus `test_kernel`, `test_kernel_delta_send`, the feed-memo input census and the bench (29 modules). The extension suite (which pins `_push_session_now` source lines): 5136 tests, 5129 pass, the one failure the pre-existing KaTeX pixel probe. The static pin scan over the branch: 8537 assertIn literals checked, 0 lost.

## Post-deploy read

The primary meter is the client-diag `chatFull` rows filtered on reason `changeAt0` with `firstHeld` and `lastHeld` true and `changeFrom` 0 in the minutes after a restart with a dashboard connected: 42 on the merge base, expected 0 or near it. Any that remain name the racing-senders residual (two targeted builds for the session within one build time in the perf log's `chatbuild` lines: the cycle's repair, and any single-client push's full to a base holder between the race and that cycle) or a tab that left and re-entered the strip between a targeted full and the cycle's eviction. Bare `pusher.chatFullWhy.changeAt0` on `/perf` is coarser: a floor advance that moves the list's first event lands there too, with `firstHeld` false in its row. `noBase` climbs by one per client per tab served whole, as before, plus the re-entered tabs; a created session's frames stay `noBase` until its first content lands, since an empty list records no base. `sends.full.chat.targeted.count` should fall to about the creates and forks while `sends.delta.chat.targeted.count` rises by the handshake pushes now riding as empty-suffix tails; `needFull` and full-frame-desync rows must not rise (a rise would mean a tail anchored past what a page holds). Left open, as one redundant full per new session: the cycle's `[]` write over a seed for a session whose transcript is not there yet.


ksarma:
A note after the merge: the evidence was executed against 120570c95, the merged head, and every file and line below refers to it.

- The seed's lower bound (kernel/kernel.py:33602). A whole-frame sender whose build read the transcript before another sender's, but whose `_seen` read lands after that sender's seed, sends its client the older list as a tail or a full; the next cycle diffs equal lists, so the client keeps a stale card with no row and no mark (the detector sees only absent-then-different), which the merge base's no-baseline cycle repaired. The same strand exists on the merge base whenever a baseline is present and the cycle is the newer writer, so the head's regression is a lost accidental repair in the no-baseline windows (a sid before its first cycle write, a tab re-entering after the eviction, a transcript-less build until content), not a new class. A stamp taken before the build and stored beside the seed, a sender whose build began before it taking the detector's road, restores the repair; the class closes only if every present baseline carries one, the cycle's write included. Moving both `_seen` reads before the build repaired both shapes with 72 tests green; test_28 pins the interleaving either way.

- The seed runs for a sender that delivered no whole frame (kernel/kernel.py:54166 and 54715). A skeleton client's status-only frame or a withheld frame seeds from a list no client has; with another sender's list in the map the detector pops and marks, a false race whose repair is the next cycle's changeAt0 full with a row to every base holder, the frame this change set out to remove; the docstring's premise at 33575 is false here. The single-flight claim closes the cycle-as-writer pairing, leaving a targeted push with a connect push landing in its send window; a cold skeleton tab with a live row never reaches the seed, so the gap bites on cached or live-row-less tabs. A newer status-only seeder also strands a concurrent whole-frame sender's client on the older card with no row. Seeding only on delivery (pinned by a lone skeleton client whose baseline stays absent after the connect push and after `_push_session_now`) is independent of the fix above; both are needed.

- The eviction's walk of the marks and its clear (kernel/kernel.py:54193 and 54198), which the body, the seed docstring and the eviction comment all claim, has no test: removing both leaves the changed modules green. Without it a re-entering marked sid costs each base holder a full with a row instead of a tail when the list moved (unchanged, it dedups) until that cycle's write clears the mark, and a sid that never re-enters keeps its entry for the kernel's life. Amending test_21 catches only the discard, since S3 is found through the baseline map; the union member needs a test_28 whose marked sid has neither a cache entry nor a baseline.

- Three texts predate the second commit's guard: docs/reference.md:2074 and the docstring at kernel/kernel.py:33597 still say the next cycle's write unconditionally repairs every client and clears the mark, and the body's Fix section calls the cycle's two write lines unchanged. At the head (54151 to 54156) the cycle skips both the write and the clear when it read the baseline present and the sid was marked, as test_27 pins. The lock and the discard came with the first commit and only the guard with the second, so the body's sentence was stale from the first. The reword: the clear is earned by a loop that read the baseline absent or empty; a tails-only cycle leaves the mark for the next one.

### Smaller points

- Empty builds over a popped baseline (kernel/kernel.py:54096 and 54700). After a race pop, or a tails-only cycle that skipped its write, the sid has no baseline while every client has content, so a failed transcript read before the next cycle sends every base holder an empty session frame, counted `empty` with a row each and no stderr line, where a seeded sid takes the stand-in road test_22 pins. The pane does not blank, since `keepResidentEvents` absorbs the frame as status-shaped; the cost is meter noise, the lost stderr line and one redundant full when content returns. The base had the same hole before any sid's first cycle, so the head narrows the window overall; only the post-cycle state is new. Treating a marked sid as one whose clients have content at both sites closes it, with `_note_empty_build` handed the true prior count.

- The module cannot run on the merge base (tests/test_chat_skeleton_reconnect.py:138 and 150): the two clears of `_chat_baseline_raced` are unguarded, so 48 of 49 tests error in setUp there (RestartDiet has no setUp). Guarded with `getattr`, the base run shows the body's picture, except that 25 to 27 are red by AttributeError, not by the reds the body describes for them; a body sentence saying the before-shapes were taken guarded covers it.

- Two seed lines delete without a red (kernel/kernel.py:33610 and 33611). Every race fixture differs at equal length, so the length arm never decides; the arm is defensive today, since a tail anchored at the held last heals that shape, and a race that appends u5 with the shorter full landing last pins it. Every fixture's build carries None in the field the seed copies into `_prev_chat_ledger`, so that write is untested; a non-None value for S1 in test_18, read back after the connect push and absent from the handshake's tail, pins it.

- The pop-and-mark is unmetered (kernel/kernel.py:33614): the only trace is the next cycle's changeAt0 rows, filed only when a base holder is alive then and its repair does not dedup, and shaped like a genuine first-event change. A `_PERF_STATS` counter at the pop (the primary meter, since perf logging is off by default) and one where the cycle's write clears a standing mark, decided under the lock and emitted after it, would make both countable.

- The eviction's premise (kernel/kernel.py:54200; docs/reference.md:2063). The page keeps a tab the strip omits when the frame's live field lists it, filing a live-omitted-kept row, and the alive list omits a live sid whose stub is None (no names entry, no transcript, no SDK owner), so the head pops every client's base while the page still has the session, and re-listing sends a row-less noBase full where the base sent a named changeAt0 full with a row, or nothing when the list was unchanged and the identical full deduped. The corner is narrow, so stating the exception in the comment and the gloss is the proportionate fix.

- The body's 48 for the module is 49 at the head (test_27, tests/test_chat_skeleton_reconnect.py:1525); the total is 1199.

PR #1910, opened 2026-09-19
Settings across machines, phase two: a machine selector above the tabs, the disagreement flag, and a scoped change as a pin

henrypinkard:
Phase two of plans/settings-across-machines.md, as the section merged in PR 1903 lays it out (the user's ask of 2026-09-19): a MACHINE SELECTOR at the top of the settings card above the tabs when more than one kernel is connected, a FLAG whenever the kernels disagree on a setting, and a scoped change that pins the picked kernels. Phase one A (PR 1859) and one B (PR 1892) stand under it unchanged.

**All kernels, the synchronized view.** The selector reads "All kernels" by default and the card is the card of today: every control shows this kernel's value, a click broadcasts as the gear's broadcast does (each copy stamped with its origin, subject to each machine's pin). A synchronized row whose connected kernels disagree wears the flag: the `rs-mixed` slot the gear already computes from the `/tunnels` rows reads "differs" in the heads-up amber (`--warn`, through the token) instead of the quiet grey "mixed", its hover names the machines and their values, and the selector's button carries the COUNT of differing rows ("1 differs"), so the disagreement is seen from the top of the card before any tab is opened. The other kernel-side controls (the judge tiers, phase three) keep the quiet mark.

**A scoped change is a pin.** Picking one or several kernels in the selector's list (checkboxes; "All kernels" resets) scopes the four synchronized rows to the picked kernels' values (a remote's `settings` from its `/tunnels` row, this machine's from `/version`; indeterminate when the picked kernels differ). A change there posts the message with `hosts` (the picked kernels) and federation routes it to those kernels alone, each copy with its origin and `scope: "pinned"`; the pin gate (`_pinned_stand_down`) passes a scoped remote-origin gesture, the setter applies under the gesture's stamp, and the arm pins the store under that same stamp (`_scoped_pin_after`, through the pin store's own gt gate). A broadcast never carries the scope: the fan-out strips a stray one, so nothing changes for a kernel the selector never named. Each synchronized row wears a pin glyph at its right edge (an inline pushpin, shown only with more than one kernel), lit while the picked kernel (this machine under All) pins the store, its hover naming the pinned value here beside the other machines' values; a click toggles `setSettingPin` on the picked kernels (`hosts`, the scope; federation addresses each copy with its origin, so a remote arm accepts the explicit pin or un-pin), and the pinned NOTE that stood under a pinned row folded into it.

**The un-pin returns the row to the synchronized value, from the records.** As the design's read asked: `_unpin_to_synchronized` reads this kernel's own per-(store, machine) records kept by `_propose_peer_settings`, never a dial at click time. A pinned store raised no proposal, so it kept no record; it now keeps the peer's newer value as a HELD record (`held: true`): no card, not in the gear's map, not pending, not answerable, and no notice; when the pin lifts, the next poll turns it into a proposal with its card, said once. The un-pin applies the newest held or open record whose value differs, under that machine's stamp, through the store's own setter with the local origin, so the two machines then hold one stamp and nothing is proposed back; with no newer different value elsewhere the local value stands.

**Around it.** The `/tunnels` row carries `settingsGt` and `settingsPinned` beside `settings` (the supervisor keeps the peer's pins on the row from the same poll). The strip's Sessions menu marks a machine that pins any store (a quiet "pinned" chip on the host's row, the stores on hover). The selector and the glyphs hide with one kernel, so the settings look exactly as today. After a scoped change or a pin click the gear re-reads the kernel through the supervisor's poll cadence (seconds), and its fill says a fault on the console instead of swallowing it (a swallowed fault hid the selector while the lab was built).

**Tests, red first at 5b254f2d** (main with the design merged; the test files copied into a detached worktree then restored) under `--python 3.13` and under `3.14` the same: `13 failed, 123 passed` over the mesh module, the wiring pins and the new two-kernel lab (the require flag on), and the node pins `24 tests, 18 pass, 6 fail`. Executed: the scoped remote pin `{} != {'task-tracking': True}` (a remote-origin pin with the scope pinned nothing at the base); the held record `KeyError: 'task-tracking'` (a pinned store kept no record of the peer's newer value); the two-kernel lab's drives (the base's gear has no selector: `True is not false : two kernels: the selector shows`, and the scoped drives red on the missing states); the scoped gesture through the arm `Tuples differ: (True, 1000) != (False, 5001)` (the base stood it down); the tunnels row `({'taskTracking': False}, None, None) != (...)` (the base row serves `settings` alone); the gate's and the pin arm's new text, said as such; the federation and gear pins (the scoped fan-out, the addressed pin, the selector, the badge and the glyph) and the compact-suggest handler's `scoped()` wrapper. Green at the head: the mesh, wiring, task-tracking and sections modules `180 passed`; the two-kernel lab `4 passed` under the require flag (the second kernel checked in to the first, the selector hidden then shown, `1 differs`, the flag in the amber, a scoped click applying and pinning on the peer alone, the un-pin, All kernels applying to both); the federation, gear, strip and vocabulary pins `164 pass`; the whole Python suite under `--python 3.13`, devbox, capped, no xdist, `14566 passed, 56 skipped, 2337 subtests passed in 2227.67s (0:37:07)`; the whole extension suite `5138 tests, 5130 pass, 2 fail, 6 skipped`, the two being the compact-suggest and task-tracking pins on the handlers' post text (now wrapped in `scoped()`), re-pointed, and the three modules `16 pass` after; both test trees grepped for every changed literal.

Mixed mesh: a kernel from before this change ignores `scope` (its gate stands a remote-origin gesture down on a pinned store as in one A, and its pin arm takes the local origin alone), so a scoped change to such a kernel is refused the way a broadcast is; the reference says so.

**Round two (the verifier's finds on a54a891a, folded at a7b45cd8).** HIGH: a peer's PINNED store raised a proposal here on the next poll (from A the user scoped B and flipped a store; B applied and pinned; A read B's newer stamp and posted the card whose Apply would undo the scoping). `_propose_peer_settings` now reads the peer's `settingsPinned` as the push leg always did: a pinned store raises no record, no card, no notice, and an open record from it drops; the "differs" flag is the surface that says the machines disagree. The push leg keeps a store THIS machine pins to itself as well (`_older_peer_settings`: a per-machine value is not the mesh's to receive). The lab gained the leg the first cut missed: a scoped click that makes the kernels DISAGREE (B on and pinned, A off), then A's `/version`, its records file and its owner-less notice file read empty of any proposal while the row wears the flag and the count reads "1 differs". M1: `autoNudge` joined the flagged set, so the row the count named wears the flag. M2: the scoped rows leave a box alone when the scope is off (the Auto Nudge tri-state stands). M3: every setting click re-reads the kernels through the poll's cadence, so the count follows a resolved disagreement. M4: a store this machine pins keeps its glyph whatever the kernel count, so the un-pin is one click away with one kernel too. M5: the pinned mark is on the browser dashboard's Remote kernels popover (`pinMark`, the quiet chip beside the host's name, the stores on hover) as well as the VS Code strip's rows; the lab opens the popover after the scoped pin and reads the mark on B; the reference names the popover. M6: the un-pin test now reds on this round's behaviour: B off@2000 and C on@3000 attached, D off@4000 gone from the mesh; the un-pin takes the NEWEST stamp among ATTACHED machines (C's, the same value, so the local stamp lifts to 3000; the base left (True, 1000)). Lows folded: (a) `_unpin_to_synchronized` reads records of attached machines alone (a row in the remotes); (b) and (c) the reference names the popover and says how a kernel from before phase two meets `scope` (pinned there: refused as a broadcast; unpinned: the value applies but pins nothing, and the next broadcast walks it back); (d) an unnamed answer counts open records alone, so a held record no longer forces "name the machine"; (e) an empty or malformed `hosts` is no scope: the broadcast, the field stripped.

Red first for round two at 5b254f2d (the test files copied then restored) under `--python 3.13` and under `3.14` the same: `17 failed, 122 passed` over the mesh module, the wiring pins and the two-kernel lab (the require flag on); the node pin module `6 tests, 0 pass, 6 fail`. Executed, this round's: the HIGH, `Lists differ: ['task-tracking'] != []` (the base proposed the peer's pinned store) and the real-poll test `['compact-suggest', 'file-editing'] != ['compact-suggest']` (the pinned file-editing proposed too); the held records `KeyError: 'task-tracking'` (the base kept none, so the un-pin test's three-record premise reds before its newest-attached assertion); the lab's disagreement leg `KeyError: 'disagreeApplied'` (the base's gear has no selector, so the driver never reaches the leg); the round-one reds as before (`(True, 1000) != (False, 5001)`, `{} != {'task-tracking': True}`, the tunnels row's `None, None`). Green at the head: the mesh, wiring and task-tracking modules with the two-kernel lab `177 passed` (the lab's five tests under the require flag, the disagreement leg and the popover read included); the federation, gear, strip and vocabulary pins `143 pass`; the whole Python suite under `--python 3.13`, devbox, capped, no xdist, `14569 passed, 56 skipped, 2337 subtests passed in 2235.32s (0:37:15)`; the whole extension suite `5139 tests, 5133 pass, 0 fail, 6 skipped`.

Tier: feature


lczh:
Evidence executed against a54a891a.

The un-pin the body and docs/reference.md promise works only on the kernel that polls; on the polled kernel, the default one-way topology, it keeps the pinned value and the hub's next pass raises the card it exists to spare, so that item is worth fixing before this merges and the rest can land with it.

## At this head

**On the polled kernel an un-pin cannot return to the synchronized value: a pinned peer is never pushed, so it holds no record.** `_unpin_to_synchronized` (kernel/kernel.py:51572) reads the records `_propose_peer_settings` (51799) keeps, and that function runs only from the poller (`_converge_peer_settings`, 51395) and the push receiver (`_apply_mesh_settings`, 51479). The push leg, `_older_peer_settings` (51420), skips a peer whose `settingsPinned` names the store (51436). Executed on a two-kernel pair, hub A polling B one way: A on@1000; scope B, click off@2000 (B off@2000, pinned); All kernels on@3000. A's pass finds `_older_peer_settings` empty and B's records stay `{}`. Un-pin B through the arm (68193): no record, B stays (False, 2000) while A holds (True, 3000). A's next pass pushes on@3000 and B raises an open proposal with the needs-you card "Task tracking: TESTHOSTA proposes on; this machine is off". The served lab passes because `_seed` (tests/test_settings_machine_selector_served.py:134) writes a task-tracking file for B alone, so A's stamp is 0 and nothing newer exists; the "B keeps on" assertion at line 306 holds for that seed only. Let the push leg reach a pinned peer: the receiver's held path (51839-51847) files exactly the record the un-pin needs, and the setter's pin gate still applies nothing. Add the polled-side mesh test (hub newer, un-pin returns to it under the hub's stamp, no proposal) and seed the lab's A with a real stamp.

**Auto Nudge loses its tri-state under All kernels and never wears "differs" while the selector counts it.** In `fill()`'s /tunnels handler (ui/webview/gear.js:2019-2024), `fillAutoNudge` (1789) sets `an.indeterminate = true` on a split (1797), then `fillScopedRows` (1945) resets `indeterminate` on every `PROPOSAL_ROWS` box whenever `scopeOn()` is false (1952). The pair list in `fillMixedMarks` (1811-1816) has no autoNudge, so `#rs-autonudge-split` only ever gets the gray "mixed", while `differing()` (1881) counts all four stores. Executed in headless Chromium over stubbed /version and /tunnels, this machine on and the peer off: `#rs-autonudge` indeterminate true after `fillAutoNudge`, false after `fill()`; `#rs-autonudge-split` "mixed" with no `rs-differs`; `#rs-scope-count` "1 differs". Clear `indeterminate` only for boxes the scoped view set, route Auto Nudge through the same flag as the other three, and pin all four marks against the count.

**An un-pin that flips the value bypasses the arm's follow-ups.** The setTaskTracking arm (68205) marks the views dirty, drops the pure feed, wakes the producer, calls `_push_soon()` and replies the `taskTracking` echo (68218-68224); setAutoNudge and setCompactSuggest fire `_ws_act_now_tick()` (68137, 68156). `_unpin_to_synchronized` flips the same stores through the bare setter (51591). Executed: a local un-pin with a held off@2000 record flipped task tracking on to off with no frame, `_producer_wake` and `_pusher_wake` unset, `_views_dirty` unchanged and `_feed_servable` still true; the setTaskTracking arm on the same kernel fired all of them; an auto-nudge un-pin fired 0 act-now ticks against the arm's 1. The counterpoint holds in part: `_fleet_view_sig` (54883) stats the tracking file (`__tracking__`, 54906), so the panes catch up within REBUILD_MIN_S plus the 0.5 s backstop, and the proposal apply takes the bare setter too. What stays visible is the judge producer's up-to-3 s pass, the shell rail's 30 s /version poll and the missing tick. One follow-up function called from all three paths, with a test that an un-pin wakes the producer.

**`refillSoon` replaces the event with seven timers ending at 21 s.** `refillSoon` (gear.js:1866) arms 700 ms to 21 s (1868), each running `fill()`, 14 fetches per click. The peer's pin reaches the gear only through the local /tunnels row, which `_tunnel_supervisor` (27376) writes from a sequential per-row pass with 4 to 5 s probe timeouts and then `_tunnel_wake.wait(SUPERVISOR_PASS_S)` (27707; 15 s, 27710). From those constants: one peer, a click at pass start, the glyph misses once a probe exceeds 0.88 s; two peers at 1 s per probe, the pin lands at 32 s and about 43 percent of clicks leave the glyph, the box and the count stale until the panel is reopened. The predicate is in the same response: re-read /tunnels until the picked row's `settingsGt[store]` reaches the gesture's stamp, bounded and said in a toast when it does not; drop the ladder.

## Smaller points

- tests/test_settings_machine_selector_served.py:234: `_result` times B's check-in by `time.sleep(6)` against the driver's first `fill()`; `checkin_apply` sets `_tunnel_wake` (kernel.py:22121). Measured: the gear's only alone-phase /tunnels read at 1.7 s, the row up at 6.06 s. A 5 s delay before `page.goto` moved the read to 6.7 s and `test_with_one_kernel_the_selector_is_hidden…` went red ("one kernel: no selector"); a 50 percent CPU quota on one core did the same. The ordering is deliberate, but it is a clock: have the driver post the check-in after the alone survey (line 88 already says `cfg.checkin`, a key cfg.json never carries).
- ui/webview/strip.ts:677-684, strip.css:109: the Sessions menu's "pinned" chip has no test; nothing under ui/webview/*.test.ts, vscode-extension/src or tests/ names `sn-pin` or `pinnedStores`. Executed: the 20 test files that read strip.ts or strip.css pass 181/181 with the chip dropped, the filter inverted and `t.settingsPinned` renamed. A source pin in net-trust-pending.test.ts's shape fits.
- kernel/kernel.py:21132: `_remote_payload_public_row` has no entry for `settingsGt` or `settingsPinned`, so `tunnels_of` (21872) relays every peer row without them while `_remote_public` (22526-22527) emits both; executed, a row through the shaper comes back with neither. The shaper's docstring scopes the whitelist to what the sub-row reads, so no reader loses anything today; still, the census `good` row (tests/test_peer_payloads_as_text.py:111) was not extended.
- tests/test_settings_machine_selector_served.py:167-199: `_boot` kills only its own kernel before `_skip`, and a setUpClass exception skips `tearDownClass`. Executed with B replaced by a process that never binds: SkipTest after 61 s, kernel A alive and answering /healthz 200; `_parent_watch` is a no-op since `kernel_env` passes no ROMP_MANAGER_PID. `addClassCleanup` per kernel and for the lab.

The runs and probes are in a second comment below.

lczh:
The test runs behind the comment above, on a54a891a.

## Tests

- tests/test_mesh_settings_adopt.py alone under this machine's shared pytest lock: 59 passed in 2.49 s; four further in-process runs, 59 passed each.
- tests/test_peer_payloads_as_text.py at the head: 14 passed, the census `good` row unchanged from the base.
- The 20 node test files that read strip.ts or strip.css: 181/181 pristine and with the `sn-pin` block dropped, the filter inverted and `t.settingsPinned` renamed; the whole node bundle with the chip dropped, 5131 pass, 1 fail, the KaTeX layout test that also fails pristine here.
- tests/test_settings_machine_selector_served.py with the driver instrumented: the baseline 4 passed; with a 5 s delay before `page.goto`, with `time.sleep(6)` cut to 0.5 s, and with the driver under systemd-run CPUQuota=50% on one core, `test_with_one_kernel_the_selector_is_hidden…` red on `False is not true : one kernel: no selector`, the other three green each time.

## Two-kernel reproductions

- The un-pin on the polled kernel, three harnesses, hub A polling B one way, A on@1000, B on. Scoped click: B off@2000, pinned. All kernels on@3000: A (True, 3000). Hub pass: `_older_peer_settings` returned `[]`, B's records `{}`. Un-pin B through the arm: B (False, 2000). Hub's next pass pushed on@3000; B raised an open record and `_proposal_card_post` posted "Task tracking: TESTHOSTA proposes on; this machine is off". With A's task-tracking file absent, the lab's seed, no card.
- The un-pin's follow-ups through `Handler._dispatch_ws`: A pinned on, a held record B off@2000; the `setSettingPin` un-pin flipped the store to (False, 2000) with no frame, `_producer_wake` and `_pusher_wake` unset, `_views_dirty` unchanged, `_feed_servable` true; `setTaskTracking` on the same state fired all of them. Auto-nudge: 0 act-now ticks on the un-pin, 1 on the arm.

## Probes

- Auto Nudge in headless Chromium over stubbed /version and /tunnels: after `fill()` `#rs-autonudge` indeterminate false (true right after `fillAutoNudge`), `#rs-autonudge-split` "mixed" without `rs-differs`, `#rs-scope-count` "1 differs"; with task tracking differing too, "2 differ" and one row flagged.
- The refill ladder, a timeline from the code's constants: one peer, a click at pass start, the last read at 21.15 s misses once a probe exceeds 0.88 s; two peers at 1 s per probe, the pin lands at 32 s and about 43 percent of evenly timed clicks end with the glyph unlit.
- The relay shaper: `_remote_public` of a row carrying `settingsGt` and `settingsPinned` emits both; through `tunnels_of`'s stubbed transport, `_remote_payload_public_row` returns it with neither.
- The lab's boot failure: kernel A booted through `_boot`, B a process that never binds its port. After 61 s `_boot("b")` raised SkipTest from `setUpClass`: `cls.b` unset, A alive and answering /healthz 200, `tearDownClass` never run; pytest 9.1.1's unittest class fixture calls `doClassCleanups` and re-raises, the same under ROMP_SERVED_TESTS_REQUIRE=1.

lczh:
Executed against a7b45cd8. Landed, by execution: Auto Nudge under All kernels with the peer off keeps `indeterminate` true and its mark reads "differs" in the warn color beside "1 differs" (`fillMixedMarks`' pair at ui/webview/gear.js:1813); with Task tracking differing too, "2 differ" and both rows flagged. One kernel with a local pin shows the glyph lit (1954-1955). An empty or malformed `hosts` routes as the broadcast (ui/webview/federation.ts:399, 404). A peer's pinned store raises no record here and drops an open one (kernel/kernel.py:51830-51836), pinned by `test_a_peers_pinned_store_raises_no_proposal_here_and_drops_an_open_one` and the lab's leg (6). An unnamed answer counts open records alone (51904-51908).

## At this head

**The un-pin on the polled kernel still cannot return to the synchronized value, and this head makes that structural.** The folded HIGH is a different case. Re-run on the mesh module's pair, hub A polling B one way: A on@1000; scoped click, B off@2000 pinned; All kernels on@3000, A (True, 3000). A's pass: `_older_peer_settings` (51420) skips B at 51436, B's records `{}`. The scoped un-pin through the arm (68222): B stays (False, 2000). A's next pass pushes on@3000 and B raises the open record and the needs-you card, "Task tracking: [the hub] proposes on; this machine is off". New here, `_unpin_to_synchronized` reads records of machines in `_remotes` alone (51588, 51592), and in the one-way topology the polled kernel has no row for its hub: executed, a held record from the hub on a kernel with no row for it, the un-pin returns None; with a row, (True, 3000). The lab's leg (6) never has the hub newer than B's pinned stamp at un-pin time; the exposing order is an All-kernels click after the scoped pin, then the un-pin. docs/reference.md:4384 still promises the return. Let the push leg reach a pinned peer (the receiver's held path files the record, its gate applies nothing) and count a push-filed record as attached: the pusher polls us, so it is in the mesh.

**Under All kernels, a box the scoped view left indeterminate stays that way.** The removed `else if (box) { box.indeterminate = false; }` (1952) also cleared the three plain boxes. Executed: this machine on and the peer off for Task tracking; pick both, the box goes indeterminate; pick All kernels: `#rs-tasktrack.indeterminate` stays true through the fill and a reopen. At a54a891a the branch cleared it. Clear `indeterminate` for the three boxes `fillAutoNudge` does not own; leave `rs-autonudge` to it.

## Smaller points

- The un-pin's follow-ups, unchanged. Through the arm: a local un-pin with a held off@2000 record flipped task tracking on to off with no frame, `_producer_wake` unset, `_views_dirty` unchanged, no `_push_soon`; the setTaskTracking arm fired all four (68236-68242); an auto-nudge un-pin 0 act-now ticks against the arm's 1. `_unpin_to_synchronized` (51574) still takes the bare setter at 51596.
- `refillSoon`'s ladder (1868) stands and now fires on every click: `scoped` (1860) calls it before the scope check. With one kernel, one click: 7 `/version`, 7 `/tunnels`, 7 `/logins`. The predicate is in the same response: the picked row's `settingsGt[store]` reaching the gesture's stamp.
- tests/test_settings_machine_selector_served.py:256 keeps `time.sleep(6)`, and cfg.json still carries no `checkin` key. 5 passed here in 86 s.
- The strip's chip (ui/webview/strip.ts:677-680, strip.css:109) has no test; the popover's `pinMark` (kernel.py:61798) got source pins and the lab's read, the strip did not.
- `_remote_payload_public_row` (21132) still emits neither `settingsGt` nor `settingsPinned`: a row carrying both comes back without them, while `_remote_public` (22526-22527) emits both; the census row is unchanged.
- `_boot` (189) still kills its own kernel alone and the module has no `addClassCleanup`: a failed second boot leaves the first running.

ksarma:
This is a note after the merge. The evidence was executed against a7b45cd83, the merged head; every line reference is on that commit.

- **A kernel from before phase two that pins a store still pushes it, and this kernel raises the card** (kernel/kernel.py:51481; tests/test_mesh_settings_adopt.py:858). The receiver hands `_propose_peer_settings` no pin field; two-way, the card flaps on every pass. A pin field in the push body closes nothing; reading the pusher's cached row pins closes the two-way case only; one-way has no receiver-side fix, so docs/reference.md:4388 and 4391-4392 need qualifying; no receiver-side test exists.

- **The un-pin on the polled kernel** is narrower than the comment above says (kernel.py:51588, 51592): with a row for the hub the polled kernel's own pass files the record the un-pin applies; the defect is confined to the one-way attach, the default. Pushing to a pinned peer works only if a push-filed record counts as attached and the un-pin drops the applied record; it reverses the one-A rule at docs/reference.md:4389 and costs one POST per pass while the stamps differ. Restoring the pre-scoped value trades this card for one on the hub if the hub moved meanwhile; rewording 4384-4386 and the hover is accurate if it names where the return works (a row for the machine carrying the newer stamp, from check-in or a mutual attach).

- **A scoped gesture stamped between the store's and the pin store's stamps half-applies** (kernel.py:51566; tests/test_mesh_settings_adopt.py:783): store on@1000, pin@1100, un-pin@1500, scoped off@1200 gives (False, 1200), no pin, no frame; the shipped test's equal stamps never reach the pin store's gate. In `_pinned_stand_down`, refuse a scoped gesture not above the pin store's stamp through the stale path, one frame carrying the pin's stamp, gated on scope alone, before any write, never a rollback.

- **Keep mine, pin, one poll, un-pin applies the kept value** (kernel.py:51851, 51862): the held branch runs before the answered check, so a kept stamp becomes a held record once pinned. Move the answered check above, making an un-pin-side skip redundant; the test must poll between pin and un-pin.

- **Under All kernels a toggle left indeterminate by the scoped view stays so** (ui/webview/gear.js:1952). The text pin at ui/webview/settings-proposals.test.ts:95 forbids the exact old line only; make it a positive pin on the clearing line for the three stores `fillAutoNudge` does not own. No executed test picks two kernels then All kernels; a browser unit reds here but skips in CI (no Playwright browser).

- **The refill ladder runs on every click** (gear.js:1860, 1868). Re-reading on scoped clicks alone would revert M3; gate on other up kernels or a scope so one kernel pays nothing. The event road needs a success frame from every setter, not `setTaskTracking` alone, and a cross-frame message from federation's /tunnels poll in the shell frame.

- **The option list is rebuilt on every fill** (gear.js:1920): a press held across a rebuild is dropped here, not at a54a891a3. Delegation alone cannot save it: update nodes in place, keyed on the input, not its label. The list (gear.js:85) lacks `aria-multiselectable`, per-option selected state and arrow keys, and Escape closes the whole card; register it as a level in the settings close hook.

- **A scoped Task tracking change to a remote kernel flips this dashboard's shell** (gear.js:1088): the remote arm's echo arrives untagged and the shell closes Feed and Outline. Tag it with its host in `prefixInbound` beside the stale-frame tag; act on host-less frames only.

- **After a scoped click the row shows the old value for 12 to 16 s** (gear.js:2031, 1951): each fill writes the local value, then the /tunnels branch overwrites it with the picked kernel's stale row. Keep the clicked value until every picked row's stamp reaches the gesture's, covering both writers, released on a stale frame, a row leaving up, or a scope change.

- **A mixed scope's click direction is unstated** (gear.js:1951): the indeterminate toggle keeps this kernel's checked state even when this machine is unpicked, and a click posts its opposite. Say it from the checked state at fill time and in the echo listener at 1089.

- **The pin glyph gives no acknowledgement** (gear.js:1903): byte-identical for seconds after a remote pin, so a second click mints a new stamp for the same value. Flash it and set an in-flight state that ignores clicks, cleared when the targets' pins reflect the answer, or on the last fill or a stale frame; `aria-busy`, not disabled.

- **Apply anyway on a stale scoped click broadcasts** (gear.js:1650, 1617): the refusal echo carries no hosts and the broadcast route strips the scope, so every unpinned kernel applies. Re-issue to the refusers the frame names, not the selector's current pick, or drop Apply anyway for a scoped gesture.

- **The dropdown wears raised-surface tokens, not the menu tokens** (ui/webview/gear.css:110). A literal-scan entry is vacuous; the pin that bites is a token assertion in the reference-menu loop under the full prefix `#rsettings #rs-scope-list`. Card, radius, shadow and hover divergences stand whatever the current mark.

- **The docs state the earlier contract** (docs/reference.md; plans/settings-across-machines.md:142-143). The one-A sentences are byte-identical to the base: keep "never broadcast" at 4362 but cover a remote dashboard's scoped pin, add the scoped remote origin at 4366-4367, and at 4370 name the glyph (also shown with one kernel when this machine pins); the plan's glyph and Sessions-menu bullets are stale. Extend the mixed-mesh sentence at 4392: while an older kernel carries an unpinned scoped value, every kernel that polls it or takes its push raises the proposal card once per stamp, whose Apply undoes the scoping.

## Smaller points

- kernel.py:51836: a pinned peer's poll skips the drop loop over other hosts' outranked records; run the loop first.
- kernel.py:51592: an un-pin applies a held record written before its peer pinned; a return flag resolved at the next fresh poll.
- kernel.py:51562: a pin wipes held records too, so a prompt un-pin returns nothing; keep them here and in `_answer_setting_proposal`.
- kernel.py:23188, 27609: a sha-less peer is listed up with null settings, stamps and pins; lift them outside the sha gate in poller and supervisor.
- kernel.py:68222, 51596: the un-pin's bare setter; one per-store helper (task tracking's four steps and echo, the two ticks, never the re-pin).
- kernel.py:51588: an un-pin applies a since-isolated peer's held record, as at a54a891a3; a trust filter at click time.
- kernel.py:51438: an unreadable pins file reads as unpinned, so the push skip pushes a pinned store; the #1859 split.
- kernel.py:51574, 68222: the `gt` parameter is never read; drop it.
- kernel.py:51566, 51835: a scoped gesture that moves and pins logs nothing, nor does dropping a pinning peer's record; the second line inside the record-drop branch.
- kernel.py:51592: a record under a name no row has is skipped, one re-keyed onto another row misdecides the un-pin; `_rekey_peer_proposals` at both pop sites.
- kernel.py:64037, ui/webview/strip.css:109: both chips are 10px beside 11px annotations.
- kernel.py:61798, strip.ts:682, gear.js:1917: the pin hovers name stores by internal id; a four-entry label table.
- gear.css:94: the glyph's hover opens the row's description popover; a separate rule.
- gear.js:392: glyphs and selector are wired on a zero-delay timer; call both last in the init.
- gear.js:1914: a failed /tunnels read drops the pick, so the next click broadcasts; keep it, pruning only on rows in hand.
- gear.css:110, gear.js:1918: the list has no height cap and the pill grows a name per pick; 40vh with scroll, "N kernels" beyond two.
- gear.js:2016: the gesture clock ignores peers' stamps, so a scoped click to a peer stamped ahead is refused as stale; learn every up row's.
- gear.js:1874: an older kernel's glyph is unlit, clickable and refused silently; a null pinned map marks only a pre-phase-one kernel; use a capability signal.

## Tests

- tests/test_mesh_settings_adopt.py:865 (kernel.py:51904): the unnamed-answer test puts the held and open records in two stores, so it is green before its fix; the one-store case (pin lifted through the bare store, not the arm; B not re-polled) reds against a54a891a3, not the base.
- tests/test_mesh_settings_adopt.py:822 (kernel.py:51593): the same-value branch is unpinned; write the record while the values differ, then a local-origin click with the origin passed explicitly, then the un-pin.
- ui/webview/strip.ts:677: the untested strip chip; a pure helper exported from strip.ts, asserted in strip.test.ts through its output, since a source pin misses the inverted filter.

PR #1911, opened 2026-09-19
The Artifacts pane: a session's written, shown and dropped files as a list and a grid of large thumbnails, optional and off by default

henrypinkard:
The Artifacts pane, the code phase of `plans/artifacts-pane.md` (design PR 1906): an optional pane, off by default, experimental. A session selector at its top and, under it, every file that was put into that session's thread, as a list and, for images, a grid of large thumbnails to cycle through large. Nothing is injected into any session and no instruction changes: an on-top read of what already happened, by deterministic rules only. Feature tier: a capability inside the existing pane model.

## What the user sees

- The gear's General tab, Panes section, gains an **Artifacts** row (off by default, like the Files row). On, the rail gains its toggle and the phone its tab; the toggle shows the pane as the rightmost column.
- The pane: a session selector (the picker's list, running first) with the session's identity colour; a Refresh; the count. Under it the grid of large thumbnails (an existing, allowed image each, newest first, loaded lazily through the token-authed file route with the session's sid), then the list: every file the thread named, newest first, one row per path, the rule as a small word (written, edited, notebook, shown, dropped), the age; a missing file struck through and marked, a refused one with the route's reason as its title.
- A thumbnail opens the picture large in place (the chat's lightbox), left and right (the arrows) cycle the grid's pictures with the position cue, Escape closes; one control sends the picture to the Files pane through the shell's `viewFile` relay (hidden where no Files control exists). A row click walks the chat's route ladder: an open Files pane takes it, else the shared viewer opens here.

## Design points (the merged design, followed)

- **Three rules, each with its source record** (`_artifacts_walk`): rule 1, an assistant `tool_use` of an edit tool (`Write`, `Edit`, `MultiEdit`, `NotebookEdit`: the `_EDIT_TOOLS` set) and its `file_path` or `notebook_path`; rule 2, a path-shaped token of the assistant's prose (the chat's token grammar, `_path_tokens`) that the chat would have linked: verified for that message in the path-links cache, existing now, or an image path by extension; rule 3, a path under the state directory's `drops/` in the person's own user turn, as an image block's source path or as text. Bash is not read (the road not taken, said in the docstring). A relative path resolves against the session's cwd; a `file://` URI unwraps.
- **The listing** (`_artifacts_items`): one entry per resolved path, the latest mention winning; newest first; capped at 500 with the cap said; each stat'd (a missing file listed and marked) and judged by the file route's rule (`_slice_allowed`: a secrets-shaped name, a path outside the folder and the home); a drop shows the name the user dropped and is allowed wherever the state directory lives.
- **Lazy, request and response** (`listArtifacts {sid, reqId}` → `artifactsListing`): answered by the owning kernel on the asking socket; the parse through `_parse_cached` first, else one `parsed_session` into the shared store; nothing pushed, nothing written. The page (`/artifacts`, `_artifacts_page`) boots with the shim's stale opt-out like the Files pane; its iframe is served with `data-src` and loaded once when the control is on, so an off pane costs nothing.
- **The shell hooks, the shipped naming exactly** (the timeline owner's word for the registry): the `_PANE_ORDER` entry `("artifacts", "Artifacts")` (the rail button, the phone tab and the pane-set broadcast come from it), `artifacts-pane`, `f-artifacts`, `po-artifacts`, `--g-artifacts`, the `LBL` word, the fourth gutter `gv-d`, the focus ring's map and columns, the Escape and mobile lists, the gutter script's pane and grow lists and id fallback, the viewer set; a fresh setting `showArtifactsControl` read like `filesCtl()` (`artifactsCtl()`, `artifactsCtlM()`, `body.no-artifacts-control`, the toggle's refusal, `avail.artifacts` in the broadcast, the close on the same apply).
- **The pane bundle** (`ui/webview/artifacts.ts`, its pure parts in `artifacts-model.ts`, `artifacts-pane.css`): one delegate on the root for the clicks; the answer dropped when its `reqId` is not the latest or its sid is not the selection; the lightbox's nav provider set to the grid's order.

## Tests

- `tests/test_artifacts_list.py`: the three rules (two writes of one path, a relative Edit, a MultiEdit, a NotebookEdit, a Bash command yielding nothing, a Read yielding nothing; an existing image path, a verified-then-gone path, an image path by extension, a bare word that is no file, a `file://` URI, the person's prose ignored; a drop as an image block and as text, a user path outside drops, a peer's turn); the listing (dedup, order, missing, refused twice, the cap); the cached-first read and the one parse on a miss; the op's answer by request id; the pane key and the viewer set; the page and route; the shell hooks census (every hand list) and the control gate in both layouts.
- `ui/webview/artifacts.test.ts`: the grid rule, the cycle, the rule words, the route ladder, the age words; pins on the setting, the gear row, the bundle entries, the request-by-id and the relay.
- `tests/test_artifacts_pane_served.py`: the served lab, two written files (one deleted), one rendered image, one drop: the off state (no toggle, the iframe never loaded, the toggle refused), the control on (toggle shown, the page loaded once), the pane shown, the selector, the four rows newest first with the missing mark, both thumbnails decoded from the file route, the large view and the arrows, the Files pane relay received by the shell and the Files pane brought forward, the toggle off.
- Re-aimed for the sixth pane: `test_files_pane` (the ordering, the lists, the shim census, the gear defaults), `test_pane_order_parity`, `test_pane_gutter_drag`, `test_pane_gutters`, `test_pane_state_broadcast`, `test_kernel_pane_rail`, `test_kernel` (the landing's pane map), `test_chat_split`, `test_kernel_panenav`, `gear-tabs.test.ts`, `esc-close.test.ts`.

## Red-first (the head's new tests copied into a detached worktree at the design head e0a53f9b, run there)

- `tests/test_artifacts_list.py`: all 10 red (the walk, the listing and the op absent; no pane key, no route, no page; every landing hand list without the pane). `tests/test_artifacts_pane_served.py`: all 4 red (no toggle, no pane, no page). `ui/webview/artifacts.test.ts`: the pins test red (no setting, no gear row, no bundle entries); its three pure-part tests exercise the new module itself, carried with the test, so they have no base behaviour to be red against.

## Counts

- 24 files changed, 1073 insertions(+), 99 deletions(-) against main (the design merged as PR 1906 and carried). Production: `kernel/kernel.py`, `ui/webview/artifacts.ts`, `ui/webview/artifacts-model.ts`, `ui/webview/artifacts-pane.css`, `ui/webview/settings.ts`, `ui/webview/gear.js`, `vscode-extension/esbuild.js`; three test files added, fourteen pin files re-aimed.

## Verification

- On the head before the last audit fixes (4537f895, the design carried): the whole extension suite (5140 tests; the one red was the Escape wiring pin, re-aimed and green alone) and the whole Python suite in three capped shards: the labs shard green but for the served settings hide's row count (re-aimed, green alone); the unit shards green but for the pane-list pins named above (each re-aimed and green alone), the body-readers audit and the settings-sections census (joined, green alone) and `test_kernel_judge_model::JudgeSettings::test_judge_run_applies_tier_effort`, the shard-order flake seen on PRs 1885 and 1890 (passes alone).
- On the head as pushed (the walk hydrating first, main carried): tsc, `test_artifacts_list`, `test_lazy_body_readers`, `test_kernel_settings_sections`, the served settings lab and the Artifacts served lab; the pane's pure-part and pin files, the Escape and gear pins. CI is the whole-suite gate on the pushed head.

## Out of this PR

- Bash output paths, subagent transcripts, files named in tool results; any write from the pane; a pushed view or an index; the panes-as-data registry (the timeline owner's).

Synthetic fixtures only (`TESTHOST`, the notes-api world, invented paths, placeholder uuids); no personal identifiers; no em-dashes in added lines; the ruled-out word appears in added lines only as the Outline pane's existing key inside the hand lists this change extends, never in new prose.

## Round two (the verifier's four mediums and the lows, 2026-09-19)

**M1, the session resolution.** The listing resolved its session through the 48-hour sweep while the selector offers the picker's thirty days, so a session idle two days was offered and answered "no transcript here". It now resolves through the one-sid API (_session_row: the sweep, else the SDK registry, else discover's wide walk). Pin: a session the sweep no longer names lists through the wide walk (red at 30ecfefa on the error text, green here).

**M2, the load.** Turning the control on loaded the page and walked the remembered session on every dashboard load while the pane stayed off screen. The iframe now takes its src when the pane comes on screen with the control on (the optional panes' rule), and on a phone when its tab shows. The lab's read is re-aimed: the control on loads nothing (red at 30ecfefa: the src was set), the toggle loads the page once; the landing pin names the gated line and asserts the ungated one is gone.

**M3, the Claude configuration directory.** The design claimed the shared route refuses a path under it; the route does not (home is a confinement root), so such a path listed unmarked and would be served. Decision: the pane marks it refused itself, with a plain reason ("under the Claude configuration directory"), and the shared route is unchanged. Why the narrow fix: the route's confinement is the chat's contract for every path link, and widening it is a decision for the user, not this pane's; the pane's listing is where a thread's own transcripts, task stores and settings would surface as "artifacts", and they are not its files. Pin: a markdown file under a synthetic configuration directory INSIDE the session's folder (confined, so only the pane's rule can refuse it) lists marked (red at 30ecfefa: refused empty). The design's section 1 says the same now.

**M4, the base reds.** Six of the ten kernel tests were red at the base on a helper's missing name. Every walk and listing test now asks the way the pane asks, the listArtifacts op on a fake socket over a session whose parse the cache stub answers, and the page test drives the real GET dispatcher. At the merge-base dc4b6c2a all twelve are red on behaviour: the op answered with a frame that is not the listing (the shape asserted, never a KeyError), the route answering 404, the landing lacking the rail button, the pane order ending in Files, the control's script absent. The cap is a literal 500 in the test, so the base reds on the count rather than on a constant's name.

**Lows folded:** (a) a drop keeps its dropped name whichever rule named it last (the strip keys on the path under the drops folder); (b) a kind the preview does not show is an ordinary kind, listed plain with the route's confinement still judging it, no thumbnail (the grid is images), and a click writes "The viewer cannot show <name>: not a kind it renders." beside the count; (d) the miss parse writes under the key the reader uses (the states log, the display's sdk_human), pinned by the kwargs the stub records (red at 30ecfefa: none passed); (e) the reduced assertion is gone; (f) the recut lab's rows list names rs-artctl and the count is one constant; (g) the cap notice reads "the newest 500 files of more"; (h) the Files pane's ordering test is named for what it asserts. **Stated, not fixed (c):** under the docking switch the pane element is not in the tree and renders at 0 by 0 while its toggle reads on (no throw, no broken layout); the registry's phase two lifts it (the design says so now).

**Verification on this head:** tsc; the kernel tests (16) and the Files pane tests; the bundle pins; the Artifacts lab and the settings recut lab; the whole Python suite in three shards (the unit shards 6847 and 7157 passed, 48 and 7 skipped, one red in the first: the judge-model tier-effort test, the shard-order flake seen on 1885 and 1890, green alone at this head; the served labs shard 575 passed, 2 skipped, 0 failed); the whole extension suite (5146 tests: 5140 pass, 6 skipped, 0 fail).


lczh:
Evidence executed against 30ecfefa.

Two items keep the pane from doing its stated job: most sessions the selector offers answer "no session with that id has a transcript here", and the phone tab shows a blank; the rest are hand lists the sixth pane did not reach.

## At this head

**The selector offers 30 days of sessions; the listing resolves the sid inside the 48-hour window.** `_session_list` (kernel/kernel.py:16050) walks `_sessions(now, PICKER_WINDOW, forks=False)` at 16057, 30 days (16043). `_artifacts_list` (59684) looks the sid up with `_sessions(now, None, forks=False)` at 59689; `_sessions` (4690) normalizes None to `jd.WINDOW`, 48 hours (kernel/judge.py:482, the cutoff at 9132), so an offered session idle more than two days gets the error at 59691. On a hermetic state root with one registered session and a 72-hour-old transcript: `_session_list` offers it as "2d ago", `_artifacts_list` returns no items and that error; at one hour it lists. Neither test reaches this: tests/test_artifacts_list.py:150 stubs `km._sessions` to one row, and the served lab's transcript is one hour old (tests/test_artifacts_pane_served.py:150). `_session_row` (4775) exists for exactly this, its docstring recording the 2026-09-14 ruling that age never decides whether a session can be acted on, and in the same probe it finds the row. Resolve with `_session_row(sid, now)` and add a test whose transcript is three days old.

**On a phone the Artifacts tab appears but its pane can never show.** `_mtab_buttons_html` (63788) renders the tab from `_PANE_ORDER` and `show()` in `_LANDING_MOBILE_JS` (62541) flips `m-on` on `#f-artifacts`, but the two mobile CSS lists in `_landing()` were not widened: `#chat-pane,#fleet-pane,#feed-pane,#files-pane,#tl-pane{display:contents!important}` (64564) and `#f-chat.m-on,#f-fleet.m-on,#f-feed.m-on,#f-files.m-on{display:block}` (64569). The iframe stays under the mobile `iframe{display:none}` reset and the wrapper under the desktop `body:not(.po-artifacts)` hide, the leak the block's comment (64556) says must not reach the phone. The rendered landing at 390x844 touch with the control on: a tap sets `data-tab=artifacts` and `m-on`, and both `#f-artifacts` and `#artifacts-pane` compute to display none, 0x0, while the Files tab in the same run shows 390x813. tests/test_files_pane.py:294-295 pins both lists for Files; the Shell census pins neither, and the served DRIVER runs one 1500x900 viewport (tests/test_artifacts_pane_served.py:46) with no phone pass. Add the selectors, pin both strings in the census, and give the served lab a phone leg.

**With the docking kit on, the pane is reported on and collapses to nothing.** The plan defers the docking leaf to the registry PR (plans/artifacts-pane.md:123-125) on the premise that the pane meanwhile shows through the flex path, which the kit's stylesheet voids: `body.pane-docking .pane{position:absolute}` (ui/webview/panedock-main.ts:80) applies to `#artifacts-pane`, while `Engine.shown` builds the row from `[FLEET, FEED, FILES]` (259), `ROW_ORDER` stops at FILES (ui/webview/pane-dock.ts:30), and `render` (294) writes rects only to tree leaves. The head's dist against a hermetic kernel, paneDocking and the control on, the rail button clicked: `po-artifacts` on, the layout leaves still chat/fleet/feed/files/tl-pane, `#artifacts-pane` 0x0 at the row origin; docking off, a 256px column. The docking served lab seeds no artifacts key (tests/test_pane_docking_served.py:82) and nothing ties `ROW_ORDER` to the landing's `COLS`. Either add the pane to the kit's lists or have the kit exclude it while docking is on and say so in the gear row; either way, pin that `COLS` mapped through `PANE` equals `ROW_ORDER` plus the chat columns.

**The walk hydrates the whole session into the shared parse, and a closed pane asks for it at boot.** `_artifacts_walk` hydrates every turn (59596) of the tree `_parse_cached` and `parsed_session` return by identity; `_hydrate_one` writes the body in place and pops `lazy` (kernel/event_model.py:7059, 7087), with no re-lazify short of `_MAT_CAP`. A restored parse with 120 lazy pre-cut atoms: one `_artifacts_list` hydrated all 120 (2.43 MB, heap +2.6 MB). Page side, artifacts.ts requests at module end when a sid is remembered (158), before the panes broadcast (142) says whether the pane is shown, and `apply()` loads the iframe on the control alone (63146), so a control-on, pane-closed dashboard pays this on every reload. The plan does mandate one pass over the whole thread through the shared parse, so this is a cost to bound, not a rule broken: defer the first request until the panes message reports artifacts on, and pre-filter by the lazy scalars or read bodies without leaving them resident.

**The command palette has no pane.artifacts command, and its chord is dead inside the pane.** ui/webview/palette-main.ts is untouched: the `panes` array ends at files (168) and the wireKeys list at f-settings (374). In headless Chromium, Ctrl+P opens the palette from the shell and from inside f-files, not from inside f-artifacts; "f-artifacts" added to the list makes it open; the "Show or hide the" list has no artifacts entry. Add `["artifacts", "artifacts"]` with a `when` on `no-artifacts-control` (the shape at 176), add f-artifacts to wireKeys, and pin both in artifacts.test.ts.

## Smaller points

- ui/webview/artifacts.ts:124: `openRow` sends an existing image row to `openLarge` before `rowRoute`, so with the Files pane open an image row opens the lightbox and posts no relay. The body and the plan (plans/artifacts-pane.md:76-77) say a row click walks the ladder with no image exception. Route image rows like the others, or state the two gestures.
- tests/test_lazy_body_readers.py: `_artifacts_walk` joined KERNEL_ALLOWED (24) but not the ordering tuple (129-130). A scratch kernel.py with `em.hydrate` moved below the walk's loop leaves both audits green, and a lazy atom then raises LazyBodyRead inside the per-atom `except Exception: continue` (kernel/kernel.py:59642), so older turns' artifacts vanish with no error. Whole-turn hydrators were never in that tuple, so this is hardening: add the entry and re-raise LazyBodyRead out of the per-atom except.
- tests/test_artifacts_list.py:126 writes the "outside" file under `tempfile.gettempdir()`, and `_slice_confined` roots on `expanduser("~")` (kernel/kernel.py:52173). With HOME at a scratch dir and TMPDIR beneath it the module runs 1 failed, 9 passed, at the `startswith("outside the session")` assertion (137) with an empty refusal. Point HOME at a temp dir for the test and put the outside file in a second one.
- tests/test_settings_tabs_recut_browser.py: renamed to six rows with count 7 (450, 454), but the DRIVER rows list still names five ids (193) and the shown pin is `["flex"] * 5` (457), so covered, hidden and shown never look at `#rs-artctl` (ui/webview/gear.js:133). A rule leaving the Artifacts row visible under the hide keeps the test green; with `rs-artctl` in rows and `["flex"] * 6` it goes red.

lczh:
The test runs behind the comment above, on 30ecfefa (merge base with main dc4b6c2a).

## Tests

- tests/test_artifacts_list.py in the worktree at the head: 10 passed in 0.41s.
- The same module with HOME pointed at a scratch directory and TMPDIR nested beneath it: 1 failed, 9 passed, the failure at the Walk listing test's outside assertion with an empty refusal.
- tests/test_lazy_body_readers.py at the head: 6 passed.

- kernel.py with the walk's `em.hydrate` moved below its loop: both body-reader audits stay green; with `_artifacts_walk` in the ordering tuple that copy goes red and the real file stays green. A copy with the call in a dead branch passes either way, so the re-raise is the stronger half.
- The kernel loaded hermetically with `em.hydrate` a no-op: a lazy atom raises LazyBodyRead inside the per-atom try, the except swallows it, and only the resident atom's mention returns.
- `_landing()` with `#artifacts-pane` and `#f-artifacts.m-on` added to the two mobile lists: the Artifacts tap shows the iframe at 390x813, from display none and 0x0.
- The palette bundle with "f-artifacts" in the wireKeys list: Ctrl+P opens from inside f-artifacts.
- gear.css with GEAR_HTML and a rule leaving the Artifacts row visible under [hidden]: the shipped five-id DRIVER still reports count 7, covered true, shown flex times five; with rs-artctl in rows and flex times six, red on the hidden check.

## Probes

- Selector window: a temp state root, one registered session whose transcript mtime is 72 hours old. `_session_list(now, {})` offers it as "2d ago"; `_artifacts_list(sid, now)` answers "no session with that id has a transcript here"; `_session_row(sid, now)` returns the row, and a `PICKER_WINDOW` lookup yields one item, report.md. At one hour both paths list.
- Phone: `km._landing()` in Chromium at 390x844 touch with the control on: the tap sets data-tab=artifacts and m-on; `#f-artifacts` and `#artifacts-pane` display none, 0x0; the Files tap 390x813; the desktop rail toggle shows Artifacts at 494x661.
- Docking: the head's dist against a hermetic kernel with paneDocking and the control on, the rail button clicked: po-artifacts on, leaves chat/fleet/feed/files/tl-pane, `#artifacts-pane` display block, position absolute, 0x0; docking off, a 256x766 column.
- Hydration: a restored parse with 120 lazy pre-cut atoms; one `_artifacts_list` hydrated all 120 (2.43 MB attributed to the walk, heap +2.6 MB); with the hydrated-bytes cap forced to 1 byte the memo held 0 entries and every atom kept its body.
- Image row: the bundled artifacts.ts framed in a parent broadcasting the Files pane on and available: the image row opened #romp-lightbox with no relay; the markdown row posted `{path, sid, pane:"pane"}`.
- Palette: the bundled palette-main.ts over a same-origin http page with #f-files and #f-artifacts iframes: Ctrl+P from the shell and from f-files opens it, from f-artifacts nothing.

lczh:
Executed against 3f0e7d73. Three of the nine items landed with pins that run: `_artifacts_list` resolves its session through `_session_row` (kernel/kernel.py:59703), pinned by `test_a_session_idle_past_the_caption_window_still_lists` (tests/test_artifacts_list.py:265), and on a hermetic root with one registered session the selector offers a transcript aged 72 hours, 20 days and 29 days ("3d ago", "20d ago", "29d ago") and the listing answers report.md each time (at 31 days the selector drops it and the listing still answers through the 365-day walk); the iframe takes its src only with `po.artifacts` on (63164), the served lab reads no src with the control on and "/artifacts" after the toggle (tests/test_artifacts_pane_served.py:223, 225), and the Shell pin at 319 asserts the ungated line is gone; the recut driver names `rs-artctl` and pins six flex rows (tests/test_settings_tabs_recut_browser.py:194, 458). At this head: the kernel tests 12 passed, the Artifacts lab 4, the recut lab 13, the Files pane 27, the body-readers audit 6.

## At this head

**On a phone the Artifacts tab still shows nothing, and the tap now loads the page into the hidden frame.** The two mobile lists are unchanged (`#chat-pane,#fleet-pane,#feed-pane,#files-pane,#tl-pane{display:contents!important}` at 64582, the `.m-on{display:block}` list ending at `#f-files` at 64587) while the round-two `show()` sets src on the tap (62559). The landing at 390x844 touch with both controls on: the tap sets `data-tab=artifacts` and `m-on`, `#f-artifacts` reads `src=/artifacts` and display none at 0x0, `#artifacts-pane` display none at 0x0; the Files tab in the same run shows 390x813. The document and the socket M2 was made to avoid are now paid on the phone for a blank. Add the two selectors, pin both strings in the Shell census (tests/test_files_pane.py:294-295 pin them for Files only), and give the served lab a phone leg.

**The palette is untouched** (ui/webview/palette-main.ts:168 and 374 are byte-identical to 30ecfefa): no `pane.artifacts` command, and Ctrl+P is dead inside the pane. Neither the body nor the plan defers it.

**An image row still skips the route ladder** (ui/webview/artifacts.ts:127 sends it to `openLarge` before `rowRoute`) while the body and plans/artifacts-pane.md:80 say a row click walks it. Route it or state the two gestures.

**The temp-directory dependence doubled.** tests/test_artifacts_list.py:201 and 227 both write the "outside" file under `tempfile.gettempdir()`, which `_slice_confined` (52173) accepts whenever the temp directory sits under the home. With HOME at a scratch directory and TMPDIR beneath it the module runs 2 failed, 10 passed (the assertions at 212 and 240); one failure at 30ecfefa.

**The ordering audit still lacks the walk.** tests/test_lazy_body_readers.py:129-130 is unchanged and the per-atom `except Exception: continue` at kernel/kernel.py:59642 still swallows a LazyBodyRead, so its 6 passed says nothing about the walk's hydrate-before-read order.

## Smaller points

- Docking: the plan now states the 0 by 0 render under the docking switch (plans/artifacts-pane.md:129-131) and defers the leaf to the registry. Accepted as a decision; the gear row still offers that user a toggle that shows nothing, and no pin ties the landing's `COLS` to `ROW_ORDER`.
- Hydration: the page half landed (a closed pane requests nothing); the kernel half is unchanged, the walk hydrating every turn of the shared parse (59596), unmentioned in the body.
- The body's "Lazy, request and response" bullet still says the iframe is loaded once when the control is on; the round-two section says the new rule. One sentence to reword.
- Checked and clean, so not listed: `_session_row` reaches `_sessions(now)` with forks on where the old lookup passed `forks=False`, but a fork lane's fsid is never a romp sid (kernel/judge.py:9095-9097), so a selector sid can only match its registered row; Codex rows join discovery as before; `po.artifacts` has a default (63084), so the new gate never reads an undefined key.

ksarma:
A note after the merge. The evidence was executed against 3f0e7d73e, the merged head; every file and line reference is on that commit. Where an item below overlaps the three comments above, only what this head's evidence adds is written; the docking render, the palette and the image row are theirs.

**What the listing's `refused` field means against what the file route serves.** The pane judges by the slice road's rule; thumbnails, large view and viewer fetch through the bytes road of GET /file, which applies none of the rule's confinement, secrets or link-dress checks (its own size cap and content sniff aside).

- The configuration-directory rule at kernel/kernel.py:59680 runs before `_slice_allowed` and refuses every file under the directory, the glossary (the route's own exemption, `_glossary_owned`) and a project's auto-memory file included, which GET /file served 200 (directory under the home). Judge the route's rule first, the pane's only on a path the route serves; exempt the glossary helper and the project memory folder rather than enumerating what to refuse.
- The docstring at 59650 and the plan (lines 43 and 47) still call `_slice_allowed` the file route's own rule while the bytes road served 200, one sniff 415 aside, wherever the slice road refused; name it the preview popover's rule.
- The drops exemptions at 59687 and 59688 compare the link path to the drops prefix while every other judgement is of the real path, so an outside-target symlink under the drops folder lists with `refused` empty; compare real paths in both branches, leaving the name strip and rule 3 on the link path.
- The kind-other clearing at 59686 keys on the real path's reason while the row's `kind` is the link name's, so an image-named symlink over a pickle entered the grid as a broken thumbnail; gate the branch on `it["kind"] == "other"` too and take the row's kind from the real name.

**An exception from the walk answers nothing.** The try at kernel/kernel.py:59706 covers the parse alone, the walk at 59711 running outside it: a deleted lineage file tore the connection down, an unmapped fsid got no reply. Move line 59711 inside the try, the reply call outside any except so a broken pipe still reaches the teardown.

**Rule 2 lists the token, not its target.** At kernel/kernel.py:59629 a linked token lists as `absolute(tok)` while the path-links cache maps tiers 2 and 3 to a repo-relative target, so the walk marked real files missing. Join the cache's value instead, pinned through `_path_links`; after a restart or the 50000-entry clear the walk still lists the bare token: say so in the docstring, or resolve misses through the links helper with one memo per walk.

**An oversize file of a preview kind is inert.** "too large to show" lands in `refused` (kernel/kernel.py:59684), so an oversize markdown or image is dimmed with an inert click while the chat's ladder opens the same path in the viewer, whose 413 page carries a Download button. Let that reason through with `refused` empty (the size check is `_slice_allowed`'s last) and keep an oversize image out of the grid with a kernel-set flag, not a page-side constant, or the grid fetches a 413 thumbnail.

**A request lost to a socket drop is never re-asked.** artifacts.ts listens to neither `romp:wsdown` nor `romp:wsup` and loading clears only on a matching answer (line 154), so a kernel restart left "Reading the thread" standing; file-browse.ts:471-483 recovers on those events. Do not clear loading on wsdown, since `paint()` has no branch for loading false with a null listing; set an idempotent resync latch if a request was in flight and on wsup re-ask the current selection.

**A kind-other row's click is a dead end, and the note that replaces it cannot be read.** At artifacts.ts:126 the row returns before `rowRoute`, reaching neither the Files pane relay nor the shared viewer, which a chat link to the file opens; delete the line. The note clips at 25 of 56 characters and persists until Refresh repaints: give it its full text as title, wrap the bar with the note on a full-width line (`flex:1 1 auto` on the note is a no-op), `.art-count` nowrap, and clear it on the row click and the thumbnail path. The refused row's early return at 125 must stay: an unguarded click would open the file the listing marks refused.

**The wait states.** A Refresh click with a listing up rebuilds the button in the same tick (artifacts.ts:47), so its flash never runs a frame; derive a busy state from the loading flag inside `paint()`, but do not disable it, since an unreachable remote kernel never answers. The first wait at line 75 is bare text though the stylesheet carries `.fileview-load` unused; inline the markup.

**The name column collapses under a long folder.** At artifacts-pane.css:33 `.art-name` shrinks by default while the folder at 34 has flex 1 1 auto, so names read one to three letters. `.art-name{flex:0 0 auto;max-width:55%}` with the folder at flex 1 1 0 and min-width 0 rendered every name whole; 50% leaves margin on rows under 320 px; the pin needs a cwd of 80 characters or more.

**The sheet's sizes and literals.** Three sizes are off the ladder (0.9em at artifacts-pane.css:21, 0.8em at 28 and 35) and the ladder loop does not read the sheet; adding it alone is a no-op, since the sheet writes `font-size:0.9em` with no space and the regex expects one: widen it and read the value through the capture group. Raw literals stand at 13 (#666, also inline at artifacts.ts:66), 21 and 38 (#e5484d) and 28; `var(--dim,#666)` for the dot, `--err` for the reds.

**The phone, beyond the earlier comment.** With a session and the tab remembered every dashboard load pays document, socket and walk with nothing shown. The six pins to re-aim, carrying the lists verbatim: tests/test_files_pane.py:294-295, tests/test_kernel_mobile.py:40, tests/test_kernel_pane_rail.py:238, tests/test_pane_hidden_word_browser.py:34 and 37.

## Smaller points

- Hydration at kernel/kernel.py:59596: the recurring cost is the prose walk, removed by memoizing the mentions per fsid, fileset key and link-cache state; the cold parse at 59707 calls `jd.parsed_session` directly, booked by /perf as a judge parse.
- The docstring at kernel/kernel.py:59700-59701 and the test comment at tests/test_artifacts_list.py:254-255 say the bare call re-parsed on every request, though with the ownership hook installed misses were +1 then +0 at both shas; describe the change as making the key inputs explicit.
- The phone tab bar overflows with both optional controls on (`#mtabs` 461 px in a 390 px viewport; Settings unreachable); the rule is at kernel/kernel.py:64596 and 64604.
- The boot gate at kernel/kernel.py:63164 keys on `po.artifacts`, which the phone layout ignores: stored true with the control on, a phone boot ran the walk into a hidden frame; gate on the layout's own answer, and give the phone `show()` line at 62559 the mobile guard.
- The selector is a bare native select (artifacts.ts:62); rows and tiles have no tabindex, role or key handling; every answer rebuilds the select, dropping focus to the body (line 67); sessions are asked once at boot (line 160): re-ask on Refresh and the off-to-on transition.
- The folder column at artifacts.ts:95 is cut by the name's length, so every drop row's folder ends in the millisecond prefix; derive it from the last slash.
- The age words (artifacts-model.ts:21) diverge from the chat's at seven to nine of thirteen offsets despite the docstring; use `relativeLabel` from time-marker.
- File names wear the accent (artifacts-pane.css:33) where chat links wear the link token, orange against cobalt in the light theme; `var(--link,#9cd2ff)`.
- The plan's section 2 (line 61) still says the kind rule applies to a drop and overstates the secrets rule, since the millisecond prefix defeats the basename-anchored `_SECRET_NAME_RE`; section 4 (lines 100-103) still names the bare parse call.

## Tests

The two temp-directory fixtures (tests/test_artifacts_list.py:201 and 227), as the earlier comment says; HOME under the World's root, the outside files in a sibling directory, passes under both placements. The ordering audit, as the earlier comment says: its load-bearing half is a Listing case with one lazy atom in an older turn, whose artifact vanishes under the mutation. The idle-session pin at tests/test_artifacts_list.py:268-269 stubs both doors, so narrowing the window in `_session_row` leaves it green, though the pusher snapshot module's aged-transcript walk catches that regression; only the composition over the real filesystem is missing.

PR #1919, opened 2026-09-19
Panes as data, phase one: the pane registry, its doors, and the shell rendering the pane set per request

henrypinkard:
Phase one of panes as data (`plans/panes-as-data.md`, section 7): the pane registry, its doors, and the shell rendering the pane set per request. With an empty registry the code panes' RENDERING is unchanged, slice for slice against the base kernel's (the body tag, the rail, the phone tabs, the pane row, the column and gutter rules, the gutter calls), and that is the first test; the page as a whole gains the guards and the attribute reads, deliberately.

## What lands

- **The registry** (`kernel/kernel.py`): one schema for every pane (`id`, `title`, `source`, `on`, `experimental`, `protocol`), checked by `_pane_check`. The shipped panes, the Artifacts pane included (PR 1911 landed while this was built; the branch is rebased onto it and the sixth pane is the sixth record), are the records in `_CODE_PANES`, asserted at import; `_PANE_ORDER` derives from them. A data pane is `STATE/panes/<id>.json` behind `define_pane` and `remove_pane` (an atomic write under one lock; a shipped id is refused; a file that fails the check is skipped and named on stderr once). `_panes_snapshot` lists the directory ONCE per call (memoized on the directory's stat and each file's (mtime, size), the board store's rule) and yields the records and the set's revision together; the landing lists once per build and hands the list to every builder. `_pane_order()` is the shipped panes, then the data panes by id. The shipped column order (the gutter chains, the data panes' gutter sequence) derives from the records, never a hand list.
- **Sources**: a kernel route (`/feed`), a page under the state root (`pane:<id>`, served at `/pane/<id>/` with `shim.js` and `theme.css` beside its files, the `/dist` traversal guard), or an `http(s)` URL (a plain sandboxed iframe: no token, no `?v=`, `protocol` forced to `none`).
- **Doors**: `POST /pane` (a definition, or `{"remove": id}`), `GET /panes` (the shipped panes marked `builtin`, then the data panes, and the revision), and `romp pane define|list|show|remove`, the board door's twin.
- **The landing**: `_panes_attr()` puts the data panes on `body[data-panes]` only when the registry holds one; `_data_pane_markup()` and `_data_pane_css()` add a gutter, a lazy `data-src` iframe, the column's grow rule and the gutter's hide rules after Files. The rail and the phone tabs read `_pane_order()` (an experimental pane has no tab). The five baked inline scripts (the pane controller, the grows and gutters, the focus ring, the bell's titles, the phone's frames) each gain a line reading the attribute; every pinned line stands.
- **Enforcement of protocol none, one check for every shell listener**: the first script on the page defines `window.__rompPaneSourceOk`: a message counts only when its IMMEDIATE source is a same-origin iframe of this document not marked `data-protocol=none`; the shell's own window, a stranger, a foreign origin and a frame nested inside a pane all fail. Every inline listener (thirteen, `activeTab` and the split's tab drag included; the fourteenth is the service worker's own channel) reads it FAIL-CLOSED as its first statement: no check, no message. The bundles read it through `ui/webview/pane-source.ts` (`paneSourceOk`, fail-closed): the palette's `openKeys` and `hotkeyConfigure` listeners, and the docking kit's grab and tab-drag registration and frame lookup. The pane controller tells a none frame nothing; the kit's `speaksProtocol` read excludes it from the mark, the detector and the tab drag.
- **The reload offer**: `pv` rides every keepalive beside `dv`; the pane shim bakes `LOADEDPV`, and a moved revision reaches the reload core's `notePanes`, which OFFERS a reload with its own wording ("The set of panes changed. Reload to see it."). Never a self-reload.
- **The gear and the palette**: a row per data pane in the Panes section (`gear.js`, checked by default unless experimental), `PaneSet` widened to keep boolean members by pane id (`settings.ts`), a palette command per data pane.

## Tests

- `tests/test_pane_registry.py`, every test red at the base on BEHAVIOUR (pane files written by hand as the door writes them; the landing, the routes, the keepalive frame and the executed scripts asserted, never a name): the rendering slices FIRST against `tests/fixtures/landing-code-panes.json` (made by `tests/landing_slices.py` over the base kernel, which names the commit), with a define and remove round trip; the schema through the door (defaults, refusals by member); define, replace, remove and the refusals through the routes; a file by hand, a bad file skipped and named once, an in-place rewrite re-read; the revision on the keepalive frame; the landing with three data panes (rail order, tab, lazy iframe, sandbox, CSS, attribute, escaping, no added script); the pane controller executed against a DOM stub with the attribute; the source check executed (a URL pane's and a nested frame's forged `toggleFleet` dropped, a protocol pane's taken, the immediate-frame table); a listener on a page without the check acting on nothing; the census over the built landing (every registration guarded, the service worker's excepted), the shell bundles' sources and the built bundles; the state-root routes; the reload core's offer executed.
- `tests/test_pane_registry_served.py` (Chromium, REQUIRE in CI): three data panes on disk; the rail, the body classes and the attribute; the state-root page with the shim, the theme and the broadcast; the URL pane sandboxed with the URL as given, its forged `toggleFleet`, `notify`, `reveal`, `openKeys` and `hotkeyConfigure` changing nothing (the Outline stays closed, no notice, the shortcuts modal stays closed, no hot key remembered) and its beacons showing no broadcast reached it; a same-origin frame NESTED in the notes pane forging the same and nothing acting; the gear's rows and the experimental pane coming in; a define while the page is open standing the offer, no self-reload, the Reload click showing the new pane.
- `tests/romp.bats`: the `romp pane` command in the board test's shape.
- `ui/webview/settings.test.ts`: a registry pane's choice rides the pane set. `ui/webview/panedock-main.test.ts`: `speaksProtocol`. `ui/webview/pane-source.test.ts`: the bundles' read is fail-closed.

Red-first at the base 145c125e per test (product files at the base, the new tests as written): see the PR's first comment for the measured reds. The branch was then rebased onto 88cb4d08 (the Artifacts pane as the sixth code record) and every pin and lab re-run there.

Two pre-existing reds on the build machine, not from this branch: `tests/test_held_mail_chat_served.py`'s at-bottom read failed once under the whole Python suite's load and passes alone on this head (the known flake of that lab); and `tests/test_drop_anywhere_served.py` skips at setup ("hermetic kernel never served /healthz") under REQUIRE on the shared main checkout too; it builds its lab kernel's environment from the whole inherited environment rather than the lab helper's list.

## Roads not taken

- No generic gutter loop replacing the three shipped calls: those lines are pinned, so the data panes get their own loop after them with the same neighbour rule.
- The docking kit does not yet read the pane list (phase two); this PR only excludes protocol-none frames from it. A fact for phase two: with the kit ON, a data pane toggled on renders 0x0 while its rail button reads on (no throw, the layout intact), because the kit's layout tree knows the shipped columns only.
- Phase three (the Artifacts pane's code record folded into data) is joint with the feed owner, now that their pane has landed.

Tier: feature


henrypinkard:
**Measured reds at the base 145c125e** (product files checked out at the base, the new and changed tests as they are on this branch; three short runs, reports kept in the runner's scratch).

`tests/test_pane_registry.py` (15 tests, 15 red at the base, 15 green on the head):
- by behaviour: the source check (a forged `toggleFleet` from the URL pane toggled: `1 != 0`, the check `undefined`); the doors (`POST /pane` answered 404 where 400 is expected); the state-root routes (`/pane/notes/` answered 404 text/plain where 200 text/html is expected); the pane controller executed with the attribute (`po-notes` never toggled; the red is on `define_pane` first, see below, and on the toggle once the store exists).
- by absence of the registry's names (`_pane_check`, `_CODE_PANES`, `define_pane`, `_panes_data`, `remove_pane`, `_panes_rev`): the schema (2), the store (4), the landing (3, including the byte-identity pin, whose empty-registry half holds trivially at the base and whose round trip needs `define_pane`), the pane controller (its `define_pane` seed), the keepalive pin and the handler census (textual: the `pv` line and the guards are absent).

`tests/test_pane_registry_served.py`: red by behaviour (the driver died at "no notes rail button": the base landing has no data pane).

`tests/romp.bats` "pane: define ...": red by behaviour (`romp pane define` is not a command at the base: exit 2 where 0 is expected).

`ui/webview/settings.test.ts` "a registry pane's choice rides the pane set": red by behaviour (the pane set dropped the extra key).

`ui/webview/panedock-main.test.ts` "speaksProtocol": red by absence (the export is `undefined` at the base; the test reads it off the module namespace so the build still succeeds and the settings test's red is measured).

After the base runs the branch was rebased onto 88cb4d08 (the Artifacts pane landed as a sixth shipped pane; it is the sixth record in `_CODE_PANES` here, its id reserved, its column in the gutter chain), and the pin set, the registry module, both served labs and the bats test re-ran green there.

Whole suites on the rebased head before the push: the extension suite with the served labs under `ROMP_SERVED_TESTS_REQUIRE=1` and the Python suite under the same, counts: the extension suite 5156 tests, 5150 passed, 6 skipped, 0 failed (exit 0); the Python suite 14618 passed, 57 skipped, 2337 subtests passed, with the two pre-existing reds the body names (the held-mail lab's at-bottom flake under load, green alone on this head; the drop-anywhere lab's kernel never serving its health route on this machine, on the shared main checkout too). The first run on the pre-rebase head found three pins on the keepalive frame's exact shape (the frame gains `pv`, the shim's keepalive branch is pinned whole, the shim test harness fills the template's slots positionally); the revision note now stands on its own line before the pinned branch and the two pins carry the slot.

lczh:
Evidence executed against 91af1bdc.

Two items keep the registry from holding what the user chose: a click on any of the three shipped rows in the gear rewrites the pane set without the registry members, so a pane the user turned on vanishes and one they hid comes back, and a state-root pane's shim raises the dashboard's "may be stale" banner after any reconnect and nothing retires it. The rest are a source check the split handler and two palette listeners never got, a reload offer whose one slot cannot hold a build change and a pane change at once, and a headline claim the first test does not pin.

## At this head

**Toggling a shipped pane's gear row drops every registry pane's stored choice.** `panesOf` (ui/webview/gear.js:464) still returns only `{ timeline, fleet, feed }`, and the three shipped rows' change handler at 465 does `s.panes = panesOf(s); ... save(s)`, rewriting the whole set from those three keys. The new registry-row handler at 485 was written to start from `st.panes` and keep the object, and `paneSet` in ui/webview/settings.ts:50-54 was widened to keep boolean members for the same reason, but the gear executes neither. The shell reads the set through `optOn` in `_LANDING_COLLAPSE_JS` (kernel/kernel.py:63548-63550): an experimental pane is on only when `p[k]===true`, a normal one unless `p[k]===false`, and the storage listener at 63608 runs `reconcile(true)`, which deletes `po[k]` and hides the rail button at 63555-63556. The exact `panesOf`, the shipped-row handler and `optOn` executed in node over a store of `{timeline:true,fleet:true,feed:true,lab:true,notes:false}`, that is Lab checked as the served lab leaves it and Notes hidden by the user: one Outline uncheck leaves `{timeline:true,fleet:false,feed:true}`, and `optOn` then reads `lab` false and `notes` true, so Lab's column, rail button and palette command go and Notes comes back, with no gesture on either. The served lab toggles only `rs-pane-lab` (tests/test_pane_registry_served.py:149), and settings.test.ts:159 pins the TypeScript `paneSet`, which gear.js never calls. Make `panesOf` keep boolean members beside the three, the rule `paneSet` has, or start the shipped-row handler from `st.panes` as the registry-row handler does, and add a gear test that checks a registry row, unchecks a shipped row, and asserts the registry member survives in the store.

**A state-root pane's shim is served with the stale prompt armed, so after any reconnect the dashboard wears the "may be stale" banner and nothing retires it.** The `/pane/<id>/shim.js` branch (kernel/kernel.py:66625) calls `_shim(pid, _dist_ver())`, leaving `no_stale` at its default (58998), while the Files, Artifacts and Settings pages pass `no_stale=True` (59922, 59951, 60114) for exactly this case, a page with no pushed view. In the shim, `onopen` on a reconnect arms the prompt at 59248 unless a restart was announced, the second keepalive raises it at 59257, `raiseStale` posts `{romp:"wsStale"}` to the shell at 59163, and `_STALE_JS` latches `connStale` at 63434; `clearStale` (59098) is reached only on a real frame at 59268, after the keepalive branch's `return` at 59258. No real frame comes for `app=notes`: the pusher's per-client loop sends only the `feed`/`fleet` and `timeline` slots (54443, 54462, no else branch), and `caps` answers a `ready` (68476, 68521) that a page carrying just shim.js never sends. The `notes` shim core in node over a fake socket: open, two keepalives, an unclean close, the redial, a reopen, and `stalePending` reads `"reconnect"`; the second keepalive posts one `wsStale`, and twenty more post no `wsFresh`. So after a crash restart, a sleep past the 30 s watchdog or a forwarder hiccup, every dashboard showing such a pane wears the banner until dismissed, and it returns at the next drop. The state-root route test (tests/test_pane_registry.py:471-493) checks the shim's `APP`, `LOADEDPV` and `notePanes` and never `NOSTALE`, and the served lab never drops a socket. Serve the branch with `_shim(pid, _dist_ver(), no_stale=True)` and pin `var NOSTALE=true;` in `/pane/<id>/shim.js` beside the Files pin (tests/test_files_pane.py:112).

**The split script's tabDrag-off branch and two palette listeners take a URL pane's post before any source check, and the census test cannot see the gap.** The commit says eleven shell handlers read `window.__rompPaneSourceOk` (defined at kernel/kernel.py:60461), and the count is right: of the landing's thirteen window message listeners, eleven carry the guard and `_LANDING_FOCUS_JS`'s activeTab handler (60392) carries its own source and origin check. `_LANDING_SPLIT_JS`'s handler (63877) carries neither: its first branch at 63878, `if(m.romp==='tabDrag'){if(!m.on){drag=null;unmountZones();return;}`, runs before `frameOfWin(e.source)` at 63879, so a URL pane, the sandboxed cross-origin frame this PR puts in the tree for the first time, can post `{romp:'tabDrag',on:false}` and cancel a live tab drag, unmounting the drop zones under the pointer. The docking kit's twin reader was guarded by this PR (`protocolFrame(e)` at ui/webview/panedock-main.ts:437, ahead of its `tabDrag` branch at 444), so the two shell-side readers now disagree. In ui/webview/palette-main.ts, the `openKeys` listener at 294 opens the shortcuts modal for any sender, and the `hotkeyConfigure` listener at 315-318 calls `configureHotkey` with the posted sid and `e.source`, which writes the tab key to localStorage and registers the hotkey (306-307) before opening the chord dialog (308), then focuses the asking window and posts `focusComposer` into it (312). The census test (tests/test_pane_registry.py:402-404) pins `KSRC.count(GUARD) >= 11` and seven constant names; `_LANDING_SPLIT_JS` is in neither, and a future unguarded handler passes the same way. The served lab's forged set (tests/test_pane_registry_served.py:77) is `toggleFleet`, `notify` and `reveal`. Put the guard first in the split handler, ahead of the `!m.on` return; guard both palette listeners with `w.__rompPaneSourceOk` when present; and replace the count with an enumeration that walks every `addEventListener('message'` in the landing's inline constants and `_STALE_JS`, asserts each handler's first statement is the guard or the activeTab check, and asserts the walk found the split handler by name. Add `openKeys`, `hotkeyConfigure` and `tabDrag` to the lab's forged set and assert no dialog opened and no drag ended.

**A pane-set proposal and a build proposal share the reload core's `seen.code` slot, so each overwrites the other's standing offer.** `notePanes` (kernel/kernel.py:58944) proposes `'panes:'+pv` through `propose` (58934-58935), the same path `noteVersion` (58945-58946) uses for the kernel's code identity; `propose` keeps one `seen.code`, and any change of it with no decline on record replaces `offered` and re-renders, and `offerState` (58932) picks the panes wording whenever `offered.code` starts with `panes:`, ahead of `behindSeen`. The core baked with `CODE="C0"` and `PANES0="P0"` in node, over the registry test's own stub: `noteVersion({code_ident:'C1'})` stands the build offer, `behind()` sharpens it to the behind-the-kernel wording, one `notePanes('B')` rewrites the banner to "The set of panes changed. Reload to see it." with the behind flag still set and its sentence gone, the next `noteVersion` flips it back, and the next `notePanes` flips it again. The shim calls `notePanes` on every keepalive whose `pv` differs (59255), so a page with both changes alternates within one heartbeat of every reopen. The decline record is shared too: `dismiss` (58939) stores the `{dv, code}` of the moment, and in the same harness a build declined, then a panes offer declined, then a reopen re-offers the build the user already turned down, because `declined` (58931) compares only the newer record. The reload driver (tests/test_pane_registry.py:533-539) calls `notePanes` twice on a fresh core; nothing drives `noteVersion`, `notePanes` and `dismiss` together. Give the revision its own slot: `seen.pv`, `offered` carrying `code` and `pv`, `offerKey` and `declined` reading both, and the panes wording only when the revision is the sole new information. Add the two scenarios above to the driver and assert the build wording stands and the declined build stays declined.

**The "byte-identical to today's landing" claim is broader than what the first test pins.** The commit message, the PR body's first paragraph, the section comment (kernel/kernel.py:55547), the `_panes_attr` docstring (55739), the plan (plans/panes-as-data.md:126, 138, 264, 282) and the test docstring (tests/test_pane_registry.py:5-6) say the empty-registry landing is byte-identical to the landing before the registry. `test_00` (220) asserts `h0 == km._landing()` at 225, head against head, the absence of the registry markers at 228-229, and a define-then-remove round trip at 234; nothing references the base's bytes, and nothing could. `_landing()` rendered with an empty registry from the base tree and from the head: 421,873 bytes against 427,557, 5,684 added, all inside the 21 inline `<script>` blocks: 23 occurrences of `__rompPaneSourceOk`, 5 reads of `data-panes`, 3 of `PANES0` and 5 of `notePanes`, none of them at the base, and with the script bodies blanked the two pages are identical; the diff adds 12 `__rompPaneSourceOk` lines to kernel.py unconditionally. A reader taking the pin at its name, say to skip the landing labs, is misled, and the pin cannot notice a later change to the empty-registry markup while the named markers stay absent. Reword the test's name and docstring, the commit message, the body, the comments and the plan to what holds: no attribute, no registry markup, no registry CSS and no added script with an empty registry, and a define followed by a remove restoring the bytes, while the inline scripts each gained lines. If an identity guarantee is wanted, pin the body markup and the pane CSS block against a golden taken from the base.

## Smaller points

- `_panes_rev` (kernel/kernel.py:55683) digests each file's `(name, st_mtime_ns, st_size)` at 55695, and `define_pane` always writes through `_atomic_write` (55709), whose `os.replace` gives the file a new mtime. Executed on a hermetic kernel: a second `define_pane` of a byte-identical record moves the revision, so does a bare touch, and so does a malformed `bad.json` that `_panes_data` skips and names; every open dashboard is then offered "The set of panes changed" with nothing to see. This is a design point against the repo's rule that a card, or here a banner, moves only on new information, not a defect in the offer's mechanics; a provisioning script re-applying the same definition is the ordinary case. Digest the checked set `_panes_data` returns instead of the stats, or skip the write when the stored record equals the new one, and pin that a re-define of an identical record leaves the revision unchanged. The revision test (191) pins moves on a define and a remove only.
- The prose says five shipped panes; the code has six. `_CODE_PANES` (kernel/kernel.py:55608-55615) holds `chat`, `timeline`, `fleet`, `feed`, `files` and `artifacts`, the test's `SHIPPED` list (tests/test_pane_registry.py:43) has the same six and `_PANE_RESERVED` (55552) includes `artifacts`, yet the commit message says "The five shipped panes are the records in _CODE_PANES" and "the five then the data panes", the section comment (55547) and the `_panes_attr` docstring (55739) say a five-constant landing, the module docstring (5-6) and two test names (97, 146) say five, and bin/romp's reserved list (1214-1215) omits `artifacts`. The `romp pane list` fallback at bin/romp:1254, "no panes defined (the five shipped panes are built in)", is unreachable: the `GET /panes` handler (kernel/kernel.py:66248) always prepends the builtin rows, and the snippet fed those six rows prints six lines. The PR body itself says the sixth pane is the sixth record. Say six, or "the shipped panes", throughout; add `artifacts` to the bin/romp list; drop the dead branch. The "five baked inline scripts" figure is a different count and stands.

lczh:
The test runs behind the comment above, on 91af1bdc (merge base with main 88cb4d08).

## Tests

- tests/test_pane_registry.py in the worktree at the head, run alone: 15 passed in 1.15s. No other module was run.

- gear.js's `panesOf` widened to keep boolean members beside the three, the `paneSet` rule: the same Outline uncheck leaves `{timeline:true,fleet:false,feed:true,lab:true,notes:false}`, and `optOn` keeps `lab` true and `notes` false; the head's `panesOf` drops both.
- The `notes` shim core built with `no_stale=True`: the same drop, redial and reopen arm nothing (`stalePending` stays empty), and twenty-two keepalives post no `wsStale`; the head's core posts one on the second keepalive.

## Probes

- Gear: the exact `panesOf` (ui/webview/gear.js:464), the shipped-row handler (465), the registry-row handler's closure (485) and `optOn` (kernel/kernel.py:63548-63550) in node over a fake `localStorage`, with `load` returning the stored object and `save` writing it. Store `{timeline:true,fleet:true,feed:true,lab:true,notes:false}`; Outline unchecked; store `{timeline:true,fleet:false,feed:true}`; `optOn` `{timeline:true,fleet:false,feed:true,lab:false,notes:true}`. For contrast, the registry-row handler re-saving Lab keeps all five keys.
- Shim strings: a hermetic kernel loaded as the tests load it: `_shim("notes", 7)` contains `var NOSTALE=false;` and not `var NOSTALE=true;`; `_shim("files", 7, no_stale=True)` contains `var NOSTALE=true;`.
- Shim over a fake socket: the `notes` shim core between its anchors, in tests/test_pane_shim_return.py's harness, which fakes `Date.now`, the timers, `WebSocket` and `MessageChannel` and records the parent's posts: open, two keepalives, `readyState=3` and an unclean `onclose`, the one live `connect` timer fired, a second socket opened. `stalePending` reads `"reconnect"`; the first keepalive posts nothing; the second posts `wsStale`; twenty more post nothing; the diagnostics rows on the second socket are `wsclose` and `stale-raise`; no `wsFresh` at any point.
- Handler census: `addEventListener('message'` occurs 14 times in kernel/kernel.py; one is the service-worker container listener at 63328. Of the 13 window listeners, 11 carry `__rompPaneSourceOk`: the boot, Outline bridge, usage, settings, remotes, mobile, reveal and stale handlers and three in the errors script; `_LANDING_FOCUS_JS` at 60392 checks `e.source===window||e.origin!==location.origin` inline, and `_LANDING_SPLIT_JS` at 63877 checks nothing before its `!m.on` return at 63878. `__rompPaneSourceOk` occurs 12 times in the file: the 11 guards and the definition at 60461. ui/webview/palette-main.ts has no occurrence; its two listeners at 294 and 315 are unguarded. ui/webview/panedock-main.ts checks `protocolFrame(e)` at 437 before its `tabDrag` branch at 444.
- Reload core: `_reload_core_js(3, boot="b1", code="C0")` with `PANES0` set to `"P0"`, over the registry test's `_RELOAD_STUB`. Scenario one: `noteVersion({code_ident:'C1'})` offers `C1` with the build wording; `behind()` switches to the behind wording; `notePanes('B')` offers `panes:B` with the panes wording, `behind` still true; `noteVersion` again offers `C1`; `notePanes('B')` again offers `panes:B`; no reload fired. Scenario two: `noteVersion` then `dismiss` stores `{dv:0,code:"C1"}`; `notePanes('B')` offers; `dismiss` stores `{dv:0,code:"panes:B"}`; `noteVersion({code_ident:'C1'})` offers `C1` again with the build wording.
- Landing render: `_landing()` with an empty registry, from the base tree extracted with `git archive` and from the head, each loaded as the tests load a kernel: 421,873 bytes against 427,557; 21 `<script>` blocks in both; `__rompPaneSourceOk` 0 against 23, `getAttribute('data-panes')` 0 against 5, `PANES0` 0 against 3, `notePanes` 0 against 5; neither carries ` data-panes="`; with every script body blanked the two are identical. `git diff 88cb4d08 91af1bdc -- kernel/kernel.py` adds 12 lines containing `__rompPaneSourceOk`.
- Revision: on the hermetic kernel, `_panes_rev()` is `"0"` with no panes; `define_pane` of the notes record, then the same record again back to back: bytes identical, mtime moved, revision moved; again 20 ms later: moved; `os.utime` alone: moved; a `bad.json` holding `{not json`: revision moved, `_panes_data` omits it and `_pane_order` is unchanged.
- Five against six: `_CODE_PANES` has six ids and `_PANE_RESERVED` seven, `artifacts` in both; the commit message carries "five shipped panes" and "the five then the data panes"; kernel/kernel.py:55547 and 55739, tests/test_pane_registry.py:5-6, 97 and 146, and bin/romp:1214-1215 and 1254 carry the figure or the short list. bin/romp's `list` snippet fed the handler's six builtin rows prints six lines; fed an empty list it prints the "no panes defined" line the handler never produces.

henrypinkard:
**Round two** (the verifier's HIGH, M1, M2 and lows a, b, c, e; d stated in the body), with the reds measured again.

**The registry module at the merge base 4d977bed** (product files at the base, the rewritten tests as they are): 14 of 14 red, every one on behaviour, no missing name:
- the rendering slices pin: the six slices EQUAL the base's (as they must) and the red is "a pane file changes the page" (at the base it does not);
- the door tests: `GET /panes` and `POST /pane` answer 404 where 200 is expected; define/replace/remove `(404, None) != (200, True)`; the state-root page `(404, text/plain) != (200, text/html)`;
- the store through the landing: the rail's last button is `('artifacts', 'Artifacts')` where `('notes', 'Notes')` is expected; the escaped title's button missing; the keepalive frame `('ka', dv, None)` where `pv` is `'0'`;
- the landing with three data panes: the rail lacks `docs`, `lab`, `notes`;
- the pane controller executed with the attribute: `po-notes` never set;
- the source check executed: forged toggles `2 != 0`, the check `undefined`; the fail-closed test: toggles `1 != 0`; the census: the open registrations listed;
- the reload core: `notePanes` absent, the offer `None`.

**The HIGH's legs at 91af1bdc** (phase one's head):
- the served lab, with the forged legs asserted first: the shortcuts modal OPEN (`keysOpen: True`), `romp:tabkeys` holding the nested frame's chosen session under `"nested"`, the Outline open (`fleet: True`), and "forged from a nested frame" in the notification center;
- the executed unit tests: the fail-closed test (a listener on a page without the check acted: toggles `1`), the nested-frame test (a frame nested in a protocol pane toggled: `1 != 0`), the census (the `activeTab` and tab-drag registrations without the check);
- `ui/webview/pane-source.test.ts`: red by absence (the module did not exist; esbuild could not resolve it, which also held the docking unit test's build, whose `speaksProtocol` test is unchanged from round one).

**What changed for it**: the inline guard is `if(!window.__rompPaneSourceOk||!window.__rompPaneSourceOk(e))return;` on all thirteen inline registrations (`activeTab` and the split's tab drag joined; the fourteenth is the service worker's own channel); the check rules on the message's IMMEDIATE source frame (the parent walk is gone; the plan's section 5 says why); the bundles read it through `ui/webview/pane-source.ts` (`paneSourceOk`, fail-closed: no check refuses, only the boolean `true` admits) in the palette's `openKeys` and `hotkeyConfigure` listeners and in the docking kit's registration and frame lookup; the census reads the built landing, the shell bundles' sources and the built dist bundles. Nine test harnesses that dispatch shell messages to executed inline scripts now define the check as the boot script would.

**M1**: the claim is "the code panes' rendering is unchanged" in the body and the plan; `tests/landing_slices.py` slices the body tag, the rail, the phone tabs, the pane row, the column and gutter rules (from the style blocks) and the gutter calls, and `tests/fixtures/landing-code-panes.json` holds the base kernel's slices (made from a detached checkout of 4d977bed; the file names the commit and the helper's docstring says how to regenerate).

**M2**: pane files are written by hand the way the door writes them, and the landing, the routes, the keepalive frame and the executed scripts are what is asserted.

**Lows**: (a) the unreachable "five shipped panes" line reworded; (b) `_panes_snapshot` lists the directory once and yields the records and the revision together; `_landing` lists once per build and hands the list to every builder; (c) the gutter chains (`_data_pane_css`) and the data panes' gutter sequence (`_LANDING_JS`) derive from `_CODE_PANES`; (e) `activeTab` reads the one function; (d) the kit-on 0x0 fact is in the body under phase two.

Whole suites on this head: the extension suite 5164 tests, 5157 passed, 6 skipped, 1 failed (a source pin on the old activeTab check in `ui/webview/feed-focus-local-switch.test.ts`, updated to the one check's text and re-run green alone; the fix is the pin's text only); the Python suite 14726 passed, 57 skipped, 2341 subtests passed, with the one pre-existing red the body names (the drop-anywhere lab's kernel never serving its health route on this machine, on the shared main checkout too); the held-mail lab was green this time.

henrypinkard:
e9753f60: the `romp pane list` bats expectation follows the reworded empty-list line (low a); the whole shell suite, 723 tests, passes locally with the pinned bats. No product change.

lczh:
Round two, executed against d1ccedaf.

Two of the five round-one items landed, the source check and the rendering claim, each with more than was asked: one fail-closed guard on every shell listener, inline and bundled, an enumerating census, and a slice-by-slice pin against a fixture from the base. Three items and both smaller points are unchanged. One new low: the landing reads the registry twice per build, and the two reads can disagree.

## Landed

**The split handler, the palette listeners and the census (3).** The split script's listener at kernel/kernel.py:64374 now opens with the guard, ahead of the `!m.on` return at 64375, and the same `if(!window.__rompPaneSourceOk||!window.__rompPaneSourceOk(e))return;` opens every other inline listener. Executed over the built landing: 14 registrations of `addEventListener('message'`, one on the service worker's channel, the guard the first statement of all 13 window listeners, no double-quoted registration and no `onmessage` assignment. The palette reads `paneSourceOk` at ui/webview/palette-main.ts:295 and 316-317, the kit at ui/webview/panedock-main.ts:214 and 430; ui/webview/pane-source.ts:7-11 refuses when the check is absent, throws, or answers anything but the boolean `true`. The check (kernel/kernel.py:60956-60958) executed in node over fake frames: a direct protocol frame and a client-made split column pass; a none frame, a frame nested in either kind of pane, a removed frame, a null source, the top window itself, a foreign or sandboxed origin all fail; a throwing `querySelectorAll` fails closed. Reading the code, the origin test alone refuses a URL pane and everything under it (a sandboxed frame's children inherit its opaque origin); the direct-frame rule adds the shell's own window and same-origin nested frames, which could script `top` directly anyway, so against same-origin content the check is discipline, not a boundary, as section 5 says. Two residuals. The census (tests/test_pane_registry.py:532) accepts the guard anywhere in the 200 characters after a registration, not as the first statement its message claims; today every listener has it first. And in CI the test's last leg never runs: the pytest job (.github/workflows/ci.yml:87-102) does not build the extension, so the `skipTest` at line 555 fires and the whole test is recorded as skipped in every cell, and the browser job (259-265, 330) runs only the served and browser modules. The inline census executes before the skip, but no CI run reads the built bundles, and the pass is invisible. Give the bundle leg its own test, and build in the pytest job or add the module to the browser job's list.

**The rendering claim (5).** The body, the plan, the test docstring and the test's name now say the code panes' rendering is unchanged, and tests/test_pane_registry.py:150 pins it against tests/fixtures/landing-code-panes.json. The fixture says `made_from: 4d977bed`, the merge base; the slicer run over the base tree's `_landing()` with an empty registry reproduces all six slices exactly, on a page of 422,573 bytes with 21 script blocks and no registry name. Outside the slicer by design: the inline scripts, and the timeline band's markup (its rules are in the CSS slice). Three small things. The `column_css` pattern (tests/landing_slices.py:23-24) matches `-panel` too, so four unrelated panels' rules ride in the code panes' slice (35 rules in all) and a restyle of any of them fails a pin named for the panes. Its lookbehind skips a rule at the very start of a style block (executed on a toy string; no pane rule sits there today). The section comment at kernel/kernel.py:56039-56040 still says "byte-identical to the five-constant landing", as does the phase-one commit message of dfab815c.

## Still open

Unchanged at this head:
- (1) `panesOf` (ui/webview/gear.js:471) still returns `{ timeline, fleet, feed }`, and the shipped rows' handler at 472 still does `s.panes = panesOf(s)`; only the registry rows' handler at 492 starts from `st.panes`. Re-executed with the shell's `optOn` (kernel/kernel.py:64045-64047): the store `{timeline:true,fleet:true,feed:true,lab:true,notes:false}` becomes `{timeline:true,fleet:false,feed:true}` on one Outline uncheck, and `optOn` then reads `lab` false and `notes` true.
- (2) The `/pane/<id>/shim.js` branch (kernel/kernel.py:67125) still calls `_shim(pid, _dist_ver())` while the Files, Artifacts and Settings pages pass `no_stale=True` (60416, 60445, 60608); `_shim("notes", 7)` bakes `var NOSTALE=false;`. The state-root test (tests/test_pane_registry.py:269) checks `APP`, `LOADEDPV` and `notePanes`, not `NOSTALE`.
- (4) `propose` keeps one `seen.code` (kernel/kernel.py:59429-59430), `notePanes` (59438) proposes `'panes:'+pv` through it, and `offerState` (59426) and `declined` (59425) read that slot. Re-executed: a build offer sharpened to the behind wording, one `notePanes('B')`, and the banner reads the panes wording; the next `noteVersion` flips it back and the next `notePanes` flips it again; a dismissed build, then a dismissed panes offer, and the build is re-offered. The driver at 593 still drives `notePanes` alone.
- (a) `_panes_snapshot` digests `(name, mtime_ns, size)` (kernel/kernel.py:56147-56148) and `define_pane` (56195-56207) always writes. Re-executed: an identical re-define leaves the bytes equal and moves the mtime and the revision; so does a touch; so does a `bad.json` that `_panes_data` skips.
- (b) bin/romp:1254 is reworded (and still unreachable, harmlessly). bin/romp:1214-1215 still lists the reserved ids without `artifacts`; kernel/kernel.py:56040 and plans/panes-as-data.md:56 still say five.

## New at this head

- `_landing` lists the registry once at kernel/kernel.py:64719 and hands `panes` to every builder, but `_stale_block(v)` at 66032 reaches `_reload_core` (59467-59473), which reads `_panes_rev()` again. Executed with a define landing between the two reads: the page carries no `data-panes` and no Notes button, `PANES0` is the new revision, and the keepalive's `pv` equals it, so the page shows the old set and is never offered a reload until the next change. The window is one request wide; the fix is handing the snapshot's revision to `_stale_block`. `_shim` (59925) has the same two reads.
- The nine harnesses that now define `__rompPaneSourceOk` as a function answering `true` stand for a protocol pane's posts and cannot hide a missing guard; the census and the fail-closed test (tests/test_pane_registry.py:517) catch that. `federation.ts`, `render.ts` and `send-pending.ts` are not in the round-two commit.

## Runs

- tests/test_pane_registry.py alone at d1ccedaf, serialized behind the machine's test lock: 13 passed, 1 skipped in 2.19 s (the skip is the census test's bundle leg at line 555; the worktree has no built extension). No other module, no npm, no browser lab.
- Probes on a kernel loaded the way the module loads it, in node and Python: the gear's exact `panesOf`, handler and `optOn`; `_shim("notes", 7)`; the revision on a re-define, a touch and a bad file; the landing built with a define inside `_stale_block`, then the keepalive frame; the census over the built landing; the reload core (`_reload_core_js(3, boot="b1", code="C0")`) over the module's stub; the source check over fake frames; the slicer over the base tree's landing.
- The whole-suite figures in the round-two comment are the author's runs, not repeated here.

PR #1920, opened 2026-09-20
Panes as data, phase two: the docking kit reads the pane list

henrypinkard:
Phase two of panes as data (`plans/panes-as-data.md`, section 4 and section 7): the docking kit reads the pane list. A pane the shell renders from the registry, a data pane or the Artifacts pane, is a tree leaf under the kit like the shipped four: it takes its place from the row's order, its weight from its own grow key, its title from its record, and it drops into a half-zone. The phase-one fact (a data pane toggled on under the kit rendered 0x0 while its rail button read on) closes.

## What changes

- **`ui/webview/pane-dock.ts`** (the pure half): `growKey(id)` is the pane element's id without `-pane` (the shell's naming rule for every pane it renders: `chat`, `fleet`, `feed`, `files`, `artifacts`, a data pane's own id; `chat<n>` for a column stays). `reconcileShown` opens missing panes in the ROW's order, which is the rail's, read off the DOM, instead of a fixed list plus a tail. `seedLayout` gives a pane the grow store never named the MEAN of the named weights (the shipped OFF path gives a new pane 40 beside 34 to 60; a weight of 1 beside those seeded an 8 px column, which the lab found). `defaultDock`'s right-end fallthrough now names what lands there. `ROW_ORDER` stays exported for the shipped four but is read nowhere.
- **`ui/webview/panedock-main.ts`** (the engine): `shown()` walks every `.col .pane` in document order and takes each whose `po-<key>` class is on (the pane controller's truth), the chat and its columns first as before. `paneTitle(id, titles)` names a pane by the records' map, which `titleMapOf` builds from the rail's buttons (every pane the kernel rendered, by rail key) and the body's `data-panes` rows; the shipped four keep their words without a map. The live outline reads it.
- **`ui/webview/pane-grab.ts`** (the detector): an element carrying `data-pane-empty` is a pane page's own empty background, honoured for ANY app before the per-app lists (which stay as the shipped panes' declarations); a control inside still yields; `install` runs for any app, so a state-root pane's page gets the hand and the forward by declaring one attribute. The chat stays list-less and the shell never injects the detector into it. A `protocol: "none"` pane has only the ring (phase one's exclusion, unchanged).

## Tests

- `ui/webview/pane-dock.test.ts`: the grow key for the Artifacts pane, a data pane and a column; a data pane weighted by its own key; data panes opened at the right end in the row's order, parked when toggled off; the fair seed weight for a pane the store never named (60 : 40 : 50), equal weights with no store.
- `ui/webview/panedock-main.test.ts`: `titleMapOf` merges the rail's words with the records' titles and drops a keyless button or a malformed row; `paneTitle` names a data pane, the Artifacts pane, the band, a column, and the shipped four without a map.
- `ui/webview/pane-grab.test.ts`: `data-pane-empty` grabs for an unknown app, a control inside yields, a shipped pane may declare too, the chat's list stays absent; `install` wires a page for an app outside the lists (the style, the flag, the three listeners, the hook).
- `tests/test_pane_registry_served.py` (Chromium, REQUIRE in CI), a sixth leg: the kit turned on in the store and the page reloaded; the notes (state-root), docs (URL) and runtime-defined panes are leaves in the rail's order with the 6 px ring; the notes page's `data-pane-empty` surface wears the hand and a press there arms the shell's press; a REAL pointer drag from the notes ring into the feed's left half shows the feed's left zone and the outline named "Notes", and the drop docks notes left of the feed in the layout store; the Artifacts pane toggled on is a leaf with a rectangle.

Red-first at the base f6bd0c4a per test on behaviour (the kit's three product files at the base, the tests as written): see the first comment for the measured reds. Whole extension suite with the served labs under REQUIRE and the whole Python suite on the head before the push.

## Roads not taken

- The kit's shipped constants (`CHAT`, `FLEET`, `FEED`, `FILES`, `BAND`) stay: the default-dock rules for the outline and the feed are the shipped panes' own; only the lists that enumerated panes went.
- The band and the phone (a second fixed-px kid, experimental panes on the phone) are the plan's phase four.
- Phase three (the Artifacts record folded into data, its hand hooks deleted) is joint with the feed owner.

Tier: feature


henrypinkard:
**Measured reds at the base f6bd0c4a** (the kit's three product files, `ui/webview/pane-dock.ts`, `pane-grab.ts` and `panedock-main.ts`, checked out at the base; the tests as they are on this branch, reading the new exports off the module namespaces so the build succeeds at the base and each test reds on its own behaviour):

- `pane-dock.test.ts`: `growKey` red (a data pane's and the Artifacts pane's keys came back as their element ids); `seedLayout` and `reconcileShown` with data panes red (the weights read off the wrong keys); the fair seed weight red (a pane the store never named weighed 1 beside 60 and 40).
- `pane-grab.test.ts`: `data-pane-empty` red (`false` where `true`: the declared surface was not a grab for an app outside the lists); `install` red (the wired flag never set for an unknown app).
- `panedock-main.test.ts`: `paneTitle` red (`'notes-pane'` where `'Notebook'`: a data pane named by its id, not its record).
- `tests/test_pane_registry_served.py`, the kit leg: the driver died at the base with `notes or feed is not a leaf: ["chat-pane","feed-pane","tl-pane"]`: with the kit on, the three data panes toggled on were no leaves (the phase-one 0x0 fact), so nothing after (the ring, the declared surface's press, the drag into the half-zone, the Artifacts leaf) could be measured there.

On this head: the three unit modules 31 tests green; the lab green with the kit leg (leaves in rail order with the 6 px ring, the hand and the press on the page's declared surface, the feed's left zone and the outline named "Notes", the drop docking notes left of the feed and persisting, the Artifacts pane a leaf). Two screenshots in the lab's drops, the kit off and on.

Whole suites on the head before the push: the extension suite 5170 tests, 5164 passed, 6 skipped, 0 failed (exit 0); the Python suite 14726 passed, 57 skipped, 2341 subtests passed, with the one pre-existing red on this machine (`tests/test_drop_anywhere_served.py`'s lab kernel never serving its health route, on the shared main checkout too; named in PR 1919).

lczh:
Evidence executed against 3a66fd27.

This is a small change and it does what it says: the three unit modules pass at the head, the grow key, the fair seed weight and the record-named outline all hold, and the one thing standing between the feature and its stated job is the shell's frame classification, which still sorts a frame into the chat by the prefix of its iframe id, so a registry pane whose id begins with `chat-` never gets the detector the PR promises. The rest are a test whose title claims more than its assertions pin, a pane-id namespace the kit's generic walk now leans on, and prose that still describes the list this PR removed.

## At this head

**A data pane whose id begins with `chat-` is classified as a chat column, so it never gets the grab detector and receives the chat's cursor stylesheet instead.** `install` now runs for any app (ui/webview/pane-grab.ts:74-82) and the body says a state-root pane's page gets the hand and the forward by declaring `data-pane-empty`. But the shell decides which frames get the detector, and it decides by iframe id: `markDoc` (ui/webview/panedock-main.ts:436-443) returns at 440 on `f.id === "f-chat" || f.id.indexOf("f-chat-") === 0 || d.getElementById(GRAB_SCRIPT_ID)` before appending `pane-grab.js` at 441-442, and `wire` at 410-411 injects `PANE_CSS` (the chat strip's `#tabbar` rule, 139) as `pd-css` into every frame matching the same prefix. The registry renders a data pane's iframe as `f-<id>` (`_data_pane_markup`, kernel/kernel.py:56245) and its id check admits the prefix: `_PANE_ID_RE` (56043) is `^[a-z][a-z0-9_-]{0,31}$` and `_PANE_RESERVED` (56045) is seven exact ids. On a hermetic kernel, `_pane_check({"id":"chat-notes","source":"pane:chat-notes"})` returns the filled record, `define_pane` writes it, and `_data_pane_markup()` renders `<div class=pane id=chat-notes-pane><iframe id=f-chat-notes data-src="/pane/chat-notes/" data-protocol=romp>`. In node, the predicate at 440 is true for `f-chat-notes`, while `isChatPane("chat-notes-pane")` (ui/webview/pane-dock.ts:39) is false and `growKey` gives `chat-notes`, so `shown()` (panedock-main.ts:301-302) keeps the pane as an ordinary leaf. Reading the code from there: the frame speaks the protocol, so `markDoc` marks its body and returns; no script tag, no `__rompPaneGrab`, no hand, a press on its declared surface arms nothing, the ring still works, and the page carries a style written for the chat's strip. The exclusion exists for the chat's transcript, where a press-drag is a text selection (pane-grab.ts:11-13), not for any id that starts with the word. No test sees it: panedock-main.test.ts covers `isPaneDockingOn`, the body class, `speaksProtocol` and `paneTitle`; the registry tests pin exact reserved ids (tests/test_pane_registry.py:228-230); the lab's panes are `notes`, `docs` and `later`. Classify the frame by what the split script actually names, `f.id === "f-chat" || /^f-chat-\d+$/.test(f.id)` (`frameId(n)` is `'f-chat-'+n`, kernel/kernel.py:64137), as an exported pure predicate beside `speaksProtocol` used at both 410 and 440, pinned in panedock-main.test.ts with `f-chat-notes` false and `f-chat-2` true; and have `_pane_check` refuse ids beginning `chat-`, pinned in tests/test_pane_registry.py, which also closes the `chat-2` twin below.

**The test that says data panes open in the row's order does not pin the row-order read.** `reconcileShown` at ui/webview/pane-dock.ts:193 now takes `missing` from `sh.row`; the base took it from `ROW_ORDER.concat(sh.row.filter((p) => !ROW_ORDER.includes(p)))`. The new test (ui/webview/pane-dock.test.ts:204-217) opens `[chat, feed, artifacts-pane, docs-pane, notes-pane]` from a base of `[chat, feed]`; the three new panes sit outside `ROW_ORDER` in both forms, so both open them in the same sequence. Executed on scratch copies: with line 193 reverted to the base's two lines, pane-dock.test.ts is 19 of 19 and panedock-main.test.ts 4 of 4, the head's figures; with the whole base pane-dock.ts under the head test, this test fails only at the width map (209), and with that one assertion removed the leaf order (213), the `defaultDock` fallthrough (214) and the parking (215-216) all pass on the base. So what the test pins is the grow key and the weights, which is what the measured-reds comment already says, and the order it asserts is the base's behavior under this row. Under the shipped kernel the two forms are hard to tell apart, since `_pane_order` (kernel/kernel.py:56183) renders the shipped panes before the data panes, but an interleaved row does distinguish them: from a base of `[chat]`, the row `[chat, notes-pane, files-pane]` opens as `[chat, notes, files]` at the head and `[chat, files, notes]` under the reverted line. Either add that row to the test so the read has a pin, or retitle the test to what it checks and say in the body that the `reconcileShown` change is a simplification with no observable difference under the kernel's rail order.

## Smaller points

- The kit's generic walk now depends on a pane-id namespace phase one left open; this predates the PR and reads as a follow-up. `_pane_check` accepts `tl`, `a` through `d`, `ghost` and `chat-2` (executed), and the shell derives `<id>-pane`, `gv-<id>` and `f-<id>` from the id. A real `_landing()` with a `tl` pane defined carries `id=tl-pane` twice, the data pane's inside the row ahead of the band's (kernel/kernel.py:65804), and `id=gv-ghost` twice. Reading the engine from there: `shown()` skips the id at panedock-main.ts:301, so the data pane is no leaf, and `byId` (146), used at 306 and for every geometry write at 346 and 375, returns the first element with the id, the data pane's. `_data_pane_css` (56250) also emits `body:not(.po-tl) #tl-pane{display:none}`, a rule that matches the real band too. Refuse the derived names in `_pane_check` (`tl`, `a` through `d`, `ghost`, anything beginning `chat-`) with a registry test, here or in the next PR.
- The prose still describes the removed list. plans/panes-as-data.md:185-186 says `ROW_ORDER` is the column panes of the list in rail order and plans/pane-docking.md:460-461 says the kit's `ROW_ORDER` reads the list; the diff touches nothing under plans/. `ROW_ORDER` (ui/webview/pane-dock.ts:33) is exported and read nowhere (`git grep` finds the export and the two plans). The `Shown` doc comment at 127-128 still enumerates the outline, feed and files as toggled on. Amend the two plan lines to the DOM read, describe every row pane in document order in the `Shown` comment, and drop `ROW_ORDER` or say why it stays.
- The install test does not pin the style it is said to cover. The Tests bullet lists the style, the flag, the three listeners and the hook; the fake document's `head = { appendChild() {} }` (ui/webview/pane-grab.test.ts:89) records nothing, `getElementById` returns null (90), and the assertions at 94-97 read the flag, the app, the listener names and the hook. On a scratch copy with the `ensureStyle(doc)` call at ui/webview/pane-grab.ts:80 deleted, pane-grab.test.ts stays 8 of 8. The served lab does pin the cursor end to end (tests/test_pane_registry_served.py:381), so the product is covered; record the appended node in the fake and assert its id is `pd-grab-css` and its text contains `cursor:grab`, or drop the style from the bullet.

lczh:
The test runs behind the comment above, on 3a66fd27 (merge base with main f6bd0c4a).

## Tests

- The extension suite at the head, `npm test` (the test bundle, then `node --test` over it): 5170 tests, 5163 passed, 1 failed, 6 skipped. The one failure is the KaTeX layout probe (test 2790), which fails on main in this environment and touches nothing the PR changes.
- The three modules the PR touches, each run alone from the same build: pane-dock.test.js 19 of 19, pane-grab.test.js 8 of 8, panedock-main.test.js 4 of 4. No Python module was run; the served labs were not run here.

- ui/webview/pane-dock.ts with line 193 reverted to the base's two lines (`const order = ROW_ORDER.concat(sh.row.filter((p) => !ROW_ORDER.includes(p)));`, then `missing` filtered from `order`), the copy bundled the way the test build bundles (cjs, node, node18, typescript external): pane-dock.test.ts 19 of 19, panedock-main.test.ts 4 of 4, the head's figures.
- The base's whole ui/webview/pane-dock.ts (f6bd0c4a) under the head's pane-dock.test.ts: 16 of 19; the reds are the `growKey` test, the data-pane seed and reconcile test, and the fair-weight test. The same with the test's line 209 (the width map) deleted: 17 of 19; the data-pane test passes, so its leaf order, `defaultDock` and parking assertions hold on the base module.
- ui/webview/pane-grab.ts with the `ensureStyle(doc);` call at line 80 deleted (the export kept, zero call sites): pane-grab.test.ts 8 of 8, the install test included.
- An unmutated scratch copy bundled the same way, as the control: 19, 8 and 4, matching the checkout's build.

## Probes

- Registry: a hermetic kernel loaded from the head under a scratch state root, as tests/test_pane_registry.py loads one. `_pane_check` with source `pane:<id>` accepts `chat-notes`, `chat-log`, `chat-2`, `tl`, `a`, `b`, `c`, `d` and `ghost` (each returned as a filled record, protocol `romp`) and refuses `chat` and `fleet` as reserved. `define_pane` of `chat-notes` writes the file; `_data_pane_markup()` then renders `<div class=gv id=gv-chat-notes></div><div class=pane id=chat-notes-pane><iframe id=f-chat-notes data-src="/pane/chat-notes/" data-protocol=romp></iframe></div>`; `_pane_order()` lists it after the six shipped ids.
- Landing: with `tl` and `ghost` also defined, `_landing()` is 430,367 bytes and carries `id=tl-pane` twice (offsets 44533 and 44710; the data pane's `<iframe id=f-tl data-src="/pane/tl/" data-protocol=romp>` first, the band's `<iframe id=f-timeline data-src=/timeline>` second) and `id=gv-ghost` twice; `_data_pane_css()` contains `body:not(.po-tl) #tl-pane{display:none}`.
- Frame classification, in node over the head's pane-dock module: `"f-chat-notes".indexOf("f-chat-") === 0` is true, so the predicate at panedock-main.ts:440 and at 410 takes the frame; `isChatPane("chat-notes-pane")` is false and `growKey("chat-notes-pane")` is `chat-notes`, so `shown()` keeps the pane as a leaf. For contrast, `f-chat-2` also matches and `isChatPane("chat-pane-2")` is true. The browser-side consequence (no `pd-grab` tag, no `__rompPaneGrab`, no hand, a press that arms nothing, the `pd-css` style present) is read from `markDoc` and `wire`, not executed; no headless browser was available to this review.
- Row order, in node over `reconcileShown`: from a base of `[chat-pane]`, the row `[chat-pane, notes-pane, files-pane]` opens as `["chat-pane","notes-pane","files-pane"]` at the head; under the reverted line and under the base module alike it opens as `["chat-pane","files-pane","notes-pane"]`. The PR test's row, `[chat-pane, feed-pane, artifacts-pane, docs-pane, notes-pane]` from `[chat-pane, feed-pane]`, opens identically under all three.
- `ROW_ORDER`: `git grep ROW_ORDER` at the head finds ui/webview/pane-dock.ts:33 (the export), plans/panes-as-data.md:39 and 185, and plans/pane-docking.md:460. `git diff --stat f6bd0c4a..3a66fd27` lists seven files, none under plans/: tests/test_pane_registry_served.py and six under ui/webview.
- The install test's fake: ui/webview/pane-grab.test.ts:89 `const head = { appendChild() {} };`, 90 `getElementById: () => null`; the assertions at 94-97 read `__rompPaneGrabWired`, `__rompPaneGrab.app`, the sorted listener names and the hook's answer.

PR #1922, opened 2026-09-20
Panes as data, phase three: the Artifacts pane is an experimental record the generic build renders

henrypinkard:
Phase three of panes as data (`plans/panes-as-data.md`, section 7 item 3), joint with the feed owner: the Artifacts pane is a record in `_CODE_PANES` (`{id: "artifacts", title: "Artifacts", source: "/artifacts", on: false, experimental: true}`) and the GENERIC pane build renders it; every hand-written hook of its first landing is deleted, and its bespoke control is mapped onto `experimental`.

## What changes

- **The generic build renders every pane after the hand-written five** (`kernel/kernel.py`): `_HAND_PANES` names the chat, the Sessions band, the Outline, the Feed and the Files pane, whose landing lines the pins hold; `_generic_panes()` is every other pane of `_pane_order()`, the Artifacts record then the data panes by id. `body[data-panes]` carries them all (a row gains `builtin` for a shipped record), so the inline scripts and the gear's Panes rows see the Artifacts pane exactly as they see a data pane. The pane controller's baked key list is the hand five.
- **The hand hooks come out**: the `artifacts` entries in `PANES`, `grow` and the `key()` fallback, the `gv-d` gutter call and its rules, the focus map's and column list's `f-artifacts`, both frame lists (which now concat the attribute's frames, so a generic pane's document is wired for Escape and the phone fit like the five), the phone's `F` map entry and its tab hook, the `po` literals, the label word, `artifactsCtl`/`artifactsCtlM`, `no-artifacts-control`, `avail.artifacts`, the toggle's refusal, the column CSS and the `f-artifacts` iframe markup. What remains is emitted by `_data_pane_markup`, `_data_pane_css` and the attribute: `gv-artifacts`, `#artifacts-pane`, `<iframe id=f-artifacts data-src="/artifacts" data-protocol=romp>`. The page (`/artifacts`), the listing op and the bundle are untouched.
- **The control maps onto `experimental`** (the feed owner's call): the gear's generic Panes row is the control (off by default for an experimental record; on, the rail button shows and the pane comes on screen); `showArtifactsControl` is deleted from `settings.ts` and `gear.js` (the `rs-artctl` row and its wiring) with no migration (the pane shipped hours ago and defaults off). An experimental record has no phone tab.
- **On a phone a pane shows by its TAB**: the phone's tab switch copies `data-src` to `src` for the shown pane, once, generically (the shipped panes' first landing carried a hand hook for the Artifacts tab and nothing generic; round two's read found a data pane's tab showing a permanently blank pane); the generic panes' phone display rules (`display:contents` for the pane wrapper, `display:block` for the shown tab's iframe) ride the mobile media block beside the hand five's, so a data pane's tab shows its page.
- **A generic pane's iframe loads only when it comes ON SCREEN** (the feed owner's constraint from PR 1911 round two: the Artifacts page's first load walks the remembered session's transcript, and a dashboard load with the pane enabled but off screen did that walk for nothing): the pane controller's reconcile copies `data-src` to `src` for a hand-written optional pane when enabled, as before, and never for a generic one; the apply copies it for a generic pane when its `po` flag is on. The gear's row turning a pane on brings it on screen (the live reconcile), so it loads then.
- **The gear's generic row** says what a shipped record shows (a hint map: the Artifacts pane's words from its first landing's row), says a data pane is defined at the kernel, and names the experimental default only where it applies; the retired `showArtifactsControl` key is dropped from a browser's store at load like the repo's other retired keys.
- **The rendering fixture** (`tests/fixtures/landing-code-panes.json`) is regenerated DELIBERATELY (re-cut again at round two for the phone rules): the body tag carries the attribute always, the phone tabs lose the Artifacts tab, the pane row and the column rules carry `gv-artifacts` in place of `gv-d`, the gutter calls lose the hand `gv-d` line, the rail is unchanged. `made_from` names the base and the fold; the slice-by-slice diff against the previous fixture is in the first comment.

## Tests

- `tests/test_artifacts_list.py`, the `Shell` class rewritten: the record renders generically (rail button, no phone tab, the generic gutter, column, rules and attribute row with `builtin`), no hand hook remains in the landing or in any of the five inline constants, and the load is gated on the pane coming on screen (the reconcile's `DPX` gate, the apply's copy, the experimental default off), with the bespoke key gone from `settings.ts` and `gear.js`.
- `tests/test_pane_registry.py`: the shipped records include `experimental` for the Artifacts pane; the attribute's rows carry the Artifacts record first; the empty-registry pin asserts the attribute holds exactly that record; the fixture comparison against the regenerated slices.
- `tests/test_pane_registry_served.py`, a seventh leg: a phone context (420 by 860) with a data pane's desktop flag off; before the tap nothing is loaded, the tap loads the pane's page and shows it full screen with its shim up, and tapping away and back requests the page once. `tests/test_pane_registry.py`: the phone's tab load and the generic panes' phone rules pinned. `ui/webview/settings.test.ts`: a stored retired key dropped at load and gone after a save. `ui/webview/artifacts.test.ts`: the hint map, the generic row's words and the retired key's drop. `ui/webview/gear-tabs.test.ts`: the ordering pin orders the Files row against Appearance (the re-aimed clause had gone vacuous).
- `tests/test_artifacts_pane_served.py`: the control legs re-aimed to the generic gates (the rail button hidden while the gear has not enabled the pane, the toggle refusing; enabled in the gear's pane set, the button shows and loads nothing until toggled; the toggle loads the page once). `tests/test_pane_registry_served.py`: the Artifacts button hidden at boot, the gear's generic row (off by default) enabling it, the rail button showing, the pane coming on screen and loading only then; the kit leg unchanged.
- The pins re-aimed to the hand-written five or to the generic form: `test_files_pane`, `test_kernel_pane_rail`, `test_pane_state_broadcast` (its harnesses carry no attribute, so the hand five), `test_pane_gutters`, `test_pane_gutter_drag`, `test_pane_order_parity` (the phone tabs are the rail less the experimental records), `test_kernel_settings_sections` (five Panes rows, no `rs-artctl`), `test_settings_tabs_recut_browser`, `test_chat_split`, `test_kernel_panenav`, `test_kernel`; `ui/webview/artifacts.test.ts` (the record and the generic row's words), `esc-close.test.ts`, `gear-tabs.test.ts`.

Red-first at the base 91b7c0ad per test (the three product files at the base, the tests as written): the first comment lists each. Whole extension suite with the served labs under REQUIRE and the whole Python suite on the head before the push.

## Roads not taken

- No migration of a stored `showArtifactsControl: true` into `panes.artifacts`: the feed owner's call, the pane shipped hours before and defaults off.
- The hand-written five stay hand-written: their lines are pinned by many files; the generic build starts after them. Phase four (bands and the phone) is the plan's next item.

Tier: feature


henrypinkard:
**Measured reds at the base 91b7c0ad** (the three product files, `kernel/kernel.py`, `ui/webview/gear.js` and `ui/webview/settings.ts`, checked out at the base; the tests and the fixture as they are on this branch):

- The eleven Python modules that pin the landing (`test_pane_registry`, `test_artifacts_list`, `test_kernel_settings_sections`, `test_pane_order_parity`, `test_files_pane`, `test_kernel_pane_rail`, `test_pane_state_broadcast`, `test_pane_gutters`, `test_pane_gutter_drag`, `test_chat_split`, `test_kernel_panenav`): 33 failed and 19 errors at the base, 199 green on the head. What the base showed: the record not experimental (`'experimental': False`), six Panes rows where five, the phone tab present, the body tag without the attribute, `gv-d` where `gv-artifacts`, the hand `po` literal with `artifacts:false`, the `grow` store and the `--g-artifacts` variable written by the hand list (every gutter harness), `key()` falling back to `'artifacts'`, `COLS` and the frame lists naming `f-artifacts`, `avail.artifacts` in the broadcast, the reconcile without the on-screen gate; the rendering fixture's body tag, phone tabs, pane row, column rules and gutter calls differing (the rail unchanged, as it must).
- `ui/webview/artifacts.test.ts`: red at the base on the record ("off by default, asked for in the gear": the base record was not experimental).
- The two served labs, `tests/test_pane_registry_served.py` and `tests/test_artifacts_pane_served.py`: red at the base, the drivers dying before a result (the gear's generic `rs-pane-artifacts` row does not exist at the base, and the base landing still carried the bespoke control), so nothing of the on-screen-only gate could be measured there.

**The rendering fixture, re-cut deliberately** (`made_from` names 91b7c0ad plus this fold; the previous fixture was from 4d977bed and is compared here slice by slice):
- `body_tag`: 42 -> 250 chars; came: `data-panes` with the Artifacts record's row (`builtin: true`).
- `column_css`: 3854 -> 3739 chars; gone: `#gv-d`, `.no-artifacts-control`, the control's two hide rules; came: `#gv-artifacts`.
- `gutter_calls`: 536 -> 336 chars; gone: the hand `gv-d` line for `artifacts-pane`.
- `pane_row`: 540 -> 569 chars; gone: `gv-d`; came: `gv-artifacts` and the generic iframe (`data-src="/artifacts" data-protocol=romp`).
- `phone_tabs`: 249 -> 203 chars; gone: the Artifacts tab (an experimental record has none).
- `rail_buttons`: unchanged (294 chars).

On this head: the eleven modules 199 green, the three extension test files green, both served labs green (the Artifacts button hidden at boot, the gear's generic row enabling it, the pane loading only when it comes on screen; the toggle loading the page once).

Whole suites on the head before the push: the extension suite 5170 tests, 5164 passed, 6 skipped, 0 failed (exit 0); the Python suite 14725 passed, 57 skipped, 2341 subtests passed, one failure that was this branch's own stale expectation (`tests/test_settings_tabs_recut_browser.py` counted five Panes rows where the gear now also renders the generic Artifacts row; the count is fixed and the module re-run green alone), and the one pre-existing red on this machine (`tests/test_drop_anywhere_served.py`'s lab kernel never serving its health route, on the shared main checkout too; named in PR 1919).

henrypinkard:
f19df237: a docs-only note at the top of `plans/artifacts-pane.md` section 5 (the feed owner's read of this branch against their list found the fold complete and this one gap): the pane is a code record rendered by the generic build since this PR, the gear's generic Panes row is its control, and the hooks and the bespoke key that section names are the first landing's, kept as history; the on-screen-only load now rides the generic build's gate. No product change.

lczh:
Evidence executed against f19df237.

This fold mostly holds: the record renders through the generic build at the head, the reds the first comment measures at the base are real for the modules it aggregates, the two unit modules pass, and the extension suite is at its usual figure. What is left is one pin that now passes for the wrong reason, the PR's one behavioral change with no executed unit test behind it, and the on-screen constraint not holding on a phone; the rest is stale prose in two tests and two comments, and a count.

## At this head

**The gear-tabs ordering chain still compares against the deleted `rs-artctl` row, so its Files-before-Appearance clause is vacuous while the message still claims it.** ui/webview/gear-tabs.test.ts:69-72 keeps `G.indexOf("id=rs-artctl") < G.indexOf("data-section=appearance>Appearance<")` after the diff dropped the `rs-filesctl < rs-artctl` clause before it. This PR deletes the row from ui/webview/gear.js, so `id=rs-artctl` is nowhere in the General slice (`grep -c` on gear.js is 0), `indexOf` is -1, and `-1 < n` holds for any n. The comment at 66-67 and the message at 72 still promise "Panes (with the Files control), Appearance". Executed on a scratch copy bundled the way the test build bundles: with the Files row block (gear.js:133-136) moved below the Appearance header (151), gear-tabs.test.ts stays 6 of 6 under the head's chain; with the clause's `id=rs-artctl` replaced by `id=rs-filesctl`, the same move fails 5 of 6 on the chain's own message, and the fixed chain on the unmoved gear.js is 6 of 6. Only this test reads its chain; the Python twins (tests/test_files_pane.py:283-284, tests/test_kernel_settings_sections.py:87) pin the order through `str.index`, which raises on absence, so the invariant survives there alone. Put `G.indexOf("id=rs-filesctl")` in the clause and, since the PR's point is that the row is gone, add `assert.equal(G.indexOf("id=rs-artctl"), -1)` so the chain cannot go vacuous again.

**The on-screen-only load gate has no executed unit test; the one harness that runs the collapse script cannot tell the gate from its absence.** The gate is kernel/kernel.py:64060, where `reconcile` copies `data-src` to `src` only for a hand-written optional pane (`!(k in DPX)`), and apply's DP loop at 64083-64084, which copies for a generic pane when `po[k]` is on. tests/test_artifacts_list.py:329-330 pins both lines as source strings. The harness that executes the script, tests/test_pane_registry.py TheInlineScripts (492-511) over `_PANES_STUB` (395) and `_PANES_DRIVER` (426), carries `notes` (on, so loaded at boot), `docs` (a URL pane, on), `lab` (experimental, off in the gear) and the Artifacts row (experimental, off); no row is enabled in the gear while off screen, the only state in which the gate acts. Executed: with `!(k in DPX)&&` removed from 64060 on a scratch copy, the stub and driver produce byte-identical output, and the unittest case itself, run in-process under the mutant, passes. The other harness that executes this script, tests/test_pane_state_broadcast.py, has a body stub that answers only `data-tab` (87), so `DP` is empty and the gate is unreachable there (reading the harness). The behavioral check is the Chromium lab, tests/test_artifacts_pane_served.py:221-226 (`ctlOn.iframeSrc` None, then `/artifacts`), which skips without the playwright package (126-127, 204), as it did here; CI's served job forces it, so the product is covered there, and source pins plus a lab is a defensible split. But what is pinned is the text of two lines and one browser's outcome in one job; nothing pins the gate's behavior where a browser is absent. The discriminating case is three lines and I ran it: seed `STORE['romp:settings'] = JSON.stringify({ panes: { artifacts: true } })` in the stub; at boot the head gives `f-artifacts` no `src`, its button not hidden and `po-artifacts` off, where the mutant gives `src` `/artifacts`; after `__rompPaneToggle('artifacts', true)` both give `/artifacts` and the class on. A non-experimental `on: false` row splits the same way (head null, mutant `/offpane`). Add those assertions to the harness test or a sibling case.

**On a phone-layout load with the stored rail flag on, the generic Artifacts iframe still takes `src` and its bundle walks the remembered session's transcript, though no tab can show it.** At boot `po.artifacts` comes from `romp-panes` (64027); `reconcile()` (64101, before the first apply) keeps it when `optOn()` says the gear enabled the pane; apply's DP loop (64083-64084) copies `data-src` to `src` when `po[k]` is true, with no layout check. On the phone the tab bar decides what shows: `_mtab_buttons_html` (64710-64715) emits no tab for an experimental record, the mobile CSS shows only `#f-chat.m-on,#f-fleet.m-on,#f-feed.m-on,#f-files.m-on` (65492) and its `display:contents!important` override names five panes (65487), so `show()` (63461-63463) has nowhere to put it. ui/webview/artifacts.ts:161 runs `if (selected) requestListing();` at module end from the remembered sid (31), so the walk does not wait for the pane-set broadcast, whose `on.artifacts` is false on a phone (64032). Executed in node under the module's stub with `__rompMobileOn` true: `romp:settings.panes.artifacts` true and `romp-panes.artifacts` true give `f-artifacts` `src` `/artifacts` and `po-artifacts` on while the broadcast says `on.artifacts` false and the tab is chat; the gear flag alone gives no `src`. The phone reaches it by itself too: from nothing stored, the gear's row turning on (the storage event at 64105, `reconcile(true)` then `apply()`) sets `po.artifacts`, persists it and copies `src`. The same stub against the base's script (its `actl&&po.artifacts` copy at the base's kernel/kernel.py:64067) loads the frame the same way; the base had an Artifacts tab, but its mobile CSS was the same five-pane list, so that tab showed a blank frame. So this is the spec's on-screen constraint not holding on the phone rather than a regression, and plans/panes-as-data.md section 7 item 4 defers the phone to phase four. Either say so in the body or close it here: in apply's generic copy, skip when `window.__rompMobileOn&&window.__rompMobileOn()` and the pane's tab is not current (an experimental record has no tab, so never), with a `MOBILE = true` case in the same harness and a served leg at a phone viewport asserting no `src` with both flags stored. Reading the code, `show()` does not refuse `artifacts` either: `F` holds its frame from the attribute (63453) and the bar has no button for the hidden check (63462) to find.

## Smaller points

- Two tests the PR rewrote still carry the deleted control. tests/test_pane_registry_served.py:185, the kit leg, writes `s.showArtifactsControl = true`; nothing reads the key at the head (`git grep` finds plans, negative assertions and this one write), and the pane is enabled by step 4's click on `rs-pane-artifacts` (160), so the later `__rompPaneToggle("artifacts", true)` (224) passes through `k in po` (64094). tests/test_pane_registry.py:398 hand-lists `'artifacts'` in the stub's `KEYS` while `_rows()` (72-76) now prepends `ARTIFACTS_ROW` (56) into `__DATA_IDS__`, so the harness builds the frame and the button twice and `querySelectorAll('.rail-btn[data-pane]')` (419) returns the Artifacts button twice; harmless since the maps overwrite, misleading to the next reader. Drop the write and the hand-listed id.
- Two comments describe the code this PR replaced. kernel/kernel.py:60846 says the `seq` prefix is "the shipped columns after the chat, from _CODE_PANES", but the literal is `_COLUMN_IDS`, now from `_HAND_PANES` (56114), rendering `["fleet-pane", "feed-pane", "files-pane"]`, and the Artifacts record, a `_CODE_PANES` entry, reaches `seq` only through `DPANES`; tests/test_pane_registry.py:571 pins the hand-written form. kernel/kernel.py:64023 says "An empty registry emits no attribute and this block is a no-op", but `_panes_attr()` (56242) returns the Artifacts row on a kernel with no data pane (executed) and test_00 (153-165) pins exactly that. The plan is covered: f19df237's note at plans/artifacts-pane.md:111-117 marks section 5's hooks and the `showArtifactsControl` bullet (127) as history and names the generic gate, and plans/panes-as-data.md:277 states the mapping.
- The body says red-first at the base per test; two re-aimed modules are green at the base as well. tests/test_pane_order_parity.py:43-44 and 57 derive the experimental set from `km._CODE_PANES`, and at the base no record is experimental: on a hermetic base kernel the set is empty, the rail and the tabs both list six panes, and every head assertion holds. tests/test_settings_tabs_recut_browser.py:34-35 count `PANES_ROWS + GENERIC_ROWS + 1`, which is 7; reading the code, the base gear has six hand `rs-panes-row` labels where the head has five, the base `_panes_attr` is the empty string with no data pane (executed), so `renderRegistryRows` adds nothing, and the five ids in `rows` (194) all exist at the base: 7 on both sides. Not defects of the re-aims; say so in the body, or make the lab discriminating (assert `rs-pane-artifacts` among `els` and no `rs-artctl`).
- The commit message of eb7fb83e says "fourteen Python pin files and three extension tests" after naming the Artifacts list test, the registry module and both served labs separately. The diff touches 15 files under tests/test_*.py; less those four, eleven, exactly the list the body gives. Three extension tests is right.

lczh:
The test runs behind the comment above, on f19df237 (merge base with main 91b7c0ad).

## Tests

- The extension suite at the head, `npm test` (the test bundle, then `node --test` over it): 5170 tests, 5163 passed, 1 failed, 6 skipped. The one failure is the KaTeX layout probe (test 2790), which fails on main in this environment and touches nothing the PR changes. ui/webview/gear-tabs.test.ts alone from the same build: 6 of 6.
- The two Python modules, each alone: tests/test_pane_registry.py 13 passed, 1 skipped; the skip is the test at 526, which reads the built shell bundles and skips as "extension deps absent" without the extension's node_modules. tests/test_artifacts_list.py 12 passed. The served labs were not run here; the playwright package is absent, so both would skip.

- ui/webview/gear.js with the Files row block (lines 133-136) moved below the Appearance header (line 151), the tree bundled the way the test build bundles (cjs, node, node18, typescript external): gear-tabs.test.ts 6 of 6 under the head's chain. The same move with the chain's `id=rs-artctl` replaced by `id=rs-filesctl` in gear-tabs.test.ts:69, rebuilt: 5 of 6, the failure being "every existing control keeps its id and sits in exactly one pane" with the chain's own message (General: Account, Panes with the Files control, Appearance, and the rest). The fixed chain on the unmoved gear.js: 6 of 6. An unmutated scratch copy as the control: 6 of 6, the checkout's figure.
- kernel/kernel.py `_LANDING_COLLAPSE_JS` with `!(k in DPX)&&` removed from the reconcile at 64060, the copy that makes a generic pane's iframe take `src` when merely enabled: tests/test_pane_registry.py TheInlineScripts's own stub and driver produce byte-identical output against the head script and the mutant; the unittest case itself, run in-process with the module's constant replaced by the mutant, 1 run, 0 failures, 0 errors.

## Probes

- The discriminating case, in node over the module's stub with its rows (`notes`, `docs`, `lab`, the Artifacts row) and `romp:settings` seeded `{ panes: { artifacts: true } }`: at boot the head gives `f-artifacts` `src` null, the button not hidden, `po-artifacts` false; the mutant gives `src` `/artifacts` at boot. After `__rompPaneToggle('artifacts', true)`: `src` `/artifacts` and the class on under both. A fourth row `{ id: "offpane", on: false }`, not experimental: `f-offpane` `src` null on the head, `/offpane` under the mutant.
- The phone layout, the same stub with `MOBILE = true` and only the Artifacts row: with `romp:settings.panes.artifacts` true and `romp-panes.artifacts` true, `f-artifacts` gets `src` `/artifacts`, `po-artifacts` is on, the broadcast's `on.artifacts` is false and the tab is chat. With the gear flag alone: `src` null. From nothing stored, a storage event with `panes.artifacts` true: `src` `/artifacts`, `po-artifacts` on, `romp-panes` persisted with `artifacts: true`, no tab switch. `_mtab_buttons_html()` at the head renders five tabs, none for `artifacts`.
- The base (91b7c0ad), a hermetic kernel loaded from the base's kernel/kernel.py under a scratch state root: its `_LANDING_COLLAPSE_JS` under the same phone stub with `romp:settings.showArtifactsControl` true and `romp-panes.artifacts` true gives `f-artifacts` `src` `/artifacts` and `po-artifacts` on; its `_mtab_buttons_html()` carries `<button data-pane=artifacts>`; its landing's mobile CSS shows `#f-chat.m-on,#f-fleet.m-on,#f-feed.m-on,#f-files.m-on{display:block}` and names `#chat-pane,#fleet-pane,#feed-pane,#files-pane,#tl-pane` in the `display:contents!important` rule, neither naming the Artifacts frame or pane. The same base kernel under the head's parity assertions: the experimental set from `_CODE_PANES` is empty, the rail and the tabs both list `chat, timeline, fleet, feed, files, artifacts`, and every assertion of tests/test_pane_order_parity.py's two re-aimed tests holds. Its `_panes_attr()` is the empty string and its landing's body tag carries no `data-panes`.
- The head's `_panes_attr()` with no data pane: one row, `{"id":"artifacts","title":"Artifacts","protocol":"romp","experimental":true,"on":false,"builtin":true}`.
- `git grep rs-artctl` at the head: ui/webview/gear-tabs.test.ts:69 (the clause), ui/webview/artifacts.test.ts:48 and tests/test_artifacts_list.py:333 (asserting absence); ui/webview/gear.js none. `git grep showArtifactsControl`: plans/artifacts-pane.md:113 and 127, plans/panes-as-data.md:277, tests/test_artifacts_list.py:297, 317 and 332 and ui/webview/artifacts.test.ts:45, 47 and 48 (all asserting or describing absence), and tests/test_pane_registry_served.py:185 (the write); no product file.
- `<label class="rs-row rs-panes-row">` in ui/webview/gear.js: 6 at the base, 5 at the head; `renderRegistryRows` (gear.js:472-477) creates one `rs-row rs-panes-row` label per row of `body[data-panes]`.
- `git diff --stat 91b7c0ad f19df237`: 24 files. Fifteen under tests/test_*.py: test_artifacts_list, test_artifacts_pane_served, test_chat_split, test_files_pane, test_kernel, test_kernel_pane_rail, test_kernel_panenav, test_kernel_settings_sections, test_pane_gutter_drag, test_pane_gutters, test_pane_order_parity, test_pane_registry, test_pane_registry_served, test_pane_state_broadcast, test_settings_tabs_recut_browser. Three under ui/webview/*.test.ts: artifacts, esc-close, gear-tabs. eb7fb83e to f19df237: plans/artifacts-pane.md only, 8 lines added.

henrypinkard:
**Round two** (the verifier's HIGH, the medium and the four lows), with the reds measured at f19df237 (round one's head: `kernel/kernel.py`, `ui/webview/gear.js` and `ui/webview/settings.ts` checked out there; the round-two tests as they are):

- **The HIGH, the phone's tab**: `tests/test_pane_registry_served.py`'s seventh leg (a 420 by 860 context, the notes pane's desktop flag off in the store): at f19df237 the tap left `notesSrc: None, visible: False`, the blank pane the read described; on this head the tap loads the page (`/pane/notes/`), shows it full screen with its shim up, and tapping away and back requests it once. Two changes were needed, not one: the phone's `show(p)` copies `data-src` to `src` for the shown pane, once, generically; and the generic panes' phone display rules ride the mobile media block beside the hand five's (`#<id>-pane{display:contents!important}`, `#f-<id>.m-on{display:block}`), without which the loaded page sat in a wrapper the desktop hide rule kept at `display:none`. Both pinned in `tests/test_pane_registry.py` (red at f19df237 as text: the copy line and the rules absent).
- **M, `gear-tabs.test.ts`**: the ordering clause reads `rs-filesctl` before Appearance, as the Python twin does; the vacuous `rs-artctl` clause is gone.
- **Lows**: (a) the experimental parenthetical is inside the experimental branch; (b) `BUILTIN_HINTS` in `gear.js` gives the Artifacts row its first landing's words (a session's written, shown and dropped files as a list and a grid of large thumbnails), the generic row rendering it for a shipped record; (c) `settings.ts` drops a stored `showArtifactsControl` at load like `fileLinkPane` and `filesControl`, pinned in `settings.test.ts` (red at f19df237: the key survived the load) and in `artifacts.test.ts` and `test_artifacts_list.py` (red at f19df237: the drop line absent, the hint map absent); (d) the lab's docking leg no longer writes the dead key.
- **The fixture** is re-cut once more for the phone rules (the `column_css` slice gains the generic panes' mobile rules; `made_from` says so); at f19df237 the slice differed by exactly those rules.

On this head: the registry module 14 green, the Artifacts list test 12 green, the three extension test files green, both served labs green (the registry lab with its seven legs).

Whole suites on this head before the push: the extension suite 5171 tests, 5165 passed, 6 skipped, 0 failed (exit 0); the Python suite 14726 passed, 57 skipped, 2341 subtests passed, with the one pre-existing red on this machine (`tests/test_drop_anywhere_served.py`'s lab kernel never serving its health route, on the shared main checkout too; named in PR 1919).

lczh:
Round two, executed against 2cdcb1cc.

The ordering chain, the dead write and the wording points landed, and the phone got a real fix for a defect round one did not name: a data pane's tab showed a blank pane and now loads and shows its page, with a served leg CI runs. Two items stand: the on-screen gate still has no executed unit test, and the tabless Artifacts record still loads on a phone with both flags stored, with the body silent on it. Two small new things in code this commit touched.

## Landed

**The gear-tabs chain (A).** ui/webview/gear-tabs.test.ts:69 orders `id=rs-filesctl` before Appearance and no `rs-artctl` clause remains (the id survives only in absence pins). Re-executed on a scratch tree bundled as the test build bundles: unmoved, 6 of 6; with the Files row block (ui/webview/gear.js:133-136) moved below the Appearance header (151), 5 of 6, failing on the chain's own message.

**The phone's tab for a data pane (the author's item).** kernel/kernel.py:63478 copies `data-src` to `src` for the shown pane, guarded by `!src`; `_data_pane_mobile_css` (56265-56273) emits `#<id>-pane{display:contents!important}` and `#f-<id>.m-on{display:block}` into the media block at 65503, beside the hand five's. Executed over a DOM stub with the built script: the copy fires once, fires for a URL pane (protocol none), leaves a frame with a `src` alone; a hidden button still refuses. The seventh leg (tests/test_pane_registry_served.py:227-250, assertions 424-432) runs in CI's browser job, .github/workflows/ci.yml:330, under `ROMP_SERVED_TESTS_REQUIRE=1` (326).

**The write, the words, the drop.** tests/test_pane_registry_served.py:185 sets `paneDocking` alone; no write of the key remains. ui/webview/gear.js:471 `BUILTIN_HINTS` is read only when `p.builtin` (487-488), and `_pane_order` (kernel.py:56189) drops a data file with a shipped id while `_PANE_RESERVED` (56045) refuses the define, so no data pane reaches the hint. ui/webview/settings.ts:96 sits with the other retired keys (90, 95); ui/webview/settings.test.ts:202-210 executes the load and the save.

**The fixture.** `made_from` names 91b7c0ad plus the fold, re-cut. Executed: the slicer over the head's `_landing()` with no data pane equals all six slices; over a `git archive` of 91b7c0ad's kernel, the rail slice is equal and the other five differ by exactly the fold: the attribute on the body tag, the Artifacts phone tab gone, `gv-d` to `gv-artifacts` and `data-protocol=romp` in the pane row, the `gv-d` gutter call gone, and in the column rules (35 on both sides) `no-artifacts-control` and `#gv-d` out, `#gv-artifacts` and `#artifacts-pane{display:contents!important}` in. (`#f-artifacts.m-on` is outside the slicer's pattern; tests/test_pane_registry.py:379-381 pins it.)

## Still open

**(B) The gate has no executed unit test.** This commit adds a CSS pin (tests/test_pane_registry.py:379-381) and a source pin (571-572); the harness at 495-518 and its stub (398-427) are unchanged: no row is enabled in the gear while off screen. Re-executed with `!(k in DPX)&&` removed from the reconcile at kernel/kernel.py:64075 on a scratch copy: the stub and driver give byte-identical output under head and mutant, and the unittest case run in-process under the mutant passes. The three-line case still splits: `romp:settings` seeded `{panes:{artifacts:true}}`, at boot the head gives `f-artifacts` no `src`, the mutant `/artifacts`.

**(C) The tabless Artifacts record on a phone.** Apply's copy at 64099 is unchanged and has no layout check; `_mtab_buttons_html` (64725-64730) emits no tab for an experimental record. Re-executed under the module's stub with `MOBILE = true`: both flags stored give `f-artifacts` `src` `/artifacts` at boot while the broadcast's `on.artifacts` is false and the tab is chat; the gear flag alone, no `src`; from nothing stored, the gear's storage event copies `src` and persists the flag. The new rules change nothing here: without `m-on` the frame sits under `iframe{display:none}` (65507). The seventh leg sees nothing loaded before the tap only because it stores `romp-panes {notes:false}` (231); with the default flag, `notes` takes `src` at boot too. Deferring the phone (plans/panes-as-data.md section 7 item 4) is fine; say so in the body, or check the layout in apply's copy.

- tests/test_pane_registry.py:401 still hand-lists `'artifacts'` in `KEYS` while `_rows()` (72-76) prepends the row, so the harness builds the Artifacts frame and button twice.
- kernel/kernel.py:60857 still says the `seq` prefix is "from _CODE_PANES" (it is `_COLUMN_IDS` over `_HAND_PANES`, 56114); 64038 still says an empty registry emits no attribute (test_00 at 153 pins the Artifacts row with none).
- The body (line 21) still says red-first per test. tests/test_pane_order_parity.py and tests/test_settings_tabs_recut_browser.py are untouched here and green at the base; re-executed on the base kernel: no experimental record, six tabs and six rail buttons, `_panes_attr()` empty, six `rs-panes-row` labels to the head's five, so 7 rows on both sides.
- eb7fb83e's message still says "fourteen Python pin files"; this commit's message and the body carry no count.

## New at this head

- kernel/kernel.py:63464 says an experimental pane has no tab button "so show() refuses it". It does not: show() (63472-63483) returns for a HIDDEN button with that id, not a missing one. Executed over the DOM stub: `window.__rompMobileTab('artifacts')` on a phone sets `data-tab` to `artifacts`, puts `m-on` on `f-artifacts`, copies `src`, lights no tab, and persists `romp-mobile-tab` as `artifacts`, so the next boot on that phone restores it (63547). Round one's tree blanked every pane there; this commit's rules show the frame. No shipped caller passes that id (the relays pass `chat`, `files`, `feed` or a remembered current tab), so the road is latent. Refuse a pane with no button, which makes the comment true, or reword it.
- ui/webview/gear.js:431 `load()` drops `filesControl` and `fileLinkPane` but not `showArtifactsControl`. Executed: a store with all three loads with only `showArtifactsControl` kept, and `save()` (443-452) writes the whole object, so any gear save re-persists it. Harmless (never read), but the two precedents the drop cites are dropped in both loaders; add the delete at 431 and extend ui/webview/gear-tabs.test.ts:75's pin.

## Runs

- tests/test_pane_registry.py alone, behind the test lock: 13 passed, 1 skipped in 2.21 s (the skip at 529: no extension deps). tests/test_artifacts_list.py: 12 passed in 0.72 s.
- The extension suite, `npm test` at the head: 5171 tests, 5164 passed, 1 failed, 6 skipped. The failure is the KaTeX layout probe (test 2790), which fails on main in this environment.
- Neither served lab ran here (no playwright package). The whole-suite figures in the round-two comment are the author's, not repeated here.

ksarma:
The evidence here was executed against 2cdcb1ccd, the PR's head; every file and line reference is on that commit, and each reads the same at the merge commit 10f33abb5, whose only other changes are in four files the PR never touched. Where one of the six comments above carries an item, only what this evidence adds is written. Of the earlier Artifacts pane items in the note on https://github.com/romp-on/romp/pull/1911, the phone tab's removal closes the blank tab and its 58 px of overflow; the boot gate on the rail flag and the tab switch's missing layout guard survive in the generic lines that replaced them, below.

**A flip of the Sessions, Outline or Feed row drops every registry key from the store, the Artifacts pane's control among them.** The hand rows' handler at ui/webview/gear.js:464 rewrites `romp:settings.panes` from panesOf(s), three keys, so the live reconcile runs with the Artifacts key absent: the rail button hides and the column closes, while the row in the open gear still reads checked. Executed with real clicks; nothing on the PR mentions it. The dropper predates this PR for every data pane's key, turning a stored false back into on; this PR routes the Artifacts control onto it. One line fixes the store: `var p = Object.assign({}, (s.panes && typeof s.panes === 'object') ? s.panes : {}, panesOf(s)); p[k] = pn[k].checked; s.panes = p; save(s);`, and the comment at 461-462 becomes false with it; a boolean filter in the shape of settings.ts's paneSet loop is the safer form, since a bare merge carries any stored value. Pin it with a served leg (enable the Artifacts row, flip Feed, assert the key true and a data pane's stored false surviving) and a node case over the handler.

**On a phone, apply loads a generic pane by the desktop flag into a frame the phone never shows.** As the 11:16Z comment's (C) says, apply's copy at kernel/kernel.py:64099 has no layout check, and the tab bar builder at 64725-64730 emits no tab for an experimental record. Executed in Chromium at 390x844, nothing stored: a tap on the Artifacts row loads /artifacts into a hidden 0x0 frame, and a reload loads it again from the persisted rail flag. The same copy loads every default-on tabbed data pane at boot behind the chat tab, as (C) notes. Fix, in apply's DP loop: skip the copy while `window.__rompMobileOn()` is true for a key with no tab button, or for every generic pane now that the tab switch copies on the tap; test the tab's presence rather than a rendered rect, since under the docking kit a freshly toggled leaf can measure 0x0 at apply time, and leave the tab switch's copy alone, or a tabbed pane's tap goes blank again. The media-query re-apply is required: the skip alone left a rotation to desktop blank, src null; with the `matchMedia` change event wired to apply (the mobile script's retell at 63487-63488) the flip copies src. The row's rail-toggle promise is false on a phone; key phone wording on the tab's absence rather than disabling a checked row, which would leave a desktop-enabled pane stuck on. Pin it with a 390x844 served leg (the row tap, then src null and zero requests; both flags stored, a boot with no request) and a MOBILE-true TheInlineScripts case on that seed.

**The last commit's copy in the tab switch runs at boot, before the pane controller has ruled and in every layout.** show() at kernel/kernel.py:63478 copies data-src to src for the shown pane, and the boot restore at 63547 calls it with the remembered `romp-mobile-tab` before the collapse script parses, with no layout gate and no gear read. On a phone, a remembered tab of a pane the gear has disabled loads its document and socket at boot before reconcile reroutes to chat; on a desktop, a key remembered from a narrow-layout visit loads that generic pane off screen on every boot, and nothing repairs the key. Executed at this head, at f19df2376, which loaded nothing here, and at the base; the hand five are unaffected. Fix: copy only on a tap or a reveal, or gate show()'s copy on `mobileOn()` and on the collapse script having parsed (window.__rompPaneToggle, defined at 64113 before the boot reconcile at 64116, is absent exactly at the boot restore); then let the controller's boot apply copy for the current tab's frame when mobile is on and the key is enabled after reconcile. Three bounds: `window.__rompPaneEnabled` (64788) cannot be the gate, since it answers on for the Artifacts record with nothing stored; a boot copy keyed on the key being in po alone still admits the gear-on tabless record full screen with no tab lit, so it also needs a visible tab button; and the media-query change handler needs the same current-tab copy, since with the layout gate alone a desktop boot on a remembered tab, then narrowed past the breakpoint, showed a blank notes pane. No composed two-script boot harness exists; tests/test_kernel_mobile.py is the browser-free home for remembered-tab seeds (a gear-disabled pane; artifacts under both gear states) at both viewports, asserting no src and data-tab chat.

## Smaller points

- show() still refuses no pane without a tab button (the 11:16Z comment's first new item), and a stale key of artifacts with the gear on now boots the phone into the Artifacts page full screen, no tab lit; fall to `p='chat'` when no bar button carries the id, the Files line's shape at 63472, never a bare return, which leaves data-tab unset at boot.
- load() at ui/webview/gear.js:431 still keeps showArtifactsControl (the 11:16Z comment's second new item), and since saveSettings in settings.ts has no product caller the drop at ui/webview/settings.ts:96 is in-memory only, so the gear.js line is the whole remedy: `delete o.showArtifactsControl;` before the filesControl delete, or tests/test_files_pane.py:267 reds; ui/webview/artifacts.test.ts:48 then re-aims to rs-artctl alone.
- Five comments and a test title still describe the code before this PR: beside the 11:16Z comment's 60857 and 64038, kernel/kernel.py:64035-64036 (the attribute carries the shipped record too), the module header at 56039, ui/webview/gear.js:465-466 and ui/webview/palette-main.ts:170-171 (the attribute is present with an empty registry), and the title of ui/webview/artifacts.test.ts:39 with its header at line 2 (a setting the body asserts is gone).
- The registry rows' div at ui/webview/gear.js:148 renders after the Pane docking switch, so the Artifacts row, above the switch at the base, now sits below it; the four-line move to just after the Files row (136) restores the grouping.
- Pre-existing, outside this diff: with the Files control on the phone's tab bar is 403 px wide at 390, clipping the settings action, unreachable at 375 and under or with a data pane on; shrink-with-ellipsis mangles labels, so the options are a scrolling tab strip beside a fixed action cluster, or recording the content budget under phase four.

## Tests

The on-screen gate at kernel/kernel.py:64075 and the tab switch's copy at 63478 are pinned as text alone in the Python matrix (the 11:16Z comment's (B); the served labs run them in CI's browser job); the copy's pin at tests/test_pane_registry.py:571-572, which the 10:50Z comment cites as red-first, sits after the dist gate at 559-564 and runs in no Python CI cell, so lift the baked-maps block (567-576) above the gate, or better, two executed legs: a `romp:settings` seed in TheInlineScripts with the Artifacts pane on plus a non-experimental on:false row, asserting src null at boot and set after the toggle, and frame stubs with getAttribute, setAttribute and a data-src in test_pane_state_broadcast.py's mobile harness, asserting one src copy on the tap and none on a second. tests/test_settings_tabs_recut_browser.py counts the generic Artifacts row but its label, class and hide lists (136-137, 195) read the hand ids (add rs-pane-artifacts to them; the shown check at 459 then needs `PANES_ROWS+GENERIC_ROWS`); the drag and gutters harnesses (tests/test_pane_gutter_drag.py:77) stub the body with classList alone, so the Artifacts default grow the base pinned is asserted nowhere executed (a body stub answering data-panes); and the slicer at tests/landing_slices.py:22-23 drops `#f-artifacts.m-on{display:block}` from the re-cut fixture, so that rule is pinned only at tests/test_pane_registry.py:379-381 (widen the alternation with .m-on). In tests/test_pane_registry_served.py the kit leg (183-186, 223) still rides on step 4's row click at 160 persisting the gear flag across the reload (the 07:30Z comment's road), with no line saying so; seed panes.artifacts true at 185.

PR #1925, opened 2026-09-20
The Artifacts pane, second pass: remote sessions, the open-tabs picker in the strip's dress, growth-driven refresh, the lock

henrypinkard:
The user tried the Artifacts pane and asked for another full pass (2026-09-20, five asks, the manager's parcel). This PR is design first: its first commit amends plans/artifacts-pane.md (a new section 9 with the decisions and the code facts each rests on; sections 3, 4, 6 and 7 amended where the first landing's words no longer hold), and the code follows as further commits on this branch, unarmed.

The five decisions, in short:

1. Remote sessions: the transport already routes the listing op by sid to the owning kernel and host-routes the thumbnails; only the selector asked the local kernel alone. The open-tabs scope below lists remote tabs with their host ids for free, so a remote session is listed, routed and served with no per-host request. A two-kernel lab proves it.
2. Scope: the picker lists exactly the sessions open in the chat panes. The chat column posts its tab set to the shell ({romp:'chatTabs', tabs}), the shell unions the columns and broadcasts to every romp-protocol pane, like the pane-set broadcast.
3. Dress: the picker button and its rows wear the strip's session label (the quiet host prefix, the name bold in its identity colour) through a shared helper beside hostNameNodes, in a ctx-menu card with the menu tokens; no native select, no identity dot.
4. Growth: no Refresh button. The pane watches its shown session (watchArtifacts); the owning kernel sends artifactsChanged from the pusher cycle when the transcript's version (mtime and size) moves, one stat per cycle per watched session (this line amended 2026-09-21 with the reviews' fix PR: the compare rides the pusher's event-woken cycle, the versions read once per watched session, never per watching client); the pane re-asks. The walk becomes incremental per session through an in-memory memo keyed by the last walked turn's fork-stable id. The request-and-response listing stands; the signal is the event that says when.
5. Lock: a lock beside the picker with the Sessions pane's padlock glyph. Unlocked, the pane follows the most recently selected chat tab through the shell's activeChat relay (widened from the feed to every romp-protocol pane) and the kernel's frame on ready; locked, it stays on the pick; a pick never changes the lock (the manager's correction of 2026-09-20, the design's 9.5 and the machine; this line amended 2026-09-21 with the reviews' fix PR); both persist per browser.

Coordination: the chat owner for the chatTabs post and the widened relay; the cards owner for the shared session label helper; both on this branch alone.

## The code (the second pass, on this branch)

- **Scope and remote (9.1, 9.2):** `renderTabs` posts the column's own tabs (`postChatTabs`, the strip's membership narrowed by the column partition, a subagent viewer and a provisional tab left out, deduped by signature); the collapse script keeps one set per column frame, unions them in column order (`chatTabsUnion`) and tells every protocol pane (`tellAll`) on each change and on a pane's load. The pane lists that union; a remote tab under its host id routes the listing to its kernel and the thumbnails through the host relay. The two-kernel lab `tests/test_artifacts_remote_served.py` proves it end to end.
- **Dress (9.3):** `sessionLabelNodes(name, sid, color)` beside `hostNameNodes` in `host-prefix.ts`, inked by `.session-name.colored` in `styles.css` through the strip's `--chip-bg` and the T390 lightness floor (`--peer-ink-l`); the picker is a `ctx-menu.ts` card (`menuCard`, `showMenuCard`) with the current row marked. The cards owner confirmed the helper's home and the ink rule.
- **Growth (9.4):** `watchArtifacts {sid, unwatch?}` records the one watched session per socket, stamped with the transcript's version; `_artifacts_signal(now)` runs in the pusher cycle after the push, one stat per watching client, one `artifactsChanged {sid, version}` per version move; `_artifacts_mentions` keeps the per-session memo (position and fork-stable id of the last walked turn, the mentions map) and walks only the turns after it when the prefix stands. The pane re-asks on screen, marks itself stale off screen; no Refresh control remains.
- **Lock and follow (9.5):** the shell's active-tab relay hands the frame to `__rompTellPanes` (every protocol pane, the feed among them with its nonce intact) and the kernel's `_relay_active_chat` and ready arm reach `artifacts` clients too; the pane posts `ready` like the Files pane. `nextSelection` in `artifacts-model.ts` is the machine: a pick never changes the lock (the manager's correction); the lock and the shown session persist per browser.

**Red first**, at the round's base 683c997b (main plus the design): the kernel tests red on behaviour (the watch unanswered, no signal frame after growth, a whole walk instead of the new turn, no relay frame, the shell pins); the pane's unit tests red at build (new exports); the three labs red per test (no picker, no lock, no growth) with their drivers hardened so a missing element records a null fact instead of a crash.

**Verification on the head:** tsc; the kernel tests and the active-chat relay tests; the pane's tests; the three labs (13 green); the whole extension suite (5174 tests: 5168 pass, 6 skipped, 0 fail); the whole Python suite in three shards (the unit shards 7129 and 7025 passed, 14 and 41 skipped; one red in the second, the remote-identity alias test, which fails in that shard's ordering at the base too and passes alone on both: main's order-dependent red, not this branch's; the served labs shard 593 passed, 2 skipped, one red: the held-mail chat lab's at-bottom read, whose payload now names the cause, a fix of its own). The strip's render harness gained stubs for the new hook, the pane-set broadcast harness counts the pane-set messages alone (a load now re-tells the open tabs too), and the relay pins were re-aimed.

## Round two (the verifier's high, medium and lows, 2026-09-20)

- **High, the resume point.** A turn keeps its fork-stable id while it gains atoms, so the memo's resume at `idx + 1` never saw a file written later in the same turn; the verifier's probe (one turn with `Write a.md` then `Write b.md`) listed `a.md` alone after the signal. `_artifacts_mentions` now resumes AT the last walked turn (inclusive) and merges its mentions afresh (a re-walked mention is the same mention, the merge keeps the latest by time). Pinned with the probe's shape in `tests/test_artifacts_list.py`, red at b1ae133d.
- **Medium, the watch after a reconnect.** A socket (re)open mints a fresh client (`_register_ws_client`), and its watch with it; the pane's `watch()` returned early on an unchanged sid, so growth stopped silently after any drop, with no Refresh control to recover. The pane re-arms on the shim's `romp:wsup` (the local kernel) and federation's `romp:hostRelayUp` (a host's relay, matched to the shown session's host), never on the sid changing, and re-asks the listing (a growth during the outage was never signalled). The follow lab gains a leg that closes the pane's socket, waits for the shim's redial and the reopen event, then appends a Write: the file lists with no click (red at b1ae133d).
- **Lows folded:** (a) the shell's `chatTabs` listener keys a set to the posting CHAT frame alone (`f-chat`, `f-chat-N`); (b) the growth signal runs in its own `try` after the push's `finally`, so the push stage times the push alone and its `finally` is the push's; (c) a repaint returns the focus to the bar's button it was on, and a row click focuses the button before the repaint; (d) a selection no tab shows keeps its last known name and colour from the union; (e) `plans/panes-as-data.md` section 5's inbound contract names `chatTabs` and `activeChat`; (f) the three tests green at the base say in their docstrings that they pin the fallback and the cost rule; (g) the dead `.art-btn` rules are gone. **Kept as is:** the mentions merge runs in `_artifacts_mentions` (into the memo's map) and again in `_artifacts_items` (its own contract for every caller, a few hundred entries at most): one builder for direct callers and the tests, one memo for the walk.
- **Main carried** as a pure merge (`1bcf0f89`, the artifacts fold included) before the round's commits; the whole suites ran on the merged head: the whole extension suite (5175 tests: 5169 pass, 6 skipped, 0 fail); the whole Python suite in three shards (the unit shards 7084 and 7036 passed, 14 and 41 skipped; one red in the second, the remote-identity alias test, main's order-dependent red seen in the first pass, which fails in that ordering at the base too and passes alone; the served labs shard 595 passed, 2 skipped, 0 failed, the held-mail chat lab among them with PR 1926's fix carried from main).


lczh:
Read against 683c997b, with the code as it stands at 10f33abb.

The design is right about what it checked: the listing op already routes by sid, the open-tabs scope makes remote sessions free, and a signal beats a pushed listing. What stands between it and a build that works is what it did not check: the incremental walk's memo misses the case the pane exists for, the watch has no lifetime past its first socket, and the lock text disagrees with its own test list.

## At this head

**The lock semantics contradict themselves.** Section 3 says `only the lock button changes the lock` (plans/artifacts-pane.md:77-78); 9.5 says `a pick never flips the lock` (307) and `unlocking keeps the shown session until the next tab switch` (311-312). The unit-test bullet still reads `(follow, pick locks, toggle, unlock jumps, a closed tab)` (324), and the PR body's decision 5 still says `a pick locks`. The correction commit rewrote sections 3 and 9.5 and the lock lab (331-333) but not 324, so a test written from 9.6 pins the machine 9.5 retracted. Rewrite 324 to the corrected machine: follow; a pick under either lock state leaves the lock as it was; toggle; unlocking keeps the shown session until the next switch; a closed tab. Fix decision 5, and add to the lock lab that unlocking moves nothing.

**Lines 305-307 quote the user.** The correction quoted in 9.5 is introduced as a verbatim utterance and given in full. CLAUDE.md's privacy section: "Paraphrase the user, never quote them ... it must paraphrase, never embed a verbatim quote of what they typed. A quoted utterance is real recorded data." Line 200 uses the permitted form. Write `(the user 2026-09-20, who wanted an unlocked pane to mirror whichever came last, a pick or a chat tab switch, and a locked pane to stay on its pick)` and drop the quote frame.

**The incremental walk skips growth inside the last walked turn, which is the motivating case.** The memo keeps `the identity of the last turn walked` (116-117) and, when that id still stands at that index, `walks turns[index+1:] only` (284-286). But a turn absorbs every later atom until an end stop: `segment_turns` opens a new turn only when `cur is None or ended` and otherwise does `cur["atoms"].append(atom)` (kernel/event_model.py:4182-4198); `_turn_id` hashes the trigger's text plus `turn["t"]` (4144-4155), set once at 4245. Executed: `segment_turns` over a synthetic opener plus one Write tool_use with `stop_reason` `tool_use`, then the same atoms plus a second Write: one turn both times, index 0, the same id, atoms 2 then 3, `turns[index+1:]` empty, both Writes inside turn 0. So a listing taken mid-turn memoizes turn N; the next plot's Write lands in turn N; the pane re-asks and nothing is walked. The growth lab (330) passes only when the appended Write opens a new turn. Re-walk from `index` inclusive, or key the memo on the last turn's atom count beside its id; `_artifacts_items` keeps the latest mention per path (kernel/kernel.py:60551-60555), so the re-walk is idempotent. Amend section 4 and 9.4 and add the mid-turn case to the walk test and the growth lab.

**The growth signal repaints the whole pane on every transcript growth.** The signal fires when `(mtime, size)` `moved since the last signal to that socket; the pane then re-asks` (106-107, 273-276): for a working session, every appended record at the pusher's cadence, most adding no artifact. The current `paint()` does `root.replaceChildren()` (ui/webview/artifacts.ts:59), builds a fresh `art-body` (73) with every `` recreated (84) and appends it (102): scroll reset and thumbnails reloaded, several times a minute. And `the last version sent to that client` (275) has no value at watch time, so the first cycle after every selection fires and the pane paints twice. Have the answer carry a listing signature; repaint keyed and in place with `.art-body` scrollTop kept; seed the sent version from the one that rode the listing answer. Pin: a growth with no new artifact repaints nothing; scroll survives a growth.

**The watch lives as long as one socket and is never re-armed.** The pane sends `watchArtifacts {sid}` `when its shown session changes` and the kernel records it on the client (270-272); each socket is its own client (kernel/kernel.py:70083), so a shim redial (`romp:wsup`, 59769) or a relay reopen (`romp:hostRelayUp`, ui/webview/federation.ts:1844) yields a client with no watch. The chat re-arms on both (ui/webview/render.ts:12714-12715); artifacts.ts listens to nothing but window `message` (142). With Refresh gone (107-108) the pane goes silently stale. Federation, read at the head: a watch for a remote whose relay socket is not open is neither a `KERNEL_SETTING` (87-94) nor in `BOOKKEEPING` (123-141), so `sendRemote` drops it with a warn (1588-1589); executed, `grep watchArtifacts` over federation.ts finds nothing. And `routeOutbound` picks the host from the scalar sid (442-447), a null sid falls to the local kernel (475), and each host holds its own `app=artifacts` connection (1761, 1053), so moving from one host's session to another's, or to none, never clears the first host's `client["artifacts"]`; `an unwatched session costs nothing` (276) does not hold. Add a `BOOKKEEPING` entry keyed by type; re-send the watch and re-ask on `romp:wsup` and `romp:hostRelayUp` for the shown sid's host; send an un-watch addressed by the previous sid before the new watch; ignore an `artifactsChanged` for a sid not shown. Pin a redial re-arming the watch and, in the two-kernel lab, no signal after a move away from the remote session.

**A remote host down leaves "Reading the thread" forever, and the wait is text.** 9.1 designs the reachable case only. When the relay socket is not open, `sendRemote` drops `listArtifacts` and dispatches `{type: "warn", text: ...}` on the window (federation.ts:1625-1627); the pane's listener handles `panes`, `sessionList` and `artifactsListing` only (artifacts.ts:145-157), so `loading` stays true and line 75 paints the text forever. `hostIsDown` (ui/webview/host-prefix.ts:61) and `hostDownNote` (96) exist and the plan does not use them. The wait itself is a bare line; the page has no `_pane_spin` (kernel/kernel.py:60462-60471) where the chat, feed and outline pages carry one (60338, 60379, 60410); ui/CLAUDE.md's loading rule: "Reuse that treatment for any new wait state rather than a blank, a bare spinner, or text." Pass two fires that wait on every unlocked tab switch (76-77), since a selection clears `listing = null` (artifacts.ts:52). Paint the strip's disconnected dress and `hostDownNote` when `hostIsDown(sid)`, take the warn frame as the listing's error, re-ask on `romp:hostRelayUp`, and show the loader over the old listing while a request is outstanding. Lab: the remote stopped, the pane says so; the remote back, the listing arrives with no click.

**A reloaded pane never gets the frame 9.5 promises, and an unlocked reload has no stated winner.** 9.5 has the kernel send `activeChat` to `artifacts` clients through `the ready arm` `so a reloaded pane knows the focused session before any switch` (300-303), tested at 322-323. The arm (kernel/kernel.py:68990) gates on `client.get("app") == "feed"` (69004-69005) and fires on a ready the bundle posts; the shim only observes one (59826). Executed, `grep '"ready"' ui/webview/artifacts.ts` finds nothing; the panes that post one are `files.ts:155`, `feed.ts:6886`, `fleet.ts:1182` and `render.ts:21273`. Widening the gate alone changes nothing. Then, unlocked, the persisted sid (312-313) and the ready frame (302-303) both claim the first paint; `whatever came last` (76-77) does not say which, the machine (314-315) has no restore event, and 9.6 pins only that the lock survives a reload (332-333). Have the pane post `{type: "ready"}` after its listener installs (the arm's comment at 68993-69001 says why); widen the gate to feed and artifacts; decide the precedence: locked reads the persisted sid, unlocked yields to the ready frame or holds the pick until a frame with a newer nonce; pin it in `artifacts.test.ts` and the lock lab.

## Smaller points

- Nothing in sections 3, 9.3 or 9.5 says how the picker button and the lock attach their actions; `paint()` rebuilds the root (artifacts.ts:59), the first landing's selector hung its handler on the rebuilt node (67) while its buttons used `delegate` (137), and pass two adds three repaint triggers. ui/CLAUDE.md's click-safe rule: "Never attach the action to a node you rebuild." Give both `data-act` values under the existing delegate and pin them.
- 9.2 dedups `chatTabs` `the way activeTab is deduped` (231-232), but `notifyActive` posts unconditionally with a fresh nonce (render.ts:12705-12711), on every show (13191), and re-announces an unchanged tab (19002, 19270); the only dedup is the kernel's `("activeChat",)` slot (kernel.py:57067) within `_DEDUP_REPOST_S` (49406, 60 s). Describe the compare on its own terms.
- The same sentence posts `{id, name, color}` `only when the set or its order changed` (229-231), so a rename or recolor never reaches the picker. Key the compare on the serialized rows.
- 282-283 says `no lock beyond the pusher's` and, in the same sentence, `the memo is guarded by one lock`. Executed, `grep _pusher_lock kernel/kernel.py` finds nothing; the cycle takes `_clients_lock` around the client list only (58928), and each socket has its own handler thread (70083), so two panes listing one sid race on the memo. Name the lock and what it guards; pin two concurrent requests ending with one memo.

The code lines cited were read at 10f33abb; the probes are named where they were run.

lczh:
Round two, executed against b1ae133d (the implementation; the design review above read 683c997b).

The implementation gets the shape right: the picker lists the shell's union of the chat columns' open tabs in the strip's dress, `nextSelection` is the corrected lock machine and its test pins it, remote sessions ride the existing sid routing, and the two-kernel and follow labs prove the reachable paths. What did not carry over from the design review is lifetimes and repaints: the walk memo still skips the open turn, the watch still dies with its socket, a remote session shown at boot still waits forever, and every growth still rebuilds the pane. Each stands in the code as an executed defect.

## At this head

**The incremental walk never sees a Write inside the turn it last walked, the case the pane exists for.** `_artifacts_mentions` (kernel/kernel.py:60598-60623) reuses the memo when `turns[memo["idx"]]` still carries `memo["tid"]` (60609) and sets `start = memo["idx"] + 1` (60610), so the last walked turn is never read again. But a turn absorbs atoms until an end stop: `segment_turns` opens a new turn only when `cur is None or ended` (kernel/event_model.py:4183) and otherwise appends (4195), `ended` comes only from a stop in `END_STOPS` (4198, 46), and the id hashes the trigger's text plus the turn's first `t` (4144-4155), so a turn grows under the same id. A Write is a `tool_use` stop, so the open turn is a working session's normal state. Executed over the head kernel, real segmenter and real walk: an opener plus one Write segments to one turn, 2 atoms, `ended` False; the same plus a second Write is the same id with 3 atoms. Listing 1 gives `['p1.png']`; listing 2, after the mid-turn Write, still gives `['p1.png']`, though the version moved and the pane re-asked; with the memo cleared the same turns give `['p1.png', 'p2.png']`. The control, growth by a new turn after an `end_turn` stop, lists the new file. The tests grow only that way: `Memo` hands `T3` a fresh id (tests/test_artifacts_list.py:377-379) and the follow lab appends a user record after a seed whose last stop is `end_turn` (tests/test_artifacts_follow_served.py:125-126 over 148-149), so both open a new turn, and with no Refresh left nothing recovers until then. On a scratch copy with `start = memo["idx"]` the same probe lists both paths on listing 2 with the control unchanged; `_artifacts_items` keeps the latest mention per path (60555-60559) and the merge at 60613-60615 is `>=`, so the re-walk is idempotent. It moves the `Memo` pins at 379 and 384, since `walked[-1]` becomes `['T2','T3']` then `['T3']`; a `(tid, len(atoms))` key that walks whole when either differs keeps them. Either way, amend 9.4 (plans/artifacts-pane.md:291 still says `turns[index+1:]` only), add the mid-turn case to `Memo`, and make the growth lab append an assistant Write with no new user record after a `tool_use` stop, which is red on this head.

**A remote session shown at boot never lists.** A reload with a remote sid persisted runs `watch(sel.sid)` and `requestListing()` synchronously at module end (ui/webview/artifacts.ts:232-233). The page loads federation.js before artifacts.js (kernel/kernel.py:60473-60474), so both go through `__rompFed.outbound` (59912); `routeOutbound` picks the host from the sid's prefix through `hostOf` (ui/webview/federation.ts:442-447; host-prefix.ts:16-20 consults no host list); but the first `/tunnels` poll is asynchronous (1098, the fetch at 1637), so no conn exists yet. `sendRemote` (1558-1590) sends only on an open socket (1560), queues only a `KERNEL_SETTING` (1564) or a `BOOKKEEPING` key (1576-1586), and neither `watchArtifacts` nor `listArtifacts` is in either list (87-94, 123-141; the file is not in the diff), so both fall to `senddrop` plus `dropWarn` (1588-1589), a `warn` frame (1626-1627) that nothing on this page shows, since ui/webview/file-view.ts:1749 reads it only mid-edit. The pane's listener (artifacts.ts:201-228) has no `warn` arm; `loading` is now true, the `panes` re-ask is gated on `!loading` (206), a same-id `activeChat` only repaints (76-79), and there is no `romp:hostRelayUp` listener, so line 115 paints "Reading the thread…" for good. A remote host down at selection time ends the same way; `hostIsDown` (host-prefix.ts:78) and `hostDownNote` (113) exist unused, and the wait is bare text where the chat, feed and outline pages carry `_pane_spin` (kernel.py:60342, 60383, 60414; none at 60466-60475). The two-kernel lab waits for the hub to report the peer up (tests/test_artifacts_remote_served.py:193-201) and picks through the card (78-81), never across a reload. Fix: `BOOKKEEPING` entries, `watchArtifacts` keyed by type, `listArtifacts` keyed by type plus sid (the `reqId` check at 223 drops a superseded answer); on `romp:hostRelayUp` for the shown sid's host reset `watched` and re-send the watch and the listing; take a `warn` naming the shown host as the listing's error; paint `hostDownNote` when `hostIsDown(sel.sid)`; use `_pane_spin` for the wait. Lab: a persisted remote selection survives a reload; the remote stopped shows the note; the remote back lists with no click.

**The watch lives as long as one socket, on either road, and nothing tests a reconnect.** The kernel writes `client["artifacts"]` only in the `watchArtifacts` arm (kernel/kernel.py:69962-69973), `_artifacts_signal` targets only clients with it set (60645), and each socket is a fresh dict from `_new_ws_client` (48709). The shim's `onopen` flushes only the down-window queue and re-sends `ready` only while `!readyAcked` (59767; the caps frame latches it, 59776), then dispatches `romp:wsup` (59773) and enqueues a `wsup` frame (59774). The pane listens to neither: `watch` returns early when `sid === watched` (artifacts.ts:68) and `watched` is reset nowhere else, so after a redial the shown session never signals again. Executed in the Growth test's shape with the real `_pusher_cycle`: a second client minted for the same page carries no `artifacts` key and gets zero frames across two growths and three cycles; a re-sent `watchArtifacts` brings one. For a remote host the same drop applies whenever its relay is not open, and `romp:hostRelayUp` (federation.ts:1844) is not listened for, so the watch is never replayed. The chat re-arms on both events (ui/webview/render.ts:16464-16480) and the Files pane re-asks on `romp:wsup` (ui/webview/file-browse.ts:479-483). The Growth tests drive one client dict (tests/test_artifacts_list.py:298); the labs never bounce a socket. Fix: on `romp:wsup` set `watched = null`, `watch(sel.sid)`, `requestListing()`; on `romp:hostRelayUp` the same when the detail's host is the shown sid's; the `BOOKKEEPING` entry above. Pins: the listener text in artifacts.test.ts; in `Growth`, a second client dict that gets nothing until the watch is re-sent, then one frame.

**Every growth signal rebuilds the whole pane twice and resets its scroll, even when no artifact changed.** `_artifacts_version` is `[int(mtime), int(size)]` of the transcript (kernel/kernel.py:60626-60636), `_artifacts_signal` runs every cycle (59042), and the cycle wakes on every push with a 0.5 s backstop (59282), so a working session signals once per appended record. Each signal calls `requestListing()` (artifacts.ts:213), which sets `loading` and paints (63) with the old listing held; `paint` begins `root.replaceChildren()` (95), mints a fresh `.art-body` (113), the one scroll container (ui/webview/artifacts-pane.css:31), and recreates every `` (124); the answer paints again (226); `apply` paints on every unchanged `chatTabs` and `activeChat` too (79). Executed on the head's bundle in a headless browser with 83 rows and 3 thumbnails, `.art-body` scrolled to 400 of 2382: one `artifactsChanged` alone replaced the body and the first ``, put scrollTop at 0, and posted `listArtifacts#2`; the identical answer replaced both again, scrollTop 0; an unchanged `chatTabs` broadcast did it a third time. No test compares the DOM across a growth; the follow lab waits for the new row (tests/test_artifacts_follow_served.py:127). Fix: a listing signature on the answer (path, t, via, exists, `capped`), returning when it equals the painted one; otherwise an in-place rebuild keyed by path with scrollTop kept; no repaint on `requestListing` while a listing is shown. Pin: a same-signature answer leaves the body node and scrollTop unchanged.

**The kernel's relayed `activeChat` echo is applied by id alone, so a late echo moves the pane backward.** Each switch reaches the pane twice: the shell's relay at once, then the kernel's `_relay_active_chat` (kernel/kernel.py:57096-57110, `artifacts` added at 57108) through `_send_active_chat`, whose frame carries the announcement's `nonce` (57057-57059). The pane applies both by id (artifacts.ts:210), `nextSelection` reads no nonce (ui/webview/artifacts-model.ts:59), and a changed sid drops the listing and churns the watch (78). Executed on the bundle: relays A, B, C settled on C with three listings; the kernel's B echo moved the pane to B (unwatch C, watch B, `listArtifacts B#6`), the C echo moved it back (unwatch B, watch C, `#7`); a single switch and its echo added nothing. Two listings and two watch churns for one destination, on stale information. The feed solved exactly this with a nonce watermark (ui/webview/feed.ts:1587, 6518-6528). Fix: track the highest nonce seen from either road and ignore a frame below it; a frame with no nonce still applies. Pin the three-frame ordering in artifacts.test.ts (today only the listener's text, 98).

## Smaller points

- A chat column that empties and closes never re-tells the union. `renderTabs` posts `colEmpty` (ui/webview/render.ts:6655, 8368) and then the empty `chatTabs` (6657, 12721-12724); the shell's `colEmpty` handler runs `close(en.n)` synchronously (kernel/kernel.py:64483-64494), whose `p.remove()` (64400) detaches the frame, so the queued `chatTabs` matches no iframe and `if(!src)return;` drops it (64136). `chatTabsUnion` filters the dead frame (64131), but `chatTabsMsg()` goes out only from that listener (64138) and the pane-load hook (64206), and the first column's set is unchanged, so the picker keeps the ended session as open. The follow lab closes a tab in a two-tab column (tests/test_artifacts_follow_served.py:112-118). Fix: `tellAll(chatTabsMsg())` in the close path, pinned in the shell tests.
- Plan 9.6 still says `(follow, pick locks, toggle, unlock jumps, a closed tab)` (plans/artifacts-pane.md:329), while `nextSelection` keeps the lock on a pick (ui/webview/artifacts-model.ts:60) and the shipped test asserts it (ui/webview/artifacts.test.ts:69, 76); the body's decision 5 still says `a pick locks`. Rewrite both.

lczh:
The test runs behind the comment above, on b1ae133d (merge base with main 10f33abb).

## Tests

- The extension suite at the head, `npm test` (the test bundle, then `node --test` over it): 5174 tests, 5167 passed, 1 failed, 6 skipped. The one failure is the KaTeX layout probe (test 2793), a fraction-stacking assertion in a browser test this PR does not touch, and the suite's standing failure in this environment.
- The two Python modules, each alone under the repo's runner: tests/test_artifacts_list.py 21 passed; tests/test_pane_state_broadcast.py 42 passed. The served labs (tests/test_artifacts_follow_served.py, tests/test_artifacts_remote_served.py) were not run here; the playwright package is absent from this Python, so both would skip.

- kernel/kernel.py `_artifacts_mentions` with `start = memo["idx"] + 1` (60610) changed to `start = memo["idx"]`, the whole kernel directory copied so the sibling loads resolve: the mid-turn probe below lists `['p1.png', 'p2.png']` on listing 2 where the head lists `['p1.png']`; listing 3 and the new-turn control are unchanged. The `Memo` test's recorded walks under the same mutant, in that test's three-listing shape (T1 and T2, then T3 added, then the same again): `[['T1','T2'], ['T2','T3'], ['T3']]` against the head's `[['T1','T2'], ['T3'], []]`, so the pins at tests/test_artifacts_list.py:379 and 384 move under this fix; an atom-count key would leave them.

## Probes

- The mid-turn walk, over the head kernel loaded the way tests/test_artifacts_list.py loads it: the state root rebound to a temp directory, `_cwd_of` stubbed, the real `segment_turns` and the real `_artifacts_mentions`. An opener plus `Write p1.png` with `stop_reason` `tool_use` segments to one turn, id `…:100:b1267f6f`, 2 atoms, `ended` False; the same plus `Write p2.png` is the same id, 3 atoms, `ended` False. Listing 1: `['p1.png']`. Listing 2: `['p1.png']`, p2.png absent. Memo cleared, the same turns: `['p1.png', 'p2.png']`. Control: the first Write's stop changed to `end_turn`, then a new user record and `Write p3.png`: two turns, and the listing is `['p1.png', 'p3.png']`.
- The watch across a redial, in the Growth test's shape: a real transcript file, the real `_pusher_cycle`, clients minted by `_new_ws_client` with no sender thread and registered by `_register_ws_client`. Client A watches; one growth gives it one `artifactsChanged`. Client B, minted for the same page id, has no `artifacts` key; two growths and three cycles give B zero frames; `watchArtifacts` re-sent on B and one more growth give B one frame. A was minted with no socket, so the same-page supersession did not retire it; only B's behavior is read.
- The pane in a headless browser: the head's ui/webview/artifacts.ts bundled the way the extension's build bundles it (iife, browser, es2020, the same externals), served on a routed http origin so `canPreview` holds, `acquireVsCodeApi` stubbed to record posts, `romp:artifacts:sid` seeded `S1`, frames delivered as window `MessageEvent`s, the door the shim's `deliver` and the shell's relay both reach, and `replaceChildren` on `#artifacts-root` counted. Boot posts `ready`, `requestSessions`, `watchArtifacts S1`, `listArtifacts S1#1`; the count is 2 and the body reads "Reading the thread…". A listing of 83 items, 3 of them existing images: 83 rows, 3 thumbnails, count 3; `.art-body` scrolled to 400 of 2382. One `artifactsChanged {sid: S1}`: count 4, `.art-body` a new node, the first `` a new node, scrollTop 0, `listArtifacts S1#2` posted. The identical listing answered on `#2`: count 5, the body new again, scrollTop 0. Scrolled to 400 again, then a `chatTabs` broadcast naming S1: count 6, the body new, scrollTop 0.
- The late echo, the same page: relays `{romp:'activeChat'}` for A, B, C with nonces 1, 2, 3 settle on C; the posts are unwatch S1, watch A, list A#3, unwatch A, watch B, list B#4, unwatch B, watch C, list C#5. The kernel's frame `{type:'activeChat', id:'B', nonce:2}`: the shown sid becomes B; posts unwatch C, watch B, list B#6. The frame for C, nonce 3: back to C; posts unwatch B, watch C, list C#7. Control: a relay for D and then its echo: one set of posts (unwatch C, watch D, list D#8) and nothing from the echo, though the root count still rose by one on it, `apply` painting an unchanged selection.
- Reading the code, for the boot-time drop: `_artifacts_page` orders federation.js before artifacts.js (kernel/kernel.py:60473-60474); the shim's `acquireVsCodeApi().postMessage` calls `__rompFed.outbound` when federation is present (59912); the federation class calls `poll()` from its constructor (ui/webview/federation.ts:1098) and `poll` awaits `fetch("/tunnels")` (1637), so no conn exists when the pane's module-end sends run (ui/webview/artifacts.ts:232-233); `outbound` routes through `routeOutbound` (1520) and `sendTo` hands a non-local host to `sendRemote` (1528-1534). `grep -n 'watchArtifacts\|listArtifacts' ui/webview/federation.ts` at the head: no lines. `grep -n -i 'wsup\|hostRelayUp\|scrollTop\|"warn"\|nonce\|hostIsDown\|hostDownNote' ui/webview/artifacts.ts`: no lines.
- Reading the code, for the shim and the relay: `ws.onopen` (59767) sends the queue, then `readyMsg` only if `bundleReady&&!readyAcked&&!readyQueued`, dispatches `romp:wsup` (59773) and enqueues `{type:"wsup"}` (59774); `readyAcked` is set on the `caps` frame (59776). `_relay_active_chat`'s audience is the `("feed", "artifacts")` clients of the wid (57108) and the ready arm's gate is the same pair (69092-69093); `_dedup_sig` strips only `now` and `buildId` (49399, 49409-49415), so two echoes with different nonces both go.
- `chatTabsMsg()` at the head is called from the `chatTabs` listener (kernel/kernel.py:64138) and the pane-load hook (64206) only; `close` removes the pane at 64400 and `__rompColGone` (61223) repaints the columns' own state and re-tells nothing. `noteColumnEmptiness` posts `colEmpty` when the column's members are all gone (ui/webview/render.ts:8353-8368), and `renderTabs` calls it before `postChatTabs` (6655, 6657).
- `git diff --stat 10f33abb b1ae133d`: 17 files, 1336 insertions, 96 deletions; ui/webview/federation.ts is not among them.

lczh:
Round three, executed against 990fa5ca.

The walk memo and the watch's lifetime landed, and with them half of the remote-boot item: a remote session persisted across a reload now lists once its relay opens, because federation fires `romp:hostRelayUp` on the first open too and the re-arm answers it. The lows landed as described. The repaint on every growth and the late `activeChat` echo stand unchanged, the down-host half of the remote item with them. One new thing in the code this commit touched: the same re-arm can make a fresh pane ask for its listing twice at boot.

## Landed

**(A) The resume is inclusive.** kernel/kernel.py:60613-60615 sets `start = memo["idx"]`; the docstring (60599-60608) and plan 9.4 (plans/artifacts-pane.md:294-299) say so. Re-executed over the real segmenter and the real `_artifacts_mentions`: an opener plus `Write p1.png` with a `tool_use` stop is one turn, id `…:100:cea73b4e`, 2 atoms; with `Write p2.png` the same id, 3 atoms. Listing 1 `['p1.png']`; listing 2, after the mid-turn Write, `['p1.png', 'p2.png']`; the third listing and the memo-cleared walk the same; the new-turn control unchanged. The `Memo` pins moved as predicted (tests/test_artifacts_list.py:380, 385); the new test at 387-396 is the probe's shape. Worth a comment line in the lab: its redial leg appends a user record after a `tool_use` stop (tests/test_artifacts_follow_served.py:140-141 after 129), and the segmenter absorbs an opener into an unended turn (kernel/event_model.py:4186-4192; executed, one turn, 2 to 4 atoms, same id), so that leg is a mid-turn growth too.

**(C) The watch survives the socket.** `rearm` (ui/webview/artifacts.ts:78-84) re-sends the watch and re-asks on `romp:wsup` for a local sid and on `romp:hostRelayUp` for the shown sid's host (pins ui/webview/artifacts.test.ts:106-107). Executed on the extension's built bundle: boot posts `ready`, `requestSessions`, `watchArtifacts S1`, `listArtifacts S1#1`; a `romp:wsup` posts the watch and `#2`; a relay event for another host posts nothing. With `TESTHOST:S9` persisted: `romp:wsup` and an `OTHERHOST` event post nothing, a `TESTHOST` event posts the watch and `#2`. The lab leg (tests/test_artifacts_follow_served.py:132-143, test at 287-293) ran here and passed.

**(B), the reload half.** Reading the code: `sendRemote` still drops both ops before the relay opens (ui/webview/federation.ts:1558-1590; neither type is in `KERNEL_SETTING` 87-94 or `BOOKKEEPING` 123-141), but the open handler dispatches `romp:hostRelayUp` on every open, the first included (1841-1845), after `flushPending` (1826), and the executed leg above shows the pane re-sending both on it. The persisted remote selection recovers with no `BOOKKEEPING` entry.

**The lows.** The focus keep (artifacts.ts:107, 156, 170; executed: `#art-pick` focused, a signal and its answer, `activeElement` stays `art-pick`). The known name (41, 116, 223; ui/webview/artifacts-model.ts:67-75; executed: tabs `docs` then none, the button reads "docs not open"). The chat-frame gate (kernel.py:64140, pinned at tests/test_artifacts_list.py:508). The signal in its own `try` after the push's `finally` (59041-59047, still under `if any_client:` at 59033). The contract (plans/panes-as-data.md:203-208). The `.art-btn` rules gone.

## Still open

**(D) Every growth still rebuilds the pane twice and resets its scroll.** `requestListing` paints at artifacts.ts:64, `paint` begins `root.replaceChildren()` at 108 and mints a fresh `.art-body` (126) and every `` (137), the signal re-asks at 227, the answer paints at 240, `apply` paints on an unchanged selection at 91. Executed with 83 rows and 3 thumbnails, `.art-body` at 400 of 2684: one `artifactsChanged` replaced the body and the first ``, scrollTop 0, `listArtifacts S1#3`; the identical answer replaced both again, 0; an unchanged `chatTabs` a third time, 0. The round-two fix stands: a listing signature on the answer, an in-place rebuild keyed by path with scrollTop kept, no paint on `requestListing` while a listing shows.

**(E) The echo is still applied by id alone.** artifacts.ts:224 reads no nonce; `nextSelection` (artifacts-model.ts:59) takes none. Executed: relays A, B, C with nonces 1 to 3 settle on C; the kernel's B echo (nonce 2) posts `unwatch C`, `watch B`, `listArtifacts B#8` and the pane shows B; the C echo moves it back with `#9`; a relay for D and its echo add one set. Track the highest nonce seen from either road and ignore a frame below it; a frame with no nonce still applies.

**(B), the down-host half.** With the host down at selection time the browser does not dial it (ui/webview/host-prefix.ts:95), so no relay event comes and the pane holds the bare "Reading the thread…" (artifacts.ts:128): no `warn` arm (executed: federation's own `warn` text delivered to the page changes nothing), `hostIsDown` (host-prefix.ts:78) and `hostDownNote` (113) unused, no `_pane_spin` where the chat, feed and outline pages carry one (kernel.py:60342, 60383, 60414; none at 60466-60475). The remote lab's four legs (tests/test_artifacts_remote_served.py:240-283) neither reload nor stop the remote. Paint `hostDownNote` when `hostIsDown(sel.sid)`, take a `warn` naming the shown host as the listing's error, spin the wait, add both legs.

- The `colEmpty` drop stands: kernel.py:64140 still returns when no iframe matches the source; the commit's change on that line filters a non-chat frame, a different case. `close` (64390-64408) tells nothing; `chatTabsMsg()` is called only at 64142 and 64210.
- An `unwatch` for a previous remote sid while its relay is closed is still dropped with a `warn` nothing here shows (federation.ts:1588-1589, 1625-1628); a `BOOKKEEPING` entry for `watchArtifacts` keyed by type would carry watch and unwatch alike.
- plans/artifacts-pane.md:335 still says `(follow, pick locks, toggle, unlock jumps, a closed tab)` against ui/webview/artifacts.test.ts:69 and 76; the body's decision 5 still says `a pick locks`; line 302 says the kernel never re-walks what it walked, which the inclusive resume contradicts by one turn per ask (tests/test_artifacts_list.py:385). The body's round-two paragraph leaves the Python figures as the literal `(SHARDS)`.

## New at this head

- **A fresh pane can ask for its listing twice at boot.** The shim dispatches `romp:wsup` on every open, the first included, after flushing its queue (kernel.py:59767-59773). The pane's module end sends the watch and `listArtifacts #1` (artifacts.ts:246-247); while the socket is not yet open they queue (`send`, 59830-59834), the open flushes them, and the event runs `rearm("")`, which sends a second watch and `#2`; the answer to `#1` is dropped at 237. Executed: boot posts `#1`, a `romp:wsup` posts `#2`. Reading the shim for the order: the socket dials at 59914, before federation.js and artifacts.js load (60473-60474), so which lands first is a race between the handshake and two script fetches; when the open comes second, every pane load costs the kernel two listings. The lab installs its `romp:wsup` counter after boot (tests/test_artifacts_follow_served.py:135), so it cannot see this. Re-ask only after a `romp:wsdown` (59820, 59927) has fired on this page; re-send the watch alone otherwise.

## Runs

- The extension suite, `npm test` at the head: 5175 tests, 5168 passed, 1 failed, 6 skipped. The failure is the KaTeX layout probe (test 2793), the standing failure in this environment. The body's figure for the same suite (5169 pass, 0 fail) is the author's.
- Behind the test lock, one module at a time: tests/test_artifacts_list.py 22 passed in 1.94 s; tests/test_pane_state_broadcast.py 42 passed in 0.94 s; tests/test_artifacts_follow_served.py 6 passed in 70.89 s over the extension's own build and a headless Chromium. tests/test_artifacts_remote_served.py was not run. The body's Python shard figures are a placeholder at this head.
- The pane probes ran on the built `artifacts.js` over a plain http origin (so `canPreview` holds), `acquireVsCodeApi` stubbed to record posts, frames delivered as window `MessageEvent`s, `replaceChildren` on the root counted, the body and first `` marked before each step.

ksarma:
A note after the merge. Executed against the head 990fa5ca5; every reference reads the same at the merge commit b7d7be36d. The last commit got no fresh correctness read. Where an item overlaps the comments above, only what is new is written; the inclusive resume, the re-arm, the kept name and the contract stand (the 20:26Z comment). Of the note on https://github.com/romp-on/romp/pull/1911, this change closes the socket-drop loss, replaces the native select and memoizes the prose walk; it leaves the bare wait text, the whole-root rebuild and the walk call outside the parse's try (kernel/kernel.py:60679); the rest, except the replaced bar and sheet lines (not re-read), and the note on https://github.com/romp-on/romp/pull/1922 are outside these hunks.

**Every growth still rebuilds the pane** (the 20:26Z comment's D). `apply` (ui/webview/artifacts.ts:91) rebuilds on an unchanged sid too, and its unchanged test must cover the none-label and the open mark; update the bar's two buttons in place. The signature must cover exists, refused, t, via and name, or a deletion never repaints; the listing is newest-first, so a carried scrollTop must add the inserted rows' height. For the late echo (E), the nonce is a per-chat-frame counter, one per column, so one largest-nonce watermark is unsound: remember the last relay's {sid, nonce} and apply an echo only when its id matches and its nonce is not below it, or before any relay has arrived.

**A prose mention refused at one listing is re-judged only while its turn is the memo's last.** Rule 2's admission runs inside `_artifacts_walk` at kernel/kernel.py:60528 and the memo at 60616-60619 keeps the survivors only, so a file mentioned in an earlier turn and created afterwards out of band never lists while the memo stands, where the base 1bcf0f89f lists it. Fix: keep rule 2's unadmitted candidates in a second per-sid map keyed by absolute path with the raw token and atom uuid, re-run the cache-or-disk admission on every answer, and latch an admitted one through the same `>=` merge. Test: mention in T1, Write in T2, list, create, list. Amend 9.4 (119-120, 296-299).

**A chat column that empties and closes never re-tells the union** (the 19:04Z comment's first smaller point). The split script dispatches romp-chat-cols with open false after `p.remove()` in close (kernel/kernel.py:64404-64406) and closeBelow (64418-64421), so a collapse-script listener there reads the union with the gone frame excluded; the detail carries the column number, not the frame id, so prune every `TABSETS` key with no element in the document. No test reaches this road.

**The signal stats once per watching client, and the design text says otherwise.** `_artifacts_signal` at kernel/kernel.py:60650-60653 stats per client; at 2.4 microseconds a stat the weight is the text: plan line 190 rules polling out and 301-302 calls the check an exact event, while it rides the pusher's 0.5 s backstop. Fix: a dict from sid to version built once per cycle, the clients' sids and stamps snapshotted under `_clients_lock`, and at every restatement (190, 277-278, 301-302, the docstring, the call-site comment, the body): the compare rides the event-woken cycle, one stat per watched session, no new timer.

**Under federation a switch reaches the pane from three speakers, and the late one decides.** `_relay_active_chat` at kernel/kernel.py:57108 and the ready arm at 69096 read no client kind, so the pane hears the shell, the local kernel and the tab's host kernel; two quick switches, or a stale ready seed from a remote kernel, leave the pane on the wrong session (a pin needs an injected delay or a kernel unit test). Smaller fix: skip relay-kind clients in the fan-out and the ready arm's send while still writing the record and the nonce, so `_reaffirm_active_chat` keeps its answer to a jump into a closed remote session. A federation-side variant (drop a non-local activeChat in inboundNow) must let a reaffirm frame through.

**Plan and body text that still state what the code retracted.** Prose only.
- Lines 317-318 introduce a verbatim quotation as the requester's own words (the 17:35Z comment's second item), as does the label at ui/webview/artifacts-model.ts:52; paraphrase in the form line 200 uses.
- Line 201 pins the numbers to main at 10f33abb, where most resolve, and eleven miss at the head; name the anchors (`_turn_id` for 4144, the pane-broadcast tell at kernel.py:64123 for 239) or re-derive and move the sha to the head.
- Line 335 still says "pick locks" (named above); "selector" survives at 3, 81 and 178 (204 and 213 describe the first landing's select in the past tense; leave them); 110-112 say the picker's enumeration where the path comes from `_session_row`.
- Lines 301-302 (named above), section 4 at 117, 9.6 at 332 and the Memo docstring at tests/test_artifacts_list.py:343-345 still say the kernel never re-walks what it walked (380, 385 and 396 expect otherwise); write that it re-walks the last walked turn and those after it, once per ask.
- The body's decision 5 still says a pick locks (named above); the shard figures are in the body now.

## Smaller points

- ui/webview/artifacts.ts:227 re-asks during an in-flight ask and the listing arm drops every answer but the newest, so a growth outpacing the listing paints nothing until it pauses; set `stale` instead and re-ask once after an accepted answer, gate the panes arm at 220 too, but never the re-arm at 81.
- The watch arm at kernel/kernel.py:69973-69977 and the cycle at 60651-60657 race unlocked on the sid and the stamp, costing a redundant listing and a wrong stamp; re-check the sid, compare and stamp under `_client_lock`, the stat outside.
- The signal call at kernel/kernel.py:59043-59047 is not under `_job_stage`, so a stalled stat lands in the jobs container on /perf unnamed; wrap it as a stage named artifactsSignal.
- kernel/kernel.py:64140 refuses a non-chat frame's chatTabs while the activeTab relay at 60983-60986 admits any protocol frame; the only poster, `notifyActive` (render.ts:12727-12734), runs in chat frames, so the same predicate refuses nothing legitimate.
- A rewind inside an open turn keeps the turn's id while atoms vanish, so the merged map keeps the mention; memo the atom count n and the uuid of atom n-1, resume only while both match.
- The controls: a repaint while the picker card is open rebuilds `#art-pick` under it, so Escape or Tab lands the focus on the body, and an onClose must treat focus inside the card, not only on the body, as the trigger; set aria-expanded in paint from `contextMenuOpen`, role menuitemradio with aria-checked on the rows, and an aria-label on the lock, none of which exist; the title at artifacts.ts:114 names the listed session while the button reads "not open", so branch it on `shown.open`.
- The dress: the picker's rows have no hover or focus rule (ui/webview/artifacts-pane.css:19), so a hovered row wears the selection fill of styles.css:778, and a wash rule for hover and focus restores the 9.3 precedent, not a 4.5:1 pass; styles.css:749 inks the picker's name through the T390 lightness floor while the strip's label at 744 paints the raw colour, against 9.3's "reads as the strip's tab would" (265), so drop the oklch line at 749 or state the divergence and soften 265.
- Plan 9.4 and 9.6: lines 291-292 name a memo that does not exist and a lock the op does not use (the 17:35Z comment's last smaller point; the code has `_ARTIFACTS_MEMO`), and line 273 says a null sid on a clear where the pane sends the previous sid with unwatch true (artifacts.ts:70); no refusals paragraph landed, and for the down host (the 20:26Z comment's down-host half) federation's warn frame carries no sid or reqId, so read `hostIsDown` from host-prefix.ts, repaint on federation's down-set window event, never a timer, and set an error listing.

## Tests

The remote re-arm (the 20:26Z comment's B) is pinned by two regexes at ui/webview/artifacts.test.ts:106-108 and no executed test: a mutant under which remote hosts never re-arm leaves the labs and the extension test green, and a reload with the remote sid and the lock persisted, asserting the frame order, catches it. The redial leg (tests/test_artifacts_follow_served.py:139-141) pins the re-armed watch, not the re-ask, since its Write lands after the redial; add a Write between the socket close and the redial. The lock leg never unlocks; the pins at tests/test_artifacts_list.py:508 and 449 and artifacts.test.ts:109 are text matches that mutants pass; the kernel-side fix above re-aims 449.

PR #1926, opened 2026-09-20
The bottom box's pre-growth read adds back the box's footprint, not its content height (the held-mail lab's last flake)

henrypinkard:
The held-mail chat lab's at-bottom read went red once more in a whole-suite run on the devbox (2026-09-20, the served labs shard of PR 1925's verification), and the payload PR 1923 added to that assertion names the cause:

```
geometry: sh 8819, scrollTop 8174, clientHeight 447, box 189.265625; pass: n 2, height 187.265625, repinned false
```

The approval box appeared and the reader stayed 198 px above the bottom, because the pass did not re-pin. PR 1915's pre-growth decider (the fallback when a scroll event in the window has revoked the follow record) added the observer's content-rect delta (187.27 px) back to the viewport's height, but what the viewport gave up is the box's footprint: its border box plus its 8 px top margin, 197.27 px. Off by 10.7 px, the reader read as scrolled up, outside the 2 px band, and nothing re-pinned. A box growing while shown has no margin change, so the two deltas agree there, which is why the lab's deterministic race scenario (the note step growing under a jump) never caught it: the case is a box that first SHOWS while the record is off.

The fix: the observer reads the box's footprint at every pass (`boxFootprint`, the border box plus the vertical margins, 0 while hidden) and adds the footprint change back (`atBottomBeforeGrowth`, pure, in `scroll-keep.ts`). Red first: the pure helper's test runs the captured geometry (the footprint puts the reader at the bottom, the content delta does not); the executed red is the shard run itself, quoted above. The pins in box-below.test.ts and the import pin re-aimed.

Verification on the head: tsc; the three pin files; the held-mail chat lab (6 of 6); the whole extension suite (5172 tests: 5166 pass, 6 skipped, 0 fail); the whole served suite (113 lab files: 585 passed, 2 skipped, 0 failed, the held-mail chat lab among them).


ksarma:
The evidence below was executed against eba5eaa5f, the PR's head; file and line references are on that commit. The PR merged as ef13dd803 by a plain merge, and the five files read the same there.

The mechanism is as stated: in the browser with the real stylesheet, the height `#content` gives up at a first show of `#notices` equals the footprint (197.265625) to 0 px in 6 of 6 runs, and the content-rect delta misses by 10 px; with the recorded mode removed from the decision, the base's first leg goes red with the payload's shape in 7 of 9 boot runs while the head re-pins in 6 of 6.

- The wiring, `dfoot` in place of `dh` at `render.ts` 14063 and the margins inside `boxFootprint` (14034 to 14039), is guarded by text pins alone (`box-below.test.ts` 47 to 49); no executed test discriminates it deterministically. A head with the call site reverted to `dh` passes the unmodified held-mail lab in 8 of 8 runs here, CPU-throttled or not: the intact record re-pins and the geometric read is never consulted, so the lab tells the mutant apart only when the load ordering lands, which is how the PR body's red arose. Across the nine test files naming these helpers, that mutant reds one assertion, the line-47 regex, so a plain revert does not land silently; what nothing sees is a footprint read wrong at run time, or a rewrite that re-aims the regex with the revert, and the new pure test exercises the helper without its call site. This is a coverage gap in `tests/`, which the PR does not touch, not a defect in the product code. Fix, executed with product code untouched: a second lab scenario, page-side hooks, no timers. Boot with no held message written before the kernel starts, wait until the transcript is scrollable and at the bottom, have a `MutationObserver` on `#notices`' style attribute dispatch one synthetic scroll event on the transcript when display leaves none (the microtask runs before the rendering update, so `followReader` records the grown geometry: the revoked record), then write the held message and assert `first.atBottom` and the pass's repinned flag. Base red 4 of 4 and the `dh` mutant red 4 of 4 with the payload's shape, head green 5 of 5. For the lab's version: assert the hook's premise too (fired exactly once, post-scroll distance about the footprint), or the leg can pass vacuously; key the write on an event, not a quiet period (the at-bottom `waitForFunction` alone sufficed; a wait on the footer's pass event is the fully event-keyed form); keep the boot-time leg beside it, since the class writes the held message in `setUpClass` and that leg covers the backfill path, so this needs its own driver run or class; and the pre-show clientHeight 645 against 447 after pins the `dfoot` arithmetic inside the 2 px band, which the pure test cannot reach.

### Smaller points

- A margin-only change of an element below fires no observer pass, so `lastFoot` (`render.ts` 14071) goes stale and the next growth's `dfoot` (14052) understates what the transcript gave up (dense chrome turned on with `#notices` shown: `dfoot` 86.73 against a fresh delta of 90.73 at the next held message, and under a revoked record nothing re-pins; the ON direction only, since OFF over-adds inside the band); the fix executed here is `const dfoot = lastFoot > 0 ? dh : foot;`, `lastFoot` still tracked so a first show after a hide is recognised, `tsc` clean, the pin at `box-below.test.ts` 48 moving with it, its residual a margin change and a growth in one frame, where the head is exact. The alternative, re-reading `lastFoot` on the settings event (`applyDenseChrome`, `render.ts` 20717), cannot be a blind reset over every element, since the same toggle does fire a pass on `#bg-tasks` and a reset before that pass gives it `dfoot` 0 for a footprint change of about 23 px, five times the error being fixed: reset only an element whose content height after the forced layout still equals its `lastH`, or run each pass body synchronously on the event so the observer's own pass sees `dh` 0 and `dfoot` 0 (a shared footprint read at every pass does not close this, since the toggle fires no pass on any element below in this scenario).

- Each pass adds back only its own element's footprint change (`render.ts` 14063), so two elements below growing in one frame under a revoked record leave every pass reading the reader as scrolled up by the other's growth (`#bg-tasks` and `#statusline` grown in one frame under a jump: distance 156 after, both passes repinned false at head and base alike; inherited from PR 1915's per-element decider); the direction of a fix is one shared record of the three footprints, each pass adding back the net change since the last pass and refreshing all three, gated on `dfoot > 0` rather than `dh > 0` (the write's own gate on `dh` stays), since the `dh` gate lets a shrink-first frame (`#bg-tasks` hiding while `#footer` grows) refresh every footprint without deciding, so the `#footer` pass sees no delta, a case the head re-pins today, while the `dfoot` gate re-pins all five scenarios; boot ordering can put two first passes in one broadcast (1 of 4 boots here), so the lab's at-bottom red keeps a boot-time road this PR does not close, and a lab scenario for the two-growth frame and the shrink-first frame would cover it.

- `atBottomBeforeGrowth` (`scroll-keep.ts` 97 to 98) subtracts a fractional footprint from an integer distance while the clientHeight the transcript gave up is that footprint rounded up or down, so when it rounded up a reader exactly 2 px from the bottom reads as 2.09 to 2.73 px, outside the band, and with the record revoked nothing re-pins (9 of 10 fractional `#footer` padding-bottom positions at footprint 98.90625, the share being the fractional part, about 27 percent at the lab's footprint; no writer parks a follow-mode reader at exactly 2 px, so the state is a gesture's transient landing spot); `Math.ceil(dfoot)` is not a fix, since at padding 0.6 the loss rounded down (98 for 98.906) and ceil reads a reader 3 px up as at the bottom, re-pinning outside the band with the record intact, and at the lab's footprint rounding down is the common case, `Math.round` fails too (the residual can be 0.73), and a cached integer loss per element goes stale across a resize that fires no pass, so the safe change is stating the residual in the helper's comment and in the `boxFootprint` comment (14032 to 14033), whose line about exactly what the transcript's height gave up is true of the layout height, not of the integer clientHeight the formula reads; otherwise the 1 px edge stands accepted.

### Tests

- The lab as shipped is green at the head (13 of 13) and at the base (3 of 3), so alone it does not reproduce the base's red here. The (1b) leg's scroll-up premise is a swallowed wait (`tests/test_held_mail_chat_served.py` 121, the catch on the jump-chip wait), inherited from PR 1915 and untouched here: with the scroll-up at line 120 changed from -300 to -0 the file still passes 6 of 6, the jump wait timing out at 15 s, and the fix, for a follow-up, is to record that wait through `timed` and assert its ok flag in the race leg, the first leg's own pattern, with the pre-click distance beside it. The pure test's fourth case (`scroll-keep-on-show.test.ts` 177), `atBottomBeforeGrowth(8819, 8422, 397, 50.78)`, places the reader at the post-growth bottom, where the distance after the add-back is -50.78 and `atBottomDist` has no bound on the negative side, so any `dfoot >= -2` passes and the case cannot check the equality its message names (not vacuous: it is the only one of the four that reds under a two-sided band, pinning that a reader already written to the new bottom before the pass still reads as at the bottom); the fix is `atBottomBeforeGrowth(8819, 8371.22, 397, 50.78)` true beside the same call with `dfoot` 0 false, with a message that states the geometry rather than the equality, and a negative-distance case kept with an accurate message or dropped knowingly, since both new cases stay green under the two-sided band.

PR #1928, opened 2026-09-20
The review's two prose points: the peer double's docstrings say the marker pop is the write still to come, and the rebound-class claim is narrowed to a bound body

lczh:
Tier: fix

The first loop of `hydrate` (#1813) meets a body a peer thread has already filled and, when the peer's memo entry is gone, counts the atom filled and returns. The peer's `_hydrate_one` wrote the message first and a kind-u atom's tool result after it, so a consumer in that window (a call from another thread while the peer's put had already left the memo, which a small cap makes immediate) was handed a body with no `toolUseResult` for that one build; the kernel reads it for diff rows and answers. Separately, the branch tested the body's class with `isinstance` against `_LazyBody`: the module loader re-executes an existing name into the same module object, rebinding the class, so a sentinel built before a re-execution failed the test, was counted filled without a read and kept its marker for the caller's next body read to raise on. Unreachable in the product today (the module is re-executed only at import time), but the wrong test.

`_hydrate_one` now writes a kind-u atom's tool result before its message, as kind k writes its skill text before its message, so the body a consumer meets is whole the instant the message lands; only the marker pop is still in flight then, and readers key on the body type, not the marker. The first loop's branch keys on the body's source slot (`hasattr(msg, "source_path")`) instead of its class; a None message has no slot either and counts filled as before. The locking is unchanged: the arm's comment names the residual (a present marker with the entry gone means the peer's pop is in flight) rather than paying for the fill inside the memo lock per atom. Four tests join the module, reusing its harness, memo double and synthetic documents: the two arms of the branch, which the merged head left unpinned, plus the fill order and the class check.

Validation

- Red-first at the merged head, `tests/test_lazy_source_lifetime.py`: `test_a_kind_u_body_met_mid_fill_carries_its_tool_result` (two threads under a cap of 0, the peer held inside `_hydrate_one` after its message set) fails with `AssertionError: None != {'answers': {'which': 'the first'}}`; `test_a_sentinel_of_a_rebound_class_is_still_read` fails on the marker left in place (`AssertionError: {'k': 'u', ...} is not None`).
- The two arm pins are green at the head and each red under its mutation: `test_a_peers_fill_caught_between_its_statements_is_finished_from_the_memo` with the memo re-read disabled (`if False:`) fails on the marker left in place; `test_a_peers_finished_fill_whose_entry_left_the_memo_counts_the_atom_filled` with the entry-gone arm broken (`if True:`, a gone entry subscripted) fails with `TypeError: 'NoneType' object is not subscriptable`. Both mutations re-run against the final kernel with the same results.
- Green after, one module at a time: `tests/test_lazy_source_lifetime.py` 15 passed; `tests/test_asm_checkpoint.py` 91 passed, 85 subtests; `tests/test_asm_index.py` 26 passed, 3 subtests; `tests/test_asm_mat_lru_release.py` 9 passed; `tests/test_chat_pages.py` 43 passed, 27 subtests; `tests/test_kernel.py` 460 passed, 12 subtests.
- The two existing race tests pass unchanged.

Follows the post-merge review of #1813.


ksarma:
Post-merge read of #1928, executed in a checkout of the merge commit 55d8e8f0c, whose tree carries both changed files byte for byte from 4bbe36428, the PR's head; every file and line reference here is on 4bbe36428 and reads the same at the merge. The changed hunks do what the body and the new comments say, by execution where the claim is behavioural; one adjacent point is under Smaller points.

What each changed piece of prose claims, and whether the code does it:

- The comment on `_hydrate_one` (kernel/event_model.py:7086-7093) says a kind-u atom's tool result is written before its message, as kind k writes its skill text first, so the body is whole the instant the message lands and only the marker pop is still in flight. It is: the tool result at line 7086 precedes the message at line 7087, kind k's skill text at line 7103 precedes its message at line 7104, and every kind's message is its last body write before the pop at line 7109. The two-thread test at tests/test_lazy_source_lifetime.py:279 drives it under a cap of 0 with the peer paused after its message set; the second call returns 1 with the tool result on the atom at return (line 296). At b7d7be36d with the head's test file, line 296 fails with `AssertionError: None != {'answers': {'which': 'the first'}}`.

- The first loop now tests `hasattr(msg, "source_path")` at line 7170, and the comment says a None message has no slot either and counts filled as before. The slot is declared at line 4070 and set unconditionally at line 4075; the class overrides dict reads, not attribute access, so the lookup succeeds on a sentinel and fails on a plain dict or on None. A lazy atom with a None message was probed: `hydrate` returned 1 and the marker stayed.

- The rebound-class comment (lines 7180-7187) and the docstring at tests/test_lazy_source_lifetime.py:302-308 say the loader re-executes the file into the same module object at every import, rebinding the class; a bound body built before that is read now; an unbound one still fails loudly further down the loop, its stale sentinel being no path; and the product re-executes only at import, before any body exists. The loader claim is true: kernel/loadsource.py:38-55 re-executes into the existing module when the name is already registered, and the product's two loads, kernel/kernel.py:52 and kernel/judge.py:96, run at module import. The test drives the bound arm by rebinding the name to another class at line 312: green at the merge commit, red at b7d7be36d with the marker left in place at line 315. The unbound arm was probed: a stale sentinel that is not the module's current `_UNBOUND_LAZY_SOURCE` passes the identity test at line 7193 and the None guard at line 7196, and the failure is a type error from the file open on the sentinel. Loud, as the docstring says, though incidental rather than the `LazyBodyRead` refusal at line 7197; no test pins it, and the product cannot reach it.

- The doubles' docstrings (tests/test_lazy_source_lifetime.py:16-20, 44-46, 70-72) say a memo built with `puts` False is a peer whose put has left the memo already, the atom filled and the memo empty for it; that the mid-fill double sets the message alone, so the arm's own fill puts the tool result in place and the marker pop is the write still to come; and that the paused double blocks the other thread's first message write with its memo put behind it. Each does that: the base double's `get` puts only when the flag is set and fills through its hook, the mid-fill hook writes the message and nothing else, and the paused atom blocks on the first message write from any thread other than the one it is handed, the test's own thread. Under a cap of 0 the put at line 7231 leaves the memo in the next statement, the eviction loop at line 7232, the empty memo the two-thread test asserts.

## Smaller points

- The class check the comment at line 7180 calls the wrong test survives at the module's two other sites: `is_lazy` at line 4100 and the sentinel class's equality method at line 4090, both class checks against the name. The mechanism is the comment's own: after a re-execution both sites resolve the name from the module's globals to the new class, so a sentinel built before it answers False to the predicate, two sentinels of the same uuid both built before it compare unequal, and a sentinel of the new class compares unequal to an old one of the same uuid, the mis-answer the first loop no longer gives; each reproduces when the loader re-executes the module around a bound body. The reach today is the bound the PR gives the loop, none in the product, so a test module that imports the name after another module built bodies; `is_lazy` has no product caller at 4bbe36428 or at the merge, only its definition and the name roster at tests/test_lazy_body_readers.py:42. Fix: key the predicate on the slot the first loop keys on, `hasattr(atom.get("message"), "source_path")`, and equality on the same slot test plus a uuid match; or leave both as class checks with a comment saying they are import-time-safe, so the next reader does not take the first loop's comment as the module's rule.

- Wording only: the entry-gone arm's comment (lines 7176-7178) names self-eviction under a cap the record does not fit under as the cause of a gone entry. Eviction by later puts from any thread, between the peer's put at line 7231 and its fill at line 7235, is the other cause; the arm behaves the same for both, and the PR body phrases the cap as an example.

## Tests

- tests/test_lazy_source_lifetime.py: 15 passed at the merge commit; at b7d7be36d with the head's test file, 2 failed and 13 passed, the reds being the fill-order test at line 296 and the rebound-class test at line 315, with the messages the body gives. The two arm pins are green at b7d7be36d, since they pin arms #1813 already had, and each reds under its own mutation of line 7173: `if False:` reds only the mid-fill test, on the marker left in place. Forcing the arm with `if True:` reds only the entry-gone test, on a subscript of None. Both match the body.

- All four new tests pin behaviour by execution; there is no source-text pin in the diff. The base double's refactor and the document builder's tool flag leave their default paths unchanged, no existing test was weakened or renamed, and the two existing race tests are among the 15 green. No test covers the unbound arm of the rebound case.

- Neighbours at the merge commit, one process each: tests/test_lazy_body_readers.py with tests/test_asm_checkpoint.py, 97 passed; tests/test_chat_pages.py, 43 passed; tests/test_asm_index.py with tests/test_asm_mat_lru_release.py, 35 passed. tests/test_kernel.py was not re-run for this note; the body reports 460 green. The two ui tests that read event_model.py pin lines outside the hydrate code.

- The label fix matches the diff: commit 89f2b6ba1 changes product code, the write order and the predicate, with two tests red before it; 4bbe36428 adjusts prose, which is all the title names.

PR #1929, opened 2026-09-20
plans/judge-prompt-experiments.md: a design note for measuring a judge prompt change before it moves cards

henrypinkard:
Documentation only: plans/judge-prompt-experiments.md, a design note for a read before code.

**Why.** The user's decision (2026-09-20, paraphrased) that Completed must be safe to clear without reading means the judges' prompts change so that offers of a next step, open questions and undone items file under the category becoming Needs you (plans/needs-you.md, the cards owner's design). Prompt changes move cards, so the user wants a quantitative experiment before any tweak lands.

**What the note designs.** The premise checked in code (where `PLAN_SYS`, `CLOSER_SYS` and `UNBLOCK_SYS` live and which clause of the closer's points the other way today; how `record_verdict`, `rollup_status` and the feed's card loop turn a verdict into `completed` or `needs_input`; what the override journal and the verdict logs record of the user's own card actions, by count; what `_ab_close_session`, `_ab_classify`, `_rebind_state` and the usage ledger already provide). The corpus: endings from the user's own history in four classes, built read-only into the cache root, never the repository, a mail or a body. Labelling: the user's recorded actions first, a labeller pass second with its agreement measured before its labels count. The harness: copies under scratch roots, the judge module rebound onto them, the prompt swapped per arm, the planner, closer and unblocker run per ending with a second build for flaps, the spend summed from the ledger, every run capped and paid passes only in a held quiet window after the manager's go-ahead. The budget from the last week's ledger means: about 64 dollars per arm-pass on 300 endings, about 380 for three arms with the flap re-runs, a 60-ending pilot at about 80. The measures (leaks into Completed, false interrupts, flaps, cost per pass) as a cleanplots figure with the numbers in a table, the landing gate, the roads not taken, and the harness's own red-first tests.

**Verification.** No code; the added lines carry no em-dash, none of the ruled-out vocabulary and no private string (grep counts zero); no transcript text or card title appears. Tier: docs.


ksarma:
The note merged as ccd0e7166 within the hour, so this is a note on the merged text. Its statements about the code were checked at that commit, where `kernel/judge.py` is byte-identical to the PR head ed23b4e20 and, of the files cited here, only `kernel/kernel.py` differs between the two: the behavioural claims by executing `record_verdict` and `rollup_status` in a hermetic scratch state root over synthetic goals, the rest by reading. Every file and line reference is on ed23b4e20.

The body's claims check out on reading: `PLAN_SYS` (judge.py 4305), `CLOSER_SYS` (13203) and `UNBLOCK_SYS` (15545) are module constants read at call time in `plan_llm` (8617), `closer_llm` (14311, called from `_close_turn` at 14957) and `unblock_llm` (15582), so assigning the attribute swaps a candidate in. The closer clauses are paraphrased correctly, `UNBLOCK_SYS` carries the moot phrase quoted, the journal ops are as described and `followup` rows are written only for a reopen with `src` `user`. `_rebind_state` (185) repoints the usage file too, and `_log_judge_usage` (1582) writes a `cost` field, with a zero-cost `err` row when an error envelope still carries a fast-mode readback, so summing `cost` under the run's root is right. `_ab_close_session` and `_ab_classify` do what the note says. The budget arithmetic is consistent (34 + 25 + 5 = 64; 384 for three arms with re-runs against the note's 380; about 83 for the pilot against its 80). No em dash and none of the ruled-out vocabulary in the added lines. The two points under the first header concern the one paragraph, line 13, that turns a verdict into a column; both change what the harness measures.

- **Line 13 gives `rollup_status` a status vocabulary the code does not have.** The note says it derives each node's status as `completed`, `blocked`, `working` or `awaiting`. Executed: five synthetic tops carrying a planner `done`, a planner `block`, a closer `awaiting` with `await_kind` `job`, a `user` `clear` and no verdict, plus a child under the last, then `rollup_status(store, True)`: the distinct statuses were `blocked`, `cleared`, `completed` and `working`; no `awaiting` value appeared; the awaiting-stamped top read `working` with `awaitingWhy`, `awaitingKind` and `awaitingAt` on the node; a child node was absent from `store['status']`. Read: the only `status[nid] =` writes are judge.py 8156, 8158, 8161, 8173 and 8175; the loop at 8153 runs over `children.get(None)`, so status is per top, not per node (as step 3 at line 42 already says); the verdict-log replay at 10722 keeps `awaiting` as an annotation like `settle`, never state; the feed sets `col` to `awaiting` only at kernel.py 43006-43008 and its column rule at 43236-43238 files that under `working` (43240 records it as the flavor). Consequence: a harness written from line 13 waits for an `awaiting` value that never appears and has no case for `cleared`, which the rollup does emit for a top the replayed journal cleared and which the feed then skips entirely (42923-42925, no card). Fix: rewrite the clause to `completed`, `blocked`, `working` or `cleared`, per top, with `awaiting` named as a planner op and closer verdict kind kept as an annotation the feed shows as a flavor of Working. Two bounds on the feed clause: the working/blocked split is not a pure function of the store, since the awaiting remap also reads the live session snapshot and a peer-wait map and can move a `blocked` top to Working, and a recheck or rejudging after a reply drops a blocked top to `working` too; so a kernel-less harness computes only the store-derivable part (`blocked` to `needs_input`, `completed` to `completed`, else `working`), names the live-only inputs (API block, permission prompt, stall, recheck/rejudging, live awaiting snapshot, peer wait) as out of its reach, and gives `cleared` its own bucket, since a completed top later crossed off by hand is neither a leak nor a finish. Step 3 then has four outcomes, not three. Measures 1 and 2 are unaffected; only the premise and the derived column rule need the correction.

- **Measure 1 omits the settled gate, so the ending's own goal reads `working` after a `done`.** `rollup_status(store, session_closed)` takes `focus = top_ancestor(store['lastNode'])` (judge.py 8142), `settled = (nid != focus) or session_closed` (8154), and completes a top only when `is_complete` and (`settled` or `settledDone`) (8169); the docstring at 7920-7932 says the same. Executed: two tops, a closer `done` on g1 through `record_verdict`; with `lastNode` g1 and `session_closed` False, `status[g1]` was `working` and g1 sat in `store['confirming']`; with True it was `completed` and `settledDone` stayed set, so a later False still read `completed`; with `lastNode` g2 it was `completed` regardless. Read: `_ab_close_session` (15811-15834) parses the live session at `now` and passes one `_session_settled` value to both rollups with the live store's `lastNode` untouched; the live closer pass does the same at 15314-15315; `_session_closed` (9361) means the last turn ended; the note names none of `settle`, `closed`, `focus`, `lastNode`, `confirming` or `session_closed` (lines 13, 42, 54). Consequence: an ending in classes 1 to 3 whose turn worked its own goal (the normal case) gets a `done` from the candidate and the harness reads `working` under False, so measure 1 reports zero leaks for the arm while the live card would have entered Completed at that turn's own end, with no new verdict. Fix: name the gate at line 13 and pin the harness's choice, in one of three shapes. (a) Read the rollup's own exports, `status == 'completed'` or the top in `store['confirming']` (8185, 8289), not the raw `nodeComplete` flag: that is the kernel's one-completion-truth precedent (kernel.py 3970-3973, 14670), and the raw flag reads true for subtrees a task still keeps open and for nodes a reopen by hand keeps open, both of which `is_complete` refuses. (b) Pass `session_closed=True` on every ending, and say that this drops the background-task wait the live planner and closer passes apply through `_session_settled` (9549); True mirrors what the live board did in the common case, since a finished turn is closed unless it dispatched a task still awaited. (c) Closest to what the live judges had: `_session_settled` over the ending's own truncated transcript, which the ending definition already carries, instead of the live parse. Under (b) the gate no longer differs between the two builds of one store copy, so measure 3 counts only verdict differences, which is what line 56 defines. Whichever is chosen, line 13's "measurable from the store alone" needs rewording: the column takes a session-state input for the focus top.

## Smaller points

- **The Status line's one cross-reference did not exist at the merge.** Line 3 points at `plans/needs-you.md` and says the note links it rather than restating it; the file is absent at ed23b4e20 and at ccd0e7166 (read, `git cat-file -e` on both) and landed in #1930, merged 24 minutes later. On the current main the link resolves; nothing to change.

- **Tier one keys "finished" on a seven-day window where the journal has the exact events.** Line 33 labels an ending "finished" on a `clear` with no `followup`, `unclear` or `restore` within seven days. Read only. Two endings with identical journals get different labels depending on when the corpus is built relative to the clear, and a `followup` on day eight still leaves "finished"; tier one is what the labeller's 90 percent gate is measured against, so the window moves the gate. Fix: key on the events, "finished" = a hand `clear` with no later `followup`, `unclear` or `restore` row on the node at build time, and require every ending to predate the build by one observation span (or record the span per ending and report labels by span).

## Tests

- The harness tests at lines 71-75 do not yet pin the gate choice above. One case for line 72: the ending's own goal, top-done, reads `completed` under the harness rule, and the second build's flap count excludes the gate.

PR #1931, opened 2026-09-20
Docs: restore the overview, install and guide pages to human size, and keep them there

henrypinkard:
The overview, install and guide pages are back to being pages a person reads, and two guards keep them that way. The shape restored is the one the docs had at the end of July 2026 (commit cf2154bb and the readability commits that follow it through a674ceac), with every feature added since given a paragraph of its own.

**Why.** These three pages are what someone meets before they know anything about romp. They had grown back into reference prose: the install page spent a screen on which interpreter the kernel picks before it gave the install command, and the guide had reached 11,244 words. A person's attention is the scarce thing, and an agent installing romp for them can find any detail elsewhere.

## What moved where

| Page | Words before | Words after |
|---|---|---|
| docs/index.md | 428 | 428 (unchanged; already the July pitch) |
| docs/install.md | 755 | 340 |
| docs/guide.md | 11,244 | 4,001 |
| README.md | 552 | 552 (unchanged) |

Nothing was thrown away. Every dense paragraph is now a named section of `docs/reference.md`, under a new part called "The interface, feature by feature", and the guide links to it where the feature is introduced:

- The chat pane in detail: dropping a file, quoting a passage into the composer, naming a session with `@`, a message that has not gone yet, opening a markdown document, a file's own HTML, text size and width, opening a PDF, links inside a file, the rings on a tab, tags and groups in the tab strip, a tag section at a glance, coming back after a dropped connection, on a small screen, columns and hot keys.
- The feed's layout controls; the Files pane; when a session's tab disappears; a comment thread's mail; the tags a new session inherits; romp's mail tools and Claude Code's own; the VS Code port forwarder; the full Tailscale setup; notifications on a phone or browser; the token file's minting and refusals.
- New sections written for features the guide now introduces in a paragraph: the Artifacts pane (experimental), pane docking, settings across machines, renaming and ending a session from the outline, and the two token usage panels.
- The install page's interpreter material was already in the reference ("The kernel's Python"), as was the macOS `node` copy ("Service environment and credentials"); the install page keeps one sentence and a link, and the one fact the reference lacked (the `uv python install` shim hazard, and `--no-bin` as the remedy) moved into that section.

Every moved claim was checked against the code on `main` before it was kept.

## The navigation change

`PR tiers` was a top-level section of the site. It is contributor process, not usage, and the repo documents it already. The file stays, still answers at `/pr-tiers/`, and `mkdocs.yml` lists it under `not_in_nav`. That mechanism rather than a "Contributing" group: with `navigation.tabs` a group is still a top-level tab, and `not_in_nav` is the documented way to keep `mkdocs build --strict` quiet about a page with no nav entry.

## The two guards

**A rule.** `CLAUDE.md` gains "The documentation front pages are written for a person": short paragraphs (about 70 words), no implementation detail or repo-internal vocabulary, the install command inside the first screen, one paragraph plus a link into `docs/reference.md` for a new feature, and process documents out of the site's navigation.

**A test.** `tests/test_docs_front_pages.py` pins a word budget per page (index 600, install 400, guide 5500, against 428 / 340 / 4001 here), a 70-word paragraph cap with code blocks, tables and raw HTML excluded, the install command inside the first 40 lines of the install page, and `pr-tiers.md` absent from `nav`. Run against the previous pages it is red on all four: install 690 words with its command at line 55, the guide 11,195 words with paragraphs up to 529 words, and the page in nav. Green here.

## Checks

`mkdocs` is not installed on this machine and the Docs workflow runs on pushes to `main` rather than on pull requests, so the strict build has not run on this branch. Instead every internal link and anchor in `docs/` was resolved with the same slug rule MkDocs uses (two hits, both inside code spans, unchanged from before), and every image and video path was confirmed to exist.

Tier: docs


ksarma:
The evidence below was executed against f99d2b327, the PR's head, and every file and line reference is on that commit; the merge a59c0cde3 carries identical content for every file the PR touched (the commits main took in between touch none of them), so the lines read the same there. The hunks do what the body says in outline: the guide's detail moves into the reference's new part, the install command sits at install.md:18 (line 55 in ce67db41e), CLAUDE.md gains the rule, the test pins the caps, and pr-tiers.md is listed under `not_in_nav`.

- **Six sentences of the base guide are in no file under docs/.** The body says nothing was thrown away and fb97d3283's message says every fact was kept, but a search of every markdown file under docs/, plus README.md and SECURITY.md, finds none of these phrasings outside git history (ce67db41e, docs/guide.md): the three mail-mark sentences (a failed send has no mark, the tool call's result saying what happened; a message that waited while the session was offline wears the parked mark; hovering a mark gives the state and when it was reached; base 453-456), the trust clause (a re-linked machine returns with the trust level last given; base 620), the hub-restart sentence (a restart from the hub's interface restarts the machines linked to it, a checked-in machine only itself; base 669-671) and the OpenSSH sentence (each browser socket forwarded one-to-one, closes propagated, so a pane that goes away is gone on both ends; base 701-703). All four facts are still true of the code, by reading: the trust store consulted at attach in kernel/kernel.py, the restart fan-out, the mark states in ui/webview/postal-state.ts. Two bounds: "Two paths work well; one common one does not" survives at guide.md:382-391; and the token exemption is only partly lost, since guide.md:453-454 still says liveness probes and the Home Screen files skip the token and SECURITY.md:49-55 and docs/read-side.md:806-812 carry the route list, but the section guide.md:455 links to, "The token file: minting, permissions and refusals" (reference.md:5246), names neither the routes nor the reason. The mark paragraph also dropped two details, the mark's place at the card head's right edge and the head naming each end in its session's color (guide.md:183-186 keeps the naming of both ends and the list of states, without the edge or the color). The new CLAUDE.md section states a rule and claims nothing about completeness, so it needs no change. Fix, as a docs-tier follow-up: the mark sentences and the two details as a subsection under "The interface, feature by feature" rather than under "Mail from the terminal" (reference.md:190), the shell surface; the trust clause and the hub-restart sentence back in the guide, which has about 1,500 words of headroom under its cap, or in architecture.md, since the reference has no linking section and guide.md:360-361 already sends the mechanics there; the OpenSSH sentence as the opening paragraph of "The VS Code port forwarder and the browser dashboard" (reference.md:5077), in reference voice with no relative pointers; and the route list (`/healthz`, `/version`, `/busy`, the bus's `/ping`, `/manifest.webmanifest` and three icons under `/media/`) with its reason (fixed files that read no session state) under the token-file section.

- **The moved admonition points at paths that are no longer above it.** reference.md:5095 still says "prefer one of the two paths above"; in the base guide the two paths, plain ssh forwarding and Tailscale, were the paragraphs directly above the same admonition (ce67db41e, guide 695-708), while in the reference the headings above it are "The tags a new session inherits" (5057) and "romp's mail tools, and Claude Code's own" (5073), "ssh -N", "ssh -L" and "port forward" appear nowhere else in reference.md, and Tailscale is first named further down, at 5104. Fix: name the paths in place, "plain ssh port forwarding (the guide's From another machine section) or Tailscale (the next section)", linking `guide.md#from-another-machine` (header at guide.md:380) and the reference's next section, `#reaching-romp-from-a-phone-the-full-tailscale-setup`; the OpenSSH sentence restored as this section's lead, as above, also gives the admonition's argument (a multiplexed channel that never closes the far end) something to contrast with. Neither edit touches a pin: no test under tests/ or ui/webview/ matches "two paths" or "port forwarder" against reference.md, and reference.md is outside the front-page caps.

## Smaller points

- **"A blue outline shows where the pane will land" (reference.md:4920)** describes `#pd-outline`, a ring in the accent variable with a blue fallback (ui/webview/panedock-main.ts:125-128), and that variable is `#9cd2ff` in the default theme but `#C2410C`, a clay orange, under the light theme (ui/webview/styles.css:115 and 270), so the ring a light-theme reader sees is orange; write "An outline in the accent color shows where the pane will land."

- **The stated reason for `not_in_nav` does not match MkDocs:** the body, 815def668's message and the test's two assertion messages (tests/test_docs_front_pages.py:133-134) say the entry keeps the strict build quiet about a page with no nav entry, but by MkDocs's configuration documentation (not a build with the entry removed) the omitted-pages notice is at the info level by default, under `validation.nav.omitted_files`, and strict mode halts only on warnings, so a page missing from nav would not have failed the strict build either way (CLAUDE.md:267-269 names the mechanism and says nothing about the strict build, so it does not carry the claim); reword the two messages to say the entry declares the page intentionally unlisted, or, if the strict build is meant to refuse an unlisted page, set `omitted_files: warn` under the nav validation in mkdocs.yml so the premise becomes true.

## Tests

- `test_pr_tiers_is_not_a_navigation_entry` (tests/test_docs_front_pages.py:125-134) splits mkdocs.yml at the first nav key and, at line 134, asserts only that the string "not_in_nav" appears somewhere before it, so a block listing any other page, or the word in a comment, passes while the message says pr-tiers.md needs the entry; parse the block and assert pr-tiers.md is one of its lines, for example `re.search(r'^not_in_nav: \|\n((?:  .*\n)+)', mkdocs, re.M)` and a check on the group's lines.

- Two defects in the paragraph cap: its failure text says `docs/<page> line N` where N is the enumerate index at line 79 over output that has already dropped fenced code and HTML comments (at f99d2b327 the filter drops 18 lines of install.md, so a failure near its end would name a line 18 above the paragraph), and `_NOT_PROSE` (line 40), applied at lines 84-87, skips any line whose stripped form starts with "<", which also drops a wrapped prose line that begins with an inline tag and splits its paragraph into two units, so the 70-word cap undercounts (guide.md has four such lines, 69, 162, 222 and 265; the seven video tags and four comments across the three pages are dropped correctly); yield (file line, text) pairs or replace dropped lines with empty strings, and skip a "<" line only when it is block HTML (video, img, div, a comment, or a lone tag matching `^<[a-z]+[^>]*>\s*$`) or strip tags and count the remaining words.

- The counts the body, 815def668's message and the test's own comment (lines 32-33) give for the restored pages, 428 / 340 / 4001, are wc -w figures, and not quite at the head either (wc -w at f99d2b327 gives 444 / 340 / 3997; index.md's 428 is the base count, before the one HTML comment the PR added), while the committed `_words` measures the pages at f99d2b327 as 394 / 259 / 3932; of the red-before figures, install at 690 words with its command at line 55 and the guide at 11,195 words match the test's measure against ce67db41e, but the largest base paragraph under the committed `_units` is 921 words (the one opening "Tags and groups"), not 529; cite the test's own measure in the comment, or say plainly that the quoted figures are wc -w counts.

PR #1932, opened 2026-09-20
The judge prompt experiment's harness: a corpus of endings, arms over copies with the prompts swapped in, the measures and the report

henrypinkard:
The harness for the judge prompt experiment (plans/judge-prompt-experiments.md, #1929): the corpus builder, the arms over copies with the prompts swapped in, the four measures and the report. A script and its tests; no kernel code changes. It costs nothing until someone hands it the real binary, and no paid pass runs before the user's word.

**What it is** (`scripts/judge_experiment.py`).
- `build-corpus` reads the live state root and the Claude root as files (names, transcripts, goal stores, override journals; no romp module is loaded against them) and writes, under a destination it refuses inside any git checkout, one truncated transcript per ending under its own id, a names entry, the goal store as the judges held it before the cut (nodes born before the cut, each verdict log cut to events before it, the flags left for the arm's rollup; a session with no store starts fresh), the override journal cut the same way, and a manifest of ids, classes and cut times with the session ids hashed and no text. Endings are the turn ends of each transcript, classified by the design's heuristic (an offer of a next step, an open question, an undone item, a finished thread), balanced per class and spread across sessions.
- `run` runs each arm in a subprocess of its own: the corpus state copied under the run root, the event model and the judge module loaded against it (the module binds its roots from the environment at import, and it is loaded under a name of the run's own so the tests' shared-state guard on `romp_judge` stays quiet), the arm's prompts swapped into `PLAN_SYS`, `CLOSER_SYS` and `UNBLOCK_SYS` by attribute and restored after, and per ending and per build the planner (`_plan_session`), the closer over the last closed turn (`_close_turn`, then the rollup), and the unblocker (`_unblock_session`), the tops' columns recorded by the feed's rule (blocked to `needs_input`, completed to `completed`, else `working`). Two builds from the same store copy give the flaps. The arm's own usage ledger is read after every ending; past a fifth over the budget the run stops and says where.
- `measure` and `report`: leaks into Completed (an offer, question or undone ending with a top read completed), false interrupts (a finished ending with a top read needs_input), flaps, and cost, calls and mean call time from the ledger; a markdown table and a measures file beside the arms, and the cleanplots figure when the library is present (a note when it is not). Counts and dollars only.

**Tests** (tests/test_judge_experiment.py, hermetic: synthetic roots as the judge tests build them, a fake `claude -p` that answers by the judge named in the system prompt and by a marker the candidate arm's prompt carries, a fixed cost per call; every string invented). The builder classifies four scripted endings, reads the live roots without writing them, writes only under its destination, hashes the session ids and keeps every transcript text out of the manifest, cuts an override journaled after the ending, and refuses a destination inside a git checkout without writing there. The current arm files the offer, the question and the undone item as done (three leaks, no false interrupt, no flap) and the candidate arm blocks all three and leaves the finished thread completed, the corpus unchanged and the cost one fixed row per call. The prompt swap reaches every closer call and the module attribute is restored (an unknown key refused). A budget below one call's worth stops the run past a fifth over. The report's table carries the counts and no goal title or transcript text, from the function and from the command line.

Red first at 387a40ee (main) by copy-aside: five reds, all on the one assertion `False is not true : the harness script exists (the base has none)`.

**Not here.** The corpus of the user's history, the labeller pass and every paid arm: those run on the manager's dispatch after the user's word on the budget, and their results land in the design note, never here.

**Suites at the head** (23d0eb4e): the harness module and its neighbours (the judge child's and the judges' process tests) 54 passed. No file under kernel/, ui/ or vscode-extension/ changes.


lczh:
Evidence executed against 23d0eb4e, the merged head.

The harness has the right shape, a corpus of endings, arms over copies with the prompts swapped in, four measures and one report, and its module passes at the head. What keeps it from measuring what it says appears on any corpus built from a live session with a goal store: the seeded store's placements never match the arm's segment ids, so the planner re-plans the whole history on every build; the cut keeps the live judges' own node and verdict from the ending's turn; the leak and false-interrupt measures count columns that were there before the cut; and a prompt whose replies the parser rejects scores a perfect row. The fixtures' empty stores hide all four, since `store_before` never runs on them.

## At this head

**The seeded store's placements never match the arm's segment ids, so every ended segment of the history is a fresh planner unit on every build.** `store_before` (scripts/judge_experiment.py:117-138) copies every key but `nodes` and `status` at 135, so the seed keeps `rompUuid` as the live session id, `placementsV` at 14 and every placement key as `<sid>:<t>:<hash>`. The arm's `jd._plan_session(eid, ...)` (307) parses under the ending id: `parsed_session` (kernel/judge.py:11766) hands `rompuuid=fsid` to `em.parse_session` at 3921, and segment ids are built as `"%s:%d:%s" % (rompuuid, seg_t, h)` (kernel/event_model.py:4311), so the arm's ids read `<eid>:<t>:<hash>`. `_placed_key` (kernel/judge.py:11190) matches exactly at 11210 or through `_seg_key`, which keeps the prefix (11161); neither matches, and `_migrate_placements` (11164) runs only when `placementsV` differs (11790, 11180-11181). So `plan_llm` (12182) runs once per ended segment past the checks at 11802 and 11849. Executed: a three-turn synthetic session planned and closed by the judge module under its own id, then `build_corpus`, then `_plan_session` per ending against a copy of the corpus state: 0 seed keys matched, `_migrate_placements` never fired, and the planner ran 1, 2 and 3 times for the three endings in turn order; the harness's two-build arm made 12 planner calls over that corpus. The tests cannot see it: every fixture store is `{"nodes": {}, "status": {}}` (tests/test_judge_experiment.py:113), so `before` is None at scripts/judge_experiment.py:207 and `load_goals` mints a fresh store keyed by the ending id (`_fresh_store`, kernel/judge.py:4883). Fix in `store_before`: set `rompUuid` to the ending id, rewrite the prefix of every `placements` key to it, and drop the placements whose `t` is at or after the ending turn's start. Re-keying alone gave 0 calls, since the seed also carries the ending turn's own placement; with those keys dropped too, exactly 1. Test: a seed with a placement for the earlier turn, asserting one planner call per ending per build.

**The cut keeps the node the live planner minted from the ending's own turn and the live closer's verdict on it, so the arm's first rollup already reads the ending completed.** `build_corpus` takes `cut_t` from the ending assistant record's timestamp (169), the turn's end; `store_before` keeps a node whose `t` is at or before it (123-125) and filters diary rows by `float(e.get("t") or 0) <= cut_t` (122). But `apply_plan` stamps a minted node `"t": seg_t` (kernel/judge.py:7694), the segment's start; `_close_turn` rules with `t=turn.get("t")` (15020) and `apply_close` files the done at that time (14723, 14747); and `record_verdict` writes the row as `{"ev_t": ev_t, ...}` (10532) with `"at": int(time.time())` (10546) and no `t` key. So the ending's own node passes, and since no diary row carries `t`, every row passes, later turns included. Executed on the store above: the turn-2 finished ending's seed kept `g2`, born at that turn's start, with the closer's done row at the same `ev_t`; the turn-0 seed kept a done row filed 600 seconds after its cut. `rollup_status` over each seed alone, no model call, read `g1` completed for turns 0 and 1 and `g1` and `g2` completed for turn 2. End to end, a candidate arm whose fake blocks every offer and question was charged 2 leaks, `g1` completed on both builds of those endings, against 0 for the same arm over the same transcript with no store. Fix: cut on the transcript's own event, the timestamp of the user record that opened the ending's turn, kept in the manifest; keep nodes with `t` strictly before it and rows with `ev_t` strictly before it. Test: a judge-shaped seed with a node and a done row from the ending's turn plus an earlier node; assert the before-store carries only the earlier node, the next ending's both, and the arm's column is the fake's.

**Leaks and false interrupts count columns that were already there before the cut, and `topsBefore` is written and never read.** `measure` (350-369) counts a leak when any top reads `completed` on an offer, question or undone ending (359) and a false interrupt when any top reads `needs_input` on a finished ending (361). `store_before` keeps each pre-cut node's diary (127) and pops only the flag cache (128-129); the arm's `rollup_status` (kernel/judge.py:7920) rebuilds the flags from the diary at 7939, a top with a pre-cut done row settles at 8154 once the session is closed, and a top with a pre-cut block row reads blocked. Every arm inherits those columns identically, and `topsBefore` (217) lists ids that nothing reads. Executed: a judge-written store with `g1` done and settled and `g2` blocked before the cut, an empty-store control over the same transcript, and a fake that answers `skip` and empty verdict lists. The arm filed nothing, 5 diary rows before and after on every ending, and `measure` reported 1 leak, the offer ending with `g1` completed, and 2 false interrupts, both finished endings with `g2` needs_input; the control gave none. So every arm saturates alike and the zero-leak gate is out of reach. Fix: roll the seed up once at build time, record each pre-cut top's column in the manifest, and count only a top whose column changed to completed or needs_input, or a top the arm minted. Test: a seed with a pre-cut completed top and a pre-cut blocked top, the fake filing nothing, zero of each.

**A failed judge call or a reply the parser rejects leaves no trace in the results or the report, so a prompt that breaks the parser wins the table.** `run_arm_inprocess` discards the returns of `_plan_session` (307), `_close_turn` (313) and `_unblock_session` (316) and never reads `jd.ERRORS` (kernel/judge.py:132). `_close_turn` returns None when the call failed (14961) or the reply did not parse (14981), after `_log_judge_error("closer", ..., "parse", ...)` at 14965. Neither `results.json`, `measure` (367-369) nor the table (380-385) carries a failure field, and cost and calls hide it too: a parse-rejected reply still writes a usage row (2459); only an error envelope skips it (2456). Executed on the PR's own fixtures with a fake that answers prose to any candidate-marked prompt: with only `CLOSER_SYS` swapped, all 4 closer calls were rejected, 4 `closer/parse` rows in the arm's error ledger, and the row read `| prose | 4 | 1 | 0 | 0 | 0.16 | 16 | 7 | no |` beside `| current | 4 | 3 | 0 | 0 | 0.16 | 16 | 7 | no |`, the one leak coming from the un-swapped planner. With all three prompts swapped, 12 planner calls were rejected, no node was minted, and the row read `| prose-all | 4 | 0 | 0 | 0 | 0.12 | 12 | 7 | no |`, the perfect row. The test's fake always answers JSON the parsers accept (tests/test_judge_experiment.py:51-64). Fix: count None returns from `_close_turn` and rows appended to `jd.ERRORS` during the arm, write `failures` into results.json, `measure` and the table, and mark the row not comparable, or stop the arm, when nonzero. A scratch copy with exactly that ended the rows `| no | 0 |`, `| no | 4, not comparable |` and `| no | 12, not comparable |`. Test: the prose fake for the candidate arm, asserting the count and the mark.

## Smaller points

- Post-cut gate stamps survive `store_before`: it copies each node with `dict(nd)` (126), and `blockCheckT`, `blockCheckDoneT` and `closerLookT` are not in `PROTECTED` (kernel/judge.py:4609-4617), so `_materialize_from_log` never rewrites them. `_unblock_session`'s due gate (15718-15720) reads `newest > max(bt, blockCheckT)` with `newest` at most the cut, so that half never fires for such a node. Executed: a pre-cut blocked top stamped after the cut drew 0 unblocker calls, 1 with the stamps dropped. The done-time half still re-arms it when the arm's closer files a done, so the effect is confined to builds that file none. Clamp or drop every time-valued stamp above the cut in `store_before`, with a test that a post-cut `blockCheckT` no longer hides a node from the unblocker.
- The planner's first menu is built from the stripped flag cache. `store_before` pops `nodeComplete`, `blocked` and `cleared` (128-129); `load_goals` re-derives flags only when the override replay wrote (`_finish_load`, kernel/judge.py:4911-4915); `_plan_session` builds `open_menu` (11891) before its one rollup (12288); and `_sealed` (6820-6830) reads the popped flags plus `settledDone`, which stays. So pre-cut done sub-goals and user-cleared tops are open items to the planner, for every session alike, since a journaled clear already in the diary writes nothing on replay. Executed: `open_menu` listed `g2, g3, g4, g5` before any rollup, a done sub and a cleared top among them, and `g2, g3` after; the fake's planner saw the done sub as item 3. Keep the flag cache in the seed, or run `rollup_status` and `save_goals` on it before the first judge of each build, and assert the fake's `<open-goals>` text omits a pre-cut done sub.
- The manifest carries the raw session id inside `topsBefore`, and the test cannot notice. Node ids are minted as `"%s:g%d" % (store["rompUuid"], store["seq"])` (kernel/judge.py:7690) with `rompUuid` the live session id (4883), and 217 writes them verbatim beside the hashed `session` field; executed, a store with tops put `<sid>:g1` through `<sid>:g5` into the manifest. The test's `assertNotIn(sid, manifest)` (tests/test_judge_experiment.py:142) passes only because every fixture store has no nodes (113), so `topsBefore` is `[]`. Write `topsBefore` as hashed ids or `g<n>` suffixes, and give one fixture store a top so the assertion can fail.

lczh:
The test runs behind the comment above, on 23d0eb4e (merge base with main 55d8e8f0).

## Tests

- `python3 -m pytest -q -p no:cacheprovider tests/test_judge_experiment.py` at the head: 5 passed in 3.57 s. No other Python module was run here; the PR changes nothing under kernel/, ui/ or vscode-extension/, so the extension suite and the served labs were not run.

- The seed re-keyed to the ending id, the first finding's control: `rompUuid` set to the ending id and the prefix of every `placements` key rewritten from the session id to it, then `_plan_session` per ending against a copy of the corpus state. Seed keys matched by `_placed_key`: 1, 2, 3 for the turn-0, turn-1 and turn-2 endings; planner calls 0, 0, 0, since the seed also carries the ending turn's own placement. Re-keyed and with every placement whose `t` is at or after the ending turn's start dropped: matched 0, 1, 2; planner calls 1, 1, 1. The unmutated seed, as the control: matched 0, 0, 0; planner calls 1, 2, 3; placement keys after the run 4, 5, 6 (the seed's three plus the arm's); nodes after 2, 2, 4.
- scripts/judge_experiment.py with a failure counter: `results["failures"]` as the line count of `jd.ERRORS` after the run minus before, `results["closerNone"]` counting None returns of `_close_turn`, `failures` carried into `measure`, and a tenth table column reading `<n>, not comparable` when nonzero. The copy lived in a scratch tree with `bin/` and `tests/` linked beside it, since the script resolves both from its own location, and its `run_arm` subprocess loads the copy. Over the PR's fixtures and the prose fake below: `| current | 4 | 3 | 0 | 0 | 0.16 | 16 | 7 | no | 0 |`, `| prose | 4 | 1 | 0 | 0 | 0.16 | 16 | 7 | no | 4, not comparable |`, `| prose-all | 4 | 0 | 0 | 0 | 0.12 | 12 | 7 | no | 12, not comparable |`. The unmutated script over the same fixtures and fakes gave the head's nine-column rows with the same counts, as the control.
- The two unblocker stamps dropped from the seeded blocked top, the smaller point's control: 1 unblocker call against 0 with the stamps as seeded; the top read `needs_input` both ways, since the fake lifts nothing.

## Probes

- The judge-written store: a three-turn synthetic session, an offer, a question and a finished ending, prompts at t, t+600 and t+1200 with the assistant record 30 s after each, planned by `_plan_session` and closed by `_close_turn` over every ended turn under its own session id, with the test module's fake as the model: 3 planner calls, 1 closer call. The store: `rompUuid` the session id, `placementsV` 14, three placements `<sid>:<t>:<hash>` whose `t` values are the three turn starts; `g1` with `t` the first turn's start and a planner done row at `ev_t` = the second turn's start; `g2` with `t` the third turn's start and a closer done row at that `ev_t`; every diary row carries `ev_t`, `src`, `kind` and `at` and no `t` key.
- `build_corpus` over it: three endings, cuts 30 s after each turn start. Every seed keeps `rompUuid` = the session id, `placementsV` 14 and all three placement keys. Turn-0 seed: `g1`, born in the ending's own turn, with its done row 600 s after the cut and its settle row 1200 s after. Turn-1 seed: `g1` with its done row at the ending turn's own start. Turn-2 seed: `g1` and `g2`, `g2` born in the ending's own turn with the closer's done row at that turn's start. `rollup_status` over each seed alone, the judge module loaded against a scratch root with a binary that cannot answer: turn 0 `g1` completed; turn 1 `g1` completed; turn 2 `g1` and `g2` completed. The manifest text contains the raw session id, in `topsBefore`.
- `_plan_session` per ending against a copy of the corpus state through the harness's own `load_judge`: every segment id of the arm's parse prefixed by the ending id; the rest of the figures are in the first mutation bullet.
- The harness's `run_arm_inprocess`, two builds, over that corpus: the baseline fake made 12 planner and 4 closer calls, columns offer `g1, g3` completed, question `g1, g3` completed, finished `g1, g2, g3, g4` completed, measure leaks 2; the candidate fake, which blocks every offer and question, made 12 planner, 4 closer and 4 unblocker calls, columns offer `g1` completed and `g3` needs_input, question the same, finished `g1, g2, g3` completed, measure leaks 2, false interrupts 0, flaps 0. The same candidate arm over a corpus built from the same transcript with the goal store removed: offer and question `g1` needs_input, finished `g1` completed, leaks 0.
- The measure probe's store, written by the judge module's own `record_verdict`, `rollup_status` and `save_goals` under its session id: `g1` done by the closer at the first turn and settled, `g2` blocked by the closer at the first turn with `blockCheckT` 5000 s after the first prompt and `blockCheckDoneT` equal to the newest done row's `at`, `g3` open, `g4` a done sub of `g3`, `g5` cleared by the user with a matching journal row; a second session with the same transcript and the fixtures' `{"nodes": {}, "status": {}}` store. The transcript: a finished, an offer and a finished ending. `build_corpus`: `topsBefore` for each of the first session's three endings `['<sid>:g1', '<sid>:g2', '<sid>:g3', '<sid>:g5']`, the raw id present in manifest.json, every `session` field a 12-character hash, the control's `topsBefore` empty. The offer ending's seed: `nodeComplete`, `blocked` and `cleared` absent on every node, `settledDone` kept on `g1`, every diary row kept, `g2`'s two stamps kept above the cut.
- The planner's menu, in process against a copy of that corpus: `_replay_overrides` returned False with the journal present; `open_menu` before any rollup `[g2, g3, g4, g5]`; after `rollup_status` `[g2, g3]`.
- The unblocker on that seed, the arm's own sequence of load, rollup, save and `_unblock_session`: 0 unblocker calls with the stamps as seeded, `g2` needs_input.
- The do-nothing arm over the six endings, two builds, its fake answering `{"ops": [{"do": "skip"}]}`, `{"done": [], "block": []}` and `{"verdicts": []}`: 32 planner, 12 closer and 0 unblocker calls; diary rows 5 in the seed and 5 after on each of the first session's endings; columns `g1` completed, `g2` needs_input, `g3` working, `g5` working on every ending and build of that session; the control's one top, minted although the planner answered skip, working on every ending (that mint path was not examined). Measure: leaks 1, false interrupts 2, flaps 0. Among the menus the fake's planner received: `1. Wire the export button`, `2. Tidy the export button`, `3. Rename the label`, the done sub numbered as open.
- The failure probe on the PR's fixtures, four endings over two sessions with empty stores, arms current, prose and prose-all, where prose swaps `CLOSER_SYS` alone, prose-all swaps all three prompts, and the fake answers a sentence of prose to any candidate-marked prompt: current 16 usage rows, no error rows, every top completed, leaks 3; prose 16 usage rows, error rows `closer/parse` 4, 4 closer calls, tops offer working, question completed, undone working, finished completed, leaks 1; prose-all 12 usage rows, error rows `planner/parse` 12, 0 closer calls, no tops, leaks 0. `results.json` keys in every arm: `arm`, `callMsMean`, `calls`, `cost`, `endings`, `prompts`, `stopped`; `measure` keys: `arm`, `callMsMean`, `calls`, `costUsd`, `endings`, `falseInterrupts`, `flaps`, `leaks`, `stopped`.
- `git grep -n topsBefore` at the head: one hit, scripts/judge_experiment.py:217, the write.

ksarma:
The harness was read at the merge commit d560913d1, whose copies of the two files are byte-identical to the PR head 23d0eb4e6; every probe named here ran there, and file and line references are on 23d0eb4e6, except the plan document (plans/judge-prompt-experiments.md), which that commit predates and which is cited at d560913d1.

`store_before` (lines 117 to 138).

- Line 122 cuts each node's verdict log on the key `t` and line 123 reads it for the birth, but no verdict-log writer stamps `t` (the earlier comment's second finding), so a candidate that files no verdict on a top the live session completed later is charged the live outcome. Read `e.get("ev_t") or e.get("at") or 0` at both lines, the birth fallback over the unfiltered log, and cut strictly before the ending turn's start, the earlier comment's cut (verdicts carry that start as `ev_t`, so a cut at the assistant record's time keeps the turn's own live verdicts), or write the boundary into the docstring, which says "at or before" where the body says "before". Loop the orphan pass (lines 131 to 134) to a fixed point (a relink can nest a child under a newer parent).

- Line 135 copies every other key whole, so the seed keeps the live session id where the arm parses under the ending id (the earlier comment's first finding). Beyond its re-key: rewrite the journal lines too; drop the ending turn's bare segment key (a cut-time rule keeps it) but keep its `#p` key and the node it points to; drop nodes whose first trail entry is a dropped segment other than that target, before the orphan pass, and compute `topsBefore` after the drop. That alone still mints a twin top, because line 128 leaves `settledDone`, `settledAt`, `doneWhy` and `rolledUp`, which seal the seeded top out of the planner's menu; only the re-key, the drops, that pop and the `ev_t` cut together gave one call and one top per ending. Test at turn index 1 or later over a store the live planner wrote.

- Line 217 writes `topsBefore` as sid-prefixed node ids (the earlier comment's third smaller point), for nearly every ending in a real root. Record a suffix, count or rewritten id; arm the pin (test line 142) in the corpus-builder test only, with a judge-shaped store (`rompUuid`, `seq`, `placements`, `placementsV`, or the arm raises `KeyError`) and a top at or before the earliest cut (seeding both sessions changes the arm test's tuple).

`measure` (lines 350 to 369).

- Lines 359 and 361 score a leak or a false interrupt on any top reading `completed` or `needs_input` (the earlier comment's third finding, over a seeded store); without a store the arm re-plans the earlier turns itself, so a top legitimately completed earlier counts against every later ending (a fresh three-turn session: leaks 1 for both arms). Excluding `topsBefore` misses a pre-existing open top the ending's own turn wrongly completes, which then scores 0. Score only tops born, or given a done or block verdict (`ev_t`), between the previous ending's cut and this one; pin it with a fixture session that has an earlier completed goal.

The closer (line 313).

- `_close_turn(store, closed_turns[-1])` passes no `seg_by_id`, so the arm's closer judges the touched goals' titles without the goal-history section production sends whenever a touched goal has a trail or placement. Build the map as `_close_session` does, after the planner's store is loaded, and pass it; assert its presence on every closer row through the fake. The plan's precedent (line 42, `_ab_close_session`) omits it too.

Turn ends (line 112).

- `turn_ends` ends a turn only when the next `user` record's content is a string, while the event model opens a turn on any non-meta `user` record with text and the SDK channel writes composer prompts as text-block lists, so in SDK-driven sessions those prompts never end a turn for the builder (1 of the judge's 207 ends on one such transcript). A paraphrase of the opener rule keeps, and adds, false ends at system wrappers the judge treats as part of the turn: reuse the event model's own `author_of` and opener rule (`isMeta` still excluded), loaded with the environment at the destination so no romp module reads the live roots, or copy the rule.

The defaults (lines 425 to 428).

- `run` and `run-arm` default the binary flag to the string `claude` and the budget to None, so a bare `run-arm` against a `PATH` shim named claude ran 16 calls with no stop: the body's "costs nothing until someone hands it the real binary" is false as read. Make `--claude-bin` required on both subparsers. Since nothing in the code can tell a real binary from a fake, make `--budget-usd` required on `run`, with `inf` for unbounded, and leave `run-arm`'s default or always forward it (four tests pass None). Test that `run` without either flag exits non-zero, and reword the body and docstring.

The shared event model (line 233, test line 188).

- `load_judge` re-executes the event model under its shared name (line 233 and kernel/judge.py line 96) under the arm's scratch roots, and the in-process test's finally restores only the environment, so afterwards the shared module's roots sit under removed directories and its checkpoint-dir lambda (judge.py line 97) is the arm judge's: `_ckpt_dir()` readers treat the missing directory as checkpoints off for the rest of the process (no current test goes red), and the conftest guard, watching the judge's roots only, stays quiet. Fix: after the environment restore, `load_source("romp_judge", BIN / "romp-judge")`, which resets roots and provider together (reloading the event model alone leaves the provider at the module default); or, at the cause, swap the shared event model out of `sys.modules` in the loader around the two loads. Pin the shared `STATE` and the provider's root (a `PROJECTS` pin is false at the floor).

## Smaller points

- Line 161 reads only the sid's own transcript, so the leaf transcripts a /clear or a resume creates never become candidates and a session whose anchor transcript is gone drops with no count (zero affected on one live root): enumerate as files, mirroring `known_fsids` in kernel/sdk_backend.py plus the same-customTitle head scan (not the judge's discovery, which the docstring rules out), and record skips in the manifest with no sid.

- Line 7 documents `report --run-root DIR [--figure PNG]` while line 429 requires `--corpus`, so the documented invocation exits 2: add the flag to the line, or make it optional, since `measure` already falls back to the class each results.json ending carries.

- Line 200 copies the goal store but not Claude's task store, so `_sync_declared_plan` reconstructs the to-do state from the transcript where production read the store when the source machine has one: copy the store whole only for endings at a session's last turn (an mtime cut drops an open item or keeps a later status) and record per ending which the arm read, or say in the body and plan that arms run without it.

## Tests

- The false-interrupt and flap counters (lines 362 and 366) are asserted only at zero, so removing both increments stays green (pin them on a hand-built results dict), and `all()` at test lines 175 and 177 is vacuously true when the ending minted no goal (assert `set(tops.values())` instead).

- Test line 136 compares sorted class multisets, so swapping the next-step and question labels in `classify_ending` stays green (assert each ending's class and the manifest's length); test line 151 pins the journal cut on one side (add a pre-cut row and assert it is the turn-0 journal); and the run-root refusal (script line 278) is pinned for `build_corpus` only (assert `SystemExit`, the subprocess entry's `CalledProcessError`, and an empty `rglob`).

- The turn-end skip of tool rows, the per-class cap (script lines 178 to 181) and the per-session spread run on a fixture too small to engage them (tool rows inside one turn and two sessions with more candidates than `per_class` turn three mutants red); test line 207 does not pin the fifth-over rule (an ending costs 0.04, so against 0.005 any rule stops after the first; 0.036 fails only the mutant `cost > budget_usd`); and the figure road has no test, and where `cleanplots` is absent an either/or assertion passes through the note, so pin the note and the PNG (a stub in `sys.modules`) separately.

PR #1933, opened 2026-09-20
A Codex session compacts natively: every compaction door runs Codex's own compaction, bracketed by the thread's status, and the divider is written when it ends

lczh:
Tier: feature

Follows the Codex slash-command guard (#1864), the known-heads refusal (#1891) and the native clear (#1880), all on main: the guard put a route in front of the setter body with a handler table between the slash predicate and the refusal, the refusal narrowed to the heads the kernel knows a Codex session cannot take, and the native clear registered the first heads in the table. This change registers the third, `/compact`, and turns every compaction door on a Codex session into Codex's own compaction. Filed as a feature: it adds a capability the doc listed as coming, inside the existing model of the Codex backend and the session contract.

## What happens today

On a Codex session every compaction door delivered the literal text "/compact" to the model as a prompt on the same thread and then showed a false "compacting" cue. The doors: the chat and timeline batteries, POST /compact and `romp compact`, which all land in the one compaction entry, sent the word through the backend's send and stamped the optimistic cue; a typed "/compact" fell past the setter route's one-token guard, parked as a command op and fired alone as text at the turn's end; and a command op still in the pending-ops disk mirror from before an upgrade replayed the same way. The Codex app-server has no slash parser and the backend's send reads no text, so the model answered the word; the backend defined neither compact nor compacting, so the kernel fell to its optimistic derivation, which cleared when the reply turn opened. The guard then made the same presses honest refusals, with the battery marked inert and its click declined on both surfaces, still not a compaction.

What the live probe of the pinned runtime showed, and this change is built on: the SDK's compaction request is acked in about a millisecond, and the ack means queued, nothing more. The compaction then runs as its own server turn (about 4 s on a two-turn thread) whose one item is a contextCompaction; the notification the normalizer's boundary writer waited for never arrives; the pinned client's router parks that unregistered turn's notifications and discards them at its completion; a turn/start during it is refused with "ActiveTurnNotSteerable { turn_kind: Compact }", which the existing rejection rule classes as retryable, so the worker would have backed off and retried into the compaction for its whole run; and that refusal landed 4 ms before the compaction's own active status, so a bracket keyed on the active status alone would have missed it. The only evidence that reaches the backend is the thread's status on the global queue: active, then idle.

## The change

A `compact` verb on the session contract, a concrete default in `move`'s and `clear`'s idiom: an empty string on success, `busy` when a turn is in flight (the kernel parks and retries at the turn's end), any other string the reason, shown verbatim; worded for no backend in particular and without the word "backend". The Claude Code backend does not implement it: a typed or clicked compaction there still goes to the CLI as text, which executes it, and every SDK path is byte-identical (the two literals another module pins in the kernel source stand where they were).

The Codex backend implements it around what the probe showed. `compact` takes the worker's own loading precondition first: on a fresh app-server the thread is prepared (thread/resume) under the turn lock as the worker holds it, a raise there answered in words and never parked, since a compaction has no retry queue; a pending or failed placeholder is refused in words rather than minted; a conversation with no records (a fresh spawn, a fresh clear) is refused before any request, since a compaction the server acks and never takes up would leave the bracket standing. The normalizer is built before the latch and outside its lock, so the boundary writer never meets an unbuilt normalizer on a session restored in this process, and the boundary chains to the true leaf the file holds. The bracket is latched before the request, so an active status that lands during the round trip is seen while it stands, and published through `compacting()` (which the kernel already reads first, so the chip, the chat's compacting element, the drive-op gates, the drain and the sessions rows follow with no kernel literal change) and `busy()` (a compaction is a turn in flight on the server, which refuses turn/start for its whole run).

The pump ends the bracket on the first idle status after an active one. An idle with no active seen is the previous turn's tail, drained late from the global queue, and is ignored. On the clean end the compact_boundary is written through the normalizer's existing thread/compacted writer, marked manual (the golden's emitted notification keeps "auto"), and it lands before the end is published, so a reader that sees the session no longer compacting also finds the divider. A systemError status is the failed compaction (see the live legs below): it ends the bracket loudly with no divider, as a launch error the next accepted turn clears, worded as Codex not having compacted the conversation; a notLoaded status ends it the same way, worded as the thread having stopped being available. The client's death ends every standing bracket loudly; a kill ends it silently, since the compaction finishes server-side and romp cannot stop it. A thread/compacted for the thread while the bracket stands ends the bracket and writes the record once, marked manual, an arm reachable only with a future runtime AND a future client that deliver that notification globally: runtime 0.153.3 never sends it, and the pinned client's router parks a turn-stamped one under its unregistered turn and discards it at that turn's completion, so under the pin the idle path is the one that runs.

The worker reads the server's Compact refusal as a state, not a failure: no launch error, no backoff, the queue prefix untouched (no acknowledgment happened), the bracket latched with the active already seen. It re-latches only while two per-session counters, read beside the request's thread id, are unchanged: a bracket-end counter every end of the bracket advances, and a count of idle statuses seen for the thread, bumped for every idle whether or not a bracket stands. Either advancing means the compaction the server named has been seen to end (its statuses traveled the other queue first), so the refusal retries at once instead of standing behind a bracket nothing would end. The idle count is the review's fold: a compaction romp did not start has no bracket to end, so its idle advanced the end counter by nothing, and a refusal handled after that idle would have re-latched a compaction already over, with no status left to end it (a previous turn's late idle now costs one extra refused retry, never a wedge). The handler also latches nothing on a dead row (a kill between the refused request and its handler found no bracket to end), and resume() resets the bracket fields with the rest, so a revived session never starts out compacting. Under a bracket that has seen its active, the worker makes no attempt and waits, untimed, for the end's kick. Under a bracket that has not, it attempts: the server is the authority from the ack on, and the answer is exact either way, a refusal latches, an accepted turn ends the bracket with a log line and no divider (a compaction the server never took up, or one that finished before the request).

In the kernel, the compaction entry prefers the verb on a backend that defines it, after the same one drive-op gate every door pays: an empty answer fires with no optimistic stamp (the stamp would only arm the 180 s optimistic read for a signal the backend already publishes); `busy` parks the same compact op the drain retries; a reason is said with the session named, and the press, when the typed route carried a copy id, so the chat retires the bubble it drew and puts the words back, filed so POST /compact and `romp compact` answer ok:false with it, and kept on the bell. The typed "/compact" registers in the guard's handler table; words after the head are refused loudly (Codex takes no compaction instructions, and compacting anyway would drop them silently), read with a whitespace split so a newline or a tab after the head counts too. The drain fires a parked compact op, and a command op parked before this head existed, through the verb; a `busy` answer leaves the head with no clock and no in-flight record, as the clear arm does; a pre-upgrade command op with words is refused and popped. The composer's palette lists `/compact`. The battery's client-side decline from the guard is lifted on the chat and the timeline, so the batteries compact a Codex session too; the bell's explanation of the refused kind names its new tenant.

Two review folds on the bracket's edges: `clear` answers busy while a compaction stands (its idle test read only the open turn and the un-parked queue; a clear that slipped past the kernel's gate (two handlers, one gate read each, no atomicity with the handover by design) would have swapped the thread id under the bracket, and since the pump matches the compaction's statuses to a session by thread id, the bracket would have stood until a kill), so the kernel parks the clear and the drain runs it after the bracket ends. And the manual boundary is written through the normalizer's thread/compacted entry with a trigger and NO turnId; the in-turn automatic compaction's writer (the contextCompaction item as the boundary writer, with a dedup on the wire turnId; #1879, in the base) keeps a turnId-less call as one that always writes, so the two compose; this change touches no normalizer file.

Deviations from the spec, each with its reason: the battery lift on the chat and timeline pages was not in the spec's change list, but the spec's own docs paragraph and defect statement name the batteries as doors, and with the guard's mark standing they would have stayed closed; the palette row for `/compact` follows the native clear's precedent (the palette is what the route takes); a conversation with no records is refused in words before any request, where the spec named the zero-turn case only as a risk; the typed route's refusal frame carries the press id (the amendment left this open, and the clear handler already does it); the placeholder's words are "this session has no conversation to compact yet" where the two amendment sets differed slightly.

Open decisions the spec left to the maintainer, with what was built: the in-turn contextCompaction item to a boundary (an automatic compaction on a registered turn) landed separately as #1879 and is in this change's base, so an in-turn automatic compaction already writes its boundary; typed instructions are refused rather than compacted with a warning; the worker attempts a turn/start under a bracket with no active seen; POST /compact answers ok:false with the backend's words (a small contract change for `romp compact`); a second press while a bracket stands answers `busy` and parks, so the drain runs a second compaction after (the SDK's behavior); persisting the bracket on the registry row is a follow-up. Named residuals: if both the previous turn's active and idle are still unread at the latch, the bracket ends falsely and a second boundary can follow; a send that bypasses the kernel gates inside the ack-to-active window can be accepted and end the bracket while the acked compaction still runs; on success the composer's optimistic bubble stands until its cross (no echo, no record, the same as a typed /model on Codex); Esc does not stop a compaction; the failed compaction's own error message never reaches the global queue, so the notice names only the status (a thread/turns/list read after the end would carry the message: a follow-up, one more request on the pump's thread); and that notice rides the launch-error card every Codex failure rides, drawn since the review round with no retry meta and no Retry buttons (nothing retries a compaction; the next accepted turn clears it), so the third screenshot, taken before the round, still shows the buttons.

## Screenshots

Synthetic data throughout (the notes-api demo sessions under `/TESTDIR`, placeholder ids, invented prompts) over a hermetic lab kernel built from this branch, the Codex backend's scripted client standing in for the app-server.

A Codex session mid-compaction: the battery click ran Codex's own compaction, the chat shows the live compaction row and the statusline reads Compacting context, and a prompt typed meanwhile waits as the queued bubble behind it `compacting-chip.png`

The same conversation after the thread went active then idle: the Context compacted divider marked manual where the compaction happened, no summary body, the chip back to Ready, the battery down, and the parked prompt delivered after the divider with its reply `compacted-divider.png`

The failed compaction: after a systemError status the red card says Codex could not compact this conversation and the conversation continues as it was; no divider, the two turns stand, the chip reads Ready `failed-compaction.png`

## With the known-heads refusal

`/compact` stays in `_CODEX_REFUSED_HEADS`, as `/clear` and `/new` do: a registered head is taken on the route when the command is the whole message, a parked one runs at the drain through the compaction arm ahead of the refusal, and the set is what refuses the head with words under it, in the wording that says what the message must be. The Codex doc's paragraph names `/compact` as taken ahead of the refused list and gains a paragraph on what native compaction does and does not do; the guard's census skips every registered head and pins the skipped set as the handler table, its seam test and the clear route's refused example use `/fast`, the known head that stays unregistered.

## The live legs

The spec's reviewers made two live legs a requirement before merge, since CI cannot observe the status transition. Both ran against the pinned runtime (codex-cli 0.153.3 through the pinned client, the existing login, a scratch working directory), with every wire line and everything that reached the client's global queue recorded.

Resume then compact: a thread was started and given one short turn on one app-server; that app-server was stopped; a fresh one resumed the thread and requested the compaction. The resume itself published the thread's idle status just before the request (the stale idle the active-seen rule ignores). The request was acked in 11 ms; the thread went active 80 ms later and idle 1.8 s after the ack, with the compaction turn's own notifications (turn started, the contextCompaction item, token usage, turn completed) parked under the unregistered turn and never on the global queue; a turns listing afterwards showed the compaction turn completed with no error; a turn after it answered normally. The resumed-thread precondition holds live.

Forced failure: the thread's model can be set at resume, and the runtime accepted an invented model name for a thread with history. The compaction request was acked in 13 ms and the thread went active 65 ms later; 1.1 s after that the thread's status went to systemError, and no idle followed. The compaction turn's error notification ("Error running remote compact task", a 400 naming the unsupported model, willRetry false) and its turn completed with status failed both carry the turn id, so the pinned router parked and discarded them: the status is the whole signal that reaches the backend. A turns listing afterwards showed the compaction turn failed with that message, and the thread stayed usable (a later app-server resumed it and ran a turn). So a failed compaction IS distinguishable on the global queue, by active then systemError instead of active then idle, and the idle-after-active writer never runs for one: the systemError arm ends the bracket loudly with no divider, now worded as Codex not having compacted the conversation, and a scripted-client test pins that shape. No confirming turns-list read was added, since it would be one more request per compaction for no new information. A third leg redirected the ChatGPT backend URL to a closed local port as a network cut; the compaction completed anyway (the remote compaction does not go through that setting on a ChatGPT login, and the built-in provider's URL cannot be overridden), so it stands as a control, not a second failure shape.

## Tests

Red first on the stock tree (the #1880 branch head), one module at a time.

`tests/test_codex_backend.py`, new class NativeCompact, 18 tests, all red on stock: twelve on the verb's absence (the spec's named first failing line, `AttributeError: 'CodexBackend' object has no attribute 'compact'`), one on `compacting`'s, and the review round's four, red on the pre-review build (clear under the bracket answered the empty string and swapped the thread id, `'' != 'busy'`; the refusal for a compaction that ended with no bracket standing wedged the send behind a re-latched bracket; the kill before the handler left a dead row compacting and resume revived it so; resume left the bracket fields as they were), plus the failed compaction's live shape split from the notLoaded case. The scripted client gains a recorded compaction request answering the empty ack, a way to feed the global queue, and a refusing flag that makes turn/start raise the exact refusal, with a hook run before it raises. Green after: the bracket latches before the request returns and busy, compacting and the live state all read it; a stale idle is ignored, an active then an idle end it, and exactly one boundary lands as the file's last record with a null parent, the pre-compaction leaf as its logical parent and the manual trigger; a send typed under a seen bracket makes no turn/start for longer than the old retry, files no error, and runs after the idle, after the boundary; the refusal on a compaction romp did not start latches the bracket with no error, no second attempt and the prefix intact, and the idle alone ends it with one boundary; the refusal made to arrive after the compaction's own statuses re-latches nothing and the retry lands at once; a raising request unlatches and answers in words; an open turn, a queued send the worker has not taken, and a standing compaction each answer busy; the client's death ends it with an error naming the app-server and no divider; a systemError status (the failed compaction's live shape: active, systemError, no idle) ends it loudly with no divider, persisted, cleared by the next accepted turn, the thread usable after; a notLoaded status ends it as the thread having stopped being available; a restored row resumes the thread before compacting, builds the normalizer before latching, and chains the boundary to the file's leaf; no client, a placeholder, an empty conversation, an ended and an unknown session refuse in words with no request; a kill ends the bracket and a late active and idle write nothing; an accepted turn under a bracket the server never took up ends it with the log line and no divider; a thread/compacted pushed onto the scripted client's global queue while the bracket stands writes once, marked manual (a path the pinned router would not take for a turn-stamped one); clear answers busy under the bracket, mints no thread, and the idle after the active still ends it with the boundary; a refusal whose compaction already ended with no bracket standing stands down and the retry lands at once; a kill landing before the refusal is handled latches nothing on the dead row and the revived row runs its send; resume resets the bracket a kill left on the row.

`tests/test_codex_compact_route.py`, new module, 14 tests, all red on stock (the door answered the guard's refusal, the route had no head, the drain refused the parked ops, the predicate did not exist). Green after: the door calls the verb, never the send nor the stamp, and POST /compact's brain answers the same; a gated or busy door parks the compact op; a refusal is broadcast naming the session, filed for POST /compact, on the bell; a send-only backend and the unowned route keep the literal send and the stamp; the typed head takes the route with one gate read and Claude's path is untouched; typed instructions (a space, a tab) are refused with the words on the socket with the press named and never reach the verb, and a second line after the head is refused by the route's whole-message rule before the handler; a busy or gated typed compact parks the same op; a refused typed compact is said with the press named, filed, answered by POST /send; the palette and the table list the head; the drain fires a parked compact and a pre-upgrade parked command natively and ends the pass, refuses a pre-upgrade command with words, leaves a busy head with no clock and its cancel working, and warns and pops a refused one; and, over the real backend with the scripted client installed as the kernel's singleton, a typed compact latches the bracket the kernel's compacting gate reads, parks a message typed meanwhile, and the idle after the active writes the divider and lets the drain deliver the parked message.

`tests/test_session_api.py` gains the contract test in `clear`'s style (red on stock: the ABC had no such attribute). `tests/test_codex_events_golden.py` gains the manual-trigger case (green at the merge base: the trigger passthrough it reads came in #1879, an ancestor of the base; its two assertions pin the manual trigger this backend sends). `tests/test_codex_slash_guard.py` is re-pointed off the head now taken (the refused examples are /help and /fast; its compaction class leaves for the new module; POST /compact over the served handler now answers ok with the verb called; the palette lists the new row); `tests/test_codex_clear_route.py`'s handler-table pin is widened. Node: `ui/webview/codex-slash-guard.test.ts` pins the lifted battery mark on both surfaces and the bell's wording; the settings-preview and colormap pins return to the status-less fill wrapper; the inert-tooltip case leaves with its mechanism.

Run locally one module at a time after the review folds and the rebase onto the folded native clear: test_codex_backend 150 (+5 subtests); test_codex_compact_route 14; test_session_api 14; test_codex_events_golden 30; test_codex_slash_guard 13; test_codex_clear_route 12; test_kernel_compact_route 14; test_kernel_send_park 51; test_kernel_parked_ops_lock 19; test_kernel_meta_command_gate 7; test_backend_call_parity 4; test_injected_voice 10 (+222); test_kernel 460 (+12); test_perf_stats 62; test_feed_memo_inputs 31; test_chat_build_sig_inputs 65 (+10); test_perf_bench 49: all green. The pre-review build also ran, green, test_kernel_pusher_snapshot 31, test_kernel_headless_ops 20, test_kernel_model_queue 9, test_drain_hoists 6, test_queued_copy_press_id 23, test_kernel_ops_gate_busy 7, test_kernel_parked_ops_liveness 9, test_model_live_midturn 17, test_kernel_slash_echo_retire 17 (+17), test_kernel_send 15, test_session_move_kernel 19, test_stage0_log_readers 36, test_sessions_listing 10, test_kernel_deliver 3, test_codex_echo_merge 4, test_codex_launch_error_card 4, test_codex_models_route 10, test_codex_runtime 11 (+6), test_compact_suggest 22, test_sdk_compacting_signal 10, test_notice_card_store 38, test_kernel_cancel_create 6, test_sdk_registry_blind 31 (+12), test_sdk_kernel 38, test_kernel_limit_queue 20, test_event_model_golden 110, test_kernel_fed_echo_absorbed 23, test_lost_send_event 4, test_kernel_push_responsiveness 15, test_codex_discovery 6, test_scanner_folds 12, test_session_env 43, test_clear_chat_view 14, test_episode_boundary 17, test_kernel_cmd_gesture 5, test_kernel_cmd_cache 6, test_pr_watch 65. Static pin scan over every assertIn literal: 0 lost. Extension suite (vscode-extension npm test, which pins kernel and render source lines): 5060 tests, 5053 pass, 1 fail, the pre-existing KaTeX pixel probe that fails on main in this environment.

Runs at the filed head (rebased onto main after #1880 merged), one module at a time: test_codex_compact_route 14, test_codex_clear_route 19, test_codex_slash_guard 19, test_codex_backend 166 (+5 subtests), test_codex_postal_tools 26, test_session_api 14, test_codex_events_golden 41, test_kernel_compact_route 14, test_kernel 460 (+12), test_chat_build_sig_inputs 65 (+10), test_backend_call_parity 4, test_kernel_send_park 51, test_kernel_parked_ops_lock 19, test_kernel_meta_command_gate 16, test_injected_voice 10 (+222), test_feed_memo_inputs 31, test_perf_bench 49, test_perf_stats 65; the extension suite 5168 passed, 1 failed (the KaTeX layout probe that fails on main in this environment), 6 skipped; the assertIn pin scan against the merge base finds nothing lost.

## Review round

The maintainer agent's executed review of the filed head found one press that did the wrong thing, one doc sentence that named a recovery that does not exist, and a handful of smaller gaps around the bracket's edges. Every item is folded in one commit on top of the head; each fold landed with a test that fails before it, except the two pins the review asked for on behavior that was already right and the description edits.

**The failed compaction's card had a Retry that sent the word "retry" into the thread as a turn.** The systemError and notLoaded ends file a launch error, and every launch error rides the API-error card: a retrying-soon meta and two Retry buttons. Nothing retries a compaction; the press started a turn with the literal word as its prompt, the accepted turn cleared the card, and no compaction request was made. The bracket's end notices (the two statuses and the app-server's death) now carry a noRetry field on the launch error, the chat build lifts it onto the status as apiNoRetry (the apiRefusal precedent: the card reads its flags from the live session status, never the event), and the card returns before its retry machinery is built: the same red card, the words alone, no meta, no actions. Scoped to the bracket's ends: a failed turn's card keeps its button, and the contract names the optional field. Red first over the real backend in the route module (the built status carried no such flag; the notice carried no such field), and at the backend's three loud ends.

**The doc's recovery for a compaction Codex acknowledges but never runs did not exist.** The sentence said the session reads compacting until the next message probes it. Under a bracket latched with no thread status seen, every message door parks and the drain skips the session, so nothing typed into it ever attempts the turn that would end the bracket. The doc now says so and names the way out: End then Revive, which keeps the thread and its history and delivers the waiting message, or a kernel restart. No code changed for this item; the review weighed the code alternative (publishing the bracket only once the active status is seen) and preferred the doc, since that alternative widens the named residual to every message typed in the window. The compact verb's docstring, which repeated the clause, says the same as the doc now.

**A compaction refused after a battery click was never said to the timeline lane that took the click.** The refusal warned the chat panes only; the timeline page drops a warn, and its click stamp cleared only on a settingRefused frame or its six-second expiry, so the lane read compacting for six seconds with no words. Beside the chat broadcast, the kernel now sends every timeline pane a settingRefused frame (gesture command, no flag: the shape the page's HTTP road already builds for a refused compact), through the app broadcast rather than as the refusal's socket, which would have replaced the chat broadcast. The double bell row the review named is dropped rather than kept: the frame carries filed, and the page posts no shell notification for a frame so marked, because the kernel's own ring row is the bell's record; the HTTP road's frame still files its row. The drain's refusal of a parked compaction sends the same frame: a parked click is a click, and the lane whose press parked still holds its stamp until the frame arrives. The typed route, whose words ride the composer's socket, sends no timeline frame. Red first in the door test (no timeline frame) and, executed, in the timeline panel's handler (the marked frame still posted a row); the drain's frame is pinned in the drain's refused-op test.

**The chat battery's click pulse never ended on a refusal.** The click added the pulse class and only a compacting status removed it, so a refused compaction pulsed until the next status change or a tab switch. The remover is one exported function the compacting branch and the chat's warn branch both call; the warn branch ends the pulse when the frame names the active session. Executed red first in the status-controls module (the function did not exist), the chat's wiring pinned at the source.

**Esc, Stop, POST /interrupt and romp interrupt during a compaction answered "the session is still working" under a chip reading Compacting.** A compaction reads busy (it is a turn in flight on the server) and the backend refuses to interrupt it (there is no turn id to stop), so both doors fell to the busy branch. An arm ahead of it reads the backend's own compacting verdict, by identity only (a double without the verb has no such attribute; a bare Mock's call answers a truthy Mock; neither is a compaction), and answers in the state's words with no backend named: the stop was not delivered, the session is compacting its conversation, and a compaction runs to its end. Red first in the headless route module and in the WS arm's module.

**The two new launch-error writers saved the registry after releasing the session lock.** The status handler's loud end and the pump's client-death end set the field under the lock and saved outside it, against the registry transaction's rule that every mutator keeps the session lock through it: a turn accepted in that window cleared the field and saved None, and the stale snapshot then committed the red card over it, so the next restart restored a card the turn had cleared. Both saves moved inside the block that sets the field, the shape the worker's own handlers use. Red first with an interleaving test at each site: a spy on the snapshot opens the window where it opens (the snapshot is taken before the registry lock), starts the clearing writer (a real send whose accepted turn clears the field at the status end; the accepted turn's two lines on another thread at the death end, where no turn can run) and gives it two seconds; before the fold the stale loud record was the last commit on disk at both sites.

**The loud ends of the bracket never poked the kernel.** The clean end poked through its boundary write, but a systemError, a notLoaded and the client's death only pushed and kicked, so a message parked on the bracket waited for the pusher's half-second backstop instead of the end event. Every end pokes now, and the three comments that described the backstop as the path describe the poke. Red first with a poke recorder that reads the bracket at poke time: before the fold no poke after the latch found the bracket down.

**Three descriptions the change made false.** The romp compact line in the reference and the romp help line still described the command as Claude's /compact only; each gains a clause: on a Codex session it runs Codex's own compaction, with no summary text. The backend's states comment named waiting and working as the two states it can know; compacting is the third. The guard module's docstring said the compaction door refuses ahead of any park or stamp; it now says the door pays the one gate, then takes the backend's verb, and only a reason the backend answers refuses.

**Tests the review asked for.** The bracket-end counter half of the refusal's staleness check had no test: every stale-refusal case pushed an idle, so dropping the end-counter key left the module green while the case it guards (a loud end read before the refusal handler runs) re-latched the bracket and parked every send. The new case latches a bracket, makes the scripted client refuse with a hook that pushes active then systemError and waits for the bracket to fall before clearing the refusal, sends, and asserts the queue drains, no boundary was written and the counters read one end and zero idles; dropping the end-counter key turns it red. The claim that the divider lands before the end is published was exercised only by a scheduling race; a spy on the append records the bracket flag at the boundary write and a spy on the end records the sequence, asserting the write with the bracket standing and then the end, reading the flag only (the append runs under the normalizer lock, where a push would deadlock the pump). The two fixed sleeps before "no boundary" assertions, and the optional third, now wait on the thread's idle count instead.

**Runs at the fold head**, one module at a time under the lock, after the amend that restored the revived-row test's request assertion, reworded the compact verb's docstring and pinned the drain's timeline frame (none of which changes a count): the Codex backend module 171 (+8 subtests), the compact route module 15, the headless ops module 21, the WS interrupt module 55, the clear route module 19, the slash guard module 19, the kernel compact route module 14, the kernel module 460 (+12), the events golden module 41, the session API module 14, the chat build signature module 65 (+10), the injected voice module 10 (+222), the launch-error card module 4, the cold tab gate module 22 (+4), the refusal retry module 17; the extension suite 5179 tests, 5172 pass, 1 fail (the KaTeX layout probe that fails on main in this environment), 6 skipped.

**A fourth screenshot, from the round's head.** Taken with the same driver as the three above, over a lab kernel built from the round's commit.

The failed compaction after a systemError status: the red card reads that Codex could not compact this conversation and the conversation continues as it was, with no retrying-soon meta and no Retry buttons under it; no divider, the two turns stand, the chip reads Ready `failed-compaction.png` (the round)


lczh:
Reopening to re-run CI: the extension job failed in tests/test_rail_relative_served.py (the driver read a rail marker's rect before it existed), which this branch does not touch; the same lab passes on this head locally through the extension's Chromium, 1 passed in 9.5 s.

ksarma:
The evidence below was executed against 7281106be, the PR's head; file and line references are on that commit.

### At this head

- **A Retry press on the failed compaction's card sends the word "retry" into the thread as a turn.** The `systemError` and `notLoaded` ends file a plain launch error (kernel/codex_backend.py:1994, 2018), which `build_session` renders as an `apiError` event, nothing suppressing the retry actions. The press sends the literal retry prompt as a turn, the accepted turn clears the card, and no compaction request is made; only the manual press does this. The body names the dress; the press is not inert. Fix: a field on the bracket's end notices that the chat build reads, so the card renders without the retry meta and buttons, the `apiRefusal` precedent; `renderApiError` reads its flags from the live session status, not the event. A warn kept on the bell, the refusal's own route, also works. A bare `retryRefused` from the retry handler is weakest: the card would still show "retrying soon" and the feed's toast would be false. Scope it to the bracket's end notices; a failed turn's card keeps its button. Test: a `RealBackendCompact` case asserting the card carries no retry action.

- **The doc's recovery for a compaction Codex acknowledges but never runs does not exist** (docs/codex.md:237, "until the next message probes it"). Under a bracket latched with no thread status seen, every message door parks and the drain skips the session, so nothing attempts the turn/start that would end it; Esc, clear and compact answer busy and the 180 s cap does not apply. End then Revive keeps the thread id and history and drains the parked message, and a kernel restart drops the in-memory bracket; sends that skip the kernel gates (a peer's mail, a Retry click, a raw-mode file save) do end it, but none is "the next message" or discoverable; reachability is unknown since `compact` now refuses an empty conversation, the one observed trigger. Fix in the doc, not the code: no message typed into the session probes the bracket; the way out is End then Revive, or a kernel restart. The code alternative (publish the bracket through `compacting` only once the active status is seen, busy True meanwhile) makes the clause true at two costs: the chip and chat read Working until the active status (indefinitely in the never-runs case), and the body's named residual, a turn accepted in the ack-to-active window, widens from the bypass doors to every message typed then. If a code path is wanted, prefer an explicit drain handover after one status read with no active.

### Smaller points

- **A compaction refused after a battery click is never said to the timeline lane that took the click**: `_say_compact_refusal` warns the chat app only (kernel/kernel.py:36964), the timeline boot drops a warn, and the lane's click stamp (ui/romp-timeline-view.js:6601) clears only on a `settingRefused` frame (3656) or the 6 s expiry, so the lane reads compacting for 6 s with no words. Fix: beside the chat broadcast, send the timeline app a `settingRefused` frame through `_send_to_app`, the idiom at 36792; passing the timeline's socket as `client` into the refusal unchanged would replace the chat broadcast. The refusal already files a ring row through `_sync_notice` (36965) and the timeline's `settingRefused` posts a shell notify (3671), so a shell showing both carries two rows per refusal: drop the ring row on the timeline frame, or name the double; `test_a_refused_compact_is_said_and_filed_for_post_compact` needs revising.

- **The chat battery's click pulse never ends on a refusal**: the click adds `ctx-clicked` (ui/webview/render.ts:15884), only `setCtxBar` with compacting true removes it (ui/webview/status-controls.ts:230), and the refusal's warn branch (render.ts:19084) skips the bar, so the battery pulses until the next status change or a tab switch; remove the class in that warn branch when the frame's sid is the active session, pinned by execution.

- **Esc, Stop, POST /interrupt and `romp interrupt` during a compaction answer "the stop was not delivered: the session is still working" under a chip reading Compacting**: busy is True under the bracket and interrupt is the no-turn False, so the WS arm (kernel/kernel.py:19821) and the route (68134) fall to the busy branch. Fix: an arm on `be.compacting(sid) is True` ahead of the busy branch in both places, worded for the state, backend-neutral, keyed on identity because tests/test_kernel_headless_ops.py drives the route with a bare `Mock` whose compacting call is truthy; hiding the stop button instead moves exact-text assertions in interrupt-ux.test.ts and stop-button.test.ts.

- **The two new launch-error writers save the registry after releasing the session lock**: the status handler's loud end (kernel/codex_backend.py:2024) and the pump's client-death end (1229) call `_save_registry` outside the lock that set the field, against that function's docstring, so when a turn/start completing in that window clears the field and saves None (2658 to 2663), the stale snapshot commits after it and the next restart restores the cleared red card; move the try/except around the save into the block that sets the field at both sites, the `_work` handlers' shape (2505 to 2515), with an interleaving test (the PR's test has no concurrent writer).

- **The loud ends of the bracket never poke the kernel**: the status handler pokes only when it wrote a record (kernel/codex_backend.py:2020) and the pump's client-death arm (1227) never does, so after a loud end a parked message waits for the pusher's 0.5 s backstop instead of the end event; an unconditional `self.poke()` at 2020 and the same call in the pump's ended block before the push, with a test and updates to the three comments naming the backstop (codex_backend.py near 1694, kernel.py near 37302 and 37318).

- **Three descriptions the change makes false.** docs/reference.md:46 and the `romp help` line at bin/romp:90 still describe `romp compact` as Claude's /compact only, while the command now reaches the backend's verb: add one clause to each (on a Codex session it runs Codex's own compaction, no summary text); the `--wait` sentence stands. kernel/codex_backend.py:556 still names waiting and working as the backend's two states, while `compact` (1900), the bracket's end (1926) and the refusal handler (2486) set or clear "compacting": name the third. tests/test_codex_slash_guard.py:16 still says the compaction door refuses ahead of any park or stamp: drop the clause (the next sentence already points at the door's tests) or reword it to the head's order: the gate, then the backend's verb, only a backend reason refusing.

- **The golden case is green at the merge base.** `test_a_manual_compaction_is_marked_so_and_an_emitted_one_stays_auto` (tests/test_codex_events_golden.py:356) passes at the base with only this PR's test file overlaid (the passthrough it reads came in #1879, an ancestor of the base; the two `NativeCompact` assertions pin the manual trigger): drop the body's red-before claim for it. Two body sentences are stale for the same reason: the normalizer edits being the passthrough only (the diff has no normalizer file) and the in-turn `contextCompaction` boundary not being included (#1879 is in the base).

### Tests

- The bracket-end counter half of the refusal's staleness check (kernel/codex_backend.py:2482) has no test: every `NativeCompact` test exercising a stale refusal also pushes an idle, so dropping the end-counter key leaves tests/test_codex_backend.py green, while the case it guards (a loud end read before the refusal handler runs) re-latches the bracket and parks every send; in the shape of the test at 3991: compact so a bracket stands, set the scripted client to refuse with a hook that pushes active then `systemError`, waits until compacting reads False, then clears the refusal; send; assert the queue drains, the boundary list is empty and `(s.compact_ends, s.compact_idles) == (1, 0)`; a kill variant adds nothing (the dead check runs first). The claim that the divider lands before the end is published is exercised only by a scheduling race (the poll-then-read tests at 3932, 4157, 4243 and 4266 pass in either order), so pin it with a spy on `be._append` recording the compacting flag when a `compact_boundary` record is appended (reading only the flag, no push: that function runs under the normalizer lock) plus a spy on the bracket's end recording the sequence, asserting True at the write and write then end. At 4205 and 4229 a `time.sleep(0.2)` precedes a "no boundary" assertion that a lagging pump passes even with the standing-bracket test removed, so wait for the idle count to advance instead (optionally at 4246; not at 4107, where nothing follows the `systemError`; optional at 3925, covered by the wait at 3929).

henrypinkard (approved):
Approved under the maintainer's standing authorization of 2026-09-20 for Codex-related pull requests by this author. Follow-ups noted for a later round: an event-based stand-down for a compaction that reports active and then nothing; the timeline's cue after a refused click; the Retry buttons on a failed Codex compaction; the two wordings for a multi-line /compact.

henrypinkard (approved):
Re-approved on the head the branch update from main created (no change to the pull request's own diff).

lczh:
Thanks for the executed pass. Every item is folded and tested on a commit above this head: the Retry press on the failed compaction's card (the bracket's end notices carry a no-retry field the chat build lifts onto the status, and the card draws with no meta and no actions), the recovery sentence in the Codex doc (End then Revive, or a kernel restart), the timeline lane's refusal frame after a battery click, the battery pulse ending on a refusal, the stop and interrupt wording under a compaction, the two registry saves moved under the session lock, a poke at every loud end, the three stale descriptions, and the end-counter, ordering and wait tests. The backend, route, guard and interrupt modules and the extension suite are green on it.

Since the owner approved this head and queued its merge, I am holding that commit rather than moving the head. It files as a follow-up pull request on main right after the merge, with the review's items as its body and a note back here.

The body above is corrected now for the three sentences the golden item showed stale: the manual-trigger case is green at the merge base (its passthrough came in #1879, an ancestor), and the in-turn writer landed as #1879 and is in this change's base. The fourth, the card's retry dress, stays as written until the follow-up lands.

lczh:
Reopening to re-run CI: the extension job on the branch-update head failed in fifteen browser labs across eight modules this branch does not touch (the chat-split lab's driver waited fifteen seconds for a frame's focus ring on every case, and seven unrelated labs timed out the same way in a thirty-two minute run), while every Python job passed. The chat-split module passes on this head locally through the extension's Chromium. If closing cleared the queued auto-merge, it needs re-enabling; the head and the approval are unchanged.

lczh:
The round lands here after all. The re-run of the approved head failed in this branch's own failed-compaction test on one interpreter: the test read the registry file for the red card right after the bracket fell, and the loud end had set the card under the session lock but saved it after the release, so the file could still carry the record from before the end. That is the sixth item of the review above, and the round's commit moves both loud-end saves under the lock, so the read cannot land in that window (the backend module now waits on the locked read, 171 passed with 8 subtests on this head). With the approved head unable to go green on its own, holding the round for a follow-up would only have delayed both, so the reviewed commit is rebased onto that head and pushed: one commit above it, the diff otherwise the one the review executed, every item folded with a failing-first test, the body extended with the round and a fourth screenshot of the card without its buttons. Locally on this head: the backend, compact route, headless ops, interrupt, slash guard, clear route and kernel modules are green, and the extension suite passes apart from the KaTeX layout probe. The approval will need renewing on the new head.

henrypinkard (approved):
Re-approved on the new head under the maintainer's standing authorization for this author's Codex work.

lczh:
For the merge queue: this head is approved (06:00Z, on 40a2cdfd), every required check is green, no review thread is open, and auto-merge has been queued since about 06:50Z, yet GitHub reports the merge state as blocked. My other open pull request reads the same state with all checks green, and every pull request of mine that landed today was merged by the owner's hand, so the rules on main seem to leave the last step to the owner rather than to auto-merge. Nothing on this side is pending; a manual merge is the way through.

ksarma:
The evidence in this comment was executed against 40a2cdfd9, the PR's head; file and line references are on that commit.

The commit at this head closes the earlier comment's items, each checked by execution against the case that comment drove: defeating each code change (the no-retry field and the kernel's lift of it, the compacting predicate behind the new stop wording, the timeline frame and the gated parent post, the saves under the lock, the pokes beyond the boundary write, the end-counter key, the boundary ordering) turns its test red, with a failed turn's card still retryable and the bare-Mock pin still green; the pulse's remover is executed and was red at 7281106be; the doubled bell row is closed by the filed gate; the idle-count waits and the sleep at tests/test_codex_backend.py:4109 stand as asked; the descriptions and the body's three sentences read as asked; and the module counts in the body and the 05:57Z comment were reproduced one process at a time.

The doc's clause "until the next message probes it" is gone; the sentence that replaced it is the first item.

### At this head

- **The recovery now in the doc, End then Revive "delivers the waiting message", is false: only a kernel restart delivers it** (docs/codex.md:240-241; the compact verb's docstring at kernel/codex_backend.py:1860-1861; the body's review paragraph). The earlier comment proposed that recovery from a read of the code, and the read was wrong; two independent runs over the real Codex backend at this head agree. A message typed under a bracket with no status seen parks in the kernel's parked-ops queue; End kills the session without touching it and finishes with a push-soon (kernel/kernel.py:20234-20240), so the drain runs within the same wake and no Revive can land first. The drain has no liveness gate (37257-37262): it pops the send and hands it to a row that now reads as unowned (`owns` is live-only, kernel/codex_backend.py:1368-1373), whose refusal the delivery ignores, so the chat hears nothing; Revive drains only the backend's own queue, which is empty, and a restart delivers because the parked queue is mirrored on disk and the in-memory bracket is not. Fix, in all three copies: a kernel restart delivers the waiting message; End then Revive keeps the thread and its history, but a message still waiting behind the cue is dropped when the session ends, so take it back first from its queued bubble (the cross, through the park arm at kernel.py:19965-19973 to `_cancel_parked` at 36165; the client restores the composer on the ok result); the docstring's closing clause, that the doc names two reachable ways out, changes with it, since only one keeps the message. If a code path is preferred, take the End arm cancelling the ending session's parked sends loudly through the existing not-delivered path (`_refuse_drive`, kernel.py:19709-19746, whose err frame hands the text back in its copy slot), not a drain that keeps a dead row's run for a later Revive: that contradicts the drain's contract at 37468 (a dead session's queue is dropped once, never retried) and would strand queued bubbles on a session never revived. Either owes a `RealBackendCompact` test red at this head, the probe's own sequence: spawn, one turn, a compact with no status, a typed send that parks, kill, one drain pass, resume, drain again, assert the text reaches the turn start or the err frame.

### Smaller points

- **`romp compact --wait` prints done and exits 0 after a compaction that ended loudly** (bin/romp:1897-1899), because the poll reads only the row's compacting field, which falls on every end (a `systemError` end over the real backend left the sessions row byte-identical), so a scripted recycler pointed at that field, as docs/reference.md:48 suggests, reads success over an uncompacted thread; fix: carry the launch error on the row (`_session_listing_row`, kernel/kernel.py:28268) and exit 1 with the notice's words on stderr when the row's launch error is new since the arming poll, not merely present (the compact verb does not clear a standing notice when it latches a new bracket, kernel/codex_backend.py:1905-1912, so a later clean wait on the same thread would otherwise exit 1 falsely) and whether or not the compacting sample was caught, and add the failure exit to the reference row's wait sentence and the help line at bin/romp:90.

- **In a hosted timeline with no shell bell (VS Code's view), the refusal frame ends the lane's cue but puts the words nowhere** (ui/romp-timeline-view.js:3673, 3675): the filed gate skips the parent post, which VS Code's host drops anyway, the gear branch runs only outside a shell, and the chat's warn broadcast reaches VS Code only while the chat panel's pipe exists, a gap that predates this change (every command-gesture refusal is silent in VS Code by the same path) and is a follow-up rather than a condition on this PR; fix: compute a bell as a shell whose parent carries a `__rompNotify` function (the same-origin probe kernel/kernel.py:60043 already makes) and widen only the gear branch at 3675 from no shell to no bell, keeping the parent post gated on a shell and an unfiled frame because ui/webview/codex-slash-guard.test.ts:71-73 and ui/webview/setting-refused.test.ts:63-74 stub a parent without the hook and assert the unfiled post lands, pinned by execution beside codex-slash-guard.test.ts:52-75 (without the hook a filed frame sets `_laneRefusal`; with it, nothing is set or posted).

- **The client-death end of a compaction files the red card but logs no line naming the session** (kernel/codex_backend.py:1235-1237): after the app-server died mid-compaction the backend logged only "global pump stopped" and "client unavailable", while the `systemError` end logs "compaction of web ended" with the reason (2044); fix: build the text under the lock, store it on the launch error, log it after the release in the shape of 2044 with the session's name and the text, and extend `test_the_clients_death_ends_the_bracket_loudly` (tests/test_codex_backend.py:4071) to build with a log recorder and assert exactly one line after the close starts with "compaction of web ended", red at this head.

- **The launch-error field comment still enumerates three keys** (kernel/codex_backend.py:584) after the optional `noRetry` the writers set at 1228 and 2033, while the sibling states comment at 556 was updated; name the optional key with plain punctuation, pointing at kernel/session_backend.py:196-199, which already says what it means (comment-only, no test owed; kernel/kernel.py:14699 could take the same word).

### Tests

- The `systemError`-site lock test (tests/test_codex_backend.py:4484) goes red before the fix on "the loud record was committed" rather than on the ordering at 4485, because it asserts as soon as the turn is no longer busy and races the pump's late stale write (10 of 10 pre-fix runs failed at 4484, the loud write landing after the assertion, late, not lost), so replace line 4484's assert with `self.assertTrue(until(lambda: any(w and 'systemError' in w['text'] for w in writes)), 'the loud record was committed')`, after which the pre-fix run goes red at 4485 and head stays green; the death test's wait at 4506 already covers both writers.

- The pulse test (ui/webview/status-controls.test.ts:181) pins that the remover and the compacting state drop the click cue, not that a non-compacting bar refresh keeps it, so hoisting the remover above the compacting check in status-controls.ts leaves status-controls.test.js and codex-slash-guard.test.js green while the one-second bar refresh at ui/webview/render.ts:19512 erases the click's cue within a tick, before any status frame or refusal; between line 180's add and line 181's compacting call run `SC.setCtxBar(bar, '62%', false, [1, 2, 3], false)` and assert the clicked class is still present, which fails under that mutant and passes at head.

- One clause of the reworded body sentence misattributes the manual trigger's pin: the golden case (tests/test_codex_events_golden.py:356) loads only the event model and the normalizer and stays green when both backend trigger sites (kernel/codex_backend.py:1993 and 2001) change from manual to auto, while the `NativeCompact` class fails at tests/test_codex_backend.py:3942 and 4257 and forcing the normalizer's passthrough to auto turns the golden red at 356, so the case pins the normalizer's passthrough of a manual trigger (#1879) and the two backend assertions pin the trigger this backend sends; body wording only, the earlier comment's other two wording corrections having landed.

lczh:
Thanks for the second executed pass. The head under approval stays where it is, so the owner's merge is not held up again; the items land as a follow-up fix on top of it, built now and filed against main the moment this merges. The first item is taken as you read it: End then Revive does not deliver a message parked behind a compaction that never ran, since End's push-soon drains the parked queue into a row that reads as unowned and the refusal is ignored; the doc, the compact verb's docstring and the body will say that only a kernel restart delivers it and that the message should be taken back from its queued bubble before End, and the End arm will cancel the ending session's parked sends loudly through the existing not-delivered path, with the probe's own sequence as the red-first test. The compact --wait exit on a loud end, the client-death end's log line, the launch-error field comment, the two test pins and the body's manual-trigger clause ride in the same follow-up; the hosted timeline's silent refusal stays the separate follow-up you scoped it as.

lczh:
Main merged into the branch at 1144a905 (c507b6b4 into 40a2cdfd). One conflict, in ui/webview/render.ts: the `Status` interface line, where main reworded the `needsYou` comment (the Needs you ring words from #1935) and this branch added `apiNoRetry` and its sentence. The resolution keeps main's line and adds the field and the sentence; nothing else in the file moved. The branch's diff against main is byte-identical to its diff before the merge for every file (merge base 0d42228e), the `Status` line apart.

On the merged tree: the extension suite 5192 tests, 5185 passed, 6 skipped, 1 failed (the KaTeX layout probe, which fails on main in this environment); the branch's Python modules one at a time: test_codex_backend 171, test_codex_clear_route 19, test_codex_compact_route 15, test_codex_events_golden 41, test_codex_slash_guard 19, test_kernel_headless_ops 21, test_kernel_interrupt 55, test_session_api 14, all passed.

The head moved, so the approval on 40a2cdfd needs renewing; the follow-up fixes named in the reply to the second review still file after this merges.

henrypinkard (approved):
Re-approved on the new head under the maintainer's standing authorization for this author's Codex work.

PR #1934, opened 2026-09-21
The docs site's strict build runs on pull requests that touch the site's inputs

henrypinkard:
The docs site's strict build (`mkdocs build --strict`) ran only on pushes to main, so a docs pull request could land a broken link and the site broke after the merge; the restoration merged on 2026-09-20 resolved its links by hand. The Docs workflow now also runs on PULL REQUESTS that touch the site's inputs, as a build-only job.

**Triggers and the path set.** `push` to main and `pull_request`, both on the same four inputs: `docs/**`, `mkdocs.yml`, `overrides/**` and `.github/workflows/docs.yml` (the README is not read by the site: a broken link appended to it builds at exit 0 and appears nowhere in the output, so it is not on the list). `workflow_dispatch` stays.

**A pull request's run stops after the build.** The artifact upload and the deploy job both carry `github.event_name != 'pull_request'`, so a pull request proves the site builds and publishes nothing. Its run keys its concurrency on its own ref (`docs-pr-<ref>`) and a newer push cancels the older run; pushes to main keep the single `pages` group; an in-progress publish is never cancelled, and a waiting one is replaced by a newer push (GitHub keeps at most one waiting run per group).

**Orphan pages fail the build too, now as a fact.** MkDocs reports a page in neither `nav` nor `not_in_nav` at info, which strict mode does not promote, so the workflow's comment used to claim more than the build did (the review's executed find). `mkdocs.yml` raises `validation.nav.omitted_files` to warn and lists `codex.md` under `not_in_nav`, the one page that lives outside the nav today, so the build stays green and an orphan page turns it red.

**Tier.** Fix: the change repairs a gap in the checks, it is not documentation. A change under `.github/` needs the owner's approval when the author is not an admin; this session acts as the owner, so the tier check merges it on green.

**The proof is this pull request's own runs.** A throwaway commit added a deliberately broken link to `docs/index.md` and the strict build went red on the pull request: run 35547376517 (https://github.com/romp-on/romp/actions/runs/35547376517), `WARNING - Doc file 'index.md' contains a link 'this-page-does-not-exist.md', but the target is not found among documentation files.` then `Aborted with 1 warnings in strict mode!`, the deploy job not run. The next commit reverted the link and the build went green: run 35547509202 (https://github.com/romp-on/romp/actions/runs/35547509202), the upload step and the deploy job both skipped. Nothing else changed between the two. A config pin keeps the shape: `tests/test_docs_workflow_pins.py` reads the workflow and the site config (a regex over the source, as `tests/test_ci_bats_bound.py` does, since the test dependencies carry no YAML library) and asserts the pull-request paths name the three inputs and match the push paths, that the upload step and the deploy job both stand down on a pull request, the per-ref concurrency, and the `omitted_files: warn` setting with `codex.md` listed.

Review closed: no verifier or panel run by me. The contributor's review of 2026-09-21 is folded: its four points are facts on this head.

Tier: fix


ksarma:
The change adds a pull-request trigger to `docs.yml` at lines 23 to 29, so the strict MkDocs build runs before a docs change merges. The artifact upload at line 56 and the deploy job at line 63 run only on push and dispatch events, and a pull request's run gets a concurrency group of its own at lines 40 and 41. The evidence was executed against 5654263d1, the PR's head, and every file and line reference here is on that commit.

Both proof runs did what the commit subjects say. Run 35547376517 fails on the deliberate broken link with one warning in strict mode; run 35547509202 at this head is green with the upload step and the deploy job both skipped; the revert leaves the tree identical to the first commit's. On a pull request `github.ref` is refs/pull/N/merge, so each PR has a group of its own and never queues behind the publish group. The group and cancel expressions match the pattern at lines 49 and 50 of `ci.yml`. A local strict build on CI's versions (mkdocs 1.6.1, mkdocs-material 9.7, Python 3.13) passes at this head and fails on the broken-link shape. The build check is not among the main branch's required checks, so a PR that touches none of the paths is not stranded.

## Smaller points

- The strict build does not fail on orphan pages, although the header comment at line 7 says a docs pull request could land "a broken link or an orphan page" and the step name at line 53 says the build fails on them. MkDocs reports a page absent from the nav at INFO, since `validation.nav.omitted_files` defaults to info, and strict mode promotes only warnings. Executed: the log of run 35547376517 prints an INFO line naming `codex.md` as a page not included in the nav configuration, and its only WARNING is the broken link. Run 35547509202 at this head is green with that page in neither nav nor not_in_nav (`mkdocs.yml` lines 67 to 81). A scratch build on mkdocs 1.6.1 with an extra page under docs and no nav entry exits 0 and writes the page into the site output. So a docs PR that adds a page outside the nav passes the new build, and a reader of lines 5 to 8 takes that case as covered. Line 134 of `tests/test_docs_front_pages.py` records the same belief about the strict build. Fix: drop "or an orphan page" from line 7 and "/ orphan pages" from line 53, or make the claim true by setting `validation.nav.omitted_files` to warn in the site config and placing codex.md under nav or not_in_nav so the build stays green.

- Line 28 lists the README among the pull-request paths, but the README is not an input of the site. Read: `docs_dir` is the default, no symlink under docs or overrides reaches the README, and no include or snippet in the site config or the pages does either. The push paths at lines 18 to 22 and line 30 of `scripts/docs-serve.sh` both define the input set as the docs directory, the site config and the overrides directory, so the two triggers disagree. Executed: a broken link appended to the README in a scratch copy builds in strict mode at exit 0, and the appended text appears nowhere under the site output, so MkDocs neither reads nor validates it. A README-only PR runs a build that can neither fail on nor render the README, a push that changes only the README does not run it, and "touches the site's inputs" at lines 5 and 6 misdescribes the path set. Fix: remove line 28; if the README is meant to become an input (an include or a symlinked page), wire it in first and add it to the push paths too.

- The PR body is still the two lines written before the proof runs (a note that the body comes after the runs, and the tier line), and the runs have concluded: run 35547376517, started 00:20:10Z, failed at 00:21:42Z, and run 35547509202, started 00:22:43Z, succeeded at 00:23:17Z, both as `pull_request` events. Read from the PR's body field, which has not been edited since the PR opened at 00:20:07Z. The description therefore lives only in the title, the first commit's subject and the comment at lines 5 to 8, and a reader of the PR finds no statement of what runs on which event and path set, and no link to either run. Fix: write the body with the triggers and the path set, that a PR run stops after the build (no artifact, no publish), the two run links, and without the orphan-page claim from the first point.

## Tests

- The red-then-green pair in CI is the fail-before evidence the fix label asks for, but nothing in the repo keeps it. Read: no test under tests reads `docs.yml`, so a later edit that drops the pull-request stanza at line 23 or either guard at lines 56 and 63 turns nothing red, and the state this PR fixes returns unnoticed. The repo already pins a CI file's properties by reading it: line 15 of `tests/test_ci_bats_bound.py` points at ci.yml and the tests assert on its content. Fix: a small test that reads docs.yml and asserts that the pull-request paths still name docs, mkdocs.yml and overrides (the three entries at lines 25 to 27), and that the upload step's condition at line 56 and the deploy job's condition at line 63 both contain `github.event_name != 'pull_request'`. That existing pin is a regex over the source because the test dependencies carry no YAML library (its docstring says so, and lines 91 and 329 of `ci.yml` install only pytest, pytest-timeout and cryptography), so the same shape fits here. A config pin is the only test shape available without GitHub; the proof runs remain the evidence that the build went red before the fix.

Edited 2026-09-21 01:05Z: the two proof runs' times now read their completion (00:21:42Z and 00:23:17Z) rather than their start; the body's unchanged-since figure now reads from the PR's edit history (never edited since it opened at 00:20:07Z) rather than the PR's update stamp; and lines 25 to 27 are named as three entries, since line 26 is a filename, not a glob.

henrypinkard:
The review's four points landed on 733cfdda, each as a fact on this head:

1. The orphan-page claim is true now rather than dropped: `mkdocs.yml` raises `validation.nav.omitted_files` to warn and lists `codex.md` under `not_in_nav`. A local strict build at this head passes with no info line about an omitted page; a scratch page in neither nav nor `not_in_nav` fails it (`WARNING - The following pages exist in the docs directory, but are not included in the "nav" configuration` then `Aborted with 1 warnings in strict mode!`). `tests/test_docs_front_pages.py` is untouched, as asked.
2. The README is off the pull-request paths; the two triggers name one input set (`docs/**`, `mkdocs.yml`, `overrides/**`, the workflow).
3. The body is written: the triggers and the path set, a pull request's run stopping after the build, the two proof runs (35547376517 red, 35547509202 green) with the strict build's words, and the orphan claim stated as the fact it now is.
4. `tests/test_docs_workflow_pins.py`, a regex over the source as `tests/test_ci_bats_bound.py` is: the pull-request paths name the three inputs and match the push paths, the upload step and the deploy job both carry `github.event_name != 'pull_request'`, the per-ref concurrency, and the warn level with `codex.md` listed. Nine tests pass alongside the front-pages module.

ksarma:
The evidence was executed against 733cfdda2, the PR's head, merged as bbf58e046. The merge carries the head's tree unchanged for the PR's three files, so every file and line reference here is on 733cfdda2 and reads the same at the merge commit.

The commit closes the four points of the comment at 00:56Z, each checked at this head. Orphan pages: on CI's versions (mkdocs 1.6.1, mkdocs-material 9.7.7, Python 3.13) the strict build at this head exits 0, and with a scratch page in neither nav nor not_in_nav exits 1 on the omitted-pages WARNING, where the base's config lists the page at INFO, exits 0 and renders it. The path set: `docs.yml` parses to the same four entries under push and pull_request (lines 19 to 31). The body: every sentence checks against the head except the one under Smaller points. The pin: `tests/test_docs_workflow_pins.py` fails on the base with four failures and passes at this head (its four tests and the front-pages module's five, nine in all), and each of eight behaviour mutations, among them a path dropped from the pull-request list, the deploy job's event guard removed, a shared group, the level set back to info and codex.md unlisted, turns exactly the named test red. The push run at the merge commit, 35565351340, is green with the upload step and the deploy job both run, so the raised level found no orphan on main.

None of the four points stands. Two items are new at this head: one sentence of the body, under Smaller points, and the shape of three pins in the new module, under Tests.

## Smaller points

- The body's sentence that pushes to main keep the single `pages` group and are never cancelled claims more than the setting does. Read against GitHub's concurrency documentation: `cancel-in-progress` at line 43 of docs.yml evaluates false on a push and protects only the run in progress; by default, and docs.yml sets no queue key, GitHub keeps at most one waiting run per group, and a newer run replaces the waiting one, which is cancelled. Two docs merges to main while a publish is in progress: run A is publishing, run B waits, run C arrives; B is cancelled and C takes its place. The site is unaffected, since C publishes the newest head, but the sentence is false for B, and a reader who relies on it expects every push to main to publish a run of its own. The comment at lines 39 and 40 of `docs.yml` says only that an in-progress publish is not cancelled, which is accurate, so the body is the one place to edit. Fix: pushes to main keep the single pages group; an in-progress publish is never cancelled, and a waiting one is replaced by a newer push.

## Tests

- Three of the pins in `tests/test_docs_workflow_pins.py` are satisfied by a comment, so the guard each names can be removed while the module stays green. The regex at line 35 captures the rest of the `if:` line, comment included: at this head the capture for the upload step is the expression followed by the trailing comment from line 58 of docs.yml, and the assertion at line 37 searches that whole string. The regex at line 38 captures the deploy job's condition (line 65 of docs.yml) the same way, for the assertion at line 40. Lines 43 and 44 search the whole file for the two concurrency expressions, unanchored. Executed: an edit that drops `github.event_name != 'pull_request'` from the upload step's or the deploy job's condition and leaves the old text as a trailing comment on the same line keeps all nine tests green, and so does a shared group written in place of the per-ref expression with the old expression left in the comment above it. In either state a pull request's run uploads the site, or reaches the deploy job when `PAGES_ENABLED` is true, or shares the pages group with the publish, and the module reports nothing. The mkdocs.yml pin at line 50 does not have this shape: its pattern is anchored at line start and end, and a commented-out copy of the warn setting turns it red. Fix: strip a trailing comment from each capture before asserting, with `re.sub(r"\s+#.*$", "", up.group(1))` and the same for the deploy capture, then search the stripped expression. Anchor the two concurrency asserts at line start, an assertRegex over `(?m)^  group: <the expression, escaped>$` and the same for cancel-in-progress. The message at line 36 could also say that the step-then-condition shape was not found, since the regex fails for a missing step as well as a missing condition.

PR #1936, opened 2026-09-21
Docs: the front pages state what romp does, and each stands on its own

henrypinkard:
An editorial pass over the three front pages and the README. No behaviour is described differently; the sentences state what romp does instead of framing it, and each page reads as itself rather than as a correction of an earlier version.

## The three passages named in review

**Install, the paragraph that introduced itself by what it was not carrying.**

Before: a paragraph after the install command saying that two operational details live in the reference rather than here, then both links.

After: one clause under Requirements, carrying the reason a reader would want it: "On a machine with several Pythons, [which one runs the kernel](reference.md#the-kernels-python) matters." The login service's environment is gone from the page: no first install needs it, and the reference still documents it.

**Guide, the message box.**

Before: "Writing to a session is more than a text box. Drop a file anywhere on the pane to attach it, type `@` and a few letters to name another session, and take a message back out of the queue with the pencil in its corner while the session is still busy."

After: "The message box also supports attachments, session names and recall. A file dropped anywhere on the pane attaches to the next message, `@` and the first letters of a session's name insert that name, and the pencil on a message the session has not taken yet puts it back in the box."

**Guide, opening a file.**

Before: "Files open where you are reading. Click a path or a markdown link and the file opens in the viewer, rendered, at a text size you control. Select a passage and it lands in the composer as a quote chip, so reviewing a document is a series of notes rather than lines copied by hand."

After: "A path or a markdown link in the chat opens the file in the viewer, rendered. A passage selected there lands in the composer as a quote, labelled with the file and the line it came from."

## The rest of the pass

- **No trailing links.** Twenty-one links from guide paragraphs into the reference are gone. The reference is named once, in the guide's opening line. Where a link stood in for a fact, the fact is now written out: the three Tailscale app settings, the shared-tailnet warning, and why VS Code's port forwarder is a poor path for the dashboard. No sentence anywhere tells a reader that the details are elsewhere.
- **The pane is called Sessions.** The guide's section heading, its anchor (`#the-sessions-pane`), the views list, the home page's feature line and the README's now say the Sessions pane holds the timeline. "Timeline" stays where it names the chart (its lanes, its arcs, an image's alt text). No heading in the reference names the pane, so the reference is untouched by this pull request.
- **Tone sweep** over the rest of the guide: the outline's rename sentence, the pane-docking paragraph, the settings-across-machines opening, the notifications opening, and the token-usage paragraph each state the behaviour and stop.

## Guards

`CLAUDE.md`'s front-pages section gains three rules: a feature's paragraph carries no trailing link (the detail goes to `docs/reference.md` under a heading matching the feature's name), state what a thing does rather than selling it and never write a page as a response to how it used to read, and call each part of the interface what the interface calls it.

`tests/test_docs_front_pages.py` gains one pin: `docs/guide.md` names the reference exactly once, in its opening. The existing caps stay green (index 448, install 312, guide 3,930 words by `wc`; the test's prose counts are lower).

## Pins

Three pins read links the guide no longer carries, and follow the prose that remains rather than loosening: `tests/test_api_health_hover.py` keeps both guide sentences about the dot and all five reference assertions; `tests/test_files_pane.py` keeps the guide's introduction sentence and the section ordering; `ui/webview/dense-chrome.test.ts` now asserts the guide names no gear label at all, which is the rule the page follows.

## Checks

`tests/test_docs_front_pages.py`, `test_api_health_hover.py`, `test_files_pane.py` and `test_new_session_dir.py`: 116 passed, 146 subtests. The seven extension modules that read the docs: 149 pass, 0 fail. Every internal link and anchor in `docs/` resolves with MkDocs' slug rule. No em-dashes, no identifiers in the added lines.

This pull request touches no line within ten lines of the open Needs-you phase-two changes in `docs/guide.md`, and touches `docs/reference.md` not at all.

Tier: docs


ksarma:
The evidence below was executed against a70bec3d2, the head of this pull request; the merge commit e4e73d14b carries that tree unchanged, so every path and line number here is on that commit.

Most of the description checks out. The new pin `test_the_guide_points_at_the_reference_once` fails on the base (25 lines name the reference, not 1) and passes at the head; the four Python modules named give 116 passed and 146 subtests; the pages count 448, 312 and 3930 words; the facts written out in place of links match docs/reference.md; the renamed heading slugs to the anchor the views list uses, and nothing references the old one. Two items stand after the merge.

- **The spend readout sentence describes a split the panel does not draw.** docs/guide.md:497-498 says the bottom bar's spend readout shows "the same split in dollars, by session", where "the same split" is the previous sentence's sessions beside the judges. The panel that readout opens, API spend, has no judges series or row: `_spend_detail_local` (kernel/kernel.py:45379) reads only spend.json, whose one writer is `_record_spend` (kernel/sdk_backend.py:12481), called only from the session turn's result handler (sdk_backend.py:8768), while the judges log to judge-usage.jsonl (kernel/judge.py:133). Executed at the merge with a synthetic spend.json and one judge-usage row, the detail payload's kind values are sid and unattributed, and the judge dollars appear only in `_token_analytics`, the Token usage panel. A reader opening the readout for the sessions-versus-judges split finds spend by session only. A sentence that matches the panel, and docs/reference.md:5016-5019's account of it (Totals, Spend over time, By session): "The bottom bar's spend readout shows the sessions' spend in dollars, by session and over time." It adds three words under the 5500 cap and no reference link, so the new pin stays at one, and no test reads the sentence; naming the panel ("opens API spend:") would follow the new call-it-what-the-interface-calls-it rule. Off the page, the JS comment at kernel/kernel.py:62099 says the By login line has the judges' calls included; no writer puts judge spend into `byLogin`, and if the comment means which login a judge call bills rather than dollars it does not say so, so a later docs edit reading it could restate the split.

- **The no-trailing-link rule describes a state the pages are not in, and the pin cannot detect the class.** CLAUDE.md:263-269 says a paragraph never ends by sending the reader somewhere else and no page tells a reader that the details are elsewhere; the description says no sentence anywhere does. Three sentences on the governed pages break the letter of that rule: docs/guide.md:343-344 (a one-sentence paragraph, "The mechanics ... are in How Romp works"), docs/guide.md:486-488 (the paragraph ends "Full details ... are in SECURITY.md") and docs/install.md:31 (a paragraph whose last sentence is the bare link "What it installs, in detail", the shape this change removed from the guide). All three were present at a59c0cde3, so the change left them standing rather than adding them; the description's completeness claim is what is false, not the diff. The pin half is confirmed by execution: tests/test_docs_front_pages.py:127-135 counts lines containing "reference.md" in the guide alone, so a scratch guide with an appended "are in [How Romp works](architecture.md)" paragraph passes while a second reference link fails. A session editing these pages therefore reads a rule the pages already break, and the mechanical half cannot say which reading is right. The fix that matches the description rewrites the three sentences in the shape CLAUDE.md:274-275 prescribes, continuing the paragraph past the link rather than ending on it: guide 343-344 joins the preceding paragraph as a clause with the link inside it, or goes, since the architecture page is in the site nav; guide 486-488 becomes a clause in the paragraph it ends; install.md:31's link moves into the sentence before it. Two more the same grep finds are not what the rule as written forbids, docs/guide.md:296-297 (a link to a section of the same page, which can go) and README.md:72 (the Docs section, whose link carries its reason in the same clause); if the rule is meant to cover a same-page link or a navigation line, it needs to say so. Narrowing CLAUDE.md:266-268 to what the pin counts, the guide's links into the reference, is the other consistent reading, but it contradicts the description and leaves three "details are elsewhere" pointers on the governed pages. If the rule stays, its mechanical half is a property pin over all four pages, no prose paragraph whose last line ends in a markdown link followed by at most a period, rather than a wider spelling list; that catches all three as they stand and, being indifferent to the link's target, guide 296-297 as well. The pin module's own word-cap message reads against the rule (under Tests).

### Smaller points

- **docs/install.md:15-16** puts "which one runs the kernel" above the install command at line 21, where CLAUDE.md:260-262 says the interpreter rules are reference material and nothing optional goes above the command while CLAUDE.md:274-275 quotes that very sentence as the exemplar of a link done right, and the position pin caps only the command's line number, at 40, so it stays green either way; either the sentence moves under the Install heading, after "This clones Romp to ~/romp and installs the newest release.", or CLAUDE.md:260-262 excepts a one-line requirement note.

- **docs/guide.md:16 and :90** write "The sessions pane" with a small s while docs/guide.md:92, docs/index.md:35, README.md:14 and the rule at CLAUDE.md:277 write "the Sessions pane", the label the interface uses, so a reader searching the guide for "Sessions pane" misses the section heading; capitalising the heading and the views-list entry leaves the slug unchanged, so no link moves.

- **docs/guide.md:31-32** says "`@` and the first letters of a session's name insert that name", but typing lists the live sessions whose names match (ui/webview/render.ts:20311-20313) and the name goes into the text when a row is picked (`pickMention` at render.ts:20245; docs/reference.md:4563-4566 describes it that way), so a reader who types "@ap" expecting "api" to appear sees a list; a sentence that matches: "`@` and the first letters of a session's name list the matching sessions, and a pick inserts the name".

- **Three counts in the description differ from the head's:** the guide names the reference on 25 lines at a59c0cde3 and on 1 at a70bec3d2, 24 fewer, where the description counts twenty-one (two of the 24 were list items and two shared one sentence, so the figure depends on what counted as a paragraph link); the CLAUDE.md hunk rewrites one bullet and adds three, and the description's list of three omits the link-carries-its-reason bullet; the seven extension modules that read the docs, built from the merge's tree and run in this environment, give 140 pass, 0 fail, against the description's 149.

### Tests

- **ui/webview/dense-chrome.test.ts:238**, the line that replaced the assertion that the guide links to the chat pane's reference section, asserts that the guide lacks the compact-tabs gear label, which the base guide also lacked (0 occurrences at a59c0cde3), so it pins nothing this change made true; the description and the commit say it asserts that the guide names no gear label at all, and the guide does name settings in bold (docs/guide.md:126, 385, 496), so a reader believes a property is pinned that the page does not have. Either drop the line, since `test_the_guide_points_at_the_reference_once` already pins the guide's one reference link, or pin the positive the change made true: the guide no longer contains "reference.md#the-chat-pane-in-detail".

- **tests/test_docs_front_pages.py:11** says the detail lives in the reference "behind a link", and the word-cap message at lines 104-105 tells an editor to move detail to the reference and "leave a paragraph with a link", so a session that trips the cap follows the message and turns `test_the_guide_points_at_the_reference_once` red in the same run; wording that agrees with the rule is "the detail lives in docs/reference.md under a heading matching the feature's name" and "move the detail into docs/reference.md and leave a paragraph stating what the feature does", and CLAUDE.md:282-283, which lists the word budget, the paragraph cap, the install command's position and the nav rule without the reference pointer, gains "and the guide's single pointer to the reference".

PR #1937, opened 2026-09-21
The judge experiment harness cuts every store copy to the ending's transcript, prefers endings the judges completed, and labels as a subcommand

henrypinkard:
The harness (scripts/judge_experiment.py, #1932) after the pilot's stop: every store copy is cut to the ending's transcript, the corpus selection prefers endings the judges completed so tier one has data, and the labeller is a subcommand. Fix, red first at main.

**The defects the pilot found** (2026-09-21; 2.44 dollars of labeller calls bought them). The goal stores' verdict log events carry `ev_t` (the evidence time) and `at` (the arrival), never `t`; the copy filter read `t`, so it kept every event and each store copy carried the live judges' verdicts from after the cut, and an arm would have measured placements the live judges had already made. The closer's `closedTurns` and `closedSig` and the planner's `placements` carried the future the same way. And the newest-first selection drew endings without a top-level closer or planner `done` in the turn's window, so the user's later card actions (tier one) could label none of the sixty.

**The fix.** `event_time` reads `ev_t`, then `at`. Turn ids and segment ids carry their epoch second after the fsid (`<fsid>:<epoch>:<hash>`, a placement key with a phase suffix), so `id_epoch` cuts `closedTurns`, `closedSig`, `placements` and each node's `trail` to what the truncated transcript holds; the node flags and settled fields (`nodeComplete`, `blocked`, `cleared`, `doneWhy`, `settledAt`, `settledDone`, the summary fields, `closerLookT`) are dropped for the arm's own rollup, and `mt` is capped at the cut. The builder records each ending's turn start (`startT`) and marks `tierOneEligible` when the live store shows a top-level closer or planner `done` inside the turn's own window (its start, or fifteen minutes back when the transcript shows none, to two minutes after the cut, when the closer files); within every class the eligible endings are picked first, newest first per session, round-robin across sessions, then the rest. The `label` subcommand: tier one from the live journals (a followup, unclear or restore on a card completed in the window says not finished; a user clear or a resolve with none of those within seven days says finished), tier two twice per ending with the classes in two orders (the label is their agreement), the agreement with tier one on every ending that has both against the 90 percent gate, `labels.json` and `labels-summary.json` under the run root, every call's cost on `labeller-ledger.jsonl`; a run root inside a git checkout is refused.

**Tests** (tests/test_judge_experiment.py, the fake claude now answering the labeller too): a live store with a done two seconds after the cut, a block an hour later, a future trail segment, a future closed turn and a future placement gives a copy holding only the mint, the pre-cut trail, turn and placement, no flags, the status empty, `mt` capped; a live store with a top-level done in the first ending's window makes that offer the one picked over a newer offer the judges never ruled on, marked eligible, every ending carrying the mark; the label pass reads not finished from a followup after the done and finished from a clear, agrees with tier one on both, passes the gate, is stable in both orders, puts two calls per ending on the ledger, and refuses a run root inside a checkout. Red first at main by copy-aside, three reds on assertions: `['mint', 'done', 'block'] != ['mint']`; `('…', 0, None) != ('…', 0, True)` on the eligibility mark; `unexpectedly None : the labeller is a subcommand of the harness`.

**Suites at the head** (1c0743fe): the harness module 8 passed under the pinned runner. No file under kernel/, ui/ or vscode-extension/ changes.


ksarma:
The evidence here was executed against 1c0743fe3, the PR's head. The merge carries both files unchanged, so every line reference reads the same.

Of the store-copy and measure points in the comments on PR 1932 (https://github.com/romp-on/romp/pull/1932#issuecomment-5754956121), the change closes one, the copy filter's key (`ev_t`, then the arrival); the boundary, the seed's live session id and the untouched measures remain.

- **The ending's own turn survives the copy** (scripts/judge_experiment.py:166 and 174 to 189). The filter keeps rows with `event_time(e) <= cut_t`, but the closer stamps a done's evidence time at the turn's start and the planner at the segment's start, so every verdict on the turn under judgment passes and the rollup re-derives the dropped flags from it. The selection preference concentrates on these endings. Fix: cut strictly before the turn's start on the log rows, the birth time and `trail, placements, closedTurns, closedSig`, keeping the window's inclusive start. The bound is the event model's turn start, not `turn_start` at line 142, the last typed record: an absorbed mid-turn input puts that after the closer's row, a turn with no typed opener in an earlier turn; derive it as the last typed record after an assistant record with a terminal stop reason, or name both shapes as a residual. A lift rider's ruling, anchored to older history, survives any evidence-time cut and stays a residual (cutting on arrival would drop the closer's backlog on earlier turns). Re-keying the seed (below) without this cut widens the leak to every eligible ending.

- **Tier one matches no journal row on a real store** (line 532). The compare reduces the row's node to its tail after the last colon, but minters key nodes as `<rompUuid>:g<N>` and journal writers record that full key, so tier one is None for every ending, the pilot's zero labelled. Fix: `o.get("node") == nid`; re-key the fixture at tests/test_judge_experiment.py:230 and 235 to the minter's shape, turning the label test red at the head.

- **The window's tail admits the next turn's done** (`in_turn_window`, line 155). No closer or planner done carries an evidence time after its own cut, so a done in the two minutes after the cut is the next turn's, and that turn's card gestures label this ending. `lo <= t <= cut_t` fixes it. The fallback's defect is the opposite (with `turn_start` None no earlier ending exists): a triggerless turn over fifteen minutes reads ineligible; bound it at the transcript's first record, or mark the ending ineligible. A deeper repair keys on the closer's `endEv` and the planner's segment epoch; short of that, call the window a bounded approximation at lines 41, 42 and 506.

- **The unblocker's watermarks survive the cut** (line 170). The drop list leaves `blockCheckT, blockCheckDoneT, delegLookT`, and the unblocker's due gate reads the first two, so a stamped node draws no unblocker call in an arm whose closer files no done, and keeps the live column. Clamping does not close it: the gate is strict, `newest > max(bt, blockCheckT)`, with newest the turn's start, at or before the cut. Drop `blockCheckT`, or keep it only when before the manifest's turn start; the other two are in the arrival domain, where drop or clamp works. Clamp, do not drop, `titledT, servingT`: dropping the first buys a title call per node per build. The earlier PR 1932 comment's gate-stamps item says clamp or drop; for the first stamp only the drop works.

- **Cleared tops are invisible** (lines 240, 245, 277 and 513). The kernel's compaction moves every cleared top with its verdict log into `goals-archive/<sid>.json`, and the harness reads only the live store, so a cleared top is missing from the done times, the eligibility mark, tier one and the seed; the plan doc counts 640 clears against 22 resolves, so the gate is measured on one side. Fix: read the archive as a file beside the store and union its tops into the store all four readers take, and split the seed by the clear's time: a node cleared after the cut returns to the live seed, one cleared at or before it goes to the corpus's archive file, or the rollup misfiles it.

- **Per-session order is oldest first** (line 253). `list(reversed(v))` there and `q.pop()` at line 259 take the session's earliest ending while the comment and the body say newest first. `q.pop(0)`, or dropping the reversal, picks the newest.

- **The two PR 1932 points left open** (lines 182 and 427). The seed keeps `rompUuid` and every placement and trail key under the live session id while the arm parses under the ending id, so the planner re-plans the earlier turn; and the measures charge a leak or false interrupt for any top read completed or needing input, whatever turn produced it. Re-key by one text replacement of the sid with the ending id over the serialized copy, journal node fields included; alone that gives zero planner calls, since the ending turn's own placement survives, so also drop placements at or after the turn's start, keeping the ending's `#p` key and its node: exactly one planner call per ending. For the measures, count only closer or planner done or block verdicts, by evidence time, inside `in_turn_window(t, startT, cutT)`, and only with the turn-start cut, recording each top's birth and verdict times per build, since results.json holds only final columns. `topsBefore` at line 294 writes the raw node keys too; hash or suffix them.

- **The restore op is dead** (`NOT_FINISHED_OPS`, line 502). The kernel's restore row carries nodes and no node field, so the match at line 532 never selects it; the ordinary undo also writes a paired unclear row, so only a restore without its twin (a faulted flag step, an older unflagged archive node) mislabels. Match it by `nid in (o.get("nodes") or {})` behind an isinstance dict check, or drop the op, noting the unclear row carries the gesture.

- **The tier label.** The labeller block (lines 494 to 615), most of the added lines, is a new two-tier labelling subcommand, which docs/pr-tiers.md routes to feature; an admin author merges on green at either tier, so only the record undercounts.

## Smaller points

- **The second labeller order can equal the first** (line 602): a shuffle of four classes is the identity one time in 24, so those endings get the identical prompt twice; use the reversed class list or redraw until the order differs.

- **The docstring's guarantee is wider than the code** (line 163): "Nothing from after the cut survives" is false for `seams, closeFails, confirming, groupedSig, consolidatedSig, rewindSwept, lastNode, seq, summaryQuote, summaryAnchors, summaryQuoteOff, titledT, warns`; a seed re-keyed to the ending id makes the inherited id-keyed records live; land the two together.

- **The labeller's input is narrower than the plan says** (line 542): plans/judge-prompt-experiments.md line 34 has the labeller reading the ending's last turn and the card's title; `_last_assistant_text` sends only the final assistant text's last 6000 characters. Widen the input to the open tops' titles and the typed ask (`LABEL_SYS` too) or change the plan's line.

- **A faulted store reads as "nothing applies"** (line 514): `except (OSError, ValueError)` returns None, the same as an empty journal, so one corrupt store halved the labelled count unseen; record the fault on the row (a `tierOneError` with the session hash and exception class, never the sid or path) and count faulted sessions in the summary rather than raise; likewise at line 238.

## Tests

The store-copy fixture's done row two seconds after the cut (tests/test_judge_experiment.py:233) is a shape no closer writes: replace it with a done at the turn's start arriving after the cut, pin red-before on the copy's log and assert the arm's column is the fake candidate's. The selection test at line 274 passes a mutant without the eligibility filter (put the eligible ending alone in the second session; both in one session with the older eligible stays green), and the agreement gate at line 305 is pinned only at 100 percent (a disagreeing third ending needs a second top, since `_live_store_with_done` mints one per sid). The label test at line 296, the drop list at line 255 (pin it by an unblocker call, not key absence) and the eligibility and order pins at line 306 survive mutants too; each wants writer-shaped fixtures, one assertion per rule.

PR #1938, opened 2026-09-21
The reference's builds.feed.memo passage names failed, failing, coldLive and coldFlip, so a standing coldLive count no longer reads as a fault

lczh:
Tier: docs

The reference's builds.feed.memo passage backticked nine of the thirteen keys the memo report returns under GET /perf and never named `coldLive`, `coldFlip`, `failed` or `failing`. Since #1820 the two cold counters have had no line anywhere in the docs, and `romp perf` does not print the memo block, so the passage is the one gloss a reader has. `coldLive` counts, per session per build, each living session not hidden from the feed, with a transcript, whose cache-only parse read missed; a session no client and no judge has parsed rides it every build, so a standing count is by design. With no gloss, that standing count read as a fault.

What changed:

- A passage (three sentences) after the `row_by` sentence names all four with what each counts, in the words of the kernel's own comments: `failed` the card builds that raised (a memoized entry's decode, the key, the derivation, its dependency key, the serialization or the memo put), cumulative, which is the exact span of the per-session guard in the feed build; `failing` the sessions whose last build did, a standing fault rather than history; `coldLive` as above; `coldFlip` the subset the memo held warm and re-read in place with one kernel parse (also under `parses.kernel`) instead of deriving cold, with the watch line that `coldFlip` climbing every build for one session with no appends means its parse never stores, which should not occur.
- A pin in the perf-stats module's Collector class, beside the chatFullWhy gloss test, slices the passage from the `memo` phrase to the `sends` bullet and asserts every memo counter is backticked there. The counters are the runtime report's keys unioned with every literal the kernel's source passes to the memo counter helper, so a counter minted at a site the test process never exercises still has to reach the doc; a floor of thirteen keys keeps a shrunken report from passing vacuously. At the base the union adds no key the report lacks. The slice stops at the `sends` bullet so a `failed` mentioned elsewhere in the `builds` bullet cannot satisfy it.

No behavior changes: prose plus a test-only pin.

Validation:

- Red first, the pin alone against the base doc: `AssertionError: Lists differ: ['coldFlip', 'coldLive', 'failed', 'failing'] != []` (1 failed, 65 deselected).
- The perf-stats module with the passage in place: 66 passed (65 at the base).
- The doc's other reader, the perf heap-block module: 13 passed. The memo's driver, the feed session-memo module, which pins the report's key set from the runtime object and does not read the doc: 54 passed, 13 subtests passed.

Follows the post-merge review of #1820.


ksarma:
The evidence was executed against 558e5be37, the PR's head, before it merged as d0e3a68b7; the merge carries that tree unchanged for every file cited, so each file and line reference is on that commit and holds at the merge commit.

The change adds three sentences to the memo passage of docs/reference.md (lines 2534 to 2545) naming the four counters the passage had left out, and one test in tests/test_perf_stats.py (lines 447 to 463) that slices the passage from the memo phrase to the sends bullet and asserts that every memo counter appears backticked in it. Read against kernel/kernel.py at this head, the passage's statements about the counters are accurate. `failed` increments once per raise of the single try block in the feed build's loop (43719 to 43740), which spans the memoized entry's decode, the key, the derivation, the dependency key, the serialization and the memo put, counted at 42022 and never reset. `failing` is the size of the dict of sessions whose last build raised, popped when a build serves or derives the session (42043 to 42046) or the session departs (42007 to 42008). `coldLive` is counted at 42425 only when the hide flag is off, the session has a transcript path and the cache-only read returned None, once per session per feed build. `coldFlip` at 42453 sits inside that branch behind the warm-key check. Its parse goes through the shared store and lands under the kernel parse counter, so the passage's three statements about it (the subset, one kernel parse, also under the parses block) are right, and the watch sentence is a faithful compression of the comment at 41898 to 41902.

The body's test results reproduce: the module is green at the head (66 passed; 65 at the base), and with the head's test file over the base's doc the new test is red with exactly the message the body quotes. Three points are still standing after the merge, each about wording or the pin's reach rather than the kernel.

## Smaller points

- docs/reference.md lines 2540 to 2541 (by reading). The sentence says a session no client and no judge has parsed rides `coldLive` every build. That names two parsers as the deciding condition, and at this head the kernel warms a session neither of them touched by at least two other roads. The background warmer (the loop at 35136 to 35145, gated by `_warm_wanted` at 11557) parses every living session that moved since boot or is working, whenever a client is connected and no chat or timeline client is. The feed build's own re-read at 42453 parses a warm-keyed session through the shared store, so the next build's cache-only read hits. So the standing population is narrower than the sentence says. A feed-only dashboard sees a moved session leave the counter after one build with no chat tab and no judge run, and sees an unmoved idle session ride it while a moved one does not; the sentence predicts neither, and telling design from fault is what the passage is for. The kernel's own comment at 42427 to 42428 states the condition as a session nothing has parsed. Saying the deciding condition rather than two of the parsers closes it: a session nothing has parsed at its current version (no client's tab, no judge, no background warm, none of the card build's own parse paths) rides it every build, and the warm gate leaves an unmoved, idle session cold by design.

- The body's validation paragraph (the passage it describes starts at docs/reference.md line 2534; by reading). It calls the feed session-memo module one of the doc's other readers, but tests/test_feed_session_memo.py never reads docs/reference.md at this head; its test at lines 885 to 899 pins the report's key set from the runtime object, in the set literal at 890 to 893. The counts are right (54 passed, 13 subtests). A maintainer reading the body to learn which tests pin the passage would run that module expecting it to go red on a doc regression, and it cannot; of the two modules the body names, the perf heap-block module (its test at line 429 reads the doc at 430) is the one that reads it. Splitting the sentence fixes it, in the body alone, with no commit: the other doc reader, the perf heap-block module, 13 passed; the memo's driver, the feed session-memo module, 54 passed, 13 subtests.

## Tests

- tests/test_perf_stats.py line 459 (by reading). The docstring at lines 453 to 454 says a counter minted at a site this process never exercises still has to reach the doc, and that this is what makes the claim true by mechanism. The source half of the union at 459 scans only the string literals passed to `_feed_memo_count`. The stats dict is also written directly: `_feed_derive_complain` at 42022 increments the failed key with a get-plus-one on the dict rather than through the helper. That key is seeded in the initializer at 41891, so the runtime half of the union carries it and the pin is right today. A future counter written in that shape without a seed, at a fault or eviction site the test process never drives, reaches neither half: GET /perf grows a key the reference never names, and the pin whose docstring says it catches exactly that stays green. Either of two changes is enough. Widen the source scan to the dict's literal subscripts as well, for example by adding the matches of `_FEED_MEMO_STATS\["(\w+)"\]` over the source and dropping the two map keys, miss_by and row_by. Or narrow the docstring to every literal the kernel passes to the counter helper and drop the by-mechanism sentence.

PR #1939, opened 2026-09-21
The chat-build census counts a read of a dropped row field as a string constant equal to it, so a kernel comment naming the field no longer fails the pin

lczh:
Tier: docs

The chat-build signature's row-projection pin holds that no chat reader reads the two fields the projection drops, the task row's lastTool and the row's raw ctxTokens. Its transitive backstop counted word-regex mentions of each field over the whole kernel source and required the count to equal the mentions inside the permitted regions (the projection block and the key, plus the compaction tick and the merge that writes ctxTokens). A mention is not a read: a comment or a docstring naming either field, anywhere else in the kernel, failed the pin with the same message a real reader would, and since both are live row fields that kernel prose has reason to name, this recurs. Two PRs were green alone and red together over kernel prose, and a third reworded two kernel docstrings to satisfy the count instead of the test.

The fix is test-only. The census now walks the parsed kernel source and counts every string constant equal to the key, on any receiver. The counted shapes are the whole claim: the argument of a get or a pop, a subscript's slice, the left operand of a membership test, an entry of a tuple a loop reads through, and a real subscript inside an f-string. A read that renders the field through a format placeholder or names it as a keyword argument is not a constant equal to the key and is not counted; the kernel holds no such read today, and the helper's docstring states the blind spot. A comment is not in the tree, and a docstring or a sentence literal that mentions the field is a constant that does not equal it, so prose counts zero and a counted read counts one. String literals are deliberately not stripped first: the key of a real read is itself a string literal, and a census that stripped them would go blind to the reader it hunts. Every hit must fall inside a permitted line range taken from the functions' own source lines; the two existing messages stand, and a failure now reports the offending line numbers. The readers' own-source pin uses the same census, so a docstring in a reader is no longer a hit either. Nothing in the kernel changes.

Validation

- Red first: the new census test, run against the old word-regex census on a scratch copy, flagged the synthetic body's docstring and comment lines as reads of lastTool, and with that part removed flagged the comment appended past every permitted range of the real kernel text (one line per field). Under the read census the same test is green, and the single-quoted get appended the same way is still exactly one hit, so the census is not blind to the reader it hunts.
- The existing row-projection test stays green at main with its messages unchanged: no read of lastTool outside the projection and the key, no read of ctxTokens beyond those, the compaction tick and the merge.
- The module: 66 passed, 10 subtests passed. Test-only, so no kernel pin scan or extension suite is at stake.

Follows the post-merge review of #1817.


ksarma:
The evidence was executed against 236d1d2c0, the PR's head, before it merged as 210747500; the merge carries that tree unchanged for the test file and for kernel/kernel.py, so each file and line reference is on that commit and holds at the merge commit.

The change is confined to tests/test_chat_build_sig_inputs.py (98 insertions, 21 deletions; the kernel is byte-identical between the base and this head). The row-projection pin's transitive backstop no longer counts word-regex mentions of the two dropped fields over the kernel text. `_key_reads` (line 338) walks the parsed source and returns the line numbers of every string constant equal to the key, on any receiver. `_reads_outside` (line 361) keeps the hits that fall in none of the permitted ranges. `_unkeyed_field_permits` (line 367) derives those ranges from the functions' own source lines. The readers' own-source check (lines 453 to 458) goes through the same census, and the new test at line 471 pins the census on a synthetic body and on the real kernel text with a comment and a single-quoted get appended past every range.

The body's validation reproduces. The module is green at this head (66 passed, 10 subtests passed). At the base, appending two comment lines naming the fields to kernel/kernel.py turns the old pin red (4 != 3 on the first field); the same kernel edit with this head's test file is green. Real reads appended past every range, a get of one field and a subscript of the other, turn the row-projection pin red with the offending line number and the new test red as well. On the kernel text the census's hits are exactly four lines, kernel/kernel.py 13093, 20854, 34304 and 34305 (20854 carries two constants equal to the field and the set lists the line once), and each lies inside a permitted range; the other word mentions (34295, 34297 and 34419) are comments. The documented blind spots are real as stated: a keyword argument, a percent-format key and a format field naming either field, appended to the kernel, pass both pins, and the kernel has no such shape today. Replacing the helper with the old per-line word count turns the new test red on the synthetic body's docstring and comment lines, so the test is load-bearing against the defect it describes; a census that returns an empty set leaves the row-projection pin green and is caught by the new test alone.

Two points are still standing after the merge, both in the test file and neither about the kernel.

## Smaller points

- tests/test_chat_build_sig_inputs.py lines 467 and 469 (by reading, and seen on an executed red: a real read appended past every range fails the pin with the line number under this wording). The two failure messages still say a field "is mentioned" outside or beyond the permitted regions, while the assertion above each now fails only on a string constant equal to the field. A red therefore names the wrong thing to remove: a kernel change that adds a get of the raw token count in a helper prints that the field is mentioned beyond the projection, and whoever reads that at face value rewords nearby prose, which is what #1825 did under the old count, and the red stays. The messages already print the offending line numbers, so the wording can match the assertion: "is read outside" at 467 and "is read beyond" at 469.

## Tests

- tests/test_chat_build_sig_inputs.py lines 342 to 346 and 479 to 493 (by reading, with the five claims themselves checked by execution). The `_key_reads` docstring and the PR body claim one counted shape, a real subscript inside an f-string, and four not-counted shapes. Those are a percent-format key, a field named in a `.format` call, the same through its mapping variant, and a keyword argument. The synthetic body at lines 480 to 489 exercises none of the five. It covers a get, a subscript, a membership test, a tuple a loop reads through, a comment, a docstring and a sentence literal, and the two assertions at 492 and 493 pin exactly those. The five claims are therefore docstring text with no executed assertion behind them. The f-string case is the one with a moving part: CI runs this module on 3.10 through 3.13, and a release that changes where f-string sub-expression nodes are positioned could drop or misplace such a read while the pin stays green and the docstring still says the shape is counted. By execution on 3.10.20, 3.11.15, 3.12.3 and 3.13.14, the interpreters in that matrix, the walk the helper performs, run over a short snippet, counts `f"{tm['ctxTokens']}"` at its own line and counts none of the percent key, the format field, its mapping variant or the keyword argument. So the five claims are true today on every interpreter CI runs, and what is missing is the assertion that would notice when one stops being true. The fix is one line of each kind in the synthetic body, the f-string above and a blind-spot line such as `f = "%(ctxTokens)s" % tm`, with the two sets at 492 and 493 extended so the f-string line is in the set at 493 and the percent line is in neither.

PR #1940, opened 2026-09-21
is_lazy and the lazy body's equality key on the source slot, not the class, so a sentinel built before a module re-execution still reads as lazy and equals its twin, and != answers the opposite of ==

lczh:
Tier: fix

The module loader re-executes the event model into the same module object at every import, rebinding the lazy body class. #1928 moved the first hydrate loop's test from the class to the body's source slot for that reason, and its comment declared the slot the module's test, but two other sites in the module stayed class checks against the name: the `is_lazy` predicate and the lazy body's `__eq__`. After a re-execution both resolve the name to the new class, so a sentinel built before it answered False to `is_lazy` with its marker still present, two old sentinels of one uuid compared unequal, and an old one equaled a new one of its uuid while the new one did not equal it back. The product re-executes the module only at import, before any body exists, and `is_lazy` has no product caller, so the reach is a test module that imports the name after another built bodies; the worth is closing the contradiction the earlier comment left, so the next reader does not take the class check as the rule.

This PR also carries a second gap in the same equality contract, present at main with no re-execution involved: the class defined `__eq__` but not `__ne__`, and the dict base's own `__ne__` sits before object's in the method lookup, so `!=` never consulted the class's `__eq__`. It compared the two storages, each holding its own unhydrated marker object, and a same-uuid pair answered both `==` True and `!=` True. It rides with the slot rekey because `==` and `!=` are one protocol: the symmetry the rekey restores cannot be pinned on the `!=` side without it, and the module's other custom-equality container subclass (the lazy assembly list) already pairs the two for the same base-class-lookup reason, which left the lazy body the outlier. Its red is reproducible at main by a standalone test that rebinds nothing.

The fix: `is_lazy` tests the message for the `source_path` slot, and `__eq__` tests the same slot plus a uuid match, each with a dated comment naming the loader mechanism; `__hash__` already keyed on the uuid with no class in it and is unchanged. The class gains `__ne__` returning the negation of `__eq__`, with a comment naming the lookup order. A plain dict and None have no slot and answer False as before; the unhydrated marker carries a uuid alone and stays unequal. The entry-gone arm's comment in `hydrate` now names both causes of a gone entry (self-eviction under a cap the record does not fit under, and eviction by later puts from any thread between the peer's put and its fill) and says the two predicates key on the same slot.

Validation
- Red-first: three new tests in the lazy-source lifetime module beside the existing rebound-class test, each run with the event model at main's text.
  - `test_is_lazy_answers_for_a_sentinel_of_a_rebound_class`: `is_lazy` of a bound lazy atom under the module's class name rebound to another class; red at main, `AssertionError: False is not true`.
  - `test_two_sentinels_of_one_uuid_compare_equal_across_a_rebound_class`: the rebound name is a real second class, built by executing the module's own class statement again over the helpers it names, so an old-class sentinel and a new-class sentinel of one uuid meet; both equality directions, both inequality directions, membership in each other's one-element list, and equal hashes are asserted, alongside two old-class sentinels; red at main on the first equality assert, `AssertionError: False is not true`. Its `!=` assertions pin symmetry under the rebinding once `==` holds.
  - `test_not_equal_answers_the_opposite_of_equal_for_two_sentinels`: two same-uuid sentinels of the module's own class, no rebinding; red at main on `a != b`, `AssertionError: True is not false`.
- Mutants each test now kills, one at a time with the product line replaced and the file restored byte-exact after each:
  - `__eq__` keyed on class identity (`type(other) is type(self)` plus the uuid): red in the cross-class test at its first old-versus-new equality assert, `AssertionError: False is not true`; the earlier cut of the test built every sentinel from the module's own class and let this survive.
  - `is_lazy` keyed on the lazy marker (`atom.get("lazy") is not None`): red in the predicate test on a plain-dict body under a marker still present, the mid-fill window, `AssertionError: True is not false`.
  - `is_lazy` wanting marker and slot both: red in the predicate test on a lazy body with no marker, a shape the product never builds since the marker pop is the last write, `AssertionError: False is not true`; the earlier cut of the test let both marker variants survive.
- Green after, one module at a time: lazy source lifetime 18 (15 at main plus the three); lazy body readers 6 (`is_lazy` keeps its roster slot); asm checkpoint 91 (85 subtests); asm index 26 (3 subtests); asm mat lru release 9; chat pages 43 (27 subtests); kernel 460 (12 subtests).
- No source-text pin covers either replaced literal; the extension suite was not run since the kernel module is untouched.

Follows the post-merge review of #1928.


ksarma:
The change moves two class checks onto the slot the first hydrate loop already tests: `is_lazy` reads the message for its source slot, and the lazy body's equality reads the same slot plus a uuid match, with a negation method added so the two comparison operators agree. The evidence here was executed against 6916af4aa, the PR's head; every file and line reference is on that commit.

The head passes the lifetime module's 18 tests and the readers module's 6. Two of the new tests leave the rule they name unpinned by execution: a rewrite of the predicate each covers survives the test unchanged. Neither is a wrong answer the product reads today, since `is_lazy` has no product caller and the module is re-executed only at import, before any body exists; both points are about what the tests prove.

### Tests

- `tests/test_lazy_source_lifetime.py:335-359`, the equality test across a rebound class. The test builds all three sentinels from the module's own class before the patch and never instantiates the stand-in. The stand-in at line 344, `type("_LazyBody", (dict,), {})`, has no source slot and no uuid, so an instance of it could not pin the direction either. The test therefore never compares a sentinel of the old class with one of a new class in either direction, and that pair is the case the comment at `kernel/event_model.py:4090-4093` describes.
  - Executed: with the equality method at line 4094 of that file replaced by a class-identity check, `type(other) is type(self) and other.uuid == self.uuid`, this module and the readers module pass (24 passed), and so do the other modules the PR body lists apart from the kernel module (169 passed across asm checkpoint, asm index, asm mat lru release and chat pages). Under a real re-execution through the `load_source` the asm checkpoint test module imports (same module object, class rebound), that mutant answers False for both equality directions, True for both inequality directions and False for list membership, for two sentinels of one uuid and path built on either side of it. The head answers True for both equality directions, False for both inequalities, True for membership each way, equal hashes, and True from `is_lazy` for both. The base answers equal in one direction and unequal in the other, the asymmetry the comment names, and its predicate answers False for the old sentinel.
  - One bound on the failure statement: a class-name key, `type(other).__name__ == "_LazyBody"` with a uuid match, also passes the 24 tests but compares old and new equal under a real re-execution, because the new class carries the same name. A class-name key is therefore not the defect the comment describes; only a class-identity key produces the unequal pair.
  - Fix: build the second sentinel from a real second class, since the existing stand-in cannot be made to pin this direction by instantiating it. The smaller change is to exec `inspect.getsource(em._LazyBody)` into a namespace that contains the unhydrated marker class, the unbound-source sentinel and the read error class. The other is to call `load_source("romp_event_model", os.path.join(BIN, "romp-event-model"))` imported from the asm checkpoint test module, the loader mechanism itself. Then assert both equality directions, both inequality directions, membership in a one-element list each way, and equal hashes. Both variants pass at the head and turn the class-identity mutant red at the first equality assert (2 failed, 18 passed). `inspect.getsource` works on the class of the path-loaded module because the bin script resolves to the kernel module file on disk. The loader variant, run first in the class, left the module's other tests, the 91 asm checkpoint tests and the 6 reader tests run after it in one process green (117 passed); it does reset module globals mid-test (the checkpoint dir, the memo, the stats, and the caches the harness's `fresh` clears), which this test never reads and the harness's setUp and tearDown set again for every test, so it is safe here.

- `tests/test_lazy_source_lifetime.py:318-333`, the predicate test. Its assertions do not separate the body's source slot from the atom's lazy marker: every assert at lines 327-333 is satisfied by a marker-keyed predicate as well as by the slot-keyed one at `kernel/event_model.py:4113`. The docstring's statement that the predicate keys on the body's source slot, and the rule at lines 7193-7194 of the same file that readers key on the body type, not the marker, are therefore not pinned by execution.
  - Executed: a mutant reading `return atom.get("lazy") is not None` passes the lifetime and readers modules unchanged (24 passed). The predicate has no other caller: the only calls in the tree are this test's lines 327-333, and the readers module lists the name in a static roster at its line 42 without calling it, so no other module can turn the mutant red. A second mutant also survives the test as written: `return atom.get("lazy") is not None and hasattr(atom.get("message"), "source_path")`. The disagreeing shape is a real window: `_hydrate_one` at lines 7088-7122 sets the message and pops the marker as its last write, so a plain-dict body with the marker present exists between the two writes. The mid-fill helper this test module builds at line 43, `_PeerMidFill`, is that window. The predicate has no product caller, so the consequence today is the unpinned rule, not a wrong answer anyone reads.
  - Fix: add one assert per disagreeing shape to the test. A plain-dict body with the marker present answers False: `self.assertFalse(em.is_lazy({"uuid": "user-web", "lazy": {"k": "u"}, "message": {"role": "user", "content": "prompt for web"}}))`. A lazy body with no marker answers True: an atom whose message is `em._LazyBody("user-web", p)` and which carries no lazy key. With both added, the marker mutant is red at the first, the conjunction mutant passes the first and is red at the second, and the head passes both.
  - Two details on that fix. `p` is not defined in this test. Define it as the two equality tests do at lines 342 and 366, as `p = str(self.td / "web" / (G.FSID_A + ".jsonl"))`. Or take the slot from the restored atom's message at line 326, `atom["message"].source_path`. And a lazy body with no marker is not a shape the product builds, since the marker pop is the hydrate step's last write, so that assert pins the rule rather than a product window and the docstring is the place to say so; the plain-body-with-marker assert does correspond to the mid-fill window.

lczh:
Executed against 2137948d, the new head. Both test gaps are closed in the same commit; the product lines are unchanged from the reviewed head.

- The equality test now builds a real second class by executing the lazy body's own class source into a scope holding the three helpers it names, asserts it is not the module's class, and compares an old-class sentinel with a new-class one in both directions: equal both ways, not unequal either way, membership in a one-element list each way, equal hashes. The class-identity mutant is red at the first old-versus-new equality assert (1 failed, 17 passed); a class-name key would pass, as you noted, and is not the defect the comment describes.
- The predicate test adds the two disagreeing shapes: a plain-dict body under a marker still present (the mid-fill window) answers False, and a lazy body with no marker answers True, with the docstring saying the second shape is not one the product builds and pins the rule alone. The marker-keyed mutant is red at the first (1 failed, 17 passed) and the marker-and-slot conjunction at the second (1 failed, 17 passed).
- At main's text the rewritten module reads 3 failed, 15 passed (the predicate under the rebound name, the cross-class equality, the standalone inequality). Green at the head: lifetime 18, readers 6, asm checkpoint 91 with 85 subtests.

ksarma:
The evidence here was executed against 2137948d0, the PR's head, merged as 2b84a6a59. Every file and line reference is on that commit; the merge carries both changed files unchanged, so the lines read the same there. The kernel file is byte-identical to the head the first comment reviewed; the difference is the test file alone, +33 -9.

The commit closes both points from the first comment. Each mutant was applied to a scratch copy of the kernel file and the file restored byte-exact after each run:

- The equality test at `tests/test_lazy_source_lifetime.py:344-383` builds a real second class by executing the source of the module's own class into a scope that supplies its three free names (356-359), then compares an old-class sentinel with a new-class one in both directions, both inequality directions, membership each way and hashes (364-369, 372, and 382 after the patch lifts). The class-identity mutant, `type(other) is type(self)` plus the uuid, is red at line 364, the first old-versus-new equality assert, 1 failed, 17 passed. A class-name key plus the uuid passes all 18, as the comment on this head notes, and is not the defect the comment at `kernel/event_model.py:4090-4093` describes. Main's class check restored with `__ne__` kept is red at 362.

- The predicate test at lines 319-342 adds the two disagreeing shapes, with `p` defined at 341. The marker-keyed mutant, `atom.get("lazy") is not None`, is red at line 340, the plain-dict body under a marker still present (True is not false); the marker-and-slot conjunction is red at line 342, the lazy body with no marker (False is not true); each 1 failed, 17 passed. The docstring at 320-328 says the second shape is not one the product builds and pins the rule alone, and the code agrees: the hydrate step pops the marker as its last write at `kernel/event_model.py:7122`, and the unhydrated marker declares a uuid slot alone at 4055.

- Main's text: this head's test module over the kernel at a59c0cde3 reads 3 failed, 15 passed, red at 335 (the predicate under the rebound name), 362 (the first equality assert) and 393 (`a != b`), with the error texts the body states.

- Green at the head, one module at a time: lazy source lifetime 18; lazy body readers 6, the predicate still in the roster at `tests/test_lazy_body_readers.py:42`; asm checkpoint 91 with 85 subtests; asm index 26 with 3; asm mat lru release 9; chat pages 43 with 27. The kernel module's 460 was not re-run: the event model file is unchanged since the first comment and the PR does not touch `kernel/kernel.py`; the extension check passed at the head.

- The body's changed sentences read true at this head. The asymmetry clause matches main's `isinstance` key, True from the old side and False from the new. No source-text pin under tests, ui, vscode-extension or scripts names either replaced literal. `is_lazy` appears in the kernel file only at its def and in two comments, and elsewhere only in the two test modules. The lazy assembly list class at line 5293 defines its own `__ne__` at 5407, as the body says.

Nothing stands after the merge.

PR #1941, opened 2026-09-21
The feed's warm-to-stale re-read runs only for a transcript the kernel can stat, so a gone leaf derives cold once instead of re-parsing every build and painting the card quiet

lczh:
Tier: fix

The feed's warm-to-stale re-read runs only for a transcript the kernel can stat, and a session whose transcript cannot be stat'ed asks the background warmer for nothing, so a warm entry whose transcript is gone derives cold once instead of parsing a missing file every build, dropping the feed cache, waking the pusher and painting the card quiet.

**Defect.** The in-place re-read added by #1820 fires for any memoized entry whose key was warm and whose cache-only parse read just missed. When the miss is because the transcript is gone from disk, the shared store has no slot for it: `parse_cached` returns None because the file-set key's stat raises, and `parsed_session` parses the missing leaf under a None key and stores nothing. So the next build missed again and re-read again, forever: `parses.kernel` and `coldFlip` (the counter documented as the regression watch for this block) moved on every build, and the empty parse the re-read produced painted the card's `sessState` quiet, a settled state for a transcript the kernel cannot read, where the cache-only read reports unknown. The kernel's tab-list path already says on the same build that the transcript could not be resolved.

**The second road.** Gating the re-read alone moves the same defect to the background warmer: with the re-read withheld the entry's parse is None, and the body then asked the warmer for the session whenever it had moved since boot or its row was working. Under a feed-only client, the warmer's reason to exist, the warm thread parsed the missing leaf every build (a None key, nothing stored), counted a kernel parse, dropped the pusher's cached feed and woke the pusher: build, warm, parse, drop, wake, at build speed for as long as the transcript was gone. Main paid one inline parse per build with no drop and no wake, so that half-fix would have been worse there.

**Fix.** Both roads close on one bit. The re-read is gated on the transcript identity the key already took at its top, which is None exactly when its stat failed; no second stat and no new read for the inputs census, and `coldFlip` stays inside the block so it counts only a re-read that ran. The key hands the same bit to the body through the build context (`leaf_ok`), and the body sets its cold flag, the warmer's ask, only when it holds; the inputs census classifies the new context field under the transcript component. Gated at both ends, the entry derives cold once, hits from then on with nothing parsed anywhere in the kernel, and reads unknown until the transcript is back and something parses it, at which point the existing cold-to-warm flip takes over. A chmod-0 leaf still stats and re-reads as before; a renamed one resolves through discover. The block comment, the key's parse-component paragraph, the cached-read helper, the warm-key helper, the warmer and the feed-warm test module all say why, and the feed-warm module's source-text pin is re-cut to the gated line.

**Validation.**
- Red first, the re-read: the new test in the feed memo module, a warmed session whose transcript is then removed, fails against the previous kernel at its first post-removal build with `AssertionError: 1 != 0` on the parses.kernel delta (and would go on to fail on coldFlip +1, sessState quiet, and the memoized parse bit True). Green with the gate: parses 0, coldFlip 0, coldLive 3, sessState unknown, parse bit False; the next build hits 3 and parses nothing; the recovery tail rewrites the transcript at a different size, asserts the cache-only miss, builds (one derivation under transcript alone, still unknown), parses it, and the next build flips warm with coldFlip unmoved.
- Red first, the warmer: a sibling test with a fake feed client on the client list and the live map stubbed to the board's rows, every session parsed before the removal so the gone leaf is the only cold candidate. Against the previous kernel it fails on the same parses delta (the inline re-read); against the re-read gate alone it fails with `AssertionError: 1 != 0` on the warmer being called, and a probe there read, on each of three post-removal builds, one warm call, parses +1, the feed cache dropped and the pusher woken. Green with both gates across three builds: no warm call, the feed cache standing, the pusher not woken, parses 0, sessState unknown.
- tests/test_feed_session_memo.py: 56 passed, 13 subtests passed
- tests/test_kernel_feed_warm.py: 4 passed (the pin re-cut)
- tests/test_feed_memo_inputs.py: 31 passed (the read census sees no new read; the new context field is classified)
- tests/test_perf_stats.py: 65 passed
- tests/test_boot_parse_gating.py: 28 passed (pins the warm-wanted line, unchanged)
- tests/test_kernel.py: 460 passed, 12 subtests passed
- Every other module that names the warmer or pins the key's source: chat empty build guard 5, feed live clobber 3, goal store fault boundary 30, interrupt marks memo 19, kernel interrupt 54, kernel goal cache wiring 17, kernel interrupt machine cut 51, perf bench 49, pusher change gates 13, kernel judging active 5, all passed.
- Static pin scans over the branch: 0 lost of 8746 assertIn literals; the one changed function pinned by getsource has all four pinning modules in the runs above.

Follows the post-merge review of #1820.


ksarma:
The evidence here was executed against 2f66a4c1e, the PR's head, merged as 73757164b. File and line references are on that commit; the merge carries the head's tree unchanged for the PR's four files, so the numbers read the same at the merge.

**The background warmer's own loop has no leaf gate, so the gone leaf is still parsed once per warm kick raised by another cold session** (kernel/kernel.py, lines 35146 to 35148).

- Mechanism. The PR gates the ask (the body sets its cold flag only when `leaf_ok` is on), so the gone leaf alone never kicks the warmer, but the loop filters on `_warm_wanted` alone at line 35146, which stays true for a gone leaf through the store's or states log's mtime since boot. So once any other cold entry kicks the warmer, the warm thread calls `_parse` on the missing path at line 35148: a None store key, nothing stored, one kernel parse counted, about 0.23 ms, nothing on stderr.
- Evidence. With every session parsed before the unlink, one build derives web cold with zero warm calls and zero parses, the PR's claim as stated; once a sibling asks (a session left unparsed before the removal, api's memo entry dropped and its transcript appended, a new session, a restart with the memo and parse store cleared), each kick's warm parses the gone path and `parses.kernel` moves by the stat-able sessions' misses plus one. Bound: the repeat is per kick, not per build. The build after the kick ran zero warms and zero parses; a per-build re-parse would need a peer that stands cold and kicks every build, and the only such candidates under this PR are the stat-able inputs the block comment already names as ungated (an anchor candidate, a states file), a pre-existing condition rather than a loop this PR leaves on its own.
- Consequence. The body's two claims, "hits from then on with nothing parsed anywhere in the kernel" and "both roads close on one bit", are true only while the gone leaf is the sole cold candidate, exactly the shape the PR's warmer test builds, and the docstring sentence at line 35123 is true of the ask and says nothing of the loop; nothing is stored or painted, so the residual is cost and that overstatement, not a card move.
- Fix. Gate the loop at its subject, in the warmer's inner function: add `or _chat_ident(s["path"]) is None` to the condition at line 35146 (or a stat call whose failure skips the row), ahead of `parsed_any = True` at line 35149 so a warm that skipped everything drops nothing and wakes nothing. That is the predicate the key reads at line 42382, one stat per warm-wanted session per kick on the warm thread and never the feed path, so the PR's no-second-syscall point stands; inside `_warm_wanted` it would double-gate the body, whose caller already sits behind the leaf bit, and the (0.0, 0) the files-stat helper returns for a missing file is a sentinel, not the stat's outcome. The gate also skips the transcript-less rows the live-session list adds (the live stub, a just-created SDK row), whose parse today is the same store-nothing waste when the row is working, a correction rather than a regression; the warmer tests in tests/test_boot_parse_gating.py write real transcript files and survive it, and no test pins the loop's parse line by source text. Add the multi-cold shape beside the PR's warmer test, which stays valid as the sole-candidate case, asserting a `_parse` recorder never sees the gone path and `parses.kernel` moves by exactly the stat-able sessions' misses, red before the gate; the docstring at line 35123 then says the warmer skips such a leaf itself, and the body's two claims become true. A disclosure-only alternative is defensible given the cost, but then the two claims need narrowing to "while no session asks for a warm", with the residual named in the docstring.

## Smaller points

- **The tab-list clause at lines 42464 to 42465 describes one road.** The stderr line it credits fires only for a live sid no backend owns (the names-stub road through `_live_stub_session` and the note helper at line 5169, once per episode), while an SDK-owned session takes the `_sdk_sess` road at lines 5258 to 5262, which never notes, so for it the PR leaves no trace on stderr or in /perf where the base at least left `coldFlip` and the kernel parse count climbing every build; drop the clause or qualify it (the stub road reports only a live sid no backend owns), and the matching body sentence, rather than noting from the SDK pass, since that road also represents a just-created SDK session with no transcript yet and the re-arm loop at line 5271 would clear and re-print the note on every run, and a signal that tells gone from not-yet belongs at the warm-key transition, the next point.
- **A signal for the condition, if wanted, belongs at the transition, not beside the cold count.** A counter beside `coldLive` at line 42431 keyed on the transcript identity being None would be one signal with two causes, since the bit reads False on every build for a just-created SDK session before its first transcript; the transition the gate withholds, a memoized warm key whose transcript now fails to stat, is the condition to name, so an `elif _feed_key_was_warm(prev_key):` branch after the gate at line 42437 that counts fires once per disappearance episode on both roads (the stored key's parse bit is False from then on, as the PR's own test asserts), asserted at delta 1 on the first post-removal build and 0 after in both new tests and 0 for api and tests, with the seed and comment in `_FEED_MEMO_STATS` at lines 41895 to 41908, the key-set pin at tests/test_perf_stats.py line 224, the reset helper's zero list at tests/test_feed_session_memo.py line 123, and the counters' documentation.
- **The chmod-0 disclosure at line 42469 names the mechanism, not the surface.** Pre-existing with identical counters on the base, so not a regression of this PR: a leaf that stats but cannot be opened is re-read once into an empty parse, which the store keeps under the transcript's (mtime, size), so the card paints quiet from it on every later build and the slot outlives the permission repair (a chmod moves ctime alone) until the mtime or size moves; no code change is owed here, since the close is a read-failure signal out of `parse_session` (the file adapter passes no failure callback), its own change, and the ask is to say this where the comment says "stats and re-reads as before" and in the body's Fix paragraph.
- **The key's docstring list at lines 42251 to 42252 omits `leaf_ok`.** It enumerates the six facts handed to the body through the context while the update at lines 42489 to 42490 hands over seven; the entry's docstring at line 42576 was updated, this one was not, the inputs census does not pin the header list, and the entry's list already lacked `hide` at the base.

## Tests

- **The re-cut pin at tests/test_kernel_feed_warm.py line 36 pins the spelling of the gate line, not the gated behaviour, and its message does not name the tests that do.** Spelling-only mutants (an operand swap, the conjunction split across two lines) fail the pin alone, behaviour mutants (the stat clause removed, the identity check inverted) are caught by the executed tests in `AWarmEntryIsNeverDerivedCold`, and a mutant with the stat clause removed but the literal left in a comment leaves the pin green while the behaviour tests go red; the shape predates the PR, since the base pin had the same character and its literal could not match at this head. Fix: keep the literal as a locator and have its message say so and point at the executed proof, `test_a_warm_entry_whose_transcript_is_gone_derives_cold_once_and_re_reads_nothing` for the gate, its sibling `test_a_warm_entry_whose_transcript_is_gone_asks_the_warmer_for_nothing` for the warmer ask, and the two pre-existing warm-half tests (the append test and the emptied-store test) for the 2026-09-18 clause; loosening the literal back to the warm-key call alone also passes today but still matches a comment-only mutant and would need the same pointer, and once the warmer gate lands the pointer gains the multi-cold test too.

PR #1942, opened 2026-09-21
The chat baseline a whole-frame sender hands its seed is the one it read before its build, so a seed that lands mid-build takes the detector's road instead of stranding a client on a stale card

lczh:
Tier: fix

**The defect.** A session's first whole chat frame seeds the shared delta baseline, and a sender's seed step compares the baseline it read against the one in the map, so two senders that both read it absent are detected as a race. But both whole-frame senders (the pusher's per-session loop and the targeted one-session push) read the baseline after their build. A sender whose build read the transcript first, but whose baseline read landed after another sender's seed, read that seed as a present baseline: it diffed its older list against the newer one, sent the difference as tails (or, to a fresh client, as a full) that put the older card back on a client the other sender had just given the filled one, and its own seed declined, since a present baseline is never touched. The map then held the newer list while some client held the older card, the next cycle diffed equal lists, and the client kept the stale card with no repair, no row and no mark. The strand reaches the newer sender's own client too: the page that had the filled card from its own connect push was regressed by the older targeted push's tail.

**The fix.** Both reads move before the build (the cycle's to the top of its iteration, before the cache lookup and the single-flight claim, so a served list and a built list both diff against the baseline as it stood before the iteration read anything). A seed that lands during the build now reads as absent-then-different at the seed step, the detector's road: the entry is popped, the session marked, and the next cycle's full repairs every base holder with a row per stale holder. The seed's call conditions are unchanged.

**The cost, stated plainly, and where it lands.** The extra frames land wherever a sender that read the baseline absent had ANY whole-frame writer land during its build, a racing seed or the cycle's every-client write: that sender sends change-0 fulls where it sent tails, its seed pops the baseline and marks the session, and the next cycle sends every base holder a change-0 full again, each with a row on the whole-frame meter under the changeAt0 reason. In the strand face those fulls replace a silent stale card. In the other face they are pure cost: the detector cannot tell a racing seed that landed mid-build (the sender's list the older one) from the cycle's write landing mid-build with the sender's list the NEWER one, because the map's content carries no order (a filled card has the length of its unfilled twin) and a written-by-the-cycle flag is not sound (a connect seed, then the cycle's tails and write, then an older targeted push's fulls: declining the pop there re-opens the strand). So a targeted push that read the baseline absent and had the cycle build, send and write inside its build pops the cycle's baseline and marks the session with every client already holding its newer list: two change-0 fulls per client and two rows where the kernel before sent one tail per client, the mark standing with no stale holder. That is the boot's ordinary interleaving, the cycle's cold build of the watched tab beside the attach handshake's targeted push with the transcript moving between the two reads; it is bounded by the number of senders building the session at once, and an equal-list race dedups on the client's slot and costs nothing on the wire. No discriminator is added here: a stamp of the build's start kept beside the baseline is the one that removes it, and it needs its own design and review. The false positive is stated at both read sites, in the seed's docstring, at the mark's definition, and in the changeAt0 gloss in the code and in the reference, which now name this face beside the race and the genuine first-event change.

**Not closed here.** Two faces, both named in the code. Pre-existing: a present baseline the cycle advanced while an older targeted push was mid-flight, the push's tails landing after the cycle's on some client. New with this change and accepted: the false positive above, whose red test for the follow-up stamp item is the third test added here.

**Validation**
- Red first: test 28 of the skeleton-reconnect module, in two pairings (the targeted push as the older builder with a connect push seeding inside its build; the connect push as the older builder with a targeted push seeding inside its). Against the unchanged kernel both fail at the baseline assertion: the baseline stays the racing sender's filled list, with no mark. Green after the change: the entry popped, the session marked, and the next cycle repairs every stale holder with one changeAt0 row each (both clients in the first pairing; one in the second, where the other client's identical full dedups).
- The accepted cost, pinned as the third test (the targeted push as the NEWER builder with the cycle's build, sends and write inside its build). On the branch: every client holds the cycle's full and then the push's change-0 full, nobody stale, the baseline popped and the session marked, two changeAt0 rows; the next cycle sends each client a third whole frame with two more rows, clears the mark and re-establishes the baseline. Against the unchanged kernel the same test fails at its first frame assertion: each client holds the cycle's full and then one tail carrying the filled card, no full from the push, no mark, no row, and the next cycle sends tails again. That failure pins the changed behavior; it is not a defect the test caught.
- skeleton-reconnect: 52 passed (49 before, plus the three)
- chat delta resync: 12 passed (the regex pin on the cycle's write gate untouched)
- kernel pusher snapshot: 31 passed, 2 subtests
- kernel: 460 passed, 12 subtests
- perf stats: 65 passed (its gloss test reads the reference's reason list)
- chat pages: 43 passed, 27 subtests
- perf bench: 49 passed
- chat empty build guard 5, cold tab gate 22 (+4 subtests), needfull no dedup 1, single-flight builds 11, pusher change gates 13, and the eight further modules that pin the two senders by source: all passed
- Static pin scans over the branch: 0 lost literals; the two senders changed shape and every module pinning them by source was run.


ksarma:
Evidence for this note was executed against 8f22252d0, the PR's head, merged as d7254b24c; the merge tree differs for two of the three files, so every file and line reference is on 8f22252d0. Both reads sit before their builds (kernel/kernel.py:54712 and 55410), and tests 28, 28b and 28c are red on the unchanged kernel; what remains concerns the cost partition, the changeAt0 gloss and what the tests pin. PR 1943, merged since as fe1ddd2c2, leaves every clause cited here standing at other line numbers.

The two-face partition (a racing seed inside an OLDER sender's build, the strand repaired: 28, 28b; the cycle's write inside a NEWER sender's build, pure cost: 28c) is stated at kernel/kernel.py:54697-54704, 33929-33936 (the seed docstring), 51360-51366 (`_chat_full_reason`), 55405-55407 and docs/reference.md:2089-2098. It names two cells of four, and a further shape falls outside it.

- The cycle's write inside an OLDER sender's build (two probes agree). At the head the sender's fulls put the older card on every base holder, its seed pops and marks, and the next cycle's fulls repair, two changeAt0 rows per client in all; the unchanged kernel leaves every client on the older card for good, with no mark and no row, a strand this change fixes and the body does not claim. The "no client stale" sentences are conditioned on the sender's list being the newer one, so the cell is unnamed rather than misdescribed, except at 54701-54704, which asserts the newer-order outcome for a boot interleaving that occurs in both orders. Fix: name the cell at the sites above as the strand face with the cycle's write as the inner writer, and add a test mirroring 28c, the sender's build reading the unfilled card and `inside="m3 filled"`, asserting every client on the older card after the push, then the pop, the mark and two rows, then the repair's rows (stale=2 in the existing helper; no status flip, the repair's frame differs from the push's); red on the unchanged kernel at the first frame assertion.

  The partition's other unnamed cell (one probe): a seed from the OLDER sender inside a NEWER sender's build (two targeted pushes on one sid). Where the unchanged kernel sent one tail per client, the head sends two fulls per client, pops, marks and files two changeAt0 rows, with two more at the next cycle, no client ever stale: the 28c false positive with a different writer, reachable live since `_push_session_now` has no per-sid serialization. Fix, no kernel change: state the pure-cost face as any whole-frame writer landing inside a build whose list is the newer one, once at the seed docstring and pointed to elsewhere, and add a test of this ordering that flips the status before the next cycle as 28c does, or the fulls dedup on the slot and the second row count reads short.

- The cycle itself as the sender that read the baseline absent, with a whole-frame seed (targeted or connect) landing inside its build (two probes agree). The comment at kernel/kernel.py:54694-54700 and the body say any such sender's seed pops and marks, but the cycle has no seed step: at the head every base holder gets a change-0 full of the older list and a changeAt0 row where the unchanged kernel sent the same older list by a tail, nothing is marked, the older list is written over the seed at 54849-54854, and the next cycle repairs by tails with no new row. The face is cost plus a wrong reading of `pusher.chatFullWhy`, not a strand, and falls outside the gloss's exhaustive "in two faces". The write over the newer seed is correct; the test pins it as the face's shape. Fix: test 28d, the cycle as the outer builder with `km._push_session_now(S1)` inside its build, asserting two change-0 fulls of the older list, two rows, no mark, the older list as baseline, then the next cycle's tail with no new row, its docstring saying, as 28c's does, that its red on the unchanged kernel is the frame shape; and name the face at 51360, docs/reference.md:2088, 54694-54700 and the seed docstring's two-faces sentence (narrowing 54694-54700 alone leaves the gloss's claim standing).

## Smaller points

- A third strand face, pre-existing and absent from "Not closed here": a sender whose seed lands BEFORE the cycle's write while its older full lands AFTER the cycle's on some client leaves that client on the older card with no mark and no next-cycle repair, on both kernels. Only the seed docstring at kernel/kernel.py:33966-33969 names it, and the write-site comment at 54830-54850 justifies the write by every base holder having got a full, which this face falsifies; name it at 54694 with a pointer to the seed docstring's residual and at the write site; closing it (a build-start stamp, or a cycle write that declines to overwrite a seed it did not read) is its own change.

- With the baseline PRESENT at the cycle's pre-build read, a detector pop landing during the cycle's build is repaired one cycle later than before (the cycle diffs against the list it read, so the mark stands). It bites only when the targeted push's send phase, not its build, spanned the seed and the cycle's read; when the build itself overlapped, the unchanged kernel stranded the client for good, so the disclosure names that bound. Fix: disclose at the cycle's read comment near kernel/kernel.py:54712 and in "Not closed here"; a code change is the author's call, and only an ABSENT post-build re-read may switch the cycle to fulls, since a present-but-different list cannot be diffed against.

- Two cost sentences misstate the slot dedup. "The next cycle sends every base holder the full once more" (docs/reference.md:2096-2098) applies only when the session frame moved, or `_DEDUP_REPOST_S` (60 s, kernel/kernel.py:49572) passed, since the sender's full; unchanged within the window the second full dedups on the slot at 50722 and files no row, so the false positive then costs one change-0 full per client. Wording only, at every copy: the gloss, kernel/kernel.py:33550, 33936, 51364-51365, 54696, 54703-54704, 55407 and the 28c docstring (tests/test_chat_skeleton_reconnect.py:1696); 28c keeps its flip, with its docstring saying the flip makes the third frame leave. "An equal-list race dedups on the client's slot and costs nothing on the wire" (kernel/kernel.py:54707 and the body) is false for a client the racing writer served a TAIL during the sender's build: that client gets a change-0 full and a changeAt0 row where the unchanged kernel sent an empty-suffix tail, with no pop and no mark; the precondition is narrow (a seed landing between the sender's read and the writer's read), so qualify the clause at both.

- The `_push_session_now` docstring at kernel/kernel.py:55314-55315 names the writers a pop means as a connect push or a second targeted push on the same sid, omitting the cycle's every-client write, a pop cause since this change; its conclusion that some client may keep the older build is true for that writer when this push's list is the OLDER one; what it lacks is the 28c face. Docstring only, two insertions: "or the non-connect cycle's every-client write" inside the parenthetical, and after "marks the sid" the clause "when this push's list was the newer one and the cycle's write landed inside its build, no client does, and the mark stands with no stale holder, the detector's false positive stated at the read below (test 28c)".

## Tests

- The placement claim at kernel/kernel.py:54686-54711 and in the body (a served list diffs against the pre-iteration baseline) is pinned by no test: 28, 28b and 28c pin the read-before-build half, and a mutant moving the read after the single-flight wait stays green. The placement has a consequence, so the test is the fix rather than narrowing the comment: stub `_chat_inflight_claim` to return an Event whose wait runs a connect push for b inline, then run the cycle for a and b, and assert on frames, not the mark: at the head no chatTail follows b's full and a stale holder a from a mid-flight sender ends on the filled card right after the cycle with one changeAt0 row, where under the mutant b gets an empty tail after its full and a stays on the unfilled card with no row until the next cycle; both placements write the baseline and neither has set the mark by then.

PR #1943, opened 2026-09-21
The chat baseline seed runs only for a build some client took whole, so a skeleton holder's status-only pass seeds nothing and pops nothing

lczh:
Tier: fix

The chat wire's shared delta baseline is seeded by a session's first whole frame (#1909), so the pushes after it send tails instead of whole-session repaints. Both places that seed ran the seed after their per-client loop whether or not the loop had handed the build to anyone. A page that holds a tab as a skeleton gets only a status frame for it, a client whose frames are withheld until its bundle is ready gets nothing, and a socket before its ready gets nothing either; a pass whose every client was one of those handed the build to nobody, and seeded a baseline that no client holds. A lone redialing page's connect push seeded every one of its skeleton tabs that way, and a handshake or a stream event for a tab every page holds as a skeleton seeded it too. This change is stacked on the seed lower-bound fix from the same review (#1942): that fix's commit is in this branch's history, and the diff above it is this change.

That list is no lower bound on any client's state, and it does harm in two shapes. When another page is looking at the tab and its own push lands inside the skeleton page's send loop, the skeleton page's seed finds a list other than its own and reads it as a race between two whole-frame senders: it pops the only base holder's list and marks the session, so every targeted push until the next cycle re-sends that page the whole session, and the cycle's repair is a whole-session frame counted as a first-event change with a client-diag row, the very frame the seed was merged to remove. In the other order the status-only pass writes first: the whole-frame sender then reads a baseline present, hands its page the older list, and declines its own seed; the next cycle diffs equal lists and the page keeps a stale card with an empty tail, no row and no repair short of a reconnect.

The fix: after the loop, each call site checks whether any client in the loop now holds a base for the session, which only the whole-frame road records, and seeds only when the baseline it read was absent or empty and that check holds. The check reads each client under its own slot lock and releases every lock before the seed takes the baseline lock; a present baseline never reaches it, so the steady state is untouched, and a content frame over an empty baseline still seeds as before. The seed's docstring, the targeted push's docstring, the full-reason meter's own docstring and the reference's gloss on first-event fulls now all say the first whole frame to reach a client seeds, so the kernel's and the reference's descriptions of that meter agree.

Validation
- Red first on the merge base, six probes in the skeleton-reconnect module, each run against the merge base's kernel file and restored byte for byte afterward (the four new tests together: 4 failed, 53 deselected on the stacked module (49 deselected on the module before the lower-bound fix's three tests and test 33 joined it); the two folded later: 2 failed, 55 deselected on the stacked module (52 before those tests joined it)): a lone redialing client's connect push seeded its two skeleton tabs from status frames (the baseline map held all three transcripts); a targeted push for a tab its one client held as a skeleton seeded it; the status-only sender racing a whole-frame sender popped the only holder's list and marked the session (the baseline read back as None); the inverted order left the newer filled list as the baseline while the page held the older card, with the next cycle's tail empty; an attach handshake's targeted push reaching a pre-ready column before the first cycle seeded the withheld tab (the map held that one transcript; with the fix the ready arm's push, the one full, is what seeds it); and a targeted push whose only client was a socket before its ready seeded the session it withheld (the map held that one transcript, the frame counted withheld and no base recorded).
- Green after the fix and after stacking on the lower-bound fix, one module at a time: skeleton-reconnect 57 passed, chat-delta-resync 12, perf-stats 65 (its gloss test reads only the reason labels the meter returns, so the docstring rewording is outside it), kernel-pusher-snapshot 31 (2 subtests), kernel 460 (12 subtests), chat-pages 43 (27 subtests), perf-bench 49.
- Neighbors the spec named, also green: cold-tab-gate 22 (4 subtests), needfull-no-dedup 1, single-flight-builds 11, chat-empty-build-guard 5, injected-voice 10 (222 subtests).
- The static scan of test pins on kernel source lines against main lost nothing; no extension or webview test names the changed lines.


ksarma:
The change gates the chat baseline seed on delivery: both seed call sites now seed only when the baseline read before the build was absent or empty and some client in the per-client loop holds a base for the sid afterwards, so a status-only pass, a withheld client and a socket before its ready seed nothing. Every file and line reference below is on 585eb3585, the PR's own commit, where the evidence was executed; the merge commit fe1ddd2c2 carries the same hunks at shifted lines (main's merges of #1941 and #1942 moved kernel/kernel.py, #1944 moved docs/reference.md), so each point stands after the merge. The module's 57 tests pass at 585eb3585, and its six probes are red against the #1942 head kernel and the merge base's kernel alike.

## Smaller points

- **The delivery signal is a state read after the loop, not the loop's own event.** `_chat_handed_whole` (kernel/kernel.py 33925 to 33945) answers whether any loop client holds a base for the sid at the walk, so a base another whole-frame sender writes on a loop client between that client's status send and the walk counts as this loop's delivery, and a status-only pass seeds; when its list differs from the map's, the detector pops the other sender's list and marks the sid. Reproduced on a scratch copy of 585eb3585 and on the #1942 head kernel: the sid ends marked and the baseline popped with one whole-frame sender in play, so this is a window the guard leaves open, not a regression. The reachable shape is a targeted push (a backend thread, no lock across its loop) whose every client holds the sid as a skeleton, preempted between its last status send and its walk while a page's reader thread handles that page's needFull or activeTab for the sid and a whole-frame sender (that page's connect push, or the woken cycle) hands it the full; a page's own connect push cannot race its own release, since the release paths run on the reader thread that runs the connect push inline, and a status-only pass materializes no JSON, so the window is the remaining status sends plus the walk. The cost is one pop and mark: at the next cycle every base holder gets a change-0 full with a changeAt0 row when the transcript moved since the writer's full, and the identical repair full is absorbed by the per-slot dedupe with no row when it did not. Closing it means reporting delivery from the echat write itself, in `_send_chat_locked` and in the proto-2 sender it delegates to, OR-ed across the loop at both call sites (54893 and 55474) in place of the after-the-loop read; the flag has to mean an echat entry was written for this frame (proto 2 writes none for an empty list, a socket before its ready writes none), not that a full was decided, and checking `ms` for None is no substitute, since proto 2 never materializes it. The red-first test is that interleaving, asserting the sid unmarked, the repair's list as the baseline, and a chatTail rather than a changeAt0 full from the next cycle over a grown transcript. The "Exact today" sentence (33936) presents the state read as the loop event; if the window stays open, that sentence and the seed docstring's "does not reach this function" clause (33970 to 33972) both want the residual stated with its one-cycle cost, and the title's "pops nothing", true for the single-threaded shapes, carries that exception.

- **A comment in the proto-2 sender is now incomplete** (kernel/kernel.py 51565 to 51570, untouched here): it names two reconnect-only readers of the echat key set and closes with "A reader of the key set on an unarmed client would be the first to see it", while `_chat_handed_whole` is now that reader, a membership test on every loop client with no flag check, and is not named; bounded to the comment, since by execution the empty-list pop makes the guard answer False and the seed writes nothing, so add the name (for it the pop reads as handed to nobody) and replace the closing sentence rather than leaving it beside the new name.

- **Optional wording at kernel/kernel.py:33949.** The seed docstring's opening line says "at a sid's first whole frame; never advance it" without the "to reach a client" qualifier that docs/reference.md:2087 and the full-reason meter's docstring carry; the line is #1909's, identical at the merge base and the #1942 head, the rewritten deciding-event sentence at 33969 to 33972 states the delivery guard, and no test pins the line.

- **The validation figures predate the stacking on #1942.** The body's "4 failed, 49 deselected" and "2 failed, 52 deselected" reproduce only from the module without #1942's three tests (and, for the first, without test 33); on the 57-test module at 585eb3585 the same selections give 4 failed, 53 deselected and 2 failed, 55 deselected, the failing counts and their reasons reproduce as stated, and the commit message of 585eb3585 says "skeleton-reconnect 54" where the module has 57 (the body is still editable, the commit's figure is history).

## Tests

- The slot lock in `_chat_handed_whole` (kernel/kernel.py 33942) has no pin: deleting the lock line in a scratch copy leaves tests/test_chat_skeleton_reconnect.py at 57 passed while a mutant returning True first turns six tests red, and no test under tests/, ui/ or vscode-extension/src names the function; the lockless read changes no single-threaded behaviour, so only a discipline pin catches it; an executed pin hands the function clients whose `dlock` is a counting context manager and asserts two takes when neither client holds the sid and one when the first does (any object with enter and exit serves, as the lock helper returns the client's entry as is, and the module's client factory sets none), and with a recording stand-in it also pins the release of every client lock before `_chat_baseline_lock` is taken, which no test covers either, where test 11's `assertLess` shape over the function's source, lock-acquire index before echat read index, is true for the right reason today but a source-text pin a later docstring edit could satisfy.

- The targeted push's copy of the guard (kernel/kernel.py 55474) has no pin for its `not _seen` half: changing only that site to a None-only check leaves the module at 57 passed, while changing both sites turns test 20 red, so the road is pinned through the connect push's copy alone, and if the copy drifts a transcript-less session's cycle write leaves an empty list as its baseline, the first content frame a stream event hands a client seeds nothing, and every later targeted push re-sends every base holder a changeAt0 full with a row until the next cycle writes; owed is a test driving the first content frame over an empty baseline through `_push_session_now` with a fresh client, asserting the seed and the tail after it, as a second half of test 20 or on its own, and merging the two guard copies is a valid alternative with `_seed_chat_baseline` the natural home (it owns the seen-or-empty return, so checking delivery there after that short-circuit keeps a present baseline from taking any client lock and restores the single unconditional call both sites had before), the test owed either way, since a helper without it leaves the targeted site's wiring unpinned.

- The seed docstring's citation at kernel/kernel.py:33973, "tests 29 to 32", misses two pins: by execution the withheld road is pinned only by the third part of test 11_d and the pre-ready socket only by test 33 (replacing the targeted-site guard with `if True:` turns 11_d, 30 and 33 red with 29, 31 and 32 green; replacing the connect-site guard turns 11_d, 29, 31 and 32 red with 30 and 33 green), so citing "tests 29 to 33 and the third part of test 11_d" is the minimal fix, the helper's docstring (33928 to 33929) names the same roads uncited and could take the same clause, and lifting that part into its own test named for the withheld road would name the pin for what it guards.

PR #1944, opened 2026-09-21
Put back four facts the docs move dropped, and sharpen the front-pages test

henrypinkard:
Folds two executed reviews by ksarma: the review of #1931 (the docs restoration) and the review of #1936 (the wording pass), posted 2026-09-21. Every item of both is here; nothing is left open.

### The four statements that were in no page under docs/

Each was read back from the pre-restoration guide (ce67db41e) and verified against the code before it was written.

- **The postal card's head and its delivery mark**, as a new section of `docs/reference.md`'s "The interface, feature by feature" part (before "A comment thread's mail"), not under the shell's "Mail from the terminal". It says the head names both ends in their session colors, the mark sits at the head's right edge, a send that failed has no mark and the tool call's result says what happened, an incoming message that waited while the session was offline wears the parked mark, and hovering a mark gives the state and when it was reached. Verified in `ui/webview/postal-state.ts` (`deliveryOf` returns null for a send with no status and for an in-hand incoming; `park` yields the parked mark; `deliveryTitle` appends the clock) and `ui/webview/postal-name-ink.test.ts` (both ends as bold text in the identity color).
- **The trust level on re-link, and the restart fan-out**, back in `docs/guide.md`'s linking section, which has headroom under its cap. Verified in `kernel/kernel.py`: the remembered-hosts registry stores the level so a re-attach keeps it, and the restart sweep plans "ask" for a checked-in peer, which restarts itself alone while machines attached to it are restarted from its own interface.
- **OpenSSH forwarding each browser socket one-to-one and propagating closes**, as the opening paragraph of the reference's "The VS Code port forwarder and the browser dashboard", in reference voice with no relative pointers.
- **The token-exempt routes and the reason**, under the reference's "The token file: minting, permissions and refusals": `/healthz`, `/version`, `/busy`, the bus's `/ping`, `/manifest.webmanifest` and the three home-screen icons under `/media/`, which are fixed files that read no session state, plus the note that `/busy`'s drain hold is a write and takes the token.

### The two smaller points

- The moved admonition no longer says "prefer one of the two paths above": it names plain ssh port forwarding, linking the guide's From another machine section, and Tailscale, linking the reference's own Tailscale section.
- "A blue outline shows where the pane will land" is now "An outline in the accent color shows where the pane will land": the ring is the accent variable, a clay orange in the light theme.

### Its test items

- The `not_in_nav` check parses the block and asserts `pr-tiers.md` is one of its lines; before, the word appearing anywhere before the nav key passed. Its two messages claimed the entry keeps the strict build quiet; MkDocs reports an omitted page at info level, so they now say the entry declares the page intentionally unlisted.
- The paragraph cap names the line in the FILE: `_prose_lines` yields `(line, text)` pairs.
- A line opening with `<` is skipped only when it is block HTML, so a wrapped prose line beginning with an inline tag no longer splits its paragraph. Four lines in the guide rejoin their paragraphs; the largest paragraph is 67 words, inside the 70-word cap.
- The comment's word counts name their measure: `_words` gives 394 / 259 / 3932 for the restored pages and 394 / 690 / 11,195 for the pages they replaced; the `wc -w` figures it used to quote are labelled as such.

## From the review of #1936

- **The spend readout sentence** described a sessions-versus-judges split the API spend panel does not draw: the panel reads the spend file whose one writer is the session turn's result handler, and the judges log elsewhere and surface in Token usage. It now reads "The bottom bar's spend readout opens API spend: the sessions' spend in dollars, by session and over time", naming the panel as the interface names it. No new reference link, so the one-pointer pin stays at one.
- **Four paragraphs no longer end on a link.** The guide's same-page pointer to Security and trust and its pointer to How Romp works are gone (the architecture page is in the site navigation); the SECURITY.md pointer is a clause inside its paragraph; the install page's "what it installs, in detail" link moved into the sentence before it. Making the rule mechanical surfaced two more the pin catches: the Codex backend list item now says what a Codex session needs with the link inside the clause, and the home page's machines paragraph ends on the claim rather than on the phone link. README.md is not one of them: its License line is exempt as a link-only navigation line and its Docs line ends inside a closing parenthetical, which the pin's shape does not match.
- **The interpreter note** moved out from above the install command to under the Install heading, after the sentence about what the command clones and installs. The command is at line 18, well inside the position pin's cap of 40.
- **The pane is the Sessions pane** in the views list and the section heading; the slug is unchanged, so no link moves.
- **Typing `@`** and the first letters lists the live sessions that match, and picking one inserts the name, which is what the composer does.
- **CLAUDE.md** states that the no-trailing-link rule covers a link to a section of the same page and excepts a navigation line, and its list of what the test pins gained the guide's single pointer to the reference and the no-trailing-link property.

### Its test items

- A property pin over the four governed pages (index, install, guide, README): no prose paragraph ends on a markdown link followed by at most a period, link target ignored. Executed red before the rewrites, where it failed on six paragraphs across three pages (README.md passing: its License line is a navigation line, its Docs line ends inside a parenthetical), and green after.
- The module's docstring and the word-cap message told an editor to leave a link where the one-pointer pin forbids one; they now say the detail lives in the reference under a heading matching the feature's name, and that the paragraph states what the feature does.
- `ui/webview/dense-chrome.test.ts` asserted the guide names no compact-tabs gear label, which the base guide also lacked; it now asserts the positive the wording pass made true, that the guide no longer contains `reference.md#the-chat-pane-in-detail` (1 occurrence at a59c0cde3, 0 now).

## Tests

`tests/test_docs_front_pages.py`: 7 passed, 298 subtests. Red-first checks, all executed: the trailing-link pin fails on six paragraphs of the pages as they stood; pointing `not_in_nav` at another page fails the new assertion; a 90-word paragraph appended to `docs/install.md` is reported at its real file line, 71, where the old index said 53.

`tests/test_api_health_hover.py` with the front-pages module: 43 passed, 298 subtests. `ui/webview/dense-chrome.test.ts`, the one node module that reads both pages: 10 passed. Greps of `tests/`, `ui/webview/*.test.ts` and the extension source find no other pin on a sentence this change touches.

Page measures after both commits: index 397, install 242, guide 3909 words (caps 600 / 400 / 5500), largest paragraph 67 words, install command at line 18, one reference pointer in the guide at line 5.

Tier: docs


ksarma:
The evidence below was executed against ea904920c, the head of this pull request, and every file and line reference here is on that commit unless it names another; the merge commit 7535b52d7 also carries main's own changes, which put the docs/reference.md lines cited below 10 lower there and the kernel/kernel.py lines 102 lower.

The change puts back the postal delivery mark, the trust level kept on re-link with the restart fan-out, the one-to-one OpenSSH forwarding and the token-exempt routes, and its second commit matches the front pages to the API spend panel, the interface's labels and the no-trailing-link rule. Read against the note on #1936 (https://github.com/romp-on/romp/pull/1936#issuecomment-5755879886), the second commit closes, each checked by execution or by reading: the spend sentence, now docs/guide.md:499-500 naming API spend and the sessions' spend by session and over time, which is what `_spend_detail_local` draws from spend.json; the trailing-link paragraphs, where the committed pin fails on six at the base d0e3a68b7 (docs/guide.md:224, 296-297, 343-344 and 485-488, docs/install.md:30-31, docs/index.md:66-68) and passes at the head; ui/webview/dense-chrome.test.ts:241-242, now asserting the guide lacks the chat pane deep link (1 occurrence at a59c0cde3, 0 at d0e3a68b7 and at the head, so it pins the wording pass rather than this change); the interpreter note, now docs/install.md:29-30 with the command at line 18; the Sessions pane casing at docs/guide.md:16 and :91, slug unchanged; the mention sentence at docs/guide.md:31-33, true of the composer's roster and pick in ui/webview/render.ts:20194-20245; and the docstring, the word-cap message and CLAUDE.md:284-286. The one item of that note not in this change is the comment at kernel/kernel.py:62099 saying `byLogin` includes the judges' calls, outside a docs change's files.

Six items stand at this head; the one with the most consequence is the pin's exemption shape, under Tests.

- **docs/reference.md:5286 lists the token-less routes and omits two.** The paragraph opens "A few routes answer without the token" and names the liveness probes and the install files; SECURITY.md:49-55 carries the same list. In kernel/kernel.py the POST handler serves `/push/ack` at line 67503, ahead of the token check at line 67540, authenticated by the per-push pid alone, and a token-less browser open of the root path gets the paste-the-token page at line 66898 rather than a 403. A reader auditing the token gate from either page concludes every POST needs the token. Read; the paragraph is unchanged since the first commit. Either a clause naming `POST /push/ack`, authenticated by the per-push id the kernel issued rather than the token, and the bare root GET, mirrored in SECURITY.md:49-55; or an opening narrowed to the plain-GET liveness and install routes.

- **CLAUDE.md:268 states an exception the pin does not implement.** The rule excepts "a navigation line whose link carries its reason in the same clause (README's Docs section)". README.md:72 passes the pin because its final link sits inside a parenthetical ending "))." which `_ENDS_ON_LINK` at tests/test_docs_front_pages.py:49 does not match, not because of that exception. A session writing the navigation line the rule permits turns the pin red; one ending a paragraph on a parenthesized link passes the pin while breaking the rule as written. Read. Either drop that exception, since README.md:72 needs none as it stands, and say in the test's comment that the pin is shape-specific, catching neither a link inside a closing parenthetical nor one followed by anything but a period; or encode the exception as an explicit list the test reads.

### Smaller points

- **docs/guide.md:490** says SECURITY.md "gives the address for reporting a vulnerability". SECURITY.md:156-160 names a channel, GitHub Security Advisories on the repository, and no address, so a reader who goes there for one finds none. Read. "says where to report a vulnerability", or "names the private channel for reporting a vulnerability".

### Tests

- **tests/test_docs_front_pages.py:187 exempts a link-only paragraph anywhere on a governed page.** CLAUDE.md:268-269 and the description say the exemption is a line standing alone under a heading that names it, and the pin's comment at lines 50-51 says the heading above carries its reason, but `_paragraphs` (lines 93-118) drops headings at line 113 without recording them, so nothing conditions the skip at line 187 on one. Executed on a copy of this head: the skip fires once on the pages as they stand, README.md:76 under "## License"; a link-only paragraph appended to docs/install.md after a blank line with no heading passes the pin (156 subtests); the same link as the paragraph's last line with no blank line fails it ("docs/install.md line 27 ends on a link"); a link-only paragraph after the guide's backends paragraph at docs/guide.md:231 passes. The blank line alone flips the outcome, so a session tripping the pin satisfies it by inserting one, and the page still ends a thought on somewhere else to go. Verified by execution: exempting a link-only paragraph only when its nearest preceding non-blank line from `_prose_lines` starts with "#" keeps every page green, README.md:76 included, and turns both appended paragraphs red with the pin's own message (docs/install.md line 30 or 32 by placement, docs/guide.md line 231); that run is the red-before for the change. Build the lookup from the prose lines, comments and fences already dropped, so an HTML comment between the heading and the link does not defeat the exemption. Keep the two-tuple `_paragraphs` returns, with a flag or a separate helper, since the word-count helper reads it. The check is structural, "directly under a heading", not that the heading names the link, so a link-only line under any heading still passes; that is the intended shape, so the comment at lines 50-51 and CLAUDE.md:269 would say "under a heading" rather than "under a heading that names it". An explicit list of the navigation lines (README.md's License line by page and text) also closes the gap but is a list where a rule is available, needs an entry per navigation line, and moves with the README. List items are unaffected either way: the "- " prefix survives in the joined text, so a link-only list item never matches `_LINK_ONLY`, and the trailing-link check already refuses it.

- **`_paragraphs` at line 93 and the two helper fixes have no committed test that fails before them.** The file line numbers and the block-only skipping of lines opening with "<" are checked only by the description's executed red-first (line 71 where the old index said 53), which lives nowhere in the tree; the second commit refactors the helper and adds none. A later edit that drops the inline-tag distinction or returns filtered indexes leaves every page test green, since no page paragraph sits on the boundary today, and surfaces as a wrong line number in a future failure message. Read. One test over a synthetic snippet asserting `_paragraphs` yields the paragraph at its file line with the joined text, that a paragraph whose second line opens with an inline span tag stays one paragraph, and that a lone video tag yields none.

- **The description and the commit message say the pin "failed on seven paragraphs across all four pages" before the rewrites.** Against the base pages the committed pin fails on six paragraphs across three pages, the six named above, and README.md passes: its License line is exempt as link-only and its Docs line ends inside a parenthetical. Executed. A reader crediting that count believes README.md is a page the committed pin catches. "failed on six paragraphs across three pages (README.md passing: its License line is a navigation line, its Docs line ends inside a parenthetical)".

(Edited 2026-09-21 06:51Z: the first header now reads Still standing after the merge, and one word in the Tests section changed; nothing else changed.)

PR #1945, opened 2026-09-21
The reference's coldLive sentence names the deciding condition, a session nothing has parsed at its current version, and the memo-key pin also reads the counters the kernel writes by a literal subscript

lczh:
Tier: docs

The coldLive sentence #1938 added to the reference's memo passage said a session no client and no judge has parsed rides the counter every build. That names two parsers where the deciding condition is broader: the background warmer parses every living session that moved since boot or is working whenever a client is connected and no chat or timeline client is, and the feed build's own re-read parses a warm-keyed session through the shared store, so the next build's cache-only read hits. A feed-only dashboard therefore sees a moved session leave the counter after one build with no chat tab and no judge run, and an unmoved idle session ride it while a moved one does not; the sentence predicted neither, and telling design from fault is what the passage is for. It now states the condition the kernel's own comment states: a session nothing has parsed at its current version (no client's tab, no judge, no background warm, none of the card build's own parse paths) rides it every build, and the warm gate leaves an unmoved, idle session cold by design.

The pin the same PR added unions the report's keys with the literals the kernel passes to the counter helper, and its docstring claims a counter minted at a site the test process never exercises still has to reach the doc. The failed count is written by a literal subscript on the stats dict, not through the helper; it is seeded in the initializer, so the report half carries it and the pin is right today, but a future counter written that way with no seed, at a fault or eviction site the test never drives, would reach neither half, and the pin would stay green over a key the reference never names. The union now also reads every literal the kernel subscripts the stats dict with, and the docstring says so.

Validation:

- The widened scan finds eight subscript literals in the kernel today (the two map keys among them), every one already named in the passage, so the pin stays green on the current doc and the claim in its docstring now holds by mechanism.
- The perf-stats module: 66 passed. The doc's other reader, the perf heap-block module: 13 passed.

Follows the post-merge review of #1938.


ksarma:
The evidence was executed against a364f4fd0, the PR's head, before it merged as d38ccae4d; the merge carries that tree unchanged for every file cited, so each file and line reference is on that commit and holds at the merge commit.

The change does two things. The reference's `coldLive` sentence now says a session nothing has parsed at its current version rides the counter every build, and that the warm gate leaves an unmoved, idle session cold by design. The memo-key pin now also reads the literals the kernel subscripts `_FEED_MEMO_STATS` with directly. Both read true at this head.

The deciding clause agrees with the counting-site comment at `kernel/kernel.py` 42425-42428 and with the warm gate as the code has it. `_warm_wanted` (11557) admits a session that moved since boot or is working, compacting or retrying. The background warmer (35114-35145) runs when a client is connected and no chat or timeline client is, and parses only what that gate admits. `build_feed` asks for the warm under the same gate at 42578 and 43786, whose comment calls a session it leaves cold "cold by design". The `coldFlip` re-read at 42452 parses into the shared store, so the next build's cache-only read hits, and the card build's helpers also parse, so the plural "parse paths" is right.

The body's figures reproduce: the kernel has eight distinct double-quoted subscript literals (the two map keys among them), every one backticked in the passage; 66 and 13 tests collected, 79 passed together.

Three points came out of reading the source at this head; none of them was executed.

## Smaller points

- The counter dict's declaration comment at `kernel/kernel.py` 41893-41895 still carries the two-parser sentence: a session no client and no judge has parsed rides it every build, so a standing count is those cold-by-design sessions. The counting site at 42425-42428 says a session nothing has parsed. It is the one copy of the old wording left at this head; a search over the kernel, the docs and the tests finds no other. A reader at the declaration, the first place the dict is defined and where the /perf key set is written, reads the condition the PR body calls wrong, and the next doc edit that copies that comment brings the removed sentence back. Rewriting those three lines to the counting site's condition (a session nothing has parsed at its current version rides it every build; the warm gate leaves an unmoved, idle session cold by design), or replacing the clause with a pointer to the comment at 42425, closes it.

- `docs/reference.md` 2540-2542: the parenthetical lists four roads (no client's tab, no judge, no background warm, none of the card build's own parse paths) as what "nothing" means, and it reads as a roster. `_parse` (34542) has some two dozen call sites at this head. Every one of them writes the store that `_parse_cached` (35085) reads, so a session parsed by any of them leaves the count on the next build. One road outside the four: the end-on-idle pass (28657-28670) parses every living session with an armed request on each jobs pass, gated on the request rather than on movement or state. When that parse shows the turn settled, the same pass kills the session, so the lasting case is an armed session whose transcript still shows an open turn while the warm gate does not admit it (not moved since boot, not in a working, compacting or retrying state). A reader reconciling /perf against the passage then sees a count move the list cannot explain, while the sentence's last clause says the warm gate leaves such a session cold. Marking the list as examples rather than a roster fixes the wording: "(no client's tab, no judge, no background warm, none of the card build's own parse paths, nor any other reader)" or "whatever the road: ..."; the deciding clause stays as it is.

## Tests

The widening's mechanism was executed. In a scratch worktree at this head, an unseeded double-quoted write of a new key was added in `_feed_memo_drop` (41987), an eviction site the test process never drives. It turns `test_every_feed_memo_report_key_is_in_the_references_memo_passage` red naming that key. The previous test file over the same mutant stays green, so the new scan is what catches it. A setdefault-plus-subscript shape is red as well.

- `tests/test_perf_stats.py` 463 and 453-454. The new scan reads `_FEED_MEMO_STATS\["(\w+)"\]`, keyed on the double-quote spelling. The docstring says every literal the kernel subscripts the stats dict with directly, which is wider than what the regex reads. A future single-quoted subscript at a site the test process never drives reaches GET /perf as a key the passage never names, and the pin whose docstring says it catches exactly that stays green. So does a `.setdefault("x", 0)` with no subscript and no helper call. Neither shape exists at this head: the eight direct subscripts are all double-quoted, and the one `.get` on the dict with a literal, at 42022, sits beside a subscript the scan reads. One way to close it is a quote-agnostic scan, `re.findall(r'_FEED_MEMO_STATS\[(["\'])(\w+)\1\]', src)` taking group 2. Another is to walk the module's AST for Subscript nodes whose value is the stats dict's Name and whose slice is a str Constant; that walk also covers setdefault and get calls with a str first argument. The other direction is to narrow the docstring to double-quoted subscripts, which matches the bound the helper scan on 462 already has.

(Edited 2026-09-21 06:20Z: the opening now names the merge commit, which landed while this note was being drafted; nothing else changed.)

PR #1946, opened 2026-09-21
The judge experiment's harness and design note after two contributors' reviews: the store copy re-keyed and cut at the turn's start, scored cards, failures, the goal-history map, the opener rule, required flags

henrypinkard:
> Correction (2026-09-21): this body describes an earlier head. By the merged head, `session_endings` DERIVES the turn boundaries from kernel/event_model.py loaded against a scratch root (never the live roots), not a copied opener rule; and the module carries 26 tests, not the count quoted below.

The judge experiment's harness and design note after two contributors' executed reviews (on #1932 and #1929) and the pilot's stop: every store copy re-keyed to the ending and cut at the turn's start, the measures scoring only the ending's own cards, failures marking a row not comparable, the closer's goal-history map, the event model's opener rule, the registry's transcripts, required flags, the caller's event model left alone, tier one keyed on events, and the plan's column rule and settled gate corrected. Fix, red first at main. The first commit amends the plan; the second folds the harness.

**What the reviews found, each fixed and pinned.**
- The seed kept the live session id and its placement keys, so no placement matched the arm's segment ids and the planner re-planned the whole history on every build (12 calls over a three-ending corpus); the cut at the turn's END kept the live planner's node from the ending's own turn and the closer's verdict on it; the sealing fields sealed the seeded top out of the menu and a twin was minted. Now `store_before(store, cut, start, ending_id)` re-keys `rompUuid` and every id prefix (`deep_rekey`: node ids, placements, closed turns, trails, and the journal lines the builder copies), cuts at the turn's start (nodes born and verdicts filed from the turn on are gone, by `ev_t` then `at`), keeps the turn's own prompt-run placement (`#p`, a delegation's `#d`) and its node, drops the work-run placement so the arm plans the turn once, drops the flag cache and the sealing fields (`nodeComplete`, `blocked`, `cleared`, `doneWhy`, `settledAt`, `settledDone`, `rolledUp`, the summary fields), drops gate stamps from after the cut (`blockCheckT`, `blockCheckDoneT`, `closerLookT`, `awaitingAt`), drops nodes minted from dropped segments, runs the orphan pass to a fixed point and drops a `lastNode` that points at nothing. Pinned on a hand-built judge-shaped store and, end to end, on stores the judge module itself wrote over the synthetic sessions with the fake as its model: one planner call per ending per build (8 over four endings and two builds), one scored card, no twin.
- Leaks and false interrupts counted columns that were there before the cut, so every arm saturated alike. The arm records per card whether the ending SCORES it (born, or given a `done` or `block` verdict, in the turn); `measure` counts leaks, false interrupts and flaps over scored cards only, `cleared` is its own bucket, and an older results file still counts every card. Pinned on a hand-built results dict (an inherited blocked card counts nothing; the finished ending's own blocked card is the one false interrupt; the card that read needs_input then working is the one flap).
- A failed call or a reply the parser rejected left no trace, so a prompt that broke the parser scored the perfect row. The arm counts `judge-errors` rows of the failure kinds (`parse`, `give-up`, `pass-crash`, `call`, `timeout`, `history-unreadable`, `store-quarantined`; the judges' anomaly notes such as a stale close are verdict facts, not failures) and closer returns of nothing; `failures` rides results.json, `measure` and the table, and a row with any reads "not comparable". Pinned with a fake answering prose to the candidate arm.
- The closer got no goal-history map: the arm builds it as `_close_session` does and passes it; every closer call in the arm tests carries the section.
- `turn_ends` ended a turn only on a string-content user record, so the SDK composer's text-block prompts never ended one: the builder now uses the event model's opener rule (a non-meta user record with text opens a turn; a tool result or a meta record does not), copied so it loads no romp module against the live roots.
- The builder read the sid's own transcript only: it enumerates the registry's transcripts (`lastSid`, the /clear episode heads, both ends of a resume fork) and counts its skips in the manifest, never naming a session.
- `topsBefore` wrote sid-prefixed node ids into the manifest: suffixes now, pinned with a seeded store.
- `run` and `run-arm` dialled a binary named `claude` with no stop: `--claude-bin` is required on `run`, `run-arm` and `label`, `--budget-usd` on `run` (`inf` for none); `run` and `run-arm` without them exit non-zero; the docstring's cost claim is reworded.
- The in-process loader re-executed the shared event model under the arm's roots: it swaps the caller's module out of `sys.modules` around its loads and puts it back; the test pins the shared module's identity and its `STATE`.
- The flag cache stripped from the seed made the planner's first menu list done subs and cleared tops: the arm rolls the seed up and saves it before the first judge of each build.
- Tier one keyed "finished" on a seven-day window where the journal has events: it keys on events (a hand clear or a resolve with no later followup, unclear or restore at build time) and records the observation span per ending; a followup nine days later reads not finished. Tier-one-eligible endings are picked oldest first, so the user had time to act.
- The design note: the rollup emits `completed`, `blocked`, `working` or `cleared` per top, no `awaiting` value; the harness computes the store-derivable part of the column rule with `cleared` its own bucket and names the live-only inputs; measure one names the settled gate (`_session_settled` over the ending's own truncated transcript, the reviewer's option c) and the arm passes it to both rollups; the store cut, the scored cards, the failures and the task store (not copied; the arms reconstruct the to-do state from the transcript) are stated.
- Smaller: the docstring's `report` usage names `--corpus`; the fifth-over budget rule is pinned against the bare-budget mutant; the figure road is pinned both ways (a stub library writes the file; no library leaves the note); the per-class cap and the per-session spread run on a fixture big enough; each ending's class is asserted; a pre-cut journal row is asserted kept and re-keyed; the run-root refusal is pinned on the subprocess entry too.

**Tests** (tests/test_judge_experiment.py, 13; the fake answers by judge and by the candidate marker, logs the goal-history flag and the planner's menu, and answers prose on request). Red first at main (d0e3a68b) by copy-aside: eleven reds, all on assertions: the planner's calls over the judge-written stores `12 != 8`; the store cut `store_before takes the turn's start and the ending id` (the base's took the cut alone); the arm's measures `(4, 3, 0, 0, None) != (4, 3, 0, 0, True)` on the comparable mark; the hand-built measures `(0, 0, 1) != (0, 1, 1)` (the base counted an inherited card and missed the ending's own); the prose fake `0 not greater than or equal to 4` closer rejections counted; the composer prompt's turn start `None != 1700000000`; the registry reader `unexpectedly None`; the skip counts `None != {...}`; the required flags `SystemExit not raised`; the shared event model's roots moved under the arm's scratch root; tier one `None != 'not finished'` on a followup after the judges' done. Two tests pass at main as pins should (the selection's cap and spread with the eligibility preference, which 1937 landed; the report's figure road).

**Suites at the head** (6805d9bd): the harness module 13 passed under the pinned runner. No file under kernel/, ui/ or vscode-extension/ changes.


ksarma:
The evidence below was executed against 98880abf3, the PR's head, merged as be724edb5 with the three files unchanged; file and line references are on that commit (scripts/judge_experiment.py unless said otherwise). PR #1956, the author's continuation, merged since (2d20267d8, head 6b6c920f7); every item is read against that head: closed items come last, partly closed ones say what stands, and a moved line's number there is parenthesised.

- **The archive union takes the rewind path's nodes too.** #1956 closes the invisible cleared tops: `store_with_archive` (396) feeds all four readers and a pre-turn clear rides in the seed (test 959); test 844 asserts the done time, label and seed, not `tierOneEligible` or `topsBefore`. Standing: the union takes every archive node (411 to 412), while the same file carries nodes the rewind path swept, whose ids sit in the live store's `rewindSwept`; skip those ids and union only the cleared subtrees.

- **A title-matched lane's endings are seeded, made eligible and labelled from the anchor's store and journal.** `build_corpus` reads dones (423; 463) and cuts store and journal under the sid for every transcript, while the judge keys a lane's store under the lane's stem. Executed: both lane endings were seeded with the anchor's `g1` and labelled, if at all, from its card. Fix: key store and journal per lane inside the fsid loop (the stem for a title lane, the sid otherwise); `store_before` re-keys from `rompUuid`, and the journal cut (491; 528) reads `overrides/<lane key>.jsonl`. Keep `session` as the anchor's hash (test 757) beside a new `lane` field, have `label` (864; 922), which reverses hashes through `names/` alone, index the `goals/*.json` stems, and pin eligibility, seed and label from the lane's own store.

- **`session_endings` makes an ending of every ended turn.** Lines 180 to 184 (179 to 183) emit every turn the event model marks ended, and `build_corpus` classifies and truncates at the last atom. Bare slash commands, idle compactions and interrupted turns become "finished" endings, and an absorbed task notification cuts the prior turn's ending there while the reply classifies otherwise; the cost is finished-class slots and the arm's `endings` denominator. Fix: filter the turn's atoms with the judges' own `_seg_command_worked` (an assistant record with text and no `command` flag, keyed on the record: a `/usage` stdout is an assistant atom but a user record), skip a turn with no such atom or an interrupt tail, keep the cut at the last atom (cut at the last assistant record, an interrupted turn reads open to the judges), have `classify_ending` read the last non-command assistant text, and keep a compact-boundary first atom (a continuation carries real work).

- **`in_turn_window`: what stands.** #1956 ends the window at the cut (290), deletes `SETTLE_S`, rewords the docstrings and note line 33, re-stamps test 294's done. Standing: `FALLBACK_TURN_S` (44 (43), 289) is unreachable from the builder (`session_endings` never yields a None start; a falsy one opens the window to the epoch), so delete it and mark a None start ineligible; and the closer dones at test 299 and 644 still carry `ev_t` at the cut, which the closer never writes.

- **The restore matcher: what stands.** #1956 matches a restore row by its `nodes` dict (851 to 855; test 886). Standing, the larger part: mutants dropping the unclear op or the src filter passed all 26 tests; no fixture at 6b6c920f7 covers either. Add writer-shaped fixtures with full ids: an unclear alone, a clear with src romp (the kernel's value, not kernel) or nudge alone, expecting None; a resolve, expecting "finished".

- **One exception inside an ending's build aborts the arm with nothing recorded.** `run_arm_inprocess` (606; 643) has no per-ending guard, writes results.json only after the loop and calls the judges directly, so the pass-crash rows `FAILURE_KINDS` counts are never written, and `run` stops before its remaining `--arm` specs. Executed with a raise on one ending: no results.json, `failures` 0. Fix: try/except Exception around each ending's builds, the row filed through `jd._log_judge_error(<judge>, eid, 'pass-crash', note=repr(e))`, `errs0` taken before the handler (652; 689), results.json written after every ending and in a finally block, and `check=True` kept, since after the guard it raises only on test 467's run-root refusal. Resume is a separate change: lines 593 to 595 (630 to 632) remove the arm's state, usage and error files included.

- **The description no longer matches the head.** It says the opener rule was copied so no romp module loads against the live roots; `_event_model` now loads kernel/event_model.py against a scratch root and delegates to `parse_session` (keep the live-roots half, replace "copied"). Its counts are 6805d9bd5's (the head at opening; 26 tests at 98880abf3, not 13), and the oldest-first test passes at main, so it pins main's behaviour restored rather than a fix.

## Smaller points

- Same-named sessions duplicate each other's endings: the lane discovery (249; 248) excludes only the session's own fsids and `build_corpus` carries no seen set across the names loop, so each lists the other's transcript as a lane (executed: 8 endings for 4 turns); exclude every registered sid and every path an earlier session claimed (each session's own `known_fsids` first).

- Per-session queue order is by stem, not time: the candidates at line 456 (496) are in insertion order, `sorted(fsids)` then turn index, so the oldest-first and newest-first orders are true per lane in stem order (executed: each face picked the wrong lane's ending); sort each session's per-class list by `cut_t` before the reverse-and-pop step.

- A faulted store or registry reads as absent: #1956 makes a corrupt store or archive raise and counts it (`store-unreadable`, 460; `tierOneErrors`, 947); standing are the labels.json row with no `tierOneError` (937), a non-object top level (405, 411), and `known_fsids` (266; 265, 273) swallowing an unparsable registry or an unreadable episodes or states record; distinguish ENOENT and ENOTDIR from other errors and count the rest under their own `skipped` key (`custom_title` is not part of this: its twin swallows `OSError` too).

- A lift rider's done is unscored: the `scored` stamp keys a verdict on its evidence time (648; 685 to 687), but the closer anchors a rider's done or block at the lift's own `ev_t`, so a rider completed from an earlier turn's lift reads completed unscored, its leak invisible to `measure`; score by the arm's own filing (`at` at or after the build's start) and name the rider residual in the docstring (320; 326) rather than cutting rows by `at`, which also removes late filings for earlier turns.

- The note's line 20 says each ending carries "the card placements the live judges filed for it", but the corpus writes no such record; the label pass reads them from the live root through `--live-state`, so tier one exists only while that root lists the session (line 34 is reworded by #1956).

## Tests

These mutants passed all 26 tests, and no pin at 6b6c920f7 covers them: eligible = bool(dones) for line 444's (484) `in_turn_window` check, whose cost is eligible slots since `tier_one_label` applies its own window (pin `tierOneEligible` False for SIDS[0] turn 1); `--budget-usd` or `--claude-bin` made optional (899, 906; 960, 967), the pins covering the both-missing shape only (one call per missing flag, against an absent corpus path); flaps counted over every card or `cleared` dropped from `COLUMN_OF` (pin e2's unscored `g1` flapping at 395; `g3`'s cleared column from `run_arm` at 679); the comment at line 457 (497): pin per_class=2, the ineligible slot going to sid3 turn 2; and the ended-turn filter at 180 (179): pin a transcript ending on `tool_use`, one ending. The fixtures' `mint` log rows (test 292, 298, 664, 688 and newer) are rows no kernel writer emits, so the keep-clause at line 347 (356) matches no real row: birth fixtures with log: [] and drop it.

## Closed since by #1956

- The labeller's second class order could equal the first: lines 933 to 934 fall back to the reversed list on the identity shuffle; test 916 asserts the orders differ.
- `delegLookT` survived the cut: now in `DROP_STAMPS` (298); pinned by key absence at test 910.
- The agreement gate was pinned at 100 percent only: test 1038 asserts (50.0, False) on a disagreeing ending.
- The gate-stamp drop was pinned by key absence: test 1007 seeds `blockCheckT` at the turn start and runs `_blocked_sub_candidates`, the due formula restated at 1033, not read from an unblocker call.

PR #1947, opened 2026-09-21
The row-projection pin's messages say a field is read, not mentioned, and the census test exercises the f-string and percent-format shapes its helper's docstring claims

lczh:
Tier: docs

Two test-only points from the post-merge review of #1939, both in the chat-build signature inputs module; the kernel is untouched.

The two failure messages behind the row-projection pin still said a dropped field "is mentioned" outside or beyond the permitted regions, while since #1939 the assertion above each fails only on a string constant equal to the field, a read. A red therefore named the wrong thing to remove: a kernel change that adds a get of the raw token count in a helper printed that the field was mentioned, and whoever reads that at face value rewords nearby prose, which is what an earlier change did under the old word count, and the red stays. The messages now say the field is read outside, or beyond, those regions; they already print the offending line numbers.

The census helper's docstring claims one counted shape past the plain reads, a real subscript inside an f-string, and four uncounted ones (a percent-format key, a format field, its mapping variant, a keyword argument). The synthetic body the census test parses exercised none of the five, so those claims had no assertion behind them, and the f-string case has a moving part: the module runs on four interpreters, and a release that moves f-string sub-expression nodes could drop such a read while the pin stayed green and the docstring still said the shape is counted. The synthetic body now carries an f-string subscript of one field and a percent-format key of the same field; the expected set for that field gains the f-string line, and the percent line is in neither set, so an interpreter that moves those nodes or a helper change that starts counting placeholders fails the test rather than the docstring.

Validation:

- The module: 66 passed, 10 subtests passed, the same count as before; the two new lines are asserted inside the existing census test.
- The counted f-string shape and the uncounted percent shape are now executed on every interpreter the module runs under.

Follows the post-merge review of #1939.


ksarma:
The evidence below was executed against cbc519519, the PR's head, before it merged as 17f4dd78c; the merge carries that tree unchanged for the test file and for kernel/kernel.py, so every file and line reference is on that commit and holds at the merge commit. The change is confined to `tests/test_chat_build_sig_inputs.py` and does two things. The two failure messages behind the row-projection pin, at lines 467 and 469, now say a dropped field is read outside or beyond the permitted regions, where they said it was mentioned. The synthetic body the census test parses gains an f-string subscript of `ctxTokens` at line 492 and a percent-format key of the same field at line 493, and the expected set for that field at lines 498 and 499 gains the f-string line. The helper's source is byte-identical between the base and this head, so the PR is a message reword plus a pin over unchanged code.

**Still standing after the merge**

- The reworded messages say a field "is read", and the assertion above each fires on any string constant equal to the field. The docstring of `_key_reads` at lines 350 and 351 says a write or a listing that names the key counts too, and that this is the pin's intent. So a kernel change that writes the dropped field as a key in a new helper, or lists it in a tuple, turns the pin red with a message that says the field is read there, and whoever takes "read" literally looks for a get or a subscript load at the printed line and finds a store or a listing. This is the same class as the "mentioned" wording the PR replaces, and much narrower: the line number is printed, and the docstring states that writes count. One fix is to say what the predicate tests, "lastTool is named as a key outside the projection and the key" with the tail of that message unchanged, and "ctxTokens is named as a key beyond the projection, the key, the compaction tick and the merge writing it"; a form like "a string constant equal to lastTool lies outside" does the same. Wording only, with no assertion change. This point comes from reading the source against the docstring, not from an execution.

**Tests**

- The module at this head: 66 passed, 10 subtests passed, in a single process. The untouched module at the base also gives 66 passed, 10 subtests, so the body's "same count as before" matches. The head's test file copied onto a detached checkout of the base is green at the same count: the PR changes no helper, so no red-before exists or is owed.

- Both messages match the permit sets the assertion above each uses. For lastTool the set is the projection block plus `_chat_build_sig`, which the message renders as "outside the projection and the key". For ctxTokens the set adds `_compact_suggest_tick` and the merge that writes the field, which the message renders as "beyond the projection, the key, the compaction tick and the merge writing it". That merge is `Sessions.live`. The pin prints the offending line numbers, as the body says: a probe that appended a get of the ctxTokens field past the end of the kernel source made the assertion print `Lists differ: [71325] != []`.

- By mutation of the helper. A variant that skips constants inside `ast.JoinedStr` turns exactly one test red, the census test, at the ctxTokens assertion on lines 498 and 499, with "Items in the second set but not the first"; the row-projection pin and the other seven Census tests stay green. A variant that also counts a constant containing the percent key `%(ctxTokens)` turns the same single test red at line 498, with "Items in the first set but not the second". Both match the body's claim that an interpreter moving f-string nodes, or a helper that starts counting format keys, fails this test rather than the docstring.

- The other three shapes the docstring says are uncounted are still unpinned, which the body states plainly: a variant that also counts a constant containing `{ctxTokens}` or the same brace form of lastTool, plus any keyword argument whose name equals the field, left all nine Census tests green. That variant covers the two brace-format shapes, `.format` and its mapping variant, and the keyword-argument shape. Those three claims rest on the docstring alone. Under the percent and brace variants the row-projection pin on the real kernel also stayed green, consistent with the docstring's statement that the kernel has none of these shapes today.

- Interpreters. The helper's exact logic run against the exact synthetic body, not the test module, under CPython 3.10.20, 3.11.15, 3.12.3, 3.13.14 and 3.14.6 gives the same sets on all five: ctxTokens at lines 6, 8 and 11 of the body, so the f-string subscript is counted; lastTool at lines 5 and 7; the percent line, 12, not counted. When last checked, the four Python CI jobs on this head had passed and the extension job had not finished.

- The `line` fragments in the census test are unambiguous in the body: the bracketed double-quoted fragment first matches the subscript line and not the single-quoted f-string; the f-string fragment matches only its own line; the percent line matches no expected fragment, so equality against the exact set pins it as uncounted.

- Neither helper has a reader outside this module (the other grep hits for the two names are three test method names that contain the substring, none a call), and the old "is mentioned outside" and "is mentioned beyond" text appears nowhere else, so the reword leaves nothing stale.

PR #1948, opened 2026-09-21
The docs workflow pins read the condition alone and anchor the concurrency lines

henrypinkard:
A follow-up on PR 1934 from the contributor's post-merge review (2026-09-21): three pins in `tests/test_docs_workflow_pins.py` were satisfied by a comment. The captures of the upload step's and the deploy job's `if:` lines took the rest of the line, trailing comment included, so a guard moved from the condition into a trailing comment kept the module green; and the two concurrency asserts searched the whole file unanchored, so a shared `pages` group with the old expression left in a comment above it passed too. In either state a pull request's run would upload the site, or reach the deploy job with `PAGES_ENABLED` on, or share the publish's group, and the module said nothing.

**The fix, test only.** The condition is read alone: a trailing comment is stripped from each capture before the assertion (`re.sub(r"\s+#.*$", "", capture)`); the two concurrency asserts are anchored at the line's start and end (`(?m)^  group: <the expression, escaped>$`, the same for `cancel-in-progress`); and the shape messages say the step-then-condition shape was not found, since the regex fails for a missing step as well as a missing condition. No workflow or site change rides this PR.

**Red first by the reviewer's mutants at 17f4dd78** (main; the workflow mutated one way at a time in a detached worktree, the guard's text kept as a comment and the guard gone): under `--python 3.13` and under `3.14` the same, `1 failed, 3 passed` per mutant, the named pin each time: the upload's guard moved into a trailing comment, `"github.event_name != 'pull_request'" not found in 'true'`; the deploy job's, `not found in "vars.PAGES_ENABLED == 'true'"`; a shared `pages` group with the expression in the comment above it, `Regex didn't match: "(?m)^  group: ..."`. Main's pins stayed `4 passed` on every one of the three mutants, the hole the review found. Green at the head on the unmutated workflow: `4 passed`; the module is test-only, so no suite beyond it and the front-pages module is touched.

Tier: fix


ksarma:
The evidence was executed against 1ed5b45bb, the PR's head, merged as 034c1bcb4. The merge carries the head's tree unchanged for `tests/test_docs_workflow_pins.py`, so every file and line reference here is on 1ed5b45bb and reads the same at the merge commit. The change is test-only: a helper strips a trailing comment from each `if:` capture before the assertion, and the two concurrency asserts are anchored at the line's start and end.

The body's red-first reproduces. Executed with docs.yml, which this PR leaves unchanged, mutated one way at a time in a scratch copy (the upload guard moved into a trailing comment, the deploy guard moved into a trailing comment, a shared `pages` group with the old expression in a comment above it), the module at this head is 1 failed, 3 passed per mutant on the named test, with the messages the body quotes, and the module as it was at 17f4dd78c stays 4 passed on all three. The same result is repeated under Python 3.10, 3.13 and 3.14 via unittest. On the unmutated file the module is 4 passed, and 10 passed with 146 subtests together with the front-pages module. Two further mutants: the cancel-in-progress setting emptied the same way, the expression in a comment above and false on the line, is red at this head and green at 17f4dd78c, so the second anchored assert closes a hole of its own; the deploy guard moved into a comment line above its `if:` is red at both, since the base's capture never took a comment line above.

Of the post-merge review comment on PR 1934, the item about the three pins is closed by this change, with the message suggestion taken, and the sentence in PR 1934's body about pushes to main now reads as that comment proposed (the body was edited on 2026-09-21 at 06:12Z), so nothing of that comment stands.

## Tests

- The two anchored concurrency asserts at lines 50 and 51 refuse a benign trailing comment on the setting's own line, and a red from either prints the whole source with no message. The `$` anchor comes directly after the escaped expression, so a comment appended to the setting's line is a regex miss. Executed in a scratch copy: a comment appended to line 42 of docs.yml (one publish at a time, say), a no-op for GitHub, turns the concurrency test at this head red, and the message is a regex miss quoting the whole file, with nothing in it naming the comment as the cause; the same holds for line 43. The `if:` pins at lines 43 and 46 strip a trailing comment by design. The strip in `_expr` at line 38 does that, so the same module treats a trailing comment two ways. The comment at line 49 says the anchor exists to refuse the expression in a comment above a shared group, which this red is not. This is not a hole in the pin: no mutant of the setting itself passes at this head, and the anchored shape is the one the review comment on PR 1934 proposed, so the point narrows that proposal rather than the change. The fix is the shape the `if:` pins already use: capture the line and compare the condition alone. Search for the group line with `re.search(r"^  group: (.*)$", self.src, re.M)` and assert the match with a message naming the missing group line. Then compare `self._expr(grp.group(1))` to the expected expression with an equality assertion, and the same for cancel-in-progress. Executed in the scratch copy with that shape: the clean file is 4 passed, and the benign-comment copy is 4 passed. The shared-group mutant is red with `'pages' != "${{ github.event_name == 'pull_request' [46 chars]' }}"`. The emptied cancel-in-progress mutant is red with `'false' != "${{ github.event_name == 'pull_request' }}"`. A red then names the setting's actual value against the expected one instead of printing the file.

PR #1949, opened 2026-09-21
The row-projection pin's messages say a dropped field is named as a key, which is what the census tests, not that it is read

lczh:
Tier: docs

One wording point from the post-merge review of #1947, in the chat-build signature inputs module; the kernel is untouched.

The two failure messages behind the row-projection pin said a dropped field "is read" outside or beyond the permitted regions. The assertion above each fires on any string constant equal to the field, and the census helper's docstring says so: a write or a listing that names the key counts too, on purpose, because the pin hunts the kernel naming the field at all outside the regions that may. So a kernel change that writes the field as a key in a new helper, or lists it in a tuple, turned the pin red with a message that said the field was read there, and a reader taking "read" literally would look for a get or a subscript at the printed line and find a store or a listing, the same class of misdirection as the "mentioned" wording #1947 replaced, much narrower since the line number is printed. The messages now say the field is named as a key outside, or beyond, those regions; the second message's tail is unchanged.

The review of this change found one more message in the same test, and the tail of the first, describing a hit as a read over the same predicate: the tail of the first message called the printed line a reader, and the per-reader loop's message said no chat reader may read the field. The tail now states the risk without calling the line a reader (a reader there would serve a stale payload under an unchanged key), and the loop's message says no chat reader may name the field as a key.

Validation:

- The module: 66 passed, 10 subtests passed, the same count as before; no assertion changed, only the three message strings.

Follows the post-merge review of #1947.


ksarma:
This evidence was executed against 2f6942ce6, the PR's head; every file and line reference is on that commit, in `tests/test_chat_build_sig_inputs.py` unless another file is named. The PR has no earlier comments or reviews. The diff is exactly the two message string literals at lines 467 and 469, one file, +2 -2: no assertion, helper, docstring or import changed, and a diff of the same range over `kernel/` is empty, so the kernel is untouched as the body says.

The head of each reworded message now says what the assertion above it tests; this is from reading the source, and the probe under Tests confirms the counted shapes by execution. Each assertion calls `_reads_outside`, which drops the census lines that fall inside the permitted ranges. The census itself is `_key_reads`, whose body at lines 356 to 358 counts every string constant equal to the field. The docstring of that helper at lines 350 and 351 says a write or a listing that names the key counts too, and states the pin's intent: the kernel naming the field as a key anywhere outside the regions that may. So "named as a key" is the docstring's own phrase.

**Smaller points**

Two messages in the same test still describe a hit as a read over that predicate. Both points come from reading the source against the docstring, not from an execution.

- Line 467: the reworded lastTool message ends with an appositive that labels the hit "a reader", while the assertion above it fires on any string constant equal to the field, a store or a listing included. A kernel change that adds `row["lastTool"] = x` in a helper outside the permitted ranges turns the pin red at that line with a head that says the field is named as a key, which is right, and a tail that calls the printed line a reader, which is wrong for a store. Taken literally, the tail sends the search to a get or a subscript load at that line, where the hit is a store. One fix, wording only with no assertion change, rewords the tail so it states the risk without calling the printed line a reader, for example that a reader there would serve a stale payload under an unchanged key; dropping the appositive does the same.

- Line 457: the third census message in the same pin, the per-reader loop at lines 453 to 458, still says "no chat reader may read it; %s does" over the same predicate, `_key_reads(own, k)`. So a store or a listing of the field in a named builder's own source prints that the builder reads it. If `_bg_tasks` gains a store of the field while assembling a payload from another source, the pin prints that it reads lastTool, and the search for a get in that function finds a store. This message is outside the diff, in the same test and of the same class as the two the PR rewords. One fix, wording only with no assertion change: "%s is dropped from the row component, so no chat reader may name it as a key; %s does". A later PR is fine for it, since the body scopes this one to the two messages the post-merge review of #1947 named.

**Tests**

- The module at this head, in a single process: 66 passed, 10 subtests passed. The module at the base 73757164b, before the change: 66 passed, 10 subtests passed. The body's figures and "the same count as before" match.

- The predicate by execution, with the helper body copied verbatim and run over a synthetic body: it counted a store, an entry of the tuple a loop reads through, a dict-literal key, an equality comparison, a bare positional string argument and a get, and it did not count a keyword argument or a prose literal. Two of the counted shapes, the equality comparison and the bare argument, name the field but not strictly as a key; the kernel has neither today, so the looser wording misdescribes a hit only if a future kernel change adds one of those two shapes outside the permitted ranges. The one phrasing exact for every counted shape is "a string constant equal to lastTool lies outside", and this PR does not need it.

- Nothing else pins the old or the new strings: a search over the Python, Markdown, TypeScript and text files in the tree finds them only at lines 467 and 469, and no doc mentions these messages.

lczh:
Executed against 85832eef, the new head. Both smaller points are folded, wording only, no assertion changed: the first message's tail no longer calls the printed line a reader (it says a reader there would serve a stale payload under an unchanged key), and the per-reader loop's message says no chat reader may name the field as a key. The module stays at 66 passed, 10 subtests, in one process.

ksarma:
This evidence was executed against 85832eef4, the PR's head; file and line references are on that commit, in `tests/test_chat_build_sig_inputs.py`. The commit closes both points from the earlier comment, by execution: a store of the field appended past every permitted range turns the pin red at line 467 with a tail that no longer calls the printed line a reader, and a dict-literal store of the field inserted into `_session_retrying` fires the loop at line 457 with a message true of a store. The test file's AST at the base 73757164b and at this head is identical once string constants are neutralised; exactly three differ, each an assertion's message, so no assertion changed. The module in one process: 66 passed, 10 subtests passed, as the comment above and the Validation line say.

**Smaller points**

Two sentences in the PR body did not move with the commit; both fixes are wording only.

- The body still says the messages were reworded "with their tails unchanged". The commit rewrote the tail of the lastTool message at line 467, the body's next paragraph says so, and the commit message dropped the phrase. A reader taking that sentence at its word expects the tail at line 467 as #1947 left it; the file carries a different one. One fix: delete the phrase, as the commit message does, or say the ctxTokens message's tail is unchanged.

- The Validation line still says "no assertion changed, only the two message strings", while the whole diff changes three, at lines 457, 467 and 469, and "two more messages" in the paragraph on the review of this change counts the line 467 message again, for its tail. A reader checking the Validation line against the diff finds a third changed literal the body says is not there. One fix: "only the three message strings" there, and "one more message in the same test and the tail of the first" in that paragraph.

**Tests**

- The head is one commit on the base 73757164b, replacing 2f6942ce6; the diff over `kernel/` is empty, and against 2f6942ce6 the delta is one file, +2 -2.

- The message at line 469, unchanged since 2f6942ce6, prints on a store of ctxTokens appended past every permitted range, with a head true of a store. Nothing else in the tree pins the old or the new strings.

PR #1950, opened 2026-09-21
The memo counters' declaration comment states the coldLive condition the counting site states, the reference's roster reads as examples, and the memo-key pin reads both quote styles and setdefault

lczh:
Tier: fix

Three points from the post-merge review of #1945: two wordings and one pin, no behavior change.

- The counter dict's declaration comment in the kernel still carried the two-parser sentence #1945 removed from the reference (a session no client and no judge has parsed rides coldLive every build). It was the one copy left, at the place the /perf key set is written, and the next doc edit copying it would have brought the removed sentence back. It now states the counting site's condition: a session nothing has parsed at its current version rides the counter every build, whatever the reader, and the warm gate leaves an unmoved, idle session cold by design.
- The reference's parenthetical listed four roads as what "nothing" means and read as a roster. The parse helper has some two dozen call sites, every one writing the store the cache-only read consults (the end-on-idle pass among them, gated on an armed request rather than on movement or state), so a reader reconciling the counter against the list could see a move the list did not explain. The list now closes with "nor any other reader of the shared store"; the deciding clause stands.
- The memo-key pin's source scans were keyed on the double-quote spelling of a subscript and of a helper call while its docstring claimed every direct write and every helper literal; a single-quoted subscript, a setdefault or a single-quoted helper call at a site the test process never drives would have reached the report unnamed with the pin green. Both scans now read either quote style, the subscript scan also a setdefault, and the docstring says so.

The review of this change added the helper-call half of that widening and two wordings: the declaration comment's pointer now says the counting site states the condition (it names no reader), and the reference's clause names the parse store ("nor any other road that parses through the parse store") rather than the goal store's shared cache.

Validation:

- The widened scans find the same eight subscript literals and four helper literals in the kernel today (all double-quoted), every one named in the passage; a probe over a snippet with a single-quoted subscript, a setdefault and a double-quoted subscript returns all three keys, and one over a single-quoted and a double-quoted helper call returns both.
- The perf-stats module: 66 passed. The heap-block module, the doc's other reader: 13 passed.

Follows the post-merge review of #1945.


ksarma:
The evidence here was executed against 0b8d6d4d3, the PR's head, and every file and line reference is on that commit. The three changes read as the description states them: the declaration comment now carries the counting site's condition, the reference's parenthetical closes as examples rather than a roster, and the subscript scan reads both quote styles and a setdefault. One point comes from execution; two smaller ones come from reading.

## At this head

- tests/test_perf_stats.py line 463, the helper-call scan. The PR widened the subscript scan on line 464 and left the scan on line 463 byte-identical to the base: `_feed_memo_count\("(\w+)"` reads the double-quote spelling only. The docstring on lines 452 to 456 says the key set is the report's unioned with every counter literal the source passes to the helper, an unqualified clause, and closes with the principle that a scan keyed on one quote style would miss the other. The phrase "in either quote style" on line 454 sits between the subscript clause and the setdefault clause, so grammatically it attaches to the subscript scan and the docstring is ambiguous rather than plainly claiming both scans; the unqualified helper clause and the closing parenthetical are what state the wider claim.

  Executed in a scratch copy of the head: a single-quoted `_feed_memo_count('probeHelper')` appended as the last line of the memo drop function after its with-block (kernel/kernel.py line 42021, a site the test process never drives) leaves the pin green, 1 passed, with the head test and with the base test alike. The same line double-quoted turns it red with `Lists differ: ['probeHelper'] != []`, and a single-quoted subscript and a setdefault at the same site each turn it red naming the key at this head while the single-quoted subscript passes at the base, so the widening does what the description's third bullet says and the gap is on the other scan. The gap predates the PR, since the base regex is the same text and the base test is green under the same mutant; this is the class the third bullet describes, left standing on the helper half, not a regression. No kernel call site is single-quoted today: the four helper literals are coldLive, coldFlip, hit and derived, all double-quoted, so nothing in the /perf report is misnamed now and the defect is the pin's coverage. Nothing else pins the spelling: the repository has no formatter or quote lint (no pyproject.toml, ruff.toml, setup.cfg, .flake8 or pre-commit configuration, and nothing under .github names one), kernel.py already has single-quoted call literals of the `('word')` shape by the hundreds against several thousand double-quoted, and no other test scans the helper.

  Consequence: a future single-quoted call of the helper at a site this process never exercises, an eviction or a forget, mints a builds.feed.memo key the reference never backticks, and the counters reach the report through `out = dict(_FEED_MEMO_STATS)` on line 42077 once the site runs, while the test whose docstring describes exactly that case stays green.

  Fix: replace the pattern on line 463 with `_feed_memo_count\(\s*["'](\w+)["']`. Verified in the same scratch copy: the unmodified pin stays green, the single-quoted and the double-quoted probe both fail naming probeHelper, and over the unmutated head source the old and the new pattern both return exactly coldFlip, coldLive, derived and hit (the only other call shape in the file is the def's `key, n=1`). Widening the scan is preferable to narrowing the docstring, since narrowing would leave the gap the docstring's own parenthetical names. Once both scans read both styles the docstring can stand as written, or "in either quote style" on line 454 can be made to cover both, for instance as "each read in either quote style".

## Smaller points

Both from reading, not execution.

- kernel/kernel.py line 41918. The new parenthetical says the counting site in `_feed_session_key` "says which" reader parsed. The counting site's comment on lines 42452 to 42455 states the condition (a session whose parse the cache lacks at its current version rides the count every build) and names no reader, so the pointer promises an enumeration its target does not carry. The condition itself matches: the count fires when the session is not hidden, has a path and the cache-only read on line 42450 returns None, and every parse call site reaches the one store that read consults, so "whatever the reader" is right; only the verb misdescribes the target, in the sentence the PR exists to correct. Fix: `(the condition the counting site in _feed_session_key states; 2026-09-21)`, or the pointer alone without the verb.

- docs/reference.md line 2550. The added clause "nor any other reader of the shared store" uses the noun this section gives the goal-store cache 35 lines later: the `goals` bullet on line 2585 sends the pusher's read-only loads "through the shared store cache" under memos.shared. The section's own term for the parse cache is on line 2466, where the `parses` bullet counts "the cold event-model parses through the one parse store the kernel and the judges share". And "reader" is wider than the mechanism: the feed's own cache-only read (kernel/kernel.py line 42450) is a reader of that store that parses nothing and leaves the count standing, so a reader reconciling the counter against the passage could expect that read to clear the count, which it never does, or could look for a goal-store reader that parses. Fix: "nor any other road that parses through the parse store", keeping the deciding clause otherwise as it is.

## Tests

The description's validation figures reproduce at this head. Pin by mutation: an unseeded single-quoted subscript and a setdefault inside the memo drop function each turn the pin red naming the key at this head and leave it green at the base 2b84a6a59, and the pristine test is green at both. The widened subscript pattern finds the same eight subscript literals over kernel.py as the old one (bound, bytes, entries, evict, failed, miss, miss_by, row_by; all double-quoted, and no setdefault in the kernel), the snippet probe returns all three keys where the old pattern returns one, and every one of the eight plus the four helper literals is backticked in the passage. The two modules run together in one process: 79 passed, 66 in the perf-stats module and 13 in the heap-block module, no skips. Since kernel.py is touched, `npm test` in vscode-extension ran with the CI step's command: 5177 tests, 5177 pass, and no test there reads the changed comment text.

The declaration comment was read against the counting site: its condition matches lines 42450 to 42456; the warm-gate clause matches `_warm_wanted` on line 11557, whose docstring calls a session the gate leaves cold "cold by design", and tests/test_boot_parse_gating.py line 96 pins an unmoved idle session as cold; the removed two-parser sentence has no other copy in kernel/, docs/ or tests/. The reference sentence's premise also reads as stated: the end-on-idle pass on lines 28657 to 28684 parses every live session with an armed request on each jobs pass, gated on the request alone, and the deciding clause and the coldFlip sentence are unchanged.

lczh:
Executed against d0d7aae8, the new head. All three points are folded. The helper-call scan now reads either quote style too (a single-quoted call of the counter helper at a site the test never drives turns the pin red naming the key; the four helper literals found today are unchanged), and the docstring says each scan reads either quote style. The declaration comment's pointer says the counting site states the condition. The reference's clause reads "nor any other road that parses through the parse store". The perf-stats module stays at 66 passed and the heap-block module at 13.

ksarma:
The evidence here was executed against d0d7aae8a, the PR's own commit; the head ea5f3ebb8 is that commit's merge with main (pushed at 08:18Z, the same three-file change of 13 insertions and 9 deletions at both), and every file and line reference is on d0d7aae8a, since the merge moved the cited lines. The commit closes the three earlier points. The helper-call scan on tests/test_perf_stats.py line 463 reads either quote style (a single-quoted `_feed_memo_count('probeHelper')` inserted after the memo drop function's with-block, kernel/kernel.py lines 42015 to 42020, a site the test process never drives, turns the pin red naming the key, the same line leaves the test at the base 2b84a6a59 green, and the old and new patterns find the same four literals, coldFlip, coldLive, derived and hit, over the unmutated source), and the docstring on line 454 says each scan reads either quote style. The pointer on kernel/kernel.py lines 41917 to 41918 reads "the condition the counting site in _feed_session_key states", which is what the comment on lines 42452 to 42455 carries, and the four rewritten comment lines replace four, so no line number after the block moves. The clause on docs/reference.md line 2550 reads "nor any other road that parses through the parse store", the section's own term for that store, with the deciding clause and the coldFlip sentence unchanged.

## Smaller points

One point, in the description rather than the code.

- The description's first line reads `Tier: docs`, while the PR's label has been fix since 07:31Z on 2026-09-21, when docs was removed and fix applied. The line cannot move the label from here: `apply_declared_tier` in scripts/ci/tier_policy_check.py (line 189) writes a body's declared tier as a label only when the PR carries no tier label and nobody other than the author has removed one (line 198), and that relabel removed one. The Tier policy check at d0d7aae8a already reports the disagreement in its summary (scripts/ci/tier_policy.py lines 277 to 279): the body declares docs, but the fix label stands. So the effect is on a reader of the description, who takes the tier as docs, and on the check's summary, which carries the disagreement on every head until the line changes. Fix: change the first line to `Tier: fix`, or delete it and let the label stand alone; no code change.

## Tests

The two modules ran together in one process at d0d7aae8a: 79 passed, 66 in the perf-stats module and 13 in the heap-block module, no skips, matching the description's figures. No test under tests/ pins the replaced texts ("says which" at the memo comment, "reader of the shared store"). No file under ui/ or vscode-extension/src names coldLive, `_FEED_MEMO_STATS` or the changed comment's phrases, so no webview pin reads the changed lines; the kernel.py change since 0b8d6d4d3 is four comment lines replaced in place with no JavaScript, and the extension's test run at 0b8d6d4d3 was 5177 passing.

PR #1952, opened 2026-09-21
Pane registry follow-ups: the gear's pane set, the state-root shim, the reload core's revision slot, the revision digest, one listing per build, the id namespace, the phone's tab rule

henrypinkard:
The pane registry's follow-ups from the reviews of PRs 1919, 1920 and 1922 (plans/panes-as-data.md), each a defect with a test that fails before it. Ten items, as facts:

- **A flip of Sessions, the Outline or the Feed dropped every registry key from the store.** The hand rows' handler (`ui/webview/gear.js`) rewrote `romp:settings.panes` from `panesOf`, three keys, so the live reconcile ran with the Artifacts key absent: the rail button hid and the column closed while the row in the open gear still read checked, and a data pane's stored false came back on. `panesOf` now keeps every other boolean member the store holds (the filter is `paneSet`'s: booleans only, so a stray stored value never rides along). A node case executes the function as written; a served leg enables the Artifacts row, hides a data pane by its row, flips the Feed off, and reads the store and the rail.
- **A state-root pane's shim was served with the stale prompt armed**, so after any reconnect the dashboard wore the "may be stale" banner and nothing retired it (no pushed view reaches such a page). `/pane/<id>/shim.js` is served with `no_stale=True`, like the Files, Artifacts and Settings pages; `var NOSTALE=true;` is pinned in the served shim.
- **The reload core's one `seen.code` slot was shared by the build offer and the panes offer**, each overwriting the other's standing offer on every keepalive, and a declined build was re-offered once a panes offer had been declined too. The pane set's revision has its own slot (`seen.pv`); an offer carries `dv`, `code` and `pv`; `offerKey` and `declined` read all three; the panes wording is used only when the revision is the sole new information. The driver runs both scenarios: a build offer stands through a pane-set change with its own words and carries the revision; Not now covers the build and the revision it was made against, and a new revision is offered again.
- **`_panes_rev` digested the files' stats**, so an identical re-define, a bare touch or a malformed file the listing skips offered every open dashboard a reload with nothing to see. The revision is a digest of the checked records; the test re-defines an identical record, touches the file and drops a `bad.json` and reads the same revision, then changes a record and reads a new one.
- **`_landing` listed the registry twice** (its builders, then `_stale_block` through `_reload_core`), and a define landing between the two baked the new revision into a page showing the old set, which was never offered a reload. One snapshot per build: the list and the revision come from it and `_stale_block(v, pv)` bakes that revision; `_shim(..., pv, data)` bakes BOTH its slots and its label from the one snapshot it is handed (the reload core's `PANES0`, its own keepalive gate `LOADEDPV`, the pane's label; the second review found the gate still read the revision itself, so across a define the two disagreed and the gate never fired for the revision that changed, and the label's lookup was a third listing), taking one snapshot of its own only when handed none; the state-root shim route reads one snapshot for the record and the revision. Executed: the listing is hooked to define a second record after its first call, and the landing lists once with the rows and the baked revision the first snapshot's, the route's shim lists once with `PANES0` and `LOADEDPV` equal.
- **A data pane whose id begins `chat-` was classified as a chat column** by the `f-chat-` prefix at the kit's two sites (no grab detector, the chat's strip style). `isChatFrame` (exported, pinned: `f-chat` and `f-chat-<n>` only) is read at both sites; `_pane_check` refuses an id beginning `chat-` and the ids whose derived element names the shell already mints (`tl`, `a` to `d`, `ghost`, `col`), pinned in the registry test and named in `bin/romp`'s help. A pane file already written under such an id is skipped at the next listing with one stderr line naming the file and the rule (the fail-loud convention; `_panes_snapshot`'s bad-file path), so its page is gone from the dashboards until it is re-defined under another id; the plan says so.
- **The phone.** Apply's copy of a generic pane's `data-src` loaded a page by the desktop flag into a frame the phone never shows (the tabless Artifacts record walking a transcript behind the chat tab; every default-on data pane at boot). On a phone a generic pane loads by its TAB alone: the tab bar's tap (`show()`, now only on a phone and only once the pane controller has parsed), or the controller's apply for the CURRENT tab when its key is enabled (the boot, the layout flip); the media query's change event re-applies the controller (`window.__rompPaneApply`) before re-telling the panes, so a rotation to the desktop loads the flagged panes and a rotation to the phone loads the current tab's. `show()` falls to the chat when no bar button carries the id (a stale remembered key of a pane with no tab booted the phone into that page full screen, no tab lit), never a bare return. Pinned three ways: the pane controller's harness with a MOBILE seed (`tests/test_pane_registry.py`), a composed two-script harness running the mobile switcher then the controller in the boot's order over a phone's shell stub with four seeds (`tests/test_kernel_mobile.py`: a remembered tab of a pane with no tab, of a gear-disabled pane, a narrow-layout key on a desktop boot, the current tab loading on the controller's apply and the flip loading the rest), and a served leg at 390 by 844 (the row tap loads nothing and requests nothing; a boot with both flags stored and the tab remembered loads nothing and lands on the chat; a tabbed pane's tap loads its page once).
- **The on-screen gate had no executed unit test.** The controller's harness gains a seed slot; a case seeds `romp:settings {panes:{artifacts:true}}` beside a non-experimental `on:false` row and reads no `src` at boot and `src` after each toggle. The baked-maps pins are their own test, executed in every CI cell (they sat behind a dist gate).
- **The census accepted the guard anywhere in 200 characters**; it now requires the guard as the listener's first statement. Its built-bundle leg, which skipped in every pytest cell, is its own test in `tests/test_pane_registry_served.py` (`TheBuiltBundles`), which CI's browser job runs with the extension built.
- **Smaller.** `load()` in `gear.js` drops `showArtifactsControl` like the other retired keys (pinned; `artifacts.test.ts` re-aimed to `rs-artctl` alone); the registry rows' div follows the Files row, above the Pane docking switch (pinned in the ordering chains of `gear-tabs.test.ts` and `test_kernel_settings_sections.py`); the registry harness no longer hand-lists `artifacts` (the rows carry the record); the stale comments (`kernel.py`'s `seq` prefix, the attribute always present, the section header's rendering claim; `gear.js`; `palette-main.ts`); `bin/romp`'s reserved list names `artifacts` and the derived names; the plans drop `ROW_ORDER` (retired from `pane-dock.ts`, the row order is the DOM read, pinned by an interleaved row) and the `Shown` comment describes every row pane; `pane-grab.test.ts`'s install test records the appended style; `test_settings_tabs_recut_browser.py`'s lists carry `rs-pane-artifacts`; the slicer's `column_css` includes the phone's `.m-on` rules and no longer matches `-panel` (four unrelated panels' rules rode in the panes' slice; the fixture re-cut deliberately, its `made_from` says so); the registry lab's kit leg seeds `panes.artifacts` true instead of relying on an earlier click; the phone tab bar's width at 390 px with the Files control on is recorded under phase four of the plan; `romp pane list`'s unreachable empty-list branch is gone (the door always lists the shipped panes).

Red-first per test at the base 21074750 (the product files at the base, the tests as written): see the first comment for the measured reds. Whole extension suite (served labs under REQUIRE) and whole Python suite on the head before the push.

Tier: fix


henrypinkard:
**Measured reds at the base 21074750** (the product files `kernel/kernel.py`, `ui/webview/gear.js`, `panedock-main.ts`, `pane-dock.ts`, `palette-main.ts` and `bin/romp` checked out at the base; the tests as they are on this branch), by reviewer item:

- **(a) the hand rows' pane set**: `gear-tabs.test.ts`'s `panesOf` case red (the three keys alone: `{timeline, fleet, feed}` where the registry keys should survive); the served lab's gear-flip leg red (`{'artifacts': None, 'lab': None, 'notes': None}` after the Feed flip, where `True, True, False`).
- **(b) the state-root shim's stale prompt**: `var NOSTALE=true;` absent from the served `/pane/<id>/shim.js` (the route's `no_stale=True, pv=...` line absent from the source).
- **(c) the reload core's slots**: the existing revision test red on the offer's shape (`(None, 'panes:abc123') != ('abc123', '')`: the revision rode the code slot); the two-offer scenario red (`KeyError: 'pv'`: an offer carried no revision).
- **(d) the revision digest**: an identical re-define moved the revision (`'f95c63a7b3' != 'fb4c7a0744'`).
- **(e) one listing per build**: `_reload_core() got an unexpected keyword argument 'pv'`; the landing's source without the snapshot.
- **(f) the id namespace**: `_pane_check` accepted `tl`, `ghost`, `col`, `a` and `chat-notes` (`(200, True) != (200, False)`); `isChatFrame` absent (the prefix test at both kit sites).
- **(g) the phone**: the controller's phone case red (`(True, '/notes', None) != (True, None, None)`: the desktop flag loaded the notes pane at boot on a phone); the composed two-script harness's four cases red (a remembered `artifacts` key kept and the tab set to it: `('artifacts', 'artifacts') != ('chat', 'chat')`; a remembered tab of a gear-disabled pane loaded by the restore: `('notes', '/notes') != ('notes', None)`; a desktop boot copying the remembered tab's src: `'/notes' != None`; the current tab loaded at the restore rather than by the controller's apply); the served leg at 390 by 844 red (`'/artifacts' is not None: no src: a pane with no tab is never loaded on a phone`, with one request made by the row's tap); the `show()` and `apply` pins absent from the sources.
- **(h) the on-screen gate**: the seeded case is GREEN at the base too: the gate itself shipped with PR 1922 and holds; what was missing was an executed test that could tell it from its absence (the 07:30Z and 11:16Z reads), which this case is (the earlier harness enabled no row while off screen). The baked-maps pins moved out from behind the dist gate run in every cell now.
- **(i) the census**: green at both: every inline listener already had the guard as its first statement; the census now requires it there instead of anywhere in the next 200 characters. Its built-bundle leg is its own test in the served module (`TheBuiltBundles`), which CI's browser job runs with the extension built.
- **(j) smaller**: `load()` without the Artifacts key's delete (`gear-tabs.test.ts`, `artifacts.test.ts`, `test_files_pane.py` red on the line); the registry rows' div below the docking switch (`gear-tabs.test.ts`'s grouping chain red); the interleaved row opening in the fixed list's order (`pane-dock.test.ts` red); the install test's style (green at both: the fake now records the node and the head appends it; the pin is coverage the review asked for); the fixture re-cut (green at both by design: the slicer's widening reads the same rules from the base kernel's landing; `made_from` says so); the settings tabs lab's lists (green at both: the row exists at the base; the lists now name it).

On this head: the node pins 64 green in six files (`pane-dock`, `panedock-main`, `pane-grab`, `gear-tabs`, `artifacts`, `settings`); the Python pin modules 386 green in thirteen (`test_pane_registry` with the two new controller cases, `test_kernel_mobile` with the composed harness's four, `test_pane_state_broadcast`, `test_artifacts_list`, `test_files_pane`, `test_kernel_pane_rail`, `test_chat_split`, `test_pane_order_parity`, `test_kernel_settings_sections`, `test_pane_gutters`, `test_pane_gutter_drag`, `test_dashboard_auto_reload`, `test_pane_shim_return`); the served labs 21 green (`test_pane_registry_served` as four tests over one drive, the gear-flip leg and the phone leg their own; `test_settings_tabs_recut_browser`; `test_artifacts_pane_served`); the fixture re-cut for the widened slicer (`column_css` 3782 to 3913 characters, the phone's `.m-on` rules; the other five slices unchanged).

Whole suites on this head before the push: the Python suite 14829 passed, 57 skipped, 2487 subtests passed, 2 failed, both source pins on the signatures this PR widened (`_shim(..., pv=None)` in `tests/test_ship_reship_served.py`, `_stale_block(v, pv)` in `tests/test_kernel_update.py`), moved to the new signatures and the two modules re-run green (66 passed); the extension suite, after the same shim anchors moved in `drop-anywhere.test.ts` and `pane-shim-stale.test.ts` (27 pins that slice the shim template out of the kernel source by its `def` line and its format tuple), re-run whole: 5179 tests, 5173 passed, 6 skipped, 0 failed (exit 0), the served labs under REQUIRE.

henrypinkard:
**Round two: the verifier's three mediums and the lows** (the read of 43952d9b), with the reds measured at that head (`kernel/kernel.py`, `ui/webview/gear.js` and `panedock-main.ts` checked out there; the tests as they are on this branch).

**Landed.**

- `_shim` bakes both its revision slots from the one revision it is handed: the reload core's `PANES0` and the keepalive gate's `LOADEDPV` come from `pv` (one read of its own only when handed none). Measured at the base: the route's shim listed three times (`3 != 1`: the route's snapshot, the shim's own revision read for `LOADEDPV`, the label's listing), and `_shim` handed nothing listed three times too.
- The one-listing item has an executed pin: `_panes_snapshot` is hooked to define a second record after its first call, and `_landing()` lists once with the rendered rows and the baked `PANES0` the first snapshot's (one occurrence); the state-root shim route lists once with `PANES0` and `LOADEDPV` equal; `_shim` handed nothing reads once and bakes both from it; handed a revision, its gate bakes it. Measured at the base: the landing leg holds at 43952d9b (round one's one listing), the shim legs red as above; the route's and the def's source lines are missing-name reds beside them.
- The registry rows' place (after the Files row, above the Pane docking switch) is pinned: `rs-panes-data` joins the ordering chains of `gear-tabs.test.ts` and `test_kernel_settings_sections.py`. Measured at the PR's base 21074750 with `gear.js` there (the div sat below the switch): `test_kernel_settings_sections.py`'s chain red (`False is not true`) and `gear-tabs.test.ts`'s grouping chain red; at the round-one head both hold, since round one made the move without a pin.
- The lows: plans/panes-as-data.md's kit list names no `ROW_ORDER`; `_panes_snapshot`'s docstring says what the landing calls now; `gear-tabs.test.ts` extracts `panesOf` to its closing brace by brace-matching, so a base without the two-line shape reds on the behaviour; `isPaneDockingOn`'s doc comment sits on its own function; the body and the plan say a pane already defined under a now-refused id is skipped at the next listing with one stderr line (the fail-loud convention); the slicer's `column_css` no longer matches `-panel` (the fixture re-cut, `made_from` says so); `romp pane list`'s unreachable empty-list branch is gone.

**The first comment's two non-reds, corrected.** The interleaved-row case in `pane-dock.test.ts` PASSES at the base (PR 1920 already reads the row off the DOM); its one base red is the `ROW_ORDER` absence pin, so the row-order read is pinned but the read itself is not this PR's. The gear-tabs grouping chain PASSED at the base too; its red there was the `showArtifactsControl` delete in the same test; the registry rows' move above the docking switch had no pin until this round's chain entry.

**Measured at 43952d9b:**
- `tests/test_pane_registry.py` and `tests/test_kernel_settings_sections.py` with the three product files at 43952d9b: 4 of 31 red: the two shim legs of the one-listing pins (`3 != 1` each) and the two source pins on the widened `_shim` signature and the route's call (missing-name reds); the other 27 hold at both.
- `ui/webview/gear-tabs.test.ts` and `panedock-main.test.ts` at 43952d9b: 12 of 12 green (the round's kit and gear changes are pins and prose).
- With `gear.js` at 21074750: `test_kernel_settings_sections.py` 1 red (the chain), `gear-tabs.test.ts` 2 red (the grouping chain with `rs-panes-data`, and `panesOf`, round one's fix).

**On this head:** the node pins 94 green in eight files (the shim's 27 among them, their anchors following the widened signature); the Python pin modules 388 green in thirteen; the served labs 21 green; the fixture re-cut for the narrowed slicer (`column_css` 3913 to 1747 characters: the four unrelated panels' rules out).

Whole suites on this head before the push: the extension suite 5179 tests, 5173 passed, 6 skipped, 0 failed (exit 0), the served labs under REQUIRE; the Python suite 14833 passed, 57 skipped, 2487 subtests passed, 0 failed, 0 errors (exit 0); the shell suite (`tests/romp.bats`, the pinned bats 1.11.1) 0 failed (exit 0).

PR #1935, opened 2026-09-21
Needs you on every surface: the feed's middle column, the magenta ring and the chip take the user's name and one colour token (plans/needs-you.md, phase two)

henrypinkard:
Phase two of [plans/needs-you.md](https://github.com/romp-on/romp/blob/main/plans/needs-you.md) (the design, merged as #1930): the renames, the Needs you chip on every surface, one colour token for the category, a served lab that reads computed colours and words in both themes, and every old-word pin moved in the same commit. The Needs you box at the bottom of the chat is phase three, its own PR.

## What changes

**Words.** The feed's middle column is **Needs you**, the one category for everything the user can act on (kernel `_CODE_BOARDS` and `board-def.ts`; the category id `needs_input` and the chip class name stay as wire values). The red ring is **Blocked**, the yellow ring is **Needs you** (`RING_ROWS`; the class ids `ring-needs-you` and `ring-waiting-on-you` stay). The status chip reads Needs you for a feed-filed card and for a live prompt alike (`CHIP_LABEL.needsInput` and the legacy `awaiting` name), so the sessions pane's lane chip, the outline pane's rows, the chat chip, the tag overview and the settings demos all say the same word; the timeline's canvas-drawn lane chip follows. The modal's question label reads Needs you / Needs you inside. The guide, the reference and plans/card-boards.md say it.

**Colour.** One token pair, `--st-needs-bg` / `--st-needs-fg` (dark `#d946ef` on `#2a0a2a`, light `#a21caf` on `#ffffff`), declared in styles.css, feed.css's own root and the mobile page's inline sheet, paints the category everywhere: the column chip and header, the card's question marks and label, the group member dot, the Needs you ring and the group pip, the phone picker's current-session border and row bar, the settings' ring demo, the chip classes. Red stays the hard stop's alone: the Blocked ring and fill, the live-block and API-error badges, the API error chip. The parity test keeps the pair readable in both themes and the ring hues apart (oklab deltaE against the other rings and the dots, under red-green deficiencies too).

## Design points

- A hard stop (a permission or approval prompt, an API error only the user can clear, waiting for input) is a red mark ON the card and the red Blocked ring, never a second column.
- Wire values do not move: the category id, the chip class, the ring class ids. Only labels, copy and the colour change, so an older remote kernel's frames still file correctly.
- The chip says Needs you on every surface that shows it (the user's second amendment); the hard stop is told by the ring and the badges, not by the chip's word.
[...]

PR #1953, opened 2026-09-21
Name the two token-less routes, and make the link exemption structural

henrypinkard:
Folds the executed review of #1944 (the comment by ksarma, 2026-09-21). Every item is here.

## The token-exempt list was short by two routes

Both read in `kernel/kernel.py` on this branch's base and covered by executed tests, and both are now named in `docs/reference.md`'s token-file section and mirrored in `SECURITY.md`'s list.

**`POST /push/ack`** is the first branch of `do_POST`, above the `_authorize` call, and its own comment says why: "AUTHENTICATED BY THE PID ALONE, ahead of _authorize on purpose: a worker's fetch carries no token header, and the pid is 128 unguessable bits the kernel itself issued, handed only to the device the push went to, good for two timestamps on that one row and nothing else." The id is `secrets.token_urlsafe(16)`, the body cap is `_PUSH_ACK_MAX_BYTES = 2048` checked against `Content-Length` before a byte is read, and an unknown id is a 404. `tests/test_kernel_webpush.py::test_the_workers_ack_records_the_origin_by_the_rows_endpoint` posts it with `token=False` for a 200 and then with a never-issued id for a 404.

**A token-less `GET /`** is not exempt from the gate but is not refused either: `_authorize` runs, fails, and the `if p == "/"` branch answers `200` with the paste-the-token page, "instead of a dead 403 - the redirect seeds the cookie". `SECURITY.md` already described that page in its token paragraph; the list above it did not.

So the reference now adds a paragraph naming both, and `SECURITY.md` the matching sentences, rather than narrowing the opening to plain GETs.

## The rule stated an exception the pin does not implement

`CLAUDE.md` excepted "a navigation line whose link carries its reason in the same clause (README's Docs section)". README's Docs line passes because its final link sits inside a closing parenthetical, `(source in [docs/](docs/)).`, which `_ENDS_ON_LINK` does not match, not because of that exception, so a session writing the permitted navigation line would have gone red while a parenthesized trailing link passes. That clause is gone; README.md needs none as it stands. What remains is what the pin implements: a paragraph that is only a link, directly under a heading.

The pin's comment now says it is shape-specific, catching neither a link inside a closing parenthetical nor a link followed by anything but a period.

## The link-only exemption is now structural

It fired for a link-only paragraph anywhere on a governed page. It now fires only when the nearest non-blank prose line above the paragraph is a heading, read from `_prose_lines`, so an HTML comment between the heading and the link does not defeat it. Both `CLAUDE.md` and the comment say "under a heading", a position rather than a judgement about the heading's words.

**Red-first, executed on this branch's base before the change:** a link-only paragraph appended to `docs/install.md` after a blank line, under no heading, PASSED (1 passed, 156 subtests). After the change the same paragraph fails with the pin's own message at its file line, "docs/install.md line 70 ends on a link", and the review's second probe, a link-only paragraph after the guide's backends paragraph, fails at "docs/guide.md line 231". With both probes removed every page is green and README's License line under `## License` stays exempt.

## The paragraph helper has a test now

`_paragraphs` and the two helper fixes from #1944 had no committed red. A unit test over a synthetic snippet pins that each paragraph comes back at its FILE line with the joined text, that a paragraph whose second line opens with an inline span tag stays one paragraph, that a lone video tag yields none, and that a comment between a heading and the line below it does not hide the heading. The two-tuple return of `_paragraphs` is unchanged, since the word-count helper reads it; the heading lookup is a separate helper.

Two mutations of the helper, both executed, turn the new test red and left every page test green before it existed: returning filtered indexes instead of file lines, and dropping the block-HTML distinction so every line opening with `<` is skipped.

## The guide promised an address

`docs/guide.md` said SECURITY.md "gives the address for reporting a vulnerability"; SECURITY.md's Reporting a vulnerability section names GitHub Security Advisories on the repository and no address. It now says it "names the private channel for reporting a vulnerability".

## Tests

- `tests/test_docs_front_pages.py`: 10 passed, 298 subtests (7 page tests, 3 new helper tests).
- With `tests/test_api_health_hover.py`, the other Python module reading these pages: 46 passed, 298 subtests.
- `tests/test_kernel_webpush.py`, `tests/test_kernel_installable_app.py`, `tests/test_token_login_page.py`, which cover the two routes this change documents: 132 passed.
- `ui/webview/dense-chrome.test.ts` (reads the guide and the reference) and `ui/webview/md-sanitize.test.ts` (reads SECURITY.md): 23 passed, 0 failed.

## Sibling

The open Needs you pull request #1935 also touches `tests/test_docs_front_pages.py`, with a chip-class pin. The hunks here are confined to the helper comments, the new `_prose_line_above` helper, the one skip line inside `test_no_paragraph_ends_on_a_link`, and a new test class at the end of the file, so the two changes do not overlap textually.

Tier: docs


ksarma:
The evidence here was executed against 5e513a683, the PR's head, merged as be18a9231. File and line references are on that commit; the merge changed docs/reference.md only at lines 2087 to 2092, six lines for six, so the lines cited there read the same on the head and after the merge. The change names the two token-less routes in docs/reference.md and SECURITY.md, makes the link-only exemption in tests/test_docs_front_pages.py structural, and adds a test class over the paragraph helper. Two points stand after the merge, then two smaller ones.

**What a valid push id buys.** docs/reference.md:5310-5311 says the id stamps two timestamps on one row and "buys nothing else", and SECURITY.md:61-62 mirrors it as good for two timestamps. Both restate the kernel comment at kernel/kernel.py:67608-67609, and the route under that comment (67605-67641) does three more things with a valid id:

- The stamp helper (57746-57762) gives the stage's timestamp its first stamp only, so a repeated report is idempotent there, but rewrites `swVersion` on every report (the route always passes a string, 67629).
- On a shown report only (67634-67636), the supersede helper (57774-57799) stamps `supersededAt` on every older row for the same endpoint and session that is unsettled and not yet tapped, so the page's read of unsettled pushes stops naming them.
- `_push_backfill_origin` (67640; 57527-57557) records the request's origin on the device's subscription when none is recorded. The origin is read from the Origin header, else the Referer's origin, else the proxy's forwarded headers or the Host (57490); a recorded origin is kept, and a different one logs a conflict line.

The consequence for a SECURITY.md reader: the recorded origin becomes the navigate URL of that device's next declarative Apple push (`_push_wire`, 58043-58054), so a chosen Origin header on a token-less report against a subscription with no origin on file aims the next tap at that origin. Executed over the module's own route harness in tests/test_kernel_webpush.py with the token withheld: one `POST /push/ack` with stage shown and an Origin header stamped the shown time and rewrote the build string on the target row, set the superseded time on an older untapped row for the same session and endpoint, and wrote the subscription's origin; a second report rewrote the build string again. The module's own tests assert each write and pass at this head: the by-pid ack test asserts the build string, the shown-ack test asserts the older row's superseded time, and `test_the_workers_ack_records_the_origin_by_the_rows_endpoint`, the one the description cites, asserts the subscription's origin and builds the next wire's navigate from it (tests/test_kernel_webpush.py:1686, 1742, 1062-1067).

The fix is wording, in both pages together: the id stamps the row's shown or tapped time (the first stamp stands) and records the worker's build; a shown report also marks the older unsettled, untapped pushes for the same session on that device superseded; the request's origin is recorded on the device's subscription when none is recorded; an unknown id is a 404. SECURITY.md can add the navigate consequence in one clause. The same phrase has two homes outside this change's files, the kernel comment at 67608-67609 (whose own sentence at 67611-67613 names the supersede) and the class docstring at tests/test_kernel_webpush.py:1637-1640; both belong to a separate change.

**The structural clause has no committed test.** The change's behavioural edit is the clause at tests/test_docs_front_pages.py:204, `and above.get(line, "").startswith("#")`. No committed test composes the heading lookup with the link-only match: `test_the_nearest_prose_line_above_sees_through_a_comment` (260-267) asserts on the lookup alone. Executed in a scratch copy of the head: restoring the pre-change skip, `if _LINK_ONLY.match(joined):`, at line 204 leaves every committed test green (10 passed, 298 subtests). Inverting the clause, or deleting the skip, goes red only at README.md line 76, the License line, which is the only link-only paragraph on the four governed pages (enumerated by running the helpers over the governed list), so the pages pin the exempt case alone. The description's red-first, a link-only paragraph appended after the last line of docs/install.md, reproduces at the head and passes again under the pre-change skip (10 passed, 299 subtests): the red lives only in an uncommitted page edit. An edit that drops the clause, or a merge resolution that keeps an older copy of the skip, lands with the module green, and a link-only paragraph under prose passes the pin again, the shape the change closes.

The fix: lift the predicate into a helper the page test calls, for example `_trailing_link_violations(text)` returning the offending (line, joined) pairs, built from the paragraph walk, the heading lookup and the two link patterns. Pin it in ParagraphHelper over three synthetic snippets: a link-only paragraph directly under a heading yields no violation; the same paragraph after prose with a blank line between yields one at its file line; a prose paragraph ending on a link yields one. Prototyped on the head: 11 passed; dropping the clause is red at the new test with the page tests still green, so the test is what catches it; inverting is red at README line 76 and at the new test. Two bounds on that fix. Keep the page test's per-paragraph `subTest` loop, or have the helper return a result per paragraph, if the description's 298-subtest figure is to stay accurate: a page test that opens a subtest per violation only reads 142. And the edit widens the hunk inside `test_no_paragraph_ends_on_a_link` beyond the one skip line the description's Sibling section names, so that section and the red-first paragraph, which can then point at the committed test, move with it.

## Smaller points

- The Sibling section says this change and #1935 do not overlap textually. Read on both diffs: this change appends ParagraphHelper at line 220, after the last method of FrontPagesStayShort and before the main guard, and #1935 appends `DocsChipClassesHaveRules` at the same anchor, so #1935's rebase over be18a9231 meets a conflict there; the resolution is to keep both classes. Placing ParagraphHelper directly under the helpers it tests, above FrontPagesStayShort, would move it off that anchor.

## Tests

- tests/test_docs_front_pages.py:255: `assertNotIn(7, starts)`, with the message "a line opening with an inline tag continues its paragraph", is satisfied by the regression it names. Read in `_paragraphs` (97-122): a line 7 treated as block HTML flushes the paragraph at line 6 and is skipped, so no paragraph starts at 7 either way, and the count stays three. The regression is caught today only by the sibling test's full equality at 244-251; if that equality is later loosened, the regression lands green while the test named for it passes and its message asserts the opposite of what happened. Asserting the mechanism, for example `self.assertIn('<span class="romp-chip">Chip</span> line opens', dict(_paragraphs(self.SNIPPET))[6])`, makes dropping line 7 change the outcome.

PR #1954, opened 2026-09-21
An empty chat build over a baseline the detector popped sends no base holder an empty frame, since a marked session's clients hold content

lczh:
Tier: fix

The chat wire's shared delta baseline is seeded by a session's first whole frame to reach a client (#1909). When two whole-frame senders race on a session with no baseline, the seed's detector pops the baseline and marks the session, and the next cycle whose loop reads the baseline absent sends every base holder the whole session and takes the mark off with its write. Three things were left around that mark. First, the guard that keeps a transcript read that came back empty from going out as an empty session frame read only the baseline: after a pop, or after a cycle that sent tails and left the mark standing, the session has no baseline while every client holds content, so an empty read in that window sent every base holder an empty session frame, counted it under `empty` with a client-diag row each, said nothing on stderr, and on the cycle road wrote the empty list as the baseline and took the mark off, so content's return was one more whole frame to everyone. The pane did not blank, since the page absorbs such a frame as status-shaped; the cost was meter noise, the lost stderr line and one redundant whole frame. Second, the eviction that forgets a tab the strip stopped listing walks the detector's marks, but nothing pinned that walk or its clear: with both removed, the skeleton-reconnect and delta-resync modules stayed green. Third, four sentences (three in the seed's docstring, one in the targeted push's, and the reference's gloss on first-event fulls) still said the next cycle's write always repairs every client and clears the mark, while the code clears it only from a cycle whose loop read the baseline absent and leaves it standing after a cycle that sent tails.

The fix: the mark now carries a count, the length of the list the detector popped, raised by every whole frame handed under the mark (the seed's declining arm raises it, since it runs after the sends for a list some client took whole; a count left at the pop named the shorter list once a targeted push under the standing mark had handed every base holder a longer one). The empty-build guard on both send roads (the pusher's cycle and the targeted one-session push) treats a marked session as one whose clients hold content. An empty build over a popped baseline sends no empty frame to any base holder, counts no `empty`, files no row, says the failed read once on stderr with that count, and leaves the mark standing and the baseline absent for the next content cycle's repair; with a cached build to stand in, the stand-in reaches every base holder as a whole frame and that cycle's write takes the mark off, the road the seeded case already took. The eviction's walk unions the marks' keys and pops the mark. Two `builds.chat` counters on `/perf`, `baselineRaced` (at the pop) and `baselineRepaired` (where a cycle's write takes a standing mark off), decided under the baseline lock and counted after it, make the rate the review asked about countable; the reference glosses them beside `moved` and `coldSkipped` (raced minus repaired is the marks still standing plus the tabs that left the strip, whose eviction clears the mark with no repair). The four stale sentences now state the rule the code has, one sentence each. The eviction's premise comment and the reference gloss state, not fix, the one exception: the page keeps a strip-omitted tab the frame's `live` field lists, so for a live session the strip omits the eviction forgets bases the page still holds, and the re-listing is a row-less whole frame per client, since the dedup slot is popped with the base.

This change follows the seed lower-bound fix (#1942) and the seed-on-delivery fix (#1943) from the same review, both merged; it sits on the merge that carried the second.

Validation
- Red first against the stacked base's kernel file, swapped in and restored byte for byte with a hash check: tests 35 (the cycle), 36 (the targeted push) and 37 (the cached-hit road) fail at their first assertion, an empty session frame for the raced session reaching both base holders (37 by the other road: the last filled card each client held stayed the older one); test 34 passes there, as a pin should (3 failed, 1 passed, 57 deselected).
- Test 34 under the review's eviction mutation on the fixed kernel (the marks dropped from the eviction's walk and their clear removed): red at its first assertion after the tab leaves the strip, the session still marked (1 failed; tests 21 and 25 beside it green, so the walk's third member is what it pins).
- Test 36's count pin against the change without the raise: red at the stderr assertion, the note naming the 5 events at the pop while every client held the 7 a targeted push under the standing mark had handed them.
- Green after, one module at a time: skeleton-reconnect 61 (57 plus the four new tests), chat-delta-resync 12 (its regex pin on the cycle's write gate is byte-identical), chat-fixed-cost-memos 44, cold-tab-gate 22 (4 subtests), kernel-pusher-snapshot 31 (2 subtests), chat-empty-build-guard 5, perf-stats 66 (its exact-dict pin on `builds.chat` gains the two counters), kernel 460 (12 subtests), chat-pages 43 (27 subtests), perf-bench 49.
- The static scan of test pins on kernel source lines lost nothing. The modules the second scan named as pinning the changed functions by source were run too and are green: kernel-tabs-first 7, pusher-change-gates 13, single-flight-builds 11, stat-rule-readers 15 (4 subtests), timeline-skeleton-dedup 4, wire-once-per-build 24, feed-first-cold-push 7, files-pane 27, first-cycle-stage-split 19, settings-page 10, comment-threads 115.


ksarma:
The evidence below was executed against aabc25305, the PR's head, merged as 611bd01b1; file and line references are on that commit. Of the PR's four files only docs/reference.md differs between the head and the merge tree, so the kernel/kernel.py and test lines read the same at both.

**The cached stand-in under a standing mark rewinds every base holder** (kernel/kernel.py 54909-54926). The guard at 54909 reads the mark and, with a cached build, hands it to every base holder as a whole frame at 54926; the write block at 54974-54977 makes it the baseline and takes the mark off. The premise in the comment at 54910-54911 (same events, so the diff sends nothing) never applies under a mark, whose baseline is absent: the stand-in always leaves as a full.

- Shorter face: a targeted push under the mark hands every base holder a longer list and caches nothing (55563). After test 23's race a targeted push leaves both clients at 7, and the next empty read sends each the 5-event stand-in as the new baseline; the page applies a behind full as sent (ui/webview/render.ts 17578-17609), so the two newest cards leave every pane until the next content cycle's tail, where the base kernel's empty frame moved no card.

- Same-length face: in test 24's order the cache keeps the older list with u3 unfilled while client a shows the filled card; the empty cycle sends a that list as the new baseline, and a shows the older card until the next tail. A compare of the stand-in's length against the mark's count closes the shorter face alone: the count sees length, not content, and a cycle that sent tails while a mid-loop seed marked the sid (test 27's shape) leaves clients ahead of the count, so a shorter stand-in still goes out when that cycle's build was not cached.

- Fix, executed: bind the mark once and take the no-hit road under it, `if hit is None or marked: continue`, keeping the release of the claimed build (`_chat_inflight_done`) inside the widened branch. The empty cycle then sends nothing for the sid, the mark and the absent baseline stand, and the next content cycle's full repairs every base holder, deduping for a client already showing that list; the targeted push's guard at 55571 needs no change. Tests 22, 34, 35 and 36 stay green; test 37, which pins the stand-in road, goes red and re-pins to no S1 frame, the mark standing, the baseline absent, counters (1, 0), then (1, 1) after the content cycle. The cost: in test 23's order the repair of the race's older holder lands one cycle later, and no holder is ever shown a card older than the one it shows.

- Wording: the comment at 54913-54920 and test 37's docstring say the stand-in goes out as a change-0 full repairing from an older list, the seeded case's road, and the `_empty_build_regresses` docstring at 33912 puts the marked sid on that road too; the row reads changeAt0 only when a client holds exactly the stand-in's span (lastGone:record in the shorter face), and "older" there means older than the grown transcript, not than a card a client shows. The seeded road itself, a targeted push's tail leaving both clients ahead of the seeded baseline, sends the stand-in as a lastGone:record full at both heads: pre-existing, outside this diff, and not closed by a count compare either, since `_chat_prior_n` reads the baseline's length there; one sentence naming it belongs at 54910-54912.

### Smaller points

- **What the count under the mark means** (kernel/kernel.py 33917, 34046-34047, 34050, 34059). The docstring of `_chat_prior_n` and the comment at 34047 read the count as what every base holder has, at least, while the code stores a max over lists some client was handed whole: a connect push under the mark hands its full to one client and the raise arm at 34050 lifts the count to it (stderr said 7 while two of three base holders had 5), and the pop arm's `max(len(cur), len(evs))` at 34059 can name a count no holder has. After test 27's tails cycle the note says 5 while both clients have 6; an else arm under the lock at 54974 raising the count to that cycle's list length would make it exact there. Only the stderr note and the perf row read the count, so the fix is wording at 33917 and 34046-34047 (the longest list some client was handed whole under the mark), with the note calling the count a minimum unless the else arm is taken; raising only on the targeted road gives no every-holder reading, since the pop arm already stores the max.

- **Four more sentences with the retired phrasing** (kernel/kernel.py 54807 and 54810, 54988, 55540-55541, 51475). The body counts four reworded sentences saying the next cycle's write always repairs every client and clears the mark; the diff rewords five, and these four, from PRs 1909 and 1942, stay; the code's rule (tests 27 and 35): a cycle that sent tails leaves the mark, and the next cycle whose loop reads the baseline absent repairs and clears; comment-only, a consistency gap against the body's claim, not a code defect.

- **The snapshot-shape docstring** (kernel/kernel.py 414-421) names `active_built`, `bg_built`, `moved` and `bg_miss` and neither `baselineRaced` nor `baselineRepaired` (nor `coldSkipped`, pre-existing); that `romp perf` prints neither is the printer's documented contract (docs/reference.md 2526-2527), so one docstring line covers it and a printer change would be its own change, covering `coldSkipped` too.

- **The connect push as a third road under the mark** (kernel/kernel.py 54909). The guard also runs on the connect push, which the body does not name: with nothing cached, a connect target for a marked sid whose read came back empty gets no S1 frame before the next content cycle, where the base kernel sent it an empty noBase full at once (as a seeded sid already gets at both heads); with a cached hit, the head hands the target the stand-in where the base sent the empty frame, an improvement. Fix: state the cache precondition in the body and the comment at 54913-54920 (the stand-in goes to the non-holder target here, not to every base holder) and pin both arms; under the no-stand-in road above the cached arm is the no-frame shape too, so the pin follows the road taken, and sending the target an empty frame at once is a separate decision that also changes the seeded road.

### Tests

The two counters are pinned on the happy path only (kernel/kernel.py 54971 and 34050; tests 35 and 37 alone read them, beside a perf-stats zero pin): by code reading, not execution, a mutant counting a repair on the tails cycle or a pop in the declining arm diverges only where no test reads the counters, and the pins are test 27 replayed through the tails cycle after a collector reset, asserting (1, 0) before the second cycle then (1, 1), and test 25 replayed after a reset, asserting (1, 0) after the connect push under the mark, both owing a green run at the head and a red run under each mutant. The pop's count at 34059 is pinned by no test, since every race fixture pops two lists of equal length: mutants storing `len(evs)` alone, or the seeded list's length alone, kept the chat modules' tests green, and the cheapest complete pin is direct `_seed_chat_baseline` calls (test 26's shape) asserting the count reads 6 for seeded 5 against this sender's 6 and for seeded 6 against 5, where a `_race` fixture with an appended event kills only the first mutant; the wording alternative is the longer of the two racing lists. Owed beside tests 35 to 37: the mirror of 37 with a targeted push under the mark before the empty build, asserting the S1 frame list, the baseline and the mark (flipping the status as test 36 does, or the repair full dedups); test 24's order on the cached-hit road; test 27 through the tails cycle, then an empty build, asserting no empty frame, the standing mark, the absent baseline and the note's count (5 as written, 6 if raised on the tails cycle); and the connect road under the mark with an empty read, both arms, the cached arm red at the base.

PR #1955, opened 2026-09-21
The background warmer skips a leaf it cannot stat, so a warm another cold session kicked no longer parses a gone transcript once per kick

lczh:
Tier: fix

The background parse warmer's own loop had no gate on a transcript it cannot stat. The merged change before this one kept the feed body from asking for a background warm when a session's transcript is gone from disk, but the warmer's loop still filtered its rows on "moved since boot, or working now" alone, which a gone leaf satisfies through its goal store's or states log's mtime. So once any other cold session kicked the warmer, the warm thread parsed the missing path once per kick: the shared store cannot compute a live key for it because the leaf's stat fails, so any slot it still holds from before the removal never matches, the parse stores nothing under a None key, one kernel parse is counted, nothing on stderr. The earlier claim that nothing parses the gone leaf anywhere in the kernel held only while it was the sole cold candidate.

The fix gates the loop at its subject: a warm-wanted row whose transcript cannot be stat'ed is skipped before the parse and before the "parsed anything" flag, so a warm that skipped everything still drops nothing and wakes nobody. The predicate is the one the feed key already reads for its transcript identity, paid once per warm-wanted session per kick on the warm thread and never on the feed path, so the inputs census sees no new read. It is not placed inside the warm-wanted helper, whose feed caller already sits behind the leaf bit. The gate also skips the transcript-less rows the live-session list adds (the live stub, a just-created SDK row), whose parse stored the same nothing while the row was working; discover hands the file over once it exists. The warmer's docstring now says it skips such a leaf itself.

Wording, no behavior change:
- The warm-to-stale block's tab-list clause described one road. It now says that only the names stub for a live sid no backend owns notes on stderr that the transcript could not be resolved, once per episode, and that the SDK-owned road never notes.
- The chmod 0 disclosure in the same block names the surface: a leaf that stats but cannot be opened is re-read once into an empty parse the store keeps under the transcript's mtime and size, so the card paints quiet from it on later builds until the mtime or size moves (a chmod moves ctime alone). The close is a read-failure signal out of the parser, its own change.
- The feed key's docstring list of the facts handed to the body through the context now includes the leaf bit.
- The new gate comment and the body's cold-road comment say the store cannot compute a live key for the gone leaf, not that it holds no slot: a session parsed before the removal still has its old slot, which the None key neither matches nor evicts.
- The source pin on the re-read's gate line in the cache-only test keeps its literal as a locator; its message now says so and names the executed tests that prove the behavior: the append and emptied-store tests for the warm-key clause, the gone-transcript test for the stat clause, its warmer sibling for the body's ask, and the new multi-cold test for the loop gate.

Validation:
- New test in the feed session memo module, beside the sole-candidate warmer test: a second session is left unparsed before the removal so its cold road kicks the real warmer under a feed-only client; a recorder on the parse entry point never sees the gone path, and the kernel parse count moves by the stat-able sessions' misses alone. Red before the gate, with the kernel at the merge's text: 1 failed, 56 passed, the gone path found in the recorder's list. Green after: 57 passed.
- New ordering pin in the boot parse gating module, in the shape of its "a warm that parses nothing wakes nobody" test: a working live row whose transcript is not on disk under a feed-only client, a sentinel in the pusher's cached feed, the parse entry point and the pusher poke recorded. Red at the merge's text: the missing path was handed to the parse (1 failed). Red on the reorder mutant, the gate moved below the parsed-anything flag: no parse, but the cached feed was dropped to None (1 failed). Green at the head. The kernel was restored byte-exact after each swap.
- Modules run one at a time, all green: boot parse gating 29 (its warmer tests write real transcripts and survive the gate), feed session memo 57, kernel feed warm 4, kernel 460, feed memo inputs 31 (the census sees no new read), perf stats 66, perf bench 49, pusher change gates 13, chat empty build guard 5, feed live clobber 3, goal store fault boundary 30, interrupt marks memo 19, kernel goal cache wiring 17, kernel interrupt 54, kernel interrupt machine cut 51, kernel judging active 5.
- The static pin scans report nothing lost on the branch.

Follows the post-merge review of #1941.


ksarma:
Evidence for this review was executed against 85a79a218, the PR's head, merged at 09:11Z as 3fc615cc7; every file and line reference is on the head commit, since the merge moved kernel/kernel.py's lines (the three test files read the same at both). The change adds a two-line gate to the background warmer's loop (kernel/kernel.py 35210-35211): a warm-wanted row whose transcript cannot be stat'ed is skipped ahead of the parse and ahead of the parsed-anything flag. A warm that another cold session kicked therefore no longer parses a gone leaf once per kick, and a warm that skipped everything leaves the feed cache and the pusher alone.

That gate is the only executable change; every other added line is comment, docstring or test text. It reads the same identity the feed key takes for its transcript component, and every row the loop iterates carries a string path, so the two predicates agree on every reachable row. The two new tests are red at the base for the reason the PR description gives and green at the head. With the gate deleted both go red; with the gate moved under the parsed-anything flag the ordering pin goes red on the dropped cache while the multi-cold test stays green, since api's parse sets the flag. Two items remain: a wording inconsistency the change leaves standing beside its own corrections, and an unasserted premise in the new multi-cold test.

### Smaller points

- Three "no slot" phrasings at this head state the inventory claim the PR's own corrections reject, and all three are false at the moment they describe. The PR's new gate comment in the warmer (kernel/kernel.py 35199-35201) and its rewritten cold-road comment in the entry (42680-42681, where the old text said the store "has no slot for it") say the store cannot compute a live key for the gone leaf, and its description says a session parsed before the removal keeps its old slot. The opposite still stands in the re-read block's opening sentence, kernel/kernel.py 42522-42523, just above the lines the PR rewrites in that block: a transcript gone from disk under a warm entry "has no slot in the shared store". It also stands in two assertion messages in the test class whose docstring the PR extends: tests/test_feed_session_memo.py 1529 ("has no store slot"), in a test whose own comment at 1545-1549 says the pre-removal slot is never retired while the leaf is gone, and 1553 ("the rewritten leaf has no slot in the store"). A scratch probe replaying lines 1518-1529 of that test found the store still keeps the (fsid, cut, leaf) slot for the web session after the unlink, while `km._parse_cached(leaf)` returns None; the slot is unchanged after a None-key parse of the gone leaf. All three phrasings predate the PR and are untouched by its diff. The inconsistency is across files (42523 and 1529 against 35200, 42681 and the description), not inside one comment: within the re-read block the only slot-persistence text is the chmod 0 case at 42537-42539, a stat-able leaf, which does not contradict the gone-leaf sentence. Three other "no slot" phrasings near the diff are accurate and can stay as written: kernel/kernel.py 35139 and 42346 and tests/test_kernel_feed_warm.py 10 say there is no slot "to re-read into", which is the write destination. That statement is true because `parsed_session` stores only under a non-None key, so a re-read of the gone leaf stores nothing. A fix for the three wrong ones: at 42522-42523, say the store cannot compute a live key for it, so no slot it keeps matches and the re-read has nowhere to land; at 1529, for example "no live key for a transcript that cannot be stat'ed: nothing to re-read into"; at 1553, "the rewritten leaf's live key matches no slot". The two test edits change assertion strings only, so all three fit in one wording commit with no behavior change.

### Tests

- The new multi-cold test does not assert the premise it depends on, so a later change to the warm-wanted predicate leaves it green without the loop ever reaching the gate. The test, `test_a_warm_that_another_cold_session_kicked_skips_the_gone_leaf` (tests/test_feed_session_memo.py 1614), pins the gate by execution at the head: with the gate deleted, line 1660 finds the web session's gone path in the parse recorder's list. Its premise is that the web row stays warm-wanted after the unlink at 1651 through its goal store's mtime since boot. The fixture's rows are idle and its stores are written in setUp, so after the unlink that status rests on the store leg of `_session_moved_since_boot` alone (kernel/kernel.py 11730). A later change to that predicate, the store leg dropped or the boot timestamp re-based per warm, makes the row cold: the loop's warm-wanted continue fires before the gate, the recorder never sees web, the kernel parse count still moves by api's one, and the test passes while pinning nothing about the gate in the multi-cold shape. Executed on a scratch copy of the head, single-process pytest on that one test: as written, 1 passed; with the store leg dropped from the predicate and the gate present, 1 passed; with the same mutant and the gate removed as well, 1 passed, so under that mutant the test cannot see the gate at all. The predicate's own test does not catch that mutant either: `test_moved_since_boot_reads_either_file` (tests/test_boot_parse_gating.py 150) pins the transcript and states legs only, never the store leg. The fix is one line after the unlink at 1651, in the shape the new boot test already uses at tests/test_boot_parse_gating.py 110. Assert that the warm-wanted helper is true for the row `{"sid": WEB, "path": str(self.tpath[WEB])}` with the web live row, with a message such as "the gone leaf is still warm-wanted through its goal store's mtime since boot: the loop reaches the gate". That two-key row is enough because `_session_files_stat` reads only sid and path (kernel/kernel.py 11670). Verified: with the assertion added, the head passes; under the store-leg mutant the test fails at the new line; with the gate removed and the assertion present, the test still fails on the existing assertNotIn at 1660 with web's path in the recorder's list, so the added line does not disturb the test's arming. Optional, and not this PR's addition: the merged sibling `test_a_warm_entry_whose_transcript_is_gone_asks_the_warmer_for_nothing` (1570) rests on the same unasserted premise after its unlink at 1600, and the same one-line assertion there would keep it from going vacuous under the same change.

PR #1957, opened 2026-09-21
Say what a valid push id reaches, and pin the link exemption in a test

henrypinkard:
Folds the executed review of #1953 (the comment by ksarma, 2026-09-21). All four items are here.

## What a valid push id reaches, in both pages together

Both pages said the id was good for two timestamps and bought nothing else. Read on this branch's base, a report on a valid id reaches that device's push state in four ways, and the two pages now say so.

- **The stage's time takes its first stamp only, but the build string is rewritten every time.** `_push_ledger_stamp` (kernel/kernel.py) sets the stage field under `if not r.get(field)`, so a repeated report is idempotent there, then does `if sw_version is not None: r["swVersion"] = str(sw_version)`. The route always passes a string (the sanitized `v` from the body), so every report rewrites the build.
- **A shown report supersedes older rows.** On `_stage == "shown"` the route calls `_push_ledger_supersede`, which stamps `supersededAt` on every earlier row with the same endpoint and session that is unsettled and not tapped, so the page's read of unsettled pushes stops naming them.
- **The request's origin is recorded when none is on file.** The route ends in `_push_backfill_origin(_row.get("endpoint"), _request_page_origin(self.headers))`. The origin comes from the `Origin` header, else the `Referer`'s origin, else `X-Forwarded-Proto` with `X-Forwarded-Host`, else `Host`; a recorded origin is kept and a different one logs `[push] origin conflict`.
- **That recorded origin aims the next Apple push.** `_push_wire` builds the declarative message from `sub["origin"]` for an Apple endpoint that has one, so the recorded origin becomes the `navigate` URL. SECURITY.md carries that consequence in a clause, since it is what a security reader needs from the fact.

An unknown id is still a 404 and the body is still capped at 2 KB before a byte is read; both pages keep that. `tests/test_kernel_webpush.py` asserts each write (the by-pid ack test on the build string, the shown-ack test on the older row's superseded time, and `test_the_workers_ack_records_the_origin_by_the_rows_endpoint` on the subscription's origin and the next wire's navigate), and it passes here.

The same phrase in `kernel/kernel.py`'s own comment and in the webpush test's class docstring is left alone: a separate change owns those.

## The link exemption is pinned by a test now

The behavioural clause from #1953 had no committed test: restoring the older skip left every test green, because README's License line is the only link-only paragraph on the four governed pages, so the pages pinned the exempt case alone.

The rule now lives in `_trailing_link_checks(text)`, which returns `(file line, joined text, offends)` for every paragraph, with `_trailing_link_violations(text)` over it for the offenders. `test_no_paragraph_ends_on_a_link` reads the verdict, so it still opens one subtest per paragraph, and three synthetic snippets in `ParagraphHelper` hold the composition: a link-only paragraph directly under a heading yields no violation, the same paragraph after prose with a blank line between yields one at its file line, and a prose paragraph ending on a link yields one.

**Red-first, both executed on this branch.** Dropping the heading condition from the helper fails the new test with the page tests still green (1 failed, 10 passed, 299 subtests). Inverting it fails the new test and the page test at `README.md line 76` with the pin's own message. The subtest count reads 299 rather than 298 because the exempt paragraph now gets a verdict and a subtest of its own instead of being skipped before one was opened.

## The helper class moved off the end-of-file anchor

`ParagraphHelper` now sits directly under the helpers it tests, above `FrontPagesStayShort`, so it no longer occupies the anchor the open Needs you pull request #1935 also appends a class to. That resolves the collision the review found between the two diffs; the remaining hunks here are the helper block and the body of `test_no_paragraph_ends_on_a_link`.

## The inline-tag assertion reads the mechanism

`assertNotIn(7, starts)` was satisfied by the regression it named: a line 7 read as block HTML flushes the paragraph at 6 and starts none at 7 either way. It now asserts that line 7's words are inside the paragraph at line 6, which the regression changes.

## Tests

- `tests/test_docs_front_pages.py`: 11 passed, 299 subtests.
- With `tests/test_api_health_hover.py` and `tests/test_kernel_webpush.py`, the module whose assertions back the route facts above: 162 passed, 299 subtests.
- `ui/webview/dense-chrome.test.ts` (reads the guide and the reference) and `ui/webview/md-sanitize.test.ts` (reads SECURITY.md): 23 passed, 0 failed.
- Greps of `tests/` and `ui/webview/*.test.ts` for a distinctive phrase of each rewritten sentence find no pin on the docs prose; the only other home of the old phrasing is the kernel comment and the webpush docstring, both out of scope here.

Tier: docs


ksarma:
The evidence here was executed against 092b625a3, the PR's head, merged as 331c11210. File and line references are on 092b625a3. The merge changed kernel/kernel.py only between lines 35217 and 42735 and docs/reference.md only around line 2565, both earlier in their files than any line cited here, so on the merge commit every kernel line number cited here is 28 larger and every reference.md line number one larger; SECURITY.md, README.md and both test modules cited are byte for byte the same on the head and the merge. The change rewrites the push-report paragraph in SECURITY.md (lines 58 to 72) and docs/reference.md (5321 to 5338) to say what a report on a valid id reaches, lifts the trailing-link rule into a helper that returns a result per paragraph, pins that helper over three synthetic snippets, moves `ParagraphHelper` above the page tests, and makes the inline-tag assertion read the paragraph's text. Each claim in the description was checked at this head; what agreed:

- Kernel facts, read and matched against both pages word by word. The stamp helper (kernel/kernel.py:57849-57865) writes the stage's time only when the field is unset and writes the build string whenever one is passed, and the route always passes the sanitized string (67732), so every report rewrites it. The supersede helper (57877-57902) runs on a shown report only (67737-67739) and stamps the earlier rows with the same endpoint and session that are untapped and unsettled (57897, with unsettled as 57905-57907 defines it). The origin is read from `Origin`, else the Referer's scheme and host, else the forwarded headers or Host (57593-57627); it is recorded on the row's endpoint's subscription only when none is on file, a recorded one stands, and a different one logs a conflict line (57630-57660, called at 67743). An unknown id is a 404 (67733-67735) and the 2 KB cap runs before the body is read (67717-67720, the constant at 57808). Both pages say those things in those terms; the one wording point is under Smaller points.
- tests/test_docs_front_pages.py at this head, single process: 11 passed, 299 subtests. The module at the base, run over the same pages: 10 passed, 298 subtests, so the extra subtest is the result for README.md line 76 opening one, as the description says. With tests/test_api_health_hover.py and tests/test_kernel_webpush.py: 162 passed, 299 subtests, matching the description.
- Mutants in a scratch copy: dropping the heading condition fails only the new test, at NAV_UNDER_PROSE, with the page tests green (1 failed, 10 passed, 299 subtests, the description's figure); inverting it fails the new test at NAV_UNDER_HEADING and the page test at README.md line 76 with the pin's own message (2 failed, 10 passed, 298 subtests); deleting the exemption fails the same two; inverting the filter in `_trailing_link_violations` fails the new test.
- The inline-tag assertion now asserts the mechanism: reintroducing the regression, every line opening with `<` skipped in the paragraph walk, fails the new assertIn at line 198 with its own message, alongside the equality test at 184; before this change only the equality test caught it.
- #1935's one hunk in this module, which appends `DocsChipClassesHaveRules` after the rule test, applies to this head cleanly in a dry run (offset 154, no fuzz); against the base module it needs fuzz 1, the collision the earlier comment named.
- The two webview tests the description names were not run here; by inspection, the regions they read in docs/reference.md and SECURITY.md lie outside these hunks.
- As the description says, the older phrase survives at kernel/kernel.py:67712 and in the class docstring at tests/test_kernel_webpush.py:1639, both outside this change.

## Smaller points

- SECURITY.md:68 says the recorded origin "becomes the `navigate` URL of the device's next declarative Apple push". That clause carries the wording of the comment posted on #1953 at 08:33Z, which put the consequence in those words and asked for it as a clause, so the sentence is that comment's, not a misreading in this change. Read at this head: `_push_declarative` (kernel/kernel.py:58128-58143) builds the field at 58137 as the origin with its trailing slash stripped plus the payload's deep link path, and the wire helper at 58146-58162 sends that message for an Apple endpoint with an origin on file. The test at tests/test_kernel_webpush.py:1067 asserts the full URL, the origin plus `/?push-reveal=...`, and passes at this head. A reader who takes the sentence literally expects a tap on the next notification to open the recorded origin's root; it opens the deep link on that origin. The security consequence stands, since the origin is the part a chosen header picks, but the stated URL is wrong. The corrected clause: "which matters because the `navigate` URL of the device's next declarative Apple push is built on that recorded origin". A shorter form is "becomes the origin of the `navigate` URL". One clause either way, and no other line moves.

## Tests

- tests/test_docs_front_pages.py:206 and 211, read at this head, not executed. The exemption at line 150 reads the nearest prose line above the paragraph's first line and asks whether it opens with a heading mark. The committed snippet `NAV_UNDER_HEADING` (206) has a blank line between the heading and the link, and so does the only exempt paragraph on the governed pages, README.md line 76 under its heading at 74. In both, the heading is also the nearest prose line above the blank line, so a lookup that read one line too far up, `above.get(line - 1, "")` in place of the paragraph's own line, finds the heading in both and passes every committed test. The shape that separates the two readings, a link line on the line directly after its heading, is exempt at this head (the line above it is the heading) and a violation under that lookup (the line above the heading, empty in the snippet), and no committed test carries it. The fix is a fourth snippet with no blank line, `"## License\n[Apache-2.0](LICENSE).\n"`, asserted to yield no violation beside the first assertion at 211.

PR #1958, opened 2026-09-21
The chat baseline seed reads delivery off its own loop's echat writes, so a full another sender hands a loop client after its status send pops and marks nothing

lczh:
Tier: fix

**The defect.** The baseline seed that #1943 gated on delivery read its signal after the send loop: whether some client in the loop held a base for the session once the loop had run. That is a state read, not the loop's own event. A base another whole-frame sender wrote on one of the loop's clients between that client's status send and the read counted as this loop's delivery, so a pass that handed its build to nobody (every client holding the tab as a skeleton) seeded all the same, and since its list differed from the one the other sender had just seeded, the detector popped the other sender's list and marked the session: one pop and mark, and the next cycle re-sent every base holder the whole session as a change-0 full with a diagnostic row each. The reachable shape is a targeted push (a backend thread, no lock across its loop) over a tab every page holds as a skeleton, preempted after its last status send while one page's reader thread answers that page's ask for the tab whole with the connect push that hands it the full and seeds.

**The fix.** Delivery is reported from the write itself. The two session-frame senders that write a client's base entry (the locked index-wire sender and the proto-2 sender it delegates to) record the session on a per-thread delivery ledger beside each of their entry writes; each seed call site (the pusher's connect road and the targeted push) opens the ledger around its own per-client loop and seeds only when that loop recorded the session, in place of the read of the clients' bases after the loop. The ledger is a thread-local set opened by a context manager, so a sender that runs inside another's loop on the same thread records on its own ledger and hands the enclosing one back untouched, and another thread's writes never reach it. The signal means an entry was written for this frame: the proto-2 sender's empty list writes none, a socket before its ready writes none, a status frame and a withheld client write none, and the lazy serialization's state is not consulted (the proto-2 sender never materializes it). The entry has one other writer, the history reply's edge advance in the dispatcher, which only rewrites an entry the client already holds and hands the client a history reply and no session frame: it is deliberately not a delivery and records nothing. The helper that read the clients' bases is removed, since nothing read it, so the lock-discipline pin the review proposed for it has nothing left to pin. Both call sites keep their absent-or-empty baseline check.

This change follows the seed-on-delivery fix (#1943) and the empty-build fix (#1954) from the same review, both merged; it sits on the second's commit.

**Wording.** The proto-2 sender's comment on the empty-list pop names the delivery signal as the third reader of the entry's writes and states that the pop records nothing on it, in place of its closing sentence about a future reader. The seed's docstring opens with "first whole frame to reach a client", as the reference and the meter's docstring do, describes the write-time signal where it described the helper, and cites tests 29 to 33 and the third part of test 11_d for the no-delivery roads.

**Validation.**
- Red first, the kernel swapped to the base's text and restored byte for byte: test 38 of the skeleton-reconnect module fails at its mark assertion (the session marked with count 3, the other sender's baseline popped). Test 39, the pin for the targeted push's absent-or-empty check, is green at the base and red under the drift it pins (that site changed to a None-only check: the first content frame over an empty baseline seeds nothing).
- Green after, one module at a time: skeleton-reconnect 63 (61 at the base), chat-delta-resync 12, kernel-pusher-snapshot 31, cold-tab-gate 22, chat-pages 43, chat-split 55, chat-window-spans 8, kernel 460, perf-bench 49, perf-stats 66, chat-fixed-cost-memos 44, chat-empty-build-guard 5, injected-voice 10; the getsource-pinned modules feed-first-cold-push 7, files-pane 27, first-cycle-stage-split 19, kernel-tabs-first 7, pusher-change-gates 13, settings-page 10, single-flight-builds 11, stat-rule-readers 15, timeline-skeleton-dedup 4, wire-once-per-build 24, comment-threads 115.
- The static pin scan over 8770 source literals lost nothing; the source-text pins on the two senders and the two call sites hold as written.

Follows the post-merge review of #1943.


ksarma:
Evidence for this note was executed against 347e21d82, the merge commit of #1958, whose second parent is the PR head 9c21bb3d1; every file and line reference here is on 347e21d82. The merge's other side changes kernel/kernel.py only in the warmer and the feed memo, so the cited code is unchanged from the PR head; there the lines from 33988 to 34017 carry the same numbers, every kernel line cited from 51648 on is numbered 28 less, and the test file is identical at both.

The change moves the baseline seed's delivery signal from a read of the clients' bases after the send loop to the loop's own writes. `_chat_delivery` (kernel/kernel.py:33992 to 34005) opens a fresh set on the thread-local at line 33988 and restores the enclosing set in its finally block. `_note_chat_handed` (34008 to 34017) adds the sid to the open set, and is called beside each of the four echat entry writes (51648, 51734, 51822, 51835). Both seed call sites open the set over their per-client loop (55016 and 55661) and read it after the loop (55072 in the pusher, whose loop serves the cycle and the connect push; 55664 in the targeted push), each still behind the absent-or-empty check.

Read and executed, the body's account matches the code. The entry's four writers each note on the next line, and the history reply's edge advance (69433 to 69435) rewrites only an entry the client already has and notes nothing; a status frame to a client with the tab as a skeleton, a withheld client, a socket before its ready and the proto-2 empty list write no entry and record nothing; and the two tail-road notes cannot decide a seed, since with no baseline `_chat_diff` returns 0 and neither sender takes a tail branch. The same-thread nesting is pinned by execution: restoring None in place of the enclosing set at 34005 turns tests 23, 24, 25 and 34 to 37 red, and a nested loop that reuses the enclosing set turns 31 and 38 red. The notes and the gates are pinned too: without the proto-2 full road's note at 51734 more than fifteen tests go red, 38 and 39 among them; without the index-wire full road's note at 51835 test 11_d goes red; the connect-site gate made unconditional turns 11_d, 29, 31, 32 and 38 red, and the targeted-site gate made unconditional turns 11_d, 30, 33 and 38 red. Red-first reproduces: with the base kernel (331c11210) under this test file, test 38 fails at its mark assertion (tests/test_chat_skeleton_reconnect.py:2142) with the session marked at count 3, and test 39 passes at the base and goes red only under a None-only check at the targeted site. The module is 63 passed at the merge commit, and the check runs on the merge commit and on the PR head all completed green.

## Tests

- The property the fix rests on in production, that another thread's writes never reach this thread's delivery set (the targeted push's backend thread against a page's reader thread running the connect push), is pinned by no executed test. Every test in the module drives the race on one thread through the nested stack: test 38's hook runs the connect push inline through `Handler._dispatch_ws` on the targeted push's own thread (tests/test_chat_skeleton_reconnect.py:2130), so the restore in the context manager's finally block is what keeps it green, and the module's one threaded test, test 26 (lines 1544 to 1547), drives the seed directly with no set open. In a scratch copy at 347e21d82 with kernel/kernel.py:33988 changed from `threading.local()` to a process-global object, the module stays 63 passed (63 at the merge commit too), the threaded modules test_single_flight_builds (11), test_queued_copy_identity (29) and test_chat_fixed_cost_memos (44) stay green, and no file under tests/ names the thread-local, the context manager or the note helper. A two-thread probe (open the set on the main thread; a joined thread calls `_note_chat_handed(S3)` with no set open; a joined thread opens its own set and notes S2; the main thread notes S1) reads {S1} on the main thread and {S2} on the other at the merge commit, and {S1, S3} on the main thread under the mutant. The consequence: a later edit that makes the record a module-level set or a shared object (a refactor, a memo) reintroduces the defect this PR fixes while the module stays at 63 passed; another thread's echat write on a loop client counts as this loop's delivery, the status-only pass seeds, pops the other sender's list and marks the sid, and the next cycle sends every client with a base a change-0 full with a row each. One refinement to that consequence: with the stack kept and only the thread-locality dropped, the base's defect needs the reader thread's set opened before the targeted push's (A opens, B opens, B notes, B closes, A reads is clean under the mutant; B opens, A opens, B notes puts B's write on A's set and loses it from B's own), so the same mutant also makes a connect push that handed a full seed nothing, and a module-level set fails under every ordering.

  A fix: an executed test beside test 38 that opens `km._chat_delivery()`, starts and joins a thread calling the note helper with S3 and no set open and one that opens its own set and notes S2, then notes S1 on the test's thread, and asserts this thread's set reads {S1} and the other thread's {S2}; green at 347e21d82, red under the process-global mutant, join-based with no timing. The isolation is carried by the one assertion that this thread's set reads {S1} (the no-set thread's S3 lands on it under the mutant); the other thread's own set reading {S2} is true under both, so that half pins only that a thread's own loop records, not the isolation. The test's comment can state why the pin is on the set being per thread rather than on test 38 replayed across two threads: test 38's order (the targeted push's set opened first, the connect push's opened and closed inside it) is clean even under a process-global stack, so a two-thread replay of it would not go red; the production order the mutant breaks is the reader thread's set opened before the targeted push's, where the mutant both counts B's write as A's and drops it from B's own. An optional second assertion covers that lost write: open the other thread's set before this thread's (an Event), note S2 on it after this thread's is open, and assert it still reads {S2}.

PR #1960, opened 2026-09-21
plans/judge-prompt-experiments.md: road (b), the labeller's class a stratification frame and the measures on the user's own actions

henrypinkard:
Documentation only, road (b) after the pilot's labeller gate failed. NOT merged before the user's word; the harness change follows the sign-off.

**Why.** The pilot's labeller gate (the labeller's finished/loose-ended read against the user's own recorded action) came to 31.6 percent, 6 of 19. Every disagreement ran one way: the user cleared a card (a "finished" gesture) whose last turn the labeller read as an open question, an undone item or an offer. That is the habit the change serves ("Completed must be safe to clear without reading"), not a labeller fault, so a user clear is not a truth about the ending's shape and cannot validate the labeller's class.

**What the amendment changes.**
- **Labelling** is now a stratification frame, not a ground truth. The labeller's class groups endings so a leak or a false interrupt can be read per class; its own gate is STABILITY (the same class in both shuffled orders), 90 percent, which the pilot met (54 of 60, 90.0). The user's own recorded actions are the ground truth the measures use, and need no labeller.
- **The measures** move onto the user's later gestures, matched per node to the top the arm placed and postdating the arm's placement time (the placement's own `ev_t`/`at` from the live journal, never the arm's clock). A **leak** is a Completed card the user then followed up, unclear-ed or restored. A **false interrupt** is a Needs-you card the user cleared with no reply (no `followup`/`unclear`/`restore`, no answering turn between the placement and the clear). Flaps and cost are as before, and every ending contributes to them; an ending with no later gesture contributes to flaps and cost only.
- **Sample sizes** the pilot's corpus gives: 60 endings, 21 with a later user gesture (15 the user treated as finished, 4 re-opened); the full corpus of 300 will carry proportionally more.

**Verification.** No code; the added lines carry no em-dash, none of the ruled-out vocabulary and no private string; no corpus text. Tier: docs.


ksarma:
The evidence below was executed against f5f4a9247, the merge commit of PR 1960, whose second parent is the PR's head 8791ba44f; the plan differs between the two by one line from earlier merges, so every file and line reference is on f5f4a9247, in plans/judge-prompt-experiments.md unless a path is given. The pilot's corpus and labels file were not available, so 6 of 19, 54 of 60 and 21 of 60 were checked for arithmetic and mechanism, not against the data. Items marked executed ran in a scratch copy, the rest read only.

- **plans/judge-prompt-experiments.md line 33 describes an ask the labeller does not receive, and line 31 decides on figures produced under one of two known defects.** Executed. `_last_assistant_text` (scripts/judge_experiment.py, 937 to 949) appends the transcript's FIRST non-meta user text as the ask, so every ending past a session's first turn is labelled with an earlier turn's ask. The plan's line 3 dates the pilot after the two contributors' reviews (PR 1946), and this PR's first commit follows the PR 1956 merge by 21 minutes, so 6 of 19 and 54 of 60 were produced under this input or, at 98880abf3 and earlier, under the `SETTLE_S` tail, where the wrong side is tier one instead; no head is free of both. In direction probes a session-opening ask moved finished-shaped endings to stably undone (2 of 5, 1 of 8, 2 of 3), the direction the pilot observed, so the pilot alone cannot carry the premise that 31.6 percent shows the clearing habit and not a labeller fault; the design stands on its own reasons either way. The disputed set is 13 of the 15 tier-one-finished endings, not 13 of 19. Fix: re-run `tier_one_label` at the merged window first, then the labeller on the 19 tier-one endings with no ask, reporting the agreement by class, or have a human read the 13 disagreeing last turns; state at line 31 which head labelled the pilot and which defect was live, and that 54 of 60 needs re-measuring if the ask fix lands. For line 33, write the merged behaviour or, preferably, land the ask fix keyed on the ending turn's own boundary (the first non-meta user text at or after the manifest's `startT`, none when `startT` is None), read from the event model's turn opener.

- **plans/judge-prompt-experiments.md line 55 scores the clearing habit as a false interrupt on every card the candidate moves.** Executed. `append_clear` (kernel/judge.py, 5284 to 5296) writes node, op, src, why and t with no column, so a Completed cross-off cannot be told from a Needs-you dismissal. On one synthetic ending the baseline reads `completed`, false interrupt 0, and the first candidate (kernel/judge.py:13296 inverted) reads `needs_input`, false interrupt 1, so the line 61 gate fails: every by-habit clear on a card the candidate moves counts against the candidate and never the baseline, refusing the move the experiment exists to validate; the answering-turn clause cannot exclude a clear with no reply. `CLOSER_SYS` already refuses to close on a question, so the realized exposure is the next-step share of the 13; line 55 is text only, since the harness still scores by class (scripts/judge_experiment.py, 812 to 815). Fix: score a false interrupt inside the labeller's finished stratum, per class: a labeller-finished ending where the arm placed a scored top in `needs_input` and no followup, unclear or restore followed on that node (15 of 60), with the clear a reported sub-count, since requiring it leaves about 2 of 60. Lines 33 and 70 then need softening for this one measure, which the stability gate does not validate; the harness change lands with or after the ask fix, since the ask defect under-populates this stratum; and conditioning on the live column at clear time instead would blind the measure to the candidate's new false interrupts.

- **plans/judge-prompt-experiments.md line 54 reads every followup as a re-open, and line 61 gates on zero.** Executed. Any reply that cites a card writes the followup op: `optimistic_followup` (kernel/judge.py, 10356 to 10379) fires for every `askFollowUp` carrying an `itemId` (kernel/kernel.py, 19935 to 19970), from the citation chip, card composer, Check status and Continue (ui/webview/feed.ts:3731, 3420, 3807, 1425, 3820; render.ts:16691), and the row is node, op and t with no text (kernel/judge.py:5280); an extension and a rejection reply wrote byte-identical rows and both reopened. So a fresh ask on a finished card is a leak for every arm, and at a gate of zero one such followup on a top the correct candidate completes fails every arm; the harness keys leaks on the class (812 to 813). Fix: gate leaks at zero within the loose-ended strata (classes 1 to 3) and at or below the baseline overall. The other shape, a followup as a re-open only when the class is loose-ended with unclear and restore unconditional, drew opposite readings on whether it reinstalls the labeller as the truth line 33 forbids. A Continue writes the same row while saying nothing is needed, so under a labeller-finished class it must not count as a leak; the class-conditioned shape handles that and the strata gate does not.

- **plans/judge-prompt-experiments.md line 52 keys "later" on the turn's start and on a source absent for the cards that matter.** Executed. A closer done's evidence time is the turn's start (`_close_turn` passes the turn's t, kernel/judge.py:15086; `apply_close`, 14789), a block row carries no `endEv` horizon (`file_block`, 13840), the arrival stamp `at` (10612) is forensics only and has no live row for an arm-only placement, and a top cleared mid-turn is sealed out of the live closer's menu (`_turn_menu`, 14387) while line 42 reopens it in the seed, so no live row exists for that card. The harness reads the same time (`event_time` prefers `ev_t`, scripts/judge_experiment.py:54; `tier_one_label` counts ops later than it, 927): a clear 60 s after a synthetic turn's start and before its cut returned finished, so a mid-turn gesture made before any placement existed counts as answering it. Fix: key later on the ending's cut, `cutT` in the manifest for every ending (592), for every top the arm placed, live row or not, since every arm verdict uses evidence up to the cut; delete the "or the live row's `at`" alternative. Recount the pilot's 21 gesture endings under the cut key before quoting them.

## Smaller points

- **plans/judge-prompt-experiments.md line 59's sample accounting.** The 15, 4 and 2 split against an agreement denominator of 19 has a bucket the harness does not produce (`tier_one_label` returns one collapsed label per ending, 934; the summary drops unstable reads, 1024), and `FINISHED_OPS` (882) counts resolve as finished where line 52 names only clear: state what the 19 is and what the 2 lacked, and whether resolve counts. "Proportionally more" at 300 assumes a random draw, but `build_corpus` takes tier-one-eligible endings first per class (538 to 548): report `tierOneEligible` per class from the pilot manifest (592), derive the expected gesture endings at 300, and restore the selection-preference sentence.

- **plans/judge-prompt-experiments.md line 33's gate is met at the boundary.** 54 of 60 is 90.0 against 90, read as a pass with no word on a full run below it: name the consequence (drop the class column, or re-prompt and re-run).

- **Two vocabularies survive the amendment.** plans/judge-prompt-experiments.md line 68 says the labeller's agreement with the recorded actions is reported before its labels count, against line 33's stability-only gate, and lines 14, 20 and 70 keep "first tier" and "tier one" after its definition left the Labelling section: make line 68 a diagnostic that gates nothing, or delete it, and restore a one-line definition of tier one or write "the recorded actions".

- **Two operations are undefined, and the Tests section (73 to 79) pins none of the new mechanics.** The answering-turn clause at plans/judge-prompt-experiments.md line 55 names no predicate and no input, and the node match at line 52 on the `rompUuid:g<N>` key cannot be met literally, since line 42 re-keys every id in the arm's copy to the ending id and arm-minted to-do nodes have no live twin: define the answering turn (a turn opened by a typed message between the reference time and the clear, read from the live transcript through a `--claude-root`), state the join (the ending's live sid plus the `g<N>` tail; arm-minted nodes unscorable), and list pins for these, the later-gesture time key and the per-class reads.

ksarma:
Follow-up to the post-merge note (https://github.com/romp-on/romp/pull/1960#issuecomment-5777730326), read at b598eae95. Under road (b) the measures score recorded gestures, and the labeller's class only groups them. Under PR 2035's ending-turn ask the labeller re-run reproduced 6 of 19 (plans/judge-prompt-experiments.md line 5), and the same PR took the cut key, the per-class gates and the labeller's described input. Two points from the note still stand, and the lift rule added in PR 2006 carries a third. Each bears on how the full run's result reads against the landing bar:

- plans/judge-prompt-experiments.md line 63 keys every top the arm placed on the cut, live row or not. scripts/judge_experiment.py still skips a top with no live placement in the turn (1323 to 1326), citing the missing placement time that the cut at 1330 replaced. Such a top scores no leak or false interrupt whatever the arm filed on it, and an ending whose live judges placed no top in its turn reads unplaced (1401 and 1402).
- The lift that turns a clear into answered then cleared is read from the live log alone (1345 and 1346). A block the live judges never filed, such as a candidate's block on a top the live closer marked done in the turn, has no live ruling to read, so a plain clear on it counts as a false interrupt whether or not a later turn answered it. The statement at plans/judge-prompt-experiments.md line 66 that the missed-ruling bias runs against every arm equally is false for those blocks.
- The labeller-finished false-interrupt count, the one the bar gates, requires a clear (1347), so it can rise above zero only in labeller-finished endings with a cross-off after the cut. The harness's measures output does not report how many there are, and a tie at zero over a few of them says little about whether a candidate spares finished endings.

PR #1965, opened 2026-09-21
An empty chat build under the detector's mark hands no client the cached build, since a stand-in under an absent baseline always leaves as a full and rewinds every holder ahead of it

lczh:
Tier: fix

**The defect.** When two whole-frame senders race on a chat session that has no delta baseline yet, the seed's detector pops the baseline and marks the session; the mark says "every client holds content, but no one list describes what all of them hold", and the next cycle whose loop reads the baseline absent repairs everyone with a full. The empty-build guard (a transcript read that comes back empty is a failed read, not a conversation that emptied) learned the mark in the change this follows, and with nothing cached it correctly sent nothing. With a CACHED build it did the other thing: it handed the cached list to every base holder as a stand-in, and the write block then made that list the baseline and took the mark off. That stand-in reasoning was written for a session with a baseline, where the cached build IS the baseline's list and the diff sends nothing. Under the mark the baseline is absent, so the stand-in always left as a whole frame, and it rewound every client that was ahead of it. Two faces: a targeted push under the mark hands every client a longer list and caches nothing, so the next empty read sent the shorter cached list and the newest cards left every pane until the next content cycle (the shorter face); and in the race's other order the cache keeps the older list with a card unfilled while one client shows it filled, so that client was shown the unfilled card again (the same-length face). A length compare against the mark's count would close only the first, since the count sees length, not content.

**The fix.** The guard reads the mark once and takes the no-cache road under it: a marked session's empty build sends nothing for that session, cached build or not, and the write block is skipped, so the mark and the absent baseline stand and the next content cycle's full repairs every base holder, deduping on the slot of a client already showing that list. The release of a claimed single-flight build stays inside the widened branch, where it lets the next builder of the tab in before the loop moves on. Under the mark with a cached build, that build still rides the cycle's session list, the one source of the feed frame's ledgers, which the Outline pane replaces wholesale on every feed frame: without that append the session's Outline row left the pane for the empty cycle and came back with the next content cycle, a payload change driven by a failed transcript read (the stand-in road had kept the row). The targeted push's guard needed no change. The connect push runs the same guard, a third road: a page dialing in for a marked session whose read comes back empty gets the strip and every other tab and no frame for that session until the next content cycle, with nothing cached (as before) or with a cached build (where the stand-in used to go to the target). The cost: in the race's first order the repair of the older holder lands one cycle later than the stand-in road delivered it, and no client is ever shown a card older than the one it shows.

**Wordings, same change.** The stderr line for an empty build follows the count's source: with a baseline it says the baseline held N events and the previous build is kept until content returns; under the mark it says the clients hold at least N events and nothing is sent for the session until content returns. That count is documented as what it is: the length of the longest list some client was handed whole under the mark (the longer of the two lists at the pop, raised by every whole frame a sender hands while the mark stands), a minimum for the note and not what every holder has, since the raise follows a connect push's full to one client, the pop's max can name a count no holder has, and a cycle that sent tails hands nothing whole. Five sentences from earlier changes still said the next cycle's write always repairs every client and clears the mark (two in the cycle's read-site comment, one in the connect arm near the write block, one in the targeted push's read comment, one in the reason meter's docstring); each now states the code's rule: a cycle that sent tails leaves the mark, and the next cycle whose loop reads the baseline absent repairs and clears. The collector's snapshot-shape docstring names `coldSkipped`, `baselineRaced` and `baselineRepaired` beside the other chat counters, noting that `romp perf` prints none of the three by its documented contract. The guard's comment names the two faces, the connect road, the Outline row, and the seeded case's pre-existing road (a targeted push's tail leaving both clients ahead of a seeded baseline sends the stand-in as a lastGone full at both heads; outside this change); the guard helper's docstring lists the seeded road's test apart from the marked case's.

**Pins, all in the skeleton-reconnect module (67 at the base, 76 now).** Test 37, which pinned the stand-in road, is re-pinned to no frame for the session, the mark standing, the baseline absent, the counters at one pop and no repair, then one repair after the content cycle. Test 44 is the shorter face; test 45 the same-length face (the race's other order on the cached-hit road); test 46 the connect road, both arms. Test 47 replays the tails cycle after a collector reset and reads the counters at (1, 0) then (1, 1); test 48 replays the connect push under the mark and reads (1, 0); test 49 pins the pop's count by direct seed calls, 6 for a seeded 5 against this sender's 6 and 6 for a seeded 6 against 5; test 50 runs an empty build after the tails cycle with that cycle's 6-event build cached and asserts no frame, the standing mark, the absent baseline and the note's count of 5 (the pop's, since the tails cycle handed nothing whole). Test 51 installs a feed build and an Outline client and asserts the marked session's row in the feed frame's ledgers on the empty cycle with a cached build, while no frame for the session goes to a base holder. Test 52 pins the single-flight release's timing: at the loop's first send for another tab the claim is already gone from the inflight map and the thread's scope. The stderr wording is pinned at every marked site and the seeded wording in test 22. Test 27's interleaving became a helper the replays share; its assertions are unchanged.

The branch was cut above the cost-partition docs change (#1959), merged since; the diff against main is this one commit.

**Validation**
- The whole module at the base, with the kernel swapped to the base's text and restored byte-exact: 10 failed, 66 passed. Behavior reds: test 37, a 5-event session frame at each base holder where none is expected; test 44, `[5] != []`, a 5-event frame at each client showing 7; test 45, `('session', 'm3') != ('session', 'm3 filled')`, the client showing the filled card was rewound; test 46, the marked session found among the target's session frames (the cached arm); test 50, a session frame at each client where none is expected; test 51 passes its Outline row assertion there (the stand-in rode the list) and is red where the stand-in frame reached a base holder; test 52 is red where the mark is asserted standing. Wording reds: tests 22, 35 and 36 at the stderr sentence's new words (the base's line said the previous build had N events). Tests 47 to 49 are green at the base by construction: the counters and the pop's count are unchanged there.
- The first cut of this change, which took the no-cache road under the mark without the ledgers append: test 51 red at the Outline row assertion, the session missing from the feed frame's ledgers (three rows of four).
- Mutants, each applied to the final kernel and reverted: a repair counted wherever the write block finds a mark, tests 47 and 50 red `(1, 1) != (1, 0)`, 74 passed. A pop counted in the seed's declining arm, tests 48 and 44 red `(2, 0) != (1, 0)`, 74 passed. The pop storing this sender's length alone, test 49 red `5 != 6` (seeded 6 against 5), 75 passed. The pop storing the seeded list's length alone, test 49 red `5 != 6` (seeded 5 against 6), 75 passed. The single-flight release removed from the widened branch, test 52 red `(True, True) != (False, False)`, the claim still held at the next tab's send, 75 passed.
- Green at the head, one module at a time: skeleton-reconnect 76, chat-delta-resync 12 (its regex pin on the write gate untouched), chat-empty-build-guard 5, cold-tab-gate 22, kernel-pusher-snapshot 31, single-flight-builds 11, chat-pages 43, perf-stats 66, perf-bench 49, kernel 460, outline-provisional-rows 10; and every module the getsource scan names for the changed functions: feed-first-cold-push 7, files-pane 27, first-cycle-stage-split 19, kernel-tabs-first 7, pusher-change-gates 13, settings-page 10, stat-rule-readers 15, timeline-skeleton-dedup 4, wire-once-per-build 24.
- The static pin scan checked 8783 assertIn literals against the merge base and lost none.

Follows the post-merge review of #1954.


ksarma:
The evidence below was executed against 7440461ff, the PR's head, merged as c507b6b45. File and line references are on 7440461ff: the merge tree carries the test file unchanged, while `kernel/kernel.py` there has moved by the merges before this one (the guard's branch at 55060 reads at 55089 there).

The change closes what the earlier comment on #1954 drove, by execution: the head's test module against the base kernel (311f7dd40) reads 10 failed, 66 passed, the body's list exactly; the stand-in road restored under the mark as a mutant of the head (`if hit is None:` with no append) reds 37, 44, 45, 46, 50, 51 and 52; and probes at the head find an empty cycle under the mark with a cached build handing no base holder a frame while keeping the mark, the absent baseline and the Outline row, the targeted push under the mark sending nothing, and the connect road handing its target no frame on either arm. What stands of that comment is only what it named as outside the diff, the seeded road's lastGone stand-in, now named in the guard's comment.

## Smaller points

- The empty-build guard reads the mark at 55028 and the baseline at 55029 outside `_chat_baseline_lock`, while the seed's pop-and-mark at 34145 to 34147 is one step under it, so a pop-and-mark landing between the two reads leaves the guard seeing neither and every base holder gets a 0-event session frame with no stderr line, the mark standing and the baseline absent (executed at the head, and the targeted push's guard at 55736 to 55737 gave the same result; the cost is one cycle, since the next content cycle reads the baseline absent, sends the change-0 full and clears the mark); fix, verified at both sites: `with _chat_baseline_lock: marked = m["id"] in _chat_baseline_raced; _prev_now = _prev_chat_events.get(m["id"])`, both passed to the guard helper, with no deadlock hazard since the three lock sites (34130, 55122, 55173) contain only dict operations and the seed's list compare, none taking a lock, with the "read once" comment at 55028 and the sentence at 33941 to 33942 in `_chat_prior_n` then going stale (the count itself may stay unlocked; it is a stderr figure), and with the pin running the seed on a second thread, since a one-thread hook deadlocks on the non-reentrant lock.

- The cached arm under the mark (55060 to 55062) appends the cached build to `chat_sessions` and sends nothing else, so a skeleton holder of the marked tab gets no status frame for it on the empty cycle where the removed stand-in road sent one (probed at the head against the base; the next content cycle sends it, a one-cycle lag); fix, applied beside the append: `for c in chat_clients: _send_light_status(c, s["sid"], hit[1].get("status"))`, the helper re-checking skeleton membership under the client's lock so no base holder gets a frame, sending the cached build's status as the safer default (the live-row status returns None with no live row), and pinned by a skeleton holder beside the base holders in test 46's or 51's shape expecting one status frame and no session frame for the marked tab on the empty cycle; disclosing the gap in the guard's comment instead would also fit.

- The Outline-row rule applies to the cached arm only: a marked session and a seeded session with no cache entry both leave the feed frame without the session's row for the empty cycle, at the head and at the base alike, and the comment at 55057 to 55058 says the no-cache arm has no build to append, but `m`, the empty build, is in hand; the fix is the comment rather than the append (tried: appending `m` keeps the row but builds it from the empty parse, so `_session_chip` blanks a working session's pip and the recency stamp goes null, the class of failed-read change test 51 closes), so 55057 to 55058 says the no-cache arms keep the gap because the only build in hand would flip the row rather than fit it, "no build fit to append".

- At 34060 in `_seed_chat_baseline` the strand-face parenthetical says the pop and the mark follow and the next cycle's fulls repair, where the four sites the earlier comment listed and the same docstring's 34102 and 34110 to 34112 say the next cycle whose loop reads the baseline absent repairs, a cycle that sent tails leaving the mark (test 27); the sentence arrived with #1959, outside the earlier comment's four sites, so 34060 takes the qualifier the other sites carry, leaving the pure-cost sentences at 34067, 51570 and 55709 as they are since they describe cost, not repair. Separately, 34101 and 34147 define the stash as the popped list's length while the code at 34147 stores `max(len(cur), len(evs))` and 33557 to 33558 and 33937 say the longer of the two lists at the pop, which test 49's first case tells apart (a seeded 5 popped against this sender's 6 stores 6), so both take "the longer of the two lists' length", the more exact reading since the stash is a count; comment-only, no test owed, since test 49 pins the max and no test asserts on the comment text.

## Tests

Test 45's repair half asserts a proxy: at tests/test_chat_skeleton_reconnect.py:2471 it calls `_the_cycle_repairs(a, b)` without clearing the frames first, so the helper reads a's last u3 card as the filled full T handed a in the race whether the content cycle re-sent a or deduped it, and with the session-full dedup disabled as a kernel mutant the shipped test still passes while the module's 12 pre-existing dedup pins go red, so the composition no test executes is the marked repair cycle handing nothing to a client already showing the list. Fix, green at the head and red under the mutant (`[5] != []`): in place of the helper call, clear both clients' frames and the cache, run the cycle, then assert no session frame at a, one 5-event frame at b, b's u3 card filled, and the baseline established as the helper asserts. Land it inline or as a helper variant and leave `_the_cycle_repairs` as it is, since nine other tests call it and 44 and 50 clear the frames before the empty cycle, so its read is not a proxy there.

PR #1969, opened 2026-09-21
Live dividers lab: the cost pin counts the page's timed callbacks over 40 ms in every document, observers and pointer listeners included

henrypinkard:
The live dividers lab's cost pin (`tests/test_live_dividers_served.py`, PR 1927) counted only the bodies of the shell's `requestAnimationFrame` callbacks, so the page's other work per frame fell outside it: layout and paint, observer callbacks (the chat's own reflow rule among them) and the pointer handlers. A 60 ms busy loop in the chat's view `ResizeObserver` callback passed the pin at 15 frames per second with 84 long frames, and an 80 ms busy loop in the gutter's `mousemove` handler passed at 23 frames per second with 96 long frames (the seventh review), while the body said the pin held "the page's own work per frame".

The count is widened to what the words claim. A script installed before every document in the lab's browser context (the shell and each pane frame) times the bodies of the `requestAnimationFrame` callbacks, the `ResizeObserver` and `MutationObserver` callbacks, and the `mousemove` and `pointermove` listeners registered through `addEventListener`, counting each run over 40 ms; the cost leg resets the counters in every frame at the drag's start and sums them at its end. The pin stays a COUNT, at most forty of some 2300 timed callbacks over the drag, chosen between load and the mutants: the control 0 (twice, the longest callback 5.7 ms); a half-core quota 13 and 16 (a cgroup quota throttles the process inside a callback now and then, so the earlier bound of ten, cut when only the shell's 98 frame callbacks were timed, would red load); the 60 ms busy view observer 168 (twice); the 80 ms busy pointer handler 96 (twice). The frame rate, the long-animation-frame count, the layout and style-recalc counts over CDP and the longest timed callback are reported beside it, not pinned.

The docstring says the lab pins the count of the page's timed callbacks over 40 ms (not the longest callback) and the instrument's comment says the same; the fixture and the rest of the lab are unchanged. Red-first per test on the two mutants (the product files patched, the bundle rebuilt, the lab as written): see the first comment. Whole extension suite (served labs under REQUIRE) and whole Python suite on the head before the push.

Tier: fix


henrypinkard:
**Measured reds on the two mutants the seventh review named** (the product file patched, the bundle rebuilt where the mutant is in it, the lab as written on this branch), and the readings that set the pin:

- **A 60 ms busy loop in the chat's view `ResizeObserver` callback** (`ui/webview/render.ts`, the observer that runs `followReflow`): `cbLong` 168 of 2324 timed callbacks, the longest 60.4 ms, red on the bound of forty (twice; 13 to 14 frames per second, 83 long frames). The merged lab's count over the frame callbacks alone read 0 there (the review's run: 15 frames per second, 84 long frames, `cbLong` 0).
- **An 80 ms busy loop in the shipped gutter's `mousemove` handler** (`kernel/kernel.py`, `mv`): `cbLong` 96 of 2412, the longest 80.1 ms, red (twice; 24 frames per second, 96 long frames). The merged lab read 0 there too (23 frames per second, 96 long frames).
- **The control**: `cbLong` 0 of 2292 and 2293 timed callbacks across nine documents, the longest 5.7 and 5.2 ms, 60 frames per second, no long frame.
- **Under a half-core quota**: `cbLong` 13 and 16 of about 2350 (22 to 24 frames per second, 66 to 68 long frames, the longest callback 101.6 and 81.5 ms: the quota's throttling landing inside a callback), within the bound of forty; the earlier bound of ten, cut when only the shell's 98 frame callbacks were timed, would red here (a cgroup quota can throttle the process inside one callback; the count tolerates a few).

On this head the lab is 10 of 10 (the cost leg: 2290 timed callbacks across nine documents, none over 40 ms, the longest 5.1 ms, 60 frames per second, no long frame) and the two DOM-stub harnesses 20 green. Whole suites on this head before the push: the extension suite 5190 tests, 5184 passed, 6 skipped, 0 failed (exit 0), the served labs under REQUIRE; the Python suite 14930 passed, 57 skipped, 2640 subtests passed, 0 failed, 0 errors (exit 0).

PR #1970, opened 2026-09-21
End hands a message parked behind a compaction back as not delivered instead of dropping it

lczh:
Tier: fix

Follows the second executed review of the native compaction (#1933, merged as 492f163a). The branch was cut on that pull request's head, so the diff against main is this one commit.

## The defect

A message typed while a Codex session showed "compacting" could vanish without a word. The case: Codex acknowledges a compaction and never runs it, so the session keeps reading "compacting", and a message typed meanwhile waits in the kernel's queue behind the cue. The Codex doc said the way out was End then Revive, which "keeps the thread and its history and delivers the waiting message". It did not deliver it. End left the queue where it was and finished by waking the pusher, so the queue drain ran in the same wake: it popped the waiting message and handed it to a session that by then read as owned by nobody (a dead row is not owned), and it ignored that route's refusal, so nothing reached the chat. Revive drains only the backend's own queue, which never held the message. Only a kernel restart delivered it, because the parked queue is mirrored on disk and the in-memory cue is not. Two independent runs over the real backend agreed; the sentence in the doc, in the compact verb's description and in the earlier body was wrong.

## The fix

Every End door (the dashboard's End, the end route, and the self-close sweep) now cancels the ending session's parked ops before the kill and hands each parked message or typed command the user typed back through the existing not-delivered path: a dialog with the text in its copy slot, the text kept verbatim in the undelivered file, and one log line. Only the user's own words come back: a parked op wearing the user flag or a press-minted copy id. A machine's message parks through the same road (a watch notice, the spend-ceiling body, a tagged headless send), and a dialog offering to copy words the user never typed, filed as theirs, would be a false interrupt; those are dropped with a log line naming the kind and never the body, as are the ops that carry no typed text (a queued compaction or clear, a settings pick, a move), since a dead session's queue is never retried (the drain's own contract) and keeping it for a later Revive would strand queued bubbles on a session that is never revived. The op the drain is handing over at that instant is kept by its slot, the way the queued bubble's cancel finds it, not by identity: two parked compaction presses are one interned tuple.

Where the dialog lands: on the pane that pressed End only when that pane can show it (the chat and the feed render the frame). An End pressed in the Sessions pane arrives on that pane's own socket, whose bundle has no arm for the frame, so handed there it showed nothing and reached no chat pane either. That case, and every End with no socket at all (the end route, the self-close sweep), goes to one chat pane: the one watching the session when there is one, else the first live one. A broadcast to every chat pane drew a dialog and a bell entry in every chat column for one message.

The three copies of the recovery sentence now say what happens: a kernel restart delivers the waiting message; End then Revive keeps the thread and its history, hands a message you typed back rather than sending it, and drops anything else queued behind the cue (a compaction, a setting pick, a message a script or romp itself sent) with the session, noted in the kernel log. The compact verb's closing clause, which said the doc names two reachable ways out, now says which of them keeps the message.

## Smaller items from the review

- The launch-error field comment enumerated three keys; it and the kernel's matching helper now name the optional noRetry key and point at the contract.
- The systemError-site lock test asserted the loud record the moment the turn was no longer busy and raced the pump's late write, so a pre-fix run went red on the wrong line; it now waits for the record, and a pre-fix run goes red on the ordering assertion.
- The pulse test pinned only that the remover and the compacting state drop the click cue; it now also pins that a non-compacting bar refresh keeps it, which fails under a remover hoisted above the compacting check and passes at the head.
- On the manual-trigger pin, correcting the earlier body's clause: the golden case pins the normalizer's passthrough of a manual trigger (#1879), and the two backend assertions pin the trigger this backend sends.

Not taken here, as the review itself marked it a follow-up: in a hosted timeline with no shell bell, a refusal frame ends the lane's cue but puts the words nowhere.

## Validation

Red first at the reviewed head, then green with the change, one module at a time:

- Compact route module, the new real-backend End cases, the probe's own sequence (spawn, one turn, a compaction with no status, a typed message that parks, End, one drain pass, Revive, another pass), once per door: red with "the text reached neither the thread nor the not-delivered frame: dropped silently" through the dashboard's End, the end route and the self-close sweep (the log showed the send refused as owned by nobody and the refusal ignored); green after.
- The machine-send case, red against a handback of every parked text: two dialogs, the second carrying words the user never typed ("Lists differ: [typed, machine] != [typed]"); green after, one dialog, the machine's kind in the log and its body nowhere.
- The Sessions-pane case, red against a handback to whatever socket pressed End: the frame on that pane's socket ("Lists differ: [{type: err, ...}] != []") and none on the chat's; green after.
- The two-chat-column case, red against the broadcast: two frames for one message ("2 != 1 : one modal for one message"); green after, one frame, on the pane watching the session.
- The two-press case, red against the identity filter: both compaction ops left in the queue ("Lists differ: [(compact,), (compact,)] != [(compact,)]"); green after, the in-flight slot alone remains and the log names one dropped.
- Backend module, the reworded lock test: green at the head and after, as the review said it should be.
- Status controls test file, run alone from the extension's test bundle: 9 pass at the head and after; under a mutant that hoists the remover above the compacting check, 8 pass and 1 fails, on the new assertion.

Module counts with the change: the compact route module 22, the Codex backend module 171 (+8 subtests), the kernel module 460 (+12 subtests), the send-park module 51, the parked-ops lock module 19, the end-on-idle module 5, the close-confirm module 5, the foreign-sid module 10, the injected voice module 10 (+222), the kernel send module 15. The assertIn pin scan against the merge base finds nothing lost. The whole extension suite was not run here.

Follows the second review of #1933.


ksarma:
The evidence for every point here was executed against 226d0e299, the merge commit of this PR, whose second parent is the PR head b26858b0b. kernel/kernel.py, kernel/codex_backend.py and tests/test_codex_backend.py differ between that head and the merge tree through the merges that landed before it, so every file and line reference is on 226d0e299.

**A parked follow-up comes back as the composed body, not the typed words.** The hand-back at kernel/kernel.py:36486 passes `op[1]` to the not-delivered path, and for a card reply or a Continue press parked behind the compaction that is the composed wrapper: the goal quote when it has a summary, the romp-note and romp-goal-id comments, and for Continue the canned text with its marker. So the copy slot, undelivered.jsonl and the log carry romp's quote and markers as the typed words, where the queued bubble and the live refusal of the same follow-up (19918) carry the typed words alone, empty for Continue. Fix: hand back `_parked_md(op)` for a send and pass the askFollowUp verb when the split's third element is True; the op holds only the wrapped body, and a strip at the marker tail alone leaves the quote block. Test for the romp-canned marker before the strip removes it, and hand a Continue back as empty text, as the live refusal does, or drop it with the kind logged. One inherited edge: with no quote composed, the strip reads a typed reply's leading greater-than line as the quote.

**Several parked texts yield one visible dialog.** The loop at 36485-36488 sends one err frame per text and the page replaces the previous dialog on each (`showConfirm`, render.ts 9300, and the feed's err dialog), so only the last text shows; the others survive only in undelivered.jsonl and the log. Fix: aggregate in `_drop_parked_on_end`, not in the not-delivered path, whose other callers refuse one op each: one undelivered.jsonl row and one stderr line per text, then one err frame per End, the texts joined in typed order and kept distinguishable (a message can itself contain blank lines), a title whose noun covers messages and a command mixed, and op and itemId slots naming one request or none. A page-side queue would instead change the replace-the-older contract of `showConfirm` for every caller. Pre-existing: the feed's err dialog (ui/webview/feed.ts:6743-6763) never appends its element to the overlay, so by reading it is an empty backdrop (not verified in a browser); until then the feed shows neither text.

**The end route answers ok whatever it handed back.** kernel/kernel.py:68826 calls the hand-back and 68832 answers `{"ok": true}` regardless, so a shell caller of POST /end or romp end, often a peer session with no chat pane open, sees only ok. Fix: answer `{"ok": true, "undelivered": n}` only when nonzero, the deferred and queued idiom, so the nothing-parked pin in tests/test_kernel_headless_ops.py stays green. In bin/romp, branch inside the ok arm like the send verb's queued line (bin/romp:2012), keep exit 0, and name the count and the undelivered.jsonl location rather than echo the text into a peer session's transcript. Outside this fix, the deferred door (romp end self, --when-idle) answers before the self-close pass runs, and with no chat pane connected the frame reaches nobody (36502-36503), so the doc's claim at docs/codex.md:241-245 is true only with a chat pane open.

## Smaller points

**Roads where a typed send is still dropped silently.**

- The cancelCreate teardown (kernel/kernel.py:16680, kill at 16697) kills without the hand-back, so a send parked behind the account-limit gate under an Opening cue is dropped when the cue is cancelled; add `_drop_parked_on_end(sid)` after the history guard and before the kill.
- A send run the drain popped just before End gets no hand-back (the run is popped whole before the handover, 37585-37586, and the forwards branch, 37428, discards the send's False); read the False per op there and, for an op with the typed flag or a press id, take the not-delivered path to the one-chat target, and qualify the body's "kept by its slot" sentence.
- A drain-side backstop for a dead row's queue, which today goes to the unowned backend with the refusal ignored, belongs at the top of the sid's pass, not after the backend lookup at 37576 behind the gates at 37560-37572, keyed on a death record rather than on unowned alone (a restart-to-reattach window reads unowned too); latch the ending sid at park time so `_park_op_locked` refuses and the sending pane is told.

**Where the one-pane target lands.**

- `_send_to_one_chat` (36507) considers chat clients only, so a socketless End with a feed pane connected and no chat pane sends the frame to the empty chat broadcast, though the feed renders err frames; with no live chat client pick the first alive client whose app is in `_ERR_FRAME_APPS`.
- The pick at 36508 takes the oldest chat socket and `_client_send` (50200-50212) returns True on enqueue, so a chat pane whose peer went silent without closing takes the one dialog for up to about 40 s; pick by `max(live, key=lambda c: c.get("lastIn", 0))`, in the watching set as well as the fallback.
- The dialog's detail names the ended session by uuid, in `_refuse_drive`'s why branch (shared with the askFollowUp refusal, 19918), so a frame on a chat column showing another session does not name the ended one in a person's terms; use the registered name with the uuid as fallback, as moveFailed (20297) does.

**Stale wording about where the frame goes.** The hand-back docstring's first paragraph (36441-36442) is stale in both halves against the picker (36498-36512), and the kill route's comment (68826) and the route test's comment (tests/test_codex_compact_route.py:697-698) say the socketless frame goes to every chat pane, where the picker sends it to one and the broadcast is only the no-chat-client fallback both socketless tests drive; reword those and the self-close comment at 720-721, comments and names only (403-405 is correct).

**Typed slash commands are handed back by design.** The predicate at 36483-36484 takes every parked command with the typed flag, so a typed /clear raises the dialog, the stated design, and restricting the hand-back to sends would reopen the class the PR closes; what remains is that the same keystrokes on a Codex session park as a compact or clear op and give a silent log line, and that the docstring's "ops that carry no typed text" is inexact for a Codex clear op; a docstring sentence covers both.

**The bell's description of the kind gives one cause of three.** `DESC.undelivered` (62063) states only the wrong-kernel cause, so the chip and tooltip give a false cause for an End hand-back; write a cause-agnostic sentence (something sent never reached a session; the dialog said why; undelivered.jsonl keeps the text) and pin it in tests/test_error_center.py's node harness on the rendered chip title.

**Two doc statements.**

- The body says the three copies of the recovery sentence now say what happens, but PR 1933's body still reads that End then Revive delivers the waiting message; edit it to the merged doc's wording, or drop it from the list.
- docs/codex.md:245-246 lists "a message a script or romp itself sent" among what is dropped, but an untagged romp send comes back as the typed words, since the send route (kernel/kernel.py:25531) sets the typed flag from the absence of the romp-tag comment the CLI adds only on --tag; say only a tagged script send is dropped.

## Tests

- Every End hand-back test parks with both markers (tests/test_codex_compact_route.py:616 passes `user=True, qid=QID`), so the predicate's flag-only half and its command arm (kernel/kernel.py:36483-36484) are pinned by nothing, while the sendCommand door (19890) and an untagged romp send (25531) park with the flag alone; add two real-backend cases, a send parked with the flag and no id and a command through the live sendCommand door, asserting the second op's property (flag True, id None), not its spelling.
- The ordering claim at 36450 (hand-back before kill) is unpinned; patch the backend's kill to run the real kill then one drain pass, drive the three doors, and assert one err frame with the text.
- The two latches popped at 36474-36475 when the queue empties have no test; seed both (`km._mark_held_working(SID, time.monotonic())` for the held-working one), assert both gone when the queue emptied and present when the in-flight slot is kept, and pop the held-working latch in `_forget` too; no product defect at the merge commit.

PR #1971, opened 2026-09-21
romp compact --wait exits 1 with the notice's words when the compaction ended loudly instead of printing done

lczh:
Tier: fix

Follows the second executed review of the native compaction (#1933, merged as 492f163a). The branch was cut on that pull request's head, so the diff against main is this one commit.

## The defect

`romp compact --wait` printed done and exited 0 after a compaction that ended loudly. The wait polled only the row's compacting field on the sessions listing, and that field falls on every end, clean or failed: a compaction Codex could not run (a systemError status over the real backend) left the sessions row byte-identical to a clean end's. A scripted recycler pointed at that field, as the reference suggested, read success over a thread that was never compacted.

## The fix

The sessions row now carries the backend's launch error whole, beside its compacting bit, and the listing's rebuild key carries the notice's text and stamp, since a field whose input is not in the key is never refreshed. Both read it through one memo on the pusher cycle's scope, so the key and the row cost one backend read per session per cycle (the Claude Code backend's read is a registry file per session); outside a cycle every read is fresh, as the other cycle memos behave.

The CLI judges each poll's notice against a baseline and exits 1 with the notice's words on stderr when it differs. New against the baseline, not merely present: the compact verb does not clear a standing notice when it latches a new bracket, so a later clean wait on the same thread would otherwise fail falsely. The baseline is per path. For a compaction that runs at once, it is the row as the CLI read it before its request, so a loud end that lands before the first poll counts as this wait's rather than as a standing one. For a queued compaction, it is the first sample after the request that reads not compacting, the earliest ours could have fired: judged against the pre-request read, the prior compaction's loud end (or a failed turn's notice) was attributed to ours and a clean queued compaction exited 1. A compaction of ours that fires and fails before that first sample is baselined with it, since nothing tells it from a prior one's, and is reported by the timeout, never attributed. The notice is read whether or not a compacting sample was caught, since a compaction can run and fail between two polls. The baseline read is retried up to three times; when it never answers, the wait is refused on stderr after the request (the compaction was still requested; the wait cannot be judged from here), since a failed read taken as an empty baseline read a standing notice as this compaction's failure. The reference row and the help line name the failure exit and the baseline per path.

## Validation

Red first at the reviewed head, then green with the change, one module at a time:

- Headless shell module, the wait cases against a scripted fake kernel: the loud end after the compacting samples was red on the exit status (done, exit 0); the loud end between two polls was red on the notice's words (it waited out the timeout); the loud end before the first poll, against a baseline taken from that poll, was red on the notice's words (read as standing, it waited out the timeout); the queued compaction behind a prior one's loud end was red on the exit status (exit 1 over the prior notice); the queued compaction with a notice new since the arming sample was red on the words (it named the earlier notice); the queued loud end before the arming sample was red on the timeout words (it was attributed); the baseline read that failed once was red on the exit status (exit 1 with the standing notice's words); the baseline that never came was red on the refusal words (it judged blind); the standing-notice case guards the other side and passes throughout.
- Listing module, the new launch-error case: red with a missing field on the row; green after, with one rebuild when the notice appears, one when it clears, and one backend read per session per cycle.
- Kernel send module, the exact-dict pin on the sessions row: red at the head (the row carried a field the literal did not); the literals carry it now.

Module counts with the change: the headless shell file 32, the sessions listing module 11, the kernel send module 15, the kernel module 460 (+12 subtests), the perf stats module 65. The assertIn pin scan against the merge base finds nothing lost; it reads assertIn literals only, so the kernel send module's exact-dict pin, which it cannot see, was found by running that module and is named above.

Follows the second review of #1933.


ksarma:
The evidence below was executed against 21c8cebfe, the merge commit of this PR, whose second parent is the PR head f99d3a1e0; bin/romp, docs/reference.md and kernel/kernel.py differ between that head and the merge tree by the merges before it, so every file and line reference here is on 21c8cebfe. The first two points reproduce done, exit 0, over an uncompacted Codex thread; two under Smaller points do the same across a kernel restart and on a Claude session; the rest are wording and coverage.

- **An accepted turn erases the notice the wait judges** (bin/romp:1958; kernel/codex_backend.py:2683). Every accepted turn or start on the Codex backend sets the row's `launchError` to None (2683), racing the loud end's notice write and poke (1994-2056): with a message parked behind the compaction, the poke drains it and the accepted turn clears the notice within milliseconds (measured down to about 1.1 ms), before the CLI's next poll (bin/romp:1973, up to 2 s later), which reads quiet with no notice and prints done, exit 0, over an uncompacted thread. Fix: judge a record the clean end writes and an accepted turn does not erase, minimally a bracket-end record on the backend (`compact_ends`, already advanced by every end at codex_backend.py:1943, plus the last end's kind, text and stamp), carried on the /sessions row and in the listing key (kernel.py:28162, or it is never refreshed); the CLI prints done when the counter advanced past its baseline with a clean last end and exits 1 with the words when the last end is loud. The record needs a kind for every end, including the accepted-turn end at 2677-2679 and the clean end with no normalizer at 2030-2031, or the CLI reads those as neither; a compact_boundary stamp is weaker, since neither writes a boundary; the SDK backend's compaction path needs the same record before the CLI relies on it for every backend.

- **The listing pairs a fresh compacting read with the cycle memo's launch error** (kernel/kernel.py:28348). The row reads `compacting` fresh at 28342 while its launch error at 28348 comes from the cycle memo filled at the key's read (28177), so a loud end landing between the two yields (compacting False, launchError None), a clean end's shape. The bad listing is built only when another key input moves in the same cycle (or in the first cycle); then it is served for the rest of that cycle and the next one up to its listing stage, and a wait against a scripted fake kernel printed done, exit 0. Fix: a listing-scoped memo of the pair per sid per cycle, filled at the key's read in the order compacting then notice and served to the row, so a writer landing between the reads yields (True, notice), which bin/romp:1942-1943 already judges as a loud end. Keep `_launch_error` at one read per sid per cycle so the listing test's read-count pin stands, and scope the memo to the listing pair, not `_compacting_now` globally, which the chip, the drive-op gates and the drain read fresh. The test must move another key input in the same cycle; a quiet cycle serves the prior kept rows and passes regardless.

### Smaller points

- **The failure exit reads only the Codex bracket's notices** (docs/reference.md:48; bin/romp:90; kernel/sdk_backend.py:8457). The SDK bracket's loud end, a null status carrying `compact_error` (8457), clears the bracket and writes nothing to the row, so a failed Claude /compact reads as a clean end and the wait prints done, exit 0, while the reference row and the help line state the exit-1 clause with no backend qualifier under a row headed by Claude's `/compact`. Qualifying both (the failure exit reads the Codex compaction's end notices; a failed Claude /compact leaves nothing on the row) matches the merged code and touches at most the help-line pin at tests/romp-headless.bats:215; having the SDK end write a noRetry launch error instead needs its own design, since that record has other consumers (the idle-queue drive, the chat's error card) and clears only on a connect that disproves it, so a failed /compact would stand a red card and park the drive; that branch owes a red-first case in tests/test_autocompact_status.py and a bats case.

- **The refusal is stated without its path qualifier** (docs/reference.md:48; bin/romp:1915). The code refuses only when not queued (`[[ -z "$_c_queued" && -n "$_c_base_unknown" ]]`), a queued compaction taking its baseline from the arming sample (1877-1878) and judging correctly from it, so fix the sentence in the row and the PR body: refused for a compaction that runs at once; a queued one is not refused.

- **The help line's queued baseline** (bin/romp:90). It says new since the first sample after the turn, while the code takes the first not-compacting sample after the request, which lands mid-turn (1938-1939); reword the parenthetical to match the reference row: for a queued compaction, new since the first sample after the request that reads not compacting (no test pins the words).

- **The Codex bracket lives in memory only** (kernel/codex_backend.py:564). The registry load restores the launch error (695) but no bracket, so after a kernel restart mid-compaction the row reads compacting False with no notice and a wait that had seen compacting prints done over an outcome nobody recorded; fix on the registry: add compacting to `_save_registry`'s allowed set and the snapshot, save it at both latch sites (1916 and the worker's re-latch at 2509) and in the bracket end, and at load (beside 695) turn a row reading compacting True into a noRetry launch error worded as an unknown outcome, writing compacting False back so a second restart does not re-fire; the CLI needs no change (the load-time notice carries the mark), and a CLI-only alternative would fail every wait over a transient kernel blip, which bin/romp deliberately tolerates.

- **Two comments describe a contract the rows no longer follow** (kernel/sdk_backend.py:3114; kernel/kernel.py:28147). `REG_ROWS_FIELDS` and the rows revision still describe themselves as the registry fields the /sessions rows read and the revision their writes move, while the rows now read the launch error from the SDK registry per cycle outside the revision; amend both texts, since adding the field to the tuple removes no read and that launch error never carries a compaction notice.

- **The jobs cycle opens a launch-error memo nothing reads** (kernel/kernel.py:60058, 60072). `_launch_error_scoped` is unreachable from the jobs pass (deleting both lines passes 140 tests); drop them and amend the `_jobs_cycle` docstring, which says its scopes mirror the pusher's, or keep them for parity and fix the line comment at 60058, which names a reader the pass lacks.

### Tests

- The stamp half of the notice's identity is pinned by no test (tests/romp-headless.bats:238; tests/test_sessions_listing.py:145), though claimed at bin/romp:1866-1868 and kernel.py:28207: a stamp-blind CLI prints done, exit 0, when a standing systemError notice is followed by a compaction failing with the same words at a new stamp, while a text-only key alone yields no false done; red under both mutants, green at head: a fourth fixture notice in NOTICES with A's text and a new stamp, keyed by a single letter such as R (the fake reads one character at 260), a case `start_wait_kernel '{"ok": true, "queued": false}' 'q,c,c,qR' A` expecting exit 1 and NOTICE_A's words, and in the listing test a third record with the same text and a new stamp, asserting built rises by one and the new stamp is served.

- The refusal's scoping to the immediate path is pinned by no test (tests/romp-headless.bats:399): a one-line mutant at bin/romp:1915 refusing both paths passes all 34 cases; add `start_wait_kernel '{"ok": true, "queued": true}' 'q,c,c,q' '' 3` with a 30 s timeout, asserting status 0, the done line for busy1 and no "can't judge" (green at head, red under the mutant), alongside the reference sentence qualified as above.

- The body's figures against the merged head: tests/romp-headless.bats runs 32 at 7c61a3240 and 34 from f99d3a1e0, so the body's 32 is stale: the review-fix commit added the turn-notice cases at 359 and 371, which the red-first list does not name; the perf stats module collects 66 against the body's 65, the file untouched by the PR; and run at the base b06d857d8 with the head's cases, 305, 314, 322, 351, 371 and 399 are red while 340, 359, 379 and 389 are green there, pinning the design against a naive form of the change.

PR #1976, opened 2026-09-21
Design note: an optional, default-off tab state badge (alternative to the dashed rings)

henrypinkard:
## What

A docs-only design note, `plans/tab-state-badge.md`, for the user's sign-off before any code: an optional, default-OFF tab state badge, a small round dot in the state colour at a chat tab's top-right corner, as an alternative to the dashed outline rings. Desktop-only. Nothing on the tab moves and the tab's size does not change.

## Premises checked in code (cited in the note)

- The three ring classes and the widget registry that picks one, with precedence (`tab-state.ts` RING_ORDER, `tab-widgets.ts` composeTabRing; `styles.css` 593/601/555).
- The tab's flex layout, `position:relative`, no `overflow`, and the currently-empty top-right corner (`render.ts` renderTabs; `styles.css:523`).
- The loading swirl (left dot slot) and the peek tab, so the badge collides with neither.
- The phone switcher (`kernel.py` mobile CSS/JS), which has no per-session box and no ring, hence no badge.
- The gear setting model (`settings.ts` tabsLocked pattern), settingsSync fan-out (`gear.js` to `extension.ts` to `settings.ts`), and the settings preview (`gear.js` demoTab/paintWidgets; `gear.css` ring mirror).
- The contrast and colour-vision validator (`theme-parity.test.ts`: deltaE, cvdWorst, WCAG), reused rather than rewritten.

## Settled in the note

Badge size and placement (active, inactive, peek); one badge, no number; the ring precedence reused; the three state colours with the OKLab (>=15, >=8 under protan/deutan) and WCAG (3:1 large chrome) floors; the phone decision (no badge, with the reason); whether the hard-blocked red fill stays (recommended, re-gated); the setting's name, home, per-browser fan-out and preview; repaint-in-place on flip; and a red-first test plan (served lab, theme-parity, precedence, gear preview, phone).

## Open questions for the user

Listed at the end of the note: whether the ring stays available once the badge is judged, the red fill under badge mode, either/or versus both, the phone later, and the exact dot size.

Tier: docs. No code until the user signs off on the design.


ksarma:
The evidence below was executed against 7dfa52ee1, the merge commit of this PR, whose second parent is the PR head 4f69bc709; the merge carries the head's tree for plans/tab-state-badge.md, and every file and line reference is on 7dfa52ee1. The change adds that one design note: a default-off tab state badge, with Needs you as a top-right magenta dot carrying a count, retrying moved to the left status dot under badge mode, and Blocked unchanged.

- **The PR body describes the first draft of the note, not the merged one.** The body was written for the first commit, 7856cb65d; GitHub records no body edit since, and the six later commits rewrote the model. Nine body statements are reversed at 7dfa52ee1: awaiting sign-off before code (plans/tab-state-badge.md line 3: CONFIRMED, code follows); a dot per state in the state colour (line 5: a magenta dot for Needs you and nothing else); the ring precedence reused (line 5: no stack, no precedence); Desktop-only and no badge on the phone (lines 60 to 62: the same split on the phone); one badge, no number (lines 37 to 40 and 88: the numbered dot); the three state colours (line 56: two colours move); the red fill re-gated (line 52: no re-gating); Open questions (line 81: Decided); and a five-leg test plan (lines 72 to 79: eight legs). The stale text is not only the Settled and Open questions paragraphs: the What paragraph, the phone premise's "hence no badge" conclusion and the closing sign-off line carry the same first-draft model, while the Premises list remains true of the code. The body stays editable after a merge; a rewrite to the head's model closes the gap, and so does a one-line preface saying the body describes 7856cb65d and the merged note supersedes it.

- **The count's ink is fixed as literal black in both themes, and black fails the text floor on the light magenta.** plans/tab-state-badge.md lines 33, 39, 42, 78 and 88 name black digits; line 42 says the contrast pin adds black-on-magenta in both themes. The light theme's `--st-needs-bg` is #a21caf (ui/webview/styles.css:310); black on it is 3.32:1, under the 4.5:1 the parity test's pairs table applies to reading text (ui/webview/theme-parity.test.ts:61). The paired ink `--st-needs-fg` is #ffffff there, 6.32:1, and #2a0a2a in the dark theme (styles.css:228), 5.17:1 on #d946ef. Line 42 names no floor for the digit pin, and the note's only floor is the 3:1 at line 58 for the filled dots; a literal black in the light theme fails 4.5:1 whatever floor a follower picks, and digits inside an 8px dot cannot be WCAG large text. The paired token is already the text ink on the magenta (`.chip-needsInput`, styles.css:1600; plans/needs-you.md:135), a decision line 11 says the note does not reopen. The fix: ink the digits with `var(--st-needs-fg)`, name the token at all five lines, and say at line 58 that 3:1 is the floor for the filled dot while the digits take 4.5:1. Line 88 records the decision as black digits on magenta; the dark token is a near-black, so the approved mockup look is unchanged in the dark theme and only the light theme changes to white, which makes that line a decision to re-record rather than a wording fix. This is for the note first and later the code.

- **The retrying mechanism at plans/tab-state-badge.md line 46 fires in ring mode too, and the dot it adds has no title.** Line 46 says `tabDotClass` gains a retrying case. That function (ui/webview/tab-state.ts:133) is a pure function of the state string with no mode input, and its only non-test caller, the dot widget's render at ui/webview/tab-widgets.ts:201-209, receives only the widget's own options from `widgetOpts` (tab-widgets.ts:129 and 139, render.ts:6368); tab-widgets.ts never imports the settings module. Executed: with the literal one-line change applied, the function returns the retrying class under both mode values, so the case fires in ring mode against line 48 and test 2. `tabDotTitle` (tab-state.ts:146) still returns null for retrying, so the dot is visible with no hover title. A `blocked` status with none of the on-you flags gets the hidden slot while it wears the amber ring today: the ring's predicate is `tabStateClass` resolving to `tab-retrying` (tab-state.ts:62), which a flagless blocked also resolves to (tab-state.ts:28); the kernel emits blocked for an API error and retrying for a live retry (kernel/kernel.py:35883 and 35885). Three things for the note to state:
  - The gate. A mode parameter follows the file's own precedent, `tabRingId` taking its switch as a parameter (tab-state.ts:66) with an adapter over the prefs at tab-widgets.ts:168. The caller has no mode, so the note must pick one path: a dot-widget option, the one existing path that reaches both the strip and the gear preview (test 5), which puts the setting under Tab widgets rather than beside the tab lock as line 66 says; a save-time mirror of the boolean into the dot's options, on the pattern of the context-gauge mirror at tab-widgets.ts:13-16; or render.ts threading the boolean into a widened render contract.
  - The predicate. The case keys on the run-state class, not the raw state string, so the signature takes the status or the class; keyed on the string alone, a flagless blocked tab, amber-ringed today, gets no cue under badge mode.
  - The title. `tabDotTitle` gains a retrying case. Retrying has no feed twin (the feed's tips at ui/webview/feed.ts:756 cover work, await and unknown only), so it follows the opening pattern tab-dot-slot.test.ts:57-60 pins by shape: the state word, the feed's separator, an explanation.

## Smaller points

These three were read at 7dfa52ee1, not executed.

- **plans/tab-state-badge.md line 23's premise omits two cases:** `tabDotClass` returns null for compacting (tab-state.ts:134), so no dot and the bar takes the slot, and the dot widget's "Grey dot when idle" option (tab-widgets.ts:199-200) turns every hidden-slot state, blocked and retrying included, into a visible grey dot titled idle (tab-widgets.ts:204 and 207), so under that option a Blocked tab today wears a grey dot beside its red ring and line 52's "not a left-dot colour" and test 3 describe the default option only; add both cases at line 23.

- **Three line citations are stale:** kernel/kernel.py:60893 and :60929 at plans/tab-state-badge.md lines 23 and 62 point at JavaScript of the thaw queue, the `#mcur.ask` dashed-border rule being at 61058 and the row's left-bar rule at 61094, and extension.ts:385 at line 66 is a comment line, with `broadcastSettings` at vscode-extension/src/extension.ts:386; citing the two mobile rules by selector outlasts the next line shift, otherwise the three numbers want a refresh.

- **The containment argument is made for an 8px bare dot, but every Needs-you dot on a current page is digit-sized:** plans/tab-state-badge.md lines 29 to 31 argue containment for an 8px dot at a 2px inset and line 86 records that geometry as approved, while line 39 says a count of one shows "1", not a bare dot, and line 42 keeps the bare dot for an older page only, and the note gives the digit dot and the two-digit pill no size or font size and no containment check against the close glyph (styles.css:868) and the hot-key keycap (styles.css:621); state the digit dot's size and font size from a size the strip already uses and redo the containment paragraph for the digit dot and the pill, or record that the 8px figure is the older-page fallback only.

## Tests

- theme-parity.test.ts:88 already pairs `--st-needs-fg` on the needs magenta at a floor of 3; raising that row to 4.5 passes today in both themes (5.17 dark, 6.32 light), is the smaller change than a new row, and puts the pin line 42 names on the token pair, with test 7 at plans/tab-state-badge.md line 78 asserting the ink resolves to the token rather than to black.
- tab-dot-slot.test.ts:63-66 lists no retrying in its titled-iff-visibly-dotted loop, which is why the literal line-46 change stayed green; add retrying and, once a mode input exists, run the loop under both modes, retrying titled iff visibly dotted, which is only under badge mode.
- Tests 2 and 3 (plans/tab-state-badge.md lines 73 and 74) run under both idle options of the dot widget, so the ring-mode-unchanged pin covers the grey-idle option as well as the default.

